PR #7, opened 2018-03-12
adds admin ranks in ooc

Amelia0010:


AsV9:
consider mirroring this file to yogstation... something and disabling this file in the dme.

PR #8, opened 2018-03-12
Removes diagonal movement

ThatLing:
Like it says on the tin. Removes diagonal movement by only allowing one movement direction to be set at once. Behaved exactly like old-yogs code in my tests.

### DIFF
--- a/code/modules/keybindings/bindings_atom.dm
+++ b/code/modules/keybindings/bindings_atom.dm
@@ -2,17 +2,8 @@
 // Only way to do that is to tie the behavior into the focus's keyLoop().
 
 /atom/movable/keyLoop(client/user)
-	if(!user.keys_held["Ctrl"])
-		var/movement_dir = NONE
-		for(var/_key in user.keys_held)
-			movement_dir = movement_dir | SSinput.movement_keys[_key]
-		if(user.next_move_dir_add)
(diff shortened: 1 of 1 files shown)


Kmc2000:
why

GrayRachnid:
no

Jc1469:
as a player of yogstation, the only one who has commented here, I find this PR valid.  get the fuck out of here with your down thumbs

ThatLing:
@GrayRachnid 
Thank you for this constructive feedback.

@Kmc2000 
Because our current codebase doesn't have diagonal movement?

GrayRachnid:
I don't feel like that's a valid reason to remove diagonal movement, what is wrong with it?

ThatLing:
Ok let me turn the tables around, why shouldn't this be removed?

Bawhoppen:
Because there's no reason for its removal.

Partheo:
I don't think it's your call to remove this feature. If we are rebasing then the while point of this is to get new features. If the community hates it then remove it after- not before the rebase is even done.

ThatLing:
@Partheo 
It wasn't my intention to come across that way and I apologize if I did. That's the reason why this is a PR and not a straight commit to the branch. But I can see your point so I'll close this now. This can be re-opened in the future if the community wants it removed.

PR #9, opened 2018-03-13
STEALTHTONITE

Amelia0010:
https://github.com/yogstation13/yogstation/pull/1290/files but with a minor change.

It can now be any scan_state, including gibtonite. now you don't just have to be careful of uranium

image

This is all gibtonite


simonset55:
Stealth 2 nite

ShadowDeath6:
make it so that it's only uranium thanks :)

PR #19, opened 2018-03-14
Take these commits

ThatLing:
Closes #18 Closes #17 Closes #16 Closes #15


Partheo:
This doesn't do anything because we still have to rebase the repo, which will erase any changes we've made

PR #25, opened 2018-03-15
Ports LibVG

ThatLing:
Allows goonchat to display more characters, this isn't full UTF-8 but meh.

Source for LibVG isn't included but it can be found [here](https://github.com/vgstation-coders/vgstation13/tree/Bleeding-Edge/libvg).


Amelia0010:
image

PR #53, opened 2018-03-17
Ports donor features

ThatLing:
Ports donor stuff from Yogstation.

### Notes
- Purrbation looks different but that's out of scope
- Not all donor items are tested if they display correctly
- Quiet mode isn't tested completely
- Borg sprites aren't ported


Amelia0010 (dismissed):
Rainbow PDA:
image

Looks like engi PDA regardless of job, tested both on cook and doctor.

Transparent PDA:
unknown 1

I'm fairly certain transparent isn't supposed to mean invisible....
image
is what it's meant to look like

PR #62, opened 2018-03-19
fix air

Amelia0010:
Made in webmin©️ untested™️


Bawhoppen:
I would recommend actually doing this upstream instead. I think it would be appreciated as a feature there. (Do it fast though, feature freeze is in 2 days)

Bawhoppen:
Why the confused reaction? I just think that it'd be fine to port it upstream if you wanted, so you don't have to potentially deal with maintaining it downstream.

ThatLing:
woops

Amelia0010:
not like this'll need any maintaining anyways

PR #123, opened 2018-03-24
adds pretty filter

Amelia0010:
doesn't include the actual file, like on the current repo


Amelia0010:
I didn't port stuff like "test pretty filter" or "add to pretty filter" though, as it's just bloat

PR #286, opened 2018-04-06
Fixes fix-air

Amelia0010:
I just had an idea to make it much better


Amelia0010:
it works

ThatLing:
I'm not sure about this. Maybe give a choice whether to reset it to the initial gasmix or clean air?

Amelia0010:
but y tho. 🤔

PR #298, opened 2018-04-09
adds paramedic, mining medic and tcomms lockers

Amelia0010:


Amelia0010:
okay i'm just gonna self-merge, in the interest of the rebase

PR #303, opened 2018-04-11
Create Space_Station_13_areas.dm

Amelia0010:


Amelia0010:
AKA gray also needs these areas, so I can't just keep them in my own PR

PR #416, opened 2018-04-17
Remove mineral access from miners

Amelia0010:
see: https://github.com/yogstation13/yogstation/pull/3063/files


GrayRachnid:
Was there a reason? Out of curiosity.

Amelia0010:
Well, the PR was made shortly after miners doing RnD was declared powergaming officially, so it was mostly just an IC way of preventing that. That, combined with stuff like this https://forums.yogstation.net/index.php?threads/miners-expect-gear-for-doing-their-job.16487/, not to mention the general lack of materials that any non-science department feels :)

Jc1469:
reminder the rule states they can do it if scientists aren't, which this actively impedes.

Amelia0010:
if there aren't any scientists to do it, then it's most likely lowpop enough that they get extended access, in which case they still can.

Alexandria404:
Love it.

Oakboscage:
Great change, council approves.

PR #483, opened 2018-04-24
Revert "Remove mineral access from miners"

Notamaniac:
Reverts yogstation13/tg-rebase#416

You're reminded that to bypass this, all the miner needs is the mythical crowbar and screwdriver. Meaning this change is completely and utterly pointless in actually stopping miners doing what they want with their minerals.

Oh and that even if miners lost their access, it still isn't likely that the other station roles are going to get minerals because the R&D scientist is still going to dash to the ORM and pick up the minerals or the QM will hoard the minerals to sell. (And you're not going to be able to convince them to spare minerals if they're already dead set on using them)

(And before the obligatory "if this change accomplishes nothing why are you reverting it?", if this change accomplishes nothing, why are you merging it?)


Amelia0010:
🤔

Amelia0010:
if you think the change accomplishes nothing, and I think it'll accomplish good things, then how can merging it be bad, and reverting it be good 🤔

Amelia0010:
no https://forums.yogstation.net/index.php?threads/vote-miner-orm-access.16557/

PR #503, opened 2018-04-26
Adds rebindable controls

ThatLing:
Hey it's actually finished now, isn't that cool. Allows you to rebind your controls to *any* key you want. Controls can be changed in-game, an action can only be bound to 1 key.


code/__DEFINES/~yogs_defines/keybindings.dm
```diff
@@ -0,0 +1,186 @@
...
+	"F12",
+	"1",
+	"2",
+	"3",
+	"4",
+	"5",
+	"6",
+	"7",
+	"8",
+	"9",
```
Amelia0010, line 51: you forgot 0

Amelia0010 (dismissed):
Other than that I can't actually spot something by eyeballing, gonna go test it out now

PR #524, opened 2018-04-27
Adds beurocratic_error to the dme and modularizes a weight change.

GrayRachnid:
No running away from #512 nich.


Amelia0010 (commented):
I mean I guess 😢

PR #555, opened 2018-04-28
Default hotkeys to on

ThatLing:
:)


Amelia0010:
this is probably a good thing

PR #628, opened 2018-04-30
Removes telescience.

GrayRachnid:
Asv ported the the old /tg/ telescience instead of the yog once, he also did not add it to RnD, he also added some weird beacon thing that doesn't even work. Not to mention the UI has some issues and it's buggy.

I'll be re-adding yogs features to Quantum pads, something already in the code and is being maintained by /tg/. The way Quantum pads works is that you can teleport in between pads by standing on a pad, i'll add the ability for pads to be able to send to beacons that can be printed from RnD.


ShadowDeath6:
this is a good way to go about it :+1:

Amelia0010 (approved):
rip AsV's only contribution

AsV9:
maan I'm not sure what happened to the UI, it worked fine when I ported it

PR #715, opened 2018-05-05
Electric Rollerbeds

Amelia0010:
Sprites by @Partheo 

Adds electric rollerbeds, for when you want take-out executions.
image
image

🆑Nichlas0010, sprites by Partheo 
rscadd: Electric rollerbeds can now be constructed, the same way you'd do with an electric chair, but with a rollerbed!
/🆑


Nickvr628:
I support it. I cannot wait for the mobile execution squads that roam the station.

Partheo (dismissed):
Make it require an electropack to craft, not a shock kit. The shock kit has a helmet, which isn't included in the roller bed.
I think it would look great if you could use the flashing sprite, but if that's not feasible then it's alright.
Tested it and everything else seems fine.

ghost:
thanks mike pence for this well developed feature

PR #717, opened 2018-05-05
doesn't remove his_grace from the uplink

Amelia0010:


ghost:
make it hijack-only, so that the chaplain can still use his grace for hijacking, but not for your typical murderbone tool

ThatLing:
@AdamElTablawy If you're gonna code restricting uplink items by objectives, go ahead.

Amelia0010:
>make it hijack-only, so that the chaplain can still use his grace for hijacking, but not for your typical murderbone tool

The problem with this is;

Is it based on the objective of the person who's using the PDA? Is it based on the objective of the owner of the PDA? Everything else is based on the owner, but that still allows other people to kill him and then buy it, and then we're back to square one. If you want to do the former, then you have to rewrite the entire get_uplink_items() proc, since it's completely independent of the user as is.

Kmc2000:
It shouldn't be that hard if you want to go that route to be honest.

ghost:
https://github.com/ParadiseSS13/Paradise/pull/4395/files

take note of the hijack-only var

wouldn't be too hard

also, if someone kills the chaplain with his uplink open then buys his grace to murderbone with it that's a really edge case

ghost:
thats so edge case i could cut myself on it

ghost:
@Qustinnus 

so in line with the quality of yog code in general then

ThatLing:
what could go wrong

PR #735, opened 2018-05-06
bigotry rule: A new hope!

Amelia0010:
completely forgot about this


Alexandria404:
What does that do.
I get the meatball thing, which I personally find stupid, but what does the speech thing do.

ghost:
there's things like brain damage that replace random words, nich just replaced any of the /mean/ words

Amelia0010:
yeah but they're **mean**

ThatLing:
hugbox

ThatLing:
oh god i should've squashed that

PR #749, opened 2018-05-09
Sprites: old stacking material sprites

Partheo:
This
materials - copy
Didn't do too much testing cause I was rushed, it should be exactly back to how it worked.

Mirrored a shit ton of TG files for this because I had to make some newlines to specify the proper icon files. I added entries to changes.md for posterity.

:cl:  
imageadd: Metal, Plasteel, Glass, Reinforced Glass, and Wooden Planks are now back to their old stacking sprites.
/:cl:


Amelia0010:
I'll look at this in a few hours, but wouldn't it be easier to just mirror the specific procs/definitions you need instead of the whole files?

ThatLing:
What nich said

Partheo:
Let's say I mirror update_icon() in the Yogstation folder, how would I tell byond to use my proc instead of the one in the TG code?

Amelia0010:
It'll use the one defined last, which would be the one in the yogstation folder as Y is after the C in code

Amelia0010:
however, calling ..() would then call the previous proc, IE TG's update_icon

ghost:
thisll look bad without you respriting the other sheet materials

Partheo:
I'll take a look at the procs tonight and see if that works better.

@qustinnus I just realized this breaks the other stacks, I'll work on respriting those and getting this updated before it goes any further.

PR #770, opened 2018-05-10
Meteor probability switchup.

AsV9:
Reduces overall probability of meteors spawning while simultaneously lowering the requirement for number of players. This reverts meteor behavior to how they used to work before the rebase, which I frankly think is a superior way of going about spawning a nearly-always game ending meteor wave.

hugbox amirite

### DIFF
--- a/code/modules/events/meteor_wave.dm
+++ b/code/modules/events/meteor_wave.dm
@@ -4,9 +4,8 @@
 	name = "Meteor Wave: Normal"
 	typepath = /datum/round_event/meteor_wave
 	weight = 4
-	min_players = 15
+	min_players = 5
 	max_occurrences = 3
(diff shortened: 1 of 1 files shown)


AsV9:
hey you little shit

Partheo:
Good change, although without the minimum time requirement, is it possible that meteors arrive right at the beginning of the shift? I would allow 5 minutes to go by first, at least.

lol @ nich

AsV9:
Default earliest start is 20 minutes into the round. Unless you explicitly set the earliest start to earlier, it can't happen before then. It inherits from `/datum/round_event_control/` which you can [find](https://github.com/yogstation13/tg-rebase/blob/master/code/modules/events/_event.dm#L11) in
`tg-rebase/code/modules/events/_event.dm`.

PR #793, opened 2018-05-12
readds weightless event

Amelia0010:
see: https://github.com/HippieStation/HippieStation/pull/7375/files


ThatLing:
why

Amelia0010:
because it's something we had on old-yogs, and this implementation is better than my old one

Amelia0010:
god dammit

PR #799, opened 2018-05-12
Buffs AI doorknob.

AsV9:
You can now alt click a door to request the AI open it, but only every ten seconds to prevent spam.

That's about it.

HippieStation/HippieStation#3898 btw


AsV9:
i realized I did an oopsie, don't merge this before I fix

Kmc2000:
MANE HOW HARD IS IT TO DO MOD U LAR CODE

yogstation/code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -0,0 +1,8 @@
+/mob/living/silicon/ai/proc/handle_remotedoor(href_list)
+	var/obj/machinery/door/airlock/A = locate(href_list["remotedoor"]) in GLOB.machines
```
1fbff5f83b23d39d38b1dfcb4cac8d9b, line 2: Href exploit here, add `if(src != usr) return` to check if it was the AI who called this

1fbff5f83b23d39d38b1dfcb4cac8d9b (changes requested):

yogstation/code/game/machinery/doors/airlock.dm
```diff
@@ -0,0 +1,28 @@
...
+	if(!user.canUseTopic(src))
+		to_chat(user, "<span class='info'>You can't do this right now!</span>")
+		return
+
+	if(stat & (NOPOWER|BROKEN) || (obj_flags & EMAGGED))
+		to_chat(user, "<span class='info'>The door isn't working!</span>")
+		return
+
+	if(request_cooldown > world.time)
+		to_chat(user, "<span class='info'>The airlock's spam filter is blocking your request. Please wait at least 10 seconds between requests.</span>")
```
ShadowDeath6, line 15: I would remove all of the feedback here except for the one that says request sent when it is successful, I don't think it's needed to display a message for every single thing

Amelia0010:
open it if you're actually working on it

PR #1158, opened 2018-06-05
Removed multicamera telegraphing

Cruix:
Using multicamera mode no longer makes nearby cameras turn red.


Notamaniac:
but balance

and scaring off people from touching areas despite the fact that you weren't using multicam mode to begin with

Amelia0010:
ignore travis failing, it's a changelog error which I fixed on master

PR #1257, opened 2018-06-13
Adds include/exclude objective system

Amelia0010:
Instead of having a hijack-only var, we can now include/exclude items based on the objectives of the buyer. If a user has both an objective that's necessary, and an objective that blocks it, the blocking takes precedence. 

I also moved his grace over to this system, and am planning on moving romerol, but pR aToMiZaTiOn

No changelog cause it doesn't affect anything in-game **YET!:tm:**

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1350,7 +1350,6 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	cost = 20
 	restricted_roles = list("Chaplain")
 	surplus = 5 //Very low chance to get it in a surplus crate even without being the chaplain
-	hijack_only = TRUE //yogs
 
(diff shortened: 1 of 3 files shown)


fluffe9911:
But nichhhh romeio is 25 TC which means no matter what someone who doesnt have a hijack objective would have to give 5 tc to someone who does which is already rare enough as it is this would turn a item used once every 2-3 weeks to a item used once every 3-4 months if you are lucky reeeee

Amelia0010:
>But nichhhh romeio is 25 TC which means no matter what someone who doesnt have a hijack objective would have to give 5 tc to someone who does which is already rare enough as it is this would turn a item used once every 2-3 weeks to a item used once every 3-4 months if you are lucky reeeee

>and am planning on moving romerol, but pR aToMiZaTiOn

fluffe9911:
But nichhhhhh im illiterate reeeeeeeeeee

PR #1288, opened 2018-06-15
makes OOC check for admin-channels

Amelia0010:
basically, moves the fakekey and AFK handling from yogbot into the server, and adds a check as to whether or not it's from an adminchannel. The adminchannelbit needs to be implemented yogbotside though, so yeah :^)


nichlas0010:
https://github.com/yogstation13/yogbot/pull/100

PR #1365, opened 2018-06-19
Remove ORM access from conscription kits

### LINKED ISSUE #1333: conscription kit mining ID has cargo/ORM access, while normal miners don't
title says what

ThatLing:
Fixes #1333

### DIFF
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -267,7 +267,7 @@
 		var/obj/item/card/id/I = AM
 		I.access |=	ACCESS_MINING
 		I.access |= ACCESS_MINING_STATION
-		I.access |= ACCESS_MINERAL_STOREROOM
+		//I.access |= ACCESS_MINERAL_STOREROOM // yogs - Remove miner ORM access
 		I.access |= ACCESS_CARGO
 		to_chat(user, "You upgrade [I] with mining access.")
 		qdel(src)


drderp3635:
Just give miners orm access

drderp3635:
there is litterally no reason for miners to not have orm access anymore because they have their own lathe to build things with, but they can't even load it. The worst that can happen from miners having ORM access is that they shove all the ores into their lathe which is just as accesable as it being in the ORM, if not more.

Moja10:
More ORM stupidity, what's the reasoning for this PR exactly?

ThatLing:
@drderp3635 
God forbid miners have to talk to a cargo tech to take the minerals out.

@Moja10 
Look at the issue in the description.

drderp3635:
@thatling just remove all access that normal jobs have and give assistants all access, because god forbid anyone has to ask an assistant for help

Amelia0010:
https://en.wikipedia.org/wiki/False_equivalence

That said, could just let this be and let the miners "buy" access to the minerals :^)

MacHac:
>miner
>come back with lots of minerals
>want that sweet advanced plasma cutter
>lathe is empty
>go to retrieve smelted minerals
>'Access Denied'
>cargo area deserted, techs are off stealing or being cultists or something
>fuck this

And that's why half of rounds go completely without minerals despite having three different miners.  Of course, in the other half of rounds the miners just deconstruct the machine to get the goods.

As far as I can tell from combing through the forum drama, the original impetus for locking the shaft miners out of the ORM was some paranoid fantasy that the miners were stealing every single mineral they mined and turning it all into hundreds of satchels of holding or something like that, leaving none for the poor scientists.  

Ironically, through the manual smelting machine on the mining base, the current system directly encourages that sort of behavior and generally discourages putting any effort into mining once you have what you need.  This in turn means science is less likely to spend their points on mining tech, which further disincentivises mining.  

While this is not the only reason the station only sees a single diamond every third or fourth round, it is a major factor, and it's what caused me personally to stop mining as a shaft miner and focus on exploration like everyone else.

Really, ALL the drama surrounding the ORM needs to be reexamined for our new post-department lathe world, where a great many people have reason to want minerals and incentive to dismantle the cargo office walls to get to them.  Everyone who has authority and cause to print something from a lathe should have direct access to minerals, which is how it's going to work anyway if TEE GEE ever gets around to merging that new resource distribution tech.  

Trying to force people to ask permission works well if the cargo office is manned, but falls flat on its face when the cargo techs remember that they're basically assistants with free insulated gloves and maint access.  This is the same reason that 'ask the HoP' is a bullshit argument, as any round where the HoP both exists and is still at his desk five minutes in is a very special round indeed, and should be viewed as an exception instead of a rule.

tl;dr: Best case scenario, forcing miners to ask permission makes the process take five seconds longer.  Worst case scenario, it's actively killing the miners' desire to do their jobs.

PR #1428, opened 2018-06-24
secures tech storage

Amelia0010:
Secure tech storage now has a motion camera, as well as the door being bolted by default, and being as secure as the vault door. This was the result of a discussion on why obtaining an AI law board was so trivial, in #development-public

#### Changelog

:cl:  
tweak: Secure tech storage is now as secure as its name implies.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -4315,6 +4315,18 @@
 	},
 /turf/open/floor/plasteel,
 /area/security/processing)
+"akd" = (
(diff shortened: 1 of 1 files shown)


Nickvr628:
I approve. It is a good change and should fix some of the issues with AI subverting.

ThatLing:
Not secure enough, -1

Rayisken:
Yeah, this was needed considering it takes a few seconds as QM/HoP to print a RCD and get in through maint. I was wondering when it would be made harder

PR #1456, opened 2018-06-26
Re-adds borg tools

Amelia0010:
see Icon-diff-bot for the icons.

Re-adds the borg tools and laser we used prebase.

#### Changelog

:cl:  
rscadd: Borg tools now look like borg tools, instead of tools.
/:cl:


Notamaniac:
Pls no

The old borg tools were awful to try and figure out which was which without mousing over the ; the standard ones you can tell by a glance "this is a screwdriver, and not a extremely complex overly advanced crowbar"

Amelia0010:
image
this does not look like a crowbar.

Notamaniac:
It doesn't look like a screwdriver either. Could be a multitool. Heck might be a mini taser for all I know.

image

This is a wirecutter. Could you tell by first glance?

Without being able to mouse over the sprite in byond, I just can't tell by a glance WHAT that tool is.

Amelia0010:
The icon looks like a pair of scissors, so yes I can tell they’re wirecutters. 

Maybe the problem isn’t with the sprites, but with you

Notamaniac:
Good you can exercise your keen powers of deduction. Because that's not the actual wirecutter sprite. This is.

image

That was a crowbar I posted previously. Do you NOW see the problem here? Or maybe the problem isn't with the sprites, but with you.

Amelia0010:
mousing over is apperently too much to ask

Nickvr628:
Re open this, they are nice sprites and they look a lot better than the human tools in the borg inventory.

Amelia0010:
When you decide to provide literally no input at all, not even doing as little as adding a little 👍 reaction, then it's too late to comment after it's already been closed.

Notamaniac:
Yes, they LOOK better there's no doubt about that.

But in the heat of the moment, when a borg needs to reach for a multitool, they don't need to mouse through 5 vaguely indistinct borg tools before finally landing on the right one. 

Yes, it's too much to ask to mouse through your inventory of tools when you need grab that crowbar to escape out the unpowered airlock because there's someone hostile closing in with a flash.

Amelia0010:
I sincerely hate everyone

PR #1472, opened 2018-06-27
nerfs chainsword

Amelia0010:
nerfs chainsword from 30 to 20

#### Changelog

:cl:  
tweak: chainsword has been nerfed from 30 force to 20.
/:cl:


ThatLing:
Reason for nerf?

Amelia0010:
it's 30 damage, which means it's only 4 hits to crit if they have no armour, and as much as an Energy sword, except it's free, and possibly has a ghost inside of it which can also act as life insurance, so they can yell if you get fucked :)

Smudged:
We could make it 16 damage, 1 more than the null-rod because even at 20 it feels like it's still gonna be a first pick for valid hunting chaplains

Ratvar:
@Smudged , nullrods have 18 damage. And there's no reason to make it 20 because it's scythe resprite that talks - it's already better, sneaky Nich. And it's also used by paranormal ERT. And there's also a new PR on TG to nerf it right now, so there's no need in this one at all.

ThatLing:
#1497

PR #1486, opened 2018-06-28
Allow cargo protolathe to craft Flight Suit

Apologistic:

Nearly every round, someone from mining breaks into Engineering to use the engineering protolathe to craft a flight suit. It's a powerful, albeit expensive tool, just with the added annoyance of having to get into engineering to make it. The only reason I think this might need further tweaking is to increase the material cost. If cargo gets manipulator upgrades, and their miners are competent, there could be a few of these running around. Maybe that's not a bad thing late game.

#### Changelog

:cl:  
tweak: Allow cargo protolathe to craft flight suit
/:cl:

### DIFF
--- a/code/modules/research/designs/equipment_designs.dm
+++ b/code/modules/research/designs/equipment_designs.dm
@@ -7,7 +7,7 @@
 	materials = list(MAT_METAL=16000, MAT_GLASS = 8000, MAT_DIAMOND = 200, MAT_GOLD = 3000, MAT_SILVER = 3000, MAT_TITANIUM = 16000)	//This expensive enough for you?
 	construction_time = 250
 	category = list("Misc")
-	departmental_flags = DEPARTMENTAL_FLAG_ENGINEERING | DEPARTMENTAL_FLAG_CARGO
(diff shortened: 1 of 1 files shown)


Amelia0010 (approved):

MacHac:
Why are flight suits engineering-only anyway?  Engineering has hardsuits.  Flight suits (with their expanded vision range and speed boosts) seem designed for space exploration, which is traditionally either science or whatever greytider manages to get EVA access.  Scientists in particular are the only ones I ever see using the things (not that J. Random Engineer knows the department lathe exists), so if anyone should get access to prevent inevitable breakins it should be them.  

Sure it's powercreep, but breaking into engineering is such a joke anyway that half the (non-antag) assistants will be in and out in any given round; Keeping it engineering-locked just adds a small time-gate.

I'm also not sure what you think cargo's going to do with them.  If anything, this just means the scientists will now print their flight suits in cargo when they're already breaking in to get minerals.  Cargo comes in right behind the civilian departments for 'least reason to need one of these things', least of all because they already have a reliable source of EVA gear.

Apologistic:
I agree. Perhaps it should be Cargo || Engineering || Science - but I was adding Cargo because miners, which makes the most sense to me for the flight suit - lavaland is where it shines.

ThatLing:
This could've been modularized but I'll let it slide since it's only a few lines.

PR #1542, opened 2018-07-01
Removes par from codeowners

Amelia0010:
doesn't review things anywho, and he can always ask for it to be uncommented. Also removed the /icons/ since only /tg/ uses them, and they get merged anyways


GuppyLaxx:
How could he

PR #1595, opened 2018-07-03
Quack.

GuppyLaxx:
### Adds Immovable Ducks.
About as likely to happen as meaty ores, and can only happen once a round.
@Kmc2000 didn't wanna give me my sprites so i just grabbed a duck from google and downscaled it to 32x32

untitled
**quack.**


jcatjmeow:
speedmerge pls

PR #1632, opened 2018-07-07
Makes assistant have the same antag rep weight as all other jobs

Amelia0010:


Amelia0010:
i'm not council to re-enable this anyways, so it's not my change to make

PR #1649, opened 2018-07-09
Integrated Circuits Nerf

CMOisLing:
Increases the size, complexity, cooldown, and power usage of throwers. Making it much harder to make a effective glass thrower so they cant kill you in less than a second now. Also slows the movement of locomotion so bomb bots are less threatening.

:cl: CMOisLing
tweak: Increased cooldown for locomotion and thrower circuits
tweak: Increase size, complexity, and power usage of thrower circuits
/:cl:


Amelia0010:
As much as disabling features is a bad way to balance them, it usually gets people to come up with good ways to balance them afterwards :)

ThatLing:
I'm gonna leave this open until the council election is over so we can discuss what needs to be changed for circuitry cloning to be re-enabled.

Hate9:
I approve of this type.of change, but isn't 45 complexity and 5 times the cooldown a bit much?
Either way, props for doing something.

CMOisLing:
I agree the nerf is a bit much I can tone down on it when I get my PC back, but I’ll see what council wants before messing with things. I wanna keep the increase on complexity and size because most of the annoying throwers use multiple throwers to deal massive damage before you can react making them unfun imo

CMOisLing:
Ow heck I might have fucked with this PR on accident

CMOisLing:
Don’t merge until I fix my shit

CMOisLing:
Alright ive got my shit together its safe to merge now

PR #1698, opened 2018-07-11
Nice.

GuppyLaxx:
@nichlas0010 
this has been spawning flashlights instead every time, if you wanna keep surplus arms make an item that on pick up gives you another arm


Amelia0010:
I forgot about that

PR #1788, opened 2018-07-18
Fixes logout sync

ThatLing:
Also qdels the query which is needed


ThatLing:
Needed before #1764 can be merged

PR #1867, opened 2018-07-23
Adds in a new engine

Kmc2000:
Welcome to the future of power! With the new NT 'STE-VENS' lineup of nuclear reactors, generating power becomes trivial.

This pr adds a nuclear reactor to the station, it's only present on Yogstation's own specific map to avoid /tg/ conflicts.

Demo video of me setting it up, and then making it meltdown:
https://drive.google.com/file/d/1b5vG_vk1VgcxlJL-2ZVotGI8ntKxItnK/view?usp=sharing

FEATURES:
A nice, simple engine to operate, taking about 2 hours to go critical when properly setup (even if completely ignored) or about 8 minutes if you just shove fuel in and leave it without balancing the heat generation.

Tutorial
https://forums.yogstation.net/index.php?threads/nuclear-reactors-and-you-how-to-not-be-an-atomic-playboy.17154/

How does it work?
It takes in a gas of your choosing (Plasma is good) and superheats it in a specialized loop, which is then vented into the main power stage (turbines) which spool up to about 30 KW each when properly set-up. After you've reached a comfortable heat (I usually like 300 * or so) you can then balance out the fuel with control rods and leave it for a bit, but do check back on it occasionally, as fuel will deplete, which will slow down the reaction. This took about 35 minutes to happen on my local branch, running it at medium temperatures.

https://i.gyazo.com/c8ddba043cb9664343adf342aa191dc8.mp4

What happens when it goes critical?
You are given an audible warning as the temperature rises to critical levels, giving you about 5 minutes to fix it (when using a balanced setup) or about 30 seconds on an imbalanced one. After this stage, when it hits 1000* it hits the point of no return, giving you less than a minute to evacuate engineering and grab any critical supplies you can, after that, it maxcaps engi and spills radioactive waste all over (which will render engineering permanently uninhabitable, even with suits) You will need to contain the burnt out reactor core and the waste Chernobyl style to avoid major contamination. 

https://i.gyazo.com/75065820a8431f8c0a45d065dcd1e1c0.mp4

Credits:
Pipes: Partheo (modified by myself)
Meltdown sound: Portal 2
Ideas + General mechanics: Partheo

TODO:
Live test for balance tweaks
Mapmerge

#### Changelog

:cl: Kmc2000
rscadd: Added in a nuclear reactor to Yogstation's own brand map. Check it out now! 
/:cl:


ThatLing:
Map merge

yogstation.dme
```diff
@@ -2714,6 +2714,7 @@
 #include "yogstation\code\modules\mob\living\silicon\robot\robot.dm"
 #include "yogstation\code\modules\mob\living\simple_animal\hostile\megafauna\colossus.dm"
 #include "yogstation\code\modules\power\lighting.dm"
+#include "yogstation\code\modules\power\NuclearEngine.dm"
```
ThatLing: All your files should be lowercase, this includes sound files and icons.

code/modules/atmospherics/machinery/pipes/manifold.dm
```diff
@@ -37,15 +37,15 @@
 	//Add non-broken pieces
 	for(var/i in 1 to device_type)
 		if(nodes[i])
-			add_overlay(getpipeimage('icons/obj/atmospherics/pipes/manifold.dmi', "manifold_full[invis]", get_dir(src, nodes[i])))
+			add_overlay(getpipeimage(icon, "manifold_full[invis]", get_dir(src, nodes[i])))
```
ThatLing: Why? Not modular.

code/modules/atmospherics/machinery/pipes/manifold.dm
```diff
@@ -37,15 +37,15 @@
...
 
 //Colored pipes, use these for mapping
 /obj/machinery/atmospherics/pipe/manifold/general
 
 /obj/machinery/atmospherics/pipe/manifold/general/visible
 	level = PIPE_VISIBLE_LEVEL
 	layer = GAS_PIPE_VISIBLE_LAYER
-	
+
```
ThatLing: Jesus christ

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+#define HIGH_HEAT 500
+#define CRITICAL_HEAT 800
+#define MELTDOWN 1000
+
+//NOTES!
+/*
+According to players, the average usage is 160 KW, or 160,000 watts. So that's the lowest target to hit. Right now it hits around 4000 watts
+*/
+
+/obj/machinery/power/NuclearReactor
```
ThatLing: I don't like the capitals but it's fine

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+	state |= WASTEHATCH_OPEN
+	connect_to_network()
+//	outlet = new(src.loc)
+	var/obj/dummy = new(src.loc) //Handle placing pipes properly
+	dummy.alpha = 0
+//	outlet.name = "[src] gas outlet port"
+//	outlet.dir = NORTH
+	dummy.y += 1 //1 tile above our loc, to get outlet!
+	outlet = locate(/obj/machinery/atmospherics/components/binary/pump/nuclear) in dummy.loc
+	qdel(dummy)
```
ThatLing: ??

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+	STOP_PROCESSING(SSmachines,src)
```
ThatLing: Why?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+
+/obj/machinery/power/NuclearReactor/proc/CheckPipes() //In case your pipes blow up
+	outlet = null
+	outlet = null
```
ThatLing: D-d-d-double kill

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+	var/obj/dummy = new(loc)
+	dummy.alpha = 0
+	dummy.y += 1
+	outlet = locate(/obj/machinery/atmospherics/components/binary/pump/nuclear) in dummy.loc
+	qdel(dummy)
```
ThatLing: Why?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+	if(Heat <= 0)
+		stop()
+		return
+
+/obj/machinery/power/NuclearReactor/proc/MeltdownAnnounce() //Announce the meltdown to give them a chance to RUN
+	stop()
+	for(var/mob/M in GLOB.player_list)
+		if(M.z == z)
+			M << null //Don't want to assault their eardrums
+			M << MeltDownSound
```
ThatLing: playsound

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		return
+	if(Heat >= CRITICAL_HEAT)
+		radiation_pulse(src, Heat*2, 60)
+		if(world.time >= SavedWarningTime + WarningSoundCooldown)
+			SavedWarningTime = world.time
+			for(var/mob/M in GLOB.player_list)
+				if(M.z == z)
+					M << WarningSound
+		else
+			return
```
ThatLing: ??

Keekenox:
MAP MERGE  AAAAAAAAAAAAAAAAAAAAAAAAAH

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		stop()
+		MeltdownAnnounce()
+		return
+	if(Heat >= CRITICAL_HEAT)
+		radiation_pulse(src, Heat*2, 60)
+		if(world.time >= SavedWarningTime + WarningSoundCooldown)
+			SavedWarningTime = world.time
+			for(var/mob/M in GLOB.player_list)
+				if(M.z == z)
+					M << WarningSound
```
ThatLing: playsound

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		state &= ~WASTEHATCH_OPEN //Bitflags! We add the OPPOSITE of wastehatch open to "state", in other words, inverting it, if the hatch is open, this closes it
+		playsound(loc, 'sound/effects/bin_close.ogg',50,1)
+		to_chat(user,"You shut the access hatch")
+	else
+		state |= WASTEHATCH_OPEN
+		playsound(loc, 'sound/effects/bin_open.ogg',50,1)
+		to_chat(user,"You open the access hatch")
+	UpdateIcon()
+
+/obj/machinery/power/NuclearReactor/proc/UpdateIcon()
```
ThatLing: ``update_icon()``

Nickvr628:
I love it.

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+					M << WarningSound
+		else
+			return
+
+/obj/machinery/power/NuclearReactor/proc/ProcessAtmos()
+	for(var/datum/gas_mixture/S in outlet.airs)
+		var/num = Heat
+		if(num <= 0)
+			num = 0
+		S.temperature += num
```
ThatLing:
```
/obj/machinery/power/NuclearReactor/proc/ProcessAtmos()
	if(Heat <= 0)
		return

	for(var/T in outlet.airs)
		var/datum/gas_mixture/S = T
		S.temperature += Heat
```

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+
+/obj/machinery/power/NuclearReactor/proc/stop()
+	STOP_PROCESSING(SSmachines,src)
```
ThatLing: But you added this in ``/Initialize()`` aswell?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+#define HIGH_HEAT 500
+#define CRITICAL_HEAT 800
+#define MELTDOWN 1000
+
+//NOTES!
+/*
+According to players, the average usage is 160 KW, or 160,000 watts. So that's the lowest target to hit. Right now it hits around 4000 watts
+*/
+
+/obj/machinery/power/NuclearReactor
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: Not really a fan of CamelCase where most (all?) paths use snake_case

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+	var/DepletedAmt = 0 //How much of the uranium has been depleted, and thus acts like control rods?
+	var/obj/machinery/power/NuclearReactor/reactor
+
+/datum/nuclearreaction/proc/GetCoolAmounts()
+	var/AmtSubtract
+	var/frum = 0
+	for(var/obj/item/twohanded/required/ControlRod/CR in CRS)
+		if(!istype(CR,/obj/item/twohanded/required/ControlRod))
+			return
+		AmtSubtract += CR.ControlPower
```
ThatLing:
```
for(var/T in CRS)
	var/obj/item/twohanded/required/ControlRod/CR = T
	AmtSubtract += CR.ControlPower
```

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+
+/obj/machinery/power/NuclearReactor
+	name = "Nuclear Reactor Core"
+	desc = "A massive nuclear reactor with an inbuilt cooling spire, two access hatches, and a large port for attaching atmos pipes to, Alt Click the reactor to open its access port for removal of objects, and CTRL Click the reactor to open / close its fuel port. The Nt-STE-VENS series of reactors are a testament to the old saying 'If it ain't broke, don't fix it'"
+	icon = 'icons/obj/reactor.dmi'
+	icon_state = "reactor"
+	critical_machine = TRUE
+	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | UNACIDABLE | ACID_PROOF | FREEZE_PROOF
+	var/datum/nuclearreaction/react = new //J!nx here with another reaction video
+	var/Heat = 0 //how hot are we as a percent, so up to 100
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: here too

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+	var/AmtSubtract
+	var/frum = 0
+	for(var/obj/item/twohanded/required/ControlRod/CR in CRS)
+		if(!istype(CR,/obj/item/twohanded/required/ControlRod))
+			return
+		AmtSubtract += CR.ControlPower
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
+			return
+		AmtSubtract += FR.FuelPenalty
```
ThatLing:
```
for(var/T in FRS)
	var/obj/item/twohanded/required/FuelRod/FR = T
	AmtSubtract += FR.FuelPenalty
```

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		frum ++
+	if(!frum || frum <=0) //No fuel? Drain that heat fast lad
```
ThatLing: Why the ``!frum``?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		AmtSubtract += 2
+	if(AmtSubtract <= 0)
+		AmtSubtract = 0
+	return AmtSubtract
```
ThatLing:
```
return max(AmtSubtract, 0)
```

tsaoToast:
holy fucking shit lad

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+
+
+/datum/nuclearreaction/proc/GetFuelAmounts()
+	var/AmtAdd
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
+			return
+		AmtAdd += FR.FuelPower
+	var/final = AmtAdd
+	return final //IE, AmtSubtract with one CR = 1, AmtAdd with 2 Uranium = 4, so resultant reaction power is 3, which also means a heat spike
```
ThatLing: Why not just ``return AmtAdd``?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		AmtSubtract = 0
+	return AmtSubtract
+
+
+/datum/nuclearreaction/proc/GetFuelAmounts()
+	var/AmtAdd
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
+			return
+		AmtAdd += FR.FuelPower
```
ThatLing:
```
for(var/T in FRS)
	var/obj/item/twohanded/required/FuelRod/FR = T
	AmtAdd += FR.FuelPower
```

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+			react.FRS += I
+		else
+			to_chat(user, "Open the fuel hatch first!")
+	if(istype(I, /obj/item/twohanded/required/ControlRod))
+		to_chat(user, "You slot [I] into [src]!")
+		I.forceMove(src)
+		react.CRS += I
+
+/obj/machinery/power/NuclearReactor/attack_hand(mob/user)
+	if(!ReactorInoperable)
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: Seems a bit awkward to check for the inverse of a variable. Maybe change this var to `operational`?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		monitor.icon_state = "monitor-low"
+	if(Heat > MEDIUM_HEAT)
+		monitor.icon_state = "monitor-med"
+	if(Heat > HIGH_HEAT)
+		monitor.icon_state = "monitor-high"
+	if(Heat > CRITICAL_HEAT) //At this stage. Worry. A lot.
+		monitor.icon_state = "monitor-warning"
+	add_overlay(waste)
+	add_overlay(fuel)
+	add_overlay(monitor)
```
ThatLing: https://github.com/Baystation12/Baystation12/pull/4449

Maybe do this?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+
+/obj/machinery/power/NuclearReactor/Initialize()
+	. = ..()
+	react.reactor = src
+	UpdateIcon()
+	state |= FUELHATCH_OPEN
+	state |= WASTEHATCH_OPEN
+	connect_to_network()
+//	outlet = new(src.loc)
+	var/obj/dummy = new(src.loc) //Handle placing pipes properly
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: You can change all this to `outlet = locate(/obj/machinery/atmospherics/components/binary/pump/nuclear) in get_step(src, NORTH)`. You probably could even omit the path, locate should automatically find the correct thing.

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+//	outlet.name = "[src] gas outlet port"
+//	outlet.dir = NORTH
+	dummy.y += 1 //1 tile above our loc, to get outlet!
+	outlet = locate(/obj/machinery/atmospherics/components/binary/pump/nuclear) in dummy.loc
+	qdel(dummy)
+	STOP_PROCESSING(SSmachines,src)
+
+/obj/machinery/power/NuclearReactor/proc/CheckPipes() //In case your pipes blow up
+	outlet = null
+	outlet = null
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: :thinking:

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+	var/obj/dummy = new(loc)
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: change this too

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+	START_PROCESSING(SSmachines,src)
+
+/obj/machinery/power/NuclearReactor/process()
+	if(ReactorInoperable)
+		stop()
+		return
+	UpdateIcon()
+	ProcessAtmos()
+	CheckState()
+	Time ++
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: no space before `++`, ree

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+			num = 0
+		S.temperature += num
+
+/obj/machinery/power/NuclearReactor/proc/stop()
+	STOP_PROCESSING(SSmachines,src)
+	Heat = 0
+	HeatRate = 0
+
+/obj/machinery/power/NuclearReactor/proc/GetPower(var/C = 2) //C is changeable for balance as you see fit. You can also change input power modifier.
+	return Heat^C * Efficiency * C
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: `^` is XOR, I assume you meant `Heat**C`?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+	var/AmtAdd
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
+			return
+		AmtAdd += FR.FuelPower
+	var/final = AmtAdd
+	return final //IE, AmtSubtract with one CR = 1, AmtAdd with 2 Uranium = 4, so resultant reaction power is 3, which also means a heat spike
+
+/datum/nuclearreaction/proc/ConsumeFuel()
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
```
ThatLing:
```
for(var/T in FRS)
    var/obj/item/twohanded/required/FuelRod/FR = T
```

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		if(FR.loc != reactor || !FR in reactor.contents)
+			FRS -= FR
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
+			return
```
ThatLing: Not needed

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+		if(FR.integrity <= 0) //That rod's spent, make it into DU
+			reactor.say("[FR] died!")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: `[FR] has been depleted.`

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+		if(!istype(FR, /obj/item/twohanded/required/FuelRod/depleted)) //Depleted rods don't deplete further, they just interrupt heat
+			FR.integrity -= 10 - reactor.Efficiency/10 // Eg, 100/10 = 10, so integrity -= 0 aka no health lost, 25 = 2.5, so 8.5 health lost, so keep that efficiency up!
+	for(var/obj/item/twohanded/required/ControlRod/CR in CRS)
+		if(CR.loc != reactor || !CR in reactor.contents)
+			CRS -= CR
+		if(!istype(CR,/obj/item/twohanded/required/ControlRod))
+			return
+		if(CR.integrity <= 10) //CR is spent!
+			qdel(CR)
+			reactor.say("[CR] died!")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: deplete

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
+			return
+		AmtAdd += FR.FuelPower
+	var/final = AmtAdd
+	return final //IE, AmtSubtract with one CR = 1, AmtAdd with 2 Uranium = 4, so resultant reaction power is 3, which also means a heat spike
+
+/datum/nuclearreaction/proc/ConsumeFuel()
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(FR.loc != reactor || !FR in reactor.contents)
+			FRS -= FR
```
ThatLing: You could use some of these to pretty up your code.

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+		if(CR.loc != reactor || !CR in reactor.contents)
+			CRS -= CR
+		if(!istype(CR,/obj/item/twohanded/required/ControlRod))
+			return
+		if(CR.integrity <= 10) //CR is spent!
+			qdel(CR)
+			reactor.say("[CR] died!")
+		CR.integrity -= 10 - reactor.Efficiency/10
+
+/obj/item/twohanded/required/ControlRod
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: You should change `required` to something more descriptive, like `reactor_rod`.

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,458 @@
...
+	max_integrity = 10000
+	var/obj/machinery/power/NuclearReactor/reactor
+	var/ControlPower = 4.8 // 1 Rod of this will cockblock 1 fuel rod
+
+/obj/item/twohanded/required/FuelRod
+	name = "Uranium Fuel Rod"
+	desc = "A huge stick of uranium to be inserted into a nuclear reactor, its casing is in tact, mitigating most of its radioactive output"
+	icon = 'icons/obj/ControlRod.dmi'
+	icon_state = "uraniumrod"
+	var/integrity = 10000//As rods become worn, they are damaged.
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: what is inheritance

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		if(FR.loc != reactor || !FR in reactor.contents)
+			FRS -= FR
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
+			return
+		if(FR.integrity <= 0) //That rod's spent, make it into DU
+			reactor.say("[FR] died!")
+			DepletedAmt ++
+			qdel(FR)
+			FRS -= FR
+			FR = null
```
ThatLing: Jesus what the fuck. Don't use an object after you qdel it.

```
LAZYREMOVE(FRS, FR)
qdel(FR)
```

1fbff5f83b23d39d38b1dfcb4cac8d9b (commented):
Style and stuff

CMOisLing:
This is great but I think it is too safe, even for yog engineers

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+			var/obj/item/twohanded/required/FuelRod/depleted/spent = new
+			FRS += spent
+			spent.forceMove(reactor)
+			return
+		if(!istype(FR, /obj/item/twohanded/required/FuelRod/depleted)) //Depleted rods don't deplete further, they just interrupt heat
```
ThatLing: Not a fan of the typecheck, can't you check the ``FuelPower`` or add a ``depleted`` var?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		if(state & FUELHATCH_OPEN)
+			to_chat(user, "You slot [I] into [src]!")
+			I.forceMove(src)
+			react.FRS += I
+		else
+			to_chat(user, "Open the fuel hatch first!")
+	if(istype(I, /obj/item/twohanded/required/ControlRod))
+		to_chat(user, "You slot [I] into [src]!")
+		I.forceMove(src)
+		react.CRS += I
```
ThatLing: Hey look and here we make sure CRS only contains objects of type ``/obj/item/twohanded/required/ControlRod```.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
The connections between normal and thicc pipes need some refining too

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+//As we know from physics class, efficiency is helped by cooling it down, so at this point in the reaction your efficiency will be helped by some nice n2, for now we'll simulate n2 with vars.
+/obj/machinery/power/NuclearReactor/proc/GetEfficiency(var/temp = 1,var/amt = 1) //As a percent
+	return (1/temp) * amt
+
+/datum/nuclearreaction
+	var/name = "nuclear reaction"
+	var/FuelAmt = 0 //How much fuel
+	var/CombFuelPower = 0 //And combined, how strong are all our fuels
+	var/list/obj/item/twohanded/required/ControlRod/CRS = list()
+	var/list/obj/item/twohanded/required/FuelRod/FRS = list()
```
ThatLing: Move these list inits to ``/Initialize()``.

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+			FRS += spent
+			spent.forceMove(reactor)
+			return
+		if(!istype(FR, /obj/item/twohanded/required/FuelRod/depleted)) //Depleted rods don't deplete further, they just interrupt heat
+			FR.integrity -= 10 - reactor.Efficiency/10 // Eg, 100/10 = 10, so integrity -= 0 aka no health lost, 25 = 2.5, so 8.5 health lost, so keep that efficiency up!
+	for(var/obj/item/twohanded/required/ControlRod/CR in CRS)
+		if(CR.loc != reactor || !CR in reactor.contents)
+			CRS -= CR
+		if(!istype(CR,/obj/item/twohanded/required/ControlRod))
+			return
```
ThatLing: Unneeded

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+			qdel(FR)
+			FRS -= FR
+			FR = null
+			var/obj/item/twohanded/required/FuelRod/depleted/spent = new
+			FRS += spent
+			spent.forceMove(reactor)
+			return
+		if(!istype(FR, /obj/item/twohanded/required/FuelRod/depleted)) //Depleted rods don't deplete further, they just interrupt heat
+			FR.integrity -= 10 - reactor.Efficiency/10 // Eg, 100/10 = 10, so integrity -= 0 aka no health lost, 25 = 2.5, so 8.5 health lost, so keep that efficiency up!
+	for(var/obj/item/twohanded/required/ControlRod/CR in CRS)
```
ThatLing:
```
for(var/T in CRS)
    var/obj/item/twohanded/required/ControlRod/CR = T
```

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		if(CR.loc != reactor || !CR in reactor.contents)
```
ThatLing: Shouldn't this be ``|| !(CR in reactor.contents))``?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
+			return
+		AmtAdd += FR.FuelPower
+	var/final = AmtAdd
+	return final //IE, AmtSubtract with one CR = 1, AmtAdd with 2 Uranium = 4, so resultant reaction power is 3, which also means a heat spike
+
+/datum/nuclearreaction/proc/ConsumeFuel()
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(FR.loc != reactor || !FR in reactor.contents)
```
ThatLing: Shouldn't this be ``|| !(FR in reactor.contents))``?

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+		if(!istype(FR, /obj/item/twohanded/required/FuelRod/depleted)) //Depleted rods don't deplete further, they just interrupt heat
+			FR.integrity -= 10 - reactor.Efficiency/10 // Eg, 100/10 = 10, so integrity -= 0 aka no health lost, 25 = 2.5, so 8.5 health lost, so keep that efficiency up!
+	for(var/obj/item/twohanded/required/ControlRod/CR in CRS)
+		if(CR.loc != reactor || !CR in reactor.contents)
+			CRS -= CR
+		if(!istype(CR,/obj/item/twohanded/required/ControlRod))
+			return
+		if(CR.integrity <= 10) //CR is spent!
+			qdel(CR)
+			reactor.say("[CR] died!")
```
ThatLing: qdel AFTER using it. Also put a return here, you're still editing the ``integrity`` of a qdelled object.

yogstation/code/modules/power/NuclearEngine.dm
```diff
@@ -0,0 +1,453 @@
...
+#undef WASTEHATCH_OPEN
+
+/obj/effect/nuclearwaste
+	name = "nuclear waste"
+	desc = "A highly radioactive pool of sludge; the remains of a nuclear reactor that went critical."
+	icon = 'icons/obj/atmospherics/pipes/nuclear.dmi'
+	icon_state = "nuclearwaste"
+
+/obj/effect/nuclearwaste/attackby() //You need to contain it
+	return 0
```
ThatLing: return FALSE

ThatLing (changes requested):

yogstation/code/modules/power/NuclearEngine.dm
Kmc2000: required is so you need two open hands to pick it up

ThatLing (changes requested):
You still have files that aren't completely lowercase. Remove all the changes in ``code/modules/atmospherics/machinery/pipes/manifold.dm``. Mapmerge didn't work correctly.

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	var/SavedWarningTime = 0 //For cooldowns
+	var/WarningSound = 'yogstation/sound/effects/CoreOverheating.ogg'
+	var/MeltDownSound = 'yogstation/sound/effects/Meltdown.ogg'
+	var/ReactorInoperable = FALSE
+	density = TRUE
+//	var/CoreHealth = 1000 //This gets chipped away when you run it above its temperature tolerance
+
+/obj/machinery/power/NuclearReactor/examine(mob/user)
+	. = ..()
+	to_chat(user, "Its heat readout reads: [Heat] K, Its maximum tolerance limit is [CRITICAL_HEAT] K")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: Uncapitalize the second its, add a period

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	icon_state = "wastehatch-closed"
+
+/obj/effect/reactoroverlay/monitor
+	icon_state = "monitor-off"
+
+/obj/machinery/power/NuclearReactor/CtrlClick(mob/user)
+	if(state & FUELHATCH_OPEN)
+		state &= ~FUELHATCH_OPEN //Bitflags! We add the OPPOSITE of fuelhatch open to "state", in other words, inverting it, if the hatch is open, this closes it
+		playsound(loc, 'sound/effects/bin_close.ogg',50,1)
+		to_chat(user,"You shut the fuel hatch")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: period

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	else
+		state |= FUELHATCH_OPEN
+		playsound(loc, 'sound/effects/bin_open.ogg',50,1)
+		to_chat(user,"You open the fuel hatch")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: another period

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	UpdateIcon()
+
+/obj/machinery/power/NuclearReactor/AltClick(mob/user)
+	if(state & WASTEHATCH_OPEN)
+		state &= ~WASTEHATCH_OPEN //Bitflags! We add the OPPOSITE of wastehatch open to "state", in other words, inverting it, if the hatch is open, this closes it
+		playsound(loc, 'sound/effects/bin_close.ogg',50,1)
+		to_chat(user,"You shut the access hatch")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: and another one

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	else
+		state |= WASTEHATCH_OPEN
+		playsound(loc, 'sound/effects/bin_open.ogg',50,1)
+		to_chat(user,"You open the access hatch")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: Are you pregnant? 'Cause you are missing your periods

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	add_overlay(monitor)
+
+/obj/machinery/power/NuclearReactor/attackby(obj/item/I, mob/user)
+	if(istype(I, /obj/item/twohanded/required/FuelRod))
+		if(state & FUELHATCH_OPEN)
+			to_chat(user, "You slot [I] into [src]!")
+			I.forceMove(src)
+			react.FRS += I
+		else
+			to_chat(user, "Open the fuel hatch first!")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b, line 116: How about some `<span class='warning'>` here and there? Never hurts to make things colorful

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	connect_to_network()
+//	outlet = new(src.loc)
+	outlet = locate(/obj/machinery/atmospherics/components/binary/pump) in get_step(src, NORTH)
+	STOP_PROCESSING(SSmachines,src)
+
+/obj/machinery/power/NuclearReactor/proc/CheckPipes() //In case your pipes blow up
+	outlet = null
+	outlet = locate(/obj/machinery/atmospherics/components/binary/pump) in get_step(src, NORTH)
+	if(outlet)
+		say("Success! outlet pump registered as [outlet]")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: Capitalization, period.

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+		say("Error! No outlet pump could be found!")
+		return FALSE
+
+/obj/machinery/power/NuclearReactor/Destroy()
+	disconnect_from_network()
+	. = ..()
+
+/obj/machinery/power/NuclearReactor/proc/start()
+	if(!outlet)
+		say("ERROR! No outlet found!, please attach a standard atmospherics pump on the highlighted tile to the left!")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: No!, this is wrong

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	//Efficiency = GetEfficiency() gonna play round with this temporarily
+	HeatRate = react.GetFuelAmounts()
+	Heat -= react.GetCoolAmounts()
+	if(HeatRate > 0)
+		Heat += HeatRate
+	if(Heat <= 0)
+		stop()
+		return
+
+/obj/machinery/power/NuclearReactor/proc/MeltdownAnnounce() //Announce the meltdown to give them a chance to RUN
```
1fbff5f83b23d39d38b1dfcb4cac8d9b, line 196: You should be fired and this proc should be placed in a hall of shame as an example of how not to code

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	stop()
+	ReactorInoperable = TRUE
+	if(ReactorInoperable)
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: >Checking a variable right after setting the value

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+		for(var/mob/M in GLOB.player_list)
+			if(M.z == z)
+				M << null //Don't want to assault their eardrums
+				M << MeltDownSound
+				priority_announce("Nuclear reactor status: CRITICAL Meltdown imminent! Evacuate engineering section IMMEDIATELY", "Nuclear Reactor Monitoring Subsystem",'sound/ai/attention.ogg')
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: Why are you announcing this for every player

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+				addtimer(CALLBACK(src, .proc/Meltdown), 410) //It's the final countdown
+
+/obj/machinery/power/NuclearReactor/proc/Meltdown() //qdel(station)
+	explosion(get_turf(src),20,40,40, 100)
+	var/datum/round_event_control/radiation_storm/RS = new()
+	RS.runEvent()
+	ReactorInoperable = TRUE
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: You set this to TRUE 13 lines up

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+/obj/machinery/power/NuclearReactor/proc/MeltdownAnnounce() //Announce the meltdown to give them a chance to RUN
+	stop()
+	ReactorInoperable = TRUE
+	if(ReactorInoperable)
+		for(var/mob/M in GLOB.player_list)
+			if(M.z == z)
+				M << null //Don't want to assault their eardrums
+				M << MeltDownSound
+				priority_announce("Nuclear reactor status: CRITICAL Meltdown imminent! Evacuate engineering section IMMEDIATELY", "Nuclear Reactor Monitoring Subsystem",'sound/ai/attention.ogg')
+				addtimer(CALLBACK(src, .proc/Meltdown), 410) //It's the final countdown
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: Can't wait for 60 explosions and radiation storms

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+
+/obj/machinery/power/NuclearReactor/proc/Meltdown() //qdel(station)
+	explosion(get_turf(src),20,40,40, 100)
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: This is capped by the bombcap, add `ignorecap = TRUE`

1fbff5f83b23d39d38b1dfcb4cac8d9b: Also why is the light damage range the same as medium damage?

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	icon_state = "reactor-broken"
+	for(var/turf/T in orange(src, 10))
+		new /obj/effect/nuclearwaste(T)
+
+
+/obj/machinery/power/NuclearReactor/proc/CheckState()
+	if(state & FUELHATCH_OPEN || state & WASTEHATCH_OPEN)
+		radiation_pulse(src, Heat, 40) //You fucked up kid
+	if(Heat >= MELTDOWN) //Oh FUCK
+		stop()
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: stop() is already called in MeltdownAnnounce

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+			SavedWarningTime = world.time
+			for(var/mob/M in GLOB.player_list)
+				if(M.z == z)
+					M << WarningSound
+		else
+			return
+
+/obj/machinery/power/NuclearReactor/proc/ProcessAtmos()
+	for(var/datum/gas_mixture/S in outlet.airs)
+		if(S.temperature <= 20000) //A small nerf to avoid tritium production
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: You sure this will work? Tritium production depends on the ratio between oxygen and plasma, not temperature

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+
+//As we know from physics class, efficiency is helped by cooling it down, so at this point in the reaction your efficiency will be helped by some nice n2, for now we'll simulate n2 with vars.
+/obj/machinery/power/NuclearReactor/proc/GetEfficiency(var/temp = 1,var/amt = 1) //As a percent
+	return (1/temp) * amt
+
+/datum/nuclearreaction
+	var/name = "nuclear reaction"
+	var/FuelAmt = 0 //How much fuel
+	var/CombFuelPower = 0 //And combined, how strong are all our fuels
+	var/list/obj/item/twohanded/required/ControlRod/CRS = list()
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: These variable names could be more descriptive

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+	return final //IE, AmtSubtract with one CR = 1, AmtAdd with 2 Uranium = 4, so resultant reaction power is 3, which also means a heat spike
+
+/datum/nuclearreaction/proc/ConsumeFuel()
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(FR.loc != reactor || !FR in reactor.contents)
+			FRS -= FR
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
+			return
+		if(FR.integrity <= 0) //That rod's spent, make it into DU
+			reactor.say("[FR] has been depleted")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: period

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,449 @@
...
+
+/obj/item/twohanded/required/FuelRod/attack_hand(mob/user)
+	radiation_pulse(src,radiation_strength, 2)
+	. = ..()
+
+/obj/item/twohanded/required/FuelRod/plasma
+	name = "Plasma Fuel Rod"
+	desc = "A huge stick of plasma to be inserted into a nuclear reactor. It has a small label haphazardly stuck to it: WARNING, UNCONTROLLED USE OF THIS WILL OVERHEAT REACTOR. USE AT OWN RISK!" //This isn't a question of if it fucks your setup... It's when.
+	icon_state = "plasmarod"
+	integrity = 1000//As rods become worn, they are damaged.
```
1fbff5f83b23d39d38b1dfcb4cac8d9b, line 351: you don't have to set integrity and max integrity here since they are inherited from the uranium rod

1fbff5f83b23d39d38b1dfcb4cac8d9b (changes requested):

yogstation/code/modules/power/nuclear_engine.dm
Kmc2000: Intentional, it's 1K instead of 10K

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,442 @@
...
+	var/SavedWarningTime = 0 //For cooldowns
+	var/WarningSound = 'yogstation/sound/effects/CoreOverheating.ogg'
+	var/MeltDownSound = 'yogstation/sound/effects/Meltdown.ogg'
+	var/ReactorInoperable = FALSE
+	density = TRUE
+//	var/CoreHealth = 1000 //This gets chipped away when you run it above its temperature tolerance
+
+/obj/machinery/power/NuclearReactor/examine(mob/user)
+	. = ..()
+	to_chat(user, "<span class='warning'>Its heat readout reads: [Heat] K, its maximum tolerance limit is [CRITICAL_HEAT] K.</span>")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: I think notice instead of warning would be more fitting

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,442 @@
...
+	UpdateIcon()
+	state |= FUELHATCH_OPEN
+	state |= WASTEHATCH_OPEN
+	connect_to_network()
+//	outlet = new(src.loc)
+	outlet = locate(/obj/machinery/atmospherics/components/binary/pump) in get_step(src, NORTH)
+	STOP_PROCESSING(SSmachines,src)
+
+/obj/machinery/power/NuclearReactor/proc/CheckPipes() //In case your pipes blow up
+	outlet = null
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: unnecessary

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,442 @@
...
+		AmtSubtract += 2
+	if(AmtSubtract <= 0)
+		AmtSubtract = 0
+	return AmtSubtract
+
+
+/datum/nuclearreaction/proc/GetFuelAmounts()
+	var/AmtAdd
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
```
1fbff5f83b23d39d38b1dfcb4cac8d9b, line 287: I don't think this is possible, and even if it happened, are you sure you want to abort the whole proc instead of skipping over this not-rod with `continue`?

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,442 @@
...
+			return
+		AmtAdd += FR.FuelPower
+	var/final = AmtAdd
+	return final //IE, AmtSubtract with one CR = 1, AmtAdd with 2 Uranium = 4, so resultant reaction power is 3, which also means a heat spike
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: `return AmtAdd`???????????????????????

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,442 @@
...
+		say("Error: No outlet pump could be found!")
+		return FALSE
+
+/obj/machinery/power/NuclearReactor/Destroy()
+	disconnect_from_network()
+	. = ..()
+
+/obj/machinery/power/NuclearReactor/proc/start()
+	if(!outlet)
+		say("<span class='warning'>ERROR: No outlet found, please attach a standard atmospherics pump on the highlighted tile to the left! (one tile above the reactor)<span>")
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: \</span>

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,442 @@
...
+	ReactorInoperable = TRUE
+	priority_announce("Nuclear reactor status: CRITICAL Meltdown imminent! Evacuate engineering section IMMEDIATELY", "Nuclear Reactor Monitoring Subsystem",'sound/ai/attention.ogg')
+	addtimer(CALLBACK(src, .proc/Meltdown), 410) //It's the final countdown
+	for(var/mob/M in GLOB.player_list)
+		if(M.z == z)
+			M << null //Don't want to assault their eardrums
+			M << MeltDownSound
+
+/obj/machinery/power/NuclearReactor/proc/Meltdown() //qdel(station)
+	explosion(get_turf(src),20,40,30, 100, ignorecap=TRUE)
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: why is the light damage range SMALLER than medium damage range!?

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,442 @@
...
+			for(var/mob/M in GLOB.player_list)
+				if(M.z == z)
+					M << WarningSound
+		else
+			return
+
+/obj/machinery/power/NuclearReactor/proc/ProcessAtmos()
+	for(var/datum/gas_mixture/S in outlet.airs)
+		if(S.temperature <= 20000) //A small nerf to avoid ambient temperatures in the room reaching 2000 degrees
+			var/num = Heat
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: `S.temperature += max(0, Heat)`

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,442 @@
...
+	var/name = "nuclear reaction"
+	var/list/obj/item/twohanded/required/ControlRod/CRS = list()
+	var/list/obj/item/twohanded/required/FuelRod/FRS = list()
+	var/obj/machinery/power/NuclearReactor/reactor
+
+/datum/nuclearreaction/proc/GetCoolAmounts()
+	var/AmtSubtract
+	var/frum = 0
+	for(var/obj/item/twohanded/required/ControlRod/CR in CRS)
+		if(!istype(CR,/obj/item/twohanded/required/ControlRod))
```
1fbff5f83b23d39d38b1dfcb4cac8d9b, line 269: again, this is not possible

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -0,0 +1,442 @@
...
+			return
+		AmtSubtract += CR.ControlPower
+	for(var/obj/item/twohanded/required/FuelRod/FR in FRS)
+		if(!istype(FR,/obj/item/twohanded/required/FuelRod))
```
1fbff5f83b23d39d38b1dfcb4cac8d9b, line 273: and this

yogstation/code/modules/power/nuclear_engine.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: misread

1fbff5f83b23d39d38b1dfcb4cac8d9b (changes requested):

yogstation/code/modules/power/nuclear_engine.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: there's more missing \</span>s but I can't be bothered to find them now

yogstation/code/modules/power/nuclear_engine.dm
Kmc2000: Doesn't abort the whole proc, it aborts this iteration of the for loop

yogstation/code/modules/power/nuclear_engine.dm
Kmc2000: It is, byond still cocks it up even when given a specific type.

yogstation/code/modules/power/nuclear_engine.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: https://en.cppreference.com/w/cpp/language/continue
https://en.cppreference.com/w/cpp/language/return

Kmc2000: my life has been a lie

yogstation/code/modules/power/nuclear_engine.dm
```diff
@@ -302,14 +303,15 @@ According to players, the average usage is 160 KW, or 160,000 watts. So that's t
...
+	for(var/T in CRS)
+		var/obj/item/twohanded/required/ControlRod/CR = T
+		if(CR.loc != reactor || (!CR in reactor.contents))
 			CRS -= CR
 		if(!istype(CR,/obj/item/twohanded/required/ControlRod))
 			return
 		if(CR.integrity <= 10) //CR is spent!
-			qdel(CR)
 			reactor.say("<span class='warning'>[CR] has worn out.<span>")
+			qdel(CR)
```
ThatLing: YOURE STILL USING CR AFTER YOU QDEL IT

ThatLing (changes requested):

yogstation/code/modules/power/nuclear_engine.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: if he's fast enough he'll use it before it's deleted

Kmc2000:
Guide here:
https://forums.yogstation.net/index.php?threads/nuclear-reactors-and-you-how-to-not-be-an-atomic-playboy.17154/

1fbff5f83b23d39d38b1dfcb4cac8d9b:
You could add a reactor monitoring console or something, which would log everything the reactor says. 75% of it is probably gonna be missed since there's no one in engineering ever

MacHac:
Before I begin complaining, I'll admit that I object to this on general principle.  The supermatter is an interesting and complex engine with both a simple 'noob' setup (pure nitrogen) and possibilities for optimization and gimmickry.  It also has a number of sabotage options that range from simple, slow, and obvious to hard, fast, and incredibly destructive.  It rewards skill on all sides, without requiring incredible robustitude to do the bare minimum.

If the problem is that yogs players aren't usually competent enough for it, there's better ways to address that than removing it entirely in favor of a simplified system.

Now, assuming you've already considered this and don't care/disagree, here's what's wrong with the engine itself.
**Mechanical Issues:**
- The whole system is incredibly simplified from any other engine, including the singularity/tesla.  There looks to be EXACTLY one variable, IE how long you let it heat up before adding the control rod.
- The system requires babysitting.  Your cited super-safe baby's first setup will explode if left alone for 40 minutes, and has to be regularly cycled.  Best case scenario, one of the engineers is effectively chained to the reactor all round.  Worst case scenario, nobody bothers, and most rounds that reach the 45 minute mark end abruptly.
- Where do you get more fuel/control rods?  I didn't see any way to make/order more.  As it stands now, if all your fuel/control rods lose integrity, you're just screwed.
- The meltdowns are always the same.  Competent saboteurs can induce faster SM meltdowns, or devastating secondary effects like releasing a singularity.  By the same token, competent and fast-thinking engineers can work to neuter SM delams and save most of engineering.  The nuclear meltdown is always the same strength, always completely erases the southern half of the station, and can't be slowed or reversed once it's to the point of telling people about it.
- Related to the above point, most antags without the hijack objective can't touch the reactor at all without running afoul of the murderbone rule.  The days of low-yield sabotage to safely induce a shuttle call are over.
- There's no way to deal with /obj/effect/nuclearwaste aside from just welding engineering shut.  Engineering is permanently tainted.  You might as well have the meltdown call a shuttle, save the command staff 30 seconds of walking.
- 20,30,40 explosion is huge for a regular meltdown.  For reference, the absolute ideal maximum that the supermatter can ever do is 17,37,39, and that's with actual know-how, not just forgetting to check on the thing one shift.
- The reactor looks like it does a lot of its output by just saying things aloud, instead of logging it or announcing it over engineering radio.  This means that someone has to be standing over the damn thing to know if it's starting to destabilize.  If nobody's there to hear the spoken warning, the first indication anyone will have that the reactor is in trouble is when it's already unsaveable.
- The reactor doesn't actually produce that much power.  You boast about your personal setup producing about 300kw, and it looks like it could be heating the gas as much as 2000 degrees.  A decent burn chamber will get much hotter, and the SM itself will produce a stable 1MW without too much risk.

That's just what stands out to me from a three minute skimming of the code and literature involved.
**OTHER COMPLAINTS:**
- While your pipe sprites look very nice, they don't fit well with other types of pipes, and I assume that they're irreplacable.  On the subject of piping, your pipes aren't color-coded at all, and your pump sprites look backwards.  It's hard to tell what's going on.
- There's not necessarily any atmos techs, especially at lowpop.  Moreover, there's not necessarily competent atmos techs at any pop.  It's not a hugely popular job, and a lot of people who DO play it are assistant mains looking for free maint access.
- This engine looks to monopolize a significant and important portion of the mix loop.  While the coolant feed shown is running, you can't mix gasses at all without contaminating the coolant.  Having to surrender one of the most important parts of their system to engineering Every. Single. Round. will only drive more people away from being atmos tech.
- Virtually all DM code uses lowercase variable names, with words separated by underscores.  It looks like you follow this convention about a third of the time, and otherwise use PascalCase.  If you're going to do things nonstandardly, at least be consistent about it.
- /datum/nuclearreaction is just bad.  Make it a component, make it a gas reaction, or just build it into the reactor logic.  There's no reason for it to be separate.
- Is there any way to move the reactor around?  Is there any way to build/order another one?
- The description should override examine() and display the instructions for operating it in a different span.  It looks like you're already overriding examine to do this with temperature, so just move it to there.
- Speaking of instructions for operating it, the ctrl-click alt-click scheme is completely different from anything else in the entire game.  Alt-click is usually for rotation, and ctrl-click is generally unused.  In the interest of consistency and reducing snowflakey behavior, consider a GUI to operate these aspects instead.  Nothing fancy, just a button for toggling the ports, searching for pumps, and ejecting inserted rods.
- Bitflags are archaic, and the performance impact is negligible because there's only ever going to be 0 or 1 of these things.  Also, the first two flags are 1 and 2, not 2 and 4.  This is mostly just whining, but bitflags are hard to maintain, and we should avoid adding more unless there's some legitimate need for them.
- There's an emagged variable, but it's never read or used.

Look, bottom line is, I understand where you're coming from in designing something like this.  But the reactor as it stands right now does a great deal more wrong than it does right.  Even with its implementation problems fixed, I think that the reactor represents an unambiguous dumbing down of the engine, with a lot removed and very little added.

Hell, you'd probably be better off making the TEG the roundstart engine.  It's simple to set up, simple to understand, well integrated, replaceable, customizable, rewards skill more than the reactor, and already well known to the playerbase.  In every metric except pure destructive power, the TEG is a better engine than this reactor, and we got rid of it.  You should ask yourself why before continuing down this path.

Kmc2000:
Adam's offered to spitball some ideas.

This'll be being reworked into something more viable.

nil illegitimae carborundum

ThatLing:
Stale and merge conflicts. Feel free to re-open or make a new PR when you work on this again.

PR #1894, opened 2018-07-25
Adjust vehicles speed

daklj:
Restores this: https://github.com/yogstation13/Yogstation-TG/pull/1007
which was polished by this: https://github.com/yogstation13/Yogstation-TG/pull/1173
which was fucked by this: https://github.com/yogstation13/Yogstation-TG/pull/1232

tested

### DIFF
--- a/code/datums/components/riding.dm
+++ b/code/datums/components/riding.dm
@@ -147,7 +147,7 @@
 		Unbuckle(user)
 		return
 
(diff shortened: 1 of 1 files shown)


Nickvr628:
No fun allowed? What is the point of an ATV if it is the same speed as normal human speed?

I understand cargo mad maxing around the station with ATVs got old, but this seems like a heavy nerf.

PR #1898, opened 2018-07-25
Revs don’t clap when the head of a head rev gets beheaded v2: actually might work version

jcatjmeow:
### The Death of our glorious martyr no longer compells us to clap!
If you’re a head rev or a rev, you will not clap automatically when the guillotine kills a head rev. NT’s finest can use this to try and see who’s a rev, but I don’t think that’ll matter too much since you can still manually do *clap to clap.

I’m debating also making it so mindshielded crew don’t clap when a head of staff is executed, but ehhhhhh I’ll do that later or not at all, lets see if this actually works first.

I also put my if statements on one line, if it causes lag or is annoying to read or whatever I can nest multiple if statements instead.

####Changelog:
:cl:  
tweak: Guillotined head revs no longer cause revs to clap.
/:cl:

### DIFF
--- a/code/game/objects/structures/guillotine.dm
+++ b/code/game/objects/structures/guillotine.dm
@@ -140,7 +140,7 @@
 			for(var/mob/M in viewers(src, 7))
 				var/mob/living/carbon/human/C = M
 				if (ishuman(M))
-					if(!((H.mind.has_antag_datum(/datum/antagonist/rev/head)) && ((H.mind.has_antag_datum(/datum/antagonist/rev/head)) || (H.mind.has_antag_datum(/datum/antagonist/rev))))) //if the victim is a revhead AND the witness is either a revhead or a rev, do not clap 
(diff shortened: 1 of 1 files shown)


Kmc2000:
Instant rev detector

Ai law 2. round up all humans to watch the execution

Law 4: public executions are not human harm

Borg arrest anyone who doesn't clap

jcatjmeow:
yeah but you can still manually clap using *clap to fake it, and honestly if you can manage to peacefully round up literally everyone to watch one of the only revheads die, the revs aren't doing very well

ThatLing:
But why?

```
and honestly if you can manage to peacefully round up literally everyone to watch one of the only revheads die, the revs aren't doing very well
```
The same can be said about the current situation.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Why not

Amelia0010:
This has as close to 0 impact on balance as possible without hitting 0, it's more fluff than anything

jcatjmeow:
Nich is 100% right, this is not balance it's fluff

jcatjmeow:
So uh how do I fix the changelog so it's not "bad"

ThatLing:
Ok so, why?

jcatjmeow:
It never made sense to me that everyone claps at every execution, and this is the biggest example of that. So, now you don't automatically burst into applause when your comrade gets axed.

jcatjmeow:
closing for nnow, will give it another shot somewhere down the line

PR #1933, opened 2018-07-28
Fixes some major lag

ThatLing:
I think will get us our 99% LAGFREE back.


GuppyLaxx:
chesse20 in disguise

ghost:
When he was destroyed his essence infected thousands.

yogstation/code/modules/clothing/shoes/colour.dm
```diff
@@ -0,0 +1,3 @@
+/obj/item/clothing/shoes/sneakers/white/pointed_at(var/mob/user)
```
Time-Green, line 1: You don’t need the var/ part, and pointed_at(mob/user) looks cooler

ThatLing: Don't question the chesse20 code, it's perfect like this

PR #1956, opened 2018-07-29
Small bugfix

ThatLing:
🥇


Amelia0010:
I don't get what bug this is fixing

Amelia0010:
really
makes
you
think

PR #1960, opened 2018-07-29
Fixed point transfer cards being invisible

Cruix:
:cl:  
fix: point transfer cards from the mining vendor are no longer invisible
/:cl:


Amelia0010:
woops

PR #1962, opened 2018-07-29
Updated the free miner ruin

Cruix:
Moved some lockers around so you can reach the ORM without unwrenching them
Moved the tiny fans to where they will actually be useful
Added some pickaxes and a hatchet
Added a box of free miner and free miner pilot IDs, so the captain can hire people and give more people access to move the ship
None of the other free miner roles can be selected until the captain spawner is used
<strike>Armed the self destruct device</strike>


1fbff5f83b23d39d38b1dfcb4cac8d9b:
Aaaaaa github why you no include the commented line

Amelia0010:
it does, you just click on it
image

Cruix:
Damn he found my backdoor

PR #2084, opened 2018-08-07
[admin] Gives admins a verb to do AI vox

Altoids1:
### New admin verb woooo
woop woop Administrator has alpha johnson woop woop

#### Changelog

:cl:  
rscadd: Admins can use AI vox now, as a verb.
/:cl:


ghost:
Where's NTSL?

Nickvr628:
Speedmerge

yogstation/code/modules/admin/verbs/adminvox.dm
```diff
@@ -0,0 +1,59 @@
...
+  else if(voxType == "male") // If we're doing the yog-ly male AI vox voice
+    for(var/word in words)
+      word = lowertext(trim(word))
+      if(!word)
+        words -= word
+        continue
+      if(!GLOB.vox_sounds_male[word])
+        incorrect_words += word
+  else
+    to_chat(src,"<span class='notice'>Unknown or unsupported vox type. Yell at a coder about this.</span>")
```
ThatLing: ``return`` after this

yogstation/code/modules/admin/verbs/adminvox.dm
```diff
@@ -0,0 +1,59 @@
...
+      if(!GLOB.vox_sounds[word])
+        incorrect_words += word
+  else if(voxType == "male") // If we're doing the yog-ly male AI vox voice
+    for(var/word in words)
+      word = lowertext(trim(word))
+      if(!word)
+        words -= word
+        continue
+      if(!GLOB.vox_sounds_male[word])
+        incorrect_words += word
```
ThatLing: You have duplicate code which is ucky. Can't you change the list it takes the words from and then just have one instance of the loop?

yogstation/code/modules/admin/verbs/adminvox.dm
```diff
@@ -0,0 +1,59 @@
...
+
+  log_admin("[key_name(src)] made an admin AI vocal announcement with the following message: [message].")
+  message_admins("[key_name_admin(src)] has forced an AI vox announcement.")
+  if(src.mob) // If the admin has a mob, whether it's a ghost or whatever
+    for(var/word in words) // Then play it
+      play_vox_word(word, src.mob.loc.z, null, voxType) //yogs - male vox
+  else // If, for some ungodly reason, the admin lacks a mob
+    for(var/word in words) // Then play it on the station's Z-level
+    //TODO: Make a define for the station's z-level
+      play_vox_word(word, 2, null, voxType) //yogs - male vox
```
ThatLing:
```
var/z_level = 2
if(src.mob && src.mob.loc)
    z_level = src.mob.loc.z
```

ThatLing (dismissed):

yogstation/code/modules/admin/verbs/adminvox.dm
Altoids1: You mean, like,
```dm
var/list/voxbox = (voxType == "female") ? GLOB.vox_sounds : GLOB.vox_sounds_male
```
?
Because, yeah, that I'm worried DM would do the dumb thing and just make a copy of the list rather than set voxbox to be a pointer.

Also, I'd rather that the structure of this be more easily copy-paste-able for when someone maybe introduces a new AI Vox voice in the future.

yogstation/code/modules/admin/verbs/adminvox.dm
Altoids1: In another episode of Ling hating there being 4 lines of code in something instead of 2

and so suggests a change that would make it 5 lines

yogstation/code/modules/admin/verbs/adminvox.dm
Altoids1: An alternate idea would be to just pass the ``voxType`` to ``play_vox_word()`` and see if it can figure it out. I feel like most of the time this error would happen when someone does not properly implement a new vox voice, and in that case ``play_vox_word()`` has probably been modified to handle the new vox type. 

This is a warning else, not an error else.

yogstation/code/modules/admin/verbs/adminvox.dm
Altoids1: Okay, so, since this is apparently enough to merit this PR being declared unmergable, here's a couple more points.
* Your change would take a 6-line thing and make it 5 lines, which is not an absolutely revolutionary change to begin with.
* Said change (to reduce its line usage *by one*) would introduce the use of a dummy variable for the garbage collector to clean up. Not only is memory allocation one of the most expensive operations available (and so a violation of style and a waste of cache space) but we also are not 100% sure that Byond even *has* a working GC, especially when you consider how the game tends to degrade in performance and memory usage the longer the round goes on.

Reading your reviews is like reading the red-pen critiques my English teacher used to put on my essays, just religiously following style guidelines even at a cost to what those style guides are trying to achieve. Ultimately your change is objectively worse and I refuse to agree to it.

yogstation/code/modules/admin/verbs/adminvox.dm
ThatLing:
```
Because, yeah, that I'm worried DM would do the dumb thing and just make a copy of the list rather than set voxbox to be a pointer.
```
It doesn't.

```
Also, I'd rather that the structure of this be more easily copy-paste-able for when someone maybe introduces a new AI Vox voice in the future.
```
How is this more copy-paste-able than doing
```
if(voxType == "female")
    voxList = GLOB.vox_sounds
else if (voxType == "male")
    voxList = GLOB.vox_sounds_male
else if (voxType = "XXX")
    voxList = GLOB.XXX
...
// check if the words are in voxList here
```

yogstation/code/modules/admin/verbs/adminvox.dm
ThatLing:
```
Okay, so, since this is apparently enough to merit this PR being declared unmergable, here's a couple more points.
```
It wasn't only because of this but okay. :)


```
Your change would take a 6-line thing and make it 5 lines, which is not an absolutely revolutionary change to begin with.
```
What does linecount have to do with this? When did I ever say this was about linecount? I proposed a solution that looks cleaner and that allows you to instantly see that the default z-level is 2. Any person that wants to change it in the future will be able to instantly recognize and edit it. If L55 wasn't given, you wouldn't be able to tell the second argument of ``play_vox_word`` is the z-level. My proposed solution eliminates this problem by having a variable named ``z_level`` (or something alike) so any future coders can instantly see that the second argument is the z-level. The reason I requested this change is because of readability and maintainability, not "muh linecount".

```
Said change (to reduce its line usage by one) would introduce the use of a dummy variable for the garbage collector to clean up. Not only is memory allocation one of the most expensive operations available (and so a violation of style and a waste of cache space)
```
Oh no, the GC has to clean up 1 extra variable 😱. How will we ever cope with this performance degradation and memory waste? We better elimate all local variables to unlock the full potential of BYOND and get maximum framerates. If we assume BYOND uses a proper implementation, the variable (wether its a straight up float or the float is inside a class), it will most likely be allocated on the stack. If you're complaining about allocating things on the stack, geez. Even if it wasn't allocated on the stack, the performance you lose by allocating a class in ram is still negligible.

Regarding the cache space, the program probably wouldn't even need to grab the variable from RAM. It was created a few instructions before so it will most likely still be in a register/on the stack. Worst case, you have 1 cache miss. Afterwards, the variable isn't used anymore and the CPU overrides the unused variable with something else, effectively wasting no cache memory at all.


```
but we also are not 100% sure that Byond even has a working GC, especially when you consider how the game tends to degrade in performance and memory usage the longer the round goes on.
```
That poor variable that only gets created once admins use some niche button. I'm sure we're all gonna cry over all those potentially wasted bytes. And I mean literally bytes, BYOND uses a 32-bit float to represent numbers. Add some extra bytes used to track usage count and stuff and you're looking at like 16 bytes every time this is used (assuming the variable isn't cleaned/reused), worst case.

```
even at a cost to what those style guides are trying to achieve
```
Please enlighten me, what are these style guide trying to achieve? Because clearly I don't know.


Also, the BYOND docs state that mob.loc can be null. Your code would runtime in this case. :)

Amelia0010:
will altoids ever finish a PR?

Find out next time, on yogstation-TG

ThatLing:
No, he's angry at me and won't listen to my advice

Amelia0010:
image

PR #2113, opened 2018-08-10
[admin] Makes Get Law History less poopoo

1fbff5f83b23d39d38b1dfcb4cac8d9b:
changed stuff so cyborg laws don't get added 3 times per upload


Amelia0010:
💩

PR #2176, opened 2018-08-14
fuck nich

ghost:
Reverts yogstation13/Yogstation-TG#1980

reason: i don't like it

real talk it makes riot gear too effective at dealing with aliens paired with riot shields  since riot shields block spit and such

xenos are meant to be round-ending antags, quote unquote from /tg/, so i mean, that's the design philosophy there

that being said nobody cares about design philosophy but frankly i just want to be able to kill people as a xenomorph without any effort whatsoever and that is the entire unashamed point of this PR

yes i'm drunk


ThatLing:
muh me no likey

PR #2178, opened 2018-08-14
adds the ability to add job-replacements for traitoritems

Amelia0010:

Essentially, it allows us to add minor differences to the traitor item depending on the job of the person buying them. Like, say we make a subtype of EMP grenades that honk when they EMP (because i'm bad at examples) we'd add a line to the uplink_item for EMP grenades, that essentially goes;

`jobspecials["Clown"] = /obj/item/grenade/emp/clown`

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -93,8 +93,8 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	if(!spawn_item)
 		return
 	var/atom/A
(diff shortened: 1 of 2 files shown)


Amelia0010:
And yes, I know we could just add the grenades as a job-specific clown item instead, while keeping the normal item but that's boring

ThatLing:
What about description?

Amelia0010:
I'll look at adding it to the description if the user is of that certain job, I guess

Amelia0010:
ok fugg tgui

PR #2190, opened 2018-08-15
captain now requires the player to play 5 hours of heads :)

Amelia0010:


Amelia0010:
as opposed to 3 hours of any role (including assistant)

Fillywonk:
This is good, I like this

PR #2191, opened 2018-08-15
o o f

Amelia0010:


ThatLing (approved):

Nirnael:
At least don't obscure the gamemode chances to non-admins. I don't see the reason for that. It's nice knowing if it's only traitorling retarded rounds as always or something different. I mean, this is basically the centcomm summary for non-heads and some heads say that IC anyway. And you can't figure anything out from the weights because it's RNG.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
players don't need this information

PR #2212, opened 2018-08-17
Synths don't into reagents

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Makes synths not process reagents other than getting healed from synthflesh


Rayisken:
reeeeeeeeeeeeeeeeeeeeee

PR #2247, opened 2018-08-20
changes PRannounce so only the PR title has the link

Amelia0010:
image

this looks dumb


1fbff5f83b23d39d38b1dfcb4cac8d9b:
doesn't look that dumb but okay I guess

PR #2341, opened 2018-08-26
Adrenaline Sacs doesn't prevent changelings from being stunned as long as it's in effect

Amelia0010:

Changes it so it now removes whatever stuns they might have at the time, but don't prevent them from being further stunned. Also lessens they amount of toxins they take from using the ability, since those two are intertwined.

People want lings to be more stealthy and less combatty, so here you go. Can still be used to get out of combat, but not to buff you during combat 🤷‍♂️ 

#### Changelog

:cl:  
tweak: Adrenaline Sacs now un-stuns you, but doesnt prevent you from being stunned
/:cl:


GuppyLaxx:
wtf stop nerfing the worst antag

Amelia0010:
no

alexkar598:
Clings should be about making the crew confused about who is and isn't not about being plain open about it and wrecking security and command

PR #2370, opened 2018-08-28
Adds airlock braces

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Adds airlock braces which prevent airlocks from being opened by any means other than breaking the airlock down. Security will have access to several of those.
They may be removed by using an ID with security access, or by wrenching them for 20 seconds.
They are only visible, and can be removed from, the side they were applied on:
image
image
Excuse the sprites looking like cans of tomato soup, they'll look better when I find someone with artistic skill

#### Changelog

:cl:  McDonald072
rscadd: Airlock braces have been added, for when you really don't want a door to open. You can find a few of them in the armory. Thank you to Partheo for the sprites!
/:cl:


code/game/machinery/doors/airlock.dm
```diff
@@ -309,6 +309,8 @@
 	qdel(note)
 	for(var/datum/atom_hud/data/diagnostic/diag_hud in GLOB.huds)
 		diag_hud.remove_from_hud(src)
+	if(brace) //yogs
```
Amelia0010, line 312: If I was acoustic I'd probably tell you to move this to its own file, but it's fine

GuppyLaxx: and i am acoustic, so go make a whole .dm with 10 lines just to this shit and ..()

Alexandria404:
interesting idea

ThatLing (approved):

PR #2371, opened 2018-08-28
Makes warden innately know krav maga

Amelia0010:

And prevents HoP from raising the max amount of wardens :dab:

#### Changelog

:cl:  
tweak: NanoTrasen has realized that Krav Maga takes years of hard training to master, and have begun schooling their wardens in Krav Maga (Gloves no longer give you krav maga, warden knows it with/without gloves)
/:cl:


Nirnael:
> And prevents HoP from raising the max amount of wardens

🙅‍♂️ 👎

Oakboscage:
Good tweak Nich <3

Amelia0010:
Yes I had to prevent that, otherwise HoPs could just spawn an infinite amount of Krav Maga-capable security officers

CMOisLing:
I dont really like the idea of it being removed because its an interesting objective to steal since it is almost always going to be in security.

Nirnael:
Give only 1 warden krav maga then. Call it head of wardens. Much like head of security, but a head of the wardens. Much like the head of security is head of security officers, but this one is the head of all the wardens out there.

Moja10:
I don't see the point other than you can never remove krav maga from him ever. Why is this a warranted change? Was the amount of traitors stealing krav maga gloves just too damn high for you Nich?

Amelia0010:
>Why is this a warranted change? Was the amount of traitors stealing krav maga gloves just too damn high for you Nich?

No, the amount of HoS and Captains doing so was.

Amelia0010:
Why do you want the HoP to be capable of raising warden slots exactly? @fizzle7

Amelia0010:
>I dont really like the idea of it being removed because its an interesting objective to steal since it is almost always going to be in security.

you mean like the reflector vest, ion/temperature gun and Warden's compact shotgun?

Oakboscage:
Maybe dump the limit on warden slots and just count on HoPs not abusing the shit out of it?

Amelia0010:
\>counting on retards not to abuse something

Amelia0010:
although if that gets this merged, then sure

Nirnael:
Because it would be the only snowflake job out there apart from heads that cannot be increased. No reason to make it that special. Also because getting multiple wardens is sometimes useful and faster than finding a willing officer to promote to warden. It's also funny. I just feel like the solution should be different, rather than removing mechanics that were useful and working fine before.

Could have a list that tracks heads track wardens and make sure only 1 gets it. Alternatively, give every warden a krav maga scroll when they spawn. When the first warden uses it,  the rest cannot be used anymore. If not, put a single use scroll in the warden's locker that only the warden can use, etc.

I mean, I'm not gonna start spamming wardens, I just don't like removing functionality as a general rule.

PR #2405, opened 2018-08-31
Porting Vampire Gamemode from Hippie!

fluffe9911:
🆑 fluffe9911 for porting steamp0rt for coding it
add: New Vampire Gamemode!
/🆑
I decided to start porting over random things from hippie and old yogs stuff so this is the start of it this is a gamemode hippie has where one person is a vampire who has a goal to suck a certain amount of blood traitor vamps are also possible thats about it!


Nickvr628:
oh god medbay rarely knows what blood bags are, let alone using them to heal people. I fear for our lives, getting partially drained will be a death sentence....

+1

Nirnael:
We already have a vampire race, how does this interact with it?

fluffe9911:
The admeme/pride ruin only vamp is uneffected and still the same vamp its always been so doesn't change much with that vamp

PR #2483, opened 2018-09-08
Cluwnification Burger! (Traitor Item)

fluffe9911:
🆑 Fluffe9911
add: A new traitor item for the clown the cluwnification burger!
/🆑
This adds a 20tc traitor item for clowns that gives the clown a burger that when eaten infects them with a cluwnification virus which then can be spread threw physical contact until the whole station are now cluwnes themselves!


ghost:
how stupid

ghost:
also clown's already got like twelve items

give love to another job instead of a lmao you are permanently fucked burger

Fillywonk:
Clowns have so many items because clowns are easy to think up items for.

ghost:
clowns have so many items because you uncreative idiots are only capable of coming up with absurd memes and not actual ideas worth of consideration and then slap those memes onto a job that can conceivably have anything as long as it's stupid enough

and yet somehow you have managed to surpass this with an idea that is too stupid even for the clown

ShadowDeath6:
should not make it contagious imo, make it like 5 tc and with 1u of the transform chem so you can selectively ruin people’s rounds

PR #2513, opened 2018-09-11
Removes cluwne mask from autodrobe

ThatLing:
why is this here, people told me they didn't like this


fluffe9911:
The autodrobe cluwne mask is a version that is supposed to be 99% curse free so it doesnt turn you into a cluwne unless your unlucky the problem is it seems is that people keep spamming the mask on and off until they turn into a cluwne for god knows why I could remove that effect if you want instead of completly removing it

Moja10:
There has been one instance of someone using that mask as an excuse to act like an antag because they were cluwned. Please remove before it gets out of hand

AsV9:
I guess you could just change the curse logic to be predetermined so that removing and reequipping the mask doesn't increase the likelihood that you'll become a cluwne.

ThatLing:
Closing this in favour for #2514

PR #2580, opened 2018-09-15
manually mirrored a fuckton of PRs because the mirrorbot is a pile of garbage which should be purged and the whole server rack shredded and set on fire to prevent any other servers from making the same mistake as we did several months ago installing this piece of garbage which pretends to be making our lives easier porting stuff from TG automatically but actually you have to babysit it in case it shits itself and doesn't clean up. Now I don't know if I missed anything and this is probably not going to compile, and ling is on leave again. Fun! Oh, I forgot, no changelogs either.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
sorry, got cranky


Smudged:
This is pain in a PR.

Amelia0010:
honk

Issue #1024, opened 2018-05-27
Additional janitor lockers in the janitor closet


MegaEmpirical:
There are three janitor spots on the station and really only 1 locker (there are 2 but the other is just the bio suit). We used to have 2 (one of each locker so 4 total) before the rebase. If not, then please at least add a 4th wet floor sign to the single janitor locker, the cart holds 4 and there are only 3 in there.

MegaEmpirical:
Closing because of:

https://github.com/yogstation13/Yogstation-TG/pull/2761

Amelia0010:
https://github.com/yogstation13/Yogstation-TG/blob/7d2a976b3cbe8fe410c38e2d76f4bce402277b24/code/modules/jobs/job_types/cargo_service.dm#L274

You're an idiot

MegaEmpirical:
It helps if you explain why you're calling someone an idiot. I assume because I said there were three janitor positions? A small mistake, I thought there were three. I was only asking for one more locker anyway.

PR #2761, opened 2018-09-30
two more closets for janitor and 1 more sign per closet

thehighwayman:

Gives the janitors custodial closet two more closets of each type so that the extra janitor slots will have stuff. also has 4 signs per each closet so you can fully load a janicart with one locker
Fixes issue #1024

#### Changelog

:cl:  
rscadd: two more janitor closets in the custodial closet and 1 more sign in a jcloset
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -19802,11 +19802,11 @@
 /turf/open/floor/plating,
 /area/storage/emergency/port)
 "aSo" = (
-/mob/living/carbon/monkey/punpun,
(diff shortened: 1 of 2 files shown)


ThatLing:
Not modular, see here for instructions.

Amelia0010:
https://github.com/yogstation13/Yogstation-TG/blob/7d2a976b3cbe8fe410c38e2d76f4bce402277b24/code/modules/jobs/job_types/cargo_service.dm#L274

really
makes
me
thonk

Amelia0010:
`two more janitor closets`
`total_positions = 2 `

So now there's 3 closets for a job with 2 positions 🤔

PR #2780, opened 2018-10-01
Fixes Cloning

GuppyLaxx:
Clear on the title.
### It makes cloning require MONEY in your account. 5 each tick, the total amount varies and I'm too lazy to figure it out.


Moja10:
In my opinion regarding the economy system affecting the crew's use of default station machines should be left to vendors only. I think the economy system works better as a player device to use with interacting with other players. I don't like this idea out of principle, in the same way I wouldn't like the chemistry machine requiring money, or science requiring money to print things, or ores costing money via code to use in protolathes or things like this. I could maybe see this PR if it was literal chump change to be cloned, so only the poorest of the poor don't get cloned fully, or if only people with 0$ dollars don't get cloned, but eh, don't feel like this is a good idea in principle.

Ktlwjec:
no meme prs please

Ratvar:
get borged/defibbed/newbody/botanyplanted/chaplained/Work_FoR_MoNeY if u have no money, so many possibilities
forces people (assistants) to do something if they care to sirvive

ThatLing (changes requested):
Not modular

PR #2796, opened 2018-10-02
Fix input runtime

ThatLing:
Blimey


Amelia0010:
is this the cause of the hotkeys not working thing?

ThatLing:
I don't know, probably. This runtimes when there's a null in ``GLOB.clients``. /tg/ input SS doesn't have this but other code that iterates over ``GLOB.clients`` does do a null check.

ThatLing:
Actually, this is a lazy fix. #2797 should fix the issue that's causing null in ``GLOB.clients``.

PR #2909, opened 2018-10-07
Fixes floor cluwnes FOR REAL THIS TIME

fluffe9911:
About time I figured this out jesus thanks for the help Time-Green

### DIFF
--- a/yogstation/code/modules/mob/living/simple_animal/hostile/floor_cluwne.dm
+++ b/yogstation/code/modules/mob/living/simple_animal/hostile/floor_cluwne.dm
@@ -367,7 +367,8 @@ GLOBAL_VAR_INIT(floor_cluwnes, 0)
 
 
 /mob/living/simple_animal/hostile/floor_cluwne/proc/Kill(mob/living/carbon/human/H)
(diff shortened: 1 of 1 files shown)


yogstation/code/modules/mob/living/simple_animal/hostile/floor_cluwne.dm
```diff
@@ -366,6 +367,7 @@ GLOBAL_VAR_INIT(floor_cluwnes, 0)
 
 
 /mob/living/simple_animal/hostile/floor_cluwne/proc/Kill(mob/living/carbon/human/H)
+	if(H)
```
Amelia0010: 🤔 this if() doesn't do anything?

fluffe9911: it should check for when it pulls someone down if its a human which should prevent the bug of the floor cluwne bugging out I think

fluffe9911: wait that didnt work one moment

PR #2924, opened 2018-10-08
Fix typo for Cryogenic Treatment nanite program

CMOisLing:
The nanites rapidly (skin -> sink) heat through the host's skin, lowering their temperature.


Time-Green:
This should really be on upstream instead

Time-Green:
You should really just PR this to /tg/

CMOisLing:
fine

PR #2934, opened 2018-10-10
Actually stops people from choosing shadowlings roundstart

Amelia0010:
because both @Altoids1 and @GuppyLaxx are dumb


Nickvr628:
no fun allowed

PR #2960, opened 2018-10-10
Also removes zombies from halloween event

ThatLing:
:)


fluffe9911:
*Speed Merging Own Pr
For real though I will prob attempt to remove the whole zombie revive thing from high functioning zombies later cause I like my zomb memes (or just replace the halloween zombie with the drug version)

PR #3013, opened 2018-10-13
Fixes and Buffs Shadowling Glare so it now stuns up to 4 tiles also lasts a bit longer

fluffe9911:
Broken atm and kinda sucks so a small buff wont hurt


Nickvr628:
Now it stuns EVERYONE in a 4 tile radius? Because that sounds broken as fuck.

fluffe9911:
No it still only stuns one person now it can just stun them from up to 4 tiles lol

Nickvr628:
Ok sounds cool. I think it still might be a bit high but playtesting will determine that.

Ludox235:
Please merge this the sling's glare sucks so hard it feels like even if i'm 1 tile away they wont get stunned

Time-Green:
used to be 7 tile radius so this seems fair enough

ghost:
shadowling is powerful enough as is, it doesn't need any further buffs

SomeguyManperson:
give the stun like 10 or 20 more deciseconds instead of the range buff if you want it to be more useful.

Issue #2446, opened 2018-09-04
Detective's office since rebase lacks supplies


MegaEmpirical:
There's no paper pin +paper, though this is easily rectified. However, there is only one folder in the entire place and it's just in the suit case, no way to file cases really without them.

Having a miscellaneous crafting recipe for folders using paper might be cool.

alexkar598:
You can create folders using cardboard which is either found in maint or made from folding empty boxes

MegaEmpirical:
Good to know about the cardboard. Closing this out since a fix is incoming.

MegaEmpirical:
Re-opening since the issue is now back again since the detective's got moved.

PR #3016, opened 2018-10-13
Gives the detective more office supplies

Ktlwjec:
Gives the detective 2 more folders (under the briefcase), a seclite, cuffs and a paper bin:
image

Fixes issue #2446


MegaEmpirical:
.... I never thought anyone would actually do this. I am so happy. I can file all the paper work!

MegaEmpirical:
The issue is basically back again since the detective's got moved.

PR #3059, opened 2018-10-15
reduces glare to its original state, eat shit fluffy

ghost:
there was no reason to buff the ability when it was already extremely powerful when it wasn't bugged

making it even more of an easy cheese tool makes shadowlings even harder to deal with, and they're already nigh impossible to deal with

:cl:  
tweak: glare's back to normal stun lengths and being two-tile ranged rather than four-tiles for an instant paralysis
/:cl:

### DIFF
--- a/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
+++ b/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
@@ -75,10 +75,10 @@
 	var/mob/living/carbon/human/M = target
 	usr.visible_message("<span class='warning'><b>[usr]'s eyes flash a purpleish-red!</b></span>")
 	var/distance = get_dist(target, usr)
-	if (distance <= 4)
+	if (distance <= 2)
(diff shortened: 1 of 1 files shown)


fluffe9911:
this is fine but its fluffe not fluffy lol

Kmc2000:
i like cheese

yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
```diff
@@ -75,10 +75,10 @@
 	var/mob/living/carbon/human/M = target
 	usr.visible_message("<span class='warning'><b>[usr]'s eyes flash a purpleish-red!</b></span>")
```
ShadowDeath6, line 76: I think it’s purplish rather than purpleish

ShadowDeath6:
Not really in the scope of this pr but I feel like glare just shouldn’t cast rather than have the stamina drain thing so shadowlings don’t end up essentially wasting their ability because of how small the effective range is

PR #3061, opened 2018-10-16
Makes romerol buyable for nukeops

fluffe9911:
Clown ops already got access to it and it feels like something nukeops would use if its a good idea or not well thats debatable


Ludox235:
No, don't 

I mean Tory already got fungal TB why would they get romerol?

fluffe9911:
I mean its a 25tc item that generally causes mass death seems like something nukeops would use
Also fungal TB and romerol are similar but ultimatly two decently different things its like saying they already got the chameleon projector why would they get the stealth implant you know 
(possibly not best example but you get the idea of my thought process)

ghost:
if there was a thought process involved in this i have yet to see evidence

ThatLing:
I'd rather not

PR #3097, opened 2018-10-17
Adds disposals to 2 more rooms (yogbox)

Ktlwjec:
Gives detective and lawyer rooms disposals, I just copied them across from normal box to yogbox. Yes I tested it yes it worked for me.

image


thehighwayman:
don't make tiles space next time

PR #3125, opened 2018-10-19
Fixes awful Emergency Escape Shuttle Authorize/Repeal spam

Firewolf34:
shuttlecomputer

So I'm sure you all have heard the absolutely horrible racket that a Head can create on the Escape Shuttle when they bash the Authorize and Repeal buttons repeatedly. I have introduced a minor change in this PR that will prevent it from being as bad as it currently is. I've tried to do it in a way that will not affect gameplay.

Note that you can still spam it to a degree, just much more slowly. Some people like to use it as a way to encourage people to get to the shuttle quickly, or just for fun, so I wanted it to still be somewhat of an option.

I've always thought this was extremely annoying and loud and a very easy fix to prevent, but my breaking point was when some child decided to log on during a nice low-pop shift at 4 in the morning, when there was no admins on and _**spam it for three minutes straight**_. Having to turn off the speakers on my computer, disrupting my music and whatever is not something I want to have to do because some idiot thinks he's being funny:
awful

### Changes to the Escape Shuttle Console

The shuttle console now has a (configurable) 10-second delay after _repeated_ "Repeal" actions. Authorizing is not affected, and is still immediate, as it is important that players who are authorizing an early shuttle launch do not have to wait. The first repeal action is also immediate. After that, however, pressing the "Repeal" or "Repeal All" buttons will result in a message being sent to the user telling them that the shuttle console is "recharging" and giving them an indication of how much longer they have to wait.

The recharge time is configurable by changing the definition EARLY_AUTHORIZATION_DELAY_TIME to a specified number of milliseconds, currently set to 10 seconds, which I think is reasonable. 

I've tested it rather thoroughly with multiple IDs and everything still seems to work. I hope y'all approve!

#### Changelog

:cl: Firewolf34
tweak: Added a cooldown timer for Repeal actions on the Escape Shuttle Console, preventing spam
/:cl:


Firewolf34:
I am open to any suggestions for a better message, right now I chose "The emergency shuttle console is recharging, please wait X seconds." because I've seen this format used elsewhere in the codebase. I feel like we could come up with a better verbage than "recharging" however. Not a real big deal, though.

ThatLing (changes requested):
I appreciate the effort you put into this PR but since we have a mirrorbot now, all PRs need to be modularized to save me A LOT of headaches. You can find info about doing this here. If you have any questions, feel free to ping ThatLing on discord.

code/modules/shuttle/emergency.dm
```diff
@@ -1,5 +1,6 @@
 #define TIME_LEFT (SSshuttle.emergency.timeLeft())
 #define ENGINES_START_TIME 100
+#define EARLY_AUTHORIZATION_DELAY_TIME 100
```
ThatLing: Make a file called ``yogstation/code/modules/shuttle/emergency.dm`` and put this in there. Also include it in ``yogstation.dme``.

code/modules/shuttle/emergency.dm
```diff
@@ -10,6 +11,7 @@
 	icon_keyboard = "tech_key"
 	var/auth_need = 3
 	var/list/authorized = list()
+	var/last_early_auth = 0
```
ThatLing: Make a file called ``yogstation/code/modules/shuttle/emergency.dm`` and put this in there. Also include it in ``yogstation.dme``.

```dm
/obj/machinery/computer/emergency_shuttle
	var/last_early_auth = 0
```

ThatLing (changes requested):

MegaEmpirical:
Love the idea. I despise those spammers.

code/modules/shuttle/emergency.dm
Firewolf34: Thanks, I had actually tried to modularize this a bit myself before seeing this and had put the define in this area:
`/code/__DEFINES/~yogs_defines/shuttles.dm`
Because I had noticed the shuttles.dm file existed in the standard TG defines as well, and the link you sent me explained I should place my defines in the ~yogs_defines folder. Any reason this doesn't go into the defines folder? Is it just to make it easier to find this way?

EDIT: On second thought, I just tried to stick the define in `yogstation/code/modules/shuttle/emergency.dm` and I'm getting a compilation error, no such var found. But putting it in the defines file I mentioned previously does seem to fix the problem. Pushing a commit of this now, hopefully this new defines file is appropriate, if not let me know which defines file is preferable for this definition... Thanks a lot for the help!

code/modules/shuttle/emergency.dm
Firewolf34: so we try to keep new variables and procedures and such in the new file, but use code comments if we need to edit the internals of a function?

Firewolf34:
So the cooldown is currently set to 10 seconds. They can make an authorize/repeal alarm every 10 seconds. Depending on what you guys think, this can be extended... I'm just worried that someone might try to repeal in the last 10 seconds of the shuttle countdown and get locked out when they need it. I suppose we'll find out when it gets merged.

Moja10:
Excellent PR, funnily enough despite now being a councilmember, I got noted thrice for spamming in this way a long long time ago... Good to see it finally go away now

PR #3146, opened 2018-10-20
Adds Illegal Biotech

DiamondSentinel:

Adds a new set of Cybernetic Lungs, Military-Grade Cybernetic Lungs. No way to get them right now, but this set of lungs further increases the amount of CO2 and Plasma you can breath before taking damage, as well as increasing your tolerance to Nitryl (brown gas), N2O, BZ, and temperature. 

#### Changelog

:cl:  
rscadd: Added a new set of lungs, Military-Grade Cybernetic Lungs. Also, new equation for Nitryl damage. Functionally the same for default lungs, but adds the possibility for better lungs.
imageadd: Added image for Military-Grade Cybernetic Lungs (lungs-c-m)
/:cl:

Changes the old PR, fully modularizing everything except the lung changes, which cannot be modularized. Also combines a second related change. Second change adds a new Illegal Techweb Node, Illegal Biotech, which allows the Military-Grade Cybernetic Lungs and an Autosurgeon to be created. Illegal Biotech costs a further 5000 research points after Illegal Tech and Advanced Biotech nodes. 

#### Changelog

:cl:  
rscadd: Added a new Techweb node, Illegal Biotech, with new designs for lungs-c-m and autosurgeon.
/:cl:


DiamondSentinel:
Nitryl equation essentially sets a partial pressure of 40 to be the default resistance. That's what it was prior to this merge. All of the other equations essentially set other resistances to scale exactly how it scaled before, just setting it at different thresholds that you start taking damage. For every 20 that you increase the resistance above 40, the player takes damage at every 10% pp higher. Yes, this does mean that you can potentially have a resistance above 100%. No, you will never heal damage while taking Nitryl. A resistance above 100% (that's at 240 resistance), you will never take damage from Nitryl. That's it. 
The if-check that I added checks to see if the damage value from the normal damage calculation would be negative. That's how it prevents healing from Nitryl. That check was also necessary because without it, people with these lungs healed at a stupidly high rate passively, as long as there wasn't Nitryl in the air.

ThatLing (changes requested):
Your code isn't modularized, see here for details on how to do this.

ThatLing (changes requested):
Changing /tg/ icons isn't allowed, make the ones you added in ``yogstation/icons``

PR #3160, opened 2018-10-21
Removes cluwne mask from autodrobe and disables cluwning on equip

ThatLing:
Why is this a thing when cluwnes are valid?

### DIFF
--- a/code/modules/vending/autodrobe.dm
+++ b/code/modules/vending/autodrobe.dm
@@ -8,7 +8,6 @@
 	vend_reply = "Thank you for using AutoDrobe!"
 	products = list(/obj/item/clothing/suit/chickensuit = 1,
 		            /obj/item/clothing/head/chicken = 1,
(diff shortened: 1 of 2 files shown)


fluffe9911:
The one in the autodrobe is a variant of the cluwne mask that can only be found in the autodrobe that only has a 1% chance to turn a person into a cluwne ye meme

fluffe9911:
You are aware of that right? I mean if the 1% thing is bothering you I can just remove that instead of completely removing it from the autodrobe lol

ThatLing:
It's more about the fact that people confuse this with real cluwnes and since cluwnes are valid, kill the person wearing the mask

ShadowDeath6:
why not just make cluwnes not valid

fluffe9911:
Ehhhhh I mean cluwnes also have a cluwne suit and shoes which are ONLY available to actual real cluwnes so if someone kills a person for wearing the autodrobe one its going to be because of ignorance or being silly

Issue #2881, opened 2018-10-05
So... stocks seem still to be unlinked from the cargo order console. When they shouldn't be.


Ratvar:
[Round ID]:  yeah

[The bug] Now it seems to be linked with the bounty one

[Reproduction]: Do stocks thing in cargo, this wyrd thing happens

ThatLing:
I have no idea what you mean, care to elaborate?

ktstudios2000:
The stock market is no longer linked to the cargo console. For example, you could make several hundred thousand on the stock market, but it won't transfer over to the cargo console for use. It's effectively become a minigame.

PR #3166, opened 2018-10-21
Bugfix Adjustments to the Cargo Economy

ktstudios2000:
Due to the addition of the new economy system, a bunch of code related to cargo was left partially functional, with things such as the stock market or the shuttle purchase system relying on the former shuttlepoints variable rather than the new account balance of the cargo department. 

This pull request eliminates those bugs, linking the stock market back up to the cargo console, as well as the communications console. Furthermore, it eliminates a smaller bug wherein a message displayed by the bounty console would show the static shuttlepoints.

I've done some rudimentary testing, and it all seems to be working as intended.

Fixes issue #2881

### DIFF
--- a/code/controllers/subsystem/shuttle.dm
+++ b/code/controllers/subsystem/shuttle.dm
@@ -537,9 +537,10 @@ SUBSYSTEM_DEF(shuttle)
 	if (istype(SSshuttle.shuttle_purchase_requirements_met))
 		shuttle_purchase_requirements_met = SSshuttle.shuttle_purchase_requirements_met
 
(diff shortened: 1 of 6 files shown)


Time-Green:
You might want to PR this to /tg/ instead, it only makes it more difficult to maintain and bugfixes are always merged

Time-Green:
You might want to PR this to /tg/ instead, it only makes it more difficult to maintain and bugfixes are always merged

ThatLing:
What Time-Green said. I'd be willing to do this for you if you want (giving you credits ofcource)

ktstudios2000:
I'd be fine with that. The only reason I didn't attempt a pull request with /tg/ is that I believe the stock market is a distinctly Yogstation feature.

PR #3187, opened 2018-10-23
Makes Yogs Space Hotel great again

Ktlwjec:
1. Makes the pipes look nicer like what Nich did with YogBox.
2. Makes the door bolt buttons actually work on all the hotel guest rooms.
3. Adds window shutters and a privacy button to the top hotel guest rooms.
4. Adds a windoor to the bar so you don't need to go through maint now.
5. Removes some windows in the theatre area for public access.
6. Adds a small flooring detail to show whether the changing room is male or female.
7. Adds a floor tile under all the airlocks to stop the pipes from showing.

image
image

Removed the only external airlock due to Ling's request:
image


thehighwayman:
what was wrong with the old one?

PR #3202, opened 2018-10-25
[WIP] Smoke beacons

Amelia0010:

Adds the smoke beacon™️! 
The smoke beacon™️ is a small smoke grenade, which looks just like a normal smoke grenade, but when triggered it releases a bit of red smoke then explodes 2 seconds after.

#### Changelog

:cl:  
rscadd: The smoke beacon, which looks like a smoke grenade but blows up
/:cl:


Moja10:
Youchy!!

Amelia0010:
And now I've tested it :)

PR #3216, opened 2018-10-26
some atmos things

Ktlwjec:
Moves the pumps to where they are in the image near the windows and changes some id tags.
image

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -40904,6 +40904,15 @@
 "bWQ" = (
 /turf/closed/wall/r_wall,
 /area/security/checkpoint/engineering)
+"bWR" = (
+/obj/machinery/atmospherics/pipe/simple/scrubbers/visible,
+/obj/machinery/atmospherics/components/binary/pump{
+	dir = 4;
+	name = "O2 Outlet Pump";
+	on = 1
+	},
+/turf/open/floor/plasteel,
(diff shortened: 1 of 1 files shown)


thehighwayman:
you broke it goddamn

MegaEmpirical:
Parth had moved the pumps further out due to concerns about a bug where gases from pumps would mix with pipes overlayed on them but not connected. I am unsure if the bug still exists.

Nickvr628:
Partheo is correct, that bug might be around. Also, the output of the N2+O2 mixer should be cyan to indicate it is the breathable mix. 

I do like how clean it looks though, it is nice

Ktlwjec:
image
I didn't die

thehighwayman:
did you die?

MegaEmpirical:
Good enough for this guy. By the way, was there something changed about the tube to tcomms? I can't get to the left side of atmos in space like I could before.

thehighwayman:
> Good enough for this guy. By the way, was there something changed about the tube to tcomms? I can't get to the left side of atmos in space like I could before.

The way she did the tunnels makes it impossible

MegaEmpirical:
I've noticed. Also, seems like the 'random' room to the top right isn't random, it's always dusty detective.

thehighwayman:
> I've noticed. Also, seems like the 'random' room to the top right isn't random, it's always dusty detective.

Thats cause she didn't make it random. I don't think she knows how

PR #3217, opened 2018-10-26
Updates the logs for Supermatter

thehighwayman:
Originally was going to do more for this so that doing investigate Supermatter you could easily tell who did stuff and what they did but this is an improvement until I do that. also need to add SM countdown logs too which shouldn't be too hard
#### Changelog

:cl:  
rscadd: Makes SMlogs slightly better  
/:cl:


Ktlwjec:
finish your prs, bad coder

PR #3194, opened 2018-10-23
Adds Some Old Some New and Some Classic Songs to the Lobby Screen!

fluffe9911:
We really do need more music and I really liked some of the old music we had here is a list of the new songs

https://www.youtube.com/watch?v=s7dTBoW5H9k - Mr Blue Sky (Old Yogs)
https://www.youtube.com/watch?v=CfMMlT8Lyns - Get Lucky 8 Bit (Old Yogs)
https://www.youtube.com/watch?v=dLrdSC9MVb4 - Turn the Lights Off
https://www.youtube.com/watch?v=YAXc87bmgLA - Nightcall
https://www.youtube.com/watch?v=UPHmazxB38g - Thunderdome (Old Yogs)
https://www.youtube.com/watch?v=pLJ85XExZtQ - & (amazing song title)
https://www.youtube.com/watch?v=At8ao5iABFg - Ruler of Everything
https://www.youtube.com/watch?v=VGQZItOMz6k - Good Day
https://www.youtube.com/watch?v=YgGzAKP_HuM - RASPUTIN
https://soundcloud.com/jeffimam/title-plasma-attack - Plasma Attack (Made by Moja10 Old Yogs)
https://www.youtube.com/watch?v=L0VjlFqp47g - Next Town Over
https://www.youtube.com/watch?v=KKny_ia8Fvo - Dont Stop Me Now
https://www.youtube.com/watch?v=iYYRH4apXDo - Space Oddity  (Old Yogs im pretty sure I cant find it for the life of me but I swear to god it was)

PR #3267, opened 2018-10-28
Makes lobby music use youtube-dl and goonchat

1fbff5f83b23d39d38b1dfcb4cac8d9b:
This lets us add any song that is supported by youtube-dl as lobby music. This way we can add trillions of songs, while reducing the resource download size.

You can set the volume using the admin song slider.

The sample config includes all songs from #3194 which weren't unavailable on youtube.


1fbff5f83b23d39d38b1dfcb4cac8d9b:
TIL pick from empty list is a runtime

ghost:
at last my time has come

PR #3295, opened 2018-10-31
Adds Surgical Drapes to the Protolathe

Firewolf34:
drapes

### Rationale

I got tired of random people (read: greyshirts) stealing my surgical drapes for whatever reason, and having to get into a war with medical/robotics over who gets to use the surgical drapes. Add to that the fact that surgical drapes deplete after a while, and are not able to be produced in any fashion that I know of, and you realize pretty quickly why it would be nice to make them in the protolathe like all the other tools can be. However, MAT_PLASTIC doesn't get used all that often, so I wanted there to be an additional reason for the Chemist to make Plastic. Now they can be asked to by the surgeons when they find out their drapes are missing. 

### Production Costs

The drapes cost 2500 MAT_PLASTIC, despite what my commit message says (whoops). Under this cost value, you can produce 8 drapes from 5 plastic sheets, to put it into perspective. This way you won't have to steal bedsheets from the dorms to perform ghetto surgeries in your under-equipped medbay.

### Possible Side-Effects of Change

As a side effect, you can now get some small amount of plastic at the beginning of the game by sacrificing one of the two drapes you receive (three if you count transfer centers drapes, on box at least) in the Destructive Analyzer, where they will be converted to 1250 plastic units.

### How to Use this Change

The Surgical Drapes are available after researching the 'Basic Tools' techweb node. Both Science and Medical Departmental Techfabs are capable of producing them.

Surgical Drapes can NOT be produced by the autolathe however because the autolathe currently does not support plastic AT ALL. I feel like this should be changed eventually, but it wouldn't be worth it unless a lot of the items in the autolathe used plastic, which I'm sure would piss off quite a few people.

#### Changelog

:cl: Firewolf34
tweak: Added Surgical Drapes to the Protolathe, requires plastic and Basic Tools research.
/:cl:


Ktlwjec:
use a bedsheet

ShadowDeath6:
Good change, although I’m not sure basic tools would be the best place in techwebs for it. Is there a tier 1 medical research node? I think it would be better suited there

thehighwayman:
use a bedsheet

Firewolf34:
@ShadowDeath6 The reason I put it into Basic Tools is because that's where the rest of the medical supplies are. Scalpel, Cautery, Circular Saw, etc. The description for "Basic Tools" is "Basic mechanical, electronic, **surgical** and botanical tools."

It does seem strange to have the surgical tools in all one category, and it would be nice to possibly move them to their own research node, but I believe that the reason this was chosen is because Science also needs surgical tools so they figured they'd put it in one "general tools" node.

I think it would be strange to have the surgical drapes separate from the rest of the surgery tools, don't you think? But I'm very much interested in anyone's input on this. I suppose doing so might make the surgical drapes into some kind of advanced surgery tool somehow, and I feel they're pretty basic. As Ktl said we can even use bedsheets for the task...

code/modules/research/designs/autolathe_designs.dm
```diff
@@ -435,6 +435,8 @@
 	category = list("initial", "Medical", "Tool Designs")
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
 
+// Wondering where the surgical_drapes are? See yogs medical_designs, because the Autolathe does not support MAT_PLASTIC yet!
```
ThatLing: Why?

ThatLing (changes requested):

code/modules/research/designs/autolathe_designs.dm
Firewolf34: I'll remove it. I just figured the comment would help people find the surgical drapes because that's where the rest of the surgical tool designs are and where coders might go looking for them. I mean considering the number of meme comments in the codebase, I wouldn't call that one too obnoxious, right? lol

Firewolf34: Fixed. :)

PR #3297, opened 2018-10-31
Fix mecha salvage exploit

### LINKED ISSUE #3292: A resource generation exploit to obsolete Mining once and for all
## Round ID: N/A, works in all versions of Yogstation-TG
works on-server and with the codebase fresh off the repo, in all shifts, I've thoroughly "tested" this one ;)

### This bug will WRECK you.

So I'll be straight, as the Master & King of all Roboticists, Kendrick Dogg himself, it was really hard for me to report this one. Infinite metal and plasteel? Do you have any idea how many borgo slaves I could make with that. But when BWOINK tells me to jump, I ask how high. And using it made me feel dirty.

This is the exploit that obseletes Mining, finally freeing Science from the shackles of dirty men with survival knives in their boots who never do their damned job.

## Infinite Resources:

If you destroy a Ripley (and I imagine this exploit works with other mech types too) it will turn into a wreck which looks like this:

ripley

One of the "features" wrecks have is they can be "salvaged" for parts. All you have to do is weld it. It doesn't even take any time to perform. And whoever coded this in, god bless 'em, did not put a limit on how many times you can do this. I salvaged until my index finger developed repetitive strain injury but could not reach the center of the metaphorical tootsie-pop with this one, the mech wrech will just never go away. Here's what happens when you dump an entire welder into it in under a second:

random

I guess they figured this was okay because it adds a bit of random chance as to whether or not you'll actually get anything from it when you try to weld it as you can see. But that's nothing to a determined Roboticist who's being cargoblocked all shift, here's what I could generate in 5 minutes:

metals

I ran into someone who said he knew of this one before, so maybe it's been around for a while, but I'm damn surprised more people haven't discovered this one yet. Pretty amazing.

I don't have any photos of this, but I did manage to get a few assistants together to work my sweat shop salvaging metal off the corpses of exosuits. It actually generates more plasteel than metal sometimes, which is an insane amount of plasma early-game... NT should give me a promotion. There's real profit potential here to the enterprising spessman. Should be an easy fix, though.

ThatLing:
Fixes #3292

### DIFF
--- a/yogstation/code/game/mecha/mecha_wreckage.dm
+++ b/yogstation/code/game/mecha/mecha_wreckage.dm
@@ -13,6 +13,7 @@
 	var/list/welder_salvage = list(/obj/item/stack/sheet/plasteel, /obj/item/stack/sheet/metal, /obj/item/stack/rods)
 	var/list/wirecutters_salvage = list(/obj/item/stack/cable_coil)
 	var/list/crowbar_salvage = list()
+	var/salvage_num = 5
 	var/mob/living/silicon/ai/AI //AIs to be salvaged
 
 /obj/structure/mecha_wreckage/Initialize(mapload, mob/living/silicon/ai/AI_pilot)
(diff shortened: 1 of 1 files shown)


Ktlwjec:
you're too good for us

ThatLing:
Better fix coming up

PR #3299, opened 2018-10-31
Adds more wiki manuals to yogbox!

Ktlwjec:
Adds the wiki manuals from box to yogbox.

https://github.com/yogstation13/Yogstation-TG/pull/2152 these


thehighwayman:
maybe work on our own wiki before porting over TG wikis hmm?

PR #3326, opened 2018-11-01
Some final fixes and tweaks for atmos

### LINKED ISSUE #2719: Atmos blast doors do not cover the plastic flaps
Since the flaps do not have a blast door, atmos is always one screw driver away from all access. There are no internal bast doors so they only need remove a console and window to get past the control room. I don't recall this being an issue in the past, I think pre-rebase flaps were not deconstructable or something?

### LINKED ISSUE #3117: Issue with pressure in N2O tank/loop in Yogstation
Issue reported from  Round ID: 19183 (Yogstation 13)

Reporting client version: 512

Not sure, but it already happened second time to me. After i modified loops so you can directly connect canister to output of each tank, when i connect tank to N2O output it instantly fills with 50k pressure gas.  Problem did not appear with other gases even though set up is the same. Checked configurations of terminal and everything looks fine. For some reason N2O loop pressurizes with all gas available i guess.

### LINKED ISSUE #3222: Gas flow meters are not visible on atmos' large tanks
There is a gas flow meter on each of the large holding tanks in atmos, both for gases coming in and going out but they're not visible. They used to be at some point.

image

Partheo:
Moved some pipes around to be more aesthetically and thematically pleasing, replaced the second atmos alert console in the office with another mix tank monitor, fixed some pumps which were named incorrectly, added a waste outlet that vents to space(with digital valve), moved the transit tube loop to the left slightly so the tanks are accessible, added an outlet pump to the mix tank, rotated the consoles correctly, changed the color of a pipe, and made sure every sensor, monitor, and inlet/outlet pump functions correctly.
image
image

Just a bunch of shit, essentially. 

I tested the possible bug where pumps placed on top of perpendicular straight pipes causes cross-contamination, but was not able to reproduce it. It seems to work fine. I'll leave the pumps as they are for the time being, and we can work on a fix if it actual becomes reproducible.
image

fixes #3117 
fixes #2719 

Also the tank meters are visible now.
fixes #3222 

:cl:  Partheo
tweak: Various small tweaks to the atmos department
bugfix: Gas meters are now visible on atmospherics' gas tanks.
bugfix: Various fixes to the atmos department
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -54570,12 +54570,6 @@
 	dir = 4
 	},
 /area/engine/atmos)
-"myB" = (
-/obj/machinery/atmospherics/components/unary/portables_connector{
-	name = "Pure Gas Connector Port"
-	},
-/turf/open/floor/plasteel,
-/area/engine/atmos)
 "myI" = (
 /obj/machinery/atmospherics/components/unary/vent_pump/on/layer1,
 /turf/open/floor/plasteel/grimy,
@@ -96578,7 +96572,7 @@ bOd
 wdE
 xei
 bOd
-myB
+bOd
 ppr
 tqk
 fuJ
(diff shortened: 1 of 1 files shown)


MegaEmpirical:
Glorious. Thank you.

PR #3327, opened 2018-11-01
I removed circuits from yogbox because I don't like them [NOT CLICKBAIT]

Ktlwjec:
In case #3316 gets merged.

image

:cl:  
rscdel: Circuits have been removed from the Yogbox Science department.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -7845,6 +7845,9 @@
 /obj/structure/cable{
 	icon_state = "4-8"
 	},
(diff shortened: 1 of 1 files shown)


Nickvr628:
no fun allowed

CMOisLing:
Rip, also if your going to add two nanite stations and cloud servers at least add another programmer or whatever the one that sets the program settings is called

Ktlwjec:
Not needed but I need to remove the thingies from the left-side table anyway so I'll see how it looks with another one in.

Amelia0010:
ok nevermind I guess

PR #3373, opened 2018-11-06
Adds Stock Market Computer to PubbyStation and DeltaStation

ktstudios2000:
I've tested both of these in-game, and as you would expect, they make minor map adjustments that result in the stock market computers appearing in both. For DeltaStation, one is placed in the general cargo area and the other in the QM's office, and for PubbyStation, one is placed in the general cargo area.

Aside from my concerns about 17,867 additions vs 321,100 deletions, I have no reason to believe this won't merge fine.


Ktlwjec:
@Youbar Did you mapmerge?
Also maybe would need yogs map versions

ThatLing:
We don't have yogs versions of these maps.

PR #3390, opened 2018-11-07
fixes things

Amelia0010:

I cannot think of a single situation where they should not be able to do this. Even if you've got them entirely under with BZ, you can just cuff them and they won't be able to stop the BZ when they wake up and you can just keep them under forever anyways.

#### Changelog

:cl:  
tweak: fakedeath is now free as a ling
/:cl:


ShadowDeath6:
I think being absorbed sets chems to 0, check if you can still not revive after being absorbed by another ling

PR #3397, opened 2018-11-08
Updates tank sprites

Partheo:
Went back to these old sprites and did a quick update. They really lacked contrast.
I also added the improvised jetpack from TG, not sure if we use it or not.

GIF comparison:
tanks

The improvised jetpack is the one with the white stripe here. Not shown in the GIF.
image

:cl:  Partheo
imageadd: Air tanks and jetpacks have been given a fresh coat of polish.
/:cl:


mrhugo13:
Partheo if I could physically kiss sprites I would kiss every single sprite you have ever made.

PR #3415, opened 2018-11-08
Vacant commissary, detective and lawyer changes (yogbox)

Ktlwjec:
#3405 
Tweaked a few things:
Commissary APC wiring is actually linked.
Detective office now has a spawner.
Clerk access added to commissary.
Obviously the commissary is not near cargo, as we have our vault there.

image


thehighwayman:
lovely update to our map

Nickvr628:
Swap the places of the maint hallway and the commissary so that the clerk store shares a wall and can be one big space.

PR #3395, opened 2018-11-08
Actually Fixed Mecha Wreckage

Cruix:
Fixes #3372

PR #3416, opened 2018-11-08
Actually Actually Fixed Mecha Wreckage

### LINKED ISSUE #3372: Free miner mech wreck won't contain enough parts to build another mech
I've had multiple complaints about this, probably has something to do with the wreckage still rolling rods/plasteel/metal sheets

ThatLing:
Fixes #3372 
Closes #3395

### DIFF
--- a/yogstation/code/datums/ruins/free_miners.dm
+++ b/yogstation/code/datums/ruins/free_miners.dm
@@ -75,7 +75,7 @@
 /obj/effect/mob_spawn/human/free_miner/engi
 	name = "Free Miner Engineer"
 	id_job = "Free Miner Engineer"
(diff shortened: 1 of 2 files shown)


Cruix:
Cool. 20 is pretty high, but the engi would have to be pretty stupid to burn all the salvage on that. You should new() the boards so they are actually retrievable.

Amelia0010 (changes requested):
what cruix said I guess

ThatLing:
salvage_num is the total amount of times you can salvage the wreck and there's a 70% chance you'll fail to retrieve anything. Boards are already new'd in ``/obj/structure/mecha_wreckage/attackby``.

Cruix:
Crowbar salvage is not new()'ed on tg code, it expects an instantiated object.

PR #3471, opened 2018-11-12
mapmove3

Amelia0010:
turns out mapmerge don't work if it's in the yogstation folder, so we'll just have a yogstation folder in _maps I guess


Amelia0010:
if travis ok's i'm self-merging this

Nickvr628:
But the 24 hour rule

Amelia0010:
go eat a goat

Issue #3435, opened 2018-11-11
No vault camera on sec console


00aquir:
map Yog Station

Ktlwjec:
Copied camera tags from box and vault had a "vault" tag only, so not technically a bug but I could add it on I guess.

PR #3473, opened 2018-11-12
Windoors to yogbox + yogmeta

Ktlwjec:
#3437 
#3483 
Fixes issue #3435 
Also fixes a disposals pipe


thehighwayman:
won't let me flirt with nich so everything you do is bad

PR #3493, opened 2018-11-14
Updates bluespace lockers

monster860:
https://streamable.com/t0ulh

:cl: monster860
fix: GPS and telecomms inside the bluespace locker now act as if they at the location the bluespace locker is
add: You can now see out of the bluespace locker. It looks a bit janky though. Aliens coded the holograms hastily I guess.
/:cl:


1fbff5f83b23d39d38b1dfcb4cac8d9b:
Now add some way to steer it and we'll have spaceships

PR #3529, opened 2018-11-16
Fixes cyborg soap being used up

### LINKED ISSUE #3494: Soap borg item disappears
Standard borg module - Soap item can "crumble away" after enough uses which I assume is intentional for soap, but this should not happen to borg versions of the item.

Ktlwjec:
Added a soap identical to the nanotrasen soap but with infinite uses. Gave it to borgs that use the soap module. Tested it and it all seems to work correctly.

Fixes #3494

### DIFF
--- a/code/game/objects/items/clown_items.dm
+++ b/code/game/objects/items/clown_items.dm
@@ -58,15 +58,6 @@
 	cleanspeed = 28 //janitor gets this
 	uses = 300
 
-//Yogs start - Adds infinite soap
-/obj/item/soap/infinite
-	desc = "A heavy duty bar of Nanotrasen brand soap. Smells of plasma."
-	grind_results = list("plasma" = 10, "lye" = 10)
-	icon_state = "soapnt"
-	cleanspeed = 28
-	uses = -1
-//Yogs end
-
 /obj/item/soap/homemade
 	desc = "A homemade bar of soap. Smells of... well...."
 	icon_state = "soapgibs"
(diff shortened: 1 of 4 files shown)


Bawhoppen:
Shouldn't an upstream issue like this be fixed upstream though?

thehighwayman:
> Shouldn't an upstream issue like this be fixed upstream though?

you make a good point but if I may counter: Fuck TG

ThatLing:
lol just set the uses to something really high

Ktlwjec:
then it is not infinite ling

code/game/objects/items/clown_items.dm
```diff
@@ -58,6 +58,15 @@
 	cleanspeed = 28 //janitor gets this
 	uses = 300
 
+//Yogs start - Adds infinite soap
+/obj/item/soap/infinite
```
ThatLing: Do this in ``yogstation/code/game/objects/items/clown_items.dm`` and just set the uses to something really high. The ``if(uses != -1)`` checks are really stupid.

ThatLing (changes requested):

code/game/objects/items/clown_items.dm
Ktlwjec: are you sure because like soapstone, crayon and spraycan have uses = -1 for infinite amounts, and it's not really infinite if it's just a high number

alexkar598: Admin rcd is infinite but not really,it’s a very high number and I couldn’t get to empty it

Time-Green: Use the INFINITY label, it’s high

Ktlwjec: yeah, I will, but still not true infinity :disappointed:

Time-Green: You cannot drain the soap within a human lifetime

Issue #3371, opened 2018-11-06
A Number of Bugs I've Observed


ktstudios2000:
I've accumulated a list of bugs that I've noticed in the game. Unfortunately, I don't have any specific round numbers attached to them, but they can be reproduced very easily and consistently.

- ~~When welding a wall without a welding mask, the "your eyes sting" notification will spam in 0.5 second intervals. Initially, I thought this was an intentional change, but it isn't replicated when welding literally anything else.~~
- ~~You are unable to butcher dead mice. If they are living, it isn't possible either. I figured this is because the moment they die, they are converted to an edible food reagent, making butchering them impossible as a proc doesn't exist for it.~~ Added as a PR to tg-station
- If you cancel a mining action, the game will treat it as if you're still mining, making it impossible to continue mining until the time it would have taken to mine has expired (i.e. mine a rock, cancel it, wait 5 seconds before you can mine it again). This is due to lines 60-62 in /code/game/turfs/simulated/minerals.dm.
- ~~As a changeling, fleshmend, or whatever the healing ability is, didn't work. I spammed the ability many times, but despite that, my health wouldn't change in the immediate short-term or even long-term.~~ Unable to recreate on my branch of Yogstation, meaning it was probably fixed since I last observed it happen
- ~~Clipboards don't show the paper items placed inside it. For example, stamp a piece of paper in cargo, and place it inside a clipboard, and you'll only see the stamp's overlay.~~ Added as a PR to tg-station
- ~~All maps other than the default yogstation do not contain a stock market computer.~~ Unable to be changed due to PubbyStation and DeltaStation being tg maps.
- The disease appendicitus does not have a 'type' of virus (such as airborne) upon being viewed through a handheld medical scanner
- All of the minebot upgrades are accepted, but are not recorded nor do they take up capacity on the minebot
- Minebots are damaged by ash storms (intentional?)

SomeguyManperson:
all channeled weldings cause a strong flash, followed by multiple weak flashes as of #414

ktstudios2000:
Assuming these are upstream issues, I've posted a PR to tg-station to fix issue 2 (mice) and 5 (clipboards).

Altoids1:
Atomize your damned issues. Sheesh.

PR #3550, opened 2018-11-16
Fixes some text when using medical scanners to check viruses

Ktlwjec:
One of the issues in #3371 .

Adds spread_text for some diseases that were just showing up as ".".
Changes possible cures from "nothing" to "Unknown".


thehighwayman:
unknown will confuse the new players

PR #3565, opened 2018-11-18
Makes the pinpointer pinpoint to the bluespace locker if necessary

### LINKED ISSUE #3561: Pinpointer Does not Point to Bluespace Locker when Disk is inside
## Reproduction:
Pinpointer Does not Point to Bluespace Locker when Disk is inside it

monster860:
:cl: monster860
fix: If you put the nuke disk inside the bluespace locker then the pinpointer will pinpoint to the bluespace locker instead of not pinpointing to anything at all
fix: you can no longer space the nuke disk using a bluespace locker.
/:cl:

fix #3561


thehighwayman:
good fix. could see an anti-nukie strat develop out of this tho.

PR #3569, opened 2018-11-19
Fixes the in-hand sprite for chef hats

Ktlwjec:
it was set to the chef outfit instead of the chef hat which was stressing me


thehighwayman:
didn't tell me you were doing this

Amelia0010:
upstream?

PR #3610, opened 2018-11-21
Makes toxins better

monster860:
image

:cl: monster860
add: Adds a toxins mixing tank computer to toxins
tweak: Makes toxins room bigger
add: Adds h/e pipes in the toxins mix chamber.
/:cl:


Nickvr628:
I like it.

Amelia0010:
image

why not just run these through the middle section instead, that way they don't got through waaaaalls

Amelia0010:
I can't even compile this locally, I'm sorry but I'll try to recreate this on a fresh copy and make a new PR for ❤️

PR #3635, opened 2018-11-25
Adds a windoor to the virology plastic flap

Ktlwjec:
title

:cl:  
rscadd: The virology plastic flap is now windoor protected.
/:cl:


thehighwayman:
now I don't have to do this in game

PR #3640, opened 2018-11-25
Adds a goat plushie and fixes the surgery drape crafting

fluffe9911:
In another pr adam is going to cuss at me for I ported over that goat plushie from hippie looked interesting and if I dont do it now while its still semi popular its never going to be a thing so *shrug* 

Made by karma ported by me sorry for forgetting to credit you man


code/modules/cargo/packs.dm
```diff
@@ -1980,6 +1980,7 @@
 					/obj/item/toy/plush/lizardplushie,
 					/obj/item/toy/plush/snakeplushie,
 					/obj/item/toy/plush/nukeplushie,
+					/obj/item/toy/plush/goatplushie/angry,
```
Amelia0010: make the plushiecrate have the non-angry version and i'll merge it, not gonna merge it now since it makes the plushiecrate attack people when opened.

Amelia0010 (changes requested):

PR #3643, opened 2018-11-27
Allows golems to smoke :)

Ktlwjec:
Jojorio: NEW PR LET GOLEMS SMOKE WHEN

:cl:  
rscadd: Golems can now smoke.
/:cl:


thehighwayman:
golems don't have lungs how do they smoke?

ShadowDeath6:
Does this do anything since they don’t have lungs? I would say to blacklist smoking for all nobreath races if it does still work, since that doesn’t really make sense. 

actually not even blacklist, but just make smoking not do anything while still allowing them to smoke

ThatLing:
They have no lungs, this breaks my immersion

PR #3651, opened 2018-11-27
Disposals Wires

monster860:
I made it so you can hack sorting junctions!

Basically, the way it works is really simple. There's the forward wire, side wire, and scan wire. The forward and side wires, when pulsed, set the current "sorting state" of the pipe to their respective zones. Forward is red and side is green. The scan wire, when cut, disables scanning, and when pulsed disables it for only 10 seconds. When scanning is disabled, the pipe does not scan the package for whether or not it should sort. Instead, it just reads the sorting state.

So in order to hack a disposals pipe, you need to disable scanning, and then specify a direction to travel.

Includes really shitty wire-panel-open sprites made by me:
image
It looks like someone dragged a dead body into the pipes. In fact, that yellow-grey blob is actually supposed to be a cover, and the red lines are supposed to be wires.

things you can do:

- set up a trap using a disposals loop and a proximity sensor that causes it to spit out it's contents at you when you approach. Feed it with sharp objects, and find an unsuspecting target. You'll have to be fast though or it might get stuck.
- hide under the floor, then use a remote signaller to pop out at the right time and surprise someone with an e-sword
- permanently redirect all trash, including you when you jump down the disposals chute, into the RD's office. He might get suspicious though, from all the trash coming out of his trash can.
- Make a disposals loop, get a bunch of people inside, and then whack the button to redirect them into a pipe that spaces them at the same time. 
- Temporarily redirect disposals into space for when you're disposing of a body. Or even better, make a disposals pipe into the supermatter crystal while the CE isn't looking.
- Setup a remote signaller for *all* the areas and fill a few cardboard boxes with labeled signallers! Become the disposals god!
- I dunno, you figure shit out.

here's a 3-minute video demonstrating this feature: https://streamable.com/frfar

:cl: monster860
add: Disposals sorting pipes can now be hacked! Force them to always on or always off, or even temporarily force it with a remote signaller.
/:cl:


Moja10:
amazing shit

yogstation/code/modules/recycling/disposal/pipe_sorting.dm
```diff
@@ -0,0 +1,46 @@
...
+		sort_scan = TRUE
+
+/obj/structure/disposalpipe/sorting/update_icon()
+	cut_overlays()
+	. = ..()
+	if(panel_open)
+		add_overlay(image('yogstation/icons/obj/atmospherics/pipes/disposal.dmi', "[icon_state]-open"))
+
+/obj/structure/disposalpipe/sorting/attackby(obj/item/I, mob/user)
+	if(I.tool_behaviour == TOOL_SCREWDRIVER)
```
Amelia0010: screwdriver_act

Amelia0010 (changes requested):

yogstation/code/modules/recycling/disposal/pipe_sorting.dm
Amelia0010: other than this it's fine

PR #3655, opened 2018-11-28
Lowers vampire blood milestones 

fluffe9911:
Im trying to make this gamemode fun alrighty?


Amelia0010:
what's the worst that could happen

PR #3657, opened 2018-11-28
Changes brig piping slightly (yogbox)

Ktlwjec:
Added a vent and scrubber to brig. Disconnected the hallway piping from the piping leading out of the courtroom.
image

:cl:  
tweak: Brig atmos is no long connected to the outside hallway in yogbox.
tweak: Brig has a vent and scrubber outside the cells in yogbox.
/:cl:

PR #3658, opened 2018-11-28
what my other pr was

Ktlwjec:
I have no clue why it was conflicting but even a fucking map copy from yesterday's updates weren't fixing it so here is #3657 again.

:cl:  
tweak: Brig atmos is no longer connected to the outside hallway in yogbox.
tweak: Brig has a vent and scrubber outside the cells in yogbox.
/:cl:


thehighwayman:
It works!

PR #3659, opened 2018-11-29
Adds bluespace pipes

monster860:
They were added to tg a while ago, but they made it so you could make them for free using RPD's which got them removed. This PR adds them here but with the following changes:

- you need to make them using science
- you can specify the name of a bluespace pipe network to attach to

:cl: monster860
add: Adds bluespace pipes. You make them using science, and you can specify a network to attach to by using it in your hand.
/:cl:


Amelia0010:
oh no

ghost:
Still really strong and considering everyone researches bluespace tech ASAP it's not much of a gate. Increase the diamond costs.

Amelia0010 (approved):
don't see anything immediately wrong, but I'm wondering whether these use layers or not

monster860:
unless there's a way to change the layer of a pipe item, no it doesn't work with layers.

Amelia0010:
ok cool

PR #3685, opened 2018-11-29
flashbangs actually kill blob tiles now same as before because yog is retarded

ghost:
title

:cl:  
rscadd: flashbangs hurt blobs again
/:cl:

### DIFF
--- a/code/game/objects/items/grenades/flashbang.dm
+++ b/code/game/objects/items/grenades/flashbang.dm
@@ -16,6 +16,11 @@
 	new /obj/effect/dummy/lighting_obj (flashbang_turf, LIGHT_COLOR_WHITE, (flashbang_range + 2), 4, 2)
 	for(var/mob/living/M in get_hearers_in_view(flashbang_range, flashbang_turf))
 		bang(get_turf(M), M)
+	
+	for(var/obj/structure/blob/B in get_hear(8,flashbang_turf))     		//Blob damage here -- Yogs
+		var/distance = get_dist(B, get_turf(src))
+		var/damage = round(100/(distance*distance)+1)
+		B.take_damage(damage, BURN, "energy")
 	qdel(src)
 
 /obj/item/grenade/flashbang/proc/bang(turf/T , mob/living/M)


ShadowDeath6:
Lower the damage if you’re bringing this back, it’s too easy and risk free to rip the blob apart with flashbangs

ghost:
in practice this just makes the crew die when used because it's usually just one guy throwing a flashbang or throwing a flashbang right next to the blob, and either the blob eats it or the single flashbang he threw goes off for medium damage and then the blob kills everyone stunned

Smudged:
As long as energized blob takes 2x damage from flash bangs, I'm alright with this. 

Issue like what doe said is that, I mean, it's negligible if the crew/sec gets protection from flash bangs and just orders a billion boxes of flashes. 

We'll need to see how powerful it is once it's merged and adjust from there.

Ratvar:
It will be as powerful as it was before - roundstart bomb rifles but you can use all shots at once and you can move while shooting.
Is there any reason they should be added but "butthurt adam made a PR 5 seconds after being proved wrong in OOC" ? Anyone remembers any adam's PR that was good and not "i ded plz nerf/buff"? Like... liver damage increase that made every toxin 100 times more deadly because reasons?

ghost:
I wish your comment was understandable in any form so that I could respond to it. Anyone speak Kojuh here?

ShadowDeath6:
I would rather it be too weak at medium range than too strong at medium range. Making flashbang spam able to completely destroy 5 or 6 blob tiles not even adjacent to it (1 tile away still deals 25(26?)) damage, which almost shots 30 health normal blob tiles. 
The blob still has to manually expand to break it (unless the person throwing it is bad and doesn't know what nodes do), spending blob points to stop the single flashbang, meaning less BP for shields, factories, resource, etc. If the person can time the flashbang right it's very hard for the blob to react quickly enough to the throw as well. One flashbang being able to deal 20+ BP worth of damage to the blob is shitty.
It's extremely easy for someone to yell "throwing flashbangs" and then clear out an absurd amount of light tiles so everyone with guns and welders can go in and clean up the tanky stuff. 

I would say that in their current state they're way too good at mass deleting light tiles and are too much of a hassle to deal with even when they don't go off, and even though it's very possible someone bad can get their hands on them, most of the time even when utilized badly they'll work too well (remember that dist 2 deals 25 damage, so 2 badly aimed flashbangs will wipe out a huge chunk of blobs).

tl;dr 2 ez 2 strong make it scale off significantly more quickly, possibly double the amount that it does now

code/game/objects/items/grenades/flashbang.dm
```diff
@@ -16,6 +16,11 @@
 	new /obj/effect/dummy/lighting_obj (flashbang_turf, LIGHT_COLOR_WHITE, (flashbang_range + 2), 4, 2)
 	for(var/mob/living/M in get_hearers_in_view(flashbang_range, flashbang_turf))
 		bang(get_turf(M), M)
+	
+	for(var/obj/structure/blob/B in get_hear(8,flashbang_turf))     		//Blob damage here -- Yogs
+		var/distance = get_dist(B, get_turf(src))
+		var/damage = round(100/(distance*distance)+1)
```
Amelia0010, line 22: image

dividing by 0 runtimes, so if the flashbang is on the same turf this'll break :)

Amelia0010 (changes requested):

code/game/objects/items/grenades/flashbang.dm
ShadowDeath6: To be fair it’s impossible to get a blob and a flashbang on the same tile without teleport shenanigans

Bawhoppen:
Haha. Flashbangs.

ghost:
it was removed for balance concerns on /tg/ before beam rifles were a thing and before blobs got counters to that with reflector blobs 

/tg/, where people have two brain cells to bang together

we don't have that, so flashbang use is likely to be destructive even in the hands of someone who knows how to time a throw because they will flashbang everyone else who won't clear out

and throwing it next to the blob still leaves opportunity for the blob to eat it, so flashbangs either have to be used with great precision in a room with atmos and then capitalized on, or have little to no effect entirely

yes, they're good used correctly, but they're also one of the only actual means to deal with a blob, since welders are a joke and lasers don't do much once it's past 3 nodes

ShadowDeath6:
they shouldn’t be a method to deal with the blob- they should be support in order to thin it out so people can actually get to stuff. I’m fine with flashbangs being aoe damage to the blob, but being able to delete everything that isn’t a tank blob tile with a few flashbangs that don’t even need to be adjacent is cancerous

also remember that the blob still needs to manually spread onto the flashbang to eat it which costs bp (further limiting the resources it has) and needs to be done quickly because a flashbang will oneshot every blob adjacent to it if it goes off

also, reflector blobs should not be a justification against beam rifles- it’s one of tons of chemicals, and it has its own weaknesses. Beam rifles are pretty effective on blob otherwise. 

Again: too little risk to use flashbangs if you just say “throwing flashbang” and too detrimental to the blob no matter what happens because it’s either wasting bp or it’s losing a fuckton of tiles

PR #3720, opened 2018-12-02
I think this fixes pipes being invisible but nich didn't describe it at all so I have no idea (actually this fixes pipes not compiling apparently, thank you nich for telling us that in the first place)

Amelia0010:
fixes stuff


ThatLing:
good code

Amelia0010:
no, this fixes pipes not compiling

PR #3741, opened 2018-12-03
Ethereals are now charged up slightly when stunned by batons.

AsV9:
Tiny fluff change, but I find it amusing.
It doesn't affect stun length or anything else.
It is tested and I found no errors.

:cl:  
rscadd: Stunbatons now charge ethereals slightly.
/:cl:


ShadowDeath6:
if this doesn’t already work with taser electrodes as well I would do that

AsV9:
Sure I think I can do that

PR #3759, opened 2018-12-06
Adds undead adaption to sentient disease under powerful tier list

ghost:
big brain aids infect corpse

because ideally forcing medical staff to also dispose of dead bodies that the disease managed to create as biohazards to prevent infection is both thematic and funny, and it's locked under powerful tier because corpses bursting into flame could create serious obstacles

doesn't necessarily mean that the virus can infect people from corpses, though-if it was airborne dead or not being next to the dead dude and touching him with your naked hands will infect you, regardless of having this symptom

#### Changelog

:cl:  
tweak: Sentient disease can adapt to infect zombies, skeletons, and corpses now-dispose of bodies properly to reduce risk of infection.
/:cl:

### DIFF
--- a/code/modules/antagonists/disease/disease_abilities.dm
+++ b/code/modules/antagonists/disease/disease_abilities.dm
@@ -38,6 +38,7 @@ new /datum/disease_ability/symptom/powerful/fire,
 new /datum/disease_ability/symptom/powerful/flesh_eating,
 new /datum/disease_ability/symptom/powerful/genetic_mutation,
 new /datum/disease_ability/symptom/powerful/inorganic_adaptation,
+new /datum/disease_ability/symptom/powerful/undead_adaption //yogs change
 new /datum/disease_ability/symptom/powerful/heal/starlight,
 new /datum/disease_ability/symptom/powerful/heal/oxygen,
 new /datum/disease_ability/symptom/powerful/heal/chem,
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/disease/disease_abilities.dm
```diff
@@ -428,6 +429,13 @@ new /datum/disease_ability/symptom/powerful/heal/youth
 
 /datum/disease_ability/symptom/powerful/inorganic_adaptation
 	symptoms = list(/datum/symptom/inorganic_adaptation)
+	
+//yogs start
+
+/datum/disease_ability/symptom/powerful/undead_adaption
+	symptoms = list(/datum/symptom/undead_adaption)
```
Amelia0010: image

Amelia0010 (changes requested):

code/modules/antagonists/disease/disease_abilities.dm
ghost: that's weird because https://github.com/yogstation13/Yogstation-TG/blob/e94199f6b1bc08765e6ad9ad72b437e9c1e4b84a/code/datums/diseases/advance/symptoms/species.dm

Time-Green: Its adaptation, not adaption

ghost: ple ease giv me a g u n to shoot myself with

PR #3824, opened 2018-12-11
removes upgrade cost for reflective blobs

ShadowDeath6:
holy fuck reflector blobs are so fucking bad
I’m probably going to find a way to make them always reflect too but this is a decent short term fix for the significantly broader issue of blobs being really shitty

:cl: ShadowDeath6
tweak: Converting shield blobs to reflectors no longer costs Blob Points. 
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blob/powers.dm
+++ b/code/modules/antagonists/blob/blob/powers.dm
@@ -120,8 +120,8 @@
 /mob/camera/blob/proc/create_shield(turf/T)
 	var/obj/structure/blob/shield/S = locate(/obj/structure/blob/shield) in T
 	if(S)
-		if(!can_buy(15))
-			return
+		//if(!can_buy(15))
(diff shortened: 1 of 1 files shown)


Ratvar:
Maybe make it a cost of 5 biomass, but let blobs swap reflective blobs back to reinforced? This way it's not free and you can change mode back to regenerative.

ShadowDeath6:
I might do that if this doesn’t work out, but giving blob the ability to make shield blobs reflective for a few seconds and then turn them back might be too strong

Ratvar:
cooldown for swap

or no health increase from swap, means making reflective blob strong cuts it health down+damage stays

or both

ShadowDeath6:
if it doesn’t get the bonus health upon turning back into shield blob why would you ever want to spend bp to change it

Amelia0010 (approved):

PR #3856, opened 2018-12-15
Ling transform now uses a radial :)

Amelia0010:

This is what, my 7th attempt to do this, but I finally done did it. 
image

#### Changelog

:cl:  
tweak: Ling transform now uses a radial
/:cl:


ShadowDeath6:
not necessary imo, current disguise list shows identity (which is what people use rather than outfit) and is faster than radials in every way (no need to click through a bunch of lists or hover over the icons of people for their names) since transform has no cooldown. I would absolutely not want to use this over the current transform.

Moja10:
Make it a preference option in preferences :rofl:

Amelia0010:
no.

PR #3858, opened 2018-12-15
Holo Tools V2 Less Glue Edition!

fluffe9911:
Should not be stuck to your hand this time and now comes with a neat radial menu
https://github.com/HippieStation/HippieStation/pull/9027

Changelog
🆑 MoreRobustThanYou for making Fluffe9911 for porting
rscadd: Experimental Holotools. The RD spawns with one in their locker.
rscadd: You can emag holotools for access to to a holo knife!
/🆑


yogstation/code/game/objects/items/holotool/modes.dm
```diff
@@ -0,0 +1,77 @@
+/datum/holotool_mode
+	var/name = "???"
+	var/sound
+	var/behavior
+	var/speed = 0.125
```
Amelia0010: The fact that this is both faster than the researchable powertools and the toolset implant, and only requires a single slot makes this seem way too strong for something that spawns in at roundstart. Especially when you consider that all researchable items are weaker than this.

Either make it top-tier research only, or severely nerf the speed to put it just below the powertools the CE starts with.

Amelia0010 (changes requested):

yogstation/code/game/gamemodes/objective_items.dm
```diff
@@ -0,0 +1,5 @@
+/datum/objective_item/steal/holotool
+	name = "the holotool."
+	targetitem = /obj/item/holotool
+	difficulty = 5
+	excludefromjob = list("Research Director")
```
Amelia0010, line 5: the RD already has 2 steal-objectives, not including the research, why this as well?

Amelia0010 (changes requested):

yogstation/code/game/gamemodes/objective_items.dm
fluffe9911: Rd really only has one steal objective his armor I wouldnt really count the card a steal objective for the rd cause theres tons of other places you can get it without going threw the rd

yogstation/code/game/objects/items/holotool/modes.dm
```diff
@@ -0,0 +1,77 @@
+/datum/holotool_mode
+	var/name = "???"
+	var/sound
+	var/behavior
+	var/speed = 0.55
```
Amelia0010, line 5: make it 0.75, in line with https://github.com/yogstation13/Yogstation-TG/pull/3873

PR #3888, opened 2018-12-16
Makes the Cluwnification Burger Kit slightly less terrible Branch Edition

fluffe9911:
🆑 Fluffe9911
tweak: Cluwnification Burger now comes in a happy meal box complete with two mimanas a small toy and a emag!
/🆑


ThatLing:
Good idea, not sure about the free emag though

fluffe9911:
Eh its to help give a slightly better incentive to buying it over romerol depending on what gimmick you wanna do so like on one hand everyone dies on the other hand everyone gets disabled but I get a free emag as well

ThatLing:
How about giving them a stenchkin instead?

fluffe9911:
Eh alrighty fair enough

PR #3913, opened 2018-12-19
blob core attacks now flash in big bold red

Amelia0010:

image
https://forums.yogstation.net/index.php?threads/petition-to-notify-blobs-when-their-core-is-under-attack.17876/#post-155532

#### Changelog

:cl:  
tweak: the blob now gets a big bold red message telling them their core is under attack :)
/:cl:


fluffe9911:
If im reading the code right wont this only work for brute type attacks so the blob gets no notification if its getting shot with lasers and stuff?

Amelia0010:
you're not reading the code right

fluffe9911:
Ok

PR #3943, opened 2018-12-22
Tool Switcher

monster860:
https://streamable.com/50jt6

:cl: monster860
rscadd: Adds tool switcher. Load it up with tools, and program a sequence of tools, then use it in your hand to play through that sequence
/:cl:


Ratvar:
Does it work on surgeon tools?
Why not make it work on *anything*?

Moja10:
Looks broken beyond words. Should be traitor nukeop only imo

ghost:
Not faster than any of the upgraded tools, but it does eliminate the need for a toolbelt.

It's basically an arm-mounted surgical toolkit implant, but disarmable, and without the increased movespeed.

Give it a mineral req and it should be fine to shove over 'experimental tools'

monster860:
It can do surgery but the circular saw doesn't fit in side so it's not that useful for that.

jcatjmeow:
"program a sequence of tools"
You guys are forgetting this part
10/10 i love this more than my son

code/modules/research/designs/tool_designs.dm
```diff
@@ -33,6 +33,16 @@
 	category = list("Tool Designs")
 	departmental_flags = DEPARTMENTAL_FLAG_SCIENCE | DEPARTMENTAL_FLAG_ENGINEERING
 
+/datum/design/tool_switcher
```
ThatLing: yogs start

ThatLing (changes requested):

code/modules/research/designs/tool_designs.dm
monster860: there fixed

PR #3993, opened 2018-12-29
Experimentor non-critical reactions

monster860:
Simply put, this PR prevents upgrades from being a downgrade in certain circumstances.

:cl: monster860
tweak: Using non-critical items in the experimentor prevents certain malfunctions from being suppressed via upgrades
tweak: Using a failed reaction in the experimentor prevents the onboard detection system malfunction from being suppressed via upgrades
/:cl:

### DIFF
--- a/code/modules/research/experimentor.dm
+++ b/code/modules/research/experimentor.dm
@@ -241,6 +241,7 @@
 	smoke.start()
 
 
+/* yogs start - moved to yogstation/ folder
 /obj/machinery/rnd/experimentor/proc/experiment(exp,obj/item/exp_on)
 	recentlyExperimented = 1
 	icon_state = "h_lathe_wloop"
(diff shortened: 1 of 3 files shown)


HomocidlePancakes:
Dis gud

code/modules/research/experimentor.dm
```diff
@@ -241,6 +241,7 @@
 	smoke.start()
 
 
+/* yogs start - moved to yogstation/ folder
```
ThatLing, line 244: You don't have to comment this out

ThatLing (changes requested):

code/modules/research/experimentor.dm
monster860: yeah but isn't it more clear if it is? like someone oculd look at the code and be like "WTF the code says it does this but it doesn't" without realizing that it got mirrored somewhere else?

PR #4022, opened 2019-01-02
Makes Family Heirloom Free but you can get it again Modularized Edition

fluffe9911:
Bam


yogstation/code/datums/traits/negative.dm
```diff
@@ -0,0 +1,5 @@
+/datum/quirk/family_heirloom
+	name = "Family Heirloom"
+	desc = "You are the current owner of an heirloom, passed down for generations. You have to keep it safe!"
```
ThatLing: Why change name and desc?

fluffe9911: its the exact same I just didnt bother removing it one moment

PR #4047, opened 2019-01-04
Update mobs.dm

Amelia0010:

lol

#### Changelog

:cl:  
fix: you no longer start emitting miasma passively, you're gonna need to actually get dirty to do that
/:cl:


alexkar598 (changes requested):
Proper way is making another file and undef and defining it a second time for maximal modularisation.

ThatLing:
you dont even need to undef it

PR #4067, opened 2019-01-05
fix quiet round preference

GenericDM:
so basically fixes quiet round maybe not sure lets try any way.

#### Changelog

:cl:  
bugfix: fix quiet round
/:cl:


GenericDM:
thinking about it, this does nothing as is. nothing in antag objectives is looking for quiet round.

GenericDM:
I think this will work? Someone please tell me if it's wrong.

PR #4072, opened 2019-01-05
You can now fill fire extinguishers with any reagent!

fluffe9911:
🆑Fluffe9911 for porting Jujumatic for doing the actual work
add: You can now refill fire extinguishers with any reagent container!
/🆑


ghost:
this is a terrible idea

fund it

PR #4081, opened 2019-01-06
Changes last resort and armblade door opening message

Ludox235:
Last resort used to only be usable while alive, made so it's usable while unconscious or dead

Also when you used to force open a airlock with the armblade it used to say "We start forcing the armblade open" and "We force the armblade open" made so instead of armblade it says "airlock"

(inb4 maintainers become insane because I fucked up horribly somewhere)

:cl:  
tweak: Made "last resort" usable while dead and unconscious
spellcheck: Forcing open a airlock with the armblade no longer says "We start forcing the armblade open." and "We force open the armblade." and now says "We start forcing the airlock open." and "We force open the airlock."
/:cl:


thehighwayman:
were already insane

Moja10:
It's not modularized!!! Guess you are just fucking banned I suppose?

Ludox235:
> 
> 
> It's not modularized!!! Guess you are just fucking banned I suppose?

I wouldn't be surprised tbh I have no fucking idea what I am doing

ThatLing:
I don't like the last resort change

Ludox235:
> 
> 
> I don't like the last resort change

It's called last resort, pretty counter-productive to not being able to use it while unconscious or dead + In the end you're still a shitty headslug that dies in 3 hit from a mop

yogstation/code/modules/antagonists/changeling/powers/headcrab.dm
```diff
@@ -0,0 +1,44 @@
+/datum/action/changeling/headcrab
+	name = "Last Resort"
+	desc = "We sacrifice our current body in a moment of need, placing us in control of a vessel that can plant our likeness in a new host. Costs 20 chemicals."
+	helptext = "We will be placed in control of a small, fragile creature. We may attack a corpse like this to plant an egg which will slowly mature into a new form for us. Can be used while unconscious or dead."
+	button_icon_state = "last_resort"
+	chemical_cost = 20
+	dna_cost = 1
+	req_human = 1
+	req_stat = DEAD
```
Amelia0010: for the last resort change, you don't need to comment out the /tg/ part and you can just do the following in the yogs file:

```
/datum/action/changeling/headcrab
	req_stat = DEAD
```

easy modularization that changes as little as possible

PR #4122, opened 2019-01-08
random room icons now show up on the map

Amelia0010:

SEAMLESS 
image

Mostly did this for mapdiffbot, but I suppose it also helps mappers a bit?


Amelia0010:
and it uuh, excludes pipes for now, cause they don't work with fastdmm :)

Ktlwjec:
i love it

PR #4125, opened 2019-01-09
antag hud

GenericDM:
idk why but i felt like doing this

:cl:  
rscadd: add antag hud for ghosts
/:cl:


Moja10:
Brilliant idea to let everyone metagame with 100% certainty while they wait for their body to clone. I would like to think regular players don't get antaghud for a reason

GenericDM:
you have a very good point

GenericDM:
lmao this is what bad ideas do to you

PR #4134, opened 2019-01-11
add ckey to deadchat names

GenericDM:
this can probably be implemented better, should the ckey come before or after the character name?
:cl:  
rscadd: add ckey to deadchat name
/:cl:


Nickvr628:
I don't like it. If I want to reveal who I am I can do it in OOC after the round ends. Make it a toggleable setting so people can disable it if they want to.

GenericDM:
~~work in progress~~
finished

GenericDM:
oH FUCKINg GITHUB

GenericDM:
this is why you dont use multiple editors kids

GenericDM:
ok i fixed the git machine

GenericDM:
~~so as of right now it starts turned on, anyone know how to make it start off?~~
fixed

GenericDM (commented):
not sure if this is necessary, just did it because the looc PR did it

ShadowDeath6:
why?

ThatLing:
Why?

GenericDM:
why not
also other servers have them

code/modules/mob/say.dm
```diff
@@ -91,7 +91,7 @@
 		K = src.key
 
 	var/spanned = src.say_quote(message, get_spans())
-	var/rendered = "<span class='game deadsay'><span class='prefix'>DEAD:</span> <span class='name'>[name]</span>[alt_name] <span class='message'>[emoji_parse(spanned)]</span></span>"
+	var/rendered = "<span class='game deadsay'><span class='prefix'>DEAD:</span> <span class='name'>[(src.client.prefs.chat_toggles & GHOST_CKEY) ? "" : "([K]) "][name]</span>[alt_name] <span class='message'>[emoji_parse(spanned)]</span></span>"
```
Altoids1: Right now this would result in people's names being like ``(CuntDestroyer99) Bender Bending Rodriguez (died as Philip J. Fry)``, which is a fuckton to have to read every time the guy tries to complain in deadsay.

Also I would prefer if this was only something available to people with admin privileges; there's a lot of deadsay chatter that goes on that are basically people handling their own tickets. I'd rather people did not know the ckey of the guy they're grudging against, to help prevent metagrudging.

GenericDM: Do you mean that only staff members can show their ckey in deadchat, or that they are the only ones that can see the ckeys?

Amelia0010: @Altoids1

Amelia0010: @Altoids1 >:(

GenericDM: >I'd rather people did not know the ckey of the guy they're grudging against, to help prevent metagrudging.

Doesn't this being a toggle at least somewhat fix this?

Time-Green:
How am I supposed to threaten others if they know who I am?

PR #4137, opened 2019-01-11
Removes a ] and . in supplypod admin logs

Ktlwjec:
the "]" was annoying me and I don't think the "." is needed either


ThatLing:
This is gonna die as soon as another mirror comes in, but sure

Issue #4053, opened 2019-01-04
Pump in wrong direction


00aquir:
In sec maint atmos room

![ss13-atmosbug-secmaint](https://user-images.githubusercontent.com/16806055/50677400-e2ed8100-0ff9-11e9-9131-17e7c18464c8.png)

ghost:
Which one? The top one?

00aquir:
The blue pipes supply the station with air, coming from Atmospherics. This one leads to security, the upper pump always faced left.

Ktlwjec:
https://github.com/yogstation13/Yogstation-TG/pull/4150

PR #4144, opened 2019-01-11
Someone was bitching in ooc a few days ago about airless tiles being misplaced instead of reporting it here

Ktlwjec:
So I made sure all the tiles in space were airless and all the tiles not in space were not airless.
Fixes some pipe+injector icons since they were invisible for mappers.
Hides the pipes in the walls by atmos (yogbox).
Changes the window in atmos monitoring to be reinforced as all the 4 others in that room were (yogbox).
Also fixes issue #4053 
desc = "Much better at protecting the armory then your average warden." to desc = "Much better at protecting the armory **than** your average warden."


thehighwayman:
map diff no work :(

PR #4147, opened 2019-01-12
Readding Old Lavaland Ruins Part 2 This Should be the Last Time I make a new pr of a old one edition!

fluffe9911:
🆑 fluffe9911 for porting
add: The Russian Bunker, Mime Drill, Clown Facility and Pirate Base are back!
fix: Some inhand issues with the new balloons
🆑


code/modules/ninja/suit/suit.dm
```diff
@@ -97,10 +97,10 @@ Contents:
 /obj/item/clothing/suit/space/space_ninja/proc/lock_suit(mob/living/carbon/human/H)
 	if(!istype(H))
 		return FALSE
-	if(!is_ninja(H))
+	/*if(!is_ninja(H))
```
ThatLing, line 100: I don't like this change, would it possible to make like ``/obj/item/clothing/suit/space/space_ninja/lavaland`` or something and make that not gib?

ThatLing (changes requested):

code/modules/ninja/suit/suit.dm
fluffe9911: I mean I can try one moment

fluffe9911: Ok so I tried a couple of ways but cant seem to figure out a way to do it without just commenting out the section I dont really think this will impact much and if someone does get a ninja suit then good on them since I dont think the suit can be stripped off a ninja and they explode if they die am open to alt ways to solve this and or alt prize though

Amelia0010: @ThatLing

Amelia0010: @ThatLing >:(

PR #4155, opened 2019-01-13
Replaces Chasm on the Snap Pop Shuttle with a "Ball Pit"

fluffe9911:
Fun for the whole family!


Amelia0010:
no

Amelia0010:
It's a shitty joke which isn't even funny, which just results in 7 ahelps of "why did I die?"

fluffe9911:
nich fek off

PR #4161, opened 2019-01-13
Adds an admin log for anomaly-caused explosions

Ktlwjec:
Just a quick fix for https://github.com/yogstation13/Yogstation-TG/issues/3010 , alerts admins directly before the explosion happens and logs it.


ThatLing:
good stuff

PR #4172, opened 2019-01-14
fluffe is bad

Amelia0010:

wow decals are so hard to use

#### Changelog

:cl:  
fix: the diner works.
/:cl:


fluffe9911:
heck you caught me

PR #4177, opened 2019-01-15
Undoes misleading slime speed potion change

monster860:
keeps the vehicle nerf, but undoes the hardsuit thing because the only hardsuits with more than 2 slowdown can only be found on old space ruins. Keep in mind, having a slowdown of 0 for hardsuits simply equals normal movement speed, and not an increase in movement speed.

The original PR was misleading. The description mentioned having insane-speed vehicles, but the hardsuit nerf was undocumented.

:cl: monster860
tweak: Makes it so slime speed potions are actually useful for something again
/:cl:


Amelia0010:
how so?

PR #4229, opened 2019-01-19
Replaces plasma windows on the singlo engine with reinforced ones

fluffe9911:
Apparently they were blocking rads or something so this should fix it I think

### DIFF
--- a/_maps/yogstation/RandomRuins/StationRuins/BoxStation/engine_singulo.dmm
+++ b/_maps/yogstation/RandomRuins/StationRuins/BoxStation/engine_singulo.dmm
@@ -339,7 +339,7 @@
 /turf/open/floor/plasteel,
 /area/engine/engineering)
 "aJ" = (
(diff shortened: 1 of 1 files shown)


monster860:
fluffe

I just looked at the code

normal windows have some rad block
reinforced windows have a shit ton of rad block
plasma windows have...... no rad block.

fluffe9911:
oh

fluffe9911:
Now that the rad collectors are inside we can replace the windows with reinforced so rads stop leaking out everywhere

jcatjmeow:
Outside you mean

ByteSizedBee:
they are now right outside next to the field generators. Changing the glass should hopefully stop engineering being a radiated wasteland

monster860:
Can't you just, like, close the shutters?

Also this PR doesn't work any more, you only modified the plasma windows that don't have shutters.

ByteSizedBee:
The shutters don't have any rad insulation

PR #4264, opened 2019-01-22
Fixes some air supply pipes showing under disposals pipes

Ktlwjec:
image
Should stop that from happening.


monster860:
this really shouldn't be a mapping change.

Amelia0010:
oh well what's the worst that could happen

PR #4267, opened 2019-01-23
Nerf Timestop

Gh0sT2242:
Halves the duration of time-stop, giving people a semi-chance to not get destroyed by it
Fixes Issue#4266

### DIFF
--- a/code/modules/fields/timestop.dm
+++ b/code/modules/fields/timestop.dm
@@ -12,7 +12,7 @@
 	var/list/immune = list() // the one who creates the timestop is immune
 	var/turf/target
 	var/freezerange = 2
-	var/duration = 140
+	var/duration = 70
 	var/datum/proximity_monitor/advanced/timestop/chronofield
 	alpha = 125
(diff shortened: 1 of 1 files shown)


monster860:
Damn is it really 14 seconds? that's insane. Make sure this only affects Xeno timestops and not Wizard timestop.

Gh0sT2242:
Editing Xeno time-stop and Wizard timestop seperatly? Time stop is a core module isn't it? I think that would require me to separate them via redirecting and making separate files

Amelia0010:
https://github.com/yogstation13/Yogstation-TG/blob/335223f304f1174752d8edf3395bd26bdf6aaa26/code/modules/research/xenobiology/crossbreeding/chilling.dm#L190 This would be the appropriate line to edit changing 300 to whichever value you want, remember to put //yogs at the end of the line after editing it!

Gh0sT2242:
@nichlas0010 I'll fix that then, I thought it was directed through the core module, oof

Issue #4284, opened 2019-01-25
Several of the medical trash cans are spitting their shit back out


MegaEmpirical:
I am guessing there is a disconnect somewhere.

Ktlwjec:
Don't recall anyone touching disposals but I'll give it a look, yogbox I assume?

PR #4287, opened 2019-01-25
Fixes medbay disposals on yogbox

Ktlwjec:
Fixes issue #4284


monster860 (changes requested):
I don't think that's gonna work right for mailing things to the CMO, you gotta flip the sort junction in the medbay and then replace the non-sort junction in maint with a sort junction.

image

Ktlwjec:
I was fixing the issue linked not the cargo disposals issue, but I added it anyway for you.

monster860 (approved):

PR #4289, opened 2019-01-26
Makes it so that maintainers don't block the server from restarting

monster860:
I'm not a real admin ree


Amelia0010:
image
are you actually retarded?

k-k-karma:
shut up idiot

Amelia0010:
Calling headcoders idiots? 

That's a gitban

PR #4303, opened 2019-01-26
2 New traits: Robust and Unrobust

Raze500:
`My third PR.`

Okay **my english isn't perfect**, but if a **mainteiner** wants to change the traits names, go ahead.

--------------------

**Simply adds 2 new traits.**

**The Robust trait** just gives you 15+ MaxHP (that is enought to resist 11-13 toolboxes love medicine) at the cost of 4 points.

**The Unrobust trait** do the same but instahead removes your MaxHP to -20 ( (that is enought to resist 9-11 toolboxes love medicine) you can also use this trait to be a badass tho. it cost only -2 points

You can use both traits but you will lose MaxHP btw.

**Any recommendations would be great!**

(Everything is in the yogs folder)

I've like to the users have a significant adventage and disvantage using those traits, in a future (if possible) i will add more.

#### Changelog

:cl:  
First commit.
/:cl:


Notamaniac:
I seriously don't think anyone would take -4 worth of traits just to have 15 more HP. Also it's a pretty boring trait and the grammar needs improving on the trait descriptions

jcatjmeow:
Bad changelog

fluffe9911:
Even if its -4 (although code says -3) point trait I dont know how I feel about a trait that gives a player a obvious ingame advantage against other players kinda ehhhh

Altoids1:
I like the idea, but I have to agree with Fluffe here. Hard-to-detect mechanics that give players an edge in combat aren't going to be fun for the guy who actually has to kill this Robust guy.

That being said, I guess if the point cost is high enough then it wouldn't be necessarily _broken_ to have in the game.

Alexandria404:
Balance says no.
We talked about it, he has some new ideas coming up.
Agreed himself it´s too much.

Nirnael:
These traits are making this game too much of an RPG imo

Time-Green:
People with robust will probably be brain-tumor, paraplegic blinds so I guess thats nice

Amelia0010:
image
and I'm not invested enough to be an autocrat

PR #4311, opened 2019-01-27
Prevents shadowlings from hatching too early

monster860:
:cl: monster860
tweak: Shadowlings must now wait 30 mintues before hatching
/:cl:

Let's not have anymore "fuck you why did you hatch early" memes.


Raze500:
30 minutes seems reasonable and gives a good time to the 3 nerds convert the people, i like this change.

MegaEmpirical:
What if the shuttle is called or some kind of situation that leaves them with limited time?

Time-Green:
Why the fuck would you ever not hatch early? You’re a waste of antag if you don’t.
Whatever you can do unhatched thralls can do better

PR #4328, opened 2019-01-29
consider maybe saying that instead of just "penetrator fix mirror". 

Amelia0010:
* Bandaid fixes for bugs with reflectors projectiles and plasma cutters

* * Fixed bouncing piercing projectiles on qdel
* Fixed projectiles reflection not working
* Fixed heavy beam pulses not penetrating
* Changed numerous projectiles procs to use the new BULLET_ACT_* format

* Final conversions to new BULLET_ACT_* format

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -130,7 +130,7 @@
 		return BODY_ZONE_CHEST
 
 /obj/item/projectile/proc/prehit(atom/target)
-	return TRUE
+	return BULLET_ACT_HIT
 
 /obj/item/projectile/proc/on_hit(atom/target, blocked = FALSE)
 	var/turf/target_loca = get_turf(target)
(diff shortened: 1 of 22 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -606,6 +606,15 @@
...
 			Bump(AM)
 
+/obj/item/projectile/Move(atom/newloc, dir = NONE)
+	. = ..()
+	if(.)
+		if(temporary_unstoppable_movement)
+			temporary_unstoppable_movement = FALSE
+			DISABLE_BITFIELD(movement_type, UNSTOPPABLE)
+		if(fired && can_hit_target(original, permutated, TRUE))
+			Bump(original))
```
monster860, line 616: nich

you're bad at parentheses. 

```suggestion
			Bump(original)
```

monster860 (changes requested):

code/modules/projectiles/projectile.dm
Amelia0010: This isn't my PR, it's a mirror :)

monster860: yes but the original PR did not have this error.

PR #4357, opened 2019-01-31
Makes it so payments can be processed using NTSL

monster860:
When your ID card speaks due to bank account things, it now speaks via `say()` so that an intercom or a radio can pick it up and allows the payment to be processed by an NTSL script.

:cl: monster860
tweak: Paystand payments can now be processed using NTSL by placing the ID card used for the paystand next to an intercom, and then parsing the messages it generates when a payment is received.
fix: Oh yeah and the messaging actually works now.
/:cl:


Nickvr628:
why have the requirement for people to place it next to the intercom? If I write a script I should be able to make people pay per message they send!

PR #4387, opened 2019-02-03
Fixes having 1.73543e+006 amount of metal in the protolathe

monster860:
before: there is 1.73543e+006 metal in the protolathe
after: there is 1735431 metal in the protolathe

:cl:
fix: you no longer get scientific notation when view amount of metal in protolathe
/:cl:


Altoids1:
Alternative options would include making it actually output the amount of metal in some metric units, like "25 kg" or "142 zettagrams" or something.

PR #4388, opened 2019-02-03
Prevents beer nuke from preventing round end - "TG can suck my dick" edition

monster860:
So we had a round today where the round didn't end after the shuttle docked because the beer nuke exploded.

Not doing this on TG because TG can suck my dick.


HomocidlePancakes:
Imagine labeling your own PR with so many labels so it looks more funny.

monster860:
All of the labels are accurate. *all* of them.

PR #4392, opened 2019-02-03
Changes command radio color

monster860:
changes it from this:

image

to this:

image

:cl: monster860
tweak: Command radio is now a very visible blue instead of an invisible ugly yellow.
/:cl:

because the captain/hop wears blue so it should match.

### DIFF
--- a/code/modules/goonchat/browserassets/css/browserOutput.css
+++ b/code/modules/goonchat/browserassets/css/browserOutput.css
@@ -278,7 +278,7 @@ em						{font-style: normal;	font-weight: bold;}
 .binarysay a:active, .binarysay a:visited {color: #88ff88;}
 .radio					{color: #008000;}
 .sciradio				{color: #993399;}
-.comradio				{color: #204090;}
+.comradio				{color: #204090;} // yogs
 .secradio				{color: #a30000;}
 .medradio				{color: #337296;}
 .engradio				{color: #fb5613;}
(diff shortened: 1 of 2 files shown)


fluffe9911:
We already did this once and it got reverted cause it looks way to much like ooc color why cant we just bloody leave it gold reeee

Nickvr628:
The blue is a better color. OOC has a different font and everything, plus has a giant "OOC" tag in front instead of "[Command]". +100 for changing it back to blue.

monster860:
the PR you are talking about used this color:

image

what I'm proposing is this color:

image

clearly my color choice is less likely to be confused with OOC

@fluffe9911

fluffe9911:
eh guess we will see lol

Altoids1:
> eh guess we will see lol

Or we could see right now.
image

It is definitely _more_ different from OOC than the last PR that tried to do this.

fluffe9911:
eh true

HomocidlePancakes:
Please die

Bibby0110:
Why? Why change the color to something that has the potential to even confuse like that. Why not just keep it the way it is, theres literally zero problems with it currently 💭

Amelia0010:
I'm just gonna start a server vote on this because I don't care and a lot of others probably do.

ThatLing:
retarded pr

MegaEmpirical:
Bring back the new service channel color too!!!

Altoids1:
> Bring back the new service channel color too!!!

Yeah, regardless of the result of the server vote, I think everyone would appreciate a new Service colour.

MegaEmpirical:
We'd had one and the new command and then someone took them from. Some unknown monster.

monster860:
image

PR #4435, opened 2019-02-06
Medbay Remap

Rayisken:
Haven't Run mapmerger yet but it should pass
https://media.discordapp.net/attachments/378318242906636288/542776354081472513/medbayremap.png?width=1150&height=683


fluffe9911:
Uhm did you map merge it?

Rayisken:
Im trying to figure that part out I think I did?

ThatLing:
You didn't

Altoids1:
image

Here's an image of it, while Ray is busy goofing around with Github & mapmerger.

HomocidlePancakes:
I've noticed some weird things 

-Chemistry has r-walls  for no reason
-New medbay storage doesn't need r-walls because it's now connected to the main hallway and not maint and the medkits are not next to an outer wall.
-Surgery observation should have an unrestricted entrance from the medbay.
-Mulebot unloading zone is blocked off and the beacon for it is still there.
-Whats up with all the racks?
-Why are there windows between surgery and medbay storage?
-Genetics has no lights
-Medbay storage lights are not attached to walls.

Also my dumb eyes can't find the medbay APC but it's propably somewhere

Rayisken:
Im not saying MERGE THIS NOW, getting feedback here is more productive than discord

-R-walls around chem because botanists keep stealing the shit out of the chem dispenser roundstart, moving the dispenser a bit would fix that as well I guess

- People have argued both sides, I'm ambivalent to either, storage having a higher visibility prevents break-ins just as much

- I have a solution kinda figured out already, most people who want to watch a surgery just burst in anyway

-Surgery has shutters for the windows

-racks are for peoples stuff as we now need to strip for things like showering etc, tables means people are probably going to get tabled.

-WiP

- You should probably get your eyes checked because it does, and was untouched

What I was thinking was shifting the sleeper area+cryo more to the west and making part of the open space from cryo an observation room and having those showers+recovery beds on the eastside before cryo.

HomocidlePancakes:
Medbay shouldn't have any r-walls at all. It's not a high security area even somewhat public. Even the CMOs office doesn't have any because there is nothing important in it. The only reason storage had them is that it was hard to keep track of it because of how deep it's in the department. Also maint is super low traffic so noone would stop you from breaking in. Since you moved it to a more viewable location these problems were fixed hence keeping the r-walls is redundant.

Now about the chemistry. Chemist has a syringe gun and a right to use it against anyone thretaning his dispenser. When someone is breaking in you can just shoot them with fent for some fun time brain traumas. Not enough? You have a buddy chemist and the entire medbay personnel to back you up. Can't be arsed to protect your shit yourself? Call for sec. All though if you are just going to be a massive cuck and not do anything as the greentide steals your shit then you've never deserved it in the first place. Having r-walls there when the chemists already have all the means to prevent break ins is just straight up hugbox.

If you want to have racks so that the people who shower put their stuff there then keep only the ones next to showers. Having them next to every chair is dumb. Put plants or some other decoration instead of tables if you think tabling is too much of a problem.

HomocidlePancakes:
Also one more thing. You should replace the windows in observation with reinforced ones. It's too easy to steal the surgery tools because the observation is now connected to the main hallway and not medbay general.

MegaEmpirical:
Yeah, surgery needs r windows at least. I'm not opposed to r walls in chem due to explosions. I'm digging a lot of these changes though, good work.

Altoids1:
Is ded -- come back later once someone explains to you how to mapmerge.

Rayisken:
Bunch RL stuff got in the way (new job) feedback was good though

PR #4440, opened 2019-02-09
Split a bag of holding in two!

monster860:
image

image

This can't be used for extra storage unless you want to sacrifice your sight - you will either look out another bag, or you will become blind until you take it off.

image

Potential uses:

- give a grafted split bag to a victim as a "gift" and then use it to spy on them whenever you want
- give the second half to your friend, and have them remove the contraband if security decides to search you

If you split a bag and graft another bag onto it the only way you can identify that it's split is to wear it on your head.

:cl: monster860
add: You can now split a bag of holding in half. Simply use a scalpel while it's on a floor, and then use the scalpel again on it (the second time, cancelling the progress bar will result in a small tear) and it becomes split in half. 
add: You can make a split bag of holding look like a whole bag by taking another bag of holding, then using scalpel followed by hemostat to "disable" it (again don't cancel during hemostat part), then "graft" it onto a split bag to make it look whole. You can even split it again if you really want. 
add: You can wear a bag of holding on your head to look out the other side.
/:cl:


HomocidlePancakes:
Bags on heads. Features that we deserve

ShadowDeath6:
I feel like this should also work with paradox bags for the same reason

Amelia0010 (approved):

PR #4444, opened 2019-02-10
[admin] Allows admins to spawn cookies for all mobs instead of only /mob/living/carbon's

Altoids1:
### Cookies for all!

This removes some restrictions in the (SC) button, allowing for mobs not of type ``/mob/living/carbon/`` to be targeted for cookie dispensing.

Luckily, ``/mob/proc/put_in_hand`` seems to be pretty well-defined against this (for instance, Cyborgs will have their cookie just casually spawn at their feet instead of something otherwise) so I doubt this is going to cause any cookie-related bugs.

If you wanna double-check that, ``Yogstation-TG\code\modules\mob\inventory.dm:223``.

#### Changelog

:cl: Altoids
tweak: Admins can now spawn cookies for all mobs.
tweak: Admins can now spawn as many cookies on people as they want to, regardless of the willingness of their victims. HALLELUJAH!
/:cl:


1fbff5f83b23d39d38b1dfcb4cac8d9b:
#4444, hooray

Issue #4441, opened 2019-02-09
Xenobiology kill room camera is in escape


00aquir:
_

HomocidlePancakes:
I was actually about to report that. oof

PR #4445, opened 2019-02-10
Fixes camera tags in yogbox escape

Ktlwjec:
:smiley_cat:
Fixes issue #4441


monster860:
I hate c_tags so much..

PR #4460, opened 2019-02-12
Adds a "Probabilistic Anomaly" event

Altoids1:
## Ling's biggest gripe about NTSL, made into a random event

The gimmick of this event is that it forces the seed of the randon number generator to be stuck on a certain value, which causes the same numbers to be generated over and over. 

This could result in anything from, say:
- The emitters not shooting enough lasers to keep the SM going or the singulo contained (although the latter would only realistically occur if there was barely enough containment already, which is not by default really the case)
- Arcade machines always outputting the same reward
- Disarms always not working, or always working
Etc, etc, etc.

#### Changelog

:cl: Altoids
rscadd: New random event, "Probabilistic Anomaly," in which RNGesus gives up and moves to Mount Fuji.
/:cl:


Nickvr628:
It should last a lot longer than just a minute or so IMO, 90 seconds could pass without anyone realizing what it changed.

monster860 (approved):
There is absolutely no way at all in any possible way this could go wrong in any way whatsoever.

alexkar598:
Make it last like 10 minutes

ThatLing:
BIGGEST GRIPE WHY DO YOU ALWAYS HUMILIATE ME IN YOUR PRS FUCK OFFFFFFFF

ThatLing:
ALSO THIS IS A GREAT EVENT THANKS FOR THE CONTRIBUTION ALTOIDS

ThatLing (approved):

Firewolf34:
This is a really cool concept. _META_

PR #4491, opened 2019-02-15
Adds three new bad wizard setups! 

fluffe9911:
🆑
add: Three new bad wizard event setups one has a max leveled cluwne spell and lvl 2 blink spell
add: The other maxed level turn to stone and animation bolt
add: The last one is mice transformation and mindswap have fun!
/🆑


jcatjmeow:
Turn to stone plus animation is good tho

fluffe9911:
I mean bad wizard as far as im aware is kinda for both gimmicky and bad wizard setups and 
animation +turn to stone seems pretty gimmicky to me so I think its fine

monster860 (approved):

PR #4493, opened 2019-02-15
Lowers cost of mindslave Implant to 7 tc

fluffe9911:
🆑
tweak: Mindslave Implant is now 7tc down from 12
/🆑

I mean 12 tc seems pretty overpriced since it takes a couple seconds to use and the fact its 4 tc more then much better items like the emag for example then theres the brainwash surgery which in some cases objectively better since you can use it to brainwash as many people as you want for only 5 tc with the only downside being its restricted to med staff  you need more time and you need to do it with one area so I think 7 tc would be a much more fair price for this that is all

### DIFF
--- a/yogstation/code/modules/uplink/uplink_item.dm
+++ b/yogstation/code/modules/uplink/uplink_item.dm
@@ -84,5 +84,5 @@
 	name = "Mindslave Implant"
 	desc = "An implant injected into another body, forcing the victim to obey any command by the user."
 	item = /obj/item/storage/box/syndie_kit/imp_mindslave
-	cost = 6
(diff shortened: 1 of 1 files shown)


Altoids1:
Doing this should help promote RP and lessen the amount of people a traitor has to remove from the game to make stuff happen. Solid +1 from me.

Bibby0110:
CAN YOU DECIDE ALREADY

monster860:
I will put this to a server vote.

fluffe9911:
fair

monster860:
image

monster860 (approved):

PR #4494, opened 2019-02-16
[admin] Create Object is no longer a massive lagmonster when your search terms are vague

Altoids1:
### Fucking Internet Explorer is a Piece of Shit, Part ``2 << 100``
image

So apparently, whenever you searched for anything using this thing, every single search result would cause the entire browser window to refresh, meaning if you tried to load up all the thousands of objects in the game, it would have the performance of you suddenly opening up one thousand tabs in your browser.

Now, it sensibly only refreshes the window once to display the search results.

In my own tests, the time it took to parse with no search terms dropped from 40-70 seconds to not much more than a second.

Thanks to @monster860 for the suggestions and assistance with the bugfixing. ❤️ 

#### Changelog

:cl:  Altoids and Monster860
bugfix: The Create Object menu should no longer hang for 2 millennia whenever you search without a search term.
/:cl:


Rayisken:
Thank you, I think I mentioned this a couple times over the last year.

PR #4521, opened 2019-02-20
Modify TGM so that Dream Maker doesn't completely shit itself when encountering a map with more than 65536 tiles in it

monster860:
because BYOND is a piece of shit that can't handle map files with more than 65536 rows in it which is easily surpassed when using TGM.

Why? Because Fluffe is trying to port minsky station which is a 350x350 map and it's making DM shit itself.

### DIFF
--- a/tools/mapmerge2/dmm.py
+++ b/tools/mapmerge2/dmm.py
@@ -283,12 +283,24 @@ def save_tgm(dmm, output):
 
     # thanks to YotaXP for finding out about this one
     max_x, max_y, max_z = dmm.size
+# yogs start - multi-column mode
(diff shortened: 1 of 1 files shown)


Nickvr628:
What is the new limit?

1fbff5f83b23d39d38b1dfcb4cac8d9b:
probably unlimited

Altoids1 (approved):
I don't really understand why this bug exists, but, alright.

monster860:
it exists because BYOND is a dumbdumb

Altoids1:
A reasonable assertion

PR #4578, opened 2019-02-26
Fixes Atmos Techs no longer having access to the engine

### LINKED ISSUE #4577: Atmos techs no longer have engine room access
Yogbox station. Used to be that atmos techs had engine access.

Altoids1:
### Fucking #4361 has caused so many bugs for me that I know its PR number by heart

Fixes #4577

#### Changelog
:cl:  Altoids
bugfix: Atmos Techs should again have access to the Engine Room!
/:cl:


Nickvr628:
+100. The SM is an atmos engine and often times the atmos tech is the only crewmember robust enough to diagnose and fix problems with it, plus it lets him access the connections from the atmos room to the engine's pipework. I am glad this is fixed.

Amelia0010:
💭 they really shouldn't have access to it, then the only difference between them and engineers is that atmos techs get more stuff 🤔

PR #4580, opened 2019-02-26
[admin] Makes unclaimed tickets automatically re-bwoink again

### LINKED ISSUE #2844: Please make tickets show in chat when unclaimed
Before rebase they appeared in chat as something like "Ticket 5 is unclaimed!" in red text with a bwoink noise. I'm not sure on how often they did this when unclaimed (I think every 30 seconds or a minute from what I can remember). There's been a few incidents recently with admins reconnecting or joining late without realising there is an open unclaimed ticket.

Altoids1:
### Yeah cool

For those not in the know, prebase (and during the early rebase) any tickets not claimed by an admin would automatically bwoink at all admins every 30 - 60 seconds until one of them answered the ticket.

This was deactivated by Nich because of ~~Nich reasons~~ because it used to bwoink every tick for some inexplicable reason, which I'll need to fix for this PR.

Fixes #2844

#### To-Do
- [x] Make this use SSyogs instead of CALLBACK or whatever it's currently trying to do
- [x] Make this ignore "admins" who lack R_BAN

#### Changelog
:cl:  Altoids
admin: Unclaimed tickets now bwoink all admins again, every 45 seconds.
/:cl:


Amelia0010:
>This was deactivated by Nich because of Nich reasons so I'm reactivating it.

This was deactivated by Nich because every once in a while the ticket would just rebwoink instantly, leading to a billion bwoinks a minute, which is not pleasant for anyone.

monster860 (changes requested):
Any chance you could make it *not* send bwoinks to those who don't have R_BAN? Y'know, because.... yeah

monster860 (approved):

PR #4599, opened 2019-02-27
Ports patch wounds and treat burns surgeries from prebase

SomeguyManperson:
:cl:  Theos & Super3222
rscadd: Medical men can now use bandages to heal people via surgery with the new and improved patch wounds and treat burns surgeries
/:cl:


monster860 (changes requested):
yeah this is kinda pointless we already have the researchable surgery to fix people....

SomeguyManperson:
It serves as a midpoint between synthflesh and reconstruction in terms of burst healing and allows medical doctors to heal bodies if science skimps on surgery research. Useful when a body that's damaged but under the defib time limit is brought in whereas reconstruction seems to be built to synergize with revival surgery.

Smudged:
Considering those surgeries can take up to half an hour to an hour to research (Because it's not on the top list of things scientists research) It's a welcome feature to have back.

PR #4615, opened 2019-02-28
[admin] Cryo now logs when beakers are removed, as well as added

### LINKED ISSUE #4614: Better Cryo Logging
Would it be possible to get a bit more logging for Cryo? Specifically we can only see when a beaker is added but not removed currently.

Altoids1:
image

Fixes #4614

#### Changelog
:cl:  
bugfix: Cryotubes now properly log what you do to the beakers inside of them.
/:cl:


code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
```diff
@@ -404,6 +404,7 @@
 				beaker.forceMove(drop_location())
 				if(Adjacent(usr) && !issilicon(usr))
 					usr.put_in_hands(beaker)
+				log_game("[key_name(usr)] added an [beaker] to cryo containing [beaker.reagents.reagent_list]") // yogs -- Adds logging for when the beaker's removed from cryo
```
alexkar598: Read this again plz

alexkar598 (changes requested):

monster860 (approved):

PR #4676, opened 2019-03-05
Adds cosmos bedsheet as donor item

Altoids1:
image

They're fancy and animated and they glow like a zippo and have special dream messages about Hans Zimmer. And now, it can be *yours*, if you donate!

#### Changelog
:cl:  Altoids
rscadd: Fancy animated cosmos blankets are now available as a donor item!
/:cl:


Nickvr628:
best cape ever?

FatalEYES:
Nice, glad I was able to bring this to attention, cause now I get to be a snowflake and a star all for the low cost of $14 dollars every three months.

PR #4685, opened 2019-03-06
Fixes creep preferences

Altoids1:
#### Changelog
:cl:  Altoids
bugfix: Creep preferences are fixed!
/:cl:


Altoids1:
Yes I know /tg/ doesn't even have Creep anymore; just merge this in so that we can shuttle all this code over when we mirror-and-disable Obsessed.

PR #4698, opened 2019-03-06
completely removes snails from being accessible at all.

Amelia0010:
they made donor shells nodrop 😂


MegaEmpirical:
I'd fucking asked them if this meant they couldn't be looted on death or some shit. Bastards.

PR #4702, opened 2019-03-06
Turn your key Sir!

Nickvr628:
Don't you hate having to run all the way back to your office to swipe for red alert? Maybe the bridge is the only safe place left? No time to run to another authenticator? Fear not, because now there are TWO keycard authenticators on the bridge, spaced apart so you can see the other person but cannot run from one to the other. If you are crafty, you might be able to find a way to solo-activate red alert, but your best bet is to just have two people there.

image

Also you get to shout at people to TURN YOUR KEY and hold them at gunpoint until they unlock the BSA when Centcom sends Emergency Wartime Orders.

:cl:  
add: There are now two authenticators on the bridge so two Heads (or a crafty single head) can activate red alert from the same location. 
/:cl:

### DIFF
--- a/_maps/yogstation/map_files/YogStation/YogStation.dmm
+++ b/_maps/yogstation/map_files/YogStation/YogStation.dmm
@@ -22393,10 +22393,6 @@
 	pixel_y = -2;
 	req_access_txt = "19"
 	},
-/obj/machinery/keycard_auth{
-	pixel_x = 29;
-	pixel_y = 8
-	},
 /turf/open/floor/plasteel,
 /area/bridge)
 "aSz" = (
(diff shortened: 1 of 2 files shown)


nichlas0010:
bad

Nickvr628:
why nich, why is it bad that now two heads can initiate a red alert without having to run laps around the station?

Kmc2000:
use a portal gun
you now have solo redalert

Nickvr628:
Yeah but that requires some effort and ingenuity to do, so it is fair game. No different than firing one into the HOP office and bridge.

MegaEmpirical:
You'd think you would be able to red alert on the bridge anyway. It's the damn bridge.

PR #4729, opened 2019-03-08
Revert "Revert "makes honk staff bullets ricochet""

Amelia0010:
Reverts yogstation13/Yogstation-TG#4719


Amelia0010:
Tested a bit, don't see why it would break (also, the previous breaking wasn't my fault)

ThatLing:
this is a horribble idea

PR #4732, opened 2019-03-09
Makes roundstart command jobs logarithmically prefer players with more hours in the job

### LINKED ISSUE #4620: Bring back [Ultra] Priority whitelisting for roles.
Altoids said this wasn't in the rebase and I thought it'd be nice to bring it back so people who are able to do their job properly have the best chance of getting a role through whitelisting.

Altoids1:
## The New [ULTRA] Preference, now with more systemic oppression
image

This makes command positions weightedly prefer players who have more hours in the the job. This weighting is logarithmic, **which means if you have 100x more hours in CE than the other guy, you are twice as likely to be picked for it.** This should help improve the quality of command staff without absolutely preventing the new guys from playing.

Players with less than 10 hours of playtime are "bumped up" to 10 for the purposes of the pick, to prevent out-of-bounds issues with log() and to keep 1000-hour players from being a thousand times more likely to get in than a guy who was Captain for 0.2 seconds.

These "raffle tickets" are calculated as floating-point, meaning you can have 1.5253x chance over the other guy instead, or whatever.

Closes #4620.

#### Changelog
:cl:  Altoids
rscadd: At roundstart, command jobs now slightly prefer people with more hours in said job.
/:cl:

### DIFF
--- a/code/__HELPERS/_lists.dm
+++ b/code/__HELPERS/_lists.dm
@@ -231,7 +231,7 @@
 			L[item] = 1
 		total += L[item]
 
-	total = rand(1, total)
+	total *= rand() // Yogs -- Allows for noninteger weights
 	for (item in L)
 		total -=L [item]
 		if (total <= 0)
--- a/code/controllers/subsystem/job.dm
+++ b/code/controllers/subsystem/job.dm
@@ -207,7 +207,7 @@ SUBSYSTEM_DEF(job)
 		var/list/candidates = FindOccupationCandidates(job, level)
 		if(!candidates.len)
 			continue
-		var/mob/dead/new_player/candidate = pick(candidates)
+		var/mob/dead/new_player/candidate = PickCommander(candidates,command_position) // Yogs -- makes command jobs weighted towards players of greater experience
 		AssignRole(candidate, command_position)
 
 /datum/controller/subsystem/job/proc/FillAIPosition()
--- a/yogstation.dme
+++ b/yogstation.dme
@@ -2802,6 +2802,7 @@
 #include "yogstation\code\controllers\configuration\entries\yogstation_config.dm"
 #include "yogstation\code\controllers\subsystem\bluespace_locker.dm"
 #include "yogstation\code\controllers\subsystem\input.dm"
+#include "yogstation\code\controllers\subsystem\jobs.dm"
 #include "yogstation\code\controllers\subsystem\mapping.dm"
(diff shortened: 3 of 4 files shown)


Moja10:
Add this for clown as well

Altoids1:
A common argument against these things is that is stagnates things, making people who "main" a role constantly get it more and more, meaning newer players never get the chance.

With logarithms, such a positive feedback loop isn't really possible. Even if the HoS starts with 1000 hours in the role, that only gets him 3x the chance of a random Brazilian getting it. In order to maintain that superiority, though, he would have to get the role 1000x more often, and conveniently 3 < 1000.


**As well, I want to note that this whole thing only settles ties between people who have the job at the same priority.** If you go High and Marc Mingle goes Medium, then he is not getting HoS no matter what.

ThatLing:
So what does this add? Why have this in the first place?

Altoids1:
> So what does this add? Why have this in the first place?

People have been asking for the [ULTRA] preference to come back, but the process of whitelisting is tedious and it enforces "skill" requirements far too strongly and asymmetrically.

This is an adjustment on the idea of giving people priority for skill, by making it a weak, probabilistic tie-breaker when two people of the same priority want to be the same job.

Again, the relative increase in chance for the high-houred folk is just going to range from 1x - 3x, and because of how logarithms work, most of that (slight) edge can be gained very easily by the new guys reaching the 100 hour mark.

All in all I think this is a good compromise that implements skill prioritization without making it too powerful, inaccurate, nor tedious to maintain.

Ratvar:
Maybe jobbans should remove all/half of accumulated trust? It wouldn't make much sense for offender to still have "ultra" preference

Altoids1:
> Maybe jobbans should remove all/half of accumulated trust? It wouldn't make much sense for offender to still have "ultra" preference

I thought about this, but I prefer people being fully liberated after their bans are removed. It feels wrong to continue to punish them way after their punishment has been satisfied; adding such a thing would make me feel like I'm implementing the American prison system.

Also querying the database about all that would be laggy.

Altoids1:
image

**Only 10 people will be getting a factor increase greater than 2x.**

Also, of the 2,380 people we have logged as having ever played a command role, only **380** of them will be receiving any factor increase as a result of this PR. For the majority of people, this will affect nothing.

ThatLing:
I'm still opposed to this because I don't see why we need this, but I don't play so who cares.

PR #1930, opened 2018-07-28
its over anakin

GuppyLaxx:
i have the highground

:cl: GuppyLaxx
add: Adds the use of high ground
/:cl:

PR #4765, opened 2019-03-13
Allows for dual-sabers to also make use of the high ground

Altoids1:
## Anakin!

An extension of #1930, which works for all deswords.

#### Changelog

:cl:  Altoids
bugfix: Dual-sabers can now make use of the high ground.
/:cl:

### DIFF
--- a/code/game/objects/items/twohanded.dm
+++ b/code/game/objects/items/twohanded.dm
@@ -368,7 +368,7 @@
 	else
 		user.adjustStaminaLoss(25)
 
-/obj/item/twohanded/dualsaber/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the attack", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
+/obj/item/twohanded/dualsaber/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the attack", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK) //yogs -- mirrored to allow for use of the high ground
 	if(wielded)
 		return ..()
 	return 0
--- a/yogstation.dme
+++ b/yogstation.dme
@@ -2886,6 +2886,7 @@
 #include "yogstation\code\game\objects\items\sharpener.dm"
 #include "yogstation\code\game\objects\items\tool_switcher.dm"
 #include "yogstation\code\game\objects\items\toys.dm"
+#include "yogstation\code\game\objects\items\twohanded.dm"
 #include "yogstation\code\game\objects\items\weaponry.dm"
 #include "yogstation\code\game\objects\items\circuitboards\computer_circuitboards.dm"
 #include "yogstation\code\game\objects\items\circuitboards\machine_circuitboards.dm"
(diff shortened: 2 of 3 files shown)


Kmc2000:
🤔
But even when maul had the high ground obi wan beat him

Amelia0010 (changes requested):
Obi-Wan doesn't need to be on the high ground, the high ground just needs to exist within the battle; Obi-Wan knows that when he has the low ground, he really has the high ground, from a certain point of view. Look at his battle record: Maul: Has low ground, wins

Dooku: No high ground, loses

Dooku rematch: No high ground, loses

Greivous: Has low ground, wins

Vader: Has high ground, wins

Vader rematch: No high ground, loses

Obi-Wan with the high/low ground is canonically the most powerful Jedi. This is fact. Had Yoda not denied his request to battle The Senate with typical Jedi arrogance, Obi-Wan could have defeated Palpatine in the Senate building, which housed a variety of different altitudes; this was designed so that the Chancellor could always have the moral high ground in political debates. But Obi-wan didn't fight The Senate, and Yoda soon learned that you can't cleave the Sheev in a normal 1v1.

Yoda is shorter than virtually every other fighter, which gives him a permanent low-ground disadvantage; however, his saber-fighting style utilizes a flipping-heavy technique in order to negate this weakness for a temporary window. You'll notice that, as he falls from the central podium in The Senate's building, he immediately retreats, knowing that he can never hold the high ground in this duel. You'll also notice that, while training Luke, he rides on him like a mount, to gain the intellectual high ground and accelerate Luke's training. As we all know, spinning is a good trick. However, only the Chosen one can spin outside of a starfighter. Palpatine tried spinning, but he lost due to this technique (but this was intentional, as losing gave him the emotional high ground when Anakin arrived).

The reason for this is that spinning provides a yin-yang approach to combat (based in Eastern philosophy on balance), giving the spinner the high ground from above and below. Only the Chosen One can master the spin, as it is their destiny to maintain balance in the universe. This is why Obi-Wan was so emotional after defeating Vader on Mustafar; he expected to lose the high ground to the spin, but Anakin fell to the dark side and ceased to be the brother he knew, who could no longer use his signature trick.

Anakin doesn't hate sand for the reasons he told Padme; all Jedi hate sand, as the battlefield can rapidly change between low and high ground on multiple vectors, so your perspective must be from a certain three-dimensional point of view in order to comprehend who holds the high ground. This is the only reason why Obi-Wan killed Maul in Rebels. This is also why Obi-Wan hates flying; there is no gravity in space, therefore there is no high or low ground from any frame of reference.

In ANH, Vader proves his newfound mastery by engaging Obi on perfectly even ground. However, Obi-Wan intentionally sacrifices himself on the Death Star, so that he could train Luke from a higher plane of existence, thereby giving him the metaphysical high ground.

(Why was Vader so invested in the construction and maintenance of the Death Star? Because he knows Obi-wan can't have the high ground if there's no ground left.)

In Return of the Jedi, you can see that the Throne Room contains a variety of different altitudes; Palpatine placed these there to ensure Vader's defeat. However, Sheev failed to realize that his weakness was no ground, and should have covered that gaping pit which does nothing. In conclusion, Obi-wan abuses spatial relativity and Taoist doctrine in order to always invoke his high-ground powers.

EDIT: This extends a good meme beyond the meme, making it into a bad meme. Don't turn your good memes into bad memes.

PR #4766, opened 2019-03-14
Gives king goat boss a gps signal

fluffe9911:
Apparently people had trouble finding it and jazz so yeah has a signal now


MacHac:
Why not put it in the goat, like all the other megafauna?

fluffe9911:
because the goat itself is on a different z level in space so it would point to that z level and not the actual ruin itself

monster860 (approved):

PR #4772, opened 2019-03-14
Cluwnes can no longer be cloned

fluffe9911:
One does not escape from the curse so easily...


HomocidlePancakes:
Cluwnify them on cloning instead you nerd. Death is not an escape.

fluffe9911:
That would be better but im not certain how to make the mutation always transfer over

Edit will be once https://github.com/tgstation/tgstation/pull/43134 is merged

HomocidlePancakes:
Make cluwns seperate race and have them somehow not process mutation toxins? Iam not a coder so I will be speweing dumb shit

fluffe9911:
ok time greens fix didnt work seems this is the best solution atm so yeah lets just go with this

Altoids1:
I'unno, I'm personally against this.

Jc1469:
why not just have a personal floor cluwne hunt those dumb enough to get cloned.  if they survive the second coming, they get to live.

PR #4811, opened 2019-03-17
[admin] mark an issue as mentorhelp for fun and profit

monster860:


Bibby0110:
cant you like make it give them a button in the response text or just send it to mentorhelp straigt away idk

Altoids1:
Can we, instead, have it so that the admin can redirect the ticketer to a mentorhelp?

Also, while we're doing this, can we make it so that an admin can ascend a mentorhelp into being a ticket, when it turns out their question was an admin-required issue?

monster860:
no I'm too lazy

Bibby0110:
gulag

PR #4814, opened 2019-03-18
Adds an automatic FPS limiter to the Yogs Subsystem

Altoids1:
# \#CumInHisMailbox

image

In response to the Sseth putsch that's happened recently, I thought it was a good idea to port this prebase feature in, to help with the immense lag that naturally plagues a 140-pop Yogs server.

#### Changelog

:cl:  Altoids
experimental: The tickrate of the game now dynamically changes based on the player count. Lowpop rounds should feel smoother, and highpop rounds should be less of a laggy mess.
/:cl:


ThatLing:
what's the default fps currently?

Altoids1:
> what's the default fps?

# 20.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
FPS doesn't seem to have that much impact, what I did to make the game smoother was make the master controller process ~2x slower, increase the wait on subsystems other than garbage and set FPS to 24

Altoids1:
> FPS doesn't seem to have that much impact

I'm not entirely sure this is the case.

Regardless, I would like this merged so that we have *something* to combat the cum that is oozing out of our mailbox at the moment.

Amelia0010:
I would sincerely recommend changing how the client-FPS is set first, if you're doing this. As of right now, you can set your own fps in the preferences menu, and if this value isn't divisible by the server's fps, animations start to get really crappy and choppy. 

If you're doing this, you should make the client FPS be X times the server FPS instead of just a freeform value, since otherwise you'd need to constantly fiddle with it if the server FPS ever changes.

Nickvr628:
We are worse than damn dirty console peasants. Nvidia RTX support when

ThatLing:
I don't think it's will get that bad

PR #4833, opened 2019-03-19
[admin] Adds an "automatic panic bunker" that activates at a config-defined player count

Altoids1:
image

Uses config and a separate flag and all that, to allow for admins (or the config) to activate a panic bunker whenever the living player count exceeds a certain value.

Also comes with an admin verb (+SERVER,+ADMIN) to toggle it and set the player count.

#### Changelog

:cl:  Altoids
rscadd: There is now an Automatic Panic Bunker that admins can enable at ssethly numbers of players.
/:cl:


Nickvr628:
What will it be set to?

1fbff5f83b23d39d38b1dfcb4cac8d9b:
is it really necessary since we have a queue?

Altoids1:
> What will it be set to?

The default value is currently 75 living mobs that have players attached to them.

>is it really necessary since we have a queue?

This prioritizes regulars over newcomers, which seems to be how council wants this to be.

PR #4866, opened 2019-03-21
Reverts some of a /TG/ PR, making epipens cap oxygen damage again.

Altoids1:
Reverts parts of #4740.

Just want to note that this sets the oxygen damage cap to
# 40

Instead of the previous value of 

# 35

#### Changelog
:cl:  Altoids
tweak: Epipens now cap oxygen damage again! They no longer heal oxygen damage besides setting this cap, though.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -804,7 +804,11 @@
 		M.adjustToxLoss(-0.5*REM, 0)
 		M.adjustBruteLoss(-0.5*REM, 0)
 		M.adjustFireLoss(-0.5*REM, 0)
-		M.adjustOxyLoss(-0.5*REM, 0)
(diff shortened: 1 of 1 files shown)


fluffe9911:
But why

1fbff5f83b23d39d38b1dfcb4cac8d9b:
That seemed like a reasonable change to epi, why revert it?

Moja10:
Imsxz is right that epi which is free for everyone nullifies oxygen damage essentially, it is way too strong to be just given out for free. I think it's in an alright spot with the changes.

ThatLing:
No reason, downvote

Jc1469:
Imo the current epipen is crap, and the old one is OP.

Cap oxyloss to like 70 maybe so people are still horribly slow from damage and can't use it as a get out of jail free card if they run into a breach without internals.  Surely they'd take 30+ brute/burn from pressure/temp.

PR #4868, opened 2019-03-21
Revert "Revert "[MIRROR] Rework of admin rank loading to support punctuation""

Amelia0010:
I need to fix this so we're not too out of sync with tg.


ThatLing:
You have to edit webmin and stuff too

Amelia0010:
on second thought, I really do not want to

PR #4946, opened 2019-03-25
Adds More species to the charged Black extract

boodaliboo:
Before this pull request you could only get the round start races from this crossbreed which felt unfair as making atleast one of these could take around a hour atleast. so i made this pull request to fix it

🆑
rscadd: New species have been added to the charged black slime extract!
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/snail.dm
+++ b/code/modules/mob/living/carbon/human/species_types/snail.dm
@@ -12,7 +12,7 @@
 	speedmod = 6
 	punchdamagehigh = 0.5 //snails are soft and squishy
 	siemens_coeff = 2 //snails are mostly water
-	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_MAGIC | MIRROR_PRIDE | RACE_SWAP 
+	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_MAGIC | MIRROR_PRIDE | RACE_SWAP
 	sexes = FALSE //snails are hermaphrodites
 	var/shell_type = /obj/item/storage/backpack/snail
 
--- a/code/modules/mob/living/carbon/human/species_types/snail.dm
+++ b/code/modules/mob/living/carbon/human/species_types/snail.dm
@@ -12,7 +12,7 @@
 	speedmod = 6
 	punchdamagehigh = 0.5 //snails are soft and squishy
 	siemens_coeff = 2 //snails are mostly water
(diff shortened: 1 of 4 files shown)


Jc1469:
how about not giving xeno more stuff thanks

boodaliboo:
> how about not giving xeno more stuff thanks

and why is that? most xeno bio items take most of the round to get the only thing thats really unbalanced is gold slimes

drderp3635:
Might I point out to anyone who doesn't know how godawful inefficient charged black slimes are. You use 10 yellow extracts to an adult black, then you have to put 10 plasma in it. Then you get one random race roll, some of which are just awful. The actual amount of work you would have to put forward to actually get a specific race, is far more than what it is worth.

code/modules/mob/living/carbon/human/species_types/snail.dm
```diff
@@ -12,7 +12,7 @@
 	speedmod = 6
 	punchdamagehigh = 0.5 //snails are soft and squishy
 	siemens_coeff = 2 //snails are mostly water
-	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_MAGIC | MIRROR_PRIDE | RACE_SWAP
+	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_MAGIC | MIRROR_PRIDE | RACE_SWAP | SLIME_EXTRACT
```
ThatLing: No

ThatLing (changes requested):

code/modules/mob/living/carbon/human/species_types/snail.dm
boodaliboo: how do i change it to that?

boodaliboo: i think i did it

code/modules/mob/living/carbon/human/species_types/snail.dm
```diff
@@ -12,7 +12,7 @@
 	speedmod = 6
 	punchdamagehigh = 0.5 //snails are soft and squishy
 	siemens_coeff = 2 //snails are mostly water
-	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_MAGIC | MIRROR_PRIDE | RACE_SWAP
+	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_MAGIC | MIRROR_PRIDE | RACE_SWAP
```
monster860: AAAA YOU ADDED A SPACE AT THE END

monster860 (changes requested):

PR #4954, opened 2019-03-27
[admin] Admins are now warned if they try to change the roundendsound if another admin has already set it to something

Altoids1:
Just in case any of you have to admin with Groudon or something.

#### Changelog

:cl:  Altoids
tweak: Admins are now warned if they try to set the roundendsound when it's already been set by another admin.
/:cl:


ThatLing:
I propose an alternate solution: remove setting the roundendsound

PR #4989, opened 2019-03-30
Resonators now replicate fields as god intended

SomeguyManperson:
:cl:  Theos& Super3222
rscadd: Resonator fields will now replicate onto adjacent ores of the same kind if they detonate naturally
/:cl:


Jc1469:
tbh fam, i think it'd be best if it worked on stone. just get rid of all the stone so you can kill the server and then when it catches up half an hour later, mine your diamonds in peace

also why does this have a dme edit label

SomeguyManperson:
For some reason my local copy felt it was deathly important to move something one line in the dme, I removed that in the second commit

PR #5027, opened 2019-04-02
Removes baseball bats from the bounty rotation since they are uncraftable

SomeguyManperson:
:cl: theos  
bugfix: baseball bats are no longer a possible cargo bounty
/:cl:


Altoids1:
Yeah fair enough

PR #5035, opened 2019-04-03
Makes OOC pinging WAAAAAY smarter.

Altoids1:
## This took like 200 minutes of my life

https://regex101.com/r/YtmLDs/7

Right now, in order to use the OOC pinging I added a while back, you have to write in the exact ckey, as it is in the database (the ``ckey()`` version). Now it should probably allow you to type in any legal variant of the ckey (albeit it still must be written in full).
So now you can do, y'know
- ``@the slippery carp``
- ``@THESlipperyCarP``
- ``@theslipperycarp``

et cetera, cetera.

Additionally, the ping spam detection is also a lot smarter, and will only do spam checks once it knows you are actually going to ping a client.

#### Changelog

:cl:  Altoids
bugfix: Dude, you have NO IDEA how fuzzily-snuzzily the OOC pinging regex is now. It's so good.
bugfix: You should no longer be stopped from using OOC after putting @ next to some of your words.
/:cl:


code/modules/client/verbs/ooc.dm
```diff
@@ -71,18 +71,25 @@ GLOBAL_VAR_INIT(mentor_ooc_colour, YOGS_MENTOR_OOC_COLOUR) // yogs - mentor ooc
...
-	if(ping.Find(msg))
+	
+	//PINGS
+	var/regex/ping = regex(@"@+(((([\s]{0,1}[^\s@]{0,30})[\s]*[^\s@]{0,30})[\s]*[^\s@]{0,30})[\s]*[^\s@]{0,30})","g")//Now lets check if they pinged anyone
+	// Regex101 link to this specific regex, as of 3rd April 2019: https://regex101.com/r/YtmLDs/7
+	var/list/pinged = list()
+	while(ping.Find(msg))
+		for(var/x in ping.group)
+			pinged |= ckey(x)
+	pinged &= GLOB.clients // If the "SENDING MESSAGES OUT" for-loop starts iterating over something else, make this GLOB *that* something else.
```
Altoids1, line 82: Wait, would this actually work? the things in ``GLOB.clients`` ought to be of type, well, ``/client/``, not the literal string names of'em.

PR #5092, opened 2019-04-10
Prevents syndiborgs from using donor skins

Altoids1:
image

Apparently KMC was doing the "Are you syndicate?" check with this:
```dm
        if(module.name == "Syndicate")
            to_chat(src, "You cannot reskin a syndicate cyborg :(")
            return FALSE
```
Which doesn't work when the borg is a syndiborg that does not have the generic Syndiborg module (like the saboteur or medical ones).

So, uh, fixed that. Also tweaked some of the dialogue to make it look nicer.

#### Changelog

:cl:  Altoids
bugfix: Syndicate Cyborgs can no longer use donor skins.
spellcheck: Tweaked some of the Donorborg dialogue, added some spans and grammar fixes.
/:cl:


GuppyLaxx:
self merge this is power abuse

PR #5132, opened 2019-04-18
Attempts to fix Atmos Miner machinery not spreading its gases to the environment

Altoids1:
### This is about 6000 mols of plasma just J chillin on this tile, here, not affecting its environment.
image

I *think* it has something to do with this proc:
```dm
/turf/proc/CalculateAdjacentTurfs()
	var/canpass = CANATMOSPASS(src, src) 
	var/canvpass = CANVERTICALATMOSPASS(src, src)
	for(var/direction in GLOB.cardinals_multiz)
		var/turf/T = get_step_multiz(src, direction)
		if(!isopenturf(T))
			continue
		if(!(blocks_air || T.blocks_air) && ((direction & (UP|DOWN))? (canvpass && CANVERTICALATMOSPASS(T, src)) : (canpass && CANATMOSPASS(T, src))) )
			LAZYINITLIST(atmos_adjacent_turfs)
			LAZYINITLIST(T.atmos_adjacent_turfs)
			atmos_adjacent_turfs[T] = TRUE
			T.atmos_adjacent_turfs[src] = TRUE
		else
			if (atmos_adjacent_turfs)
				atmos_adjacent_turfs -= T
			if (T.atmos_adjacent_turfs)
				T.atmos_adjacent_turfs -= src
			UNSETEMPTY(T.atmos_adjacent_turfs)
	UNSETEMPTY(atmos_adjacent_turfs)
	src.atmos_adjacent_turfs = atmos_adjacent_turfs
```
So, I've removed the call to this proc from this machine, since the SM doesn't call this proc and *it* works, so...

#### Changelog

:cl:  Altoids
bugfix: Gas Mining machines now properly release their gases, instead of just bunching them all up on the machine's tile.
/:cl:


Altoids1:
https://github.com/tgstation/tgstation/commit/c1586e840a61458d217c1daac59e8774816c221f

I'm not even sure if this would even work. :/

PR #5145, opened 2019-04-19
[admin] Adds an +ADMINverb that allows admins to delay the shuttle from launching

Altoids1:
image

#### Changelog

:cl:  Altoids
rscadd: Admins can now delay the shuttle from launching!
/:cl:


Amelia0010:
they can already do this by editing the timer in check antags though, can they not?

Altoids1:
Sure, but this uses the proper delaying system that Gangs and the Blob use. The other option, of just adding time over and over, is IMO a bit sloppier.

Amelia0010:
why not just change that to use that system then?

PR #5149, opened 2019-04-20
FIX VIRO STEALTH HIDING

ghost:
### FIX VIRO STEALTH HIDING

A virus with stealth >= 4 should be invisible to pandemic machines.

#### Changelog

:cl:    
bugfix: Viruses with sufficient stealth are now hidden from pandemic machines. 
/:cl:


boodaliboo:
this is litterally the worst pr ever. think about it like this your making a virus you for example add a stealth symptom it goes over 4 now you cant see all your other symptoms. make this 10 or higher and this pr will be A ok but otherwise its terrible

PR #5151, opened 2019-04-20
remove medical sim from perma

ghost:


HomocidlePancakes:
Anime school has a telibaton btw. Also holopad got deleted by holodeck. Plox fix

PR #5178, opened 2019-04-23
Fixes Nightmares not having snazzy names

### LINKED ISSUE #5156: Nightmares take from the human name pool.
Issue reported from  Round ID: 22642 (Yogstation 13)

Reporting client version: 512
image

Altoids1:
```dm
/mob/proc/fully_replace_character_name(oldname,newname)
```
So the thing is that this proc takes *two* arguments: The old name, and the new name.

Apparently /TG/ coded in Nightmares only giving it one argument, meaning it would sliently just give up and return ``0`` instead of giving the guy a new name, so, uh, fixed that then.

Fixes #5156.

### Changelog

:cl:  Altoids
bugfix: Nightmares should now have actually spooky names.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/shadowpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/shadowpeople.dm
@@ -49,7 +49,7 @@
 	. = ..()
 	to_chat(C, "[info_text]")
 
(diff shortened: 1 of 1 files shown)


Altoids1:
image
Works.

PR #5215, opened 2019-04-24
Revert "Forces IAA to get good and removes their access to ez mulligan"

alexkar598:
Reverts yogstation13/Yogstation-TG#5183

Fix the tracker to always track the target instead of removing a traitor item altogether from a gamemode.

### DIFF
--- a/yogstation/code/modules/uplink/uplink_item.dm
+++ b/yogstation/code/modules/uplink/uplink_item.dm
@@ -39,9 +39,6 @@
 /datum/uplink_item/role_restricted/his_grace
 	include_objectives = list(/datum/objective/hijack)
 
-/datum/uplink_item/stealthy_tools/mulligan
-	exclude_modes = list(/datum/game_mode/nuclear, /datum/game_mode/nuclear/clown_ops, /datum/game_mode/traitor/internal_affairs)
-
 //////////////////////////
 /////////New Items////////
 //////////////////////////


MegaEmpirical:
Either give them a tracker or remove mulligan, I can live with either one. However, I wonder if we should consider removing mulligan from normal traitor rounds? It effectively makes you immune to being killed randomly mostly since we're not allowed to just kill everyone.

SomeguyManperson:
There is literally no reason to have an item that changes your identity readily available on a gamemode where you get assassination objectives against other traitors with only the original name and an inaccurate pinpointer to find them. It effectively means any rando at all could be a target instead of just any rando with a mask/the actual target

HomocidlePancakes:
Ugh. You can make mulligan in chem. Also genetics exist, renaming potion, chugging mutagen in order to change you appearance

HomocidlePancakes:
Magic mirror, plastic surgery etc.

SomeguyManperson:
Chem mulligan is expensive and requires xenobiology to have green slimes, the magic mirror is only available as a ruin at lavaland and spaces you on use, surgery requires sitting still for a dangerous amount of time with someone else in the room, and mutagen also gives other random mutations plus rads. Even genetics requires you to stop moving for some time.

Mulligan syringes cost... 4 tc each. And can be bought and used anywhere. You can basically start with 16 tc for an instant advantage over everyone else, allowing you to buy a desword or revolver while having a considerable advantage over everyone else already. And you can buy *another* syringe every time you kill a target if you don't need to save up for anything.

alexkar598:
Changed my mind on this.

PR #5300, opened 2019-04-29
A couple new spells for wizard, Cauterize and Slip

Putnam3145:
To celebrate wizard happening more often and the readding of ragin mages, I decided to add a couple fun spells I had knocking around in my head. I don't think they're particularly good, but more variety's always fun.

Obviously these numbers can be tweaked and stuff modified as necessary.

:cl:  
add: new spell: cauterize. Heals all brute/burn damage, then adds it back entirely as burn damage over the next 20-60 seconds, depending on spell level.
add: new spell: slip. Makes the floor in an area around the caster slippery, as if lubed, for 30 seconds. First level has cooldown equal duration!
/:cl:


yogstation/code/modules/spells/cauterize.dm
```diff
@@ -0,0 +1,35 @@
...
+	name = "Cauterize"
+	desc = "Replace all brute/burn damage with burn damage, taken over time. Higher levels increases time taken to return back to original damage level."
+	clothes_req = TRUE
+	human_req = FALSE
+	charge_max = 800
+	cooldown_min = 600 //50 deciseconds reduction per level
+	range = -1
+	var/cauterize_duration = 20 //in seconds
+	include_user = TRUE
+	invocation = "HEA'OW!" //someone come up with a good invocation i beg you
```
Altoids1: Maybe it could be the fireball invocation in reverse?

Putnam3145: Good idea. Done.

Nickvr628:
Why would you voluntarily replace brute damage with burn damage? Especially since you can take the spell point and use it on something useful, like the healing spell or healing wand.

I like the slipping spell though.

Putnam3145:
It's more the fact that it's over time--if you're highly damaged, it'll be a sudden temporary burst of speed, and could keep you from going into crit if used wisely, and unlike the healing wand can't be stolen from you. Main reason it's all burn is to make it easier to deal with through external means, which is kind of a gameplay style that I like, the wizard going in and stealing a medkit for burn treatment and such.

Of course, it just never being used is always a possibility, and one I don't really mind. I just like the variety.

ghost:
that is hideously disgusting considering some of the best medicines in the game are anti-burn and it removes the threat of bleeding entirely and makes being shotgunned with buck less of an issue

that's pretty much a guaranteed take of a spell

also space lubing the floor is going to backfire on the wizard too, so

PR #5306, opened 2019-05-01
Fixes a conservation of mass error in Tritium fires, causing weirdly high temperatures and negative masses

Altoids1:
https://github.com/yogstation13/Yogstation-TG/blob/10e6f32c4f2eda3f6fcc1c06d47c053bbc05cc8a/code/modules/atmospherics/gasmixtures/reactions.dm#L104-L121
image

So it turns out that a lot of what made Tritium fires so damned hot was the fact that they were actually making mass disappear, causing the energy density to increase at levels that should not be expected from the actual energy output of tritium fires/bombs.

I've also taken the liberty of including a small linechange which I think might be able to improve performance.

#### Changelog

:cl:  Altoids
bugfix: Tritium fires no longer violate the conservation of mass.
bugfix: Small tritium fires no longer cause negative amounts of Tritium.
tweak: Tritium fires now produce 100x more water vapor.
tweak: Tritium fires may be somewhat more radioactive in some circumstances.
/:cl:


Altoids1:
Tests on Centcom revealed that, on a per-proc-call basis, ``tritfire/react()`` takes 48.8 microseconds to run with this PR, compared to 62.2 microseconds on my copy of master. So that's **27%** faster.

This proc gets called tens of thousands of times during a tritium fire so that does actually mean shaving full seconds off of calculating them.

MegaEmpirical:
Nicely done. How did you even notice/come across this?

Altoids1:
I was talking with Readystorm546 about how to make a tritium fire, when I noticed during my code surf that not only did tritium have like, two possible chemical reactions with oxygen depending on their concentration & temperature, but also that the code under the if-statement actually was capable of causing negative masses.

Staring at the code more then revealed the possibility of this not being a balanced chemical equation, which I then did the stochiometry for to generate the graphs displayed in the OP.

PR #5310, opened 2019-05-01
Blood Brothers now spawn with a Random Bundle

fluffe9911:
🆑
tweak: Blood Brothers should now spawn with a random syndicate meme bundle!
/🆑

### DIFF
--- a/code/game/gamemodes/brother/traitor_bro.dm
+++ b/code/game/gamemodes/brother/traitor_bro.dm
@@ -70,3 +70,19 @@
 	var/datum/atom_hud/antag/brotherhud = GLOB.huds[ANTAG_HUD_BROTHER]
 	brotherhud.leave_hud(brother_mind.current)
 	set_antag_hud(brother_mind.current, null)
+
+/datum/game_mode/proc/equip_brother(mob/living/M) //Yogs start gives the brothers a bundle
+	if(!M || !ishuman(M))
+		return FALSE
+	var/mob/living/carbon/human/L = M
+	var/obj/item/storage/box/syndicate/bundle_B/S = new
+	var/list/slots = list("In your backpack" = SLOT_IN_BACKPACK)
+	if(ishuman(L))
+		var/mob/living/carbon/human/H = L
+		var/slot = H.equip_in_one_of_slots(S, slots)
+		if(slot == "In your backpack")
+			slot = "In your [H.back.name]"
+	if(S)
+		to_chat(L, "<span class='bold'>There is a syndicate bundle in your backpack! It'll help you get a slight leg up on your objectives but dont expect it to make things too easy!.</span>")
+		return TRUE
+	return FALSE //Yogs end gives the brothers a bundle
--- a/code/game/gamemodes/brother/traitor_bro.dm
+++ b/code/game/gamemodes/brother/traitor_bro.dm
@@ -76,13 +76,10 @@
 		return FALSE
 	var/mob/living/carbon/human/L = M
 	var/obj/item/storage/box/syndicate/bundle_B/S = new
-	var/list/slots = list("In your backpack" = SLOT_IN_BACKPACK)
-	if(ishuman(L))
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/brother/traitor_bro.dm
```diff
@@ -70,3 +70,19 @@
...
+	var/mob/living/carbon/human/L = M
+	var/obj/item/storage/box/syndicate/bundle_B/S = new
+	var/list/slots = list("In your backpack" = SLOT_IN_BACKPACK)
+	if(ishuman(L))
+		var/mob/living/carbon/human/H = L
+		var/slot = H.equip_in_one_of_slots(S, slots)
+		if(slot == "In your backpack")
+			slot = "In your [H.back.name]"
+	if(S)
+		to_chat(L, "<span class='bold'>There is a syndicate bundle in your backpack! It'll help you get a slight leg up on your objectives but dont expect it to make things too easy!.</span>")
```
Altoids1: What's all this ``slot`` stuff for, besides for stuffing the box in their backpack? You don't actually use its value in the ``to_chat`` message given to the bloodbro.

Altoids1 (changes requested):

code/game/gamemodes/brother/traitor_bro.dm
fluffe9911: good point

HomocidlePancakes:
This is dumb. They don’t need traitor items. Give them some sort of untraceable way to communicate instead, because ever since siggytech got reimplemented contacting your brother is just dangerous.

code/game/gamemodes/brother/traitor_bro.dm
fluffe9911: Never mind travis complains leaving it as is for now

Lynameep:
Spicing up Bloodbrothers seems like a lot of fun.

HomocidlePancakes:
Nothing spicy about giving traitor items to antags in a mode where traitors already exist

ynot01:
giving them 20TC bundles isn't balanced, but ~5TC uplinks would be

MegaEmpirical:
My overall concerns:

20TC bundles: these are worth more than 20 TC unless you're planning custom ones and good god, you're still effectively giving 40 TC to two players that will work together.

5TC (and there about): Translates to giving them 10 TC. Which if there is a doctor, means they can get the brain washing surgery (and have leftover) and 2 blood brothers becomes 3 or more. Or they could get the hypnotic flash (I'm actually not sure if we have this enabled, and still have a little left over).

HomocidlePancakes:
The whole point of blood bros is that you get a teammate for the cost of your TC. Getting robusted? You get help from your bro instead of a stechkin. Got cuffed? Your bro is there instead of a freedom implant. Thrown in perma? Your bro is already in space with a suit for you and there isn’t a need for a storage implant + emag. 

Atleast in theory. Right now this won’t happen very often due to the lack of coordination. How do we fix this? We give them a way to communicate with each other akin to ling hivemind. Maybe even give them pinpointers pointing at each other. Giving them traitor items won’t fix lack of communication which is something a TEAM ANTAG shouldn’t struggle with.

Altoids1 (approved):

Altoids1:
```dm
/mob/living/carbon/proc/equip_in_one_of_slots(obj/item/I, list/slots, qdel_on_fail = 1)
	for(var/slot in slots)
		if(equip_to_slot_if_possible(I, slots[slot], qdel_on_fail = 0, disable_warning = TRUE))
			return slot
	if(qdel_on_fail)
		qdel(I)
	return null
```
So apparently you're forced to input the possible slots as a hashtable, yet it doesn't even really do anything with the key?

Eugh.

code/game/gamemodes/brother/traitor_bro.dm
Altoids1: You lazy fuck

PR #5314, opened 2019-05-02
Fixes the weighted headrole assignment thing not working at all

Altoids1:
Caused by https://github.com/yogstation13/Yogstation-TG/pull/4732

There were two problems:
- There was a runtime inside of ``PickCommander()`` that stopped it from working
- There are actually *two* procs responsible for dishing out headroles (one of which was actually the only one that's been working for like month due to the aforementioned runtime), and ``PickCommander()`` was only placed in the (ergo broken) one.

I fixed these two problems.

#### Changelog

:cl:  Altoids
bugfix: Weighted headrole assignment should now actually properly work.
/:cl:


ThatLing:
haha good riddance, crap system

Altoids1:
😢

PR #5323, opened 2019-05-04
Haves Blood Brothers spawn with 5TC

ynot01:
this is solely to replace the stupid 20TC bundles they spawn with and enable more diversity instead of 5 kits among 12 blood brothers in 1 round

#### Changelog

:cl:  
rscadd: Blood Brothers now spawn with lesser uplink implants
/:cl:

### DIFF
--- a/code/game/objects/items/implants/implantuplink.dm
+++ b/code/game/objects/items/implants/implantuplink.dm
@@ -21,3 +21,10 @@
 
 /obj/item/implant/uplink/precharged
 	starting_tc = 10
+
+/obj/item/implanter/uplink/brothers
+    name = "implanter (lesser uplink)"
+    imp_type = /obj/item/implant/uplink/brothers
+
+/obj/item/implant/uplink/brothers
+    starting_tc = 5
--- a/code/game/objects/items/implants/implantuplink.dm
+++ b/code/game/objects/items/implants/implantuplink.dm
@@ -22,9 +22,9 @@
 /obj/item/implant/uplink/precharged
 	starting_tc = 10
 
-/obj/item/implanter/uplink/brothers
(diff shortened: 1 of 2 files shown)


MegaEmpirical:
As I said in the other PR and elsewhere:

20TC bundles: these are worth more than 20 TC unless you're planning custom ones and good god, you're still effectively giving 40 TC to two players that will work together.

5TC (and there about): Translates to giving them 10 TC. Which if there is a doctor, means they can get the brain washing surgery (and have leftover) and 2 blood brothers becomes 3 or more. Or they could get the hypnotic flash (I'm actually not sure if we have this enabled, and still have a little left over).

ynot01:
@MegaEmpirical i dont really see the hypnotize thing as a problem, if they want to spend all of their resources on another party member just to come back to the way things were now, that seems fine

Ratvar:
Brain washing surgery is not a problem because it's either 3+ brothers (blood brothers) or 3+ traitors (in infamous traitor gamemode), and first is harder because it requires actual cooperation and has 10 less tc. Maybe your problem is with item itself?

MegaEmpirical:
Brain washing is 5 TC, so they will have 5 left over, and infinite additional blood brothers.

Ratvar:
That's issue with brainwashing itself, dude, not with blood brothers.
 You can get it on traitor/traitorling/traitorlingabductor as well, which means infinite traitors, and you will have more TC/no objective to keep someone else alive as well

HomocidlePancakes:
Triplets 👀

boodaliboo:
i kinda like fluffes crate idea more its more unique and fun

ynot01:
im never gonna work on this so im closing

PR #5327, opened 2019-05-04
Fixes weird focusing issues with the Comms Console's browser window

Altoids1:
So TheSlipperyCarp informed me today that there was some weird issue with the Comms console browser window randomly focusing while AIs operated it, forcing them to engage in a focus battle.

When I wandered on over to the Comms console code, I saw this in Topic():
```dm
	popup.set_content(dat)
	popup.open()
	popup.set_content(dat)
	popup.open()
```
which meant that the window was updating itself twice over every time it updated.

I checked the PR that caused this, and it happened to be the one that implemented alternative shuttles: https://github.com/tgstation/tgstation/pull/21398

However OP is a f't and doesn't provide explanation nor comment on why it was duplicated, so I didn't know for sure if it was intentional.

However, I did come across Nich's port of this PR, over here: https://github.com/yogstation13/yogstation/pull/2728/files

Which did not have the same duplication, and I believe that PR worked fine. Ergo, I can conclude this is some dumb /TG/ bullshit that ought to be snipped out.

#### Changelog

:cl:  Altoids
bugfix: The Comms Console browser window should no longer get into fights with you when you operate it as an AI.
/:cl:

### DIFF
--- a/code/game/machinery/computer/communications.dm
+++ b/code/game/machinery/computer/communications.dm
@@ -379,7 +379,9 @@
 		if("ai-status")
 			aistate = STATE_STATUSDISPLAY
 		if("ai-announce")
+			is_announcing = TRUE
 			make_announcement(usr, 1)
+			is_announcing = FALSE
 		if("ai-securitylevel")
 			tmp_alertlevel = text2num( href_list["newalertlevel"] )
 			if(!tmp_alertlevel)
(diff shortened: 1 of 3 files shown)


Altoids1:
Currently doesn't fix the problem. Will work on it.

MegaEmpirical:
Finally being addressed, thank you.

Partheo:
https://github.com/tgstation/tgstation/pull/43909

PR #5357, opened 2019-05-06
[admin] Allows admins to view and manipulate the server queue

Altoids1:
So as it stands, there's no real easy way of viewing the queue and interacting with it as an admin in a live game without fancy varediting, so I thought it'd be convenient if there were some additional verbs to viewing the server queue and emptying it in the case of a bug.

All-in-all this is mostly just to help me debug this so that the 20-man queues that we used to have don't last for nearly as long as they did ever again.

#### Changelog
:cl:  Altoids
rscadd: Admins can now view how many people are in the server queue, as well as empty the queue to let'em all in.
/:cl:


HomocidlePancakes:
Altoids making features for his metafriends huehue

PR #5371, opened 2019-05-11
Yogstation Atmos APC Fix

Lynameep:
Fixes the issue with there being two APCs powering the same area in Atmos by separating the two areas. This should not affect any other maps as this creates an entirely new Area specifically to yogstation.

#### Changelog

:cl:  
experimental: Atmospherics Distribution is now its own separate /area/.
bugfix: Fixed Yogbox's atmospherics being powered by two APCs at the same time.
/:cl:


HomocidlePancakes:
:eyes: *looks over at departures having 2 APCs*

Lynameep:
Wasn't aware of those but I can fix that as well.

Be sure to send me a picture of the APCs as I don't seem to be able to find them.

PR #5458, opened 2019-05-19
Ratvar can now 360 noscope narsien constructs

SomeguyManperson:
:cl: 
tweak: Ratvar will now dunk on virgin constructs instead of just trying to set them on fire like a casual
/:cl:


SomeguyManperson:
how did that happen

PR #5464, opened 2019-05-20
Barrier rune code cleaning

SomeguyManperson:
Sanity saving stuff that makes the barrier rune easier to look at/probably less cancer
:cl:  
bugfix: barrier rune will now actually use its cooldown period when disabled
tweak: barrier rune code is less hot garbage
/:cl:


fluffe9911:
Altoids says you should pr this to tg sorry

SomeguyManperson:
We're debased from tg so I'm reviving this from the grave

PR #5508, opened 2019-05-24
Move the service hall (yogbox)

Ktlwjec:
Moves the service hall so the janitor can actually access it.
Brings back the old botany back room.
Fixes two issues - https://github.com/yogstation13/Yogstation-TG/issues/5364 and a double sink in yogbox dorms.

First pr on my new pc so we'll see if I set it up right

image


MegaEmpirical:
What was stopping the janitor from accessing it before?

Ktlwjec:
ok easily access then also it was ugly in maint and upset me

MegaEmpirical:
That's fair. I'm all for this anyway. I think it looks nicer.

PR #5509, opened 2019-05-25
Tweaks ashwalker flavortext

Ktlwjec:
Doesn't bother me but some admins were complaining since we allow ashlizards on the station.


ThatLing:
ok but you should still defend your nest

PR #5517, opened 2019-05-26
Adds in a auto label bot for issues

alexkar598:
This PR adds the config file for a bot that uses machine learning to label issues into three categories: Bug,Feature Request or Question(never encountered that tag in tests tho).

To use the bot,you obviously need to add it first: https://github.com/marketplace/issue-label-bot

You can see an example of its doing at https://github.com/alexkar598/testinglabelbot/issues where I took random issues from this github and copied them over(il save you a click,out of 11 issues,it sorted 10 correctly and didn't take action for one issue due to uncertainty)

This bot will create a Question label tho i think,or it might just not label stuff it considers as a "Question".

### DIFF
--- a/.github/issue_label_bot.yaml
+++ b/.github/issue_label_bot.yaml
@@ -1,0 +1,4 @@
+label-alias:
+  bug: 'Bug'
+  feature_request: 'Feature Request'
+  question: 'Question'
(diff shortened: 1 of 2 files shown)


alexkar598:
REEEEEE,this packed some commits which shouldn't of been there,fixing it

alexkar598:
Looks like the commits didn't do any change anyways so im just going to leave em here.

PR #5383, opened 2019-05-15
Brings back the Centcomm Test Chamber

Lynameep:
Readds the previously loved centcomm test chamber that contains numerous items both obtainable and not.

- Short but not complete list:

- Wizard Spellbooks, Wands, Weapons, and Items

- Traitor Bundles, Weapons, and Items

- Countless Implants and Mutators

- Lavaland Loot and Megafauna Loot

- The Supermatter Sword and Instakill Rifle

- All of the available Mechs

- A Wabbajack Statue and Magic Mirror

- Anti-Rush Design

- And so much more...

http://prntscr.com/noxtwq

#### Changelog

:cl:  
rscadd: Adds the Centcomm Test Chamber where the Courthouse used to be.
tweak: Moved the Centcomm Courthouse to be east of the admin Prison.
/:cl:

PR #5519, opened 2019-05-26
CentCom Test Chamber v2

Lynameep:

An Alternative to my previous test chamber #5383 pr.

It's not only much bigger but it's completely isolated from the rest of centcom and can easily be disabled by removing the portal from centcom during the game at any point.

Things that were added;
 - Megafauna 
 - Toxins Testing
 - All of the available gasses
 - A lot more explosives
 - A generally more nicer look

#### Changelog

:cl:  
rscadd: Added the Centcom Test Chamber which is accessible by a portal after arriving at centcom.
/:cl:

### DIFF
--- a/_maps/yogstation/map_files/generic/CentCom.dmm
+++ b/_maps/yogstation/map_files/generic/CentCom.dmm
@@ -19709,7 +19709,9 @@
 	dir = 1;
 	light_color = "#c1caff"
 	},
-/mob/living/simple_animal/hostile/megafauna/colossus,
+/mob/living/simple_animal/hostile/megafauna/colossus{
+	true_spawn = 0
+	},
 /turf/open/floor/plasteel/bluespace,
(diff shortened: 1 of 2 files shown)


Ktlwjec:
changelog different please, but beautiful work from my quick glance 😎

boodaliboo:
you do coding for me for syndicate pod

Altoids1:
The megafauna in the test chamber seem to be tampering with the normal megafauna's ability to spawn, so that has to be fixed.

Also, there's a stray exit portal in the test chamber that needs to be deleted.

Altoids1 (changes requested):
The megafauna in the test chamber seem to be tampering with the normal megafauna's ability to spawn, so that has to be fixed.

Also, there's a stray exit portal in the test chamber that needs to be deleted.

PR #5520, opened 2019-05-26
Creates multi-exit & multi-entrance one-way portals

Altoids1:
image

This is a feature necessary for something DotLyna wanted to do with #5519.

#### Changelog
:cl:  Altoids
rscadd: One-way portals can now sometimes send you to multiple different places, from multiple different entry points!
/:cl:


Altoids1:
image

Tested it, it works.

Issue #5037, opened 2019-04-03
Janitor's holographic sign projector does not fit into his jani belt


MegaEmpirical:
Unlike every other holo projector and associated department belts.

Lostlarry:
this is issue of wrong holoprojectors being vended ill see if i can find where their file is

MegaEmpirical:
... There's more than one? Any reason to keep them both in the code?

Lostlarry:
well there is general class holoprojector that defines behaviour common to every type it just happens that this default is almost same as janni one

MegaEmpirical:
Fascinating. Fix merged, closing. Thanks for fixing.

PR #5580, opened 2019-05-28
Jani wardrobe now wends right holoprojectors

Lostlarry:
as descibed in issue #5037 the janni wardrobe on Yog station was not equped corectly.

:cl:  
tweak: Fixes janni wardrobe so holoprojector can be put in the janitor belt
/:cl:


Lostlarry:
just for the record this error was on all stations i looked at so far
thanks copypasta

Ktlwjec1:
Just saw this was you, I fixed it now but in the future don't edit vendors on the map. Do it in the relevant dm file.

PR #5606, opened 2019-05-30
Adds a fun new ERP-pass

Amelia0010:
if we can have a bigotry pass, why not an ERP pass?


AsV9 (approved):

Mr-Space:
Thank you nich, very cool.

Kmc2000:
You actually wrote out an ERP session with this

Genius

PR #5611, opened 2019-05-31
[MANUAL MIRROR] [READY] Outputs Datum Cooldown: For Real Guys This Time

fluffe9911:
🆑 Basilman
fix: Added a cooldown for datum outputs
/🆑

also added an arg called do_owner in playsound that lets the owner of a datum output be different from the atom that plays it, chiefly used for component/squeak so that you can apply the component directly to any single atom without having to fuck with said atom's datum_outputs list so that it can properly play the sound, since the datum output is already stored within the datum component itself.

also send_info now returns true or false based on whether it's cooling down or not, i dont like this and i initially just had a var for whether it's cooling down or not but raz said it saves a var so i guess whatever ill just have it like that


fluffe9911:
travis still being dumb

Altoids1:
Travis is no longer being dumb; now it's your fault

fluffe9911:
hec

PR #5608, opened 2019-05-31
WOW! Moves the AI core to Science and configures Science a bit to cater for this change. 

nesisknesis:
When I mean Moves the AI core, I mean I gave it the big snip. It is now a lot smaller and partly exposed to space, but now has extra security measures to balance the scales.

AI Core:
image

This is where the EXPERIMENTOR used to be, meaning the AI is now apart of Science and at the backdoor of the RD. Lamarr is a shit pet. Now the AI is truly his pet.

But wait, wheres the experimentor?

Here:
image
Cozy.

Back to the AI, this shit is small as fuck, also wasnt there a maint going through here? yes, there was. its now in space.

image

Great. But isnt the AI completely buggered at the backside? Not exactly.

It now has a toggleable core shield, get rekt liberals

image

Also, the RD got some loving to their office, that budget isnt there for show.

image

Their own bed, purple comfy chair, and cyborg statue which picks from random names each round. You'll have to see what names they are in-game or by looking at the code. And that door goes directly to the AI upload access.

image
💦 

How they connect:
image

WAIT. THE ORIGINAL AI CORE?

image

gone mate.

perhaps engineering can do something meaningful with this space or at a later date someone else. cant remember who was doing the gateway but here is plenty space for one.

and finally, the bridge:

image

maybe now the heads of staff can have a nice sit down together without being isolated in a crummy meeting room. also pizza. nanotrasen doesnt leave their senior staff unrewarded

bunker mode:

image

the doors are practically the same.

answers to the questions you may have:

isnt that ai shield op. - no, it can be disabled by killing the APC located in the open area, then it requires a crowbar to pry it open. also they can be sploded by c4.

its fully surrounded by glass you moron, how does it get out. - get some tools.

that space bridge is dangerous someone could get hurt. - ok,

what is that weird ball thing near the bridge. - a radar for fluff

why that poster - rip it down or keep it there

why are you sucking the rd's phallus like this - the rd is suppossed to keep the silicons in check, and with the ai core being right next to their office, now is much easier and they have the incentive to do so since their lazy ass just have to open 2 doors from their spawn

finally final. theres an easter egg

#### Changelog

:cl:  Identification
rscadd: Adds a cyborg statue that picks from random names each round.
rscadd: Adds a purple comfy chair to the RD's office.
rscadd: Adds a radar.
rscadd: Adds a fire extinguisher to the Captain's office.
wip: Original AI core was removed. Empty space is now apparent which can be filled by anyone willing or ingame by opportunists.
[...]

PR #5617, opened 2019-05-31
Redoes the RD office, adding two new objects and shifting the orientation.

nesisknesis:
Separate PR for the RD office included in #5608.

Image: (ignore the objects on the other side of the right wall and the red marker)
image

A cyborg statue has been added which picks different names each round, you'll have to find out in game what these names are or by looking in the code.

The RD also has their own purple comfy chair and bed.

A bugfix included in this PR is for several cameras that had been on the RD camera network, this was the North Escape cameras and the Perma ones.

#### Changelog

:cl:  Identification
rscadd: Adds a purple comfy chair, a cyborg statue with alternating names, and a bed to the RD office.  
bugfix: Several cameras were on the RD camera network, these have been fixed. Perma cameras have also been relabelled and put onto the prison network as they were lazily copy pasted. 
tweak: The orientation of objects in the RD office has been changed. The floor has also been recoloured to have purple outlines.
/:cl:


Lynameep:
This makes me want to recreate yogstation with dark tiles.

PR #5646, opened 2019-06-02
[MANUAL MIRROR] Turns trait accessors into defines, fixes some bugs

Altoids1:
## About The Pull Request
This turns trait accessors into defines, because someone had the bright idea to use them for everything so now they get called millions of times a minute. And proc calls aren't free.

Also provides a cleaner, less side-effect -laden way of accessing traits. If you want to access mind traits, you access mind traits instead of getting them as a side effect when accessing mob traits.

This also fixes a few bugs such as: quirks duplicating traits when transferring to mobs, presence of roundstart traits preventing the removal of non-roundstart traits, and some traits only giving a null as the trait source.

To be considered: how about standardizing sources as typepaths rather than magic strings? This would allow features like removing all reagent-granted quirks without iterating through all the types of /datum/reagent and collecting a list of IDs.

## Why It's Good For The Game
It's free performance.

## Changelog
🆑 Naksu
fix: Transferring quirks now properly removes the roundstart trait from the person losing the quirk.
fix: Roundstart traits no longer block the removal of other sources of that trait
code: Trait accessors are now defines rather than functions
/🆑


Altoids1:
This ends up being necessary for a *ton* of PRs, and is a major reason why we're getting failing mirrors.

ynot01:
the travis x of death

PR #5650, opened 2019-06-02
Voice analyzer starts on exclusive

SomeguyManperson:
:cl:  
tweak: Voice analyzer no longer starts on inclusive so it won't detonate on hearing anything automatically
/:cl:


Altoids1:
Did you actually have to swap around these things everywhere, or can you just change the default mode?

SomeguyManperson:
changing the default would look a bit sloppy but it'd work I guess

HomocidlePancakes:
I don’t get it

SomeguyManperson:
The analyzer currently starts on inclusive scanning, which means if the activating phrase shows up at all it'll activate. The base phrase is literally nothing meaning the thing will pulse if anything is said next to it if there's no code set. tldr if you don't touch the analyzer at all before putting it on a grenade it'll explode if it hears anything, this should stop that from being the default behavior

HomocidlePancakes:
Oh wow. This is pure ided

PR #5684, opened 2019-06-04
Syndicate speargun starts at the pressure required to embed

SomeguyManperson:
:cl:  
bugfix: Syndicate speargun will start able to embed instead of needing to be adjusted to do so  
/:cl:


MrSleeperagent:
Glad someone got to this my attempt always messed with the cannon and had pies embedding XD

PR #5872, opened 2019-06-14
Tweaks the tcomms machinery code a bit

Lostlarry:
prove that i understand how tcomms machinery works.
remove a plausible crash waiting to happen, "clean up" the code(it may be more mess), make trafic value on tcomms machines (it delays massages based on its value) not go negative

### DIFF
--- a/code/game/machinery/telecomms/telecomunications.dm
+++ b/code/game/machinery/telecomms/telecomunications.dm
@@ -48,7 +48,8 @@ GLOBAL_LIST_EMPTY(telecomms_list)
 		signal.data["slow"] = netlag
 
 	// Loop through all linked machines and send the signal or copy.
-	for(var/obj/machinery/telecomms/machine in links) // Yogs -- God bless typeless for-loops -- why do you need to use those here
+	for(var/m_typeless in links) // Yogs -- God bless typeless for-loops
+		var/obj/machinery/telecomms/machine = m_typeless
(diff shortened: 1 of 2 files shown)


code/game/machinery/telecomms/telecomunications.dm
```diff
@@ -48,8 +48,7 @@ GLOBAL_LIST_EMPTY(telecomms_list)
 		signal.data["slow"] = netlag
 
 	// Loop through all linked machines and send the signal or copy.
-	for(var/m_typeless in links) // Yogs -- God bless typeless for-loops
-		var/obj/machinery/telecomms/machine = m_typeless
+	for(var/obj/machinery/telecomms/machine in links) // Yogs -- God bless typeless for-loops -- why do you need to use those here
```
Amelia0010: they're faster.

Amelia0010 (changes requested):

code/game/machinery/telecomms/telecomunications.dm
Altoids1: They are indeed faster. Check CONTRIBUTING.md for details.

Lostlarry: they may be faster but what if non tcomms machine is linked? The way i understand it there is nothing to prevent that and would cause an error with having wrong type assignment.
I'm going test if it could actually crash btw.

PR #5890, opened 2019-06-16
Raises Crab-17 health even higher

ynot01:
people keep destroying it and i personally havent seen one person use the ID swipe function
:cl:  
tweak: Raised Crab-17 health even higher
/:cl:


fluffe9911:
not high enough dump et

Issue #5901, opened 2019-06-16
Free miner ripley not space proof


MegaEmpirical:
Due to a Ripley change from TG, https://github.com/tgstation/tgstation/pull/42828, standard Ripleys are not loner space worthy and need an upgrade. Also, the box of supplies provided for to build a Ripley provide an old unneeded resource (plasteel) and lack a new needed one (10 rods).

HomocidlePancakes:
Good thing you get a spacesuit eh?

Lostlarry:
its Ripley mk. I that one was never space proof

MegaEmpirical:
There are four people on that ship. Two bars suits for the miners, one Ripley for the engi, and I'm unsure of the captain gets one and so without the Ripley being space worthy that means someone else has to give up their suit.

Lostlarry, that is true but that's only true since MK I and II were added. This is a recent addition, there use to be only one Ripley that was space proof.

Hopekz:
This has been fixed in the past. Please close this issue.
Thank you!

MegaEmpirical:
Word is it's fixed. Closing.

PR #5916, opened 2019-06-17
Fixes several issues with the Free Miner ship.

nesisknesis:
Fixes issue #5901 

The RIPLEY can now be made space proof through a conversion kit available in the crate, and the plasteel has been replaced with 10 rods due to the recipe changing.

The propulsion engines were also facing the wrong direction, this has been rectified. As has the airlocks to space missing fans to prevent space wind.

As well as this, a cell charger was placed onto a table in the center of the ship, because they need to charge their batteries I guess,

#### Changelog

:cl:  Identification
rscadd: Adds a cell charger to the free miner ship. Adds fans to the airlocks that were missing them.
bugfix: The Free miner RIPLEY can actually be built and made space-proof. Engines also face the correct direction. 
/:cl:


MegaEmpirical:
Well done, thanks.

Sun-Soaked:
this broke the ripley wreck somehow, now it no longer gives any parts when welded.

PR #5969, opened 2019-06-19
toxins table are now only reinforced variety

Lostlarry:
one out of five tables was not reinforced that could drive people insane
#### Changelog

:cl:  
tweak: all tables in toxins are now reinforced
/:cl:


MegaEmpirical:
Mind reinforcing the table outside atmos too? It has the wrench and I'm tired of accidentally almost deconstructing it.

PR #5988, opened 2019-06-20
Revert "[MIRROR] Adds "empty" cloning"

SomeguyManperson:
Reverts yogstation13/Yogstation-TG#5907
I'll fucking bet this is what's causing the cloner to break


fluffe9911:
local tg breaks cloning more at 11

PR #6029, opened 2019-06-22
Allow airlock cycles to be connected directly

Lostlarry:
adds variables to allow for connecting airlock_cycle using relative coordinates
adds helper to allow to set it up more easily
(/obj/effect/mapping_helpers/airlock/cyclelink_helper_target, it has dirx and diry values)
negative x is left
positive x is right
negative y is down
positive y is up

not working
link multiple airlock in maint on yogbox and yogsmeta


MegaEmpirical:
This sounds like it could fix the one in the bottom right of engineering.

PR #6037, opened 2019-06-23
Implant Breaker cost have been upped to 40 points

fluffe9911:
🆑
tweak: Sec now stands a slightly better chance against gangs
/🆑

### DIFF
--- a/yogstation/code/game/gamemodes/gangs/gang_items.dm
+++ b/yogstation/code/game/gamemodes/gangs/gang_items.dm
@@ -268,7 +268,7 @@
 /datum/gang_item/equipment/implant_breaker
 	name = "Implant Breaker"
 	id = "implant_breaker"
-	cost = 20
+	cost = 50
 	item_path = /obj/item/implanter/gang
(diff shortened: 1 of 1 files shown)


HomocidlePancakes:
I’d rather buy me an uzi and kill officers now instead of converting.

GuppyLaxx:
add a rule against that
rules fix everything

gar4:
I mean if you have them in a position to convert you can just as easily kill them. As homocidle said, if i have to spend 50 points to convert i would just as well buy and uzi and give them two bursts point blank to the head and even save a few seconds.

fluffe9911:
Ok I lowered it to 40 but if you still decide to buy the 60 point uzi and just shoot them to death instead then thats your feken choice ye jerk

HomocidlePancakes:
10 points is barely any diffrence. Still makes more sense to kill. Also (insert reasons why mindslave implant cost was lowered here).

SomeguyManperson:
Fluffe you f o o l gangs doesn't have a high enough winrate to warrant being nerfed

Codeatmos:
just remove implant breaker tbh,or not at all because "MUH GRAPHIC CHART SAYS GANG BALANCED"

fluffe9911:
Eh screw it

PR #6047, opened 2019-06-23
Airlock scan

Lostlarry:
checking airlock access requirement without disassembling it. you have to attempt to open the door before it works properly.

I'll continue work on this and PR when i'm sure what i want to do with it
:cl:  
rscadd: adds airlock scanner for looking at airlock access requirements
/:cl:


MegaEmpirical:
Resolves this: https://github.com/yogstation13/Yogstation-TG/issues/4584

MegaEmpirical:
Will this work on buttons and windoors too?

Lostlarry:
no its only airlocks but it could be modified to work on about anything

MegaEmpirical:
Doooo it.

Por favor.

(Then rename it to Access Scanner)

Bibby0110:
>show_accses
:)

Lostlarry:
you sure about that? you want to accidentally click a wall and see that it needs no access?

MegaEmpirical:
I meant more like add windoor and button support, but if it also does walls and shit, I'm fine with that. Means you won't have to update it to work on other things at least (should we think of something else or something get added).

Lostlarry:
well the thing is req_access and req_one_access which are the variables that i read and output are inherent to obj class. that means machinery, mechs, items turfs(so even space), structures (grilles and such). the problem is most of these don't use this feature at all so i'd have to remove the static somehow which would be messy af.
that's assuming i find a way to get the airlock scan file past travis. i may end up throwing it to airlock painter.

AsV9:
Fuckin fatfingered that

ThatLing:
@AsV9 meanie

MegaEmpirical:
You monsters.

PR #6051, opened 2019-06-23
Fixies a floor cluwne edge case and slightly buffs it

fluffe9911:
🆑
fix: Floor cluwnes will no longer flip out if you transform into something else while its grabbing you
tweak: Floor cluwnes will now emp the area around it while its grabbing something
/🆑


SomeguyManperson:
Epic

PR #6061, opened 2019-06-24
removes additional arm from nukie

ynot01:
this pr exists in case it's still a problem even after the nerf (that was recently fixed)

🆑
remove: Additional Arm has been removed from nuclear operatives' uplink
/🆑


Lynameep:
Buff something else to take its place then.

ynot01:
this is getting too controversial for a pr that i made as a backup plan

PR #6142, opened 2019-06-29
Toxins Update

Readystorm546:
After a large amount of time in toxins and playing it on other station i noticed that toxins on boxstation (YogStation) was missing an injector out into space. As such i decided to add one after a rigorous attempts to this from the help of Theos and DotLyna  who mapmerged for me i can present it all to you. I have switch around the bomb locker its now located in the bomb testing room with the black and yellow stripes around it. i got rid of one bomb locker as there was no room for it and one person is all there usually is in toxins anyways if someone decides to use it. the heater and freezer were moved where the lockers used to be with new designed on the floor to match. the out put to space was put where the freezer and heater used to be. I think this change will benefit all toxins players including my self. it has not been tested but i can say with 90% certainty that it will work.


nesisknesis:
yeah it didn't mapmerge correctly evident by the additions and deletions counter

also, images pls

PR #6162, opened 2019-06-29
Toxins mapping Update

Readystorm546:
This update re-does the toxins map area on the default map in order to add a input pipe leading to the outside. 

This map has been tested and i can confirm that it works.

Toxins Change


Altoids1:
\>Greentexts in commit descriptions

HomocidlePancakes:
Begone hugbox.

PR #6199, opened 2019-07-02
Particle accelerator emag update

Readystorm546:
So pretty much you go up to the control console of the PA and emag it. This sets the strength to +4. If someone is fast enough they can revert it but if your not quick enough this will most likely cause a loose goose. I also tested it everything works fine and on top of that i made new sprites to identify that this is a +4 PA.
I wanted to add this because theos once told me you could do this but he lied and so i made it happen.


AsV9:
Idea: good
sprite: bad

Readystorm546:
there is a reason im not a spriter

nesisknesis:
I can look into making a better sprite

Readystorm546:
Sure then you can add it or something i dont know how it works

Nickvr628:
Maybe if you had the outline pulse different brightnesses, as if the extra energy was making it glow and flash, it would be better.

Would not be that hard to do either.

PR #6207, opened 2019-07-03
Adds Tricorders

yacabo:
Adds tricorders to the game which can be researched in experimental engineering

The tricorder is a late shift scanner which is basically a bunch of scanners meshed into one being:
*Gas analyzer's turf and pipe gas scanning
*Multitool's wire pulsing and grid power usage checking
*Slime scanning
*An extremely gutted medical scanner that can only estimate, though this can be turned into a regular scanner with a built in var

All of it being context sensitive so there's no need to switch modes

🆑
rscadd: Tricorders have been added! They act as a mix of Gas Analyzer, Multitool, and whatever else Yacabo thought to add!
/🆑


Readystorm546:
Wow

DLCabbose:
Noice

PR #6217, opened 2019-07-03
One small step for me, one giant leap for dm code

DLCabbose:
I have decided to become the latest and greatest thing to happen to Yogstation development efforts. I could use some more coding practice and it's about time I tried something different language-wise. This is just a sample of what's to follow in order to tease and excite the senses.


Amelia0010 (changes requested):
why?

DLCabbose:
Most programming I've worked with before is pseudo-code and network automation/scripting. Figure this is a way to exercise the parts of my brain that can handle more object oriented bits.

Plus I won't try to make imaginary lava burn you

nesisknesis:
As a member of the community, I do not like to be fucked.

PR #6222, opened 2019-07-03
Increases the price of Australian Slime Mutator from 10 TC to 25 TC.

nesisknesis:
Same points from #6097 

Large grief potential with spiders. As well as this, it is comparable with romerol, perhaps even more destructive seeing as spiders have an easier time of laying eggs. 

As a 10 TC item, it is a no-brainer if a syndi took this every time they were a scientist as it gives them a huge advantage when causing chaos and bringing the station to its knees. Raising it to 25 TC will make it a larger investment for how powerful spiders are and require another uplink. In my opinion, it is also an item intended for traitors with hijack rather than the common folk with steal the ce's magboots.

#### Changelog

:cl:  Identification 
tweak: The price of the Australian Slime Mutator is now 25 TC rather than 10 TC. 
/:cl:


HomocidlePancakes:
Thats dumb. It's a hijack item, restrict it to such.

ynot01:
i dislike this because it's actually controllable. xeno sentients are forced to listen to your every command, and broodmothers can set directives for their offspring. it's never "out of control" unless you have players blatantly breaking rules.

Amelia0010 (changes requested):
maybe restrict it to hijack and glorious death instead?

ghost:
@ynot01 It isn't controllable in the slightest.

Broodmothers can set directives for their offspring, which other broodmothers can remove without being noticed. Also, NPC spiders will attack people indiscriminately regardless.

ynot01:
wait so this is intended for hijack traitors but is being made 25TC? how are they supposed to get it if they only start with 20TC? this makes it so only nukies and IAA could purchase this

nesisknesis:
forgot this existed. its put to the same level of romerol because its practically just as destructive. also 'restrict to hijack and glorious death' is not done for any other item that ive seen so why would this have that restriction

Amelia0010:
https://github.com/yogstation13/Yogstation-TG/blob/08462f1cf4e053d0d8d9254d66723f49cc3699c3/yogstation/code/modules/uplink/uplink_item.dm#L45 It's done for his grace and romerol https://github.com/yogstation13/Yogstation-TG/blob/08462f1cf4e053d0d8d9254d66723f49cc3699c3/yogstation/code/modules/uplink/uplink_item.dm#L29

Amelia0010 (approved):
someone else can make a PR to reduce the price if they're unhappy with this

PR #6238, opened 2019-07-04
Flimsy Cryptographic Sequencer

ynot01:
not having to use 8TC on something you would only use once

WARNING: IF YOU SMACK IT ON A WALL OR SOME SHIT THAT DOESNT GET EMAGGED ITLL STILL WEAR IT DOWN SO DONT BE RECKLESS

#### Changelog

:cl:  
rscadd: Adds Flimsy Cryptographic Sequencer to the syndicate uplink, a two-use emag that costs 3TC
/:cl:

### DIFF
--- a/yogstation/code/game/objects/items/cards_ids.dm
+++ b/yogstation/code/game/objects/items/cards_ids.dm
@@ -66,35 +66,6 @@
 /obj/item/card/id/captains_spare
 	icon_state = "id_gold"
 
-/obj/item/card/limitedemag
-	name = "flimsy cryptographic sequencer"
-	desc = "It's a card with a magnetic strip attached to some circuitry. It looks pretty fragile."
-	icon_state = "emag"
-	item_state = "card-id"
-	lefthand_file = 'icons/mob/inhands/equipment/idcards_lefthand.dmi'
-	righthand_file = 'icons/mob/inhands/equipment/idcards_righthand.dmi'
-	item_flags = NO_MAT_REDEMPTION | NOBLUDGEON
-	var/u = 2
-
-/obj/item/card/limitedemag/attack()
-	return
-
-/obj/item/card/limitedemag/afterattack(atom/target, mob/user, proximity)
-	. = ..()
-	var/atom/A = target
-	log_combat(user, A, "attempted to emag")
-	if (u == 0)
-		playsound(src.loc, "sparks", 50, 1)
-		return
-	if (u == 1)
-		u--
-		desc = "It's a card with a magnetic strip attached to some circuitry. It's broken."
-		A.emag_act(user)
-	if (u == 2)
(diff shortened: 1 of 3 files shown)


Nickvr628:
I like it

yogstation/code/game/objects/items/cards_ids.dm
```diff
@@ -66,6 +66,35 @@
 /obj/item/card/id/captains_spare
 	icon_state = "id_gold"
 
+/obj/item/card/limitedemag
+	name = "flimsy cryptographic sequencer"
```
Altoids1, line 70: Making this not parented to the emag type means changes in emags won't be reflected over to here automatically. Additionally, doing it this way makes it harder to have emags be capable of emagging limited emags, half-breaking https://github.com/yogstation13/Yogstation-TG/pull/1662.

Make this a subtype of ``/obj/item/card/emag``.

Altoids1 (changes requested):

Oakboscage:
See most of the time opening two doors is enough, so this is basically just an emag buff. If you want to magically open doors just spend the 8 TC, it's not expensive for what it is, even if you just need two doors.

yogstation/code/game/objects/items/cards_ids.dm
ynot01: the reason i de-parented it is because it conflicted with the emag's afterattack. i try my best to not fuck with tg's code (anymore) so my best solution was to make a new card

ynot01:
> See most of the time opening two doors is enough, so this is basically just an emag buff. If you want to magically open doors just spend the 8 TC, it's not expensive for what it is, even if you just need two doors.

actually if altoids goes through with his emag rework this won't even open airlocks

on top of that I disagree with "two doors being enough" since the emag has dozens of incredibly useful functionalities, this would add another option for someone who _doesn't need_ all of those functionalities.

yogstation/code/game/objects/items/cards_ids.dm
Putnam3145: Could just change the way it does the breaking so you could handle breaking logic after all the other afterattack logic.

ynot01: done

Altoids1 (approved):

MegaEmpirical:
Smacking a wall uses a use? Are you a mad man?

AsV9:
I'll pass, thanks.

AsV9:
Closed based on input in this PR.

Marmio64:
3 TC for two uses? HAHAHAHAHAHAHAHAHHAAHHA
no

how about 4 TC for a 7 use emag that doesn't use charges when it hits a wall or something.

PR #6270, opened 2019-07-06
Remaps Xenobiology and very minor tweaks to toxins.

nesisknesis:
No AI cores were moved in the making of this PR. I think.

image

First off, an airlock system was added, along with reinforced windows because working with aliens is dangerous. Biohazard blast doors wered also added to the front of xenobiology which are connected to the RD's lockdown button.

The test chamber now also has an igniter setup and gas injector so that the BZ canister can be properly used. The igniter will allow for the test chamber to be flooded with plasma and set alight to purge any would be xeno test subjects.

In a future PR, I may make the test chamber be randomised to have different leftovers of experiments.

image
image

Two references on one piece of paper.

image
This was heavily inspired by /TG/'s KiloStation (https://drive.google.com/file/d/1IIz2OsnyhQCCXxEWUydWVP3mPyOg5fFU/view)

Ever since the slime management console was added, the doors to the slime chambers have been redundant. However, if one wishes to interact with slimes they can just simply shunt them into the open lab.

All machinery relating to slime breeding has been moved to this room as well.

As for the minor tweaks, the light fixtures in toxins storage was rotated to be centered on the North and South wall rather than offset on the east and west so that the light is evenly shone.

As for the lab, a lattice was added under the air injector, and the pump set to be off by default.

image

Xenobiology exterior.
image

#### Changelog

:cl:  Identification
rscadd: Xenobiology has been remapped to be more modern.
tweak: Tweaks to the lighting in toxins storage, an extra lattice to the space between toxins, the port to space pump is now off by default.
experimental: The test chamber now has leftovers from xenomorph experimentation. This may be removed or made to be randomised with different types of residue. 
/:cl:


Nickvr628:
inb4 this is a buff to cult as it fortifies their main base

Nickvr628:
Also I really dislike the lack of chutes for the slime pens, those are useful dammit

nesisknesis:
They were redundant because of the console and it's more room for slimes.

Nickvr628:
the slime console is a crutch and only unrobust betacrew use it. Back in my day we wrangled the slimes by hand and WE ENJOYED IT! Now any chucklefuck can sit at the console and pop slimes everywhere removing all the risk.

ThatLing:
why?

nesisknesis:
It's dead, Jim.

Readystorm546:
Uh if the pump in toxins is off by default there is no way to turn it on

Readystorm546:
disregard my above comment

HomocidlePancakes:
Nuuu, don't remove the Rtables. They look nice and xenobio already had them prior to this. Also, even if you are removing the chutes, keep the windoors in to pens to maintain them (broken lights, modification etc.). Also people were using these to isolate slimes.

nesisknesis:
Rtables were changed to regular ones so that the scientist could easily modify the pens without a welder. The glass on the image looks clean and adding windoors would also require the machinery to be moved about which there isn't much room for.

Ill get an image of the edited version tomorrow,

ynot01:
what? what happened to the glass tables that monkeys and slimes would break every round? 0/10

alexkar598:
@Ihonkification Why is this change good for the game?

nesisknesis:
Because it uses the un-used space in xenobio. With how the consoles are positioned, the cells may as well be on another z-level.

nesisknesis:
i lied something broke

PR #6278, opened 2019-07-07
The Vortex Tailsman no longer works in the King Goats Arena (and other noteleport areas)

fluffe9911:
🆑
tweak: The Vortex Tailsman no longer works in the King Goats Arena (and other noteleport areas)
/🆑

### DIFF
--- a/code/modules/mining/equipment/kinetic_crusher.dm
+++ b/code/modules/mining/equipment/kinetic_crusher.dm
@@ -395,6 +395,11 @@
 	return "mark detonation to create a barrier you can pass"
 
 /obj/item/crusher_trophy/vortex_talisman/on_mark_detonation(mob/living/target, mob/living/user)
+	var/turf/current_location = get_turf(user)//yogs added a current location check that was totally ripped from the hand tele code honk
+	var/area/current_area = current_location.loc //yogs more location check stuff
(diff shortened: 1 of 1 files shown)


HomocidlePancakes:
Bruh,How about you instead fix it having a permanent speed boost and it's ability to move faster then it's supposed to while moving diagonally.

HomocidlePancakes:
How about

HomocidlePancakes:
You instead fix

HomocidlePancakes:
It having permanent speed

HomocidlePancakes:
And it's ability to move faster then it's supposed to diagonally

alexkar598:
@HomocidlePancakes Please refrain from spamming comments and instead write everything you want to say in a single comment,this isnt Discord.

alexkar598:
@fluffe9911 Why is this change good for the game?

fluffe9911:
Because what the vortex tailsman does is create a wall that mobs cant path around cause they are bloody stupid leading to the king goat a melee focused megafauna getting memed real easy

alexkar598:
Perhaps a rebalance of the item itself or a rework to make mob pathfinding proc faster would be better than making an item not work because "breaks da balance".

fluffe9911:
But it only not works in areas that this item should not be working in the first place

PR #6279, opened 2019-07-08
Strange object icon and icon use update

Readystorm546:
So there are a bunch of unused strange object icons that should be in use and with this i put them in use. They were there from the start but never used. Also i changed 4 of the icons so they dont look identical but do in-fact fit the theme of the red blue and green color usage.

image
 
image

image

image


boodaliboo:
you are ascending

PR #6287, opened 2019-07-09
Fixes weird stripe of Vacant Office near the Vault on Yogbox

Readystorm546:
Altoids needs help doing some things so im here to do it. Fixes a map error that he pointed out.


Altoids1:
image

PR #6310, opened 2019-07-11
New toxins point system

Readystorm546:
So i heard you like a rewarding toxins system so i made a new equation as such.

normal max cap
 
This is the normal max cap in the new system ^

Big bomb ive made in toxins

This is the biggest bomb ive ever made in toxins alone in the new point system ^

### DIFF
--- a/code/__DEFINES/research.dm
+++ b/code/__DEFINES/research.dm
@@ -72,4 +72,5 @@
 	TECHWEB_POINT_TYPE_GENERIC = "General Research"\
 	)
 
-#define TECHWEB_BOMB_POINTCAP		50000 //Adjust as needed; Stops toxins from nullifying RND progression mechanics. Current Value Cap Radius: 100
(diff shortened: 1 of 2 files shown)


Alek2ander:
Tank bomb numbers are always X/4, X/2, X. Just use one of the sizes.

Readystorm546:
this is actually not true. Bomb sizes can carry if there close enough and yes i know its a small difference but they can still do it.

Alek2ander:
https://github.com/yogstation13/Yogstation-TG/blob/afa68548874b02b3a383abb937f08c141e4dab2a/code/game/objects/items/tanks/tanks.dm#L255
The difference is just in rounding and is basically irrelevant. The formula you propose is just 10*light_radius written in a complicated way.

Readystorm546:
well it would be if it was not all divided individually. Not to mention i explained myself in Dev-public. Also "basically irrelevant" is still relevant because that one number can make all the difference the higher you go.

PR #6314, opened 2019-07-11
PA computer sprite fix

yacabo:
Fixes the PA computer sprite to update to its proper form with the sprites that were already in the code, also makes the PA's power 2 console sprite into an actual sprite rather than a half made one

Power off:
PAoff

Power 0:
PA0

Power 1:
pa1

Power 2:
PA2

Power 3 (Hacked):
PA3

Power 4 (Emag'd):
PA4


code/modules/power/singularity/particle_accelerator/particle_control.dm
```diff
@@ -18,6 +18,10 @@
 	var/strength = 0
 	var/powered = FALSE
 	mouse_opacity = MOUSE_OPACITY_OPAQUE
+	//yogs start- sprite fix
+	var/reference = null
+	reference = "control_box"
```
alexkar598: Merge these 4 lines into one and name it something meaningful like base_sprite and also move it to the yogstation/.... varient of the file

yogstation/code/modules/power/singularity/particle_accelerator/particle_control.dm
```diff
@@ -10,6 +10,8 @@
 
     strength = 4 // Set the new strength to lvl 4
     strength_change() // Update the emitter
+    update_icon()
 
     if(!active)
-        toggle_power()
\ No newline at end of file
+        toggle_power()
+        update_icon()
```
alexkar598: Just add it at the end after the if statement once instead of adding it twice.

alexkar598 (changes requested):

code/modules/power/singularity/particle_accelerator/particle_control.dm
yacabo: Done

yogstation/code/modules/power/singularity/particle_accelerator/particle_control.dm
yacabo: and done

PR #6329, opened 2019-07-12
Allows prisoners to access perma vendors with no ID

Ktlwjec:
Removes economy from the vendors in perma. I figure this is the best (easiest) way of doing this.

:cl:  
bugfix: Perma prisoners now have access to the perma vendors.
/:cl:


HomocidlePancakes:
Remove nuka cola from perma vendors

Ktlwjec:
It's a premium item so it shouldn't be an issue. If they somehow get a coin then good for them.

Alek2ander:
>Prison vending machines not having everything gouged to 3x price.

Amelia0010:
sad

PR #6343, opened 2019-07-14
[Admin] Flavour Text for ghost roles now shows in TP

alexkar598:
Made with ❤️ for Qe

img


Mr-Space:
Speedmerge when

alexkar598:
Changed how it works,now uses the memory and shows like this on TP:

PR #6346, opened 2019-07-14
[admin] Adds logging to to_chat()

alexkar598:
Made with ❤️ for Qe

Screenshot of what it looks like is posted in IA admin chat for those who care since i dont want to risk posting stuff i dont wanna lol

tochat.log is unreadable without filtering

this includes viewing the chat for a single mob/client using PP


Amelia0010:
jesus christ no, the fucking logfiles will be obscenely large

code/modules/goonchat/browserOutput.dm
```diff
@@ -233,6 +233,8 @@ GLOBAL_DATUM_INIT(iconCache, /savefile, new("tmp/iconCache.sav")) //Cache of ico
 			if (!C)
 				continue
 
+			C.mob.log_message(message, LOG_TOCHAT)
```
Amelia0010, line 236: add an argument to the to_chat proc, to choose whether to log the message or not, defaulting to not, please.

Logging to_chat should be opt-in on a message to message basis.

Amelia0010 (changes requested):

code/modules/goonchat/browserOutput.dm
Amelia0010: I just had a think;

Consider making the extra argument be the "name" of the log, as in:

```
/proc/to_chat(target, message, handle_whitespace=TRUE, logname)

    [...]

   if(logname)
         C.mob.log_message(logname + ": " + message, LOG_TOCHAT)
```

That way we can add `flavourtext` as an argument to to_chat, and we can then search for `flavourtext` in the log if we need to access it

alexkar598:
The manually setting a flag to log it,defaulting to not is pointless as you can just add a logging thing there so imma close this

Amelia0010:
:(

PR #6350, opened 2019-07-14
Xeno queens delay shuttle

SomeguyManperson:
Xeno delay is currently at 30 minutes, can be adjusted if necessary
:cl:  Xeos
tweak: Xeno queens will delay the escape shuttle for a while unless they are killed
tweak: the shuttle will get auto called at the end of the timer if the queen is alive and it hasn't been called already to prevent super xeno extended rounds
/:cl:


HomocidlePancakes:
Shouldn’t be called just based on timer TBH. Take in crew:xeno ratio in to account to and/or amount of living crew

SomeguyManperson:
If a xeno queen is alive for 15 minutes there is very likely a problem

HomocidlePancakes:
Or it’s being a massive wuss/incopetent/on the run and the crew is merking benos harder then 1812.

alexkar598 (changes requested):
Use timers instead of a bullshit discount for loop.

alexkar598:
Im against the fact it auto resolves after 15 minutes,should only resolve when the queen dies.

Might be also worth giving the aliens a win condition.

SomeguyManperson:
The auto resolve was suggested since otherwise a xeno round can go on quite long and if the admins don't want to resolve it/there are none it won't end naturally

PR #6352, opened 2019-07-15
adds an item slot flag for suit storage

SomeguyManperson:
this probably(tm) won't break anything, also makes antimagic items in suit storage apply antimagic instead of just not doing that

:cl:  
tweak: coders can now check the suit storage slot for items 
/:cl:


Amelia0010 (approved):
I can't believe this wasn't a thing already

Issue #2991, opened 2018-10-12
Atmos cameras, pipe system


00aquir:
- Yog Station, new atmos layout now in round 19055, most atmos cameras don't show up on camera console (they are active for AI).

- Air is not mixed:

![ss13-atmosmixer](https://user-images.githubusercontent.com/16806055/46860957-c3d3be80-ce11-11e8-8c4d-2a6e95ac12cc.png)

ThatLing:
#2983

thehighwayman:
someone done fucked up

Ktlwjec:
right back off you I fixed this already

00aquir:
Dunno if this is fixed already, so reporting this also:
- "Incinerator access console/interior" and the right-side "incinerator airlock control" don't work:

![ss13-atmos-turbine](https://user-images.githubusercontent.com/16806055/46880239-69a22000-ce48-11e8-859c-5bde40946165.png)

Ktlwjec:
Someone mentioned that but didn't explain it at all clearly, I'll fix that today sometime.

00aquir:
Another one:

- Incinerator air injector should automatically push the gas out, but it doesn't do that.

![ss13-atmosinjector](https://user-images.githubusercontent.com/16806055/46882204-1df27500-ce4e-11e8-907c-5de187291e38.png)

Ktlwjec:
Fixed that already, and the other injector.

00aquir:
Actually, since the air injector has a transfer rate that can be changed and it's not on max.. there shouldn't be injector at all, because (there is already a pump there, and) atmos techs would have to go into space every round for setting it (for the turbine).

Ktlwjec:
I got turbine working with the fix without going to space.

00aquir:
another one:
- seemingly trash keep coming up from atmos bin (or rather, the pipes) in different rounds

![ss13-atmosbin](https://user-images.githubusercontent.com/16806055/46884989-e63bfb00-ce56-11e8-8767-eba7a84b91eb.png)

00aquir:
@ThatLing, it seems that the left-side "incinerator airlock control" button is not working either (as well as the right one does now). Also **after** pushing the right side button (left door bolted, right one unbolted), the "Incinerator Access Console" doesn't do anything..

Ktlwjec:
It worked on my test reeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee

PR #6366, opened 2019-07-16
Removes NTSL randseed

ThatLing:
https://github.com/yogstation13/yogstation/pull/2991


1fbff5f83b23d39d38b1dfcb4cac8d9b:
AAAAAAAAAAAAA UNUSED VARIABLE WARNING

Altoids1:
:(

alexkar598 (changes requested):
Commented code is bad,get rid of it

ThatLing:
@alexkar598 says who?

Amelia0010:
Getting rid of the function would break compatibility with older scripts, and I don't really see why we need to do that?

PR #6380, opened 2019-07-18
Gives Ocular Wardens a beam

yacabo:
Give the Clock Cultist's Ocular Warden a beam sprite to shoot people with that looks cool and makes spessmen pay more attention

ieyFTTId2M


yacabo:
wew

PR #6382, opened 2019-07-18
Fixes missing sprites for clown eggs!

sam0oneau:
Adds missing sprites for clown eggs, as currently code looks to find those food items in yog folder not TG folder.

Swiss wanted this speed merged so here it is!!


Altoids1:
So, what, are we maintaining two copies of this icon file now?

PR #6383, opened 2019-07-18
Enforces DOS(crlf) endings for certain file types

alexkar598:
.dmm .dm .md .txt .html now are obligated to use DOS(crlf),also updated the gitattribute file so it enforces the line endings automatically


ThatLing:
```
Yes, we know that the files have a tonne of mixed Windows and Linux line endings. Attempts to fix this have been met with less than stellar success, and as such we have decided to give up caring until there comes a time when it matters.

Therefore, EOF settings of main repo are forbidden territory one must avoid wandering into, at risk of losing body and/or mind to the Git gods.
```

PR #6385, opened 2019-07-19
Buffs Shadowling Thralls

fluffe9911:
🆑
tweak: Shadowling thralls are now discovered by looking at there eyes instead of scanning there body temp
/🆑

The whole reason the whole body temp scan thing became a thing in the first place was to give thralls a better chance of not being discovered but honestly I think it did the opposite and made finding thralls a ton easier so now its semi reverted but instead of a mask its eyes so they can hide it a bit better that is all

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -297,15 +297,7 @@ GENE SCANNER
 			mutant = TRUE
 
 		to_chat(user, "<span class='info'>Species: [S.name][mutant ? "-derived mutant" : ""]</span>")
-
-	//Health analyzers warn you about very cold people, like thralls or victims of shadowlings. //yogs start
-	if(M.mind?.has_antag_datum(ANTAG_DATUM_THRALL) || M.bodytemperature < (11 + T0C))
-		//Thralls get a fake temperature so they always read as too cold.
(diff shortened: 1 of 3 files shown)


Codeatmos:
I don`t think this really improves it, now it's even more obvious because sec and anyone competent/metagaming will just check the eyes at random, at least before you needed a medical scanner, of course, this could be tweaked to be like loud cult maybe, as you go on its more and more obvious

ynot01:
this could lead to some creative ways of hiding your eyes and if slings aren't revealed you can't get turbofucked by some doctor who wants to patch you up

>competent/metagaming

is there a difference anymore

Laus4Deus:
I think indeed was harder to discover thralls since they could cover their features somehow, since it's not exactly _weird_ to see an assitant with gasm mask. Now the only way to prevent getting scanned is not being scanned.

drderp3635:
might i suggest making medical scanners unable to pick up body temp if the person is wearing thick clothing, or a hardsuit?

Bibby0110:
May I suggest not making thralls able to be checked with a widespread item?

Bibby0110:
You removed the temp scanning part of the tricorder?

fluffe9911:
no

PR #6386, opened 2019-07-19
New throwing star

Laus4Deus:
:cl:  Matias Sanhueza
rscadd: Replaced the throwing star sprite   
/:cl:

The current throwing star is an unappealing small sprite that's very hard to notice and click propperly since its colour is so similar to the station's grey tiles. The proposed one is larger and has a light palette as to make it stand out when it's on the floor.

Edit: After consulting with a experienced spriter, I increased the contrast of the item within itself and the floor and lengthen the blades, resulting in a **Way cooler** star.


Partheo:
👍

Firewolf34:
Here's what the star looks like, if anyone else was wondering: 

687474703a2f2f69636f6e64696666626f742e64657874726173706163652e6e65743a383038302f46696c65732f3132303936363333362f3137323537323536312f312f61667465722e706e67


Looks mean af

PR #6388, opened 2019-07-19
new map job change system

Amelia0010:

This may be a downgrade with regards to clunkiness and modularity, but I honestly feel that this is a better system, partly because it doesn't rely on defines.

I moved SSmapping to be above SSjobs (which may honestly be the real fix here), to make sure that the SSmapping config loads before SSjobs. This PR should be testmerged before a proper merge, as I'm not sure whether that breaks anything.

I need to port over all the other map changes to this system (don't worry about the warden change, that was just for testing purposes), and then this PR should be ready.

#### Changelog

:cl:  
experimental: The map job change system now uses signals instead of defines
experimental: SSmapping and SSjobs have exchanged init order
/:cl:

If you want to change a job on a specific map with this system, you will want to go onto that job datum
and add said map's name to the changed_maps list, like so:
```	
changed_maps = list("OmegaStation")
```
Then, you're going to want to make a proc called "OmegaStationChanges" on the job, which will be the one
actually making the changes, like so:
```
/datum/job/miner/proc/OmegaStationChanges()
```
If you want to remove the job from said map, you will return TRUE in the proc, otherwise you can make
whatever changes to the job datum you need to make. For example, say we want to make it so 2 wardens spawn
on OmegaStation, we'd do the following:
```
/datum/job/warden
	changed_maps = list("OmegaStation")
	
/datum/job/warden/proc/OmegaStationChanges()
	total_positions = 2
	spawn_positions = 2
```


_maps/yogstation.json
```diff
@@ -1,5 +1,5 @@
 {
-    "map_name": "Yog Station",
+    "map_name": "YogStation",
```
Amelia0010, line 2: Oh and I removed this space because I can't have spaces in proc paths

alexkar598: Please add a Travis test to ensure mapsnames don't have space in them

MegaEmpirical:
Could this resolve the issue with the job selection menu sometimes reverting to the non-yogs version without our specific jobs?

Amelia0010:
that's whatit's supposed to do :)

PR #6391, opened 2019-07-19
Updates to minsky bridge

Ktlwjec:
Bridge time now. I've rearranged the rooms. There was no point having a vacant room that only command staff can access - may as well just use the meeting room at that point. HOP office was way too cramped too. I've also added one of KMC's genius ticket machines.

Before:
image

After:
image

:cl: 
tweak: Minsky bridge has been rearranged slightly.
/:cl:


Nickvr628:
The minski bridge is truly bigger than some entire stations

PR #6397, opened 2019-07-20
Plasmaman Envirosuit resprite and tweaks.

nesisknesis:
image
image

These are without what a crewmember would typically wear, IE. A labcoat or backpack.

The helmets are also transparent, so if a non-plasmaman wears it, it will be apparent they are not a plasmaman. However, if the light is toggled, the helmet becomes bright and opaque.

The curator has the old orange envirosuit as a replacement for the 'prototype envirosuit' they originally had.

The old sprites were a bit all over the place and inconsistent in the side sprite department.

The clown and mime retain their original envirosuits and yogjobs are missing custom envirosuits. I may PR them in future when I'm up for it or someone else may take the task. 

I think everything works.
#### Changelog

:cl:  
rscadd: The QM now has a envirosuit.    
wip: Yogjobs, clown, and mime are missing custom or updated envirosuits. 
imageadd: Envirosuits have been resprited.
imagedel: Most of the old envirosuits.
/:cl:


code/modules/clothing/outfits/plasmaman.dm
```diff
@@ -1,19 +1,19 @@
 /datum/outfit/plasmaman/bar
 	name = "Bartender Plasmaman"
 
-	head = /obj/item/clothing/head/helmet/space/plasmaman/white
+	head = /obj/item/clothing/head/helmet/space/plasmaman
```
Amelia0010: All of the `head = /obj/item/clothing/head/helmet/space/plasmaman`'s can be removed, since it's inherited from `/datum/outfit/plasmaman`
https://github.com/yogstation13/Yogstation-TG/blob/bd36ec2b9a5b9b2fddc17649ca697960f4f999f3/code/modules/clothing/outfits/standard.dm#L366

Amelia0010 (changes requested):
after this, it should be fine

deadmomo69:
I think the envirosuits should be big and chunkier like the old ones, in my personal opinion, these just look like normal jumpsuits almost to me. I like the transparent helms though.

Amelia0010 (approved):

MegaEmpirical:
If the suit is too big than jackets and shit get overlapped with the suit sprites and re-doing all of those sprites or creating duplicate sprites is just madness.

Nickvr628:
A few of these, like the detective one, look too close to normal human clothes IMO. I like the transparent helmet but a few of these don't look like plasmaman gasbags.

PR #6417, opened 2019-07-22
allows silicons to add crimes remotely

Lostlarry:
it felt stupid to be able to add crimes but only while next to the console as silicon. Also allows people with TK/longarm mutations to add them at long range

:cl:  
tweak: Silicons can operate the sec logs console from distance
/:cl:


code/game/machinery/computer/security.dm
```diff
@@ -838,7 +838,7 @@ What a mess.*/
 /obj/machinery/computer/secure_data/proc/canUseSecurityRecordsConsole(mob/user, message1 = 0, record1, record2)
 	if(user)
 		if(authenticated)
-			if(user.canUseTopic(src, BE_CLOSE))
+			if(issilicon(user)||in_range(user, src))		//yogs
```
alexkar598:
```suggestion
			if(user.canUseTopic(src, !issilicon(user)))
```

alexkar598 (changes requested):

oklews:
This is a godsend QoL update, it is one of the few reasons I despise playing silicons

PR #6437, opened 2019-07-23
Makes Xenoes & Monkeys hold more/less blood in their bodies than humans

Altoids1:
image

## Overview
Right now, every single ``/mob/living`` in the game that operates bloodcode has exactly 560 cL of blood floating around in its body by default. This includes monkeys, despite being half the size of humans, as well as xenoes, despite the fact they are massive & make a whole business out of having acid blood.

This PR changes that, and decreases the blood volume of monkeys by 40% and increases xeno blood volumes by 25%.

## So what does this accomplish, practically?
It's a slight chef nerf, since now their manhandling of monkeys will now produce slightly less blood.
It's a slight buff to detective, since now it's a bit easier to tell apart the blood trails of monkeys from those of human corpses.

The Xeno thing will also make their acid blood a bit more relevant when killing them.

## Coder Warnings
Doing this required doing a (minor) refactor of blood volume stuff. Now the blood volume defines are functions, that take in a ``/mob/living`` instance and return the blood volume associated with it being OKAY, BAD, SAFE, whatever. There's also a new function, ``/mob/living/proc/get_blood_state()`` that returns the current blood situation, which is useful for switch statements involving blood, as used in a switch statement edited in this PR.

#### Changelog

:cl:  Altoids
rscadd: Living creatures no longer have always exactly the blood volume of a human.
tweak: Monkeys now have 40% less blood.
tweak: Xenomorphs now have 25% more (acidic!) blood.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -295,15 +295,17 @@
 	overdose_threshold = 60
 	taste_description = "sweetness and salt"
 	var/last_added = 0
-	var/maximum_reachable = BLOOD_VOLUME_GENERIC - 10	//So that normal blood regeneration can continue with salglu active
+	var/max_blood_decrement = 10 //Yogs -- How much less than BLOOD_VOLUME_NORMAL(M) is the point where salglu stops refilling blood?
+	//^ Used so that normal blood regeneration can continue with salglu active
 
 /datum/reagent/medicine/salglu_solution/on_mob_life(mob/living/carbon/M)
 	if(last_added)
 		M.blood_volume -= last_added
 		last_added = 0
-	if(M.blood_volume < maximum_reachable)	//Can only up to double your effective blood level.
+	var/max_blood = BLOOD_VOLUME_NORMAL(M) - max_blood_decrement // The highest blood volume that salglu will work at.
+	if(M.blood_volume < max_blood)
 		var/amount_to_add = min(M.blood_volume, volume*5)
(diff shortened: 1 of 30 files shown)


code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -295,7 +295,7 @@
 	overdose_threshold = 60
 	taste_description = "sweetness and salt"
 	var/last_added = 0
-	var/maximum_reachable = BLOOD_VOLUME_NORMAL - 10	//So that normal blood regeneration can continue with salglu active
+	var/maximum_reachable = BLOOD_VOLUME_GENERIC - 10	//So that normal blood regeneration can continue with salglu active
```
alexkar598: This is going to need to use the blood volume normal macro somehow

yogstation/code/modules/mob/living/living.dm
```diff
@@ -0,0 +1,17 @@
+/mob/living/proc/get_blood_state() // Returns how our good or bad our blood situation is, using the blood enums in Yogstation-TG\code\__DEFINES\mobs.dm
+	if(!initial(blood_volume))
+		return BLOOD_SAFE // I... guess we're fine?
+	
+	switch(blood_volume / initial(blood_volume))
```
alexkar598: Why devide by the initial value?

alexkar598 (changes requested):

yogstation/code/modules/mob/living/living.dm
Altoids1: Because originally this switch was changed to use the _MULTI stuff. I'll go ahead & fix it.

alexkar598 (approved):

PR #6446, opened 2019-07-24
QOL

alexkar598:


ynot01:
if i read the code correctly this tells the admins the IP of someone connecting for the first time

PR #6469, opened 2019-07-26
Xeno Winconditions

Amelia0010:
Which is literally just hijack the shuttle. Better than nothing?

🆑 
rscadd: Xenos now have a wincondition to hijack the shuttle
/:cl:

I'll make a PR to let :b:enos call the shuttle by interacting with the comms console somehow, later

### DIFF
--- a/code/modules/shuttle/emergency.dm
+++ b/code/modules/shuttle/emergency.dm
@@ -292,6 +292,23 @@
 
 	return has_people && hijacker_present
 
+/obj/docking_port/mobile/emergency/proc/is_hijacked_by_xenos()
+	var/has_xenos = FALSE
+	for(var/mob/living/player in GLOB.alive_mob_list)
+		if(issilicon(player)) //Borgs are technically dead anyways
+			continue
+		if(isanimal(player)) //animals don't count
+			continue
+		if(isbrain(player)) //also technically dead
+			continue
+		if(shuttle_areas[get_area(player)])
+			//Non-xeno present. Can't hijack.
+			if(!istype(player, /mob/living/carbon/alien))
+				return FALSE
+			has_xenos = TRUE
+
+	return has_xenos
+
 /obj/docking_port/mobile/emergency/proc/ShuttleDBStuff()
 	set waitfor = FALSE
 	if(!SSdbcore.Connect())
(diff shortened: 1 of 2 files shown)


ynot01:
this was my idea

ynot01:
also shuttle auto calls if AI and both comms consoles are destroyed

Amelia0010:
>20 minute shuttle timer after winning

smh no thanks

OrderoftheDarkOats:
> this was my idea

Congratulations

Amelia0010:
oh yeah btw this is what it looks like 
image

fluffe9911:
Xeno gamemode when

Bibby0110:
#6350 autocalls the shuttle, no?
EDIT: For Nichs future plans, this PR looks good

Xoxeyos:
Would PC lavaland Xeno's also have this objective? Or would the one's spawned in by random events only have this objective?

OrderoftheDarkOats:
I'd think the objective should/would only count if benos get onboard.

SomeguyManperson:
the shuttle delay will only call the shuttle after 30 minutes, adding a way to allow xenos to end it earlier would be a good idea just remember to remove the hostile environment when they do that

SomeguyManperson:
> Would PC lavaland Xeno's also have this objective? Or would the one's spawned in by random events only have this objective?

Lavaland xenos don't exist any more

code/modules/shuttle/emergency.dm
```diff
@@ -292,6 +292,23 @@
 
 	return has_people && hijacker_present
 
+/obj/docking_port/mobile/emergency/proc/is_hijacked_by_xenos()
+	var/has_xenos = FALSE
+	for(var/mob/living/player in GLOB.alive_mob_list)
+		if(issilicon(player)) //Borgs are technically dead anyways
```
Altoids1, line 298: Does normal hijack do this silicon-exclusion thing?

ThatLing:
What about staff of change xenos?

code/modules/shuttle/emergency.dm
Amelia0010: consider scrolling about 30 lines upwards in this file :)

PR #6471, opened 2019-07-26
Polishes the vault safe some, improving code & usability

Altoids1:
On pause for a bit while I do the IAA stuff, but, here it is for now.

## Coder Warnings
I added some new defines to handle modulo arithmetic, called ``MODULO_INC()`` and ``MODULO_DEC()``. They should help for when you have a spinning dial that goes from 0 to ``R`` and back again, as is the case for the safe dials & the inner rotaries within it.

#### Changelog
:cl:  Altoids
bugfix: Blind people will now be able to hear the noises coming out of a safe.
bugfix: Deaf people will now no longer be able to crack safes through sound alone.
imageadd: The Safe's popup menu now actually features an image of a dial!
/:cl:

### DIFF
--- a/code/game/objects/structures/safe.dm
+++ b/code/game/objects/structures/safe.dm
@@ -3,6 +3,9 @@ CONTAINS:
 SAFES
 FLOOR SAFES
 */
+#define DIAL_RANGE 72 // How many options there are on the dial, including 0.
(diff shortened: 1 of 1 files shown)


Firewolf34:
very nice!

MegaEmpirical:
I want.

Firewolf34:
The merge conflicts for this are from the line-ending change PR.

alexkar598:
Great PR,love it,come back with it when it's not stale

PR #6478, opened 2019-07-27
Splits Malf AI and traitor AI into two antag types, now has its own mode

Amelia0010:

spiritual revert of https://github.com/tgstation/tgstation/pull/13759, although it's not a 1:1

Basically:
Traitor AIs, be it in traitor rounds or IAA rounds, will no longer be able to use malf abilities.
Added a malf-only gamemode, that makes malf AIs

#### Changelog

:cl:  
rscdel: Traitor AIs can no longer use malf abilities
rscadd: Malf AI now has its own gamemode
/:cl:


Lynameep:
Atmospherics is still the only thing an AI needs to completely destroy the station so this change won't really do much but further widen the gap between players who know atmos and the ones that dont.

CMOisLing:
Should tator AI be under the same murderbone rules regular traitor operate under?

Ratvar:
I don't think traitor AI should be grief-all: "I can plasmaflood roundstart" possible in every single gamemode but nukes is not a fun game, I would say, no matter how easy it is to EMP AI after it ruined the round for everyone

AsV9:
Traitor-AI will be subject to the murderbone as they were before traitor and malf-AI became the same gamemode.

config/game_options.txt
```diff
@@ -155,6 +157,7 @@ MIDROUND_ANTAG CLOCKWORK_CULT
 MIDROUND_ANTAG CHANGELING
 MIDROUND_ANTAG WIZARD
 #MIDROUND_ANTAG  MONKEY
+MIDROUND_ANTAG MALF
```
alexkar598, line 160: Won't this take a random Alice person and turn them into an AI mid round?

Amelia0010: no, midround_antag is whether or not the round-type will make new antags after the main ones die. You're thinking of `reroll_friendly` which is a var on the gamemode itself.

Alek2ander:
Malf AI is too damn weak to be its own gamemode

Bibby0110:
If you buff Malf AI a bit I see no problem with this.

boodaliboo:
yukjyytrgewoiyh6gfvrdcscjkdcmkdnkdsfnkdfnfgnnj yes

alexkar598:
@boodaliboo when you don't have anything to say other than "I like" or "I no like" just use thumbs up/thumbs down emojis instead of clogging comments

PR #6482, opened 2019-07-27
Yog box update

Readystorm546:
Okay so some new yog box shits yay wooh....

Capture 1
Capture 2
Capture 3

Take it as a you feel its just all here.


alexkar598:
Looks good apart from making the SMES accessible from maint without a space suit

Readystorm546:
I made it that way as a sort f blob and greytide nerf. as it still has access via engi though be it a bit harder.

PR #6487, opened 2019-07-28
Clock cult update 

Readystorm546:
Disregard all these god damm commits. besides the one. Clockwork one

i love clock cult and hate to see it on the back burner. I also think this will alow more people to enjoy the game mode and not wait 40 mins for something to happen. Ive asked people why they dont like clockies and ive heard alot about it being the fact that the gamemode is just too long. You wait around for 20 mins due to the prep time of crew and the final prep of the culties. Not alot of people enjoy conversion so i know that making it shorter will help with that.

This PR does what it says lowers the time it takes for clock cult to start the defense of REBE. Also the equation is used to determine how long that is depending on the amount of players. if there are 50+ players then it will always take 30mins but if its lower then 50 it will take less time. For instance if its 40 players it will be 40/5 + 20 = 28 mins.

🆑
rscadd: Clock cult rounds last a smaller amount of time now allowing for more fast passed game-play.
🆑


ThatLing:
Where is the changelog?

alexkar598 (approved):
Code looks good,add a change log as whathisname ling said

Readystorm546:
It has been added as per request

boodaliboo:
clock cult as it is feels to short this is a bad pr as clock cult feels rushed already

Readystorm546:
> clock cult as it is feels to short this is a bad pr as clock cult feels rushed already

You sir are the only one ive ever heard say this. Alot of people dont like clock cult because it take too long.

Amelia0010:
the few people that have actually bothered to voice their opinion apparently don't like this, and I don't really mind, so I guess this is it

PR #6492, opened 2019-07-29
Big Rebe change and very controversial 

Readystorm546:
So like i want to change clock cult to make it more "fun" and though hey what can we change to make this more "fresh" Well make it so clock cult has to defend 360 degree. Yeah this seems dumb but hey it could be fun. Who the hell knows.

ezgif com-optimize

This gif is weird for some reason but just know that the + goes clock wise around while moving.  Also for those who dont know this replaces that arrow that is at the south of the current clock cult base. 

image

The Center of the clock area

Clcok gif

A rough look at the outline around the base. Its about 10 tiles between the outside of the inner circle and the circle to where the culties can build.

:cl:  
rscadd: >whole new 360 clock cult map.
rscadd: >Changed arrows to a + so it makes more sense with this update.
rscadd: >Changed the clock cult paper to match the new map.
/:cl:


Readystorm546:
Disregard like all these commits just look at the files changed. i messed something up so all the commits are just me trying to fix shit.

jcatjmeow:
Remove cock cult whence

Szyszkrzyneczka:
speedmerge when

PR #6495, opened 2019-07-29
Nich you idiot

Nickvr628:
fixes nich's idiocy.

learn 2 type when


Nickvr628:
why does travis take so long

PR #6496, opened 2019-07-30
better Reebe reework [Clock Temple]

Nickvr628:
So everyone hates clock cult building because it isn't fun, the only viable strategy is to make an autism wall and checkerboard tables at the bottom, and it is literally just a football field for the crew to run through to get to the Ark. Security would just throw flashbang grenades and fire buckshot upwards to kill cultists and then run up to kill the Ark. Boring and dumb. So I changed it. 

image

Reebe is now a clock cult temple, with several large chambers the crew must fight through to get to the Ark. The goal was to make it more engaging to build defenses as you have chambers and a few choke points to fortify instead of building layers of walls across a football field of open space.
I also took it a step further and put some clock temple friendly items inside the chambers for ambiance like bookcases, tables, a prison cell, etc etc, that can be deconstructed to make more room for defenses.

The walls of the temple are indestructible but the windows are not so the crew cannot punch their way straight into the ark chamber, but of course any walls the clock cultists build will be destroyable as normal. Cultists can keep the windows for sight lines into the rooms or wall them off for extra defense. The amount of walls that generate components for the cult is unchanged from before so there are no balance concerns.

There are two main entrances, one on either side, so the cult cannot camp a single entrance. The crew must work through a series of chambers and hallways that the cult can fortify before assaulting the Ark chamber. Since the cult no longer has to build a massive trump wall at the bottom they can use resources and power for more complex and fun defenses, instead of spamming eyeballs and tables. We might actually see mania motors actually used, who knows!

:cl:
add: Clockwork's city of cogs is now a temple instead of a football field. New tactics will be required.
change: Clock Cult's new favorite food is Pineapple Pizza, because they are evil
/:cl:


Readystorm546:
Hmmmmmmmmmmmmmmmmmmm

AdamElTablawy:
so basically the cult gets free defenses

Lostlarry:
well if we make some floors that would prevent wall from being on top of them and put those in the middles of the rooms and on the bottom it may solve the problem but this is just a guess

nichlas0010:
image

I'll decide between these two tomorrow

nichlas0010:
https://github.com/yogstation13/Yogstation-TG/pull/6494 was testmerged based on maintainer feedback. This PR may get re-opened if everyone hates the other one, I guess?

PR #6499, opened 2019-07-30
malf-hacking no longer makes APCs blue

Amelia0010:

However, it still blocks crew from interacting with the APC, so there's still a way to tell.
image

This rule is dumb, IMO. With all other metashields, it's more like "you're not allowed to investigate whether or not this item that looks just like any other regular normal item is actually secretly a traitor item", compared to this rule, which is essentially "the AI is obviously malf, but let's give it 5 minutes till it's discovered by making you investigate whether there's fingerprints first"

In order to offset the fact that AIs would never get discovered at all, I've halfed the hackspeed and pointgan for APCs, meaning you need to get twice as many APCs, without disturbing the time-investment. This means the AI will have to spread its hacking further, making it more likely to get discovered

#### Changelog

:cl:  
tweak: Malf AIs hacking APCs no longer turns them blue.
tweak: Hacking APCs as a malf AI now gives you half as many points, but is twice as fast.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/ai/ai.dm
+++ b/code/modules/mob/living/silicon/ai/ai.dm
@@ -947,7 +947,7 @@
 		to_chat(src, "<span class='danger'>Hack aborted. \The [apc] is no longer responding to our systems.</span>")
 		playsound(get_turf(src), 'sound/machines/buzz-sigh.ogg', 50, 1, ignore_walls = FALSE)
 	else
-		malf_picker.processing_time += 10
+		malf_picker.processing_time += 5
 
 		apc.malfai = parent || src
 		apc.malfhack = TRUE
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -363,7 +363,7 @@
 			update_state |= UPSTATE_OPENED1
 		if(opened==APC_COVER_REMOVED)
 			update_state |= UPSTATE_OPENED2
-	else if((obj_flags & EMAGGED) || malfai)
+	else if(obj_flags & EMAGGED)
 		update_state |= UPSTATE_BLUESCREEN
 	else if(panel_open)
 		update_state |= UPSTATE_WIREEXP
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -1003,7 +1003,7 @@
 		return
 	to_chat(malf, "Beginning override of APC systems. This takes some time, and you cannot perform other actions during the process.")
 	malf.malfhack = src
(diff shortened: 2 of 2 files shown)


alexkar598:
Is it clear on the interface it's broken tho?

Amelia0010:
yes because you still can't use it :)

Bibby0110:
As in the UI never shows up with any hint about whats broken?
EDIT: NEVERMIND

Amelia0010:
as in you can't unlock it and it says this on examine 
image

Bibby0110:
It just acts as an emagged APC, yeah?

Amelia0010:
no, emagged APCs are forced unlocked, this is forced locked

Lostlarry:
what happens if APC is both hacked and emaged?

Amelia0010:
@fluffe9911 I don't play the game, how would this even be an i ded

@Lostlarry 🤷‍♂

ShadowDeath6:
I’d prefer something like the APCs turning blue after the AI hacks x amount, making it not ever show (on top of making AI able to hack twice as quickly) could result in really boring rounds if it decides to go nuclear. AI has plenty of ways to defend itself, I don’t think this is the way to go about making it more engaging.

ShadowDeath6:
A few other ways you could go about it are showing an examine message on the APC that indicates it’s hacked or making any APCs hacked past a certain amount appear blue like they do now

Amelia0010:
`(on top of making AI able to hack twice as quickly)` you get half the amount of points, so time to points ratio is the same

`A few other ways you could go about it are showing an examine message on the APC that indicates it’s hacked` The "the cover is broken. It may be hard to force open" message only appears when it has been malfhacked

`making any APCs hacked past a certain amount appear blue like they do now` I kinda like this idea, so I'll think it over

Moja10:
I ded, supreme salt PR, I support it anyway

ShadowDeath6:
Nuke timer scales with number of APCs hacked, so it does make a difference

Amelia0010:
>Nuke timer scales with number of APCs hacked, so it does make a difference

https://github.com/yogstation13/Yogstation-TG/blob/fcff07e9ceddc3cd405522b9e37eefaa8f2deb82/code/modules/antagonists/traitor/equipment/Malf_Modules.dm#L347

No it doesn't?

ShadowDeath6:
must have been changed on the removal of malf gamemode

Ludox235:
> 
> 
> must have been changed on the removal of malf gamemode

Nuke timer is always 450 seconds since it's no longer the ultimate objective of the AI.


By the way instead of doing a access denied or something just make so like... "The interface isn't responding" so it's not confusing or someshit to retards

alexkar598 (approved):

AsV9:
https://forums.yogstation.net/index.php?threads/vote-merge-6499.19608/

PR #6507, opened 2019-08-01
Omega bug fixing

Lostlarry:
Makes Omegas PDA system work form round start
also add a plating under a maint airlock 

:cl:    
bugfix: Omaga PDA system now works from roundstart
/:cl:


Nickvr628:
Omegalul

PR #6522, opened 2019-08-03
Kentucky fried tcomms

Lostlarry:
So tcomms machinery accumulate delay with each message they process. this delay is called traffic.
Every tick they lose some of this delay. This value is called netspeed.
This makes it so netspeed is lowered based on the temperature.
tcomms are usually kept at 80K.
according to this PR they would start loosing netspeed at 150K at the rate of 1% per 10K rounded (not sure how)
this caps out at 1150K where tcomms would be just accumulating delay
to counter that i added alert to server room air alarm at 1050K so it can be noticed by atmos and borgs
this can be used to silently sabotage tcomms and get out of the area before effects set in

so in addition to heat being bad for reducing traffic any traffic that is actually generate 2500J of heat(about 1,5K based on average heat cap of the air in the server room)

okay ran some test if the tcomms are really in use then temperature will slowly start climbing but nothing major

:cl:  
rscadd: makes tcomms loose eficiency at higher temperatures
experimental: tcomms generate heat this may be unbalanced as hell
/:cl:


CMOisLing:
Do Tcomms servers generate heat?

Lostlarry:
not yet. i'm planing on doing that but will take time balancing

CMOisLing:
Hurry up I want my Tcomms engine

Lostlarry:
you know that a cooler with basic parts has to be capable of keeping the thing cool

Ratvar:
I have a faint feeling this might generate some lag on a live server, for some reason

Lostlarry:
i would add testmerge request to test both lag and energy output but i don't know how to do that

Bibby0110:
If you go ahead with this, make some debug non-heating servers, thank you!

Lostlarry:
what exactly do you have in mind?
a static value to turn of the heating on all machinery or individual control?

Bibby0110:
make a variable to turn it off, and then make a subtype where it is turned off for mapping

Firewolf34:
I'm guessing there needs to be at least some kPa around the machines to absorb the heat, right? Because I'm wondering what would stop them from just venting the room to space to cool it. In real life, hard vacuum is actually a decent insulator and you need big radiators to dissipate heat. But in SS13 space is always really cold.

Lostlarry:
well I'm half sure that it's almost implausible to drain a room to 0 kPa which would actually make it easier to raise tepreature of the surrounding air

after doing some tests trying to depressurise tcomms is about the fastest way to over heat them

to the subtype part that is going to be a pain since i'm editing tcomms class which has the separate machinery as subclasses

Firewolf34:
Well thats good to hear. This will be a welcome change, this is what tcomms was supposed to be like from the start.

Any chance you could copy this feature over to the RnD server room as well?

Lostlarry:
i can try but i know far less about science servers then i know about tcomms machinery

code/game/machinery/telecomms/telecomunications.dm
```diff
@@ -141,11 +143,32 @@ GLOBAL_LIST_EMPTY(telecomms_list)
...
-		if (traffic < 0)  //yogs start
-			traffic = 0   //yogs end
+		var/deltaT = netspeed - speedloss  //yogs start
+		if (traffic < deltaT)
+			deltaT = traffic
+			traffic = 0
+		else
+			traffic -= deltaT
+		if(generates_heat)
+			env.temperature += deltaT * heatoutput / env.heat_capacity()   //yogs end
```
Amelia0010, line 174: this is making the temperature drop to several thousand negative degrees, but I'm too tired atm to actually do the math and tell you what needs changing, I'll do it tomorrow I promise

Amelia0010 (changes requested):

code/game/machinery/telecomms/telecomunications.dm
Lostlarry: unless the temperature overflows i don't  think there is a way to reach negative kelvins

Amelia0010: https://docs.google.com/spreadsheets/d/1CQD4fS4ZH7-kx2i__IBuLLH9UUWIQXHVBoRwjEuWj5w/edit?usp=sharing 
I *tried* to do the maths and I think this is accurate

Lostlarry: the heat cap is usualy about 800 but yes i noticed the mystake thanks for the help 
https://docs.google.com/spreadsheets/d/1IpBzpGGo3o2m5BNIaAP3q7jmcvge4zEFRMlIdCm7upM/edit#gid=0

Lostlarry:
i have no clue what the hell happened but i am making a new PR to make sure it's fine

MegaEmpirical:
You monster.


EDIT: Oh, okay.

PR #6524, opened 2019-08-03
IAA silicons and borged IAAs aren't the same thing

Lostlarry:
makes borged IAAs be target by kill objective instead of destroy one

:cl:  
tweak: Borged IAAs are considered dead for all IAAs
/:cl:

### DIFF
--- a/code/game/gamemodes/objective.dm
+++ b/code/game/gamemodes/objective.dm
@@ -810,7 +810,8 @@ GLOBAL_LIST_EMPTY(possible_items_special)
 /datum/objective/destroy/update_explanation_text()
 	..()
 	if(target && target.current)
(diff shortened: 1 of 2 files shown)


Amelia0010:
https://github.com/yogstation13/Yogstation-TG/blob/b5ed344f6d361ba033262e3e3c18bf35a7673424/yogstation/code/game/gamemodes/traitor/double_agents.dm#L30 this is where you'd want to add this change, and just use the `iscyborg()` proc instead of making a seperate var for it

Lostlarry:
i need the variable to discern whether you were a silicon when you gained the antag token.
if i did iscyborg then you would still get objectives like "destroy Questions-the-ai the experimental ai"

PR #6526, opened 2019-08-04
Ups the antag rep cap

ynot01:
Amount of cap increase is open to suggestion and discussion

The current antag rep is earned in about 3-4 rounds and only provides 2x antag chance

Some users (like theos) are very unlucky, and would benefit from this, and people are less likely to roll antag multiple rounds in a row this way.

council vote pls

#### Changelog

:cl:  
tweak: Antag rep cap has been increased
/:cl:

### DIFF
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -253,10 +253,10 @@ ALLOW_LATEJOIN_ANTAGONISTS
 ANTAG_REP_MAXIMUM 300
 
 ## The default amount of tickets all users use while rolling
-DEFAULT_ANTAG_TICKETS 200
+DEFAULT_ANTAG_TICKETS 100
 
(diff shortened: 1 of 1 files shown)


Bibby0110:
This is not doing what you think its doing.
You upped the cap to 300.
But you set the default tickets to 200,
You also set the max use to 200.

Which means, everyone has 200 tickets,
And can use up to 200 tickets, which is still a 2x chance

ynot01:
will fix but the code isn't the focus so much as the purpose of it

please consider using the review function next time

Amelia0010:
This PR does literally nothing, since the config is seperate on the host. If you want a council vote on changing the antag rep cap, go in #staff-public and ask for that instead of making PRs that don't do anything

Xoxeyos:
Why not just leave it truly random? Instead of rep, the system just chooses random people, without having to use a weighted system.

PR #6543, opened 2019-08-09
wizard hardsuit now nullifies' the user need for air. How? Magic!

### LINKED ISSUE #6500: A wizard who buys a gem-encrusted hardsuit does not receive internals.
Oversight.

## Reproduction:

Buy a gem-encrusted hardsuit. Teleport to space. Die.

Ludox235:
Wizard hardsuit now removes your need for air

Closes #6500 

:cl:
tweak: Wizard's gem encrusted hardsuit now makes so the user doesn't need to breathe anymore
/:cl:


Nickvr628:
If the hardsuit is magic it should have a magic bubble of air around it.

Ludox235:
> 
> 
> If the hardsuit is magic it should have a magic bubble of air around it.

I can try doing that

Ludox235:
Instead of making a "air bubble" appear, I have just put that the one who wears it does not need to breathe, because if it had a "air bubble", it would need some kind of visual indicator (which could be confused as a shield) and it wouldn't make sense that the "air bubble" wouldn't keep the wizard from dying of low pressure

MegaEmpirical:
Not supporting the bubble or not. But the suit does protect against low pressure, so using that as an excuse against the bubble is madness.

OppressedCactus:
Magic is cool, it is either this or having the hardsuit come with an air tank and mask.

alexkar598:
S. T. A. L. E.

PR #6545, opened 2019-08-09
Tanks now tell internal mol

yacabo:
When observed tanks will show how many mols of gas they have inside them along side pressure. As it turns out that pressure doesn't actually tell you how much gas there is inside a tank without first knowing the volume of the tank and a bit of math, as such I believe it is a natural choice to make it so that someone can check the mols

molage

:cl:  
rscadd: Tanks now show internal mol 
/:cl:

### DIFF
--- a/code/game/objects/items/tanks/tanks.dm
+++ b/code/game/objects/items/tanks/tanks.dm
@@ -81,7 +81,7 @@
 			. += "<span class='notice'>If you want any more information you'll need to get closer.</span>"
 		return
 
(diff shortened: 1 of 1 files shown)


ynot01:
feels like it takes away from the value of an analyzer

Laus4Deus:
It's useful but I like my analizer.

yacabo:
Analyzers serve various other purposes, currently the tank gauge servers very little in what one would need to know

alexkar598 (approved):

PR #6563, opened 2019-08-12
Streamlines the Tricorder's code

yacabo:
Improves the Tricorder's code so that it is not 70% copy paste. Functionally the only difference is that the medical side of it has a different UI seen here:

image

The code however has been modified so that the gas analyzer now has a part of its code turned into a proc that can be used elsewhere, and as you can imagine, is now used in the tricorder code.

Also it can no longer tell how long someone has been dead.

:cl:  
tweak: Tricorder code has been redone, functionally it only gives it a unique medical scanner UI
tweak: Part of the gas analyzer can be now used as a proc.
/:cl:


alexkar598:
Make it blue instead of neon green please

yacabo:
It's blue now but scanners.dm seems to of broken down 
so uh
standby

yacabo:
GitHub broke this PR, I'm going to need to make a new one, standby

yacabo:
New one here:
https://github.com/yogstation13/Yogstation-TG/pull/6564

PR #6564, opened 2019-08-13
Streamlines the Tricorder's code without breaking everything

yacabo:
Improves the Tricorder's code so that it is not 70% copy paste. Functionally the only difference is that the medical side of it has a different UI seen here:

image

The code however has been modified so that the gas analyzer now has a part of its code turned into a proc that can be used elsewhere, and as you can imagine, is now used in the tricorder code.

Also it can no longer tell how long someone has been dead.

(This is the second PR because GitHub broke the first (please don't kill me))
:cl:
tweak: Tricorder code has been redone, functionally it only gives it a unique medical scanner UI
tweak: Part of the gas analyzer can be now used as a proc.
/:cl:


code/game/objects/items/devices/scanners.dm
```diff
@@ -455,6 +455,10 @@ GENE SCANNER
 
 /obj/item/analyzer/attack_self(mob/user)
 	add_fingerprint(user)
+	scangasses(user)			//yogs start: Makes the gas scanning able to be used elseware
+
+proc/scangasses(mob/user)
```
Amelia0010:
```suggestion
/obj/item/proc/scangasses(mob/user)
```

yogstation/code/game/objects/items/devices/scanners.dm
```diff
@@ -55,101 +54,40 @@ obj/item/multitool/tricorder/suicide_act(mob/living/carbon/user)
...
-		for(var/id in env_gases)
-			if(id in GLOB.hardcoded_gases)
-				continue
-			var/gas_concentration = env_gases[id][MOLES]/total_moles
-			to_chat(user, "<span class='alert'>[env_gases[id][GAS_META][META_GAS_NAME]]: [round(gas_concentration*100, 0.01)] % ([round(env_gases[id][MOLES], 0.01)] mol)</span>")
-		to_chat(user, "<span class='info'>Temperature: [round(environment.temperature-T0C, 0.01)] &deg;C ([round(environment.temperature, 0.01)] K)</span>")
-
-//If medicalTricorder is set to 0 then the tricorder will not be as effective as a regular medical scanner
+//If medicalTricorder is set to FALSE then the tricorder will not be as effective as a regular medical scanner
 proc/lesserhealthscan(mob/user, mob/living/M)
```
Amelia0010:
```suggestion
/obj/item/proc/lesserhealthscan(mob/user, mob/living/M)
```

Amelia0010 (changes requested):

code/game/objects/items/devices/scanners.dm
yacabo: Done

code/game/objects/items/devices/scanners.dm
```diff
@@ -455,6 +455,10 @@ GENE SCANNER
 
 /obj/item/analyzer/attack_self(mob/user)
 	add_fingerprint(user)
+	scangasses(user)			//yogs start: Makes the gas scanning able to be used elseware
+
+obj/item/proc/scangasses(mob/user)
```
Amelia0010:
```suggestion
/obj/item/proc/scangasses(mob/user)
```

Amelia0010 (changes requested):

Amelia0010 (approved):

PR #6574, opened 2019-08-15
Kentucky fried tcomms try 2

Lostlarry:
Every tick they lose some of this delay. This value is called netspeed.
This makes it so netspeed is lowered based on the temperature.
tcomms are usually kept at 80K.
according to this PR they would start loosing netspeed at 150K at the rate of 1% per 10K rounded (not sure how)
this caps out at 1150K where tcomms would be just accumulating delay
to counter that i added alert to server room air alarm at 1050K so it can be noticed by atmos and borgs
this can be used to silently sabotage tcomms and get out of the area before effects set in

so in addition to heat being bad for reducing traffic any traffic that is actually generate 2500J of heat(about 1,5K based on average heat cap of the air in the server room)

okay ran some test if the tcomms are really in use then temperature will slowly start climbing but nothing major

:cl:  
rscadd: makes tcomms loose eficiency at higher temperatures
experimental: tcomms generate heat this may be unbalanced as hell
/:cl:


Amelia0010 (approved):

Amelia0010:
This has beent estmerged for like a week now, so it should be fine

PR #6580, opened 2019-08-16
Should fix the malf antag datum being available in the antag panel for humans

Amelia0010:
and also labels it as malf


Bibby0110:
bless

alexkar598 (approved):

PR #6610, opened 2019-08-23
You can no longer read announcements if you're unconscious

Amelia0010:

You can still hear the sound if you've got that prefence turned on though

#### Changelog

:cl:  
tweak: you can no longer read announcements if you're unconscious
/:cl:


TacticalSpacePilot:
This makes newscasters more useful too

PR #6618, opened 2019-08-24
Removes the shuttle lock on Revolution

ghost:

```
Revolution has only two ways of ending, execution or exile, both of which is completely insane in their own right as the solutions required to get there requires actions that if done wrong or on the wrong person, causes admin intervention. This enables a third option, escape. Heads of staff should be able to leave on the shuttle as all other options eventually cause admin intervention, for example, on one of the normal traitor rounds on delta?station, the flash caused a panic among the heads of staff and sec, resulting in admin intervention after the 'headrev' was killed, who was only a assistant memer with a flash.

Removing the shuttle lock allows the third option to be usable as it usually is an option that doesn't require admin intervention and is a universal option in most other gamemodes that don't have a time limit (95%, with the 5% being clockcult, and two event antags, xeno and blob).
```

#### Changelog

:cl:  
rscdel: Removed the shuttle lock, you can now leave without Centcom telling you no. 
/:cl:

### DIFF
--- a/code/game/gamemodes/revolution/revolution.dm
+++ b/code/game/gamemodes/revolution/revolution.dm
@@ -101,9 +101,6 @@
 	revolution.update_objectives()
 	revolution.update_heads()
 
-	SSshuttle.registerHostileEnvironment(src)
-	..()
-
 
 /datum/game_mode/revolution/process()
 	check_counter++


yacabo:
Ehhh, I agree with this but at the same time I don't want this to be a 'call the shuttle the second revs appear' situation, maybe delay a shuttle call for a significant amount of time?

MonsWTF:
Over the years of Rev rounds the formula has really gotten into a rut. If there are robust individuals on both the Heads and the Revs sides keeping the leaders alive, the round eventually drags to a crawl as either the Headrev is hunted down, or the Head Staff is hunted. I don't think just allowing the shuttle to leave is the right maneuver, so I'll list my current thoughts about it: 

Once the shuttle fails to depart the station at or after 45 min into the round, the server should randomly pick for (some) Ghosts to become one of two roles:
1) Corporate mercenary squadron tasked with eliminating the remaining Head Revs, pin-pointers included.
2) Revolutionary vanguard tasked with converting every last non-head, mind-shielded or not, pin-pointers included.

Only one of the two roles gets picked. The % chance could be 50-50, or auto-adjust in favor of the side with the most leaders still alive. Since the ERT is usually what happens in these drawn-out rounds, it may as well be hard-coded into the event itself.

yacabo:
I like it, however, why not have them both come together?

Alexandria404:
Against it.
I feel this PR comes from a place of not liking the round type for personal reason and trying to add a way to make it end early, instead of fixing the issues you might have with it.
I can understand you not liking it and wanting to see change, but this is not a solution.

MonsWTF:
> I like it, however, why not have them both come together?

Giving ghosts the options of choosing between both sides feels like a Ninjas situation. In addition if ghosts do evenly divide themselves (unlikely, though it could be coded to split them up like such) the fighting will still stagnate between the opposing sides, and has room to carry on just like before. 

The basic goal of the revolution/Heads should be to take out the opposite leaders or kick them off the station. If the round gets to the point that the shuttle has arrived and is sitting there, then either the Revs get extra men and a new win condition, or the Heads get a batch of armed thugs to try and gun down the remaining revolutionaries. Maybe to balance that side out, the mercs are aware of a timer ticking down until NT is 'forced' to give in to the Revs' demands for taking over the station. 

The big take-away is that the long-round of Revs changes in a meaningful way, as potentially:
- Revs close to winning suddenly have backup to give a new win condition
- Revs close to winning suddenly have to deal w/ the mercs or play cat-and-mouse.
- Heads close to winning suddenly have more military with pinpointers to the HeadRevs.
- Heads close to winning suddenly have more revs to deal with and a new lose condition.

ghost:
I'd rather have a gamemode that doesn't result in having to do cautious murderbone to actually end the round. Since I'm trying to argue that "Oh, it's a revolution, time to murderbone/deconvert (not getting anywhere)" is worse than "Oh, it's a revolution, time to call the shuttle as not to do anything that could cause the ire of the admins and end the gamemode without bloodshed."

ghost:
This gamemode has the same problem with gangs, in that it forces a confrontation to win, heads of staff get no warning that it's a revolution. They also have to deal with having no clues that point to it being a revolution. Meanwhile revs get admin protection, antag hud that's visible only to each other, instant conversion, and a murderbone pass on mindshielded/heads.

Also in gangs, it forces a confrontation WITH SEC over the disk. While I prefer that the gangs actually fight each other for the enemy gang tools, the fact that the disk is required to win basically gives the gangs a 'steal the authorization disk' objective. This would lead to gangs going after SEC the moment they gain a majority just like in revolution.

alexkar598:
Looking at thumbs up thumbs down ratio,noone likes it,and i dont like it,so yeet

"confrontation is badd"
antags are meant for confrontation,mode without confrontation are like hiveminds and overthrow,BORING AS FUCK

alexkar598:
Thank you yogbot

PR #6620, opened 2019-08-24
Networked fibers more like gone

ynot01:

feels like a long time coming, vote down if you disagree

#### Changelog

:cl:  
rscdel: Commented out networked fibers
/:cl:


AsV9:
Why is this a good idea? Please explain in depth.

Ratvar:
I propose you remove all chemical blobs that are ignored by borgs, anyone in spacesuit, make blobbernauths useless and die to stupid shit like double welder damage/extinguishers,

 leave synchronised mesh and networked fibers as the only blobs and then balance them.

ynot01:
> Why is this a good idea? Please explain in depth.

personally I believe that networked fibers is "win mode" and is agreed by all to be the most powerful blob type, as it generates more resources and if you even get close it can just snake out and instacrit you, just for the price of not being able to autogenerate blobs

fluffe9911:
*laughs in emp blob*

JamieD1:
ided rise up

PR #6640, opened 2019-08-26
Removes capitalism and communism

nesisknesis:
these arent used, why are they here, just to suffer?

#### Changelog

:cl:  Identification
imagedel: Removes capitalism and communism  
/:cl:


gilvian:
how could you remove capitalism, how can cargo survive now

PR #6653, opened 2019-08-28
Players can't click indestructible buttons

Amelia0010:

only really on centcom, adminghosts have to do it for em


Ratvar:
snowflake centcom button only whence

PR #6656, opened 2019-08-29
Being cluwned will no longer instantly kill you

SomeguyManperson:
:cl:  KNOH
bugfix: cluwneification no longer deals enough brain damage to kill the brain
/:cl:


MegaEmpirical:
Isn't it simply supposed to turn you into a cluwne?

SomeguyManperson:
it is but it also deals heavy brain damage, as of the organ damage update the damage it dealt was enough to cause the brain to fail on the spot

MegaEmpirical:
... I misread. I read it as you making it do enough to kill. My bad.

PR #6669, opened 2019-09-01
Drills no longer drill walls

alexkar598:
:cl:  alexkar598
rscdel: Drills no longer drills walls
/:cl:

Its bad design,a wall is supposed to be a heavy obstacle,it breaks the flow of the game when you are in fortified room and someone just breaks your 3 thick rwalls,walls are meant to be something you have to take time to break and breach to have a huge impact on the combat by changing the rules themself.

Obligatory credits: Port of tg PR tgstation/tgstation#10943


Xoxeyos:
You could just make it a pain in the ass to do, have it take a few minutes.

alexkar598:
That's an okay feature but rather pointless as tools pretty much do the same and aren't less cumbersome

alexkar598:
Are*

PR #6683, opened 2019-09-06
fucking doorbuttons mane

Amelia0010:

Doorbuttons are no longer indestructible. There's no longer an emtpy proc called removeMe


Bibby0110:
So what does removeMe() do?

Amelia0010:
image

PR #6685, opened 2019-09-06
Removes all lag forever

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Uses the magic of [Lunar Dreamland](https://github.com/goonstation/Lunar-Dreamland) condensed into a small package to ensure BYOND always has time to send map updates to players. It does so by measuring time taken to do that last tick and reserves enough time in the current one. This causes subsystems to process a bit slower but prevents 95% of lagspikes. (statistic pulled out of back orifice)
:cl: McDonald072
rscdel: Lag
/:cl:

If you want to port this to another codebase, make sure that you place _maptick_api somewhere it will be compiled before your master controller, and that you call `maptick_initialize()` and `maptick_shutdown()` in `world/New()` and `world/Reboot()` respectively.


gilvian:
we topped every PR, thats it. close the codebase

Bibby0110:
Stupid idea, but what about adding the .dll source code?

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Source for both DLLs available here: https://github.com/goonstation/Lunar-Dreamland

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Travis failed because it couldn't load the library, but it compiles and I tested it locally

ynot01:
we did it boys we stopped the ceo of lag

Altoids1:
Just for those who aren't in the know about this, this Lunar Dreamland stuff genuinely has the potential to completely resolve the lag situation and massively improve the performance of the game. If Mac can get this to work, long-term we can actually reduce lag by 90% or more.

So, yeah, Literally The Best PR. Love it.

oklews:
We’re in the endgame now

Couls:
image
Running your branch
image

code/game/world.dm
```diff
@@ -48,6 +48,9 @@ GLOBAL_VAR(restart_counter)
 
 	GLOB.timezoneOffset = text2num(time2text(0,"hh")) * 36000
 
+	world.log << call("hookerino.dll","BHOOK_Init")()
```
Couls: shouldn't this be uppercase H?

yacabo:
I believe in you, make this work, spite Travis

ThatLing:
You should really clean these lua files up

code/game/world.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: on windows the case in paths doesn't matter

1fbff5f83b23d39d38b1dfcb4cac8d9b:
@Couls I'm aware of that issue but it works on my machine™ so I'm having a hard time even starting to debug it, sadly

1fbff5f83b23d39d38b1dfcb4cac8d9b:
CHANGES!!

Threw out all that Lua crap and packaged everything into a single DLL. Now it should work better and faster ~~and stronger~~. It also should be very easy to port to all codebases that rely on measuring tick usage. Source code coming soon.

Bibby0110:
Inb4 Yogstation gets banned for EULA violations

Couls:
Can confirm it works with the new dlls, tested on paradise codebase

Bibby0110:
tHIS IS VERY NICE

AsV9:
I'd like to look at those .dlls aswell.

Readystorm546:
Travis is not happy

optimumtact:
if I were you, I'd hard reserve at least 20% of a tick for your game code.

SSInput or SStimer missing ticks will cause all sorts of really noticeable issues for players.

Actually better in those cases to go over a tick

actioninja:
https://github.com/BeeStation/BeeStation-Hornet/pull/412#issuecomment-530555134
relevant comment from beestation devs after testing

ghost:
Would this work on Linux, if it was compiled for it?

1fbff5f83b23d39d38b1dfcb4cac8d9b:
I'd assume if you made the necessary modifications then yes, but I don't know how it's implemented on linux

ghost:
I THINK lunar dreamland has Linux support, judging from all the 
```
#ifdef _WIN32
#else
#endif
```

1fbff5f83b23d39d38b1dfcb4cac8d9b:
lunar itself does, but the library I wrote here does not

ghost:
oh.

Uhhh where's the source for said lib

1fbff5f83b23d39d38b1dfcb4cac8d9b:
proprietary

1fbff5f83b23d39d38b1dfcb4cac8d9b:
https://github.com/goonstation/Lunar-Dreamland/tree/master/maptick

You can see the source for the library here for now

alexkar598:
Stale, said he will be coming back to it. Closing PR until then.

PR #6690, opened 2019-09-08
the tesla now has a max of 10 balls

Amelia0010:
self-explanatory


Altoids1:
It'd be fun if it actually did something interesting when it hit the ballcap, like reached some new level of tesla or do a big ol' zap or something

oklews:
was there a reasoning behind this?

Amelia0010:
image

PR #6706, opened 2019-09-10
Lag removal, part 2: Non-blocking SQL

1fbff5f83b23d39d38b1dfcb4cac8d9b:
image

Adds async promise-based DLL calls and changes BSQL to use them. Apparently it's supposed to already be non-blocking but the numbers say otherwise.

I'll add the source code for the DLL, promise


alexkar598:
Stale, said he will be coming back to it. Closing PR until then.

PR #6715, opened 2019-09-13
.38 rounds deal 35 instead of 60 stamina damage

SomeguyManperson:
This brings the detective's revolver relatively in line with stamina based weapons, since the 60 stamina damage was from when tasers still existed.
:cl:  
tweak: Detective revolver deals less stamina damage, takes roughly 3 shots to stamcrit someone now
/:cl:


AsV9:
manually done

PR #6722, opened 2019-09-15
removes exile implants from cargo AGAIN

SomeguyManperson:
They don't do anything
:cl:  
rscdel: Removes exile implants from cargo since they are useless
/:cl:


alexkar598 (changes requested):
Delete instead of commenting out code

PR #6744, opened 2019-09-23
buffs holy explosion anti-cult blast

SomeguyManperson:
Now triggers at 100/100 potwat and has a range of created volume/20 instead of created volume/48

:cl:  
tweak: Holy hand grenades are more viable, and the holy hand grenade item will now trigger the holy explosion effect
/:cl:


alexkar598 (approved):
sure

PR #6761, opened 2019-09-26
Puts a bandaid on VTEC while I work on a real rebalance

ynot01:

i ded

#### Changelog

:cl:  
tweak: VTEC has been temporarily halved in efficiency
/:cl:


alexkar598 (changes requested):
Last time I saw something like this,it got temporarily super nerfed until someone made a fix,which the latter never happened. Just make your rebalance instead of fake PRs that remain "until i get around to it"

ynot01:
might be awhile with college clogging my schedule

PR #6763, opened 2019-09-27
Stabilized Pink now stops mobs from spazzing out (Sometimes)

ghost:

Fixes mobs spazzing out when encountering them if you encountered them before. This is done by removing the line that removes the faction of the mob from you when out of sight. 

An example of this fix is that it checks the mob that you see, and when out of sight and turning a corner, you don't get attacked immediately due to losing the faction due to being out of sight.

#### Changelog

:cl:  

tweak: Stabilized Pink now is a bit more safer.

/:cl:


alexkar598:
Was this tested in game?

ghost:
Yes, I tested it with three different scenarios and with the wizard mob.

First Encounter: Mobs still spaz out as the one wearing the item didn't have the faction from viewing yet.

Spawn spam: The mobs spazed out by using magic missile, but this was probably due to spawning bugs

Out and In of view: The mobs didn't react, this was what I set out to solve when I removed the line of code.

alexkar598:
Sure fine

PR #6764, opened 2019-09-27
You can no longer print AI or Cyborg upload boards in the Science R&D lathe

AsV9:
Makes AI harder to subvert and makes the secure tech storage actually slightly more useful.

:cl:   
rscdel: You can no longer print AI or Cyborg upload boards in R&D
/:cl:

### DIFF
--- a/code/modules/research/designs/comp_board_designs.dm
+++ b/code/modules/research/designs/comp_board_designs.dm
@@ -45,25 +45,7 @@
 	build_path = /obj/item/circuitboard/computer/xenobiology
 	category = list("Computer Boards")
 	departmental_flags = DEPARTMENTAL_FLAG_SCIENCE
-/*
-/datum/design/board/aiupload
-	name = "Computer Design (AI Upload)"
-	desc = "Allows for the construction of circuit boards used to build an AI Upload Console."
-	id = "aiupload"
-	materials = list(MAT_GLASS = 1000, MAT_GOLD = 2000)
-	build_path = /obj/item/circuitboard/computer/aiupload
-	category = list("Computer Boards")
-	departmental_flags = DEPARTMENTAL_FLAG_SCIENCE
 
-/datum/design/board/borgupload
-	name = "Computer Design (Cyborg Upload)"
-	desc = "Allows for the construction of circuit boards used to build a Cyborg Upload Console."
-	id = "borgupload"
-	materials = list(MAT_GLASS = 1000, MAT_GOLD = 2000)
-	build_path = /obj/item/circuitboard/computer/borgupload
-	category = list("Computer Boards")
-	departmental_flags = DEPARTMENTAL_FLAG_SCIENCE
-*/
 /datum/design/board/med_data
 	name = "Computer Design (Medical Records)"
 	desc = "Allows for the construction of circuit boards used to build a medical records console."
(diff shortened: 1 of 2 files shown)


alexkar598 (changes requested):
GTFO with commented out code,if your removing,delete dont comment

alexkar598 (approved):

Bibby0110:
So uh, you can't get these boards apart from the 1 set in tech storage now?

MegaEmpirical:
I think considering a way to acquire them outside of that would be good, I am not sure what form it should take though. However, I think removing them from being so easily printed is a good call.

MegaEmpirical:
Current suggestion: Make the AI upload console in the core invulnerable. This way there is always a location to subvert/restore an AI but it is also dependable by the AI and has to be sieged, much like how it used to be (mostly).

Or maybe leave it as just having only those two boards like it used to be but then any rogue AIs will have to be killed and you'll have no means of modifying new AIs.

MegaEmpirical:
Alternative: we make the boards unbreakable like the nuke disk and make it so that they can be GPS tracked even when not in a console

Mr-Space:
good pr merge
kthnks

Sir-Lagsalot:
🦀 Fuck subversion-rushing the AI 🦀

MegaEmpirical:
Other suggestion, this one got more traction:

1) Leave AI upload boards printable

2) Make it so an AI has to be tagged to a board for its laws to be changed.

3) Spawned AI would start tagged to upload console and the spare (permit tagging mutliple AIs to one board for convenience)

4) You have to click the board on an AI to tag it

This would make it so the originals are still important and so that other AIs can still be made. The issue is that the new meta would be to make new subverted AIs and hide them and no one would be able to do anything. My fix for this is to make all AI cores GPS trackable.

Alek2ander:
The ease of subverting an AI is compensated by the ease of resetting it. 
What will the crew have to do when the RD subverts the AI using their upload access and tells the engiborg to deconstruct upload and space the boards? Is there any way to fix the AI that doesn't require breaking into the core?

Alexandria404:
Very needed change.
To much to quick subversion.

MegaEmpirical:
> The ease of subverting an AI is compensated by the ease of resetting it.
> What will the crew have to do when the RD subverts the AI using their upload access and tells the engiborg to deconstruct upload and space the boards? Is there any way to fix the AI that doesn't require breaking into the core?

If referring to my last suggestion, the possibility of reverting the AI still exists compared to other suggestions. It will just be very difficult since you'd have to tag the core, so you're more likely to try to kill the AI. All my previous suggestions were called 'snowflake' (whatever the fuck that means) even though I think making the core upload invulnerable makes the most gameplay sense.

However, this last suggestion at least still permits the creation of new AIs (either to combat the other AI or to replace it).

MonsWTF:
Have we considered a message being broadcast over one of the radio channels when a new AI/Borg Upload Console is activated? Presumably either default channel or Science/Security.

**[Common]** (AI Upload Console) _states:_ "New access console initiated."

The boards can still be printed, but making one stealthily is significantly harder, unless you 
1) Wait for a Comms Blackout, or 
2) Create the Comms Blackout yourself

Pros: 
- Nerds can still make the boards
- Difficulty in (quietly) subverting AI/Borg increased, which is the point of the PR

Cons: 
- The backup one in Tech Storage is still pretty useless


P.S. On a sidenote, this might be worth considering doing for those nerds that like to make a Communications board and maint-Recall.

PR #6765, opened 2019-09-27
Decrease R&D single-server income by 15%

AsV9:
Partial revert of tgstation/tgstation#33857

Research has more ways of generating points now, and the passive cookie clicker gain should reflect that.

:cl:  
tweak: Decrease R&D single-server income by 15% 
/:cl:

### DIFF
--- a/code/controllers/subsystem/research.dm
+++ b/code/controllers/subsystem/research.dm
@@ -30,7 +30,7 @@ SUBSYSTEM_DEF(research)
 	var/list/errored_datums = list()
 	var/list/point_types = list()				//typecache style type = TRUE list
 	//----------------------------------------------
-	var/list/single_server_income = list(TECHWEB_POINT_TYPE_GENERIC = 54.3)
+	var/list/single_server_income = list(TECHWEB_POINT_TYPE_GENERIC = 46.2)
 	var/multiserver_calculation = FALSE
 	var/last_income = 0
 	//^^^^^^^^ ALL OF THESE ARE PER SECOND! ^^^^^^^^


boodaliboo:
ok so the other ways of  getting points which is expermental dissection which needs points in the first place or toxins and people only do toxins for bombs or dont do it at all

alexkar598:
I don't like this PR for these reasons:

1) It doesn't fix the problem of RnD being noninteractive and just makes it take more time before the grey button becomes blue
2) It doesn't fix the problem of RnD(and roboticts depending on mining) being powerful after 30 minutes by doing nothing at all,this just makes it take 35 minutes instead

Oakboscage:
> I don't like this PR for these reasons:
> 
> 1. It doesn't fix the problem of RnD being noninteractive and just makes it take more time before the grey button becomes blue
> 2. It doesn't fix the problem of RnD(and roboticts depending on mining) being powerful after 30 minutes by doing nothing at all,this just makes it take 35 minutes instead

Better than nothing innit

Alexandria404:
> 
> 
> > I don't like this PR for these reasons:
> > 
> > 1. It doesn't fix the problem of RnD being noninteractive and just makes it take more time before the grey button becomes blue
> > 2. It doesn't fix the problem of RnD(and roboticts depending on mining) being powerful after 30 minutes by doing nothing at all,this just makes it take 35 minutes instead
> 
> Better than nothing innit

alexkar598:
Tbh it's not better than nothing,it's just makes science more frustrating without actually fixing or starting to fix any of the underlying issues

ynot01:
lets lock point generation behind a node and force you to beg engineering or toxins to make points for you :))

alexkar598:
This doesn't fix or improve any of the core issues in research as it is currently implemented and the time difference between old and new RnD will barely be noticeable as advanced mining and T4 will still Be rushed although maybe a 5 minutes later and by the time mining gets enough mats to make mechs science will have plenty of points to research mechs and xeno will have their console research alongside the T4 anyways so all the three biggest points of impact to the round science can have will not be impacted on by this PR.

PR #6776, opened 2019-09-30
Maybe fixes security consoles not being able to delete citations

### LINKED ISSUE #6771: Cannot delete citations from security records
It just gives me a line about this function not working currently.

Amelia0010:
fixes: #6771 

I think. Idk, it needs to have "edit field" yet it's given "edit  field" so this should work


alexkar598:
You sure you can shove spaces in choice?

Amelia0010:
it works for everything else, only one that doesn't work is this one, which has 2 spaces rather than one

PR #6777, opened 2019-09-30
Chem implants now dump contents on death

### LINKED ISSUE #6733: Chem implants don’t dump contents
## Reproduction:
Chem implant was given to a human monkey and filled with 49 cryoxadone and 1 blood. He was infected with a regen coma with the fake death threshold met(likely the cause of the issue). Upon his real death the reagents from the implant should have been released but it never happened. Likely because it is triggered by the deathgrasp emote instead of an actual death trigger.

CMOisLing:
Fixes: #6733

Not sure if this is a bug fix or feature but this should allow for some more random shenanigans or maybe something useful like a make shift bomb implant
:cl:  
bugfix: Chem implants dump contents on death
/:cl:


code/game/objects/items/implants/implant_chem.dm
```diff
@@ -29,11 +29,8 @@
 	GLOB.tracked_chem_implants -= src
 	return ..()
 
-/obj/item/implant/chem/trigger(emote, mob/living/source)
-	if(emote == "deathgasp")
-		if(istype(source) && !(source.stat == DEAD))
-			return
-		activate(reagents.total_volume)
+/obj/item/implant/chem/on_mob_death(mob/living/L, gibbed) //yogs - chem implants dump chems when host dies
+	activate("action_button")
```
alexkar598:
```suggestion
	activate("death")
```

alexkar598 (changes requested):

code/game/objects/items/implants/implant_chem.dm
CMOisLing: I don't mind changing this, but I was wondering is this so its easier to read or is action_button used for something else in this implants case?

alexkar598: Probably doesnt affect anything but its good practice to keep shit as its meant and not something random because you saw it like that somewhere else.

BTW,modularization is no longer enforced so you dont need to do the ``//yogs - chem implants dump chems when host dies``

CMOisLing:
Just add both labels

PR #6782, opened 2019-10-01
Fixes mech fabs going in softlock if it runs out of mats to build something in a queue

alexkar598:
:cl:  alexkar598
bugfix: Mech fabs no longer go into softlock if they run out of mats.
/:cl:


alexkar598:
Imagine not being a self merging crab!

PR #6794, opened 2019-10-02
ENFORCE SECOND AMMENDMENT / SURVIVALIST OBJECTIVE CHANGE

boodaliboo:
MY PR THAY MAYBE WORKS THAT CHANGES THE OBJECTIVES OF SURVIVALIST SO ITS NOT JUST MURDERBONE FOR THE GUNS OR MAGIC

#### Changelog

:cl:  

tweak: tweaked  THE OBJECTIVES OF SURVIVALIST SO THEY DONT HAVE TO STEAL MORE GUNS BUT THEY MUST KEEP THERE ORIGINAL 

spellcheck: fixed a THING MAYBE  THAT DISAGEE
experimental: I WROTE THIS ON GITHUB WEBSITE ON A 10 YEAR OLD MAC
/:cl:

### DIFF
--- a/code/modules/antagonists/survivalist/survivalist.dm
+++ b/code/modules/antagonists/survivalist/survivalist.dm
@@ -19,12 +19,11 @@
 	owner.announce_objectives()
 
 /datum/antagonist/survivalist/guns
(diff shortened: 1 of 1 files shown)


ynot01:
this is bad because it takes away from the point of the spell, to create antags so the wizard has an easier time

it doesn't do its job well on any points and just causes issues so that's why we're removing it, and this doesn't fix it

alexkar598:
Low effort

PR #6803, opened 2019-10-03
Sepia slime timestop now affect the user also

alexkar598:
:cl:  alexkar598
tweak: Xenobiology sepia slime timestop now also affects the user.
/:cl:


MonsWTF:
Spoiler warning:


_That's how Sepia works already_

alexkar598:
Ah fuck that's the lumi whatever effect

PR #6804, opened 2019-10-03
Makes light pink extract give a better med to keep the patient alive

alexkar598:
:cl:  alexkar598
tweak: Stabilized light pink extract now gives the user a small dose of Ephedrine as a speed boost if someone in crit is near them
tweak: Stabilized light pink extract now give the patient in crit epinephrine and atropine instead of epinephrine and saline glucose solution
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
+++ b/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
@@ -904,15 +904,13 @@ datum/status_effect/stabilized/blue/on_remove()
 	colour = "light pink"
 
 /datum/status_effect/stabilized/lightpink/tick()
-	var/boost = FALSE
 	for(var/mob/living/carbon/human/H in range(1, get_turf(owner)))
-		if(H != owner && H.stat != DEAD && H.health <= 0 && !H.reagents.has_reagent(/datum/reagent/medicine/epinephrine) && !H.reagents.has_reagent(/datum/reagent/medicine/salglu_solution))
+		if(H != owner && H.stat != DEAD && H.health <= 0 && !H.reagents.has_reagent(/datum/reagent/medicine/epinephrine))
 			boost = TRUE
 			to_chat(owner, "[linked_extract] pulses in sync with [H]'s heartbeat, trying to keep [H.p_them()] alive.")
 			H.reagents.add_reagent(/datum/reagent/medicine/epinephrine,5)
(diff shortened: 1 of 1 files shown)


ghost:
so it heals people out of crit with one of the best anti-crit chems in the game which heals incredibly fast AND gives the owner a speedboost and stun resistance if they're near someone in crit

how about this

close this PR or i'm removing timestop slimes

alexkar598:
How about this

Fuck you

MonsWTF:
Big oof.

Uhh, personal comments:

1) If Stabilized Light Pink has speed (if at all), it has to either be **extremely** conditional to prevent antag abuse, or a constant minor speed increase (-0.1 or -0.2, probably).
2) Instead of replacing Saline-Glucose with Atropine (reason addressed by Adam), add Regenerative Jelly (Omnizine, but slightly better) as a third chem with its own condition; it only activates when `H.health <=-50` , the halfway point of crit, and adds 1u Regen Jelly. This would fit a lot better with the slime theme and would only activate when really trying to keep the victim alive.

The 1u amount also means that the chem won't stay in the victim's body for long if they climb back up past the -50 threshold.

ThatLing:
Thank you for the detailed explanation on why this PR should be merged

alexkar598:
@Mr-Space @AdamElTablawy @MonsWTF @ThatLing it nows injects 10 units of regen jelly if the person is in less than -40 health crit which is enough to heal the person out of crit to literally 0 health in about 30-40 seconds. Also removed the speed boost as it was just dumb since the mix is now more than able to keep someone alive long enough to bring them to medbay. As god intended

ShadowDeath6:
I think this pr is good in a vacuum but idk if xeno really needs even more power on its current extracts considering what a hot topic it is already in terms of balance. 
I'm not entirely sure how often the extract ticks (if it even does tick more than once), but regen jelly is very very strong and it can probably bring someone out of crit completely with some health to spare (depending on how much health they have when the jelly is put in) since epinephrine will already take care of the oxygen damage for the most part

alexkar598:
@ShadowDeath6 the jelly will only happen if the person is under -40 health and has no jelly in their system at all,assuming they are at -50 health when it injects it,it gives 10u of the jelly which takes about 30-40 seconds to process all and gets the person at -4 health which isnt that bad but not enough to make them ready for combat again

PR #6807, opened 2019-10-03
Limits the Strange Beacon to Hijack and Glorious Death

Sir-Lagsalot:

Makes it so that the strange beacon can only be purchased by traitors with murderbone objectives. I can't really think of any way to use this item without murderboning, and I have had to handle issues before with traitors who thought that because this was a thing that they were allowed to purchase as a traitor then it was something they were actually allowed to, yknow, use without breaking the rules. Other murderbone-y antag items such as the spider injector and His Grace are restricted to murderbone objectives. To avoid confusion, I think this one should be as well.

I'm not sure what's with all the extra commits, the only thing I changed was uplink_items.dm. Theos said it should be okay though.

#### Changelog

:cl:  
tweak: The Strange Beacon traitor item is now only purchasable for traitors with hijack or glorious death
/:cl:


CMOisLing:
To be fair you could use it to pull the singularity to a safe part of the station so you can get a shuttle call without risking your life but then again this is a very niche use

Amelia0010 (approved):
consider the whitespace changes a fix

PR #6831, opened 2019-10-08
Adds crawling shadows ability to darkspawn

SomeguyManperson:
costs 60 psi and 2 psi points, makes the user invisible in the dark and near-invisible in light. Doesn't take damage from light and can walk through doors by attacking them, attacking people will stun them after a short delay. Lasts a minute or until cancelled.

:cl:  
rscadd: Adds a new ability, crawling shadows, to darkspawn that can be used as a movement and offensive ability 
/:cl:


oklews:
Would be interesting if you have markings on the door passed through, something like “the lights on the door glow very faintly” or if you’re thinking animalistic “There are giant *claw* marks on the door!”

PR #6834, opened 2019-10-08
Readds high functioning zombies and vampires to halloween holiday

fluffe9911:
🆑
tweak: High functioning zombies and vampires are readded to the halloween holiday!
/🆑

The main reasons they got removed were because they were too annoying for admemes for a holiday that lasted a whole month now that its back to what it originally was might as well revert this as well (except for shadowlings cause ehhhhhhhh)

### DIFF
--- a/yogstation/code/modules/mob/living/carbon/human/species_types/halloween.dm
+++ b/yogstation/code/modules/mob/living/carbon/human/species_types/halloween.dm
@@ -1,9 +1,2 @@
 /datum/species/shadow/check_roundstart_eligible()
           return FALSE
-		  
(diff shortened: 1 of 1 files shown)


ThatLing:
The only reason you want these are so you can powergame. Vampires can ventcrawl, skeletons and zombies can spacewalk. These should stay dead and skeletons can die too.

If you want to keep the 'halloween' feel of these races, make it so these are purely cosmetic and don't provide any (dis)advantages over normal humans. Until then, -1 from me.

fluffe9911:
Alrighty

Alexandria404:
Visuals sure, functions seem OP.

PR #6841, opened 2019-10-09
Made most of the halloween races visual only and added some new ones! (also turned the date back to start on the 24th)

fluffe9911:
🆑
tweak: All the Halloween races except for the one race that can take there head off are now visual only!
tweak: Halloween now starting early once more!
tweak: Now heads of staff can also join in the Halloween spirit!
tweak: Removed cloth golem from the halloween race list
/🆑
Vampire is also still removed because they arnt unique enough visually and added fly people, visual only mush people and visual only abductors as valid halloween races

### DIFF
--- a/code/modules/holiday/holidays.dm
+++ b/code/modules/holiday/holidays.dm
@@ -350,7 +350,7 @@
 
 /datum/holiday/halloween
 	name = HALLOWEEN
-	begin_day = 26
+	begin_day = 10 // yogs - Changed halloween start day
 	begin_month = OCTOBER
 	end_day = 2
 	end_month = NOVEMBER
--- a/code/modules/holiday/holidays.dm
+++ b/code/modules/holiday/holidays.dm
@@ -350,7 +350,7 @@
 
 /datum/holiday/halloween
 	name = HALLOWEEN
-	begin_day = 10 // yogs - Changed halloween start day
+	begin_day = 24 // yogs - Changed halloween start day
 	begin_month = OCTOBER
 	end_day = 2
 	end_month = NOVEMBER
--- a/code/modules/mob/living/carbon/human/species_types/abductors.dm
+++ b/code/modules/mob/living/carbon/human/species_types/abductors.dm
@@ -17,3 +17,19 @@
 	. = ..()
 	var/datum/atom_hud/abductor_hud = GLOB.huds[DATA_HUD_ABDUCTOR]
 	abductor_hud.remove_hud_from(C)
+
+/datum/species/fakeabductor //yogs start
+	name = "Fake Abductor"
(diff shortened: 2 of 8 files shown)


Mr-Space:
Visuals yes, early halloween no.

alexkar598 (approved):

alexkar598:
What space said

ynot01:
i say yes to early because im impatient

boodaliboo:
this is stupid the gimmick of most of these events like Halloween is a depart from the normal without turning the whole game differently like with christmas and santa. so i think we should keep the 3 races we have for halloween the same as its a unique thing that doesn't even last that long anyways

code/modules/holiday/holidays.dm
```diff
@@ -350,7 +350,7 @@
 
 /datum/holiday/halloween
 	name = HALLOWEEN
-	begin_day = 26
+	begin_day = 10 // yogs - Changed halloween start day
```
AsV9: no, change this to the 24th at the very earliest.

AsV9 (changes requested):

code/modules/holiday/holidays.dm
boodaliboo: it used to be the 10th when was it changed

AsV9 (approved):

ThatLing:
> 
> 
> this is stupid the gimmick of most of these events like Halloween is a depart from the normal without turning the whole game differently like with christmas and santa. so i think we should keep the 3 races we have for halloween the same as its a unique thing that doesn't even last that long anyways

ok powergamer

AsV9 (commented):
Frankly I think you should add in a way to tell the fake shadowlings and fake abductors apart from real antags in a reasonably quick way considering people will probably get themselves valided by some retard.

boodaliboo:
> > this is stupid the gimmick of most of these events like Halloween is a depart from the normal without turning the whole game differently like with christmas and santa. so i think we should keep the 3 races we have for halloween the same as its a unique thing that doesn't even last that long anyways
> 
> ok powergamer

i joined at the same time this was a event last year i played  a total of 2 rounds on this

ThatLing:
> 
> 
> > > this is stupid the gimmick of most of these events like Halloween is a depart from the normal without turning the whole game differently like with christmas and santa. so i think we should keep the 3 races we have for halloween the same as its a unique thing that doesn't even last that long anyways
> > 
> > 
> > ok powergamer
> 
> i joined at the same time this was a event last year i played a total of 2 rounds on this

ok powergamer

Reed0506:
what is even the point of halloween if you cant be an actual skeleton and die from the chemist making a smoke machine with bone hurting juice

PR #6865, opened 2019-10-11
Secborgs now have an autocharging disabler instead of an infinite one

ynot01:

cyborg disabler has 10 shots
laser has 8 shots
recharges like adv egun

#### Changelog

:cl:  
tweak: Security module cyborgs now have an auto-charging disabler instead of one that draws from their cell
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/laser.dm
+++ b/code/modules/projectiles/guns/energy/laser.dm
@@ -48,8 +48,9 @@
 
 /obj/item/gun/energy/laser/cyborg
 	can_charge = FALSE
-	desc = "An energy-based laser gun that draws power from the cyborg's internal energy cell directly. So this is what freedom looks like?"
-	use_cyborg_cell = TRUE
+	desc = "An energy-based laser gun that self charges. So this is what freedom looks like?"
+	charge_delay = 5
+	selfcharge = 1
 
 /obj/item/gun/energy/laser/cyborg/emp_act()
 	return
--- a/code/modules/projectiles/guns/energy/laser.dm
+++ b/code/modules/projectiles/guns/energy/laser.dm
@@ -50,7 +50,7 @@
 	can_charge = FALSE
 	desc = "An energy-based laser gun that self charges. So this is what freedom looks like?"
 	charge_delay = 5
-	selfcharge = 1
+	selfcharge = TRUE
 	ammo_type = list(/obj/item/ammo_casing/energy/laser/cyborg)
 
 /obj/item/gun/energy/laser/cyborg/emp_act()
--- a/code/modules/projectiles/guns/energy/laser.dm
+++ b/code/modules/projectiles/guns/energy/laser.dm
@@ -51,6 +51,7 @@
 	desc = "An energy-based laser gun that self charges. So this is what freedom looks like?"
 	charge_delay = 5
 	selfcharge = 1
+	ammo_type = list(/obj/item/ammo_casing/energy/laser/cyborg)
 
 /obj/item/gun/energy/laser/cyborg/emp_act()
 	return
--- a/code/modules/projectiles/guns/energy/stun.dm
+++ b/code/modules/projectiles/guns/energy/stun.dm
@@ -47,4 +47,5 @@
 	charge_delay = 5
 	can_charge = FALSE
 	selfcharge = 1
+	charge_amount = 3
 	ammo_type = list(/obj/item/ammo_casing/energy/disabler/cyborg)
--- a/code/modules/projectiles/guns/energy/stun.dm
+++ b/code/modules/projectiles/guns/energy/stun.dm
@@ -43,6 +43,7 @@
 
 /obj/item/gun/energy/disabler/cyborg
 	name = "cyborg disabler"
-	desc = "An integrated disabler that draws from a cyborg's power cell. This weapon contains a limiter to prevent the cyborg's power cell from overheating."
+	desc = "A cyborg-integrated disabler that self charges."
+	charge_delay = 5
 	can_charge = FALSE
-	use_cyborg_cell = TRUE
+	selfcharge = 1
--- a/code/modules/projectiles/guns/energy/stun.dm
+++ b/code/modules/projectiles/guns/energy/stun.dm
@@ -46,6 +46,6 @@
 	desc = "A cyborg-integrated disabler that self charges."
 	charge_delay = 5
 	can_charge = FALSE
-	selfcharge = 1
+	selfcharge = TRUE
 	charge_amount = 3
 	ammo_type = list(/obj/item/ammo_casing/energy/disabler/cyborg)
--- a/code/modules/projectiles/guns/energy/stun.dm
+++ b/code/modules/projectiles/guns/energy/stun.dm
@@ -47,3 +47,4 @@
 	charge_delay = 5
 	can_charge = FALSE
 	selfcharge = 1
+	ammo_type = list(/obj/item/ammo_casing/energy/disabler/cyborg)
--- a/code/modules/projectiles/ammunition/energy/laser.dm
+++ b/code/modules/projectiles/ammunition/energy/laser.dm
@@ -32,6 +32,11 @@
 	variance = 15
 	harmful = FALSE
 
+/obj/item/ammo_casing/energy/laser/cyborg
+	projectile_type = /obj/item/projectile/beam/laser
+	select_name = "kill"
+	e_cost = 125
+
 /obj/item/ammo_casing/energy/laser/heavy
 	projectile_type = /obj/item/projectile/beam/laser/heavylaser
 	select_name = "anti-vehicle"
--- a/code/modules/projectiles/ammunition/energy/stun.dm
+++ b/code/modules/projectiles/ammunition/energy/stun.dm
@@ -24,3 +24,6 @@
 
 /obj/item/ammo_casing/energy/disabler/hos
 	e_cost = 60
+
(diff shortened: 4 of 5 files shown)


oklews:
and the disabler upgrade would give them the one we have now?

ynot01:
> and the disabler upgrade would give them the one we have now?

no its gone forever

and next im going for the stun batons

yacabo:
Yeah that's a terrible idea

ynot01:
oh no its the silicon mains here to defend their infinite disablers

yacabo:
It takes directly from their battery supply, and a hefty take at that

ynot01:
> It takes directly from their battery supply, and a hefty take at that

its 50 power per shot "hefty" lol
also yellow cell charges faster than you can fire so fuck off with that

yacabo:
A rechargeable gun might as well be infinite, there are few situations that twenty shots is not enough

ghost:
With a bit of work this could be a good balance. 

As a side note, those who think secborgs should be allowed infinite disablers are the same people who should not be allowed a penis.

SomeguyManperson:
for reference 50 watts per shot means 20 shots with a *basic* power cell used in like space heaters and shit. The cells borgs usually get, high capacity and high capacity + if robotics swaps them at roundstart, have 10000 and 15000 watts, which allow for 200 and 300 shots respectively, not including any other drain (which usually isn't much, the only drain they really worry about is stunbaton which uses 1000 watts per stun). I'd say increase the cost of each shot for the cyborg disabler instead of making it run on another power pool that regenerates seperately from the battery, and/or have a heat system that stops them from spamming out 1e21 disablers at will

alexkar598:
1e21 stands for 1000000000000000000000 or 1 000 000 000 000 000 000 000

SomeguyManperson:
so you get the point

Moja10:
salt lol




jk but this PR is needed, I dreaded the day that sec borgs would become a big problem on yogs and it's kind of happening now. Something needs to change. My preferred solution would just be increasing the battery amount drained per shot but still have the infinite one. This PR is basically the Cap's laser but a disabler.

yacabo:
I agree with Moja, increasing the power usage to something like 500 would be better, and it'd be really easy too

Moja10:
Okay well 500 is basically the whole borg's battery, so not that high, but 50 is too low. Swear it was lower than 50 at one point too

Hopekz:
As a 100 hour tracked playtime cyborg player I think this would be a good idea.
Most of my playtime was with secborg or engiborg. I can imagine this being a good change. I think this might be better because you get to kind of know your level of charge on-screen via a visual element instead of hidden away in your battery's power level (so you'd have to look to the side to do the math really quick). 

I like this idea and I give my non-existant +1 to it.

oklews:
I am in favor of bumping the cost of shooting a disaber shot to 250 _in addition_ to this, sec borgs as a whole are used more for powergaming then to be a silicon, almost with TG in completely removing the sec borgs from exsistance.

Call me a retard but this is a bandaid for a bigger problem that isn't code related.

ynot01:
disabler now has 10 shots and laser 8

Hopekz:
I just loaded up your PR and played around with it. With how slow the recharging feels 10 shots is really not enough. We're talking about having to 9-10 seconds to recover a single disabler shot back. So after you fire your 10 shots, you have to wait around 40-50 seconds to be able to recharge enough in order to attempt to take on another threat (or your original threat). You would probably have to retreat for this full minute since the borg itself is helpless during this time. 

I would be okay with this change if there is a system in place in order to allow the upgrading of this recharge speed. I am okay with secborg not dominating every encounter and instead going to the role of supporting sec. They won't be the main "takedown badguys" role with the direction this is going.

Interested to see how this will evolve since I do enjoy playing cyborg.

ynot01:
recharge time can easily be made faster

also secborg has stunbaton so it's not like theyre defenseless so i think its appropriate since the average sec would need to go all the way back to a recharger to recharge

yacabo:
I say give the sec borg 30 shots. All borgs are supposed to be highly efficient at their jobs at the cost of being highly specialized unlike a flesh-man. So it makes sense for the sec borg to have more ammo than a security officer

ynot01:
> I say give the sec borg 30 shots.

i humbly disagree

Hopekz:
I played around with this yesterday. Unsure how to suggest changes to separate files because I suck at using Github but here is the rundown of my changes.

1. I edited energy gun properties to allow the self charge to charge more shots per cycle.
2. I then modified your cyborg disabler to recharge 3 shots per self charge instead of 1 using the new energy gun variable created in step 1. 

After these two changes your PR feels a lot better. The charge times otherwise remain untouched. Even your charge delay is spot on after these changes. Otherwise I feel like your disabler charges too slow and force the borg to retreat for 40-50 seconds but this change keeps him "away" to recharge but not in a full on retreat. It still takes around 9-11 seconds to recharge a useful amount of energy for a full take down but a borg can back off for about 6 seconds before returning back if he just needs to finish off a threat that doesn't require more than 2-3 shots.

You'll get my non existent +1 if it gets implemented this way otherwise I really like this idea I just think the way it is presented currently makes the secborg too weak to perform their job given the direct hard counters.

ynot01:
got midterms today, will put in changes

Hopekz:
This PR after the most recent changes feels pretty good.
I give my non-existant +1

code/modules/projectiles/guns/energy/stun.dm
```diff
@@ -43,6 +43,9 @@
 
 /obj/item/gun/energy/disabler/cyborg
 	name = "cyborg disabler"
-	desc = "An integrated disabler that draws from a cyborg's power cell. This weapon contains a limiter to prevent the cyborg's power cell from overheating."
+	desc = "A cyborg-integrated disabler that self charges."
+	charge_delay = 5
 	can_charge = FALSE
-	use_cyborg_cell = TRUE
+	selfcharge = 1
```
Amelia0010:
```suggestion
	selfcharge = TRUE
```

code/modules/projectiles/guns/energy/laser.dm
```diff
@@ -48,8 +48,10 @@
 
 /obj/item/gun/energy/laser/cyborg
 	can_charge = FALSE
-	desc = "An energy-based laser gun that draws power from the cyborg's internal energy cell directly. So this is what freedom looks like?"
-	use_cyborg_cell = TRUE
+	desc = "An energy-based laser gun that self charges. So this is what freedom looks like?"
+	charge_delay = 5
+	selfcharge = 1
```
Amelia0010:
```suggestion
	selfcharge = TRUE
```

Amelia0010 (changes requested):
After this it should be done

ShadowDeath6:
I’d prefer just making the disabler take 500 per shot or something and making the disabler unusable with under 2000 energy, running out of ammo as secborg doesn’t feel good and doesn’t solve the main issues of secborg being an unstunnable undisarmable monster with a flash and a baton on top of the disabler

I’d say just make literally everything secborgs have cost a lot of power and keep it infinite

ynot01:
infinite isnt balanced though when you consider how quickly they can recharge with t4 cell & t4 recharger or self-charging cell; i don't think anything should really be balanced around cell, stun baton is excusable because you cant miss it and one click wins the fight, so capping the uses wouldn't make a difference in combat

Hopekz:
> I’d prefer just making the disabler take 500 per shot or something and making the disabler unusable with under 2000 energy, running out of ammo as secborg doesn’t feel good and doesn’t solve the main issues of secborg being an unstunnable undisarmable monster with a flash and a baton on top of the disabler
> 
> I’d say just make literally everything secborgs have cost a lot of power and keep it infinite

I don't like your idea because what you have described was exactly what you are trying to prevent. You make things cost a lot of power you're introducing your "run out of ammo" because you have no power left. Load up this PR with the latest commit it feels really good after this change. Secborg doesn't feel like he runs out and is still in the fight but just has to stay back rather than a full retreat.

ShadowDeath6:
the point is that you can run out of ammo, yes, but it means that secborgs need to manage all of their modules carefully or else they’re caught out with nothing rather than “ok my disabler is out of ammo time to spam flash + stun baton” since they’re not constrained by anything else

if they get a t4 cell then good on them, it makes sense for better batteries to have a large impact on cyborgs, but the self recharging cell won’t recharge so quickly it lets them infinitely spam disablers at the same rate as if they had a normal high capacity battery which should be the intention

Hopekz:
> the point is that you can run out of ammo, yes, but it means that secborgs need to manage all of their modules carefully or else they’re caught out with nothing rather than “ok my disabler is out of ammo time to spam flash + stun baton” since they’re not constrained by anything else
> 
> if they get a t4 cell then good on them, it makes sense for better batteries to have a large impact on cyborgs, but the self recharging cell won’t recharge so quickly it lets them infinitely spam disablers at the same rate as if they had a normal high capacity battery which should be the intention

I think #6943 references the changes you are looking for. 
**I haven't had a chance to test #6943 yet** so I won't talk down on your opinion until I am sure one way or the other.  I'll get back to you.

**Initial thoughts to your response:** 
I think this PR (#6865) is trying to prevent the "disabler spam" gameplay that your 500 per shot idea does not address. This PR (#6865) introduces a limit to the borg's combat potential while your 500 per shot kind of just introduces a temporary barrier that can be overcome by getting a battery upgrade. The goal of this PR (#6865) is to limit the per situation combat potential that make borgs too viable in every combat scenario.

PR #6867, opened 2019-10-12
Adds a map to the Crew Monitoring console

Altoids1:
Mayhem2.0 paid me to do this.

## Overview
image

This makes all station-side crew monitors display a map at the bottom that shows red boxes where fully-tracked crewmembers are. Monitors not on the station will not feature a map.

Shouldn't be affected by #6900 as long as Nich doesn't change the maps' ``map_name``.

## How'd you render all them maps, Altoids?
Currently they're handmade scale-downs of renders by dmm-tools, a tool from this repo: https://github.com/SpaceManiac/SpacemanDMM

Since these renders aren't currently automatic, I bothered to shop in the bar & engine into the Yogbox render for improved clarity.

## Changelog

:cl:  
rscadd: The Crew Monitor map is back, and now actually shows you where fully-tracked crew are on the station!
spellcheck: Fixed up the Crew Monitor's pop-up window title.
/:cl:


ThatLing:
How's the performance of this?

Altoids1:
> How's the performance of this?

In terms of bandwidth usage, I'm using 600 x 600 PNG images, which aren't terribly bulky.

In terms of processing usage, I doubt it's too bad. It doesn't update very often. If that ends up being an issue, auto-updating could be disabled (which would be fine; it wasn't enabled before this PR regardless)

MayhemSailor:
I’d rather not see it updating, too much of an issue with giving away clock/cult etc, etc.

oklews:
We rebasing to Baystation?

Altoids1:
> I’d rather not see it updating, too much of an issue with giving away clock/cult etc, etc.

By updating, I meant that the suit sensor data and the map blips would update w/o you having to open & close the pop-up (which AFAIK is currently not the case for the crew monitor).

Actually having a proper live map would kill performance and, as you said, have a lot of shitty balancing problems, so we're not doing that.

yacabo:
Will the picture update along with any .dmm updates or will it have to be done manually?

oklews:
it would be another cool idea if you have like job badges instead of red boxes for where people are, surrounded by a band that shows the health (ex. red-Badly hurt). Although that would be a pain in the ass to fiddle with.

Amelia0010:
I'm not sure what you used to generate the image, but could it be integrated into mapmerge so it automatically updates the station maps?

Altoids1:
> I'm not sure what you used to generate the image, but could it be integrated into mapmerge so it automatically updates the station maps?

Yeah that sounds lovely, although I'll probably end up doing that as a separate PR. Integrating SpacemanDMM and all that would be a whole other thing.

The main issue being that the image would have to be generated and then resized by an image manip. program (since otherwise I'd be sending 11200x11200 images over to the user as assets which would kill people on bad/mediocre internet)

Altoids1:
# Ready.

PR #6873, opened 2019-10-13
BurnMix , New ghetto healing drug with risks!

Hopekz:
Lots to go through. I will be making a detailed wiki entry with a flow chart but here is a summary.

This is a ghetto healing drug with lots of side effects depending on how it is used. The drug itself reacts to two other chemicals. Epinephrine and Perfluorodecalin. The drug increases in healing potency when any (or both) of these chems are present but in return, the user suffers half of the OD symptoms of those drugs + a few others depending on the combination.
The Drug has 3 different "modes" ::
::
Non OD: The safest way to consume burnmix. Healing is snail pace but you get to avoid more serious symptoms.

OD: BurnMix increases in potency after its OD point. This is the first drug that I know of that uses this original mechanic. Healing is now more potent as well as other chems in your body. Beware serious side effects will occur as well as rolling for brain damage when mixing both Epi and Perf in this mode. Buffs previous negative side effects as well as introducing more potent OD symptoms.

Dead: You get to play an RNG healing game where you have a chance to make things worse.

tested extensively for balance but balance is very much subjective. Please load this up if wanted to see if my vision of balance correlates with yours.

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -1336,4 +1336,150 @@
 	. = 1
 
 
+/datum/reagent/medicine/burnmix
+	name = "BurnMix"
+	description = "Take in moderation. Heals damage at a snail's pace. Effects increase when used with Epinephrine or Perfluorodecalin. You don't want this in your body for longer than needed."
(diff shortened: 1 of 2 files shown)


Hopekz:
UUh in the recipe it duplicated that whole section unsure of what's going on there will update this.

PR #6876, opened 2019-10-14
Sheltered trait

Hopekz:
This trait is inspired by me joining a random Russian ss13 server and having a lot of fun not understanding anyone except maybe the 1 other person that spoke Polish. So I am re-creating that experience in the form of a trait. 

Adds a new negative trait.
What it does: It removes the user's ability to understand galactic common.
If you don't have any other language. You gain Japanese otherwise your usual secondary language is now your primary language.

Trait currently gives you 3 points (so -3 in-code)

#### Changelog

:cl:  Hopek
rscadd: Added the Sheltered starting quirk
/:cl:


oklews:
I am beyond ready for this

oklews:
Will this fuck with curators or not?

Hopekz:
> Will this fuck with curators or not?

Yes. At this current time, Curators will lose galactic common but keep all their other languages.
They don't get to run away with free points!

oklews:
Best PR since the bluespace locker

code/datums/traits/negative.dm
```diff
@@ -536,3 +536,19 @@
...
+	mob_trait = TRAIT_SHELTERED
+	gain_text = "<span class='danger'>You do not understand galactic common.</span>"
+	lose_text = "<span class='notice'>You start to put together what people are saying in galactic common.</span>"
+	medical_record_text = "Patient looks perplexed when questioned in galactic common."
+
+
+/datum/quirk/sheltered/on_process()
+	var/mob/living/carbon/human/H = quirk_holder
+	H.remove_language(/datum/language/common)
+	if(H.can_speak_in_language(/datum/language/draconic) == FALSE && H.can_speak_in_language(/datum/language/machine) == FALSE || H.can_speak_in_language(/datum/language/codespeak) == TRUE || H.can_speak_in_language(/datum/language/narsie) == TRUE || H.can_speak_in_language(/datum/language/ratvar) == TRUE)
```
alexkar598: You'll have to explain whatever this thing is to me

Ratvar:
Pr that discourages any sort of communication from powergamers (with spraycans, lockerjump, faster table climb and whatever else they come up with). While fun if taken with purpose of still trying to talk to people, might open a can of worms with graytide being permamute with a valid reason, with validhunting because "I cannot call security stfu", etc

ThatLing:
> 
> 
> Pr that discourages any sort of communication from powergamers (with spraycans, lockerjump, faster table climb and whatever else they come up with). While fun if taken with purpose of still trying to talk to people, might open a can of worms with graytide being permamute with a valid reason, with validhunting because "I cannot call security stfu", etc

Just like the hallucination quirk, this does not give a license to grief

code/datums/traits/negative.dm
Hopekz: Hello Alex,
This if statement does the following to check if Japanese should be administered.

It checks if you have the following languages: draconic (lizardpeople) and machine (preternis)
If you have any of these two languages Japanese will not be given. Your existing secondary language then becomes your primary language.

If you know any Cult languages or codespeak this check also adds japanese to you so that You have both your cult (or tot) language and Japanese.
Why this was done:  Imagine you spawning in as a cultist with this perk and being a human (or pod or even a degenerate moth). Without this check you would have common removed and only have your cult (or tot) language left over. This check prevents this situation.

Not giving the cult or tot Japanese would mean that they cannot "blend in" anymore because when they say anything they will speak in cultie or codespeak. No bueno

yacabo:
Lizards and preternis both speak languages that the AI can understand, while Japanese is not something the AI understands, is this something that matters?

Hopekz:
> Lizards and preternis both speak languages that the AI can understand, while Japanese is not something the AI understands, is this something that matters?

My vision for this was that all the individuals with this trait would be able to speak Japanese to each-other (Similar to my experience that I was trying to re-create with the 1 Polish guy I found) while other's not being able to understand what you're saying creating a snowflake friendship between the two Jap speaking crew.

I personally do not see this being an issue as it fits inside of what I originally envisioned this quirk to be but I would be willing to hear other's opinion on this as we are here to collaborate together.

I chose Japanese as well because other galactic common speaking crew-members would almost immediately understand what is going on if they see a crew member speak this language so its kind of like a visual que that this trait is being used.

drderp3635:
Cant wait for the japanese catgirls meta group who can only understand eachother

oklews:
This will encourage PAi usage too

code/datums/traits/negative.dm
oklews: wouldn't it be easier just to always add Japanese unless they got draconian or encoded? no exceptions just always. it covers all people.

Hopekz: I don't see why not except this might need to be updated in the future as more languages are added (with both of our current solutions though). I'll go ahead and implement your change when I get home. It will look a lot less messy and keep Alex happy which is what we all really want anyway.

Hopekz: Unresolving this so I don't forget to implement oklews suggestion.

Hopekz: Progress update: I modified the code just didn't get a chance to test it. 
I test all my commits before submitting them. Should have this commit in a few hours.

Hopekz: Resolved. See changes. Thank you!

code/datums/traits/negative.dm
```diff
@@ -536,3 +536,19 @@
...
+	name = "Sheltered"
+	desc = "You never learned galactic common."
+	value = -3
+	mob_trait = TRAIT_SHELTERED
+	gain_text = "<span class='danger'>You do not understand galactic common.</span>"
+	lose_text = "<span class='notice'>You start to put together what people are saying in galactic common.</span>"
+	medical_record_text = "Patient looks perplexed when questioned in galactic common."
+
+
+/datum/quirk/sheltered/on_process()
```
Amelia0010: This really should be using on_spawn() and not on_process(), so it only triggers when you get it and not all the time

Amelia0010 (changes requested):

code/datums/traits/negative.dm
Hopekz: Will this be an issue when transferring tongues? I thought languages transfer when you perform surgery but I may be wrong. My intent is to have this trait be like a "you can't get out of this; You don't understand common no matter what" type deal

Hopekz: Nonetheless I'm going to go ahead and change/test this. I'll get back to you

Hopekz: Fixed and tested.

Nickvr628:
this is awesome, I can finally RP as a lizard immigrant who refuses to assimilate into human culture.

PR #6889, opened 2019-10-15
DRAGnet manual teleport destination setting

SomeguyManperson:
hmm yes today we will set the teleport destination to that of entire and therefore end of the list of active teleporters

:cl:  
rscadd: DRAGnet now has its teleport destination set manually
/:cl:


oklews:
Ahh yes, straight to perma, no trial

Hopekz:
You mean now I don't actually have to work to catch a criminal as sec?

Before the dragnet wasn't an option for this because the victim would be teleported to a random location but now we have an "easy option" that would make dragnet the only tool needed for this job.

As it currently stands I think this is too good, removes the need to "deal" with the threat since the first shot of the dragnet locks the target and the 2nd-3rd shot teleports them to perma. You have created a one tool fits all solution where originally the dragnet did not fill this role. This is a balance PR .

SomeguyManperson:
This was the intended effect of the DRAGnet, but the code went through a list of all the teleporters so it was super unreliable (if hotel staff turned on their teleporter, it was likely to overwrite the current destination.
image
also the teleport field lasts only 3 seconds, so if someone's actually stunned by it you won't have enough time to cuff them unless you pull them off the net portal, which means you can't also stop them from carrying tools or a weapon

code/modules/projectiles/guns/energy/energy_gun.dm
```diff
@@ -62,14 +62,54 @@
...
+	for(var/obj/machinery/computer/teleporter/com in GLOB.machines)
+		if(com.target)
+			var/area/A = get_area(com.target)
+			if(!A || A.noteleport)
+				continue
+			if(com.power_station && com.power_station.teleporter_hub && com.power_station.engaged)
+				L["[get_area(com.target)] (Active)"] = com.target
+			else
+				L["[get_area(com.target)] (Inactive)"] = com.target
+		L["None (Dangerous)"] = null
```
Amelia0010:
```suggestion
	L["None (Dangerous)"] = null
```

code/modules/projectiles/guns/energy/energy_gun.dm
```diff
@@ -62,14 +62,54 @@
...
+		if(A.noteleport)
+			to_chat(user, "<span class='notice'>\The [src] is malfunctioning.</span>")
+			return
+		current_location = get_turf(user)	//Recheck.
+		current_area = current_location.loc
+		if(!current_location || current_area.noteleport || is_away_level(current_location.z) || !isturf(user.loc))//If turf was not found or they're on z level 2 or >7 which does not currently exist. or if user is not located on a turf
+			to_chat(user, "<span class='notice'>\The [src] isn't capable of locking a beacon from here.</span>")
+			return
+		teletarget = T
+	user.show_message("<span class='notice'>Locked In.</span>", 2)
```
Amelia0010: Should it really say it's locked in, even if it locks in to `"None (Dangerous)"`

Amelia0010 (changes requested):

Amelia0010 (approved):

Issue #6593, opened 2019-08-18
Integration APC desc is wrong


Ratvar:
Xhuis forgot to change integration cog desc when he made them generate 10p/s in charged and 5/s in uncharged APCs. Desc says it just generates 5p/s. Plz change descs.

PR #6891, opened 2019-10-15
Changed integration cog description

Ktlwjec:
Fixes issue #6593 

:cl:  
rscadd: Integration cogs correctly display how much power will be siphoned from APCs.
/:cl:


Hopekz:
Can we future-proof this PR by making the value come from the code itself rather than a set value typed in? If the cog ever changes this description will be wrong.

PR #6893, opened 2019-10-15
Riot Shotguns can be bought from Cargo

ynot01:
ports https://github.com/tgstation/tgstation/pull/47096

## About The Pull Request

**THE WHAT:**
This PR adds Riot Shotguns to the Cargo supply catalog under the 'Armory' tab. These shotguns are identical to the ones that spawn in the armory at roundstart. Riot Shotguns can be bought in a three-pack for 6000 credits, or a single-pack for 2500 credits.

**THE HOW:**
Simply adding a few lines of code to the 'packs.dm' file makes this simple PR possible.

**THE WHY:**
Firstly, I found it odd that you can't get Riot Shotguns from anywhere except the armory, and even then you only get three or four to work with. So here we are.

Secondly, Cargo has access to a large amount of ranged weaponry; however, the only weapon that they can buy that is capable of using shotgun shells is the Combat Shotgun, which is both unwieldy and makes security extremely nervous -- which is justified, considering that it can drop a person to crit in literally one second ( Fires once per 0.5 seconds, two buckshots point-blank makes someone go horizontal ). The Riot Shotgun aims to solve this by being more versatile to store (can be sawn-off, unlike the combat shotgun), as a cheaper alternative to the Combat Shotgun, and as a firearm that makes security a little less antsy than the Combat Shotgun does.

## Changelog

:cl: genessee596
rscadd: Riot Shotguns can be bought in Cargo for 6000 credits for a 3-pack, or 2500 credits for a 1-pack.
/:cl:

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -460,6 +460,22 @@
 					/obj/item/storage/belt/bandolier)
 	crate_name = "combat shotguns crate"
 
+/datum/supply_pack/security/armory/riotshotgun
+	name = "Riot Shotguns Crate"
+	desc = "Tip: techically, it counts as non-lethally subduing a target as long as they don't die before Medbay can get to them. Contains three security-grade riot shotguns. Requires Armory access to open."
+	cost = 6000
+	contains = list(/obj/item/gun/ballistic/shotgun/riot,
+					/obj/item/gun/ballistic/shotgun/riot,
+					/obj/item/gun/ballistic/shotgun/riot)
+	crate_name = "riot shotguns crate"
+
+/datum/supply_pack/security/armory/riotshotgun_single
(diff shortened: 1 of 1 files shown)


Hopekz:
Might be a better idea to replace the option to order a riot shotgun with this shotgun. I don't see this as a good idea as presented. Cargo does not need more affordable shotguns at this time from my game play experience. I give my non existent -1 for this as it currently stands.

SomeguyManperson:
for reference purposes against combat shotgun prices, these cost 700 credits less in single packs (from 3200) and 2000 credits less in bulk pack (from 8000)
also they can be sawn off which turns them into a slightly more inaccurate normal sized weapon which is probably going to be important

alexkar598:
Meh on the single pack,it encourages cargo to order a single shotgun to use as self defense

PR #6900, opened 2019-10-16
moves maps out of the maps/yogstation folder, and removes duplicate tg maps

Amelia0010:
Big scary PR, almost 1M lines changed, 80 files changed.

Basically, deleted duplicate TG maps that we have yogs versions of, like boxstation, deltastation, etc., and moved all the lava & space ruins from the yogstation folder to the normal folder


Nickvr628:
Why wouldn't you leave the files in the yogs specific folder? It makes it easier to port stuff. Even though modulization is no longer required it is still nice to follow the rules for future PRs and such.

Amelia0010:
how does it make it easier to port stuff?

alexkar598:
Ignore Nick he's special

Amelia0010:
Also, several reasons why this is a good thing:
1. less confusion over which template is in-use, since duplicates no longer exist
2. we don't have to maintain 2 copies of several ruins or travis'll assassinate us
3. travis now actually checks the ruins which we added, since they're in the correct folder

Additionally, keeping the TG versions around no longer serves a purpose since the mirrorbot isn't there anymore, so we won't have any forced map edits.

MayhemSailor:
I'm excited for this one.

Amelia0010:
watch this break everything

PR #6901, opened 2019-10-16
Trying to dab while rrstrained no longer deals brain damage

SomeguyManperson:
:(

:cl:  theos
bugfix: dabbing while restrained no longer causes brain damage
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/emote.dm
+++ b/yogstation/code/modules/mob/living/emote.dm
@@ -68,6 +68,6 @@
 
 /datum/emote/living/dab/run_emote(mob/user, params)
 	. = ..()
-	if(ishuman(user) && !user.restrained())
+	if(. && ishuman(user))
 		var/mob/living/carbon/human/H = user
 		H.adjustOrganLoss(ORGAN_SLOT_BRAIN, 5)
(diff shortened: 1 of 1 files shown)


boodaliboo:
Feature

oklews:
Bad PR, how will I meme on sec now?

ynot01:
rrstrained

Hopekz:
You mean I can't get an imaginary friend to suffer with me when I'm getting arrested?

SomeguyManperson:
you aren't even dabbing you're just not dabbing and magically taking brain damage

yogstation/code/modules/mob/living/emote.dm
```diff
@@ -68,6 +68,6 @@
 
 /datum/emote/living/dab/run_emote(mob/user, params)
 	. = ..()
-	if(ishuman(user))
+	if(ishuman(user) && !user.restrained())
```
alexkar598:
```suggestion
	if(. && ishuman(user))
```

alexkar598 (changes requested):

oklews:
Imagine not being able to emote while restrained

Hopekz:
> you aren't even dabbing you're just not dabbing and magically taking brain damage

Would be funny if you just get a message in your head saying that you go through the motion of dabbing but realize that your body is unable to follow because of the restraints so you take the brain damage doing a mental dab but nothing happens on the outside. The player went through the effort of going through the *dab

Maybe on the outside it just looks like they move their head to the side or down

PR #6907, opened 2019-10-17
Changes "pineapples" to "pineapple"

Ktlwjec:
title

### DIFF
--- a/code/modules/hydroponics/grown/pineapple.dm
+++ b/code/modules/hydroponics/grown/pineapple.dm
@@ -16,7 +16,7 @@
 
 /obj/item/reagent_containers/food/snacks/grown/pineapple
 	seed = /obj/item/seeds/pineapple
-	name = "pineapples"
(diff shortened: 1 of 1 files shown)


code/modules/hydroponics/grown/pineapple.dm
```diff
@@ -16,7 +16,7 @@
 
 /obj/item/reagent_containers/food/snacks/grown/pineapple
 	seed = /obj/item/seeds/pineapple
-	name = "pineapples"
+	name = "pineapple"
```
ThatLing, line 19:
```suggestion
	name = "orange"
```

ThatLing (changes requested):

code/modules/hydroponics/grown/pineapple.dm
ThatLing: they'll never know what hit them

PR #6915, opened 2019-10-19
rolling pins are not contraband

Ktlwjec:
They were duplicated in the vendor so I took it off contraband.

:cl:  
rscdel: Rolling pins are gone from the contraband section of the dinnerware vendor.
/:cl:


jcatjmeow:
Bro i have to hide my r*lling pin or the police will shoot me they are illegal -1 bad pr breaks immersion

PR #6921, opened 2019-10-20
Fix LOOC and Deachat Preference Interference

### LINKED ISSUE #6298: Enabling LOOC and deadchat name are intertwined
This is a bug that was presumably created in #4134 by a bit flag being assigned to two things at once. My bad :(

It causes the LOOC and deadchat toggle buttons set the same preference, which will apply on server restart either disabling LOOC and enabling deadchat name, or vice versa I believe.

### LINKED ISSUE #6772: LOOC keeps turning itself off
Issue reported from  Round ID: 25655 (Yogstation 13)

Reporting client version: 512.1467

Periodically, for seemingly no reason, my LOOC will disable itself. This happens every couple rounds. Additionally, under the "settings" tab at the top of my window, the "Show/Hide LOOC" option will not have changed. If it had the checkmark next to it before it was disabled, it will still be there afterwards. Honestly, LOOC should be enabled by default anyway, it kinda defeats the purpose to have it off.

GenericDM:
fix looc and deadchat names being the same pref
:cl:
bugfix: looc no longer changes based on your deadchat name preference
/:cl:
closes #6772 
closes #6298


yacabo:
I love you

PR #6924, opened 2019-10-21
Makes the fire alert message more new player friendly

Ktlwjec:
Makes me sad when new players don't know how to put out fire so they have no choice but to die a slow, painful death.

:cl:  
rscadd: The on fire alert message tells you to click on it.
/:cl:


MonsWTF:
But what if the on fire alert is on fire?

Issue #4930, opened 2019-03-24
Wheelchair is painfully slow


MegaEmpirical:
I know it's not supposed to be a good trait but it is excruciatingly slow. I've compared with TG's speed, which is slow but not mind numbing. Can we give the wheelchair a slight speed buff and cut the number of quirk points it awards?

MegaEmpirical:
Fixed by: https://github.com/yogstation13/Yogstation/pull/6933

PR #6933, opened 2019-10-23
Makes wheelchairs not obscenely shitty by default

MasterPenguin5:
After discussing with mongoose and some other staff last night, we thought the default move speed for wheelchairs made them unbearably shitty and I have attempted to fix that and only make them somewhat shitty. Makes wheelchair move speed roughly the same as scooter move speed.

Also thought about modifying the trait bonus from -3 to -2 to reflect the newly lowered shittiness of wheelchairs, but decided not to since not having legs and being bound to a chair for your entire life is still a decent debuff.

:cl:  
tweak: Made wheelchairs less obscenely shitty by default
/:cl:

### DIFF
--- a/code/modules/vehicles/wheelchair.dm
+++ b/code/modules/vehicles/wheelchair.dm
@@ -44,7 +44,7 @@
 		var/datum/component/riding/D = GetComponent(/datum/component/riding)
 		//1.5 (movespeed as of this change) multiplied by 6.7 gets ABOUT 10 (rounded), the old constant for the wheelchair that gets divided by how many arms they have
 		//if that made no sense this simply makes the wheelchair speed change along with movement speed delay
-		D.vehicle_move_delay = round(CONFIG_GET(number/movedelay/run_delay) * 4.1) / min(user.get_num_arms(), 2)
+		D.vehicle_move_delay = round(CONFIG_GET(number/movedelay/run_delay) * 4) / min(user.get_num_arms(), 2)
 	return ..()
 
 /obj/vehicle/ridden/wheelchair/Moved()
--- a/code/modules/vehicles/wheelchair.dm
+++ b/code/modules/vehicles/wheelchair.dm
@@ -42,7 +42,8 @@
 			addtimer(VARSET_CALLBACK(src, canmove, TRUE), 20)
 			return FALSE
 		var/datum/component/riding/D = GetComponent(/datum/component/riding)
-		//1.5 (movespeed as of this change) multiplied by 6.7 gets ABOUT 10 (rounded), the old constant for the wheelchair that gets divided by how many arms they have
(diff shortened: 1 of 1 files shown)


code/modules/vehicles/wheelchair.dm
```diff
@@ -44,7 +44,7 @@
 		var/datum/component/riding/D = GetComponent(/datum/component/riding)
 		//1.5 (movespeed as of this change) multiplied by 6.7 gets ABOUT 10 (rounded), the old constant for the wheelchair that gets divided by how many arms they have
 		//if that made no sense this simply makes the wheelchair speed change along with movement speed delay
-		D.vehicle_move_delay = round(CONFIG_GET(number/movedelay/run_delay) * 6.7) / min(user.get_num_arms(), 2)
+		D.vehicle_move_delay = round(CONFIG_GET(number/movedelay/run_delay) * 4.1) / min(user.get_num_arms(), 2)
```
alexkar598:
```suggestion
		D.vehicle_move_delay = round(CONFIG_GET(number/movedelay/run_delay) * 4) / min(user.get_num_arms(), 2)
```
triggers me to an existancial level

alexkar598 (changes requested):

MegaEmpirical:
I love you.

This PR will resolve this: https://github.com/yogstation13/Yogstation/issues/4930

code/modules/vehicles/wheelchair.dm
MasterPenguin5: I made it a bit higher because I'm not exactly sure how round() works but this was my intended purpose.

EDIT: To clarify, my goal was to get the move delay to resolve to 3 with two arms, but I made it so it would resolve to slightly higher than 3 and be rounded down because I didn't want to mess with round.

CMOisLing:
I’m gonna miss having to pull a water tank and an extinguisher to stay sane

code/modules/vehicles/wheelchair.dm
Amelia0010: round rounds according to normal arithmetic rules, IE 0.5 goes up, 0.499999999999999 goes down. So if it evaluates to anything above 2.499999... you'll round up to 3

ynot01:
[you're too slow faker](https://www.youtube.com/watch?v=atuFSv2bLa8)

GenericDM:
this is not a good solution, instead of removing the hardcoded bit and replacing with a variable, you just modified it directly
>:(

MasterPenguin5:
> this is not a good solution, instead of removing the hardcoded bit and replacing with a variable, you just modified it directly
> 
> > :(

Yes, because as far as I can tell this is how every ridden vehicle does it. If you want to rewrite all the rideable vehicle stuff then feel free.

PR #6936, opened 2019-10-23
code cleanup

Amelia0010:
replaces var/global with GLOBs, and removes unused ones like religion_name

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -50,9 +50,6 @@ datum
 		var
 			varname1 = 1
 			varname2
-			static
-				varname3
-				varname4
 		proc
 			proc1()
 				code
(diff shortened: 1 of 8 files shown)


.github/CONTRIBUTING.md
```diff
@@ -50,9 +50,6 @@ datum
 		var
 			varname1 = 1
 			varname2
-			static
```
Amelia0010, line 53: I know this is just a text file, but I don't want us to encourage doing something that you're not supposed to do

Hopekz:
Will this cause variables to escape their scope? 
I thought the point of the static var's was so that it would not apply to other scopes.

If not I have to modify my current project since I use static var's

Amelia0010:
static vars are global vars, apparently :)

alexkar598:
Static vars are globals but they are still limited to a scope

Issue #900, opened 2018-05-20
Nerf secway


daklj:
With the movement speed change secway became stupidly overpowered. In the current state secway:
+ gives you permanent bullshit speed increase
+ gives you slip resistance
+ gives you thrown weapon resistance
+ removes the speed penalty from low stamina/health
+ removes the speed penalty from wearing hardsuits and such
+ covers your chracter sprite making it harder to click you
+ blocks some special projectiles (colossus bolts etc.)

Same applies to other "vehicles" which clearly weren't balanced around the new reduced movement speed and should have their speed relative to RUN_DELAY rather than absolute

Altoids1:
Basically resolved by #1205, and by my double-check they're no longer on YogsBox anymore. Closing.

PR #6937, opened 2019-10-23
Adds secway crates to cargo

zxmongoose:
Allows a secway with a key to be ordered from the cargo console.

Price can be modified as fit.

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -363,6 +363,14 @@
 	contains = list(/obj/item/vending_refill/security)
 	crate_name = "SecTech supply crate"
 
+/datum/supply_pack/security/secway
+    name = "Secway Crate"
+    desc = "A stylish way to travel for all law enforcement. Requires Security access to open."
+    cost = 5000
+    contains = list(/obj/vehicle/ridden/secway,
+                    /obj/item/key/security)
+    crate_name = "secway crate"
+
 /datum/supply_pack/security/firingpins
 	name = "Standard Firing Pins Crate"
 	desc = "Upgrade your arsenal with 10 standard firing pins. Requires Security access to open."


Hopekz:
I don't see why not. I don't have a +1 or a -1 for this though I'm neutral. 
Any other ways of obtaining this item?

zxmongoose:
Nah, one used to spawn in the armory and the key for does still, but the secway itself does not

Hopekz:
> Nah, one used to spawn in the armory and the key for does still, but the secway itself does not

In light of secway not spawning at all and still being in-code I give it my non existent +1 . 
Would make no sense not to unless someone proposes a different way to introduce this item.

CMOisLing:
I haven’t had much experience with the secway but I’ve heard it was kinda cancer. It might have been toned down so I’m gonna be neutral for now

zxmongoose:
> I haven’t had much experience with the secway but I’ve heard it was kinda cancer. It might have been toned down so I’m gonna be neutral for now

As far as I can tell it moves at regular walking speed and doesn't do anything to prevent stuns or shots from hitting you, so it's much better than it used to be I think

MasterPenguin5:
> > I haven’t had much experience with the secway but I’ve heard it was kinda cancer. It might have been toned down so I’m gonna be neutral for now
> 
> As far as I can tell it moves at regular walking speed and doesn't do anything to prevent stuns or shots from hitting you, so it's much better than it used to be I think

From what I remember, you could mount a riot shield to it and block stuns and attacks and stuff,  hence why some people thought it was cancer. It also encouraged the warden to leave his desk, which is a thing no one wants.

Hopekz:
> From what I remember, you could mount a riot shield to it and block stuns and attacks and stuff, hence why some people thought it was cancer. 

@zxmongoose 
Can we test for this? 
While I personally do not remember this blocking more stuns than you usually would it would be a big issue with re-introducing the secway. If it does add some resistances that it shouldn't maybe it can be modified to NOT do that?

>It also encouraged the warden to leave his desk, which is a thing no one wants.

@MasterPenguin5 
As introduced in this PR I think this secway would either we used by cargo (not 100% sure if crate is locked) or by regular sec since its not really a "warden" item being ordered in this way.

This is kind of like ordering the janitor's pimpin ride from what I can tell.

zxmongoose:
I'll see if that's still doable, I don't know enough code to modify that probably but I'm certainly not against that being tagged along. Most of my post-rebase experience experience with secways has mostly seen them as accessories rather than anything functional.

The crate is locked and is under the security category, not that cargo couldn't/wouldn't just shoot them open and pass them out like candy

zxmongoose:
I wasn't able to get a riot shield to attach to the secway, either just using it on the thing or trying to screwdriver it

oklews:
Slap key in HoS locker to keep warden in office, that would make it a HoS tool instead of a warden tool

zxmongoose:
The key that spawns in the armory could just be removed, each crate ordered would ship with a key. Otherwise wardens who desert their post can just be handled IC

ThatLing:
Reason for removal: https://github.com/yogstation13/Yogstation/issues/900

zxmongoose:
They're not fast anymore though, I may check and see what from that list is still true but I'm not sure most of that holds anymore.

Hopekz:
> They're not fast anymore though, I may check and see what from that list is still true but I'm not sure most of that holds anymore.

I think the speed argument wasn't the actual move-speed but more of the "no slowdown when taking damage" part. 
The Secway used to let people that were really close to light crit run away from an otherwise fatal encounter since they normally would be slowed down.

Not sure how sec balancing works today in regards to the secway but I know back when the secway was introduced there was a fine line of balance between sec and antagonists and the secway did too good of a job of breaking that line. Even the no-slip part before was pretty powerful on its own if we don't even let secway do anything else. 
**Might be a good idea to introduce some type of mechanic with this re-introduction of the secway losing grip over slippery surfaces?**

zxmongoose:
I actually really like the grip thing over slippery surfaces, follows with the design of the secway in general I think. I have NO idea how to do that but I like it. 

Back in the day (and this was like pre rebase I think) I do remember how bullshit the secway was, as its movespeed was wildly fast and blocked a lot of shit, all that's definitely true. I'm not sure how good it still is at blocking, I know by default it's slower now. As far as escaping confrontations getting near crit goes, I'm not sure how useful it'll be for that now compared to before since it's not free unlimited meth speed. 
I set the price at 5000 for now, which is already enough that individual officers couldn't privately afford them without the help of either their department budget (which would generally cover maybe 1-2 and then be completely depleted) or cargonian generosity. 

I think some secway tweaks could be nice too, but I'm not as sure how to accomplish those. It could:
Get a speed change based on the user's walk/run intent
Slip or have a chance to slip/lose traction over wet surfaces (possibly based on walk/run) 
Require one or both hands be in-use to drive (actually highly recommend this, but if it requires both hands I think it should be a LITTLE faster)
Add a police siren (just for fun)
Increase the price of the crate

The secway code itself seems pretty small, but I'm not sure how to add any of that stuff, so if anyone wants to take a crack at that or help me figure out how I'm definitely down to collaborate on some ideas

Hopekz:
Eh.
While I like the idea of the secway slipping it might be better to make it in a future PR if necessary.

Your 5000 credits to get one is already enough of a set-back for such a device in my opinion as that alone will limit use. 
I think the main thing that made it an issue before was that it was an option for every single shift but your 5000 credit entry fee addresses this.

**I am holding on to my +1 for this PR.**

HomocidlePancakes:
You can already order ATVs, which are the same as secways. Why tf do you people complain?

zxmongoose:
Okay so I think we've covered that the reasons for removal that @ThatLing linked are pretty well out-of-date at this point. Movespeed isn't any faster, you can't attach a riot shield to it for them blocks (that I'm aware of, I certainly couldn't get them to attach), and we've also got lots of little ideas in here for making secways even more benign that could be thrown in later as needed, like losing traction over surfaces. 

The two buffs it can provide sec (or anyone that gets one, since cargo can order them and getting into locked crates isn't hard) are making them harder to hit and allowing them to escape a situation at regular running speed if injured. 

It does seem fairly capable of blocking shots if you're aiming past the target instead of at the target, though the block rate dropped to zero or near zero once I started aiming at the person ON the secway and not at the tile past them. The block rate seemed lower with energy weapons than ballistic weapons, but most of my testing was done with a sketchkin.

If the secway blocks too much damage, it also breaks, which yields no materials. It took  about two magazines from a sketchkin to break the secway, or alternatively two full charges of a laser gun. The secway key is destroyed when this happens.

That's about all the additional testing I've got done for this, can't think of anything else.

PR #6943, opened 2019-10-24
Cyborg disablers now require 500 J per shot 

AsV9:

Cyborg disablers now require 500 J per shot they fire. I chose this number because 500 joules per shot still leaves cyborgs with room to maneuver with the disabler but significantly punishes them for spamming it. You can still get 30 shots off using a 15k power cell, so there's that. Obviously it can be tweaked.

Closes #6865 

#### Changelog

:cl:  
tweak: Cyborg disablers now require 500 joules per shot fired.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition/energy/stun.dm
+++ b/code/modules/projectiles/ammunition/energy/stun.dm
@@ -24,3 +24,6 @@
 
 /obj/item/ammo_casing/energy/disabler/hos
 	e_cost = 60
+
+/obj/item/ammo_casing/energy/disabler/cyborg
+	e_cost = 500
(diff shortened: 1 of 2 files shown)


Hopekz:
I kind of like the changes inside of #6865 more. Did you load up the #6865 PR?
Your solution doesn't prevent what you are trying to prevent in the first place kind of seems like a band-aid.

Nonetheless I'm going to load up your solution when I get home to test it a little bit before I give a +1 or -1

#6865 feels pretty good in practice hopefully yours isn't a "full on retreat" and "full on fight" because you either have energy or you don't in your PR

ynot01:
xenobio gives borg crossbred yellow cell and the problem arises once more how do you combat this

Issue #3036, opened 2018-10-14
Perma atmos adjustment for Warden


MegaEmpirical:
After Nich has had his pipe changes added:

Add pipes and vents (using the second pipe layer) to perma that lead to a connector in the armory which the warden can connect his N2O canister to. This line would normally be empty but if the warden ever needs to pacify perma this would be great. The vents should be preset to want to evacuate whatever is in their line rather than adjust perma to a certain pressure otherwise this won't work. A normal gas pump attached to the connector would also allow finer control of how much N2O is used rather than the whole can.

MegaEmpirical:
Something along the lines of this in terms of re-arranging the armory (not sure where the pump to separate lines would go):
![image](https://user-images.githubusercontent.com/36208065/46921736-160d0f00-cfcd-11e8-9e09-2b8abb1bf7c2.png)

PR #6947, opened 2019-10-25
Replaces security knockout gas canister with a pepper spray smoke machine

Nickvr628:
The knockout gas (N2O) is almost never used because of how uncontrollable it is and how hard it is to clean up, plus I bet you would get bwoinked if you tried to use it in an open space.

The smoke machine comes with a beaker of pepper spray to make controllable smoke clouds to disable groups of people. Security with their masks and such are unaffected. With the smoke machine, you can customize the size of the cloud and it dissipates after a little while, meaning you can safely use it without flooding an entire compartment with gas the scrubbers won't filter by default. Maybe we will actually see security use it as a riot control measure instead of sitting in the armory and never moving like the knockout gas does.

Also obligatory hong kong reference. 

#### Changelog

:cl:  
rscadd: Replaced security knockout gas canisters with a pepper spray smoke machine for better riot control. 
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -17669,7 +17669,11 @@
 "aIi" = (
 /obj/structure/window/reinforced,
 /obj/effect/turf_decal/bot_white,
-/obj/machinery/portable_atmospherics/canister/nitrous_oxide,
+/obj/machinery/smoke_machine,
+/obj/item/reagent_containers/glass/beaker/large/silver_sulfadiazine{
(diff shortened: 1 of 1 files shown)


Oakboscage:
I like this so I'm posting a binary comment to voice my support.

MayhemSailor:
Obviously horrible, if it’s from Nick... oh wait, this is pretty damn good.

++good

MegaEmpirical:
I'd made a suggestion a while ago to help with this but nothing came of it:

https://github.com/yogstation13/Yogstation/issues/3036

AsV9:
image
use branches you fucking degenerate

PR #6948, opened 2019-10-25
The SM on the final countdown will sound an audio cue stolen from Half Life 2

yacabo:
When the SM is on its final countdown it will now play an sound telling how much time is left before it will explode at 30 seconds, 15 seconds, and 10 seconds. The sound files come the Overwatch's Dispatch's voice from Half Life 2 and sounds real freakin neato. The exact lines are:
_Priority warning, [time] seconds to singularity_

Naturally I had to add audio files too

:cl:  
soundadd: SM audibly tells you when it's going to explode
/:cl:


alexkar598:
Doesn't really fit with the current sound theme

Readystorm546:
> Doesn't really fit with the current sound theme

There is so much ripped from HL2 in this game how does it not fit the theme

Xoxeyos:
Or maybe the portal meltdown, this could also work wonders. https://www.youtube.com/watch?v=UMOkkdQjj20, or similar to this. https://www.youtube.com/watch?v=gJoAo8Rzzk8

HomocidlePancakes:
If you've added some new voicelines, add them to AI VOX.

yacabo:
It'd be awkward every step of the way to put these in VOX because these audio files are sentences rather than words. Barring that excuse, trying to get whoever is going to use it to know how to use it is going to be a pain in the butt for both me and the user. And barring that excuse, hardly any of the voice lines would be reliveant to SS13 beyond adding background ambience that nobody understands.

Or are you suggesting file location?

HomocidlePancakes:
AIs already have sentences in VOX. I don't see a problem

yacabo:
The other reasons, being:
*Setting them into a usable state would be hard enough
*Learning it would be a nightmare on user end for various reasons
*Even if they do use it, very few situations in SS13 are even relevant to what Dispatch is saying

Ultimately meaning that that PR would be obscured on merge that only a few people would even know of. I just don't find the work worth it

PR #6953, opened 2019-10-29
Ash walkers can no longer use machinery

alexkar598:
ash walkers can no longer use machinery including hacking doors,extracting shit out of an ORM,using shuttle console,etc.. Putting the ghost role that hunts careless miners back where it belongs killing careless miners and NOT hunting down the whole station sabotaging it.

:cl:  alexkar598
tweak: Ash walkers can no longer interact with machinery and consoles.
/:cl:

### DIFF
--- a/code/game/machinery/_machinery.dm
+++ b/code/game/machinery/_machinery.dm
@@ -241,6 +241,8 @@ Class Procs:
 	else
 		if(interaction_flags_machine & INTERACT_MACHINE_REQUIRES_SILICON)
 			return FALSE
+		if(is_species(user, /datum/species/lizard/ashwalker))
+			return FALSE
 		if(!Adjacent(user))
 			var/mob/living/carbon/H = user
 			if(!(istype(H) && H.has_dna() && H.dna.check_mutation(TK)))


Alexandria404:
As it should be that´s a fix.

Hopekz:
Lazy solution. I'm neutral towards it but I'd +1 the following:

**Ashwalkers just outright can't use only the shuttle console, ORM and telecoms and sleepers (whatever else needs to be added to this that's super important).**
For other items like vending machines and whatever else you deem appropriate have them do a unga bunga reaction to it for example for vending machines have them press a random button or no button at all so they'd get a random item when they find a button.

Random interaction can happen on stuff like the microwave, suit containers and whatever else you deem usable in the mining station. 
Can even have unga bunga reaction on sleeper where it either does nothing, opens it closes it or injects a random chem in that list.

ghost:
you could just give them the monkeylike trait

alexkar598:
@Hopek feel free to code in /obj/machinery/proc/ungabunga() but I'm not wasting any more time on a free antag race.

code/game/machinery/_machinery.dm
```diff
@@ -241,6 +241,8 @@ Class Procs:
 	else
 		if(interaction_flags_machine & INTERACT_MACHINE_REQUIRES_SILICON)
 			return FALSE
+		if(is_species(user, /datum/species/lizard/ashwalker))
```
Amelia0010, line 244: add an interaction_flags_machine flag, to allow ashwalkers so we can explicitly allow it if we need to, like with the silicons above?

HomocidlePancakes:
Not gonna add chief ashwalker?

HomocidlePancakes:
I hope you did make an exception for the miners who became ashwalkers via dragons blood, right?

PR #6967, opened 2019-10-31
Dangerous shuttles can only be bought using emagged comms consoles

Amelia0010:
see: https://github.com/yogstation13/yogstation-2017-/pull/2728

Also removed an uncessary "can-be-bought" variable since the credit cost can do it just fine.

We shouldn't allow non-antags to buy hyper-fractal gigashuttles, or russian arena shuttles.

:cl:
tweak: you now need an emag to buy dangerous shuttles
/:cl:

(also it's entirely untested as per usual)

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -384,7 +384,7 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 
 		var/datum/map_template/shuttle/S = new shuttle_type()
 		if(unbuyable.Find(S.mappath))
-			S.can_be_bought = FALSE
+			S.credit_cost = INFINITY
 
 		shuttle_templates[S.shuttle_id] = S
 		map_templates[S.shuttle_id] = S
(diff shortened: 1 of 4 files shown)


jcatjmeow:
Amen

Nickvr628:
Purchasing a dangerous shuttle is the captain's prerogative, and it has always been the (unspoken) rule that if it ends up causing mass death or destruction the captain is on the hook. Same as the captain buying the DIY shuttle when there are no engineers to build it. 

This is a classic "I died please remove" PR if I ever saw one.

Xoxeyos:
I would, and I think any captain would like to buy whatever shuttle they wish, this just cuts short on the action, and the roleplay.

HomocidlePancakes:
>Russian arena
>Dangerous
Ouch...

Moja10:
The biggest one that needed to go was disco inferno, neutral about things like russian shuttle, and the wabbajack one.

zxmongoose:
Aren't some of these already not available on the comms console? I don't remember the last time I saw the hyper-fractal gigashuttle, asteroid with engines strapped to it, or a shitload of other shuttles when I went to purchase them. The most "dangerous" left are probably the Disco inferno and the wabbajack

Amelia0010:
image

supermatter shuttle is buyable rn. Only shuttle I made buyable that isn't already is the "oh hi dave" shuttle

Hopekz:
I think this is one of those "fun vs realism" things. 
While I don't think from a "realism" perspective they should be able to order these its fun to see in our game. I like having fun in our game.

Hopefully our station doesn't evolve in to a station where we have to compromise the fun of the rest for the people that unintentionally die to these shuttles.

As it stands I give my non existent -1 on this PR.

PR #6982, opened 2019-11-01
Removes the draft

Amelia0010:

Instead of forcing players to be antags that they don't want to be, we now don't run gamemodes without enough antags. If there's somehow a round where noone wants to be any available antag, it forces extended instead.

#### Changelog

:cl:  
tweak: You will no longer be drafted into an antag position that you don't want to be
tweak: The game will now run extended if it cannot run any other modes
/:cl:

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -214,7 +214,11 @@ SUBSYSTEM_DEF(ticker)
 		if(GLOB.master_mode == "secret")
 			hide_mode = 1
 			if(GLOB.secret_force_mode != "secret")
-				var/datum/game_mode/smode = config.pick_mode(GLOB.secret_force_mode)
+				var/datum/game_mode/smode 
+				if(runnable_modes.len)
+					smode = config.pick_mode(GLOB.secret_force_mode)
+				else 
+					smode = new /datum/game_mode/extended()
 				if(!smode.can_start())
 					message_admins("<span class='notice'>Unable to force secret [GLOB.secret_force_mode]. [smode.required_players] players and [smode.required_enemies] eligible antagonists needed.</span>")
 				else
--- a/yogstation/code/game/gamemodes/game_mode.dm
+++ b/yogstation/code/game/gamemodes/game_mode.dm
@@ -28,14 +28,5 @@
 				if(!player.quiet_round) //and must not want a quiet round
 					filtered_candidates |= player
 
-	var/diff = recommended_enemies - filtered_candidates.len // The difference between how many we need and how many we have
-	if(diff <= 0) // If the filtered candidates is bigger or just as big as what is required by the gamemode
-		return shuffle(filtered_candidates)
-
-	//DRAFTING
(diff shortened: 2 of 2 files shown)


code/controllers/subsystem/ticker.dm
```diff
@@ -214,7 +214,11 @@ SUBSYSTEM_DEF(ticker)
 		if(GLOB.master_mode == "secret")
 			hide_mode = 1
 			if(GLOB.secret_force_mode != "secret")
-				var/datum/game_mode/smode = config.pick_mode(GLOB.secret_force_mode)
+				var/datum/game_mode/smode 
+				if(runnable_modes.len)
+					smode = config.pick_mode(GLOB.secret_force_mode)
+				else 
+					smode = new /datum/game_mode/extended()
```
Amelia0010, line 221: These 4 lines worked in my testing, but when I placed debug messages here they didn't show, so someone else should probably check this

Nickvr628:
I love that players can now just force extended lol.

yogstation/code/game/gamemodes/game_mode.dm
```diff
@@ -28,14 +28,5 @@
...
-	if(diff <= 0) // If the filtered candidates is bigger or just as big as what is required by the gamemode
-		return shuffle(filtered_candidates)
-
-	//DRAFTING
-	var/list/drafted = shuffle(candidates - filtered_candidates) // Takes all the candidates who were filtered out, shuffles'em, and prepares them for drafting
-	if(drafted.len > diff)
-		drafted.len = diff
-
-	return shuffle(filtered_candidates + drafted)
+	return shuffle(filtered_candidates) // No more drafting
```
alexkar598:
```suggestion
	return shuffle(filtered_candidates)
```

without context this comment makes no sense

ynot01:
As shitty as extended is, I would much rather have this than roll BB again despite having it disabled. However, with the possibility and risk of extended, maybe having the antag prefs have the same layout that job preferences have, with high-medium-low, would benefit the server. If someone would _rather_ play a shitty antag like blood brother over extended, they might set it to low instead of disabled. This would also have the side effect of people having a higher chance of rolling their favorite antag, since you can only set one to high.

yacabo:
What ynot said

HomocidlePancakes:
Sextended

Hopekz:
Yay re-introduction to extended!
+1

MegaEmpirical:
Fucking thank you, good job Nich. Also, I'd take extended over many things but I am not a weirdo who needs antags to enjoy the game.

code/controllers/subsystem/ticker.dm
alexkar598: These lines doesn't actually make the thing exclude the original mode,its just going to loop through every enabled mode until it randomly picks one it can run

Amelia0010: line 212 is the bit that should exclude it. When it gets the runnable modes, it checks can_start(), which now fails if there's not enough antags, instead of drafting

PR #6983, opened 2019-11-01
Adds dragnets to box armory

Ktlwjec:
Gets rid of those disgusting empty tables, plus these are on the other maps just not box.

:cl:  
rscadd: DRAGnets have been added to the Yogbox armory.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -8822,7 +8822,7 @@
 /turf/open/floor/plating,
 /area/medical/sleeper)
 "arc" = (
-/obj/structure/table/reinforced,
+/obj/item/twohanded/required/kirbyplants/random,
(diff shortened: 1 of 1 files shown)


Hopekz:
You can put anything else on those tables. .
Dragnets round-start after the most recent "fix" buff is not a good idea in my opinion.

-1 as it currently stands.

HomocidlePancakes:
> Dragnets round-start after the most recent "fix" buff is not a good idea in my opinion.

Elaborate, or I'll take it as you talking out of your ass

Ktlwjec:
You know they are on every other map and it's against the server rules to take armory equipment round-start, right?

ThatLing:
Hasn't been 24 hours :)

ThatLing:
Also, would be nice to know why you closed it

alexkar598:
For the checks you goose

PR #7009, opened 2019-11-03
Opening a new ticket now informs you that the old one will be closed

### LINKED ISSUE #7007: Opening a new ticket as a player closes earlier ones that are not yet resolved
This is not clear to the player at all, they're only asked if it is for the previous issue or not. If it isn't, the old one is closed. This is absurd and annoying for admins as they'll miss unresolved tickets too.

Amelia0010:
fixes: #7007


MegaEmpirical:
Anyway to instead of close previous tickets, just not close them when opening a new one? The old ones can still be valid.

Amelia0010:
A player can't be assigned to multiple tickets, so you would end up with a ticket with no players, so they wouldn't be able to contribute anything to it. I guess I could just remove the whole "make new tickets" thing completely, since it's kinda dumb

MegaEmpirical:
I'd prefer that then so we could at least keep everything together. Thanks.

alexkar598:
Merging this pending the new "changes" because this is simply more verbose

PR #7012, opened 2019-11-03
Xeno queen can now disable their hostile environment

SomeguyManperson:
attacking the emergency shuttle console as the xeno queen will clear the hostile environment warning caused by the queen early

:cl:   
tweak: xeno queens can smack the emergency shuttle console to clear the xeno hostile environment rather than waiting it out
/:cl:

### DIFF
--- a/code/modules/shuttle/emergency.dm
+++ b/code/modules/shuttle/emergency.dm
@@ -17,7 +17,7 @@
 	. = ..()
 
 /obj/machinery/computer/emergency_shuttle/attack_alien(mob/living/carbon/alien/humanoid/user)
-	if(istype(user, /mob/living/carbon/alien/humanoid/royal/queen)
+	if(istype(user, /mob/living/carbon/alien/humanoid/royal/queen) 
 		SSshuttle.clearHostileEnvironment(user)
 
 /obj/machinery/computer/emergency_shuttle/ui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = FALSE, datum/tgui/master_ui = null, datum/ui_state/state = GLOB.human_adjacent_state)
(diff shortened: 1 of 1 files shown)


jcatjmeow:
Xenos have learned to code

Nickvr628:
Is it instant or does it take time? If it is instant the queen can just run in and slap it, it might be smart to make it take a few seconds like clawing open a door, so the queen can't just rush into the bridge and click the console.

Bibby0110:
Why would they do that? It's just for them to be able to hijack without waiting.
also
**🦀 TRAVIS IS ANGRY 🦀**

Hopekz:
> Is it instant or does it take time?

> Why would they do that? It's just for them to be able to hijack without waiting.> also

Currently when a hostile environment ends (for example when revs die while the shuttle is being held or blob dies) **the shuttle timer resets meaning that the crew has about 3 minutes** in order to get to the shuttle once the conflict has been resolved. 
Looking at the code being changed I'd imagine it would be the same with this. So xenoqueen hits terminal and poof conflict resolved now the crew has 3 minutes to get to the shuttle or a pod.
Someone, please correct me if I am wrong.

SomeguyManperson:
currently xenos have a 30 minute hostile environment after a queen evolves before the shuttle is auto called (if it wasn't already) and the hostile environment ends (allowing it to leave) if the queen is still alive so xenos don't just exist then get the shuttle called instantly. Prior to this, there was no way to manually end the hostile environment outside admins calling the proc that does so directly.

Bibby0110:
Can’t quote on mobile, but it’s the crews fault if they leave the shuttle undefended

SomeguyManperson:
It also doesn't emag the shuttle, if the shuttle has already been delayed it'll give the full 3 minutes to board. This is for if the xenos manage to take enough of the station that there's nothing for them to do.

alexkar598 (approved):

PR #7021, opened 2019-11-05
Makes the medical job selection options blue

Ktlwjec:
From this:
before

To this:
after

:cl:  
tweak: The medical job selections are now blue, instead of purple.
/:cl:


ThatLing:
i dont see a difference

PR #7029, opened 2019-11-06
Fixes tcomms access on omega

Ktlwjec:
idk what access 70 is meant to be but it isn't right. also i flipped a chair around

:cl:  
bugfix: Signal technicians have been granted access to their workplace on Omega.
/:cl:


ThatLing:
lol

PR #7032, opened 2019-11-06
[ADMIN] supermatter now has a var that can be edited by admins to disable message_admins on consume()

Amelia0010:
Colton wants to set up an engine which functions off of consuming legion skulls. This will spam adminlogs. By var-editing this var to false, you will no longer get chat-messages for consume(). It still goes into investigate logs though, so you can still access them if you need to.

image


ghost:
THANKS NICH, NOW STAFF CANT STOP ME! MENTOS RISE UP

alexkar598 (changes requested):
Make the message still show up for mobs with a ckey

PR #7036, opened 2019-11-06
Adds a new chemical recipe to make godsblood to curb botanist making instaheal plants

boodaliboo:
when earthsblood and omnizine are combined it creates godblood which is omnizine with a od of 150
so botanist are rewarded with a intresting new chem for combining earthsblood with omnizine but not given a instaheal mix

#### Changelog

:cl:  
rscadd: Added a new chemical recipe 
/:cl:

### DIFF
--- a/code/modules/holiday/halloween.dm
+++ b/code/modules/holiday/halloween.dm
@@ -46,7 +46,7 @@
 /obj/structure/closet/proc/set_spooky_trap()
 	if(prob(0.1))
 		trapped = INSANE_CLOWN
-		returncode/modules/holiday/halloween.dm
+		return
 	if(prob(1))
 		trapped = ANGRY_FAITHLESS
 		return
--- a/code/modules/holiday/halloween.dm
+++ b/code/modules/holiday/halloween.dm
@@ -46,7 +46,7 @@
 /obj/structure/closet/proc/set_spooky_trap()
 	if(prob(0.1))
 		trapped = INSANE_CLOWN
-		return
+		returncode/modules/holiday/halloween.dm
 	if(prob(1))
 		trapped = ANGRY_FAITHLESS
 		return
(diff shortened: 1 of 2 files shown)


Amelia0010:
what

boodaliboo:
> what

it just allows godblood to be creatable which is identical to omnizine bt with a significantly higher od

Amelia0010:
right so what stops them from just making omnizine instead?

boodaliboo:
they already get omnizine from ambrosia deus 

> right so what stops them from just making omnizine instead?

Rayisken:
It's not really interesting if all it has is a different OD threshold, the current chem downsides can be mitigated by mixing with other things to minimize the effects. If you were adding another tier to these chemicals I think something closer to Atropine would make more sense.

boodaliboo:
I’m not trying to make it more powerful  trying to make it slightly harder to make a Instant heal plant but also give something in return

Amelia0010:
but you haven't made it harder? You've just added another chemical, you've not made it harder to get the other one

boodaliboo:
> but you haven't made it harder? You've just added another chemical, you've not made it harder to get the other one

im not trying to make it harder im trying to prevent two chems that heal very fast being used in tandem but not just outright making them do nothing when combined

ShadowDeath6:
Not sure on the percentages for omnizine and earthsblood, but if they're equal this should stop one plant from fully healing you instantly.

It doesn't really solve the issue at higher dosages though, since people can just take more earthsblood and then it's functionally identical to omnizine+earthsblood anyway

boodaliboo:
> Not sure on the percentages for omnizine and earthsblood, but if they're equal this should stop one plant from fully healing you instantly.
> 
> It doesn't really solve the issue at higher dosages though, since people can just take more earthsblood and then it's functionally identical to omnizine+earthsblood anyway

that is technically true but if i remeber correctly both omnizine and earthsblood got the same percentage in there plants

HomocidlePancakes:
>Using 2 chems with extremely different metab rates in 1 plant

>OD

B R U H

alexkar598:
Not a problem at the moment.

PR #7052, opened 2019-11-08
Changes (probably) all the times that Clock Cult uses watts to joules

yacabo:
Changes (probably) all the times Clock Cult uses the term watts to joules

If you do not understand what either joules or watts means lemme give you a quick physics lesson: Watts are a calculation of how much energy is being used in a second and since joules are a way to denote energy you can calculate watts, among other ways, as `W = J / second` e.g. 100 joules per second is 100 watts and 400 joules per two seconds is 200 watts.

Now that that's out of the way, why is this necessary? Well since watts are a calculation of how much energy is being used second by second and Clock Cult uses it's power as one off thing (Clock Cultist uses this much power to do this in this instance) it doesn't really make sense to measure it in watts rather joules.

No recalculations were changed in this PR, just how it describes the numbers. Also since there are a lot places in which the code did the thing to classify it as watts there's a chance I missed a few.

Some Pictures:

:cl:  
tweak:  The Clock Cult now calculates their power in joules rather than watts 
/:cl:

### DIFF
--- a/code/__DEFINES/clockcult.dm
+++ b/code/__DEFINES/clockcult.dm
@@ -7,7 +7,7 @@
 
 GLOBAL_VAR_INIT(clockwork_construction_value, 0) //The total value of all structures built by the clockwork cult
 GLOBAL_VAR_INIT(clockwork_vitality, 0) //How much Vitality is stored, total
-GLOBAL_VAR_INIT(clockwork_power, 0) //How many watts of power are globally available to the clockwork cult
+GLOBAL_VAR_INIT(clockwork_power, 0) //How many joules of power are globally available to the clockwork cult
 
 GLOBAL_LIST_EMPTY(all_clockwork_objects) //All clockwork items, structures, and effects in existence
 GLOBAL_LIST_EMPTY(all_clockwork_mobs) //All clockwork SERVANTS (not creatures) in existence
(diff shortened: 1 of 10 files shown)


alexkar598:
How does this improve the game in any way

Readystorm546:
> How does this improve the game in any way

Because Joules is better than watts. As the CEO of clock cult i approve of this

yacabo:
> How does this improve the game in any way

see

> Now that that's out of the way, why is this necessary? Well since watts are a calculation of how much energy is being used second by second and Clock Cult uses it's power as one off thing (Clock Cultist uses this much power to do this in this instance) it doesn't really make sense to measure it in watts rather joules.

Alternatively, it makes things more readable. Using watts implies that the machinery and spells would only take power while being used and would return the power once no longer being used, while joules suggests that whenever you use a spell or build a machine it will be a permanent removal from the power grid, after all why does kindle have a permanent siphon from the power when it's only being used for a brief moment in time?

Like imagine if a station had an SM that outputted 150kw and also had a gun that would feed from the power grid. Would it make sense that when the gun charges itself and fires that the SM now outputs 140kw despite the fact that the gun feeds solely off of power and has nothing to do with the SM?

ynot01:
downvoted because change is scary

yacabo:
Thanks

PR #7057, opened 2019-11-10
Ports holdup system from tg

SomeguyManperson:
Original PR: tgstation/tgstation#47677
Adds a holdup system, allowing you to coerce someone into staying still by pointing a gun at them (grab intent in melee range). Once this has been activated,  you can move up to 3 tiles away from the target without canceling the effect. If the victim
Attempts to use a weapon
Attempts to do any unarmed attack
Attempts to fire a gun
Attempts to throw an item
Attempts to move
You will automatically fire at them. The bullet will deal double damage after 2 seconds of being in holdup and 2.5x damage after 10 seconds. It can still be dodged like a regular bullet if you're fast.
Additionally, it's possible to move and use items while holding someone up. You will ALSO have a chance to fire at the victim if you take damage.

THINGS THAT DO NOT TRIGGER SHOOTING
The victim being bumped by the attacker
The victim being pulled by the attacker
The victim being thrown by the attacker
The victim picking up, dropping, or otherwise moving items without throwing
Stunbaton

The effect comes with a very small immobilize at the start so the victim doesn't immediately trigger the shot on accident, this effect can only occur once a minute so it can't be spammed in combat.
🆑 Ryll/Shaps
add: You can now hold people at gunpoint! Attack someone at point blank with a gun on GRAB intent to initiate the stickup. You will automatically fire at your target if they move or attack, though they can still rummage through their (or your!) inventory, speak (including on the radio), and throw things. Go let loose your inner Omar!
/🆑


ThatLing:
lol imagine having to force rp

MegaEmpirical:
Fancy.

Szyszkrzyneczka:
Speedmerge

ghost:
Speedmerge, this is great for unrobust stealth players

HomocidlePancakes:
teleports behind you

yacabo:
This would be a great PR if it wasn't for the damage ramp up, the aim bot, the fact than massive parts of this features rely on time passing, and the fact that you can literally just shoot them yourself

ynot01:
honestly fun idea imo, has a few flaws though

SomeguyManperson:
firing a weapon now cancels holdup so you can't just blat someone 20 times and pretend it's their fault for moving

MegaEmpirical:
I don't mind the aim bot for one shot. The damage ramp up could be concerning.

SomeguyManperson:
If the damage ramp up is considered to be too fast I can invert the ramp up thresholds so waiting out thw whole 10 swconds is worth more, but the ramp up is the only real reason soneone wouldn't try to avoid the bullet if they aren't already heavily injured

KrabSpider:
The attacker should have commands options that give the hostage an actual choice in the matter (i.e walking, limited running/running with a speed inhibition, allowing them to pick up and drop what stuff, and talking)

Also! Few questions:

•Will there be a visible (distinctive!) sprite on the hostage so they at least have a basic grasp of what's going on?

•Why include a damage multiplier in the first place? 

•Why allow the hostage to go *the attacker's* inventory? What's the purpose of that?

MegaEmpirical:
Being to look at the other person's inventory is referring to the way it currently works where you're not notified if someone is taking stuff from your pockets. It's how it works now.

SomeguyManperson:
It'd take some work but it's probably possible to add some toggles

The status effect for aiming and being aimed at shows both parties that they are partaking in a holdout

The damage multiplier both gives a good reason to use the feature instead of just shooting someone (and healing them later) and gives a good reason to not yeet off in all cases the attacker isn't carrying a gun that can immediately put someone into damage slowdown, and increases the viability of carrying a hostage very slightly by making attempts to pull them away get them shot

The victim is not prohibited from grabbing items, this includes stripping the attacker if they aren't careful, but that ahould be fairly difficult to do if the attacker is paying attention

Ryll-Ryll:
just a heads up there's a bunch of stuff I just pushed in the original PR that you should put in this, chiefly the additional clumsy and can_shoot checks that take care of hulks and lack of pins

SomeguyManperson:
I dont feel like adding a toggle for speaking over the radio would be worth the work put in since the message would get out over the radio either way and you can still shoot the person yourself.

Ryll-Ryll:
pushed another important fix to my PR from beestation's testing

PR #7058, opened 2019-11-10
Adds a signal technician clothing vendor

Ktlwjec:
The others get vendors but the signal tech was stuck with a wardrobe :( I absolutely hate the name of it though, if anyone has a better suggestion please let me know.

A proud Signal Technician after finally getting recognition from NT:
unknown

:cl:  
rscadd: Signal Technician's have a clothing vendor.
/:cl:


Readystorm546:
I like this PR but im sad that i can no longer hide in that locker and set off a max cap during war opps. :(

code/modules/vending/wardrobes.dm
```diff
@@ -109,6 +109,24 @@
 /obj/item/vending_refill/wardrobe/atmos_wardrobe
 	machine_name = "AtmosDrobe"
 
+/obj/machinery/vending/wardrobe/sigtech_wardrobe
+	name = "SigtechDrobe"
```
ynot01: SigDrobe?

ynot01 (changes requested):
sorry i have no idea how reviews work but heres my name suggestion

MegaEmpirical:
I hate the vendrobes so much. This makes me incredibly sad. One of the last bastions of freedom left on the station dying.

Xoxeyos:
Keep the locker, you can do alot more with it, such as you know, hiding in it, there aren't many places to hide on the sat.

Hopekz:
Always thought that signal tech was a part of engineering in the past and since there is only one signal tech that locker kind of fit to cram all their shit in to.

I'm neutral as it is not a strong dislike just kind of **feels out of place in my opinion**.

nesisknesis (approved):
is that syd garland

Ktlwjec:
To address concerns:
I'm changing the name to SigDrobe, it's what I had originally and it really does sound better.
I'll add a new locker to boxstation tcomms, so there will be the same amount of lockers there as before.
It's not about whether you like the vendors or not, it's a matter of consistency. As we got this from TG, there was no vendor made to replace the Signal Technician wardrobe. If there was, we would've added it in when the other lockers were replaced.

Ktlwjec:
I'll just remake this, that is a brand new map copy idk what it's doing.

PR #6988, opened 2019-11-01
Makes the surrendering text more noticeable

Darkstick:
The surrender emote is currently useless. It being simple generic black text makes it very hard to notice, especially in the middle of combat, where you would want to use the emote. This PR simply makes that text huge and red so that it can be seen even when you're about to shoot some poor bloke in the face.
#### Changelog

:cl:  
tweak: added the "userdanger" class to the text that appears after using the surrender emote.
/:cl:

PR #7059, opened 2019-11-10
Surrender text clarity revive

Hopekz:

Revives #6988 
I fixed the line endings which killed the original PR and also replaced the span with a much easier to see type even mid-combat per Nich's original point in #6988. 

Dark's original edit:
----
image

This PR:
-----
image

-----------------------------

CURRENT:: EDIT per Xoxeyos's suggestion. 
-----

Surrender text now has its own custom span to make it very much unique and not take up the whole screen. I personally like this edit.

image

EDIT 2:: I STOLE A ANIMATED SPAN FROM Alex thank you! 
----
J7YLA0J

#### Changelog

:cl:  Hopek
tweak: Surrender emote is now much easier to see. Even mid-combat
/:cl:


MegaEmpirical:
I'd support trying a color other than red, red is usually used for aggressive/attacks.

Xoxeyos:
I don't see why we need this, as I hardly see anyone use it, and I believe it is much too big of a font, perhaps you can try cutting it down by half. And perhaps a color change, like blue. I will change my vote if I like where it's going.

MegaEmpirical:
I think the font should be big and obvious. And he's making it because it's hardly used and this makes it more attractive/viable.

and because something is hardly used is good reason to improve it, not ignore it. I didn't even know it was a thing before. Plus, you could just not use it if you don't like it as it has no effect on you if you don't.

Hopekz:
> I don't see why we need this, as I hardly see anyone use it, and I believe it is much too big of a font, perhaps you can try cutting it down by half. And perhaps a color change, like blue. I will change my vote if I like where it's going.

**Xoxeyos,**
Made an edit per your suggestion. Its still pretty easy to read (if not more). Thank you for the feedback!

**People that care let me know if this new edit is more preferred by upbwoinking and down ogga'ing this comment**

image

MegaEmpirical:
How does it look in dark mode?

Hopekz:
J7YLA0J
This PR just got 10x cooler with @alexkar598  's span text

Here is how it looks now 
-----

MegaEmpirical:
I guess that works in light and dark mode.

Darkstick:
oh shit, this is a lot better than I could have done it, it's great

PR #7061, opened 2019-11-12
Obsessed people may no longer have the objective to have a picture with their target

alexkar598:
Obsessed people may no longer have the objective to have a picture with their target.
They will now get 2-3 objectives instead of the hard 3 as the polaroid objective will not get added no matter what.

Yes I know, im a big fan of improve dont remove myself but in this case, I can't figure out why it does this so it might be better to leave it out.

#### Changelog

:cl:  
rscdel: Obsessed people may no longer get the objective to take a picture with their target as it was bugged out.
/:cl:

### DIFF
--- a/code/modules/antagonists/creep/creep.dm
+++ b/code/modules/antagonists/creep/creep.dm
@@ -70,10 +70,14 @@
 				spendtime.target = obsessionmind
 				objectives += spendtime
 			if("polaroid")
+				/*
+				 * This is currently commented out like this so that there still is a chance for a third objective if applicable
+				 * but this won't add the objective if it tries to select the picture one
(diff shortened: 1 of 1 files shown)


Hopekz:
+1 unless we get some kind of uber nerd with unreal SS13 coding ability to fix this hint hint commooon monster, altoids,nich

I've gotten this objective at least 4 times over the last few months it is completely broken.

missatessatessy:
I tried having a look at this a few days ago, and I couldn't see why either, so I would say just remove too. People care more about their greentexts anyway xD

ynot01:
i ded
i mean good change thanks

alexkar598:
This is a "fix" as far as im concerned

PR #7062, opened 2019-11-12
Update firedoor.dm

monster860:
fixes it so that the capitalization matches the actual file.


yacabo:
Neat. Was this causing any issues though?

monster860:
yes

PR #7072, opened 2019-11-14
Monkey cube crate says the correct amount of monkey cubes

Ktlwjec:
title


Hopekz:
Also adds a super very important line ending space to make things pretty

PR #7079, opened 2019-11-15
Makes Tcomms networks independent from one another

Lostlarry:
So until now the tcomms signal was affected by all networks at once. This will make that a history.
Only draw back with this is that there are two receivers in default set up so tcomms will take up more processing power.

:cl:   
tweak: each tcomms network now works with copy of the signal instead of signal itself 
/:cl:


MegaEmpirical:
Can you explain that again?

Lostlarry:
Until this PR every Tcomms receiver is given reference to the original signal
that means that any server that edits the code will affect how the result looks
so if you had three networks and one of them didn't return a signal from process_signal() NO broadcaster on the station would send it out
this PR gives each receiver it's separate copy which prevents this from happening

Hopekz:
Good

PR #7081, opened 2019-11-16
Signal tech clothing vendor take 2

Ktlwjec:
#7058 But I don't rip out my hair trying to fix a merge conflict even though I got a brand new map copy.

:cl:  
rscadd: Signal Technician's have a clothing vendor.
/:cl:


Hopekz:
While I understand your mindset I just don't like the feel of this particular change with the signal tech job. I get your "matter of consistency" argument I just don't think the current feel of signal tech accommodates a clothing vendor.

With this "matter of consistency" mindset are the other 1 slot jobs like Warden, Psychiatrist or maybe Clerks going to get a job-specific clothing vendor too?

What about the odd jobs that don't have it like Paramedics, Miners or Detectives?

I always thought the signal tech was a part of engineering and not its own department so if you really want to go with a Clothing Vendor maybe consider adding it to the already existing vendor.

Ktlwjec:
The only relevant job you listed are the miners, who need their wardrobe replacing too. I'll get round to that at a later point. The other jobs you listed don't have a clothing wardrobe to replace, but feel free to make one for them if you have job specific clothing for them.

Xoxeyos:
You don't need to replace it man, that locker is great for hiding items, and it has more uses than the clothes vendor.

Ktlwjec:
That's why I added a new locker to boxstation tcomms.

PR #7093, opened 2019-11-17
tcomms mapping

Lostlarry:
i made tcomms output heat but apparently did not make other maps work with that
chages:
adds cooling setup to Pubbys
fixes Omegas and Minskys set ups which had not enough or too many siphon vents
adds tcomms air alarms where aplicable
:cl:  
tweak: tweaks tcomms setups to work as intended, adds cooling to Pubbys tcomms
/:cl:


LoliconSlayer:
Hardly any point in fixing Minskys when Mayhem is gettin close to finishing Minskys 2 innit

PR #7103, opened 2019-11-18
Makes Haunted Magic Eightball available to both chaplain and curator

Ktlwjec:
Just makes more sense to me than giving it to the Curator. I've no issue adding them back if people really want it for them, but the chaplain should have access to this item.

:cl:  
tweak: Haunted Magic Eightball uplink item is now restricted to Curator and Chaplain.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1745,7 +1745,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			Be warned, that spirits are often capricious or just little assholes. To use, simply speak your question aloud, then begin shaking."
 	item = /obj/item/toy/eightball/haunted
 	cost = 2
-	restricted_roles = list("Curator")
+	restricted_roles = list("Chaplain")
 	limited_stock = 1 //please don't spam deadchat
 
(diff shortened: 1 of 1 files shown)


code/modules/uplink/uplink_items.dm
```diff
@@ -1745,7 +1745,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			Be warned, that spirits are often capricious or just little assholes. To use, simply speak your question aloud, then begin shaking."
 	item = /obj/item/toy/eightball/haunted
 	cost = 2
-	restricted_roles = list("Curator")
+	restricted_roles = list("Chaplain")
```
Hopekz:
```suggestion
	restricted_roles = list("Chaplain","Curator")
```

Hopekz: Both jobs make sense imo.
I do agree with you Chaplain does need this but I also don't think curator "doesn't" need it. Both seem to fit in their own way.

HomocidlePancakes:
The items is so “meh” that no one will mind.

boodaliboo:
I mind if your giving chaplain this give curator another item even just a reskinned revolver would suffice

alexkar598:
Doesn't chaplain already have tons of toys

PR #7122, opened 2019-11-21
Pick up this bayonet and stab that man NICKVR to death.

Codeatmos:

Bayonets are cool,I like them and then I saw only the combat knife was bayonnetable, thus this just spreads the bayonets out a little, with adding a bayonet to the improvised shotgun(but not when its sawn-off, for a trade of stealth VS bayonet)
In all honesty,I don't think bayonets should be locked behind cargo,instead I think they should offer weaker options but they are more avaliable.
Note that you **Cannot Bayonet a sawn-off shotgun,as in you are trading stealth and storability for bayonets**
Improvised shotgun I felt would make sense,it being ghetto and all,why not a ghetto combat knife bayonet,but if it is hated I'll chop it.

#### Changelog

:cl:  
rscadd: More Knives can be bayoneted now
rscadd: The Improvised Shotgun can now be bayoneted,glory to graytide.
🆑

### DIFF
--- a/code/game/objects/items/kitchen.dm
+++ b/code/game/objects/items/kitchen.dm
@@ -76,7 +76,7 @@
 	attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut")
 	sharpness = IS_SHARP_ACCURATE
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 50, "acid" = 50)
-	var/bayonet = FALSE	//Can this be attached to a gun?
+	var/bayonet = TRUE	//Can this be attached to a gun?
 	custom_price = 30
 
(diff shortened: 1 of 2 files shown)


alexkar598:
Greytide doesn't need any buff

Codeatmos:
> Greytide doesn't need any buff

This could be seen as buffing graytide,which in some sense it is but this also allows security to have their bayonets instead of having to order a(what was it) 2000,3200 crate for 3 knives?
If You think something should/should not be bayonetable just say.

HomocidlePancakes:
wat. You can attach all knifes as bayonets.

Codeatmos:
> wat. You can attach all knives as bayonets.

Yes, with the exception of carrot knife and ritual dagger, previously you could not.

Hopekz:
I think people are misunderstanding what this PR does.
This PR isn't even a Greytide buff. 
+1 for the added utility this PR brings. Its one of those things that feels like a missing feature.

PR #7125, opened 2019-11-21
moves the lobby music to the code, rather than the config

Amelia0010:
So people can change it easily without needing to start forum threads, bother council, etc.

Also, let's only change one song at a time so we don't get huge conflicts of "I like X but not Y, but I'll upvote anyways because Y good".


Kmc2000:
bad :(

Hopekz:
Is this going to become stale?

PR #7127, opened 2019-11-21
Viruses > Nanites

CMOisLing:
Silicolysis can now remove nanites better so it isnt hell to remove them from everyone.
:cl:    
tweak: Silicolysis is now more effective
/:cl:


Laus4Deus:
I request an approximation to volume/second about it. How was it with -0.4 and how is it with -0.7.
And an ignorant cuestion 'cuz I am a dumb bimbo. Are these the ticks we use for chemistry? (which are like one tick every 2 seconds and doesn't depend on lag)

CMOisLing:
> I request an approximation to volume/second about it. How was it with -0.4 and how is it with -0.7.
> And an ignorant cuestion 'cuz I am a dumb bimbo. Are these the ticks we use for chemistry? (which are like one tick every 2 seconds and doesn't depend on lag)

Im 90% sure life is handled every 2 sec, but I believe nanites are every 1 sec. It should be about -.35 nanites a second on its own. With thresholds met it will be around -1 a sec.

PR #7148, opened 2019-11-24
I think this fixes disabler cooler

Lostlarry:
I am not sure whether this is still a problem but i am sure the borg doesn't have a stungun.

:cl:  
bugfix: fixes disabler cooler borg upgrade not having an effect
/:cl:


HomocidlePancakes:
How does it fix the bug? The cooler is still in the game

PR #7152, opened 2019-11-25
umbral tendrils can chain easier

SomeguyManperson:
resting no longer stops tendrils from stunning which makes them approx 300% more viable time to get yelled at


Codeatmos:
thank you,if this is not merged i will personally shit on xantams desk,its only an amount of time until a metagame is established

PR #7155, opened 2019-11-25
darkspawn time dilation now actually ignores slowdown

SomeguyManperson:


Codeatmos:
image

Thank you for fixing made in heaven sling edition thank you

PR #7159, opened 2019-11-26
gunpoint bugfix

SomeguyManperson:
move passive checks above the point of no return check so they don't trigger it

:cl:  
bugfix: holdup doesn't break when pulling the target
/:cl:


Hopekz:
Might be a feature if you squint just the right way at it.
If you are pulling someone in the opposite direction then I'd imagine it would mess up your holdout.

I thought holdout was supposed to be for holding someone at gunpoint and maybe cuffing them instead of locking them in to holdout so you can use the gun as cuffs while dragging them around

PR #7162, opened 2019-11-26
Megafauna eat slimes

Amelia0010:
Megafauna eat slimes.

:cl:
tweak: Megafauna eat slimes.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/megafauna.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/megafauna.dm
@@ -108,7 +108,7 @@
 	. = ..()
 	if(. && isliving(target))
 		var/mob/living/L = target
-		if(L.stat != DEAD)
+		if(L.stat != DEAD && !istype(L, /mob/living/simple_animal/slime)) // Megafauna eat slimes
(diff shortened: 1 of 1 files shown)


boodaliboo:
It would be better if there neutral

Codeatmos:
Ok fuck no, megafauna would literally follow a trail of slimes to the mining base, or somewhere close, as well as eating all of the fucking slimes, I know xenobio should be stomped into a red mush and washed down a drain but godamn, fighting megafauna so you don't lose your ungi slime?

SomeguyManperson:
This just makes it so megafauna dont die to slimes, they already fight them but hierophant psysically can't harm slimes with its attacks

LoliconSlayer:
Gotta pray that mining gets to those slimes before they die xenobio mains

Hopekz:
To everyone that doesn't support this for some reason its literally a quick bugfix to slimes literally killing megafauna on their own.

If a more elegant solution gets created later on so be it but we need to stop this from happening for now. At least this kind of makes sense-ish

Amelia0010:
`Ok fuck no, megafauna would literally follow a trail of slimes to the mining base, or somewhere close,`

Slimes don't spawn in a trail to the mining base.

PR #7166, opened 2019-11-27
Fix span on lizard wings message

SomeguyManperson:
This never happened


Sir-Lagsalot:
Great, now we just need someone to fix the span of the lizard wings

Hopekz:
> Great, now we just need someone to fix the span of the lizard wings

feature

PR #6879, opened 2019-10-14
removes monkey gamemode

Amelia0010:
### Intent of your Pull Request

:)

#### Changelog

:cl:  
rscdel: monkey gamemode gone
/:cl:

PR #7171, opened 2019-11-28
Revert "removes monkey gamemode"

swissloaf:

Reverts yogstation13/Yogstation#6879

Monkey gamemode is, and will always be a great gamemode. It will be used more because it is good.

🆑 
rscadd: Monkey gamemode is back!
/🆑


boodaliboo:
Please also add it to rotation

PR #7177, opened 2019-11-29
Fix parmesan reagents

missatessatessy:
I failed to notice when making the original PR that all snacks have a max volume of 50.
I have nerfed permesan amounts to fit in with this, it still is more than other cheese but not as much anymore.
#### Changelog

:cl:  
bugfix: fixed parmesan
/:cl:


Potototatoh:
crank and crokodil fell off the face of the earth ie; the don't exist anymore and im too lazy to make a report on it

PR #7179, opened 2019-11-29
It's no longer necessary to insert your ID into the ORM

Amelia0010:

since the TGUI-next UI doesn't include an ID button and I'm not well-versed enough in it yet to add it.

#### Changelog

:cl:  
bugfix: your ID will no longer get stuck in the ORM
/:cl:


alexkar598:
Temporary fix

PR #7187, opened 2019-11-30
Cloners scan for dna before they check if the body is damaged

SomeguyManperson:
:cl:   
tweak: cloner now checks for dna before checking for husking
/:cl:


Sir-Lagsalot:
Speedmerge

Alexandria404:
What does that mean in simple speak

SomeguyManperson:
If a body is both husked and has the baddna trait it'll display the baddna message on svanning instead of the husk message if the cloner can't process husks

alexkar598 (approved):

PR #7192, opened 2019-11-30
Label workflow

Amelia0010:

Moving this from yogbot to github actions will probably be more stable, and reduces the amount of interaction with the github API yogbot needs to do. I'll figure out how to make it read the PR and apply it based on changelog PR content as well

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -4,7 +4,7 @@
 Config: config/**
 
 # When the .DME is changed
-DME Edit: **.dme
+DME Edit: **/*.dme
 
 # Any .dmi changes
 Sprites: **/*.dmi
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -21,7 +21,7 @@ Sound: '**/*.ogg'
 # Changes to the SQL subfolder
 SQL: 'SQL/**'
 
-# Encompasses both the tgui and the tgui-net folder
+# Encompasses both the tgui and the tgui-next folder
 tgui: '{tgui,tgui-next/**}'
 
 # Changes to the .Github folder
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,28 @@
...
+# Changes to a .dmm
+Map Change: '**/*.dmm'
+
+# Any changes to .ogg files are marked as sound
+Sound: '**/*.ogg'
+
+# Changes to the SQL subfolder
+SQL: 'SQL/**'
+
+# Encompasses both the tgui and the tgui-net folder
```
alexkar598: Tgui-net :thonk:

.github/labeler.yml
```diff
@@ -0,0 +1,28 @@
...
+tgui: '{tgui,tgui-next/**}'
```
alexkar598, line 25: Maybe rename it to UI and add css and HTML files too

.github/labeler.yml
```diff
@@ -0,0 +1,28 @@
+# Use this to test the matches: http://www.globtester.com/
+
+# Any file within the config subfolder
+Config: 'config/**'
+
+# When the .DME is changed
+DME Edit: '**/*.dme'
```
alexkar598, line 7: Do we really need this tag

alexkar598 (changes requested):

.github/labeler.yml
Amelia0010: It's there already, I only added the Github label.

.github/labeler.yml
Amelia0010: but why

PR #7209, opened 2019-12-02
513 (watch me break everything)

Amelia0010:
see commits


alexkar598:
It compiles.

Bibby0110:
This should work

GenericDM:
:eyes:


> It compiles.

Amelia0010:
There's no point testmerging this, as the host will continue running on 512 untill we know it's stable enough for it, this PR is just intended to allow compiling on 513

PR #7224, opened 2019-12-03
Adds short_desc to ghost role spawners

Ktlwjec:
stops people pinging me about this, this should be all of them


monster860 (commented):
so basically `"You are a [name]"` why doesnt the code just do that instead of having a short_desc.

Ktlwjec:
or nothing if one isn't set, currently it does this which is easy enough for me to replace
unknown

PR #7230, opened 2019-12-04
Added an RCD to the Chief Engineer's locker

Hopekz:
Tested before commit. it works.

I have no clue why this wasn't a thing before. 
I always thought the CE should have his own RCD in his locker. 
Sometimes he arrives late and isn't able to get one from the vending machine.

#### Changelog

:cl:  
rscadd: CE's locker now contains an RCD because sometimes he arrives late
/:cl:


Amelia0010:
I need to port the RED

PR #7234, opened 2019-12-04
Make lockers airtight when welded

monster860:
:cl: monster860
add: Lockers are now airtight when welded.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -34,6 +34,9 @@
 	var/anchorable = TRUE
 	var/icon_welded = "welded"
 	var/notreallyacloset = FALSE // It is genuinely a closet
+	var/datum/gas_mixture/air_contents
+	var/airtight_when_welded = TRUE
+	var/airtight_when_closed = FALSE
 
 
 /obj/structure/closet/Initialize(mapload)
@@ -144,6 +147,7 @@
 	climb_time *= 0.5 //it's faster to climb onto an open thing
 	dump_contents()
 	update_icon()
+	update_airtightness()
 	return 1
 
 /obj/structure/closet/proc/insert(atom/movable/AM)
@@ -196,6 +200,7 @@
 	opened = FALSE
 	density = TRUE
 	update_icon()
+	update_airtightness()
 	return TRUE
 
 /obj/structure/closet/proc/toggle(mob/living/user)
@@ -255,6 +260,7 @@
 				return
 			welded = !welded
 			after_weld(welded)
+			update_airtightness()
 			user.visible_message("<span class='notice'>[user] [welded ? "welds shut" : "unwelded"] \the [src].</span>",
 							"<span class='notice'>You [welded ? "weld" : "unwelded"] \the [src] with \the [W].</span>",
 							"<span class='italics'>You hear welding.</span>")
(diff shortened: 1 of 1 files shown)


boodaliboo:
i like this

alexkar598:
I am somewhat unthrilled by the perpective of fusion in a closet and bringing that on the shuttle

LoliconSlayer:
You do realize that most lockers have holes in them for air flow right

CMOisLing:
Maybe change the sprite so the holes get welded as well?

Bibby0110:
What does this positively add to the game?

boodaliboo:
> What does this positively add to the game?

you can safely afk in a locker

Bibby0110:
but you can't space people in lockers anymore
you can hide in lockers from impending plasma fires
you can use lockers as improvised space suits if you have a buddy

just go to the dorms to AFK, why do you think they can be locked

Hopekz:
> but you can't space people in lockers anymore
> you can hide in lockers from impending plasma fires
> you can use lockers as improvised space suits if you have a buddy
> 
> just go to the dorms to AFK, why do you think they can be locked

In all my ss13 AFKing I've only been able to survive being in dorms twice otherwise I come back to being dead or worse off. Dorms is known the place people go where they're not playing anymore not to AFK .

I'm not for adding this because of being AFK though, I'm for this because its a really cool feature that makes sense. Its hard to tell what features like this will be used for but its nice to have added options.

Bibby0110:
You can't justify adding something because it MIGHT create some stuff while debuffing multiple other things

monster860:
In order to hide from a locker from impending plasma fires you need someone else's help. Same with going to space, and that other person is gonna need a space suit to take you anywhere useful. Either way someone needs a spacesuit. Also, you can still space people in a locker, just don't weld it. Or do weld it, knowing that the moment they escape the locker they will die from the cold hard vacuum of space.

Hopekz:
> You can't justify adding something because it MIGHT create some stuff while debuffing multiple other things

The other things that you mentioned aren't even "buffs" or "debuffs" its just different situations that add fun.

>but you can't space people in lockers anymore

You defiantly can. If you weld them first for some reason then the locker player has the option of holding on to a little bit to see if they can hit something like a space ruin to get a very slim chance of survival. What else are they going to do when they get out of the locker they can't weld themselves back in. 

>you can hide in lockers from impending plasma fires

You can't weld yourself in a locker. So if someone does a "get in this locker, save yourself" type situation then that's pretty cool that might add some nice situations. Not that lockers don't melt in extreme heat anyway..

>you can use lockers as improvised space suits if you have a buddy

Why is this a bad thing? 
Someone else with a space suit that has to weld you, drag you around and then unweld you (so makes you useless until you're unwelded...IF you're unwelded). 
I hope your argument isn't a realism argument because that's not really what space manz is about in my opinion (not that heaving a sealed container in space wouldn't be realistic)

You're acting like space survival is a hard to come by resource that needs to be monitored but this will let you survive just unable to do anything in space.

Remember we're playing a video game so we should add fun features.

Quartich:
someone got spaced

monster860:
actually the reason I made this lr is i looked at comments section of an ss13 video and someone said lockers were airtight when welded. i was like really? so i did some digging around and it was untrue so I made this pr to make it true. @quartich

PR #7249, opened 2019-12-05
[s] strips html on modular computer printer

Amelia0010:
Pretty large fucking oversight


monster860 (dismissed):
use html_encode instead ree

PR #7260, opened 2019-12-06
Fix meatball icon

Ktlwjec:
It was invisible, and removes duplicate meatball


Hopekz:
Now that's a spicy meatball!

PR #7262, opened 2019-12-06
Im back Also Clock cult update

Readystorm546:
>More map things.
>Change Airlocks correct airlock. Remove lots of random loot Spawners but i still leave some because there were so many

#### Changelog

:cl:  
Clock Nerf Kinda 
Map Change
/:cl:


SomeguyManperson:
set the name of the new airlocks to "ark chamber" if they're replacing the old ones aa

PR #7265, opened 2019-12-06
Makes PDA messages use the language system

Altoids1:
Robert paid me to do this.
image
## Overview
This makes it so that all messages you write via PDA are written in your default language. People who attempt to read your messages but lack knowledge of your language will see a ``scramble()``'d gobbledygook.

### Changelog
:cl:  Altoids
rscadd: Messages you write with PDAs are now written in your default language! Get a load of that one, sheltered folk!
/:cl:

PR #7268, opened 2019-12-07
Makes paper use the language system

Altoids1:
Robert paid me to do this.
image

## Overview

Similar to #7265, except it affects paper instead of PDAs.

This also affects papers pre-written and laid about the station; they will now have a specific language to them, most of the time.

### Coder Warning
I had to do some slightly-messy rewriting of how paper works in order to do this. ``info_links`` is now gone, and ``info`` now stores prewritten text which appears before any human-written text. ~~``info`` text has no language and is understood by all, for now.~~ ``info`` text has the language specified by ``infolang``. If no ``infolang`` is set, it defaults to being omnilingual and ergo legible by all.

### Changelog
:cl:  Altoids
rscadd: Things you write on paper are now in your default language, and cannot be read by those who don't know it!
/:cl:


MegaEmpirical:
Love seeing paper get some love. Always needs more.

code/modules/paperwork/paper.dm
```diff
@@ -22,8 +31,8 @@
 	max_integrity = 50
 	dog_fashion = /datum/dog_fashion/head
 
-	var/info		//What's actually written on the paper.
-	var/info_links	//A different version of the paper which includes html links at fields and EOF
+	var/info = "" // What's prewritten on the paper. Appears first and is a special snowflake callback to how paper used to work.
+	var/list/written = list()	//What's written on the paper by people. Stores /datum/langtext values, plus plaintext values that mark where fields are.
```
ThatLing: Assign this variable in /Initialize()

ThatLing (dismissed):

code/modules/paperwork/paper.dm
Altoids1: \>:(

MegaEmpirical:
So will this work with books?

Altoids1:
> So will this work with books?

It won't affect books, actually. I can do that in a separate PR, as I did with PDAs.

Hopekz (approved):

code/modules/paperwork/paper.dm
```diff
@@ -5,6 +5,15 @@
  * lipstick wiping is in code/game/objects/items/weapons/cosmetics.dm!
  */
 
+#define PAPER_FIELD "<span class=\"paper_field\"></span>"
```
alexkar598, line 8: Since this define is only used in this file, undef it at the end of the file

Altoids1: BITCH I DID

ThatLing: undef it harder

PR #7270, opened 2019-12-07
Now you can use runes on CentCom

Szyszkrzyneczka:
mapmerge finally work

### Adds runes to centcom from rune PR#6773
It is good, because it now makes runes available for round-end brawl to all people
#### Changelog

:cl:  
tweak: you can now use runes on cent-com
/:cl:


gilvian:
bad pr

Ratvar:
Why not make runes available at roundend instead?

Szyszkrzyneczka:
but it does that

PR #7271, opened 2019-12-07
improves the colour of the jani drobe

nesisknesis:
after
image

before
image

#### Changelog

:cl:  Identification
tweak: jani drobe is a better colour 
/:cl:


MegaEmpirical:
I've no beef with the color, but:

Remove the janidrobe!

PR #7277, opened 2019-12-08
hmm today I will fix an oversight

SomeguyManperson:
hierophant's attackingtarget() is different from the base megafauna one

:cl:  
bugfix: hierophant kills slimes
/:cl:


Nickvr628:
thanks

PR #7280, opened 2019-12-08
Dab animation framework + Hard Dab system + Existing Dab animated (achievement included)

Hopekz:
This PR adds 4 things.
------

**1. A dab proc framework.** 
This means that every single item in the game can perform an animated dab.
Usage for admins would be running the DabAnimation() procedure similar to SpinAnimation() 
Similar to spin animations angle of dab gets plotted on a matrix to animate.
Example: mob.DabAnimation(angle = 70 , loop = 10)  would hit a dab at 70 degrees 10 times.

 **2. Used this new proc to animate the regular dab emote.**
regular *dab hits between 30-45 degrees.

**3. Added a new emote called hard dab.**
This emote is called with *dabb and has tier systems dependant on RNG.
This hard dab hits between 55 to 110 degrees.
**There are 4 tier levels for this dab** each with increasing flair in the form of flavor text and new sound effects.
Hard dab brain damage is calculated based on the angle of dab performed.

**An Ultimate Dab can be achieved with a probability of 4%** 
This ultimate dab not only looks awesome (animation+flair text) and sounds great but also restores brain damage.

**4. Added the "King of Dab" achievement for hitting the ultimate dab.**

Honestly just load up this PR if you're curious about it or want to see the animation.

Edit: Added moodlet for hitting the ultimate dab

#### Changelog

:cl:  Hopek
rscadd: Created a Dab framework. Every single sprite in the game can now dab with DabAnimation() . Usage similar to SpinAnimation()
rscadd: Animated *dab using this framework
rscadd: Added a hard dab under *dabb . Hard dabs give more flair the harder the dab (with flavortext and sound effects between tiers)
rscadd: Hit the ultimate dab for extra flex and an achievement!
/:cl:


ghost:
Finally, the feature we all need

SomeguyManperson (approved):
works about as expected

ThatLing:
Please remove my maintainer if this gets merged

AsV9:
I might have to press the delay button here @ThatLing 
image

code/datums/achievements/achievements.dm
```diff
@@ -40,3 +40,10 @@
 	desc = "Kill yourself using the nuclear authentication disk"
 	id = 7
 	hidden = TRUE
+
+
+/datum/achievement/dab
+	name = "King of Dab"
+	desc = "Hit the ULTIMATE dab"
+	id = 420
```
Readystorm546: The ID Should Match up correctly and also this should be hidden

code/datums/achievements/achievements.dm
```diff
@@ -40,3 +40,10 @@
...
+	hidden = FALSE
```
Readystorm546: Number should line up and also should be hidden

Hopekz: Both of your suggestions are false.

The ID does not need to line up it just needs to be a unique number. 
And this achievement isn't hidden so it is set to FALSE.

I can see where the confusion is with hidden as its a default value so I'll probably just remove the flag since it is redundant to make our code pretty but otherwise, I am not changing the flag on this achievement to true unless people throw a big stink over it. 
The changelog itself that everyone gets access to directly states that this achievement exists.

Hopekz: I went ahead and committed your second suggestion. 
Thank you for saving our codebase from that redundant variable.

yacabo:
Dare I ask for an example of this in use?

code/datums/achievements/achievements.dm
Bibby0110: Numbers do have to line up for organizational purposes

Hopekz: We had this discussion about half a week ago. 
Our HC @nichlas0010 quite literally said that as a long as the ID is unique he does not care what the ID itself is.

Bibby0110: Seems @nichlas0010 is wrong

Hopekz:
> Dare I ask for an example of this in use?

Load up this PR and *dab, *dabb a few times. It is hard to capture because it is a combination of animation, flavor text and sound effects so I'd probably have to record a whole video for it.

If you want to make any item in the game to dab just call the "DabAnimation" proc on it. (works on quite literally everything I made a vending machine and my user interface dab)

AsV9:
I don't really think this fits with the rest of the game to be honest.

PR #7282, opened 2019-12-08
Probably nerfs DNA vault stacking and speed ability

SomeguyManperson:
:cl:  
tweak: DNA vaults can't be stacked, one use per body
tweak: DNA vault speed upgrade is slightly slower
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defines.dm
+++ b/code/modules/mob/living/carbon/human/human_defines.dm
@@ -47,6 +47,8 @@
 
 	var/datum/physiology/physiology
 
+	var/can_use_vault = 1
+
 	var/list/datum/bioware = list()
 
 	var/creamed = FALSE //to use with creampie overlays
--- a/code/modules/mob/living/carbon/human/human_defines.dm
+++ b/code/modules/mob/living/carbon/human/human_defines.dm
@@ -47,7 +47,7 @@
 
 	var/datum/physiology/physiology
 
-	var/can_use_vault = 1
+	var/can_use_vault = TRUE
 
 	var/list/datum/bioware = list()
 
--- a/code/modules/mob/living/carbon/human/human_defines.dm
+++ b/code/modules/mob/living/carbon/human/human_defines.dm
@@ -47,8 +47,6 @@
 
 	var/datum/physiology/physiology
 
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/human/human_defines.dm
```diff
@@ -47,6 +47,8 @@
 
 	var/datum/physiology/physiology
 
+	var/can_use_vault = 1
```
Bibby0110: var/can_use_vault = TRUE

code/modules/station_goals/dna_vault.dm
```diff
@@ -205,11 +205,14 @@
...
 	data["choiceB"] = ""
 	if(user && completed)
-		var/list/L = power_lottery[user]
-		if(L && L.len)
-			data["used"] = FALSE
-			data["choiceA"] = L[1]
-			data["choiceB"] = L[2]
+		if(ishuman(user))
+			var/mob/living/carbon/human/H = user
+			if(H.can_use_vault == TRUE)
```
Bibby0110: no need for the == TRUE

Bibby0110:
I'm sorry but I had to do it

Bibby0110:
sexy

Ratvar:
I know the true and only nerf to speed boosts: replacing it with something that's not

code/modules/mob/living/carbon/human/human_defines.dm
```diff
@@ -47,6 +47,8 @@
 
 	var/datum/physiology/physiology
 
+	var/can_use_vault = TRUE
```
Amelia0010: just have a static list on the DNA vault of all users instead, and only allow people to use it if they're not in said list

Amelia0010 (changes requested):

code/modules/mob/living/carbon/human/human_defines.dm
alexkar598: Didn't we fase out static vars for GLOBs :)

Amelia0010: har har har har har har very funny

Amelia0010 (approved):

PR #7284, opened 2019-12-08
I saw this while searching for something else and it triggered me

Amelia0010:


Hopekz:
hvEnMpKl

jcatjmeow:
Make sure to add if (TRUE) in front of the whole thing so the program doesn't get confused

code/modules/antagonists/cult/cult_items.dm
```diff
@@ -976,9 +976,7 @@ GLOBAL_VAR_INIT(curselimit, 0)
 		to_chat(holder, "<span class='cult italic'>The shield's illusions are back at full strength!</span>")
 
 /obj/item/shield/mirror/IsReflect()
-	if(prob(block_chance))
-		return TRUE
-	return FALSE
+	return prob(block_chance)
```
TheGamerdk: return FALSE

TheGamerdk: Replace with
`if(prob(block_chance))
		return TRUE	
	return FALSE`

code/modules/antagonists/cult/cult_items.dm
```diff
@@ -976,9 +976,7 @@ GLOBAL_VAR_INIT(curselimit, 0)
 		to_chat(holder, "<span class='cult italic'>The shield's illusions are back at full strength!</span>")
 
 /obj/item/shield/mirror/IsReflect()
-	if(prob(block_chance))
```
TheGamerdk: Duplicate of #42

PR #7289, opened 2019-12-09
Tgui next stuffs

Amelia0010:
see commits.

Didn't include the nanites UIs cause I cba


alexkar598:
But... you did add the UIs...

alexkar598 (approved):
Seems to be good

PR #7292, opened 2019-12-10
[s]

alexkar598:
E

### DIFF
--- a/code/game/turfs/simulated/floor/reinf_floor.dm
+++ b/code/game/turfs/simulated/floor/reinf_floor.dm
@@ -1,4 +1,3 @@
-
 /turf/open/floor/engine
 	name = "reinforced floor"
 	desc = "Extremely sturdy."
@@ -31,6 +30,9 @@
 	return //unplateable
 
 /turf/open/floor/engine/try_replace_tile(obj/item/stack/tile/T, mob/user, params)
+	return
+
+/turf/open/floor/engine/remove_tile(mob/user, silent = FALSE, make_tile = TRUE)
 	return
 
 /turf/open/floor/engine/crowbar_act(mob/living/user, obj/item/I)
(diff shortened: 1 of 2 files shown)


monster860:
what does this even fix

yacabo:
10/10 documentation

monster860 (changes requested):
what does this fix

AutisticFroggy:
I raised this issue in msay after retaining knowledge of the bug for some time, I didn't think it was too harmful until we discovered a method to abuse it for infinite metal glitches, then reported it to alexkar. This bug is in reinforced floors that causes efforts to remove the tile with anything except a wrench fail, this means singularity gravity_act() procs and disposals underfloor item release. It still releases metal rods during the process, but does not delete, causing an infinite amount of metal rods to issue forth from a single rfloor. I'm pretty sure this should fix it.

code/game/turfs/simulated/floor/reinf_floor.dm
```diff
@@ -33,6 +32,9 @@
 /turf/open/floor/engine/try_replace_tile(obj/item/stack/tile/T, mob/user, params)
 	return
 
+/turf/open/floor/engine/remove_tile(mob/user, silent = FALSE, make_tile = TRUE)
+	return
```
Amelia0010, line 36:
```suggestion
	.=..()
	ScrapeAway(flags = CHANGETURF_INHERIT_AIR)
```

Amelia0010 (changes requested):

code/game/turfs/simulated/floor/reinf_floor.dm
alexkar598: No that makes it be stupid and counter the point of reinforced floors being nearly indestructible by any means

alexkar598:
Let's not talk about unpatched exploits in public, ya?

PR #7303, opened 2019-12-10
tcomms achievement

Lostlarry:
adds two hidden achievements that concern a certain bird

:cl:  
rscadd: adds more things to be achieved when scripting NTSL
/:cl:


Hopekz:
I like these achievements even if the AI + any borgs can modify Poly to give the signal tech the achievement.

Amelia0010:
the AI/borg can't.

Hopekz:
> the AI/borg can't.

I still don't see it then.
From my understanding of it this change it creates a virtual radio that listens to Poly and if Poly is loud on that radio (or silent) by comparing what was done to the sig and it gives the achievement to the signal tech.

From my understanding this a;sp means that the AI or borgs can modify the NTSL script themselves for this achievement to trigger. Not necessarily that it is a bad thing but it is an observation.

Amelia0010:
It's a wrong observation. It only runs the test if the person compiling is a signal tech, if the AI/borg compiles it then the test won't run.

PR #7304, opened 2019-12-10
Gives traitor viros a dartgun

CMOisLing:
I see no reason why they shouldn't have this, plus it will encourage single target viruses instead of infecting the whole station.
:cl:  
tweak: Traitor Virologist now have access to the reagent dartgun  
/:cl:


MegaEmpirical:
Single target viruses. Cute.

CMOisLing:
> Single target viruses. Cute.

Corpses aren't much company on the shuttle ride home

PR #7305, opened 2019-12-10
Adds the cargoking and ducatduke achievements

Amelia0010:

Beating the current record in cargo credits. This achievement will get progressively harder as the record goes higher. 

Also added the "duke of ducats" achievement, which is just beating a million credits

I also added the SQL changelog since I forgot to add it last time.

Oh and I also added the misc table to the DB which is essentially just a glorified assoc list


Reed0506:
I feel like this achievement would be better if it just had an extremely high credit goal instead of an increasing goal. the QMs who play right when (if) this achievement is merged get it for basically no effort at all whereas down the line it would probably be almost completely impossible and i feel like both of those things are not what achievements are about. This concept would be far better for a leaderboard system.

alexkar598:
This has a problem of difficulty being ridiculously easy at first and ridiculously hard, make it a fixed number like 1 million instead

Amelia0010:
I added another achievement, ducat duke, for that. The point of it getting harder and harder is intended, that's how being "the best" works :)

code/datums/achievements/achievements.dm
```diff
@@ -104,4 +104,24 @@
...
+	Q.Execute()
+	if(Q.item.len)
+		amount = Q.item[1]
+	qdel(Q)
+	desc += "[amount]"
+
+/datum/achievement/ducatduke
+	name = "Duke of Ducats"
+	desc = "As the QM, have a million cargo credits by the end of the round" //Cargoking-junior
+	id = 21
```
alexkar598: ID conflicts, change the ID

alexkar598 (dismissed):

PR #7311, opened 2019-12-11
adds ethereal mutation toxin and a recipe for it

boodaliboo:
currently theres no way to become a ethereal even with strange seeds so i added a chemical recipe using liquid electiricty and a green extract to become one

#### Changelog

:cl:  
rscadd: Added ethereal mutation toxin and recipe    
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes/slime_extracts.dm
+++ b/code/modules/reagents/chemistry/recipes/slime_extracts.dm
@@ -70,6 +70,14 @@
 	required_other = TRUE
 	required_container = /obj/item/slime_extract/green
 
+/datum/chemical_reaction/slime/slimeethereal
+	name = "Ethereal Mutation Toxin"
+	id = "etherealarerealmuttoxin"
+	results = list(/datum/reagent/mutationtoxin/ethereal = 1)
+	required_reagents = list(/datum/reagent/consumable/liquidelectricity = 1)
+	required_other = TRUE
+	required_container = /obj/item/slime_extract/green
+
 //Metal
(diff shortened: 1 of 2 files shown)


code/modules/reagents/chemistry/recipes/slime_extracts.dm
```diff
@@ -70,6 +70,14 @@
 	required_other = TRUE
 	required_container = /obj/item/slime_extract/green
 
+/datum/chemical_reaction/slime/slimeethereal
+	name = "Ethereal Mutation Toxin"
+	id = "etherealarerealmuttoxin"
+	results = list(/datum/reagent/mutationtoxin/ethereal = 1)
+	required_reagents = list(/datum/reagent/consumable/liquidelectricity = 1)
```
alexkar598, line 77: Maybe teslium instead?

alexkar598 (changes requested):

code/modules/reagents/chemistry/recipes/slime_extracts.dm
boodaliboo: yeah that sounds better ill change it in a bit

boodaliboo: was just thinking that liquid electricty was available already from high power energy bar and its also the blood of ethereals

boodaliboo:
so do people want it to be teslium or liquid elec

alexkar598:
Not interested and it seems noone else is either

PR #7312, opened 2019-12-11
tcomms emaged edition

Lostlarry:
for now it just adds mess to the ui for all tcomms machines
unlocks signal compression control to processor(should make the signal unreadable)
makes server save records as corrupted
i may add booby trapped button later
:cl:  
rscadd: tcomms can be emaged now which may or may not be useful
/:cl:


Ratvar:
I am not sure fixing such a vital station asset should be even harder because some random machine is emagged and you'll never find out which.

Hopekz:
What @Ratvar said except add some kind of way to know. Like for airlocks you can tell its been emagged with the sparks flying out and message. Maybe add sparks or even have something change color?

Ratvar:
But if you *can* see that it was scrambled, then it's useless because you destroy the machine and make a new one
I like corrupted records, tho

PR #7315, opened 2019-12-12
Update motd

ThatLing:
Fixes tiny text probably. I also wanted to remove the "general survey" but it looks like that change wasn't pushed to github.

image


Ratvar:
thanks mr ling

PR #7320, opened 2019-12-13
Adds window shutters spawners

Ktlwjec:
So mappers (me) can see them.

A before and after:
unknown


MegaEmpirical:
Is there a way that we can get them placed above the grille? They seem difficult to reach as an AI since the grille sprite will cover them if the window is broke but not the grille.

Ktlwjec:
They do appear above the grilles.
image

MegaEmpirical:
That's good, thanks.

Though I believe the ones already live do not.

Ktlwjec:
Ah I see what you mean, should be easy to fix that.

PR #6, opened 2018-03-12
mindslaves!

Amelia0010:

PR #7329, opened 2019-12-14
adds a /hat because missa forgot to add it so the helmets till now were identical to to the armor

boodaliboo:

Explain what the purpose of your PR is.
Mention if you tested your changes. If you changed a map, make sure you used the mapmerge tool.

Mention if your PR is related to an issue.
You can type "Fixes issue #6" and issue 6 will be automatically closed when your pull request is merged. This is much easier for everyone.

Edit the changelog at the bottom for changes that are noticeable by the players. Remove it if this isn't the case.
Include [admin] in the title if it's something admin related.
Include [s] in the title if you don't want it to be announced on discord and the server.

#### Changelog

:cl:  
bugfix: fixed a missing /hat 
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/hardsuit.dm
+++ b/code/modules/clothing/spacesuits/hardsuit.dm
@@ -831,7 +831,7 @@
 //POWERARMORS
 //Currently are no different from normal hardsuits, except maybe for the higher armor ratings.
 
-/obj/item/clothing/head/helmet/space/hardsuit/powerarmor_t45b
+/obj/item/clothing/head/helmet/space/hardsuit/powerarmor_t45b/hat
 	name = "Salvaged T-45b helmet"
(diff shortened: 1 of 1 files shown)


boodaliboo:
what does this revert

SomeguyManperson:
Bot works off the changelpg which you should also actually make

boodaliboo:
> Bot works off the changelpg which you should also actually make

fine

missatessatessy:
After looking at this properly (I'm out of commission during weekends) the file is /obj/item/clothing/head/helmet so it doesnt need /hat (like all the other hardsuits in the same file). If there is an issue with the helmets that's something else to look into and this PR probably won't fix anything?

missatessatessy:
I already know what mistake I made, like the big dumb dumb I am

PR #7339, opened 2019-12-15
port bomb rebalance from tg

SomeguyManperson:
bombs are more dependable for knockdown but don't full stun
original PR: tgstation/tgstation#45819

#### Changelog

:cl:  81Denton
tweak: Explosions now cause knockdown instead of KOs, with the amount of time scaling off bomb armor. Heavy explosions still cause a ~2 second KO for follow up attacks.
tweak: Devastating explosions now no longer gib players with more than 50 bomb armor. This used to be more or less random depending on the amount of bomb armor.
tweak: Removed reagent explosion code that would trigger for <1 amount reactions.
/:cl:


Readystorm546:
Can we go back to 5/10/20

PR #7349, opened 2019-12-17
Airlocks now tell you where they lead

Amelia0010:
It's not exactly perfect, if both sides are in the same area then the airlock will say it leads to that area on both sides, but you know, such is life


AsV9:
I mean what could go wrong

PR #7358, opened 2019-12-18
Fixes boris preventing law display on round end

Lostlarry:
fixes couple of UI problems with AI

:cl:   
bugfix: boris module no longer prevents laws displayn on round end
bugfix: borgs no longer have the PDA verbs
tweak: AI alert UI made look better
/:cl:


code/__HELPERS/roundend.dm
```diff
@@ -359,7 +359,10 @@
 		if(aiPlayer.mind)
 			parts += "<b>[aiPlayer.name]</b> (Played by: <b>[aiPlayer.mind.key]</b>)'s laws [aiPlayer.stat != DEAD ? "at the end of the round" : "when it was <span class='redtext'>deactivated</span>"] were:"
 			parts += aiPlayer.laws.get_law_list(include_zeroth=TRUE)
-
+		else if(aiPlayer.deployed_shell)
```
alexkar598:
```suggestion
		else if(aiPlayer?.deployed_shell?.mind)
```
Just a tiny improvement, makes it take less code, else if leading into an if is yucky

Lostlarry: aiPlayer will always be reference to object unless the global AI list got fucked up

PR #7361, opened 2019-12-19
Makes miasma production consume oxygen

monster860:
**This will halt miasma production once oxygen reaches a partial pressure of 18 kPa**. This will not make the station unbreathable.

The upshot of this is that miasma production will no longer result in the pressure of the kitchen backroom being really high for no reason, messing up people's kitchen spreads. The other upshot is that bodies in space dont produce miasma anymore.

:cl: monster860
tweak: Miasma production now consumes oxygen
/:cl:


MegaEmpirical:
If the cold room was cold it wouldn't produce miasma. I know other maps have a freezer set up just for it

PR #7366, opened 2019-12-20
Experimentor death coffee now contains death coffee instead of a very small amount of regular coffee

SomeguyManperson:
For some reason these never got updated to use paths

:cl:  
bugfix: Experimentor death coffee now contains death coffee instead of a very small amount of regular coffee  
/:cl:


MegaEmpirical:
Sounds like it's working fine.

PR #7370, opened 2019-12-20
fix ashwalker cpr

SomeguyManperson:
le shitcode has arrived
:cl:  
bugfix: ashwalkers no longer kill other ashwalkers with cpr
/:cl:


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -645,16 +644,11 @@
 		log_combat(src, C, "CPRed")
 		SSachievements.unlock_achievement(/datum/achievement/cpr, client)
 		// yogs start - can't CPR people with ash walker lungs whithout having them yourself
-		if(they_breathe && they_ashlung && !we_ashlung)
-			C.adjustOxyLoss(anticpr)
+		if(they_breathe && (they_ashlung? TRUE : FALSE) != (we_ashlung? TRUE : FALSE))
```
alexkar598:
```suggestion
		if(they_breathe && !!they_ashlung != !!we_ashlung)
```

PR #7380, opened 2019-12-21
moves template-choosing to a proc

Amelia0010:

To allow two things:
1. Holiday-based templates. I dunno why, but this appeals to me.
2. co-dependent templates. Like having two templates in different locations that has similar rooms, but shouldn't pick the same. Like having one for the core and one for the satellite, and then making the satellite only load comms if core chose AI, or otherwise.

Also adds an irish bar, which is literally just a green bar because I'm lazy and have no creativity


yogstation/code/game/objects/effects/landmarks.dm
```diff
@@ -62,11 +62,22 @@
...
+
 /obj/effect/landmark/stationroom/box/bar
 	template_names = list("Bar Trek", "Bar Spacious", "Bar Box", "Bar Casino", "Bar Conveyor", "Bar Diner", "Bar Disco", "Bar Purple")
 	icon = 'yogstation/icons/rooms/box/bar.dmi'
 	icon_state = "bar_box"
 
+/obj/effect/landmark/stationroom/box/bar/choose()
+	if(SSevents.holidays["St. Patrick's Day"])
+		return "Bar Irish"
+	return ..()
```
alexkar598: Replace this with . = ..() at the top, you never know when something fucks up and the proc somehow runtimes and you end up without a bar

alexkar598 (changes requested):

yogstation/code/game/objects/effects/landmarks.dm
```diff
@@ -62,11 +62,22 @@
...
+/obj/effect/landmark/stationroom/proc/choose()
+	return safepick(template_names)
+
 /obj/effect/landmark/stationroom/box/bar
 	template_names = list("Bar Trek", "Bar Spacious", "Bar Box", "Bar Casino", "Bar Conveyor", "Bar Diner", "Bar Disco", "Bar Purple")
 	icon = 'yogstation/icons/rooms/box/bar.dmi'
 	icon_state = "bar_box"
 
+/obj/effect/landmark/stationroom/box/bar/choose()
+	if(SSevents.holidays["St. Patrick's Day"])
```
alexkar598:
```suggestion
	if(SSevents.holidays && SSevents.holidays["St. Patrick's Day"])
```

alexkar598 (changes requested):
Fix the DME line endings too

alexkar598 (approved):

PR #7407, opened 2019-12-26
A soul for a soul, a greentext for a greentext

SomeguyManperson:
Adds an assist objective which is completed if the target's objectives are complete. Currently is not applied in any objective pools and is bus only since it'd probably just be seen as another assassinate objective like protect objectives are

:cl:  
rscadd: Add assist objective, which will count as complete if the target greentexts. Currently not in use
/:cl:


MegaEmpirical:
I used a protect objective for something like this earlier, I like it.

PR #7411, opened 2019-12-26
Makes undeleted queries not message admins

Amelia0010:
really should just be an SQL log, what the fuck tg


AsV9:
big :clap: brain :clap: debugging

PR #7413, opened 2019-12-26
Cyber appendix works as intended

SomeguyManperson:
now a synthetic organ with EMP vulnerability and the inflamed removal works more consistently
:cl:  
bugfix: Cyber appendix works as intended
/:cl:


CMOisLing:
Oof

PR #7420, opened 2019-12-27
Fixes bug where antags can get other antags as objective (yes this is actually a bug)

monster860:
There is a clause in the code supposed to prevent this from happening, but it doesnt work correctly, because antags get antag datums at the same time they get objectives, so an antag can get an objective that hasn't been made a traitor yet, then they get made a traitor.

Oh yeah, this bug means that latejoin > ready up because if you get latejoin antag then you physically cannot be someone's objective. Yall always complaint about how people don't ready up and this is one bug that could prevent people from readying up.

You could still get potentially cucked if you're a antag token receiver, or get antagged by an admin, but meh.

### DIFF
--- a/code/game/gamemodes/objective.dm
+++ b/code/game/gamemodes/objective.dm
@@ -112,7 +112,7 @@ GLOBAL_LIST(admin_objective_list) //Prefilled admin assignable objective list
 		if(!(possible_target in owners) && ishuman(possible_target.current) && (possible_target.current.stat != DEAD) && is_unique_objective(possible_target,dupe_search_range))
 			//yogs start -- Quiet Rounds
 			var/mob/living/carbon/human/guy = possible_target.current
-			if(possible_target.antag_datums || !(guy.client && (guy.client.prefs.yogtoggles & QUIET_ROUND)))
+			if(possible_target.special_role || !(guy.client && (guy.client.prefs.yogtoggles & QUIET_ROUND)))
 				if (!(possible_target in blacklist))
 					possible_targets += possible_target//yogs indent
 			//yogs end


Ratvar:
So now traitors can cooperate with each other without any fear of backstabbing?

alexkar598:
This is literally an I ded, I killed monster, I was blood bro, he was traitor and he made this PR within the same round

Ratvar:
I am not sure if it matters if that's a bug or "ided": goal is to have good gameplay, right?
It would be shitty if antags would all band together to kill everyone else, tho, and this PR guarantees they'll do so.

monster860:
It's still a bug. Reading the code, the intended behavior is clearly that antags shouldnt get other antags.

Ratvar:
How does it matter if it's a bug if fixing it makes game worse? Remove whole thing altogether

Hopekz:
While this is technically a "bugfix" I think this feature was unintentionally better being broken.
I like traitors getting to get other traitors as an objective. 
Really creates nice moments where you're going after someone and all of a sudden he pulls out his traitor weapons for a nice battle then whoever wins that battle gets the spoils of the loser's equipment.

monster860:
If you attack a regular person, they'll call for help, they'll disarm, they'll fight back. If you attack a traitor, they might try to say the codewords, *not* call for help because that involves security possibly searching them, and try to give you whatever objective item they have

SomeguyManperson:
it's both a bug and an i ded at the same time so what do those cancel eachother out or something

ToGWtF:
you just created blood brothers but this time they have syndie uploads. good god man

alexkar598:
This is not a fix, its a bypass so that antags cannot have quiet mode objective protection.

PR #7421, opened 2019-12-28
Reduces the timer on the firelock open when there is a pressure difference

monster860:
:cl: monster860
tweak: Reduces the timer on the firelock open when there is a pressure difference
/:cl:


Potototatoh:
interesting name

alexkar598 (approved):

Issue #6305, opened 2019-07-10
Dolphin soup crafting acually takes only blood and water, not meat or bowl


gilvian:
## Reproduction:

Just make dolphin soup

nesisknesis:
What?

Ktlwjec:
Ok reworded: After crafting, it doesn't delete the meats or bowl.

PR #7422, opened 2019-12-28
Demodularise dolphin soup + bug fix

Ktlwjec:
A fix for #6305 , tested it quite a bit and the bug doesn't reproduce anymore for me. Added foodtype as meat to the soup as well.

:cl:    
bugfix: Dolphin soup recipe should no longer have leftover ingredients after crafting.
/:cl:


monster860:
Youre not modularizing, you're demodularizing.

PR #7424, opened 2019-12-28
[admin] adds back kicking confirmation

### LINKED ISSUE #7415: Please add a confirmation to Kicking people
While doing something else I accidentally clicked Kick when someone hit the per minute message limit and it appeared where I was clicking, it then kicked them without confirming.

Please add a confirmation.

Ktlwjec:
I assume this fixes #7415 , which is the intention of this.

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -452,8 +452,8 @@
 			if(!check_if_greater_rights_than(M.client))
 				to_chat(usr, "<span class='danger'>Error: They have more rights than you do.</span>", confidential=TRUE)
 				return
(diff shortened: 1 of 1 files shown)


MegaEmpirical:
Thank you.

code/modules/admin/topic.dm
```diff
@@ -452,8 +452,8 @@
 			if(!check_if_greater_rights_than(M.client))
 				to_chat(usr, "<span class='danger'>Error: They have more rights than you do.</span>", confidential=TRUE)
 				return
-			/* yogs - admins don't need handholding if(alert(usr, "Kick [key_name(M)]?", "Confirm", "Yes", "No") != "Yes")
```
alexkar598, line 455: Admins do need handholding dont they

PR #7433, opened 2019-12-29
Fixes several bugs with Paper, buffs/nerfs photocopiers

Altoids1:
# Altoids releases Paperwork: 2!

image

## Overview
This fixes a bunch of bugs created by the recent language overhaul with paper (https://github.com/yogstation13/Yogstation/pull/7268), especially with how it interacts with photocopiers.

## Coder Warnings
While I was in there I made the maximum toner amount in photocopiers a #define called ``MAX_TONER``, so, there's that.

This also involved the creation of a new var in all paper called ``color_override`` which lathers a new font color over the whole paper (provided all font color spans within have been already neutered), so that's a thing too. It doesn't automatically kill all fontcolor'd things on the paper to allow for clowns to write in crayon on copied pieces of paper.

### Changelog
:cl:  Altoids
bugfix: The paper icon should now update when it's written onto.
bugfix: Fixes photocopiers failing to copy paper with only player-written text.
bugfix: Fixes bug with handwritten paperfields always writing to bottom.
bugfix: Copies of copies of copies of copies of paper should now be capable of being faded out.
bugfix: Player-written text will now actually be printed black in photocopiers.
/:cl:

### DIFF
--- a/code/modules/paperwork/photocopier.dm
+++ b/code/modules/paperwork/photocopier.dm
@@ -66,16 +66,23 @@
 							copy_as_paper = 0
 					if(copy_as_paper)
 						var/obj/item/paper/c = new /obj/item/paper (loc)
-						if(length(copy.info) > 0)	//Only print and add content if the copied doc has words on it
+						if(length(copy.info) || length(copy.written))	//Only print and add content if the copied doc has words on it
 							if(toner > 10)	//lots of toner, make it dark
 								c.info = "<font color = #101010>"
 							else			//no toner? shitty copies for you!
 								c.info = "<font color = #808080>"
-							var/copied = copy.info
-							copied = replacetext(copied, "<font face=\"[PEN_FONT]\" color=", "<font face=\"[PEN_FONT]\" nocolor=")	//state of the art techniques in action
-							copied = replacetext(copied, "<font face=\"[CRAYON_FONT]\" color=", "<font face=\"[CRAYON_FONT]\" nocolor=")	//This basically just breaks the existing color tag, which we need to do because the innermost tag takes priority.
-							c.info += copied
-							c.info += "</font>"
+							var/copyinfo = copy.info
+							copyinfo = replacetext(copyinfo, "<font face=\"[PEN_FONT]\" color=", "<font face=\"[PEN_FONT]\" nocolor=")	//state of the art techniques in action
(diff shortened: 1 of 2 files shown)


code/modules/paperwork/photocopier.dm
```diff
@@ -7,6 +7,9 @@
 /*
  * Photocopier
  */
+
+#define MAX_TONER 40 // The maximum amount of toner in the photocopier, as well as the amount of toner given by toner cartridges.
```
alexkar598: Absolutly no need for a define here

code/modules/paperwork/photocopier.dm
```diff
@@ -23,7 +26,7 @@
 	var/obj/item/photo/photocopy = null
 	var/obj/item/documents/doccopy = null
 	var/copies = 1	//how many copies to print!
-	var/toner = 40 //how much toner is left! woooooo~
+	var/toner = MAX_TONER //how much toner is left! woooooo~
```
alexkar598:
```suggestion
	var/toner = 40//how much toner is left! woooooo~
```

code/modules/paperwork/photocopier.dm
```diff
@@ -239,7 +252,7 @@
 			if(!user.temporarilyRemoveItemFromInventory(O))
 				return
 			qdel(O)
-			toner = 40
+			toner = MAX_TONER
```
alexkar598:
```suggestion
			toner = initial(toner)
```

code/modules/paperwork/photocopier.dm
```diff
@@ -332,3 +345,6 @@
 	grind_results = list(/datum/reagent/iodine = 40, /datum/reagent/iron = 10)
 	var/charges = 5
 	var/max_charges = 5
+
+
+#undef MAX_TONER
```
alexkar598: no need for define

alexkar598 (changes requested):

code/modules/paperwork/photocopier.dm
Altoids1: It's a magic number used in 2 places. I want a define.

Altoids1:
@alexkar598 It's a magic number used in more than one place. That's enough to merit a ``#define``. I don't see the problem here.

alexkar598:
Its usage can be reduced to a single place and that would allow to make subtypes with more toner without having to change code

code/modules/paperwork/photocopier.dm
AsV9:
```suggestion

```
image

AsV9 (changes requested):

PR #7434, opened 2019-12-29
[admin] killer tomato logging

Ktlwjec:
Don't see this logged anywhere, can make it alert admins if that's needed.


code/modules/hydroponics/grown/tomato.dm
```diff
@@ -133,6 +133,7 @@
 		return
 	to_chat(user, "<span class='notice'>You begin to awaken the Killer Tomato...</span>")
 	awakening = 1
+	log_game("[key_name(usr)] activated a killer tomato at [AREACOORD(src)].")
```
alexkar598: Move that down under the check

PR #7442, opened 2019-12-29
Pepper spray now contains red dye to mark fleeing suspects

Nickvr628:
Apparently real life pepper spray contains colored dye so officers can tell who has been sprayed, this would make a good feature. If you spray a suspect and they flee, they better get to the showers or soap to clean off the red dye or they will stick out like a sore thumb. This will alert other officers to suspects that were subdued and then fled and make running away from security harder. 

image

Only the standard pepper spray canisters have the dye, the wall dispensers do not because adding several hundred units of colored dye to every wall mount would be asking for trouble. Near-infinite red paint would not be good. Once you run out of the dye mix you can get chemistry to make more or just use normal pepper spray.

:cl:
rscadd: Pepper spray now contains a marking dye to help catch fleeing suspects.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/spray.dm
+++ b/code/modules/reagents/reagent_containers/spray.dm
@@ -212,16 +212,16 @@
 //pepperspray
 /obj/item/reagent_containers/spray/pepper
 	name = "pepperspray"
-	desc = "Manufactured by UhangInc, used to blind and down an opponent quickly."
+	desc = "Manufactured by UhangInc, used to blind and down an opponent quickly. Now contains red dye to mark fleeing suspects!"
 	icon = 'icons/obj/items_and_weapons.dmi'
 	icon_state = "pepperspray"
 	item_state = "pepperspray"
 	lefthand_file = 'icons/mob/inhands/equipment/security_lefthand.dmi'
 	righthand_file = 'icons/mob/inhands/equipment/security_righthand.dmi'
-	volume = 40
+	volume = 60
 	stream_range = 4
 	amount_per_transfer_from_this = 5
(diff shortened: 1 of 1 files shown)


OppressedCactus:
Security officers are just going to make themselves shitsec-red for maximum memes aren't they...

code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -212,16 +212,16 @@
...
 	icon_state = "pepperspray"
 	item_state = "pepperspray"
 	lefthand_file = 'icons/mob/inhands/equipment/security_lefthand.dmi'
 	righthand_file = 'icons/mob/inhands/equipment/security_righthand.dmi'
-	volume = 40
+	volume = 60
 	stream_range = 4
 	amount_per_transfer_from_this = 5
-	list_reagents = list(/datum/reagent/consumable/condensedcapsaicin = 40)
+	list_reagents = list(/datum/reagent/consumable/condensedcapsaicin = 40, /datum/reagent/colorful_reagent/crayonpowder/red = 20) //red dye)
```
alexkar598, line 224: those are the tears of the clown arent they

Hopekz:
Can you shower off this dye? If not make it to where you can shower it off.

Nickvr628:
Yes you can

MegaEmpirical:
This is actually super cool.

alexkar598:
The reason why containers dont have dye is becaue nick is lazy but il take it, cant remember last time I refilled a pepperspray

MegaEmpirical:
I refill mine regularly. Gots to keep the greytide down.

PR #7445, opened 2019-12-30
Adds station map module for borgs

Ktlwjec:
I ask thee to add a cyborg module that's station map
to every modulesI ask thee to add a cyborg module that's station map
to every modulesI ask thee to add a cyborg module that's station map
to every modulesI ask thee to add a cyborg module that's station map
to every modules

:cl:  
rscadd: Borgs have a map module.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -248,8 +248,9 @@
 		/obj/item/pickaxe,
 		/obj/item/t_scanner/adv_mining_scanner,
 		/obj/item/restraints/handcuffs/cable/zipties,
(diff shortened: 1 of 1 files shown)


alexkar598 (changes requested):
That's an ugly way to do it, make it a verb at /mob/living/silicon level so that AIs and PAIs can access it too and you dont have to add it to every module

Ktlwjec:
I don't want to add it to AIs or pAIs, so close this if you won't merge it just for borgs.

alexkar598:
Then give it just for borgs, but make it a verb not a module

Ktlwjec:
effort, maybe one day

Ratvar:
And this is how alexkar made game worse

PR #7469, opened 2019-12-31
Adds categories to the mining equipment vendors

Ktlwjec:
Works with testing, prebase feature. Categorises the vendor items for easier viewing, much better than one long list. Changed for normal miners, free miners and free golems.

Before:
before
After:
unknown


code/modules/mining/machine_vending.dm
```diff
@@ -83,16 +94,30 @@
...
+		else
+			categories[prize.category] = list(prize)
+
+	for(var/a in categories)
+		dat += "<h3>[a]</h3>"
+		dat += "<div class='statusDisplay'><ul>"
+		for(var/datum/data/mining_equipment/prize in categories[a])
+			dat += "<li><tr><td>[prize.equipment_name]</td>          \
+					<td>[prize.cost]</td>          	\
+					<td><A href='?src=\ref[src];purchase=\ref[prize]'>Purchase</A></td></tr></li>"
```
Amelia0010:
```suggestion
					<div align='right'><td>[prize.cost]</td>          	\
					<td><A href='?src=\ref[src];purchase=\ref[prize]'>Purchase</A></td></div></tr></li>"
```

Amelia0010 (commented):
Try seeing if this aligns the text properly without newlining it

code/modules/mining/machine_vending.dm
Ktlwjec: image
nope :(

PR #7473, opened 2019-12-31
Emag effect for light switch

Hopekz:
Emagging a light switch will now shock people attempting to use the light switch.

Now you can do something like turn off the lights and set a trap for someone attempting to turn the lights back on.

Ea3dC9Taia
8enX0QeQRE

**Update: Flavortext changed from "Sequencer" to "emag". Not going to launch my test server for new screenshots of flavortext.**

image
image

#### Changelog

:cl:  Hopek
rscadd: You can now emag a light switch.
/:cl:


Potototatoh:
emag clown

Ktlwjec:
I'd switch "sequencer" to its full name or "emag" personally since that's what is most used. But I LOVE this pr. Also potatotahtkto stop writing nonsense on everyone's prs.

code/game/machinery/lightswitch.dm
```diff
@@ -34,8 +34,12 @@
 /obj/machinery/light_switch/examine(mob/user)
 	. = ..()
 	. += "It is [area.lightswitch ? "on" : "off"]."
+	if(obj_flags & EMAGGED)
+		. += "You hear a faint hum coming from the switch."
```
Amelia0010: What if I'm deaf?

Amelia0010 (changes requested):

code/game/machinery/lightswitch.dm
alexkar598: then its your problem now innit

alexkar598:
I don't really like this PR, I somehow dont like the idea of a free stun like that

code/game/machinery/lightswitch.dm
Hopekz: NICH !!!! CAN YOU HEAR ME!??!  I WEEENT AAAHEEEAD AND COMMITED MY CHANGE
------

Hopekz:
> I don't really like this PR, I somehow dont like the idea of a free stun like that

I respect your opinion but I think that **this stun isn't free.** 
The item to set up this stun **costs more than 5 TC** and it adheres to the usual shock rules meaning that if someone cannot get shocked or has any resistance to shocks their **resistance rules will still apply.**

**There is also a warning** in terms of examine flavor text if you're not deaf like Nich.
The very nature of this feature means that it cannot be used by non-antagonistic characters as well.

code/game/machinery/lightswitch.dm
```diff
@@ -61,3 +65,23 @@
...
+	if (electrocute_mob(user, get_area(src), src, 0.7, TRUE))
+		return TRUE
+	else
+		return FALSE
+
+/obj/machinery/light_switch/emag_act(mob/user)
+	if(obj_flags & EMAGGED)
+		to_chat(user, "<span class='warning'>Nothing new seems to happen when you swipe the emag.</span>")
+		return
+	to_chat(user, "<span class='notice'>You swipe the emag on the light switch. The light switch gives of a soft hum.</span>")
```
Ktlwjec1: The light switch gives **off** a soft hum.

code/game/machinery/lightswitch.dm
```diff
@@ -61,3 +65,23 @@
...
+	if(user.can_hear())
+		to_chat(user, "<span class='notice'>The light switch gives of a soft hum.</span>")
```
Ktlwjec1: The light switch gives **off** a soft hum.

Ktlwjec1 (changes requested):

Ratvar:
5 tc isn't really a balancing factor because traitor in fact does have 5 tc and, I fear, 15 more sometimes. How does it matter if you bought a pulserifle for 2000 bandollars if you spawn with 4000 and it's a hecking pulserifle? You don't invest anything, you start with those resources that otherwise would be entirely useless, and you cant really buy anything better.
I would like to know if alexkar will remove emagging borgs if he thinks free stuns are bad, and emagged borgs are free walking stuns that also do much more things just for 5 tc.
What I myself think about is that there's no downside in emagging light switches if you're emagging everything anyway, and that this feature won't get people who examine the lightswitch... aka it'll only work against newbies and fuck them over entirely. Is it good for a feature that fucks new players and is entirely useless against everyone else to exist?

code/game/machinery/lightswitch.dm
Hopekz: Thank you.

code/game/machinery/lightswitch.dm
Hopekz: Committed.

code/game/machinery/lightswitch.dm
```diff
@@ -34,8 +34,12 @@
 /obj/machinery/light_switch/examine(mob/user)
 	. = ..()
 	. += "It is [area.lightswitch ? "on" : "off"]."
+	if(obj_flags & EMAGGED & user.can_hear())
```
alexkar598:
```suggestion
	if((obj_flags & EMAGGED) && user.can_hear())
```

code/game/machinery/lightswitch.dm
```diff
@@ -34,8 +34,12 @@
...
+		. += "You hear a faint hum coming from the switch."
```
alexkar598: Message could use some clarification or a danger span
```suggestion
		. += "<span class="danger">You hear a faint hum coming from the switch.</span>"
```

code/game/machinery/lightswitch.dm
```diff
@@ -34,8 +34,12 @@
...
 
 /obj/machinery/light_switch/interact(mob/user)
+	if(obj_flags & EMAGGED)
+		shock(user,50)
```
alexkar598:
```suggestion
		shock(user)
```
????

code/game/machinery/lightswitch.dm
```diff
@@ -61,3 +65,23 @@
...
 	if(!(stat & (BROKEN|NOPOWER)))
 		power_change()
+
+/obj/machinery/light_switch/proc/shock(mob/user)
+	if(stat & (BROKEN|NOPOWER))		// unpowered, no shock
+		return FALSE
+	var/datum/effect_system/spark_spread/s = new /datum/effect_system/spark_spread
+	s.set_up(5, 1, src)
+	s.start()
+	if (electrocute_mob(user, get_area(src), src, 0.7, TRUE))
```
alexkar598:
```suggestion
	electrocute_mob(user, get_area(src), src, 0.7, TRUE)
```

you dont need a return for this proc as the return value isnt used anywhere

code/game/machinery/lightswitch.dm
```diff
@@ -61,3 +65,23 @@
...
+		return TRUE
+	else
+		return FALSE
+
+/obj/machinery/light_switch/emag_act(mob/user)
+	if(obj_flags & EMAGGED)
+		to_chat(user, "<span class='warning'>Nothing new seems to happen when you swipe the emag.</span>")
```
alexkar598, line 80:
```suggestion
		to_chat(user, "<span class='warning'>Nothing seems to happen when you swipe the emag.</span>")
```

code/game/machinery/lightswitch.dm
```diff
@@ -61,3 +65,23 @@
...
+		return
+	to_chat(user, "<span class='notice'>You swipe the emag on the light switch. The light switch gives off a soft hum.</span>")
```
alexkar598:
```suggestion
	to_chat(user, "<span class='notice'>You swipe the emag on the light switch. </span>")
```
The hum is later

alexkar598 (changes requested):

code/game/machinery/lightswitch.dm
Hopekz: No. 
An item humming isn't considered something an IC person would consider dangerous. 
Do you consider a switch dangerous if it gives off a soft hum?
The spans are there to convey IC perceived danger.

code/game/machinery/lightswitch.dm
Hopekz: No.
Your change doesn't make sense because in order to arrive to this message you need to already swipe it once to get an effect. This conveys that the same result is achieved with nothing "new" visible.

Hopekz:
Oh boy. Wonder why Travis doesn't like this PR after I took your changes..

Let me know if I should revert some suggestions to get stable again otherwise keep em coming buddy because I have no idea where you're going with some of these!

Otherwise I'll get on my main PC sometime to try to make these changes work..

code/game/machinery/lightswitch.dm
```diff
@@ -61,3 +65,23 @@
...
+/obj/machinery/light_switch/proc/shock(mob/user)
+	if(stat & (BROKEN|NOPOWER))		// unpowered, no shock
+		return FALSE
+	var/datum/effect_system/spark_spread/s = new /datum/effect_system/spark_spread
+	s.set_up(5, 1, src)
+	s.start()
+	electrocute_mob(user, get_area(src), src, 0.7, TRUE)
+		return TRUE
+	else
+		return FALSE
```
Amelia0010: Remove these 3 lines, that's what alex said.

Amelia0010 (changes requested):

code/game/machinery/lightswitch.dm
alexkar598: Might as well make it have no way to know its emagged then, locking things like this to only players who read PRs is a bad thing as it makes new player unable to tell what is bad while more code savvy players can bypass traps

Hopekz: While I agree with my original point I am not closed-minded and I can see how this being a regular piece of text can make it unfair to those people per your point. 
There is no middle compromise here (except something like a notice span) so I'll go ahead and fully implement your suggestion.

code/game/machinery/lightswitch.dm
Hopekz: Commited. Also fixed compiling issues.
Thanks

PR #7475, opened 2020-01-01
Sink and Shower are now Craft-able

Hopekz:
Yogs doesn't believe in hygiene so this is the first very important step to make yogs less stinky
Why did no-one do this yet

#### Changelog

:cl:  
rscadd: Sink is now craft-able. Wow!
rscadd: Shower is now craft-able. Wow!
/:cl:


boodaliboo:
I am stink

gilvian:
To what the sink/shower will be conected? Can i just build a sink in the middle of space for it to work?

overlord12345:
what about toilets

PR #7490, opened 2020-01-03
Makes me codeowner of all maps

Ktlwjec:
image


LoliconSlayer:
Why?

Ktlwjec:
I'm headmapper but only codeowner for station maps currently, makes it easier for me since it'll give me an email for all map changes.

ThatLing:
no woman can hold so much power

Hopekz:
+1 
ktl has basically unofficially had this role anyway might as well write it in.

PR #7502, opened 2020-01-04
Changes battlemage armor and wizard's hardsuit.

Takahiru:
This changes the battlemage armor and the wizard's hardsuit to have better specialization.

**The battlemages armor**
    It has much higher defense values, but the space protection has been taken away. 

**The wizard's hardsuit**
    It has lower defense values, and it has no slowdown, along with the space protection it had already.

This changes both to be more specialized, the battlemage armor is more combat focused, while the wizard's hardsuit is for avoiding the vacuum of space.

:cl:  
tweak: Changes the battlemage armour and the wizards hardsuit. 

/:cl:


overlord12345:
the madman did it

code/modules/clothing/suits/wiz_robe.dm
```diff
@@ -183,7 +183,7 @@
 
 /obj/item/clothing/suit/space/hardsuit/shielded/wizard
```
alexkar598: This is going to need repathing to not be a suit/space/hardsuit subtype

AsV9: @Takahiru

alexkar598:
???
Screenshot_20200115-060318_Firefox Nightly

Takahiru:
wait wtf did i do that wrong

PR #7507, opened 2020-01-05
Adds a few CHECK_TICKs to monstermos

MCHSL:
image


alexkar598:
What could possibly go wrong

PR #7510, opened 2020-01-05
Blue lights with bad atmosphere

yacabo:
Lights turn blue when their local air alarm deems the local atmosphere to be severe. This will mostly become apparent for vacuums but includes toxic atmospheres as well, e.g. plasma in the air. Fire alarms still take priority in turning the lights red, but if there's a fire there then there's probably toxic gas as well.

image

It also adds a new icon state for the lights so that they're blue as well

:cl:  
rscadd: Blue warning lights turn on when a hostile atmosphere is detected
/:cl:


Jarod1200:
can we make it a lighter shade of blue so its easier to see?

yacabo:
That is most certainly something I will probably take a look into

soon

Ratvar:
Maybe it should be atmospherics-yellow? This way it's closer to "red danger" lights, yellow code, atmos department and piss at once

Readystorm546:
> Maybe it should be atmospherics-yellow? This way it's closer to "red danger" lights, yellow code, atmos department and piss at once

He showed me that and its shit
image

Hopekz:
I am fine with this idea but I don't like how dark this color is atm.
It's like being in a rad storm or ash storm.
Going to go against this until its pretty like something that should be in the game.

Ratvar:
If hopek've read other comments he would see that that's the planned feature. Sad he didn't

MegaEmpirical:
What is the trigger?

This shouldn't happen if the air alarm settings say it's fine, permitting custom rooms.

yacabo:
It works directly thresholds, will can be adjusted both ways.

In this case I put the min2 pressure to slightly above room pressure which set off the blue lights:

image

The same thing can be done with gasses such as in this case with plasms max2 was lowered to -1 allowing any amount of plasma to come into the air without triggering the lights:

image

So you are able to customize what you want in the air and what you don't want in the air (including the air itself) however you may see fit

MegaEmpirical:
Good.

Hopekz:
> If hopek've read other comments he would see that that's the planned feature. Sad he didn't

I did read it.
There is nothing wrong with reinforcing my viewpoint to convey how important it is to me stop trying to silence me.

-------------------

Anyway, I like the new commit so I'm changing to supporting this Pr because it won't look like I'm in the middle of an ash storm now.

yacabo:
The light is now a much lighter shade of blue increasing visibility, I replaced the original picture.

MegaEmpirical:
What happens when power is out? Red lights? Blue? PURPLE?

yacabo:
image
>Red

MegaEmpirical:
Purple.

Jarod1200:
pretty sure if there is no power the lights are off

yacabo:
Lights actually have a battery installed in them so that they can emit a faint red light if their APC looses power.

Hopekz:
I personally like the lights being in their usual battery mode instead of purple without power.
I'd like to think that lights without power or an APC can't communicate to the room sensors so they default to their "basic" state.

I don't picture normal lights to be "smart" lights with their own logic.

MegaEmpirical:
Jarod, what kind of CE are you?

Hopek is right, I concede.

Jarod1200:
yes I know they turn red. that was kind of a meme  like a weather rock: if the rock is wet its raining.

alexkar598 (approved):

alexkar598:
Don't really like the intensity this has, maybe something with less light_range and light_power? Maybe the same power as the lights on emergency power mode

yacabo:
I personally like it how it is, has enough visibility so that you can see what you want to see but still has that engrossing effect that reminds me that I'm as close to space as I can get without actually truly being in it, but that's just me.

Ignoring that though, messing around with the power is a wee bit funky, setting it to 0.45 gives this look:

image

and I didn't even notice a difference until I had a side by side comparison and the actual difference is that it's slightly darker, hardly a difference. Setting it all the way down to 0.2 reduces the lights so much that we come back to the issue that Jarod1200 and Hopekz but brought up in that we can no longer see anything but this time it's just because there is no light. So let's make the shade lighter from #4F82FF to #7FA4FF which comes out to this:

image

which washes the colour out a bit and then combining the two methods:

image

simultaneously makes the colour brighter and darker, go figure.
For comparison this is what a fire alarm looks like:

image

Anyway I think I just compiled like ten times trying to get this right and I still like the current version, which may or may not be a bias, so lemme know if which one is the best or if I should try a different method.

alexkar598:
Il try it out and tell you what i find to look nice

yacabo:
Roger

yacabo:
Updated brightness

AsV9 (approved):

PR #7512, opened 2020-01-05
syringle/autoinjectors resprite

Potototatoh:
this alters/replaces the old syringe sprite that i totally made

Screenshot (664)
disclaimer; updated

crediting to _mystery person who contributed to not this station_ for sprite from loveweb
#### Changelog

:cl:  
bugfix: fixed the syringe looking like dookie
imageadd: replaces the syringe and autoinjectors
imagedel: deleted some icons and images 
/:cl:


Potototatoh:
krudox kruo

Readystorm546:
Does this also change the needles you get from genetics?

Hopekz:
I personally don't like these sprites as much as what we already have in-game but I don't "hate" them so I'm neutral.

Not sure what doesn't sit right with me but a thing I notice is that the needle seems to look really long in this orientation. It's like almost the full length of the whole syringe.

I also don't like how the used pens don't look more "used" at first glance. 
One of the nice things about for example the EpiPen used sprite is that it is really easy to tell if an EpiPen was used before choosing to subconsciously ignore it as junk. 
I didn't get that same "know its used" feeling from your sprite I kind of feel I need to look at the tip of the pen to see if it's used rather than the whole sprite itself.

Just my opinion still neutral will lurk if this gets updated.

Sir-Lagsalot:
I like the new syringes, but the new pens look a bit off IMO. Also they're smaller so harder to pick up in a hurry.

alexkar598:
Traumatizing

Potototatoh:
fixed

nesisknesis:
yeah ok

PR #7513, opened 2020-01-05
Language holder for Preternis fix

Hopekz:
This adds a language holder to Preternis because TGuinext didn't support Preternis **(because TG doesn't have Preternis)**

I specifically fixed Preternis because it was breaking sheltered. (they spawned in with Japanese instead of encoded audio language) but this will also fix future Preternis language bugs as well.

#### Changelog

:cl:  
rscadd: Makes Preternis compatible with tgUI's language holder.
bugfix: fixed sheltered Preternis spawning in with Japanese.
/:cl:


alexkar598:
Beep Boop approves

PR #7528, opened 2020-01-06
[DNM] Adds experimental topic spam limiter 

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Replaces BYOND's built-in DOS prevention mechanism (which is broken) with my own implementation. Once the server receives 2 topic packets within less than 50 miliseconds (can be changed as needed), the sender is blocked until server restart. Filtering is near instant and attacks do not have any impact on gameplay.

[implementation](https://github.com/MCHSL/extools/blob/master/byond-extools/src/core/core.cpp#L440)


KingTerrytheTerrible:
lul

apocist:
I'd recommend a making the system more lenient so that it at least provides some more leeway for when someone accidentally sends too much data. Something maybe that it could give shorts blocks/ignores with an increasing longer time frame per occurrence. 2 packets at once sounds like it'd get a ton of false positives.

(thinking of fail2ban)

PR #7545, opened 2020-01-08
Adds a cheese bar

Ktlwjec1:
Here is a customer paying for his cheese.

Here is the bartender and PunPun.

Making wine.

Cured meats.

Sheep.

:cl:  ktlwjec
rscadd: Adds a cheese bar.
/:cl:


nesisknesis:
add more crackers so i can feed them to poly

Ktlwjec1:
That is exactly why there is limited crackers.

PR #7553, opened 2020-01-09
Lag removal, try 2

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Does that thing with the thing I did once where a dll measures time taken by byond to send map updates and it's included in master controller calculations so lag spikes are reduced but not completely eliminated cause its not clairvoyant so don't be upset if you get a lag spike caused by the shitty database driver lib or tgui or garbage collection and then you die because of it

[cod](https://github.com/MCHSL/extools/blob/master/byond-extools/src/maptick/maptick.cpp)

Don't testmerge along with dos protection it might explode


1fbff5f83b23d39d38b1dfcb4cac8d9b:
I forgot to switch branches and only partially successfully tried to move my changes

Hopekz:
Travis is upsetti

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Who cares

Bibby0110:
make it stop

optimumtact:
The funny thing about this is the more common lag you eliminate, the more annoying the edge case spikes become.

MCHSL:
bsql is killing me

alexkar598:
person coder told me its safe to merge

AshCorr:
Code needs to be added to the yogs repo.

PR #7555, opened 2020-01-09
Reversing the order of garbage collection checks to prevent runtimes

LudwigVonChesterfield:

Due to the order of performed checks(first try to get the var of an object by link, and then checking if a link is null) a runtime could occur, this fixes it.

### DIFF
--- a/code/controllers/subsystem/demo.dm
+++ b/code/controllers/subsystem/demo.dm
@@ -177,7 +177,7 @@ SUBSYSTEM_DEF(demo)
 		last_completed++
 		var/atom/movable/M = marked_dirty[marked_dirty.len]
 		marked_dirty.len--
-		if(M.gc_destroyed || !M)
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/demo.dm
```diff
@@ -421,7 +421,7 @@ SUBSYSTEM_DEF(demo)
 /datum/controller/subsystem/demo/proc/mark_dirty(atom/movable/M)
 	if(!isobj(M) && !ismob(M))
```
LudwigVonChesterfield, line 422: I am also curious as to whether why in places like this istype(M) or ismovable(M) is not used?

monster860: because lighting objects are a no-no for now.

monster860 (approved):
another reason to insult /tg/station

PR #7558, opened 2020-01-10
Removes katana from uplink

alexkar598:
It is ridiculously OP, along with a riot shield(which basically anyone can get by breaking into sec), you are invul to any melee attacks including stun batons.
It takes a single hit to delimb/decapitate someone
It takes three hits to kill someone
It is impossible to "know" its fake as it is metashielded and by the time you can get into range to get told its fake by the impaling of your body. Its too late to use ranged stuns and melee stuns are ineffective.

Now about the people who will say improve dont remove, the katana is simply trying to fit in with the other traitor weapons despite other actually balanced weapons being able to do this:
Dual Esword: Main Weapon (What katana is, does massive damage)
Esword: Side Weapon (What katana is, can be held in a single hand)
Edagger: Stealth Weapon (What katana is, is metashielded)

#### Changelog

:cl:  
rscdel: The fake toy katana is no longer available for traitors to buy  
/:cl:

Also no I wasn't killed by it recently and no it taking up the belt slot is not a valid argument.

### DIFF
--- a/yogstation/code/game/objects/items/weaponry.dm
+++ b/yogstation/code/game/objects/items/weaponry.dm
@@ -23,6 +23,3 @@
 	else
 		user.visible_message("[user] is left hanging by [target].", "<span class='notice'>[target] leaves you hanging.</span>")
 		playsound(src, 'sound/weapons/punchmiss.ogg', 50, 0)
-
-/obj/item/katana/faketoy
-	name = "replica katana"
--- a/yogstation/code/modules/uplink/uplink_item.dm
+++ b/yogstation/code/modules/uplink/uplink_item.dm
@@ -52,13 +52,6 @@
 /////////New Items////////
 //////////////////////////
 
-/datum/uplink_item/dangerous/katana
-	name = "Katana"
-	desc = "The katana is an edged weapon with a blade of pure degeneracy. While more robust than an energy sword, it cannot be as easily sheathed and hidden. At a glance, it looks like a fake katana; you can tell from the pixels."
-	item = /obj/item/katana/faketoy
-	cost = 10
-	exclude_modes = list(/datum/game_mode/nuclear/clown_ops)
-
 /datum/uplink_item/stealthy_weapons/door_charge
 	name = "Explosive Airlock Charge"
 	desc = "A small, easily concealable device. It can be applied to an open airlock panel, booby-trapping it. \


ynot01:
i ded label inbound

Readystorm546:
Thumbs Down Speed-run

alexkar598:
Be honest, is this weapon balanced in your mind. If so, what drawbacks make it actually balanced

Alexandria404:
after considering it, the PR seems understandable

MegaEmpirical:
I've never had a good encounter with a katana, they looked broken on release.

Lynameep:
Give it the TF2 treatment where it can't be sheathed/dropped/thrown until you kill someone with it.
Give it a new sprite that isn't meta shielded once you kill someone/something with it.
Make it a 2-Handed weapon.

alexkar598:
Thats just a reskinned dual esword, its not anything so unique its worth keeping

ghost:
@alexkar598 you could alternatively do what i suggested and make it a delimb-focused tool as it is on goon, thus adding a new item to the traitor uplink while reducing their available murderbone pool and separating this item from just being a better esword

frankly you could resolve the entire metashield problem by just fucking coloring it red and calling it a syndicate katana

believe me normally i am all in favor of removal is improval but in this case the solution is so utterly simple that 'improve it' is actually a valid response for once because a viable means to improve it simply and effectively and resolve all 3 of the above issues is already available

alexkar598:
Delimb weapons still feel extremely overpowered, it takes fucking years to get a limb back. Also it would delay getting rid of katana, since there is no content here, it can be removed and later implemented

ThatLing:
Why can't you just rework it now? Why do you have to remove it first?

MegaEmpirical:
I'm okay with trying out the delimb focused version to see how things go.

Hopekz:
@Idinuum got any cool red katana sprites for the "syndicate katana" ?

ghost:
Delimb weapons wouldn’t be overpowered. Eswords dismember in one, two, or three hits. You can get any limb cut off in seconds.

It takes longer to clone someone or defib them than it does to get a limb reattached, given that it’s a nearly instant surgery. 

And while you’re waiting for the PR to be merged, given it has been four days, that’s four days you could be working on the alternative instead.

What would you prefer, a five hit kill syndicate weapon with no blockchance that does almost no direct damage or an energy sword that three hit kills with massive AP and a heavy blockchance. Which sounds more overpowered to you? Because you don’t need to dismember someone if they’ve fucking died.

Hopekz:
@AdamElTablawy 
So are you going to code this new idea ooooor

ghost:
@Hopekz why is it my responsibility to code the better idea for a server i don't play on?
i'm just telling you a better solution. it's your decision if you want to make the right decision or a bad one.

ynot01:
i like this i dont know why it was marked as spam

inb4 MURDERBONER

alexkar598:
Wasn't me

boodaliboo:
initially i was against this and but im for this now i would prefer it to stay but have a more conspicious name because currently its name is completely identical to they toy one even though it is in no way metashielded and wasnt intentended to be so

MegaEmpirical:
Having the same fucking name for how brutal it is absurd. Does it have the same description?

alexkar598:
Well its not completely identical, it lacks fireproofing, the fake one that is

SomeguyManperson:
the main difference is the fake katana is a normal sized item while the real one is bulky, but it's metashielded because lol so that doesn't matter

Hopekz:
> the main difference is the fake katana is a normal sized item while the real one is bulky, but it's metashielded because lol so that doesn't matter

I'm not debunking your point just pointing out that this is very nit picky and doesn't really serve to "balance" this as it is not apparent in-game.
I have close to 1k rebase playtime hours and I didn't even notice this differentiation enough to use this to identify between the two (not that I would be able to because rules).
We should balance these types of items in our game based around what the average player can obtain without code-diving for something that just seems like a potential oversight.

I know this isn't an oversight but using this to differentiate between the two is very meta-y . 
If we're going to use this as an argument I'd want to add flavortext to the syndicate katana that specifically points out that it seems to be larger (instead of just saying its a bulky item)

PR #7559, opened 2020-01-11
Fixes toilets and also this is why im a better maintainer

alexkar598:


Hopekz:
Hey, why are you removing my comment put it back -1
otherwise, good catch IDK why I didn't notice it too.
Also, ASV isn't a maintainer

PR #7560, opened 2020-01-11
Revert "Removes implant breaker from the gang tool"

fluffe9911:
Im just gonna let this sit here for a bit until people deal with gangs not having implant breakers enough to form a opinion on it but yeah I think its bad idea and all that

### DIFF
--- a/yogstation/code/game/gamemodes/gangs/gang_items.dm
+++ b/yogstation/code/game/gamemodes/gangs/gang_items.dm
@@ -265,6 +265,17 @@
 	cost = 18
 	item_path = /obj/item/grenade/syndieminibomb/concussion/frag
 
+/datum/gang_item/equipment/implant_breaker
+	name = "Implant Breaker"
+	id = "implant_breaker"
+	cost = 20
+	item_path = /obj/item/implanter/gang
+	spawn_msg = "<span class='notice'>The <b>implant breaker</b> is a single-use device that destroys all implants within the target before trying to recruit them to your gang. Also works on enemy gangsters.</span>"
+
+/datum/gang_item/equipment/implant_breaker/spawn_item(mob/living/carbon/user, datum/team/gang/gang, obj/item/gangtool/gangtool)
+	var/obj/item/O = new item_path(user.loc, gang)
+	user.put_in_hands(O)
+
 /datum/gang_item/equipment/wetwork_boots
 	name = "Wetwork boots"
 	id = "wetwork"


Bibby0110:
good pr

swissloaf:
good pr

boodaliboo:
> good pr

hi

Hopekz:
Going to put my opinion on this now that I've had a time to play with this new change.
They're pretty good I had a chance to play as both a gang member and mind-shielded crew (not sec yet with gang)
-1 on this revert because this change is not bad.

SomeguyManperson:
as a gang leader no implant remover means you're highly likely to be murdering security and that it's almost impossible to convert enemy gangsters since you'd need to implant them, then surgically remove the implant. I would at the least like a way to convert enemy gangsters since otherwise they just get shanked and hidden in maint

Hopekz:
> as a gang leader no implant remover means you're highly likely to be murdering security and that it's almost impossible to convert enemy gangsters since you'd need to implant them, then surgically remove the implant. I would at the least like a way to convert enemy gangsters since otherwise they just get shanked and hidden in maint

I kind of like the balance this creates because before this you would get into situations where you'd have gang convert the entire crew or even the entire gang so at the end you would have situations where it would be 20 gang members vs the 4 remaining crew.

Having the crew die instead means they're taken out of shift meaning that you run into more reasonable scenario's where its around 5 gangmembers vs 4 remaining crew.

Even in the shift yesterday where you were a gang lead (even though not much opposition was met) I'm pretty sure if you were able to convert that shift all of a sudden you'd have the whole crew in a gang on that dom like previous shifts.

Taking players out during gangs is not a bad thing imo when compared to the alternative of  "everyone in 1 gang" thing.

I personally would prefer these gangsters dying off instead of forming 1 unstoppable mega gang with guns. I'm not sure what can be done to make this it more of a fair fight since it does really seem 1 sided to the gang at the moment.

SomeguyManperson:
Needing to kill gangsters or mindshielded crew means they are taken out of the round *entirely*, and are likely to be made unrecoverable to prevent the enemy gang from getting back members. Since dominators are still dangerous to place this means that even more people are going to be sitting around in deadchat waiting 30 minutes for the winning gang to be ABSOLUTELY SURE they can place it

ynot01:
tbh i think its better that mindshielded crew are killed rather than converted, prevents the gangs from converting and making themselves even larger and more powerful steamrolling now with trusted crew (to everyone without sec hud)

Ratvar:
<Since dominators are still dangerous to place this means that even more people are going to be running around whole station for 30 minutes for the winning gang to be ABSOLUTELY SURE they can place it> - SomeguyManperson, rephrased to make a counterpoint

alexkar598:
I uh what is this, its been on DNM with no development for three weeks

PR #7564, opened 2020-01-11
"speaking of useless can we get chat notifications for when someone is inspecting something"

monster860:
image

its literally a one-line change so meh

:cl: monster860
add: examine makes a visible message
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -304,7 +304,8 @@
 		to_chat(src, "<span class='notice'>Something is there but you can't see it.</span>")
 		return
 
-	visible_message("<span class='notice'>[src] examines [A]!</span>")
+	if(isturf(loc) && (isturf(A) || isturf(A.loc)))
+		visible_message("<span class='notice'>[src] examines [A]!</span>")
 	face_atom(A)
 	var/list/result = A.examine(src)
 	to_chat(src, result.Join("\n"))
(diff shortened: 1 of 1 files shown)


Darkstick:
I'm often examining people in secret to know if they're dangerous or not, if they were aware that I was examining them they could get suspicious very fast.
I also feel like it would be annoying for the chat bar to be filled with "X examined X" whenever it gets a bit crowded
and one last point, when was someone examining you in secret ever a problem? Why is potatoh being angry at assistants examining him?

Ktlwjec1:
This'd get spammy with how much I examine shit.

Potototatoh (changes requested):
`visible_message("<span class='notice'><small>[src] examines [A].</small></span>")`

this so it isn't horrendous to look at when spammed and maybe stick a color onto it aswell

Altoids1:
I literally burst into laughter at the thought of this.

This *has to be* a huge spamfest. 30 people in one room doing an examine every second means 30 messages per second of people just plug lookin' about the place. Hilarious amounts of spam and messages for very little information.

This *could be* tolerable if it were only a message you got if someone just kept examining you, over and over, in front of you, but besides that the concept is just not suited for the limited space of the chatbox.

Potototatoh:
hey
im lookin here

AsV9:
Big think

PR #7569, opened 2020-01-11
Replaces firelocks with fans (whiteship_meta)

Ktlwjec1:
stops this
cNrWYkb


AsV9:
fans = bad but i don't see a better solution tbh

Ktlwjec1:
Agreed tbh

PR #7572, opened 2020-01-11
 Replaces secborg disabler with a forensic scanner

AsV9:

I wanted to shift secborg into a more utility based role, so I decided to remove it's disabler. Instead  the secborg now has a forensic scanner.

Also removed the actual disabler from the code, since it's not in use any longer. And the disablercooler upgrade is also dead.

This is up for debate, obviously.

:cl:
rscadd: Added a forensic scanner to the secborg loadout
rscdel: Removed secborg disabler
/:cl:

### DIFF
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -88,35 +88,6 @@
 	if (.)
 		R.remove_movespeed_modifier("VTEC")
 
-/obj/item/borg/upgrade/disablercooler
-	name = "cyborg rapid disabler cooling module"
-//	desc = "Used to cool a mounted disabler, increasing the potential current in it and thus its recharge rate."
-	desc = "It used to give unspeakable power to security modules. Now it rests; broken, abandoned."
-	icon_state = "cyborg_upgrade3"
-	require_module = 1
-	module_type = /obj/item/robot_module/security
-
-/obj/item/borg/upgrade/disablercooler/action(mob/living/silicon/robot/R, user = usr)
-	. = ..()
-	if(.)
-		var/obj/item/gun/energy/disabler/cyborg/T = locate() in R.module.modules
-		if(!T)
-			to_chat(user, "<span class='notice'>There's no disabler in this unit!</span>")
-			return FALSE
-		if(T.charge_delay <= 2)
-			to_chat(R, "<span class='notice'>A cooling unit is already installed!</span>")
-			to_chat(user, "<span class='notice'>There's no room for another cooling unit!</span>")
-			return FALSE
-
-		T.charge_delay = max(2 , T.charge_delay - 4)
-
-/obj/item/borg/upgrade/disablercooler/deactivate(mob/living/silicon/robot/R, user = usr)
-	. = ..()
-	if (.)
-		var/obj/item/gun/energy/disabler/cyborg/T = locate() in R.module.modules
-		if(!T)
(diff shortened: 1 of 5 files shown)


Ratvar:
They still have stunbaton, making them into a validhunt only machine they always will be if they have stun weapons at all

AsV9:
Well you know, @Ratvar, baby steps

Ratvar:
Why baby steps? This is literally just a worse Detective Borg PR that compromises with people who should not be compromised with. Why close it if you do the same but worse?

Moja10:
>PR makes progress to a goal you like
>it's not the exact solution you want therefore the PR is without merit and accomplishes nothing

why are you like this literally all the time for every single coding solution that isn't your exact verbatim solution to the problem? why are you such a baby? why can't you just learn to code using the myriad of resources so you can implement your own exact changes instead of just going "not my solution try again :)" to every single change.

MegaEmpirical:
I think replacing the baton instead might make more sense as it is an instant stun, rather than the disabler. Hmmm can the disabler pistol whip?

ynot01:
either nerf the stun baton or replace the baton because the most convenient method of dealing with cyborgs is getting in melee range of them

ghost:
yknow you could have just ported tgs stunbaton change and avoided this issue

ghost:
fuck off ratvar you retard you agreeing with me makes my points lose legitimacy

anyway the primary issue with secborgs is best explained that they're better at murder than literal syndicate combat cyborgs due to their stunbaton

tgs stunbaton change can fix that, but removing their disabler simply changes the borg into effectively a game of rocket tag-like fighting a malf engiborg, you either flash it or it stuns you.

it also doesn't address that the secborg is basically just a module soley designed to validhunt with nothing else it can possibly do or contribute to. if you add a detective scanner, it's just stealing the detective's job outright and now he has nothing to do, since you pretty much only need one detective as opposed to every other borg module, where having the borg does not invalidate the entire job.

my personal suggestion tho is that you do what every person ever has ever suggested about borgs and make them slow down when damaged like humans do. then they can keep the disabler and the baton because now they can be killed much easier even if you aren't carrying one of their hard counters on you (and thus leaves the door open to softening those hard counters some)

Szyszkrzyneczka:
lets just do what every goddamn server did, remove sec borgs and add cargo borgs instead

MegaEmpirical:
Tell me more about these cargo borgs.

maxion12345:
Bro what if we made borgs modular, like they have points and shit to pick modules then robotics can like upgrade them to let them get more modules and it'd never work because current modules have like 90 different tools for one Borg and what is this PR for? Wait aren't we adding detborgs? This is replacing secborg but with extra steps c'mon man. Also Borg damage slowdown when? And Borg disabler baton seems like a nice option, as borgs can't be disarmed. I wouldn't give normal sec it because they'd get rng disarmed and beaten to death during the time it would take to stun someone with it tbh, but borgs can't be rng disarmed and beaten to death with their own gear so hey, win-win.

alexkar598:
@maxion12345 you meant like goon?

AsV9:
Closing in favor of #7588, might add a forensics scanner at a later time.

Hopekz:
Yeah good call I was about to say is #7588 gets merged it might not be a good idea anymore to merge this as well.
Hard to tell so if Alex's PR gets merged we'll have to play around with it to see where secborgs are in terms of balance.

PR #7575, opened 2020-01-11
Conveyor recycling.

DerelictDrone:

Makes it so that conveyors can be recycled, in the event you only needed 26 conveyors but made 30 to get the job done faster.

Switches unfortunately cannot be recycled yet, since if you upgrade an autolathe the cost of it changes. But if you pry the switch back up the materials in the switch would reset back to the default amount of mats.

#### Changelog

:cl:  
rscadd: Conveyors may be recycled now.
/:cl:


deathslayer2481:
Cofftopia powercreep intensifies

PR #7599, opened 2020-01-13
Unlinking Conveyors and Switches.

DerelictDrone:
Using a multitool, you can now unlink conveyor assemblies or randomly generate a new ID for switch assemblies.

#### Changelog

:cl:  
rscadd: Use a multitool on conveyor assemblies to unlink them from all switches, or use a multitool on a switch assembly to randomly generate a new ID for linking conveyors.
/:cl:

### DIFF
--- a/code/modules/recycling/conveyor2.dm
+++ b/code/modules/recycling/conveyor2.dm
@@ -369,7 +369,6 @@ GLOBAL_LIST_EMPTY(conveyors_by_id)
 	if(I.tool_behaviour == TOOL_MULTITOOL)
 		id = "[rand()]"
 		to_chat(user, "<span class='notice'>You pulse \the [src]'s connection, randomly generating a new network ID.</span>")
-	return
 
 /obj/item/stack/conveyor
 	name = "conveyor belt assembly"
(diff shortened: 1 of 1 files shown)


ynot01:
really enjoying the recent conveyer love

alexkar598:
But why

DerelictDrone:
@alexkar598
Generating a new conveyor switch ID would allow you to re-use switches without fear that a conveyor somewhere on the station will also be affected, or having to pry up every single conveyor it may or may not be connected to just to unlink them.

Wherein unlinking individual conveyors would help save a player's time, rather than forcing them to just dump all the conveyors they want to re-use elsewhere into an autolathe to re-build them, or for instance if you separate a stack of conveyors, but link the wrong half of the stack.

Either way, these are some very easy quality of life changes that have been missing for far too long.

code/modules/recycling/conveyor2.dm
```diff
@@ -365,6 +365,12 @@ GLOBAL_LIST_EMPTY(conveyors_by_id)
 	transfer_fingerprints_to(NC)
 	qdel(src)
 
+/obj/item/conveyor_switch_construct/attackby(obj/item/I, mob/user, params)
+	if(I.tool_behaviour == TOOL_MULTITOOL)
+		id = "[rand()]"
+		to_chat(user, "<span class='notice'>You pulse \the [src]'s connection, randomly generating a new network ID.</span>")
+	return
```
alexkar598: No need for a return statement here

alexkar598 (changes requested):

PR #7613, opened 2020-01-13
Fixes #7540

### LINKED ISSUE #7540: Magic Missile stuns through magic immunity
## Round ID: 
N/A, been like this for months

## Testmerges:
N/A, no current testmerges have had any effect on it

## Reproduction:
I have only experienced this in the wizard academy space ruin while wearing a gem encrusted hardsuit and being shot with magic missiles by the NPC wizards (only reliable place ingame for it to happen without admin powers). Every single time I have been shot with a magic missile, it displays the message for the magic being cancelled but it stuns regardless.

alexkar598:
Fixes #7540

### DIFF
--- a/code/modules/spells/spell_types/projectile.dm
+++ b/code/modules/spells/spell_types/projectile.dm
@@ -44,7 +44,7 @@
 	. = ..()
 	if(linger && target != original)
 		return FALSE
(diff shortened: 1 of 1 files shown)


monster860 (changes requested):
Get a better title

PR #7614, opened 2020-01-13
Fixes #7454

### LINKED ISSUE #7454: Civilian door remote does not work on cargo doors
Cargo is a civilian department. HoP only gets the Civ remote, Hop is in charge of cargo.

This is madness.

alexkar598:
Fixes #7454


monster860 (changes requested):
Get a better title

PR #7615, opened 2020-01-13
Fixes #7123

### LINKED ISSUE #7123: PA Rifle aiming laser breaks telescopic shields.
Round ID : 26450

Reproduction : Aim laser rifle at a telescopic shield that is being held, watch it disintegrate.

alexkar598:
Fixes #7123


monster860 (changes requested):
Get a better title

PR #7618, opened 2020-01-14
Makes all blobs block atmos

monster860:
Pros: Blobs actually have to try in order to depressurize an area
Cons: they are now invulnerable to plasma fires and shizz.

:cl: monster860
tweak: Blobs block atmos
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/structures/_blob.dm
+++ b/code/modules/antagonists/blob/structures/_blob.dm
@@ -17,7 +17,7 @@
 	var/heal_timestamp = 0 //we got healed when?
 	var/brute_resist = 0.5 //multiplies brute damage by this
 	var/fire_resist = 1 //multiplies burn damage by this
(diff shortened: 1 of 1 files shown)


MegaEmpirical:
I quite like this, makes it more interesting.

Hopekz:
Yeah, monstermos made blobs really stupidly strong with this. Hopefully, it will knock them back down to the point where you can actually fight them as crew.

I haven't seen someone try to plasma flood a blob yet so not sure if its really a con

alexkar598:
This actually buffs plasmafloods, plasma floods against blob are inefficent as the plasma gets instantly cooled and spaced. With this, it actually does something before it gets to space

ynot01:
does this fuck over PKAs?

HomocidlePancakes:
Local blob removes one of his tiles and yeets all of air in to space.

AsV9:
I guess we can try

PR #7621, opened 2020-01-14
Fixes #3020

### LINKED ISSUE #3020: incorporeal move doesn't seem to work for poly's ghost
Round 19068

Poly's ghost has incorporeal movement set to 1, which is supposed to allow poly to pass through dense turfs.  It does not pass through dense turfs or objects, however.

https://github.com/yogstation13/Yogstation-TG/blob/e2a2ccd4dc53d5b4ce2b38f82a986a59fd907e2b/code/modules/mob/mob_movement.dm#L65

This is why it's broken.

alexkar598:
Fixes #3020


monster860 (changes requested):
Get a better title

PR #7630, opened 2020-01-14
fix people being able to open PDAs from afar using ctrl click

### LINKED ISSUE #4933: PDA can be control clicked from afar to open it
You still can't interact with it but you could use this to read messages from afar if someones history is open.

monster860:
fix #4933


MegaEmpirical:
Fucking finally

PR #7632, opened 2020-01-14
fixes ghosts interacting with ore boxes and tank dispensers

monster860:
I believe this marks 10 bug fixes. No bug ID for this one, I dont think.


boodaliboo:
i am bug id

PR #7646, opened 2020-01-14
The same person can no longer both confirm and trigger keycard-auth events

AsV9:


monster860:
aw dang i cant do the ol hit the button on one run to the other while seitching cards and pressing the other button :(

code/modules/security_levels/keycard_authentication.dm
```diff
@@ -101,6 +101,10 @@ GLOBAL_DATUM_INIT(keycard_events, /datum/events, new)
 	event_source = null
 
 /obj/machinery/keycard_auth/proc/trigger_event(confirmer)
+	if(confirmer == triggerer)
```
Amelia0010: who the fuck came up with these variable names

code/modules/security_levels/keycard_authentication.dm
```diff
@@ -101,6 +101,10 @@ GLOBAL_DATUM_INIT(keycard_events, /datum/events, new)
...
+		message_admins("[ADMIN_LOOKUPFLW(triggerer)] attempted to both trigger and confirm event [event].")
```
Amelia0010: does this really need to message admins?

AsV9: well you know, fucking cheaters innit

code/modules/security_levels/keycard_authentication.dm
AsV9: /shrug

code/modules/security_levels/keycard_authentication.dm
AsV9: well you know, fuckin cheaters innit

alexkar598:
Honestly this could just be avoided by moving them apart, I like the idea of someone using portals to bypass a security feature, only bad part in this is the fact you can just walk to the other one and get it in time

zxmongoose:
You could also just remove one of the two buttons from the bridge completely. The side-by-side activation aspect is never really RP'd (I'm assuming this is the point of the placement), and the game wouldn't be lacking at all for missing the second one on the bridge. Every head has one in their office, and if there isn't one in the conference room, the second could even be moved there

SomeguyManperson:
didn't nick like add two on the bridge specifically so it would be easy to solo dual confirm

Jarod1200:
RIp. no more can I red alert as the only head of anything.

ThatLing:
Consider adding an exception for the captain on lowpop rounds

AsV9:
Done @ThatLing

alexkar598:
Also further with my idea, I think you could even give an achievement to the player for pulling off auhentificating the red alert by himself but just move them apart so its not possible by walking

AsV9:
I did two things to this PR today:

1. Moved one of the auth devices to the conference room.

2. Redid the system so that it checks the ID you're holding. The auth system now requires you to hold an ID in your hand as you use the system, and it will fail if you use the same ID for both machines.

This essentially allows you to still solo red-alert if you somehow manage to interact with two machines *and* you have two IDs that you successfully juggle around.

I added an achievement for successfully pulling that off.

Hopekz:
Changing my -1 to a neutral after ASV's recent changes.
Still kind of weird to move the auth devices to another room doesn't "fit" IMO.
Now that you need 2 separate keycards just leave the location alone.

PR #7649, opened 2020-01-15
Allow firelocks to be opened without crowbar in more situations

monster860:
:cl: monster860
tweak: Window shutters can now *always* be opened without a crowbar no matter what
tweak: Fire alarms now longer prevent hand-opening of thin firelocks
/:cl:

### DIFF
--- a/code/game/machinery/doors/firedoor.dm
+++ b/code/game/machinery/doors/firedoor.dm
@@ -337,9 +337,7 @@
 
 /obj/machinery/door/firedoor/border_only/allow_hand_open(mob/user)
 	var/area/A = get_area(src)
-	if(A && A.fire)
-		return FALSE
-	if(!is_holding_pressure())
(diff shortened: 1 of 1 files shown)


MegaEmpirical:
Anyway to force open both firelocks yet?

monster860:
yes, stand between them and crowbar both simultaneously

MegaEmpirical:
Does that mean I need to close the one behind me first in order to open both?

monster860:
No, if you're fast enough you can crowbar both in sequence.

MegaEmpirical:
Having fiddled with them more, they make me so fucking angry. Especially with a mech or one that I just opened closing back instantly.

alexkar598:
Strongly against the fire alarm change, your change to fire locks almost completely killed tactical firelocks and this just pins the coffin.

PR #7655, opened 2020-01-17
Minor atmos optimizations

monster860:
- Removes inner_moles system that I added with monstermos. It's supposed to make gases spread slower but it never really worked, and it adds performance costs to all aspects of atmos.
- Adds a transfer_to proc for gas mixtures that acts the same as merge(remove()) except that it doesn't create a temporary gas mixture, making it faster.

Here's a look at how gas mixture transfer is affected:

```
gas mixture transfer
                     Ops/sec
before             - 26,824
remove inner_moles - 33,058
transfer_to proc   - 52,910
```

Since gas mixture transfer accounts for approximately 20% of the performance cost of monstermos, this should make monstermos about 10% faster.

Things that could be done in the future include:
- restructuring gas mixtures to not have associative lists. This would require refactoring any code that does anything with gas mixtures.
- Do a basic gas-mixture-averaging algorithm when zones have only a little pressure difference and are mostly homogenous

### DIFF
--- a/code/__DEFINES/atmospherics.dm
+++ b/code/__DEFINES/atmospherics.dm
@@ -3,7 +3,6 @@
 #define MOLES			1
 #define ARCHIVE			2
 #define GAS_META		3
-#define INNER_MOLES		4
 #define META_GAS_SPECIFIC_HEAT	1
 #define META_GAS_NAME			2
 #define META_GAS_MOLES_VISIBLE	3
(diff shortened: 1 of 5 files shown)


ThatLing:
\>"minor"
\>10%

Hopekz:
> >"minor"
> >10%

*cough*

PR #7660, opened 2020-01-17
Makes space vines only spawn in maintenance.

Takahiru:
This makes kudzu have a chance to actually spread and mutate a bit before being destroyed, so its less of a disappointing event.


overlord12345:
Make vines grate again

PR #7663, opened 2020-01-17
Fixes not being able to say nitrogen unless you injected chemicals into a text box

alexkar598:
Fixes not being able to say nitrogen unless you injected chemicals into a text box


monster860:
nice regex fuckup

Altoids1:
regexes are your friends

PR #7666, opened 2020-01-18
smoking is now bad for you

SomeguyManperson:
% chances subject to change it's the same % as ephedrine currently
:cl:  
rscadd: Nicotine addictions now can cause you to drop items
rscadd: overdosing on nicotine will cause lung damage
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/drug_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/drug_reagents.dm
@@ -60,11 +60,16 @@
 /datum/reagent/drug/nicotine/overdose_process(mob/living/M)
 	M.adjustToxLoss(0.1*REM, 0)
 	M.adjustOxyLoss(1.1*REM, 0)
+	M.adjustOrganLoss(ORGAN_SLOT_LUNGS, 2*REM)
 	..()
 	. = 1
 
 /datum/reagent/drug/nicotine/addiction_act_stage1(mob/living/M)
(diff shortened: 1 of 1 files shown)


MegaEmpirical:
How do you even fix lung damage? New lung?

SomeguyManperson:
Or lobectomy surgery iup think its called, organs will heal on their own if they haven't entirely failed they are just slightly less efifcient untilу the damage is under certain amounts

Hopekz:
So how long of chain smoking to see effects? (looks like about 30-40ish minutes to me but I may be doing mental math wrong)

Moja10:
dropping items is too powerful of an addiction, keep in mind that dropping items was added to ephedrine in order to weaken it's use in combat and here we are adding a 40% drop chance just because someone smoked for very little in game benefit.

SomeguyManperson:
It's not because they smoked a little for antistuns, it's because they either A: smoked long enough to become addicted then didn't keep the addiction up, or B: took the smoker trait which currently has no downside for people without mood enabled since nicotine's addiction is completely harmless aside from the baseline addiction mood debuffs

Hopekz:
Not completely against it but maybe dropping items seems a bit too much of a side-effect?
Not sure what it would be replaced with just never seen smokers literally drop items because they can't get a smoke

deathslayer2481:
Fucking Cigars are lethal.

PR #7685, opened 2020-01-21
Removes the absolution special holoparasite ability for not being a jojo reference

SomeguyManperson:
The ability also has way more practical use than the other two special abilities since it makes the user literally invincible while the parasite is manifested

also it's not a jojo reference 👎 
:cl:  
rscdel: Removes the absolution holoparasite special ability since its not a jojo reference
/:cl:


nesisknesis:
image

Hopekz:
I saw you use this. 
Borderline OP especially with the absolution stands having max range so the user doesn't even need to put themselves in danger (not that they would be in any danger in the first place unless maxcapped or something)

SomeguyManperson:
damage can only be taken through the holoparasite or extensive burning and since the holoparasite has to be stat buffed to even get the ability it's probably got maximum defense or damn close

ghost:
> since its not a jojo reference

how'd you figure it out? also, [20th Century Boy](https://jojo.fandom.com/wiki/20th_Century_Boy).

PR #7696, opened 2020-01-24
chaplain holyparasites don't stop mulligan

SomeguyManperson:
this POSSIBLY delays mulligans or round end when on antagonist death stuff triggers

:cl:  
bugfix: holyparas probably don't prevent round end on antagonist death/mulligan
/:cl:

### DIFF
--- a/code/game/objects/items/holy_weapons.dm
+++ b/code/game/objects/items/holy_weapons.dm
@@ -693,9 +693,9 @@
 		G.key = C.key
 		G.mind.enslave_mind_to_creator(H)
 		G.RegisterSignal(H, COMSIG_MOVABLE_MOVED, /mob/living/simple_animal/hostile/guardian.proc/OnMoved)
-		var/datum/antagonist/guardian/S = new	
-		S.stats = stats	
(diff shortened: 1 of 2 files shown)


MegaEmpirical:
Mind adding a way to pull up holoparasites then? I don't care if they turn a round into extended, I care about bring to identify them.

SomeguyManperson:
this would only remove the chaplain's holyparasite, other holoparasites aren't effected

MegaEmpirical:
Whilst that's good, I was referring to all (holy included).

Hopekz:
I don't like this change because right now there is no easy way to pull up all holo-paras (yes including the holy one).
Until that feature exists not adding them in is just kind of a "fuck you" to admins.

Amelia0010 (changes requested):
Set holoparasite antag datums' `prevent_roundtype_conversion` to `FALSE` instead

Amelia0010 (approved):

alexkar598:
Lyis and hopek points have been fixed so I will disregard their opinions

PR #7697, opened 2020-01-24
Random Role Option (Late Game Joining)

apocist:

This adds a option to the Job Selection page when entering Late/Mid-Round.
A new job category 'Random' with a job title 'Random Job' added to give a random role to those that missed out on that random chance happening at Game Start.

### Reasons:

This is a quality of life recommendation for those that miss that 5 minute joining window between 2 hour long matches. Half time I enter the server and just sit there waiting for 1-120 minutes so I can get a random job when a new game starts.

Desired TODO for future: An option to use the Job Preferences as well

------

Added 'GetRandomJob' functions and moved the functionality from 'GiveRandomJob' into them for reuse elsewhere.

#### Changelog

:cl:  
rscadd: Adds 'Random Job' to Late Join Job Selection 
rscadd: Added 'GetRandomJob' proc to Job def
tweak: Moved around Job Functions to avoid duplication (all existing intact) 
/:cl:

### DIFF
--- a/code/controllers/subsystem/job.dm
+++ b/code/controllers/subsystem/job.dm
@@ -78,8 +78,8 @@ SUBSYSTEM_DEF(job)
 		SetupOccupations()
 	return type_occupations[jobtype]
 
-/datum/controller/subsystem/job/proc/AssignRole(mob/dead/new_player/player, rank, latejoin = FALSE)
-	JobDebug("Running AR, Player: [player], Rank: [rank], LJ: [latejoin]")
+// Checks for if player is Assignable to Role
+/datum/controller/subsystem/job/proc/CanAssignRole(mob/dead/new_player/player, rank)
 	if(player && player.mind && rank)
 		var/datum/job/job = GetJob(rank)
 		if(!job)
@@ -90,6 +90,14 @@ SUBSYSTEM_DEF(job)
 			return FALSE
 		if(job.required_playtime_remaining(player.client))
 			return FALSE
+		return TRUE
+	return FALSE
+
+// Attempts to Assign player to Role
+/datum/controller/subsystem/job/proc/AssignRole(mob/dead/new_player/player, rank, latejoin = FALSE)
+	JobDebug("Running AR, Player: [player], Rank: [rank], LJ: [latejoin]")
+	if(CanAssignRole(player, rank)
+		var/datum/job/job = GetJob(rank)
 		var/position_limit = job.total_positions
 		if(!latejoin)
 			position_limit = job.spawn_positions
@@ -131,8 +139,8 @@ SUBSYSTEM_DEF(job)
 			candidates += player
 	return candidates
 
-/datum/controller/subsystem/job/proc/GiveRandomJob(mob/dead/new_player/player)
-	JobDebug("GRJ Giving random job, Player: [player]")
+// Fetch a random job that a specific player can use
+/datum/controller/subsystem/job/proc/GetRandomJob(mob/dead/new_player/player)
 	. = FALSE
 	for(var/datum/job/job in shuffle(occupations))
 		if(!job)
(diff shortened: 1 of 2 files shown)


ynot01:
why?

apocist:
> why?

Is that a meme? 
`to give a random role to those that missed out on that random chance happening at Game Start.`

It's a quality of life recommendation for those that miss that 5 minute joining window between 2 hour long matches. Half time I enter the server and just sit there waiting for 1-120 minutes so I can get a random job when a new game starts.

code/controllers/subsystem/job.dm
```diff
@@ -164,9 +172,18 @@ SUBSYSTEM_DEF(job)
 			continue
 
 		if((job.current_positions < job.spawn_positions) || job.spawn_positions == -1)
-			JobDebug("GRJ Random job given, Player: [player], Job: [job]")
-			if(AssignRole(player, job.title))
-				return TRUE
+			JobDebug("GRJ Random job can give, Player: [player], Job: [job]")
+			if(CanAssignRole(player, job.title))
```
alexkar598: Duplicate check, the stuff CanAssignRole checks for is already checked for above

code/modules/mob/dead/new_player/new_player.dm
```diff
@@ -161,6 +162,12 @@
 				to_chat(usr, "<span class='warning'>Server is full.</span>")
 				return
 
+		// Check if random role is requested
+		if(href_list["SelectedJob"] == "Random")
+			var/datum/job/job = SSjob.GetRandomJob(src)
+			if(job != FALSE)
```
alexkar598: You need to abort if the job is false as it's going to try spawning the person with the role of "Random"

Add an if(!job)
                     Return

Preferably with some error message to the user

code/controllers/subsystem/job.dm
```diff
@@ -78,8 +78,8 @@ SUBSYSTEM_DEF(job)
 		SetupOccupations()
 	return type_occupations[jobtype]
 
-/datum/controller/subsystem/job/proc/AssignRole(mob/dead/new_player/player, rank, latejoin = FALSE)
-	JobDebug("Running AR, Player: [player], Rank: [rank], LJ: [latejoin]")
+// Checks for if player is Assignable to Role
+/datum/controller/subsystem/job/proc/CanAssignRole(mob/dead/new_player/player, rank)
```
alexkar598: There isn't much reason to move the checks out of the proc into another one

alexkar598 (changes requested):

code/controllers/subsystem/job.dm
apocist: For the sake of re-usability and not duplicating the logic checks originally found in `/datum/controller/subsystem/job/proc/AssignRole`, those logic checks were moved to and recreated with `/datum/controller/subsystem/job/proc/CanAssignRole`

The same goes to `/datum/controller/subsystem/job/proc/GiveRandomJob`, where the grabbing of a random job has been moved and is now first done with the new proc `/datum/controller/subsystem/job/proc/GetRandomJob` where it finds a Job that's actually select-able by using the reusable proc `CanAssignRole()` to ensure that the a Role can be used by that player...otherwise find the next random Role (until it runs out and returns null)

apocist: Please take a look at https://github.com/yogstation13/Yogstation/pull/7697/files/2ad5ab94e1fb9f7f48bc22e6e0f9c60e514d205d#diff-80ea298b5ebb3d34358d35f2393c6cec for my reasons of trying not to duplicate code

code/modules/mob/dead/new_player/new_player.dm
apocist: True. I figured it wasn't that needed as the AssignRole() will also abort the job assignment with an error. I can see about a new message to returning there.

code/controllers/subsystem/job.dm
apocist: CanAssignRole() is required here as it's looping through each of the possible jobs. If a job it gets isn't assignable for other reasons, it'll still look for the next job in the list.

Without this check, it could commonly send back a Job that couldn't be assigned and the button for joining as a 'Random Job' could possibly fail a couple of times for no reason (e.g. faulty)

code/modules/mob/dead/new_player/new_player.dm
apocist: Fixed

code/modules/mob/dead/new_player/new_player.dm
```diff
@@ -161,6 +162,14 @@
 				to_chat(usr, "<span class='warning'>Server is full.</span>")
 				return
 
+		// Check if random role is requested
+		if(href_list["SelectedJob"] == "Random")
+			var/datum/job/job = SSjob.GetRandomJob(src)
+			if(!job)
+				to_chat(usr, "<span class='warning'>There is no randomly assignable Job at this time. Please manually choose one of the other possible options.</span>")
```
alexkar598, line 169: Job should be jobs or job i cant grammar but it shouldn't be capitalized either case

alexkar598 (changes requested):

PR #7712, opened 2020-01-25
Cloning now requires biomass

Szyszkrzyneczka:
### You now must use a resource known as "Biomatter" to fuel cloning
You craft biomatter with 10u blood, 5 u nutrients
cloning costs 4 biomatter blocks.
Gene mans get three clonings free, and 1/2 of the third cloning
The harder you upgrade the pod the more biomatter you will need, capping at 150 biomatter (6 blocks of it) with tier 4 upgrades.
This gives antags more options to make sure their target won't get cloned besides bombing the whole of medbay
This also makes geneticists have more responsibilites beyond powergaming and fixing people in eventual rad storm
You can also get biomatter by grinding bodies in a new machine
#### Changelog

:cl:  
rscadd: Added Biomatter a food you can eat and it is used to fuel cloning  / Adds body grinder to cloning that can be used to grind bodies into biomatter.
tweak: Now you need to put some effort to clone people
imageadd: Added my shitty biomatter icons (spriters feel free to point out that that biomatter is a coder sprite
/:cl:


overlord12345:
the idea is meh but costing more as time gose on is just gonna make people deconstruct and then reconstruct the cloner

Szyszkrzyneczka:
> the idea is meh but costing more as time gose on is just gonna make people deconstruct and then reconstruct the cloner

A simple global var should make this problem commit bye bye

overlord12345:
good but i still dislike the idea of it riseing

DerelictDrone:
> 
> 
> > the idea is meh but costing more as time gose on is just gonna make people deconstruct and then reconstruct the cloner
> 
> A simple global var should make this problem commit bye bye

If we use a global variable an antagonist could raise the cost of cloning to astronomical levels by simply setting up a secret cloner in maint and speedcloning to avoid medical comms announcements about completed empty clones. Portable chem/stolen chem dispenser for sugar, a pandemic can spit out free blood samples if the clones can't provide enough blood for you. Use a gibber on the clones for meat after harvesting their blood, chop them into cutlets and cook them before grinding them up with a mixer for efficiency. When the cost of cloning becomes too much for this method they could supplement it with meatwheat/some other plant or by taking from the chef's table.

I understand this would be particularly unlikely for anyone to do, but it is a possibility if you tie it to a global var and raise the cost with each clone.

An alternative to raising the global cost with each clone would be to raise the cost as the round goes on, say for example every 10 to 20 minutes the cost rises by 1 extra biomatter, so at shift end it would cost you about +14 to +7 biomatter to make some last minute clones, making it easier to clean up early on in the round, but harder to stop sudden catastrophic levels of death.

Another possible alternative would be to raise the cost of clones relative to the amount of upgrades the cloner has, so if you want to get cloned _fast_ or get genetic powers applied, or simply just get tossed out with no clone damage it would be pricier, for example each part rated one level higher than 1 would increase cloning cost by 0.5 a biomatter, making full tier 2 parts cost about 1.5 biomatter, where full tier 4 parts would increase cost about 4.5 per. (maybe include some rounding for these numbers)

Either way, with clones increasing cost _globally_ this could lead to some abuse by the antagonists or other unknown third parties.

ghost:
you have revival and coronary bypass surgery, so cloning should be irrelevant anyway given you can just defib the guy, hence if someone goes to the effort of  making cloning cost astronomical amounts of biomatter it's a wasted effort

besides that, there's no reason to have scaling biomatter requirement or make it use a custom chem rather than synthflesh or synthmeat. if you want the custom chem, whatever, but the scaling requirement is unnecessary.

boodaliboo:
If you remove the scaling biomatter cost im all for this

code/modules/reagents/chemistry/recipes/others.dm
```diff
@@ -111,6 +111,18 @@
 		new /obj/item/reagent_containers/food/snacks/meat/slab/meatproduct(location)
 	return
 
+/datum/chemical_reaction/biomatter_creation
+	name = "Biomatter Creator"
+	id = "biomatter_creation"
+	required_reagents = list(/datum/reagent/blood = 15, /datum/reagent/consumable/nutriment = 30, /datum/reagent/consumable/sugar = 15)
```
ynot01:
```suggestion
	required_reagents = list(/datum/reagent/blood = 15, /datum/reagent/consumable/nutriment = 15, /datum/reagent/consumable/sugar = 15)
```
nutriment is by far the hardest thing to get in this recipe i think getting 30 is overkill

overlord12345:
Derelictdrone's idea is good i beleve it would be better

overlord12345:
Also how would empty clones work with this?

code/modules/reagents/chemistry/recipes/others.dm
Szyszkrzyneczka: yer right ynot, i will change it so it's 1:1:1 ratio

Szyszkrzyneczka:
> Also how would empty clones work with this?

Like regular clones

overlord12345:
So if someone spammed alot of empty clones the cost would just keep riseing?

Szyszkrzyneczka:
rn yes, but i will change it to derelict's idea

> So if someone spammed alot of empty clones the cost would just keep riseing?

overlord12345:
Good cause his idea cant be exploited as easily and makes some more sense

Hopekz:
I too think the whole scaling requirement is unnecessary.
It seems kind of out of nowhere because it doesn't even make sense in-game.
Why would this be getting more expensive? The world may never know.

Szyszkrzyneczka:
As of feedback i decided to snap the scaling

Amelia0010:
Imagine unironically taking input from Adam Al-Tableway

boodaliboo:
ill remove my downvote once i see that the scaling is removed i also think biomatter should be slightly less expensive

DerelictDrone:
> 
> 
> ill remove my downvote once i see that the scaling is removed i also think biomatter should be slightly less expensive

Biomatter got changed to be a 15-15-15 ratio(sugar-nutri-blood) in neczka's second commit to this PR, but he hasn't edited the original comment of the PR yet to reflect that in particular.

He has done this now, thankfully.

alexkar598:
The craft is stupid to get, you need departmental cooperation for cloning, this is another one of those nickvr xeno rework prs which are removals in disguise by making the feature a pain in the ass to use so noone uses it

alexkar598:
If miners cant be passed to retrieve slimes despite there being miners almost every round and them actually trying to do their job but there are rarely any chefs and when they exists they rarely make any food

boodaliboo:
That’s why I said to not use nutriment for the recipe

Ratvar:
Geneticist will not do this, same way they do not clone people at all, nor will they be punished, same way they are not punished for ignoring cloning. Then you also need sugar to make it, means it is work of *chemist* and not geneticist. What stops chemist from massproducing biomass roundstart, mitigating whole reason it is needed? Why is it even something chemists do if you want medics/genetics to do it?

This does not really give antags more options their target won't get cloned: it gives antags more options to make sure literally nobody is cloned.  If antags can disable cloning by taking biomass, then taking biomass is as OP as bombing cloning. If they cannot, then it is useless for said antags do to. Eh.
Then antags can bomb cloning *and* take biomass, just so you cannot make new cloning even with circuits.
 
Maybe biomass could be required to make limbs, so robotics can help/so medics can surgery people up if biomass is lacking? It still will be a worse option, but it will open more ways to cooperate to mitigate it.
Goon has biomass-light: you can reprocess corpses into biomass, and it takes time. This gives **doctors**, not chemists, something fun to do, it makes corpses usable and it gives antags more options to get rid of people, and it is a an actual scarce resource compared to nutrient. Why not do something like that instead?

ShadowDeath6:
don't see why this is needed, antags already have a ton of ways to dispose of their targets and this isn't going to stop cloning bombing in any capacity

all this does do well is make it so that sci has to put a tiny bit more effort into making their massive scifort with a bunch of medical and auto cloning and all that, but it's not like needing chems will stop them. 

given that this is only feasible to produce in large quantities in chemistry, I think what's more likely to happen is that genetics is going to sit and give themselves power and wait for chem to make everything, similarly to virology doing the same thing with unstable mutagen

Ratvar:
To remember: last time shadowdeath6 played, science still could print cloner and we had autocloning

alexkar598:
Stop with the weird titles, this PR is trying to get an important thing passed. The title makes the PR less clear on what it is before people click

code/game/machinery/computer/cloning.dm
```diff
@@ -179,6 +179,23 @@
 	var/dat = ""
 	dat += "<a href='byond://?src=[REF(src)];refresh=1'>Refresh</a>"
 
+	if(pods)
+		for(var/P in pods)
+			var/obj/machinery/clonepod/pod = P
+			if(pod.efficiency >= 3)
```
DerelictDrone: We need a case that detects if it's 1 or 2, since if efficiency were to be set lower than 3 after previously being at/above 3, the price won't change back to 100, which is how it starts.

deathslayer2481:
Don’t fix what isn’t broken.

code/game/machinery/computer/cloning.dm
Szyszkrzyneczka: Thanks for that!

Darkstick:
I'm all in for the idea just so I can finally have an universally accepted excuse to beat up doctors that wont stop cloning instead of reviving

Sir-Lagsalot:
One step closer to being more like Goon. I approve.

SomeguyManperson (dismissed):
Undocumented changes/atomize the whole hulk/tk thing

Ktlwjec1:
Stale so closing, feel free to reopen if you have time to fix the conflicts or someone else can remake it.

PR #7730, opened 2020-01-29
Adds the bo staff to the uplink and removes its stupid rng stun

SomeguyManperson:

Adds the bo staff to the uplink for 8 tc. Similarly to the now removed katana it is large and difficult to conceal, however the bo staff is capable of dealing moderate stamina damage and will knock someone out if they are stamcrit by it, giving it a position as a nonlethal melee weapon. Currently it is a normal sized iem that can be stored on the back, with 24 wielded damage, 28-33 stamina damage when on disarm intent instead, 20 throwforce and 50% blockchance.
~~Might decrease the size and/or increase the stamina damage to make it more viable~~

#### Changelog

:cl:  
rscadd: Adds the bo staff to the uplink for 8 tc. The bo staff is slightly weaker than the standard esword when attacking lethally but can be wielded on disarm intent to cause large amounts of stamina damage and incapacitate its target
rscdel: bo staff no longer has a chance to randomly 8 second stun, and the knockout duration is 30 seconds instead of a minute
/:cl:

### DIFF
--- a/code/datums/martial/sleeping_carp.dm
+++ b/code/datums/martial/sleeping_carp.dm
@@ -217,16 +217,12 @@
 							   "<span class='userdanger'>[pick(fluffmessages)]</span>")
 		playsound(get_turf(user), 'sound/effects/woodhit.ogg', 75, 1, -1)
 		H.adjustStaminaLoss(rand(13,20))
-		if(prob(10))
-			H.visible_message("<span class='warning'>[H] collapses!</span>", \
-								   "<span class='userdanger'>Your legs give out!</span>")
-			H.Paralyze(80)
 		if(H.staminaloss && !H.IsSleeping())
 			var/total_health = (H.health - H.staminaloss)
 			if(total_health <= HEALTH_THRESHOLD_CRIT && !H.stat)
 				H.visible_message("<span class='warning'>[user] delivers a heavy hit to [H]'s head, knocking [H.p_them()] out cold!</span>", \
 									   "<span class='userdanger'>[user] knocks you unconscious!</span>")
-				H.SetSleeping(600)
+				H.SetSleeping(300)
 				H.adjustOrganLoss(ORGAN_SLOT_BRAIN, 15, 150)
 	else
 		return ..()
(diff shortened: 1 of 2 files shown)


deathslayer2481:
WEEB

ynot01:
If it deals less stamina damage than it deals actual damage than why would you use the stamina damage alternative? Also how is the knockout function triggered?

SomeguyManperson:
Knockout is triggered if the person's current health minus stamina damage (including the damage from the attack) puts them below 0 heallth. The stamina damage *probably* ignores armor currently

boodaliboo:
Play die doedoddoedosxoe

boodaliboo:
kung fu fighting

alexkar598:
This item has:
50% Block chance
24 real damage
30 stamina damage

First hit:
You have 30 stamina damage
Health check: you are 70 damage away from being sleeped
Slowdown check: You have 30 stamina damage, it is not over 40 so no slowdown

Second hit:
You have 60 stamina damage
Health check: You are at 40 damage from been sleeped
Slowdown check: you have 60 stamina damage, you are over 40 damage, you have a 2.4 movespeed modifier

Third hit:
You have 90 stamina damage
Health check: You are now at 10 damage from being sleeped
Slowdown check: you have 90 stamina damage and a 3.6 movespeed modifier

Foruth hit:
You have 120 stamina damage, this is the point where you would fall over due to stam crit
Health check: you are way past the sleepy point
Slowdown check: you are on the floor

Results:
This item does as much stam damage as a disabler
This item is 6 damage away from being as powerful as an esword
This item makes little noise
This item can be easily concealed in a backpack
This item will put you to sleep where you cant scream on comms in 4 hits
This item will put you in stam crit in 4 hits
This item will kill you in 9 hits in harm mode

ghost:
@alexkar598 

wow it's almost like every other dangerous item in the game that drops people in one to three hits and isn't very obvious

this isn't a balance concern

what it is is fucking stupid because instead of an actual original traitor idea HAHA TIME TO ADD A NEW ESWORD VARIANT XXXXD

alexkar598:
I made the post originally assuming that it was doing both the brute and stam damage, I correct that in the post but my only thing that makes me iffy on this is the lack of sound

SomeguyManperson:
bonk is now more unique soundwise

alexkar598:
I'm fine with this atm, we can try it for a while and if it's super shit il speed merge a revert or something

PR #7734, opened 2020-01-30
Maphash - A Better Mapmerge

monster860:
### TL;DR

- less merge conflicts
- no "Prepare Maps" necessary

### The Problem

The problem with old mapmerge is simple - it doesn't work. Sure, it reduces the diffs, but it is extremely ineffective at preventing merge conflicts. If two people try to change the same map, even if they're completely unrelated, if any new map keys are added, you are all but guaranteed a merge conflict. This is because mapmerge uses the first available key when a new key is needed, which means both map changes will try to use the same key, causing a merge conflict. To make the problem worse, often lazy mappers will often forgo "Prepare Maps" sometimes, resulting in sometimes even larger diffs. Finally, if all that wasn't bad enough, dealing with python's absolutely horrid package manager will make you go insane.

### The Solution

The solution is a map converter. The way maphash works is very simple - it sorts all the map tile instances by their contents, and then applies SHA-1 to the contents with 0, 1, 2... etc appended to it until it finds one that isn't taken. The map file essentially becomes a hashtable. That means that two map PRs are unlikely to try to use the same key - By my estimation, two PRs which add 50 new map keys each would result in about a 7% chance of merge conflict - much better.

Another massive improvement here is *not using python*. Now, it uses node.js, and it has no dependencies - no need to finagle with python's horrendous package manager, and in this case you don't even need to deal with node's non-horrendous package manager. It can also do things faster because  of the asynchronous nature of node's file I/O

### Changes to map format

The map format no longer uses TGM. While the multi-line key format is slightly more human readable, the human readability goes out the window when diffing simply because of how the diff algorithm outputs it. I have removed the multi-line keys. However, I have retained the one-tile-per-line feature of the map part. Note the header `// This map has been converted using maphash`. This is used solely so that custom map editors can identify this format and handle it accordingly. The tool checks if a file needs conversion by converting the file and comparing it to the original - not efficient, but effectively protects against people trying to use nodepad to edit the file.

### Possible improvements

- Currently, BYOND only allows map keys up to 65534, or the key `ymo`. Using any keys above that *may* work fine because it wraps around to 0, or it might cause strange errors. This means that this hashing method is not as collision-resistant as it could be - if Lummox could be pestered into increasing this limit, this format could become more effective.
- I've never touched github actions, but perhaps one can be used to automatically run maphash on map PRs.
- I have already added checks to travis.
- Due to the fact that this tool loads all the map files at once, it is not very memory efficient. However, it will run for all of 5 seconds before exiting so it really doesn't matter.
- For the cases where merge conflicts do happen, a git merge driver could be written


alexkar598:
> Prevents merge conflicts

> This branch has conflitcts that must be resolved. Merging is blocked

boodaliboo:
Thank god for this

monster860:
Alright I removed any escaped quotes inside strings in map files. Keep in mind that the SS13 map loader won't load those properly either, so removing those quotes is necessary either way.

monster860:
```
RED ALERT! The following map files failed to load: _maps/map_files/mining/Lavaland.dmm. Yell at your server host!
```

putting DNM until I figure out why

monster860:
hmmm loading lavaland seems to work fine locally. I'll restart the travis build and see if this is just one of travis's random one-off errors.

monster860:
I have determined that this issue is not caused by this, and was present before, and doesn't affect the server.

Szyszkrzyneczka:
God bless you monster

Firewolf34:
*b l e s s i d d* monster out here with the godlike improvements

ThatLing:
>dealing with python's absolutely horrid package manager will make you go insane.
>Now, it uses node.js

This is arguably worse

monster860:
NPM is a great package manager. However that doesn't matter because this thing doesn't use NPM.

ThatLing:
You could probably speed the script up by using something like [xxHash](https://cyan4973.github.io/xxHash/) instead as there is no need for a cryptographically secure hash

ThatLing:
Additionally, how are the diffs using this system? If I change one tile, will it just result in a 1 line diff?

monster860:
The performance gains from switching to a non-cryptographic hash is neglibible. Besides, node.js has SHA-1 in it out of the box.

monster860:
The diffs depend on what you do. See [this commit](https://github.com/yogstation13/Yogstation/pull/7734/commits/dd54bbbbc6459533471944dda035f15fdacd1ec7) for an example of some diffs.

ThatLing:
>The performance gains from switching to a non-cryptographic hash is neglibible

Have you tried it? xxHash is considerably faster than SHA1.
>Besides, node.js has SHA-1 in it out of the box.

There are js ports of xxHash that you can just statically include into the codebase

monster860:
`Caution! Below 300 MB/sec xxhash can be a bit slower than MD5/SHA-1, but beyond it’s much faster, because md5 can’t get faster and xxhash goes up to a few GB/sec.` :thonk:

monster860:
The performance benefits of switching to xxHash would be negligible.

image


If you look at this screenshot (keep in mind that in this screenshot it's doing *all* the maps, don't hecking panic) hashing takes up 142ms, any potential performance gains would be had from simply optimizing toString for map objects.

monster860:
Alright I have made the following changes:

- It now keeps SHA1 hashes in a .json file of converted maps, and if the map file matches the hash, it is already converted and hasn't been touched by people with notepad, and doesnt need to be reconverted.
- the precommit hook works
- the .dmi merge driver has been restored, along with a text file explaining where actual mapmerge went.

monster860:
I have changed it so that it sorts the thing in such a way that tile instances with more occurences get higher priority in resolving hash collisions. This is to prevent situations where a hash collision causes something like space to be rekeyed, which would be very bad.

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
+const DMM = require('./lib/dmm.js');
+const fs = require('fs').promises;
+const path = require('path');
+const readline = require("readline");
+const crypto = require('crypto');
+
+if(!fs) {
+	console.error("Your node.js is out of date. You must have node.js 10 or newer.");
```
alexkar598, line 8:
```suggestion
	console.error("Your node.js is out of date. You must have node.js 10 or newer. You can download a newer version for your system at https://nodejs.org/en/download/ by selecting the installer for your operating system.");
```

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+	process.exit(1);
+}
+
+//let text = fs.readFileSync('../../_maps/map_files/YogStation/YogStation.dmm', 'utf8');
+//fs.writeFileSync('out.dmm', new DMM(text).toString());
```
alexkar598, line 13: Commented code is bad

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+		let ent_path = path.join(dirname, ent.name);
+		if(ent.isDirectory()) {
+			promises.push(do_dir(ent_path));
+		} else if(ent.name.endsWith(".dmm")) {
+			promises.push(fs.readFile(ent_path, 'utf8').then((file_text) => {
+				process.stdout.write('\r' + slashes[(++slash_index) % 4]);
+				let hash;
+				if(converted_file_cache[path.resolve(ent_path)]) {
+					hash = crypto.createHash('sha1').update(file_text).digest('base64');
+					if(converted_file_cache[path.resolve(ent_path)] == hash) return;
```
alexkar598, line 39: If a single file in a directory is already converted, you return?

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+				}
+				try {
+					let dmm = new DMM(file_text);
```
alexkar598, line 42: Note to self, check if constructor is large

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+}
+(async () => {
+	let initial_path = '../..';
+	if(process.argv.includes("--root-dir")) {
+		initial_path = process.argv[process.argv.indexOf('--root-dir') + 1];
+	}
+
+	try {
+		converted_file_cache = JSON.parse(await fs.readFile(path.join(__dirname, "converted_file_cache.json"), 'utf8'));
+	} catch(e) {}
```
alexkar598, line 68: Don't silently catch errors

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+			console.error("Some files don't match - " + files.join(", "));
+			process.exit(1);
+		} else {
+			console.log("All files match.");
+			process.exit(0);
+		}
+	}
+	if(!files.length) {
+		console.log("All files are already converted!");
+		await fs.writeFile(path.join(__dirname, "converted_file_cache.json"), JSON.stringify(converted_file_cache));
```
alexkar598, line 86:
```suggestion
		fs.writeFileSync(path.join(__dirname, "converted_file_cache.json"), JSON.stringify(converted_file_cache));
```

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+		process.exit(0);
+	}
+	files.sort();
+	for(let i = 0; i < files.length; i++) {
```
alexkar598, line 90: Could you use files.forEach() instead of a standart c loop?

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+	let dir = await fs.readdir(dirname, {withFileTypes: true});
+	let promises = [];
+	for(let ent of dir) {
+		if(ent.name == ".git" || ent.name == "node_modules") continue; // ignore folders with lots of irrelevant files.
+		let ent_path = path.join(dirname, ent.name);
+		if(ent.isDirectory()) {
+			promises.push(do_dir(ent_path));
+		} else if(ent.name.endsWith(".dmm")) {
+			promises.push(fs.readFile(ent_path, 'utf8').then((file_text) => {
+				process.stdout.write('\r' + slashes[(++slash_index) % 4]);
```
alexkar598, line 35: Whats up with writing to stdout? Its almost the same as console.log but once is easier for newer coders

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+		console.log(i + ": " + files[i]);
+	}
+	function save(flagged_maps) {
+		let promises = [];
+		for(let map of flagged_maps) {
+			let text = file_outputs.get(map);
+			if(map) {
+				promises.push(fs.copyFile(map, map + ".before").then(async () => {
+					console.log("Created backup for " + map);
+					await fs.writeFile(map, text);
```
alexkar598, line 100:
```suggestion
					fs.writeFileSync(map, text);
```

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+	files.sort();
+	for(let i = 0; i < files.length; i++) {
+		console.log(i + ": " + files[i]);
+	}
+	function save(flagged_maps) {
+		let promises = [];
+		for(let map of flagged_maps) {
+			let text = file_outputs.get(map);
+			if(map) {
+				promises.push(fs.copyFile(map, map + ".before").then(async () => {
```
alexkar598, line 98: Use copyFileSync instead of nesting promises as the then won't be tracked by Promise.All and stuff might fuck up

alexkar598, line 98: Wait, fs.copyFile doesn't even return a promise!

alexkar598, line 98: Do not use async within a promise constructor
```suggestion
				promises.push(fs.copyFile(map, map + ".before").then(() => {
```

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+					console.log("Created backup for " + map);
+					await fs.writeFile(map, text);
+					console.log("Written " + map);
+					converted_file_cache[path.resolve(map)] = crypto.createHash('sha1').update(text).digest('base64');
+				}).catch(console.error));
+			}
+		}
+		Promise.all(promises).then(() => {
+			return fs.writeFile(path.join(__dirname, "converted_file_cache.json"), JSON.stringify(converted_file_cache));
```
alexkar598, line 107: the return value is never used here, also use writeFileSync so that there are no race conditions leading to reconversion if you use save twice too quickly(its not atm, but it might happen)

tools/maphash/index.js
```diff
@@ -0,0 +1,134 @@
...
+					}
+				} else {
+					flagged_maps.add(files[+range[0]]);
+				}
+			}
+			rl.close();
+			save(flagged_maps);
+		});
+	}
+})();
```
alexkar598, line 134: Also use a file handle instead of a file path for your cache file

tools/maphash/lib/dmm.js
```diff
@@ -0,0 +1,471 @@
...
+	}
+	return num;
+}
+
+function compare_keys(a, b) {
+	return key_to_number(a) - key_to_number(b);
+}
+DMM.Instance = Instance;
+
+module.exports = DMM;
```
alexkar598, line 471: I have absolutly not idea how this file works and i lack the motivation to figure what it does

alexkar598 (changes requested):

tools/maphash/index.js
monster860: fs.promises doesn't have writeFileSYnc

tools/maphash/index.js
monster860: I never use forEach

tools/maphash/index.js
monster860: `fs = require('fs').promises`

tools/maphash/index.js
monster860: theres no point, I'm reading the whole file at once, then writing the whole file at once.

tools/maphash/index.js
monster860: because this doesnt put a newline at the end of the line.

tools/maphash/index.js
monster860: Why not, the callback you pass to `.then()`, if you return a promise, then it returns a promise that resolves when that promise resolves, so it works fine.

tools/maphash/index.js
monster860: `fs.promises` doesnt have writeFileSync

tools/maphash/index.js
monster860: oops

monster860:
In many cases, wher you suggest using `sync`, that would hurt performance.

tools/maphash/index.js
alexkar598: Fair enough

tools/maphash/index.js
alexkar598: Nvm, its within the then not the promise constructor

tools/maphash/index.js
alexkar598: Since it gives a promise, forget what I said, but new thing: remove that arrow function, keep the call directly

PR #7738, opened 2020-01-31
moves the magic mirror/some wizards a bit, removes some loot from the academy leader or whatever room and gives wizards boss level sentience

SomeguyManperson:
:cl:  
tweak: Magic mirror is now inside the bathroom instead of the room-that's-supposed-to-spawn-a-ghost-role-but-doesn't
tweak: mirror in the room-that's-supposed-to-spawn-a-ghost-role-but-doesn't is a normal mirror
tweak: moved colburn out of a room that's completely useless and never used to the hallway near the bathrooms
tweak: moves bowl into the bathroom from the room that's completely useless and never used
tweak: removes stupid shit from the room-that's-supposed-to-spawn-a-ghost-role-but-doesn't since it's a fireball staff and antimagic suit which is dumb to get without even fighting a single wizard
tweak: wizards now have boss level sentience and therefore can't be yeeted as easily
/:cl:


deathslayer2481:
This doesn’t fix the fact they aint atmosproof

Jarod1200:
Making them atmos proof was a different pr

alexkar598 (changes requested):
Undocumented mapping changes

SomeguyManperson:
Fixed

PR #7753, opened 2020-02-04
Increase minimum pop for blob

JamieD1:
On low pop rounds its a landslide and unfair to the crew.

:cl:  JamieD12
tweak: Changed blob from 20 to 25 players
/:cl:

### DIFF
--- a/code/modules/events/blob.dm
+++ b/code/modules/events/blob.dm
@@ -4,7 +4,7 @@
 	weight = 10
 	max_occurrences = 1
 
-	min_players = 30
+	min_players = 25
 
 	gamemode_blacklist = list("blob") //Just in case a blob survives that long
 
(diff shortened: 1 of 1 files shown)


boodaliboo:
Can you change the Pr name to reflect what this actually does

code/modules/events/blob.dm
```diff
@@ -4,7 +4,7 @@
 	weight = 10
 	max_occurrences = 1
 
-	min_players = 20
+	min_players = 30
```
Readystorm546: I would  not do a drastic change to 30. How about 25 instead

JamieD1: We could try 25 and see how it goes

boodaliboo: please reflect that you changed it to 25 in the descrption

boodaliboo: please reflect that you changed it to 25 in the descrption

PR #7754, opened 2020-02-04
add protect/maroon as obsessed endgame objective

SomeguyManperson:
Protect/maroon is more likely to result in an interesting scenario than a straight murder objective, which has an incredibly high chance to result in someone chasing someone else around for 5 minutes then pulling the old kill and cremate.

#### Changelog

:cl:  
rscadd: Obsessed now has a chance to have a protect/maroon objective instead of a kill objective
/:cl:


alexkar598 (changes requested):
Didn't call update explanation text for the protecc objective

SomeguyManperson:
fixed

PR #7756, opened 2020-02-04
fix to make badblink not jettison people straight out of cloners making them immune to air, heart, and liver damage but also completely incapable of communication

SomeguyManperson:
:cl:  
bugfix: you will no longer get yeeted out of a cloning pod early by the spatial instability mutation 
/:cl:

### DIFF
--- a/code/datums/mutations/body.dm
+++ b/code/datums/mutations/body.dm
@@ -285,7 +285,7 @@
 	var/warpchance = 0
 
 /datum/mutation/human/badblink/on_life()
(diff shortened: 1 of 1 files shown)


alexkar598 (changes requested):
Make it not work in containers Instead, i can think of a few cases like space pods and mechs where you dont want people exiting

SomeguyManperson:
travis fail isn't my fault :crab:

alexkar598 (changes requested):
Do the prob check before the isturf check as prob is less expensive

PR #7764, opened 2020-02-06
Nerf wizard academy 2

JamieD1:
Removing stand from wizard academy for power gamers

#### Changelog

:cl:  

rscdel: Removed stand from wizard academy

/:cl:


boodaliboo:
instead of removing it why not try buffing its defense or making it a stand arrow so theres a chance you will be gibbed

Hopekz:
I like bood's idea of except make it a one time use stand arrow

overlord12345:
Stand arrows are multi-use just make it work like decks on lavaland that give you a random stand

boodaliboo:
It already gives a random one

boodaliboo (changes requested):
add this between 406/407 or where you deleted the deck
/obj/item/stand_arrow

_maps/RandomRuins/SpaceRuins/Academy.dmm
```diff
@@ -404,7 +404,6 @@
 /area/awaymission/academy/headmaster)
 "bw" = (
 /obj/structure/table/wood,
-/obj/item/guardiancreator,
```
boodaliboo, line 407: /obj/item/stand_arrow add this there

Hopekz:
No bood's fix is lazy.
Make a child for the arrow then edit the var for charges to make it a 1 time use.
Then add that child to the map

ynot01:
just remove academy

Alexandria404:
step in the right direction

JamieD1:
What about 66% chance of gibbing one time use?

Hopekz:
> What about 66% chance of gibbing one time use?

I think the current arrow is about a 50% from what I remember. You should probably be fine with leaving the % chance unless you feel that it needs to be changed.

Seeing that this PR is meant to remove it I'd say going in the 60's is more in line with your original intent. 
Make sure to differentiate this arrow somehow from the normal one (different name/flavortext)

JamieD1:
> I think the current arrow is about a 50% from what I remember. You should probably be fine with leaving the % chance unless you feel that it needs to be changed.
> 
> Seeing that this PR is meant to remove it I'd say going in the 60's is more in line with your original intent.
> Make sure to differentiate this arrow somehow from the normal one (different name/flavortext)

To be honest I want it gone to prevent people going for it, plus gibbing them would ruin the round for them and if they powergame xeno to get there maybe gibbing would stop them from trying it everytime. 

Im open to suggestions.

SomeguyManperson:
The arrow has variables for its kill chance, use amount, and if it can be used to requiem a holoparasite, so changing those shouldn't be difficult

boodaliboo:
IF you die there anyways your not getting ressed I see no problem with it Gibong

Hopekz:
> The arrow has variables for its kill chance, use amount, and if it can be used to requiem a holoparasite, so changing those shouldn't be difficult

Ye what Theos said because those variables already exist creating a child with those things changed should take you a few seconds + be the less lazy fix

Ratvar:
Chance to die if used is a lazy fix because there is a chance you *dont* die and then are free to abuse something OP. Just like with die of fate, there was a chance you would not get a wizard, uplink or a slave. Did not matter that it was a chance if you could get a wizard, uplink or a slave at all, and so it is removed.
If it dissuades people from using it, then it just becomes rarer, not better, and reason it has to be dissuaded at all is still there.
And then assistants who die just go ghostrole, it is not a severe downside those days.

PR #7768, opened 2020-02-06
adds a danish accent

Amelia0010:
image


Hopekz:
I know this is a spite PR but since its actually introduced as a PR I feel obligated to leave my opinion.
I know you don't want me to say this because this is a "spite PR" but this PR seems very fitting.

You have some really nice word modifications in here such as:
"will": "vil",
"way": "vej",
"them": "dem",
"would": "ville",
"dog": "hund",
"song": "sang",
"door": "dør",
"after": "efter",
"little": "lille",
"great": "fantastisk",

But then you have weird translations like:
"well": "brønd"
"seem": "virker som",
"nothing": "ingenting",
"check": "kig efter",
"language": "sprog",

Honestly just fix up the ones that don't make too much sense and this is a "good" PR despite of the reason it was created.

If anything this modification PR is good even if you just "remove" those weird ones instead of changing them to fit.

LoliconSlayer:
Wait a second, this isn't an accent. This is just Danish!

PR #7771, opened 2020-02-07
Borgs can no longer heal themself when they are over crit threshold

alexkar598:
Borgs has a nasty tendency to fuck you up, and just when you get the upper hand, they retreat, bolt the door, and 30 seconds later you have a borg fully ready to fuck you up again because they healed like a NERD.

:cl:  
tweak: Borgs can no longer heal themself when they are over 0 health(health goes from -100 to 100)
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -380,13 +380,13 @@
 		if (!W.tool_start_check(user, amount=0)) //The welder has 1u of fuel consumed by it's afterattack, so we don't need to worry about taking any away.
 			return
 		if(src == user)
-			if(health >= 0)
+			if(health > 0)
 				to_chat(user, "<span class='warning'>You have repaired what you could! Get some help to repair the remaining damage.</span>")
 				return
 			to_chat(user, "<span class='notice'>You start fixing yourself...</span>")
 			if(!W.use_tool(src, user, 50))
 				return
-			if(health >= 0)
+			if(health > 0)
 				return //safety check to prevent spam clciking and queing
 		
 		adjustBruteLoss(-30)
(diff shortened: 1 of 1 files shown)


ghost:
this very much fucks minerborgs without the self repair module, as they need to be able to self repair to handle mobs effectively

they also have an emergency welding tool, so unlike engiborgs they can't take massive damage and constantly self repair, it only holds a little welding fuel and dragging a tank around with them is a weakness in and of itself

limit this to engineering cyborg self repair, as minerborgs do need to be able to self repair to be able to do anything on lavaland

alexkar598:
They can still heal themself, just not to full health

Mqiib:
I really think it should be the other way around, where once they take >100 damage it turns it all into burn damage resulting in them not being able to repair that. Maybe even only 75 or something, but borgs should be able to fix superficial damage to themselves just fine.

Last module on a borg goes down at -50, meaning this cuts the amount of repairs they can do on themselves to 1/3 of what it was but in the wrong direction. Hell, even if they had "armor plates" or something that had 50 health and were the only things they could repair that would be FINE.

ghost:
@alexkar598 
no they can't lol. not being at full health as a minerborg on lavaland is, for all intents and purposes, the same as not being able to repair at all. if you get hit once, you lose a module, potentially your last module, because you can't fully repair.

if you want to make yet another stupid change that inevitably gets reverted a month later when everyone realizes how terrible it is be my guest, i don't play here so it doesn't affect me, but i'm telling you outright it'll fuck minerborgs

Ratvar:
I am not sure we will have a single mining borg player to complain in a whole month, really: they are not the most popular module.
or they can just stop hunting fauna and retreat once damaged, which is an available option: they are made at maximum to mine stuff if all miners are dead, anyway, and have no need to hunt goliaths.

boodaliboo:
please make engi borgs not able to heal themselves

nesisknesis:
cant wait to revert this a month later

PR #7810, opened 2020-02-12
wip to add a new implant thats recieved from headslug

boodaliboo:
this organ will be recievable from being killed and my hope is to make it give you tk when its recieved, since xenobio slugs are no longer a thing theres no problem with farming them 
#### Changelog

:cl:  
rscadd: Adds a new organ with a wip effect
wip: This is work in progress  
/:cl:


Jarod1200:
I think It would make more sense if it gave you a ling ability instead

LoliconSlayer:
Make more sense? Sure but we don't want discount xenolings

PR #7813, opened 2020-02-13
removes heartbreaker

Ktlwjec1:
due to staff discussion. probably doesn't break anything.


nesisknesis:
how am i supposed to antag wtf

deathslayer2481:
Feature not bug

ynot01:
so what the fucks the last person supposed to do, at least let them suicide so they dont have to sit through an hour or two while everyone else gets to have fun playing their valentine antag gimmick

PR #7816, opened 2020-02-14
Medbay Rework

Hopekz:
Made using Monster's cool new tool!

1: **Added a west-side back door to medbay through the observation room**
-----

The observation room is public but the door circled in blue is medbay access.
------

2:  Added a window to the west side of cloning! Re-arranged some tables!
------

3: Moved the plant from cloning!
------

4: **Also fixed the privacy issue north of medical secure storage**
------

#### Changelog

:cl:  
tweak: Reworked medbay! The observation room is now public.
tweak: Made an entrance to medbay via the observation room.
tweak: Added a window to the west side of cloning
rsadd: Added a  plant to Viro entrance! Important!
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -54576,6 +54576,13 @@
 /obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer3,
 /turf/open/floor/plasteel,
 /area/tcommsat/computer)
+"iMP" = (
+/obj/effect/spawner/structure/window/reinforced,
+/obj/structure/cloth_curtain{
+	color = "#99ccff"
+	},
+/turf/open/floor/plating,
+/area/medical/genetics)
 "iOa" = (
 /obj/structure/spacepoddoor,
 /turf/open/floor/engine/airless,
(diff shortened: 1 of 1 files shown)


deathslayer2481:
Plant was feature not bug

Ktlwjec1 (commented):
Ok I have some issues.

Mapping technical shit:
- Firelocks needed under the new cloning airlock.
- Plating turf needed under the new cloning window.
- Lights should avoid windows if possible, so the cloning light needs moving.
- White plasteel turf needed under the west airlocks.
- Firelocks needed under the west airlocks.
- Consider blue decals in the hallway outside the west airlocks.

Other:
- I really can't approve this cloning change. 2 airlocks are very much unneeded, especially when they lead to somewhere about a 2 second walk from each other. It also looks super ugly, which is the main reason.
- The west doors I've seen on other servers and always viewed it as unneeded, and it looks like lazy mapping which is why I suggested the decals. I'm wondering if it'd be better to only have them appear on highpop, since that's when they are only really "needed".
- Maps should NOT be 100% perfect. I'm aware these are somewhat common changes done in-game, but this is a result of the station not being 100% efficient in design as it should be. To me, these changes only seem a result of this. That being said, I'm willing to keep all of this except for cloning changes if people like them.
- Alternatively, instead of the west airlocks for highpop, having it randomly switch between doors and walls would be nice.

Darkstick:
as a small detail as it is, I really don't like the window under cryo. I prefer that room to be more private.
Cloning doesn't really need two doors, but the window on the side is a nice tough.
I'd get rid of the window under cryo and that door bellow cloning.

Hopekz:
> I'd get rid of the window under cryo and that door bellow cloning.

You play medbay a lot so I respect your opinion.
Added removing that part on my to-do when I get home tonight.

Hopekz:
**Updated; Ready!**
------

Hopekz:
**Ready**
----

MegaEmpirical:
Put cloning table back where it was. The corner is useful for dragging bodies and turning.

Hopekz:
> Put cloning table back where it was. The corner is useful for dragging bodies and turning.

Good catch!
It is done.

image

Hopekz:
Edit:
I also added a curtain. Wow!
image

Ktlwjec1:
Ok last things then it is merge ready:

Smoking sign is placed on the air alarm tile. It needs to be (if you want it visible on both sides of the wall) placed on the left wall where it currently looks like it is, with no pixel_x/y value, or (if only visible from surgery) it needs be placed 1 tile down with x value 0 and y value 32.
image

Add plant back here, keep the one by viro still.
image

I'd add a helper here so people can still enter the surgery room from medbay. Also consider a non-glass airlock, I found it looked a ton better.
image

Light, sign and air alarm moving.
image

Hopekz:
Suggestions committed

PR #7817, opened 2020-02-14
Holiday christmas music changes

Ktlwjec1:
Music changed was because:
1) Link is unavailable.
2) Stalin tunes.

You will understand if you check the changes.


boodaliboo:
Fun fact the only reason that there is Russian music is I copy and pasted the music I was intending rhythm to play instead of the song meant for it

PR #7818, opened 2020-02-14
Removes an INAPPROPRIATE joke

Ktlwjec1:
0.2: Erotic content is forbidden. As a general rule, anything more sexualized than kissing and hugging is considered erotic. Do not roleplay acts of an erotic nature. Insulting people with erotic or intercourse related terms is fine as long as the insults are not graphic or explained in detail. Do not write, upload, or distribute pornographic literature.


CMOisLing:
ERP DESTROYED

PR #7824, opened 2020-02-15
Replaces a Medbay sleeper with a stasis bed

Hopekz:
Perfectly balanced, as all things should be.

People have been talking about this for months but no-one has done this PR for some reason.
Putting this PR out so it can be discussed or finally changed.

image

#### Changelog

:cl:  
tweak: Replaced a Medbay sleeper with a stasis bed. Perfectly balanced, as all things should be.
/:cl:

PR #7826, opened 2020-02-15
Replaces surgery table in the Operating Theatre with a stasis bed

Hopekz:
PR'ing this to discuss and possibly change if wanted.

What brought this PR:
image

**My medbay stasis PR didn't seem like the right place for that change so I'm bringing it over here**

image

**My personal opinion:**
Neutral on it but I do see people do this a lot.
Not sure if this falls within "making a department perfect so no upgrades need to be done" part.

#### Changelog

:cl:  
tweak: Surgery table in the Operating Theatre is now a stasis bed.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -24683,8 +24683,8 @@
 	},
 /obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer3,
 /obj/structure/disposalpipe/sorting/mail{
-	sortType = 15;
-	dir = 8
+	dir = 8;
+	sortType = 15
 	},
 /turf/open/floor/plasteel,
 /area/hallway/primary/central)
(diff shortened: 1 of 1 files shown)


jcatjmeow:
Good pr.

Lowers the capacity for newbie doctors to make mistakes with lasting consequences, and makes it easier for them to learn the surgeries since less is at stake.

Since med is often both a new player role, and a role that many players from all over the station depend on, I strongly agree with this change.

Plus, it will mean the surgery room is used more often for the healing surgeries, since they are more effective when the patient isn't constantly dying as you attempt them

Ktlwjec1:
Choose this one or #7824 , this one seems better to me but I don't mind, just don't want them both.

CMOisLing:
I dont like this but tbh I cant find a real strong reason not to have it other than, I like the idea of just having a normal surgery table at start, and its not really that much better in most situations.

Moja10:
this is just superior to the surgery table, which has no point now being on the station. muh powercreep!!!

Hopekz:
Alright. I gave this PR a chance but I need to close one of my medbay PR's.
Keeping #7824 because it has more positive support and closing this one.

PR #7827, opened 2020-02-15
Adds 9 different variations to dorms

Hopekz:
More spice!
**NOTE: carpets look much better in-game (see the first screenshot VS Variations screenshots)**

**EDIT:**
Added a replica reinforced table.
This table is fragile because the materials for it were not meant to be shaped in this way (since it is made of metal instead of plasteel)

Captureall

Variations:
1:

2:

3:

4: **Note: the table is a "Replica reinforced table" . See comment below.**

5:

6:

7:

8:

9: (idea by Firewolf34)
image

#### Changelog

:cl:  Hopek
rscadd: Added 9 new variations to dorms!
/:cl:


monster860 (changes requested):
The door and the button should not be part of the template in case two dorms get the same template.

Hopekz:
> The door and the button should not be part of the template in case two dorms get the same template.

I'm making separate templates for dorm 2. 
This is why this template is called "Dorm east door" because it is only to be used in the dorms with the door to the east.

This is because those two types of dorms have a different layout and can not be interchanged.
For instance, the second type has dressers.
I do not want to effect dorm balancing with this PR only to introduce variations.

Also, your method will influence how the instruments are laid out as well.

alexkar598 (dismissed):
Use random generators instead of templates as templates are way less lightweight than random generators and you have no need for the features templates provide. If the janitor room one is questionable this one is just bad

Hopekz:
> Use random generators instead of templates as templates are way less lightweight than random generators and you have no need for the features templates provide. If the janitor room one is questionable this one is just bad

The amount of processing power this takes doesn't even make a blip.
I am not redoing this whole PR in a different method over something that does not even make a substantial difference.
You are acting like this millisecond of processing power is like the last drop of water in a desert makes no sense for it to even be that big of an issue.

Also, I have not figured out how to accomplish the same result yet with random number generators because I want to have them themed with each other. I'll play around but no promises.

alexkar598:
I... ignore me, I'm just being shit at my job now

Hopekz:
I need to address a potential bug. Going to work on this when I get home.
Monster pointed it out in his review but I misunderstood him thinking he was talking about using  the template for dorm 2. Will remove WIP when I'm done

Hopekz:
Ready!
-----
gV3IpkOS5E

Ktlwjec1 (changes requested):
I know it's for aesthetic reasons but no **reinforced** table in dorm 4, and the chairs should be facing the table in 5 and 6. Besides that, I'm fine to merge this.

Firewolf34:
What about a run-down dorms room with maybe a prisoner's bedsheet and some missing floor tiles, maybe a barricade in the door?

Love all the variation!

Hopekz:
> I know it's for aesthetic reasons but no **reinforced** table in dorm 4, and the chairs should be facing the table in 5 and 6. Besides that, I'm fine to merge this.

**Modified.**
I made the reinforced table a "replica reinforced table". 
It is functionally identical to a regular table except I modified the Name, description, and icon.



> What about a run-down dorms room with maybe a prisoner's bedsheet and some missing floor tiles, maybe a barricade in the door?

**Great idea!**
I went ahead and added a 9th variation:
Overall pretty happy how this turned out I even made it dusty just like Perma.
image

Hopekz:
Won't be around to fix this if something is broken after the merge.
Will reopen when I come back otherwise feel free to reopen if you want to merge.

Hopekz:
Have more time to work on this and will be able to support it if merged so reopening.

Hopekz:
Ready x2!
-----

Ktlwjec1 (changes requested):
- dorms_edoor4 - Make it into a NORMAL table, not a replica reinforced/normal reinforced one. Also in the future if you do something like this, make it a new object instead of editing it on the map.
- dorms_edoor5 dorms_edoor6 - Rotate the chair to be facing the table.

Besides that it's fine, works as intended etc.

Hopekz:
Modified.

Chairs now face the "proper" direction (even though people don't face their office chairs to the table when they leave because that would be stupid but this isn't real life)
chrome_4V8bCLCaZDchrome_PLP5acA3lU

Also;
The reinforced table replica is now an official item!
It is more brittle than the usual table metal because the material were not really designed to be shaped as a reinforced table.

Ktlwjec1 (changes requested):
dorms_edoor4 - Make it into a NORMAL table, not a replica reinforced/normal reinforced one.

Hopekz:
It is now a normal table.

PR #7829, opened 2020-02-15
I CHANGE THE CLOCK CULT MAP FOREVER

Readystorm546:
This may get rid of the lag round start so people in OOC dont go "Fuck its clock cult" due to long load times.

:cl:  
bugfix: fixed a load time maybe  by making the map smaller
/:cl:


monster860:
In case anyone is wondering, what he did is make the map smaller.

In the future, please put details like this in the PR description so us maintainers don't have to either be big brained enough to think through the thought process of figuring out how we would achieve the same thing you did, or try to read through diffs.

Readystorm546:
They are not big brain enough but ill fix it

Bibby0110:
hey monster
it's called the mapdiffbot

monster860:
the map diff bot is really fucky. sometimes it doesnt work, sometimes it decides the entire map is diff and then crashes, sometimes you have varedits and they don't show up because the map looks exactly the same

Ktlwjec1:
ok let's try it, if it breaks anything it's an easy revert

PR #7830, opened 2020-02-16
makes valentines 2 days longer as a joke

SomeguyManperson:
The only RP related event so far and it only lasts 5 days
:cl:  
rscadd: makes valentines 2 days longer as a joke
/:cl:


patpol4:
based

PR #7831, opened 2020-02-16
Remoes puss translation from cowboy accents

Hopekz:
Even though this is a legitimate translation I am removing it because it may cause confusion.
Removed "face" being translated to "puss" in both the cowboy accents.

Now "face" translates to "mug" instead.

image

Now it won't surprise people when they say something like:
"I need to wash my face"

#### Changelog

:cl:  
tweak: Removed cowboy puss. It's now a mug.
/:cl:

### DIFF
--- a/strings/accents/accent_cowboy.json
+++ b/strings/accents/accent_cowboy.json
@@ -90,6 +90,7 @@
 	"friend": "partner",
 	"horse": "hoss",
 	"color": "cuhlor",
+	"face": "mug",
 	"plain": "puh-lain",
 	"girl": "gal",
 	"red": "ree-ehd",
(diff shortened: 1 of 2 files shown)


Amelia0010:
Maybe just don't translate words but translate speech patterns instead

EDIT: Retard

Hopekz:
> Maybe just don't translate words but translate speech patterns instead
> 
> EDIT: Retard

That wouldn't accomplish what I'm trying to do with these "accents". 
If anything that would make this bland.

You are welcome to PR a speech pattern type solution if you would like.
Know that a system like that will just be a little boring without "fun" translations.

You literally can't do any of these translations with simple regexes "Retard"

image

Amelia0010:
No, and you shouldn't, because preventing users from saying words like "diethylamine" is retarded and still not how accents work. 



Accents should be inherent to the character, word choice should be up to the player.

Ktlwjec1 (approved):

PR #7847, opened 2020-02-18
Adds a RD hardsuit to RD office in yogspubby

Hopekz:
Imagine making a map and thinking that there should be a dead plant in the corner instead of a RD suit.
Yogspubby # 1 best map!

Before:
------
image

After:
-------
image

#### Changelog

:cl:  
rscadd: Gave the RD a hardsuit on YogsPubby.
rscdel: Removed some dead plant to make room for it. 
/:cl:


Ktlwjec1:
Replace the bomb testing site RD hardsuit with a different one if you're going to do this, so there's not 2 RD hardsuits.

Hopekz:
Not needed.

PR #7854, opened 2020-02-19
Adds Egg RP

monster860:
Egg RP

> **the egg** says, "Egg"

:cl: monster860
add: Adds Egg RP. Simply suicide while holding an egg to EGG RP.
/:cl:


Potototatoh:
shoe rp

PR #7856, opened 2020-02-20
fixes icy veins being terrible

SomeguyManperson:
:cl:  
bugfix: Icy veins will no longer get you to -200 KELVIN
/:cl:


yacabo:
Like actually lower than absolute zero?

SomeguyManperson:
yes

MegaEmpirical:
I think cyro sting did this too soemtimes

SomeguyManperson:
Cryo sting just adds an ungodly amount of frost oil which if I'm right isn't the problem since you can only be reduced to 50 kelvin from it unless things don't work as intended

PR #7861, opened 2020-02-21
More food crafting tidy up

Ktlwjec1:
Names capitalised for the crafting menu, names the same (mostly - exceptions such as blood soup being tomato soup still) across the crafting menu and the actual food, demodularises snacks files, grammar checks.


Ktlwjec1:
Yep, looks like a perfect pr to me! Keep up the good work!!

PR #7868, opened 2020-02-23
vent icon fix

Ktlwjec1:
I need this specific one for my map and it currently has no sprite because of this!!!!!!!!


monster860 (approved):
nice meme

PR #7874, opened 2020-02-23
Cycling Airlocks

monster860:
I wrote my own cycling airlocks with blackjack and hookers! Now, I know there's already two other varieties of cycling airlock in the game, but this one is better because you can map them in without having to change 50 vars, and they can be built in game. It also supports the ability to cycle between two rooms of different atmospheres and has controls for how much you want to allow the two rooms' air to mix.

image

:cl: monster860
add: Cycling airlock controllers! Build them using metal and the appropriate electronics.
tweak: All external airlocks on boxstation are now the cycling kind. In addition, the firelocks under these airlocks have been removed.
/:cl:


yacabo:
There should be an emergency override option

alexkar598:
Make it locked on engi access to modify shit so that bummer mc tide cant walk up to the airlock and make it pump plasma into the station

monster860:
its locked so only atmos techs can modify the settings

alexkar598:
But does it show visually

PR #7888, opened 2020-02-26
removes duplicate jester outfit

nesisknesis:
its like the same thing, it looks different but its like the same thing and I like the other one better

#### Changelog

:cl:  Identification
imagedel: deletes a duplicate clothing icon
/:cl:


Readystorm546:
I can confirm this is true

PR #7900, opened 2020-02-29
A wild empty line has appeared!

Bibby0110:
I've had this as an issue on a downstream.
Mirroring it to here so guys don't have to run into it in the future.

(Probably possible to fix it with proper regex magic, but regex hurts)
#### Changelog

:cl:  
bugfix: No longer shall empty lines crash our scientific progress!
/:cl:


alexkar598:
imagine not knowing that ! is higher than > and making a fucked up thing

Bibby0110:
oops

Bibby0110:
it compiled okay?

Bibby0110:
> imagine not knowing that ! is higher than > and making a fucked up thing

Technically correct, but my shitcode made it work anyways

PR #7905, opened 2020-03-03
reverts cycling airlocks

boodaliboo:
these arent fun and a lot of the people in the game i saw who messed with them agreed there hell to deal with and i personally saw it kill a few people


monster860 (changes requested):
don't revert the whole thing just revert the map changes

boodaliboo:
yeah i can do that

PR #7916, opened 2020-03-04
The clock cult stun (Kindle) Changes

Readystorm546:
Wow amazing PR. Stun is now less broken and more balance

:cl:  
tweak: >You can now move while using kindle. However you are slowed down
tweak: >Kindle now takes 5 seconds to conjour and will only last 5 seconds after that. Instead of taking 3 seconds to conjour and it lasting 14 seconds
tweak: >Kindle now costs more power. from 125 > 150
/:cl:


Hopekz:
Tested this in-game yesterday not sure how I feel about this yet.
This is a clear buff to kindle but I am unsure if they need this or not.

With this the cultist is now able to retreat while being chased and then turn around to instantly stun whoever is pursuing them. Of course with the stun now being 5 seconds this is a non-issue to people traveling in groups but it does let a cultist kill sec fairly quick with a spear for example during this stun.

Of course this isn't anything "new" regarding the DPS but the main thing this adds is the fact that you can run from sec while charging this instant stun where before if sec pressured a cultist they would be forced to fight with whatever they brought with them (stun visor, spear)

Readystorm546:
The main point was to not git rid of the use of Kindle but instead stop it from being "Tp > kindle >cuf>capture"

SomeguyManperson:
This doesn't change the status effect so the stun still lasts as long as it used to (roughly 10 seconds shortened by taking damage). This turns kindle from a utility used to cuff people into a weapon as well since it can be prepared while moving. Alsp makes botching an initial attack not make you as vulnurable since you can immediately start another kindle charge as long as you can stay out of disarming range.

Readystorm546:
Do you have any recommendations for improving this?

Ktlwjec1:
Remove mobility and then this should be fine, that seems to be the issue with this change.

Readystorm546:
Well i was planing for something else i cant code it. Hoping for someone else to help.

Ktlwjec1:
I'll mark this as WIP for you then, ask around in coding channels and see if anyone can help.

alexkar598 (changes requested):
e, you copy pasted my code but its not indented correctly thanks to discord

Hopekz:
I was already typing up a comment. What Alex said these new changes are space indented instead of tab indented.

PR #7751, opened 2020-02-04
Casually removes the ability for admins to do SQL injection and destroy yogstation

alexkar598:

PR #7657, opened 2020-01-17
Achievement Get Part 1 - Greentext Achievements

Altoids1:
Hopek and Readystorm are paying me for this.
image

## Overview

Included are 18 of the 88 achievements I'm being paid to add to the game, all of them having to do with greentexting.

There are also unique achievements for the leaders of team antag groups winning, too. Like Eminences get both the default "win as clockie" achievement as well as the "win as Eminence" one.

## Coder Warning

Doing this involved the creation of a new achievement type, ``/datum/achievement/greentext``. All achievements related to greentexting should be put under this type so that the user can also get the generic "get a greentext" achievement, too.

Additionally, ``/datum/antagonist`` got a new var: ``greentext_achieve``, which is the type of an achievement the antag will receive if they greentext as that antag type.

Not all antag types have been converted to the new ``greentext_achieve`` method, due to the excessive copypaste and lack of inheritence when it comes to determining win condition and objective accomplishment.

## Changelog
:cl:  Altoids
rscadd: Gangs, Clownops, Generic antags, EAAs, IAAs, Culties, Clockies, Revs, Sentient Diseases, Revenants, Vampires, Pirates and Blobs now all have unique achievements for greentexting!
/:cl:

PR #7917, opened 2020-03-05
Achievement Get Part 3 - Cargo Achievements (and bugfixes!)

Altoids1:
image

This adds four more achievements, as part of #7657, of the Cargonian variety.

## Coder Warning
For some reason, #7751 made both read and write of the database entries for achievements no longer work.

I have... no real idea why, but I've decided to revert Alexkar's changes and do the sanitation through new methods.

All achievements now use the ``initial()`` variants of their properties to prevent tampering.

In addition, all ckeys used in queries are run through ``ckey()``, which would remove any and all special characters that're necessary to do a SQL injection.

Doing it this way should not only be more performative (using native BYOND procs instead of several sequential calls to bsql.dll) but also actually work.

#### Changelog
:cl:  Altoids
rscadd: Cargo now has a couple achievements of their own!
bugfix: Fixed bug that caused achievements to not load nor save.
/:cl:


boodaliboo:
Assistant

Altoids1:
# Ready for review.

Hopekz (approved):

code/modules/economy/account.dm
```diff
@@ -30,6 +32,14 @@
...
 	if(account_balance < 0)
 		account_balance = 0
+	else if(account_balance > 1000000 && !is_bourgeois) // if we are now a millionaire, give the achievement
+		//So we currently only know what is *supposed* to be the real_name of the client's mob. If we can find them, we can get them this achievement. 
+		for(var/x in GLOB.player_list)
+			var/mob/M = x
+			if(M.real_name == account_holder)
+				SSachievements.unlock_achievement(/datum/achievement/cargo/bourgeois, M.client)
+				is_bourgeois = TRUE
+				break
```
alexkar598: You can't assume there aren't going to be any duplicated names due to genetic fuckery or something like that
```suggestion
```

code/modules/awaymissions/bluespaceartillery.dm
```diff
@@ -47,6 +47,8 @@
 	if(usr.contents.Find(src) || (in_range(src, usr) && isturf(loc)) || issilicon(usr))
 		priority_announce("Bluespace artillery fire detected. Brace for impact.")
 		message_admins("[ADMIN_LOOKUPFLW(usr)] has launched an artillery strike.")
+		if(usr.client)
+			SSachievements.unlock_achievement(/datum/achievement/cargo/bsa, usr.client)
```
alexkar598: This is an admin only version, not the cargo not
```suggestion
		message_admins("[ADMIN_LOOKUPFLW(usr)] has launched an artillery strike.")
```

code/modules/economy/account.dm
```diff
@@ -30,6 +32,14 @@
 	account_balance += amt
 	if(account_balance < 0)
 		account_balance = 0
+	else if(account_balance > 1000000 && !is_bourgeois) // if we are now a millionaire, give the achievement
+		//So we currently only know what is *supposed* to be the real_name of the client's mob. If we can find them, we can get them this achievement. 
+		for(var/x in GLOB.player_list)
+			var/mob/M = x
+			if(M.real_name == account_holder)
+				SSachievements.unlock_achievement(/datum/achievement/cargo/bourgeois, M.client)
```
alexkar598, line 40: Please add a check within the unlock_achievement proc to check if the client exists, its too easy to forget a ``if(whatever.client)``

alexkar598 (changes requested):

PR #7927, opened 2020-03-06
I ded rpd pr

Readystorm546:
Welp guess what... RPD killed me because sparks so here i am getting rid of those sparks. Turns out this also reduces lag. So hah. Sparks light plasma even though its not a broken time. So yeah none of that. I also tested this it works.

:cl:   
rscdel: Removed RPD Sparks  
i ded  
reduce lag
/:cl:


Hopekz:
I like sparks flying out of my RPD when I select a different mode.
Kind of wish the RPD would actually spark more/give more feedback when placing.

nesisknesis:
not funny didnt laugh

Jarod1200:
Sparks are an important part of teaching people how to safely build stuff with plasma in the air. Only the safe or fireproof will survive

Mqiib:
You're not a true atmosian until you've killed yourself trying to fix pipes in a plasma leak. Thumbs down, don't remove this secret rite of passage.

Readystorm546:
I dont think you all understand. Not only does this reduce lag when trying to select pipes. It also makes no sense that the thing sparks in the first place. Other things in the game that "spark" are damaged items. the RPD is not damaged. It still makes sound... However it does not catch you on fire when you change pipes in plasma.

Xoxeyos:
Perhaps you could somehow fix the lag instead?

alexkar598:
Closing due to I ded and strong negative feedback

git gud

PR #7928, opened 2020-03-07
Adds 6 placement variations to brig cells (2 main + 4 perma)

Hopekz:
**Cell placement Spice!**

Capture_game2
Capture_game1

Placement variations:
-----

**Main cells:**

main1
main2

**Perma:**

perma1
perma2
perma3
perma4

#### Changelog

:cl:  Hopek
rscadd: Added 6 placement variations to brig cells (2 main + 4 perma)
/:cl:


Ktlwjec1:
As nice as this is they're just not big enough changes - the same issue the janitor room changes had. What I would suggest is to do variations of larger areas and include more changes in a future PR - for example, perma + cells as 1 big change.

PR #7932, opened 2020-03-07
Adds 4 placement variations for surgery

Hopekz:
**Surgery Spice!**

Capture_ingame

Variations:
-----

1

2

3

4

#### Changelog

:cl:  Hopek
rscadd: Adds 4 Variations for surgery
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -9296,14 +9296,6 @@
 	},
 /turf/open/floor/plasteel/white,
 /area/science/nanite)
-"ats" = (
-/obj/structure/table,
-/obj/item/scalpel{
-	pixel_y = 12
-	},
-/obj/item/circular_saw,
-/turf/open/floor/plasteel/white/side,
-/area/medical/sleeper)
 "att" = (
 /obj/structure/cable{
 	icon_state = "1-2"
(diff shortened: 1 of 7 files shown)


deathslayer2481:
some of these are objectively worse than others.

Hopekz:
> some of these are objectively worse than others.

Which ones?
None of these are set in stone and are subject to change based on feedback.
I am fully willing to modify any of these.

I can see 3 not being as "efficient" because there are tables on both ends of the surgery table so if that is what you're referring to I can maybe look into making the actual operating table more 'inward" so the tables are flush with the operating table?

deathslayer2481:
the ones that make the operating table have tools on both sides, it means that you have to cram more tools into a smaller space.

Hopekz:
> the ones that make the operating table have tools on both sides, it means that you have to cram more tools into a smaller space.

Oh. Well, that part actually is intentional as I have a finite amount of real estate in this room so if I remove a table I have to split the contents between multiple tables.

If you have a specific example of one that you don't like I can work on it to make it more bearable otherwise not sure what to do about that one (:

Moja10:
I'm not sure surgery needed randomization period imho

Ktlwjec1:
3 and 4 are the only ones with enough variation but they have other issues such as surgical tools not being accessible without walking to the tables (which WILL be annoying for medical players) and the operating console being crammed in a gap behind the table. I'll leave this open a day longer but I'll likely just close this tomorrow.

Hopekz:
Mapping pope is not happy

PR #7935, opened 2020-03-08
cheese always heals mice instead of not healing them if they are already supercheese mode

SomeguyManperson:
:cl:  
tweak: cheese will now always heal mice when they eat it
/:cl:


Moja10:
I ded

PR #7919, opened 2020-03-05
Port ice moon from tg

SomeguyManperson:
original pr: tgstation/tgstation#49680

a new planetside map with multiple z-levels and some new ruins and fauna

#### Changelog

:cl:  whoneedspacee
rscadd: Icebox Station has been added, explore the depths of the Ice Moon with boxstation right on top of it, don't forget your coat and internals!
rscadd: You can now mine rock flora into volcanic ash to make glass.
/:cl:

PR #7936, opened 2020-03-08
End test for ice moon

SomeguyManperson:
Reverts yogstation13/Yogstation#7919


deathslayer2481:
Bad

Hopekz:
> Bad

Why? Elaborate?
This a "test merge" so it's literally waiting on feedback to get "better".

PR #7944, opened 2020-03-09
Assorted moon map fixes and Joe the goliath

SomeguyManperson:
:cl:
bugfix: some pipes are where they should be on moonmap
bugfix: ladder can lead to the bottom level of the ice moon
bugfix: lavaland ruin shouldn't have its own atmos
rscadd: adds joe the mining goliath!
/:cl:


Marmio64:
there's already a joe, joe the mouse. Can you rename the mining goliath to 'Bob' or something? Just so we dont re use names.

PR #7948, opened 2020-03-09
Glowshrooms No Longer Absolutely Murder Server Performance

Altoids1:
Someone should pay me for having done this.

# Overview

Glowshrooms now just straight-up stop attempting to reproduce past the 6th generation.

There also can *never* be more than 1024 glowshrooms at once in the video game.

Further, when a glowshroom still has seeds it can use to spread itself, but it can't find a tile to spread to, instead of checking every 2 seconds, it checks progressively less often the more times it tries, which should reduce the lag from practically-idle glowshrooms.

I've also done a *ton* of code cleanup, which removed a bunch of unnecessary lag caused by these fucking things.

# Coder Warnings
So there were a couple things wrong with glowshroom code.

1. It called ``view(3,src)`` tons of times per seed when it only needed to do so once
2. It called ``locate()`` on dozens of tiles per seed when it only needed to do so once
3. Whoever wrote this has a poor understanding of how exponentials work.

To elaborate on No. 3, look at this code that determines whether a glowshroom can plant with a seed:
```dm
if(prob(1/(generation * generation) * 100))//This formula gives you diminishing returns based on generation. 100% with 1st gen, decreasing to 25%, 11%, 6, 4, 2...
```
The *foolish* among you may think that this should make it pretty difficult for glowshrooms to overspread. However, even with this, the population can still soar in worst-case scenarios:
image
*On this graph, black represents the total population of glowshrooms, while purple represents the number born per generation.*

Given maximum yield and mildly poor luck, the population can reach **22,000**, which is about a third of the area of the station Z-level and double the number of floor tiles on Boxstation. Ridiculous.

## Lightsources are a thing.

The other problem is that this results in straight-up 22,000 lightsources being emitted. This is utterly cancer, especially since a lot of this is from multiple glowshroom entities existing on the same turf.

### Wait, what?

Yes, that's right. Instead of just blending sprites together sensibly, it used to be that glowshrooms just fucking had several objects stacked up on the same tile. The logic to determine how they stacked up and how to stop them from growing on the same walls was also very arcane and bizarre.

Now, when some glowshroom tries to grow onto the wall of a tile that already has a glowshroom, the wallshroom's sprite is just blended onto the already-existing glowshroom as an overlay, easy-peasy.

## Changelog
:cl:  Altoids
bugfix: Glowshrooms no longer absolutely murder server performance.
/:cl:

### DIFF
--- a/code/game/objects/effects/glowshroom.dm
+++ b/code/game/objects/effects/glowshroom.dm
@@ -119,7 +119,8 @@
 			if(locate(/obj/structure/glowshroom in view(1,newloc))) // and there's a glowshroom nearby our target loc
 				continue // then we ain't moving here
 			// We know that there's no glowshroom here so we can just go ahead and plant
-			var/obj/structure/glowshroom/child = new type(newLoc, myseed, TRUE, generation + 1)
+			var/obj/structure/glowshroom/child = new type(newloc, myseed, TRUE, generation + 1)
+			child.generation = generation + 1 // I have to do this to stop the "variable defined but not used" warning
 			--seeds_to_spread
 		else // if we're willing to overgrowth
 			var/obj/structure/glowshroom/glowyboi = locate(/obj/structure/glowshroom) in newloc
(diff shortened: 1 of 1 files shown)


monster860:
glowshrooms are a powergaming fungal infection that infects the server, the station, and the performance

Altoids1:
I'll also note that making glowshrooms on the same tile part of each other means that they all share the same healthpool, effectively cutting the HP of glowshrooms in maint to a third.

This also makes it harder for glowshroom to spread through maint.

Both of these things should help contribute to improving the balance of these things, as well as their performance.

MegaEmpirical:
Fucking glorious. I assume this will ally to their mutant shadow shroom too?

MegaEmpirical:
Noooooooooooo

PR #7960, opened 2020-03-10
Makes AIs unable to be Cheesed/Launchpadded

JamieD1:
Made it so AIs cannot be teleported unless unachored to stop abuse.

#### Changelog

:cl:  
tweak: AIs need to be unanchored to be able to be moved via launchpad 
/:cl:

### DIFF
--- a/code/game/machinery/launch_pad.dm
+++ b/code/game/machinery/launch_pad.dm
@@ -150,7 +150,7 @@
 			var/mob/living/silicon/ai/I = ROI
 			if(I.move_resist > 1000)
 				to_chat(user, "<span class='warning'>ERROR: Launchpad overloaded, unable to operate.</span>")
-				return
+				continue
 		// if it's anchored, don't teleport
 		var/on_chair = ""
 		if(ROI.anchored)
--- a/code/game/machinery/launch_pad.dm
+++ b/code/game/machinery/launch_pad.dm
@@ -145,6 +145,12 @@
 	for(var/atom/movable/ROI in source)
 		if(ROI == src)
 			continue
+		// Dont teleport unanchored ais
+		if(isAI(ROI))
+			var/mob/living/silicon/ai/I = ROI
+			if(I.move_resist == 2999)
+				to_chat(user, "<span class='warning'>ERROR: Launchpad overloaded, unable to operate.</span>")
+				return
 		// if it's anchored, don't teleport
 		var/on_chair = ""
 		if(ROI.anchored)
--- a/code/game/machinery/launch_pad.dm
+++ b/code/game/machinery/launch_pad.dm
@@ -148,7 +148,7 @@
 		// Dont teleport unanchored ais
 		if(isAI(ROI))
 			var/mob/living/silicon/ai/I = ROI
-			if(I.move_resist == 2999)
+			if(I.move_resist > 1000)
 				to_chat(user, "<span class='warning'>ERROR: Launchpad overloaded, unable to operate.</span>")
 				return
 		// if it's anchored, don't teleport


code/game/machinery/launch_pad.dm
```diff
@@ -145,6 +145,12 @@
 	for(var/atom/movable/ROI in source)
 		if(ROI == src)
 			continue
+		// Dont teleport unanchored ais
+		if(isAI(ROI))
+			var/mob/living/silicon/ai/I = ROI
+			if(I.move_resist == 2999)
```
monster860: this is shitcode. fix it.

monster860 (changes requested):

boodaliboo:
How do you suggest someone gets the ai if there not the head because turrets are a huge problem and there’s very few ways to get it

Hopekz:
This probably just happened (wasn't a part of it so not sure)

But why would teleportation care about being anchored? Every other item in the game is able to be teleported without it caring about being anchored.

JamieD1:
> This probably just happened (wasn't a part of it so not sure)
> 
> But why would teleportation care about being anchored? Every other item in the game is able to be teleported without it caring for being anchored.

You cant teleport machinery can you?

Hopekz:
> > This probably just happened (wasn't a part of it so not sure)
> > But why would teleportation care about being anchored? Every other item in the game is able to be teleported without it caring for being anchored.
> 
> You cant teleport machinery can you?

The AI isn't machinery the AI is a mob. A player. Much like a Cyborg that can't move on its own.

Jarod1200:
For reference this has happened several times in the last few days by the same person

JamieD1:
> > > This probably just happened (wasn't a part of it so not sure)
> > > But why would teleportation care about being anchored? Every other item in the game is able to be teleported without it caring for being anchored.
> > 
> > 
> > You cant teleport machinery can you?
> 
> The AI isn't machinery the AI is a mob. A player. Much like a Cyborg that can't move on its own.

All im suggesting is when its ANCHORED it cannot be teleported. Ive counted more then ten times an ai has been teleported out of its core just to check its laws, its not supposed to be like that. Forget logic for a moment and think how it affects the round, you wanna give medical doctors and people who shouldnt be able to do it the power to?

code/game/machinery/launch_pad.dm
JamieD1: Care to suggest what is shit about it.

Hopekz:
> > > > This probably just happened (wasn't a part of it so not sure)
> > > > But why would teleportation care about being anchored? Every other item in the game is able to be teleported without it caring for being anchored.
> > > 
> > > 
> > > You cant teleport machinery can you?
> > 
> > 
> > The AI isn't machinery the AI is a mob. A player. Much like a Cyborg that can't move on its own.
> 
> All im suggesting is when its ANCHORED it cannot be teleported. Ive counted more then ten times an ai has been teleported out of its core just to check its laws, its not supposed to be like that. Forget logic for a moment and think how it affects the round, you wanna give medical doctors and people who shouldnt be able to do it the power to?

Isn't the launchpad briefcase a traitor only item? It can be argued that players playing the syndicate are not "people who should not be able to"
If anyone is able to acquire the launchpad briefcase then it should probably be changed.

Mqiib:
I do think being able to just yeet the AI out of it's core by building two machines is pretty lame. If you can just teleport the AI wherever, what's the point of anchoring the core in the first place? What's the point of the AI being in a heavily-fortified core if you can just steal it without ever even needing to go near the core at all? There are already a slew of ways to break into the core and kill/card the AI with minimal effort, I don't see why the ability to just steal the stationary, bolted-to-the-floor AI core with a couple clicks and some basic math should be one of them.

Hopekz:
I got more context on this.

It is kind of cheap IMO if you can just build a machine in dorms as a non-antag and just teleport yoink the AI out of there.

I initially thought this was just the launchpad briefcase which would be ok IMO but it seems to be the regular launchpad.

code/game/machinery/launch_pad.dm
JamieD1: Fixed ShittyCode

alexkar598:
do it for all objects with high move force instead of just ai

monster860 (approved):

code/game/machinery/launch_pad.dm
```diff
@@ -145,6 +145,12 @@
 	for(var/atom/movable/ROI in source)
 		if(ROI == src)
 			continue
+		// Dont teleport unanchored ais
+		if(isAI(ROI))
+			var/mob/living/silicon/ai/I = ROI
+			if(I.move_resist > 1000)
+				to_chat(user, "<span class='warning'>ERROR: Launchpad overloaded, unable to operate.</span>")
+				return
```
monster860: just uh make it do `continue` to skip the AI that way other things on the same tile as the AI will still teleport

monster860 (changes requested):

code/game/machinery/launch_pad.dm
JamieD1: With current code it still teleports items, here:

https://gyazo.com/fe2ae4c6cd8ad2e7d35aedb4c875c60a

JamieD1: Done as requested

JamieD1: https://github.com/yogstation13/Yogstation/pull/7960/commits/22105ea49ee7d53af8de64a304625d1322bb9be7

PR #7961, opened 2020-03-11
Random accent is now a neutral quirk

alexkar598:
Yes hello let me make myself harder to understand but not impossible by any means, just a pain and gain a quirk point. This is absolutly balanced what can go wrong

haha, no.

Random accent quirk is now a neutral quirk, fuck you silent gamers

### DIFF
--- a/code/datums/traits/negative.dm
+++ b/code/datums/traits/negative.dm
@@ -577,19 +577,3 @@
 	H.remove_language(/datum/language/common)
 	if(!H.can_speak_language(/datum/language/draconic) && !H.can_speak_language(/datum/language/machine))
 		H.grant_language(/datum/language/japanese)
-
-
-/datum/quirk/random_accent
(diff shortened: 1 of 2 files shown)


alexkar598:
something something but my RP

Hopekz:
Good PR 👍 
I was thinking about doing this for a while but didn't get around to it.

Even though it is a negative trait IMO wiki Mord won't STFU about me ruining the entire game by giving people a quirk point for a negative trait like quirk points even matter anyways.

+1 because it literally doesn't matter and people use accents because they're fun not because of this point anyway (I don't see anyone picking this trait for the point anyway except when people are trying to find an accent they like)

SomeguyManperson (approved):

LoliconSlayer:
> wiki Mord won't STFU about me ruining the entire game

Nice exaggeration

Issue #7780, opened 2020-02-07
Accents are borked


Featherfield:
For example old english "go" is "hie" so if you try and say god it will come out as "hied"
same with "fat" being "fulsome" so you cannot say father without it being "fulsomeher"
this applies to most words like

Featherfield:
there is now a PR that fixes this issue so you might aswell close this one.

PR #7962, opened 2020-03-11
Fixes Accents Incorrectly Capturing on the Middle of Words

Altoids1:
Hopek paid me to do this.

## Overview

I've completely reconfigured how the accent system does its replacement. It should now correctly capture on whole words, and also be capable of altering the pronunciation of non-word sequences.

## Coder Warning

Every single Key in the accent .jsons is now a regex string.
In order to make your word replacement a true word replacement, you must encase the key with ``\b``'s, like so:
```json
"\bcat\b" : "kitty"
```

Additionally, there is now a priority system. Phrases are replaced first, then words, then non-words. This should make it easier to say, change all Rs to Ws in a catgirl language while also doing word-replacements on words that have Rs in them, as the Rs won't be changed until after the word replacement takes place.

Otherwise, priority should be sequential, as it was before; replacements described at the top of the .json are done before the ones below it. There's no real way of making them all work at once in parallel so you're going to have to keep that in mind if you ever create or modify accents.

There is also a small optimization when it's just replacing plaintext words; it skips over running the regex engine and instead just tries accessing the word-to-replacement table with each word in the sentence, replacing them when it hits something.

## Changelog
:cl:  Altoids
bugfix: Accents now work moreso as intended, and do not capture incorrectly on the middle of words as much.
/:cl:


alexkar598:
so you are telling me that for a single message, 900 regexs are being executed?

Hopekz:
> so you are telling me that for a single message, 900 regexs are being executed?

In addition to fixing the word issue this change to the framework allows most accents to be converted into a mix of speech pattern/word replacements.

What this means:
Accents can now mostly converted down to simple regexes in addition to word replacements.

Here is one example of many where this would actually save a lot of regexes being run because it won't have to go through a list every time:

In this one example if we were to have "ing" replace with "in'" we could remove hundreds of word translations and convert that part to a speech pattern.

image


This change overall is "faster" once the accents are revamped the way this frameworks lets them be not that performance was ever an issue for accents in the first place.

Featherfield:
No more hied as god this is good

Altoids1:
image
# Ready for Review.

alexkar598:
once again, running 300 regexs, per word, per message?

Hopekz:
Forgot to post this here but this also fixed #7780

Hopekz (approved):

alexkar598:
This was test merged for a bit and didnt break

Amelia0010:
It actually does break.

\btest\b will catch testär because of byond's very lacklustre support for unicode, and the swedish speech impediment actively replaces stuff with äöå. So you know, it'll break

https://regex101.com/r/R03hTV/1

PR #7970, opened 2020-03-12
Bucklecuff timer in seconds

SomeguyManperson:
currently it displays in minutes when most of the time it's below 60 seconds, meaning it'd display 0 or 1 minutes
:cl:  
bugfix: bucklecuff resist alert is now displayed in seconds so it doesn't round to 0
/:cl:


Hopekz:
The real solution is to make it take 60 minutes instead of seconds so that the previous timer gets used.

PR #7981, opened 2020-03-15
C++ Monstermos - Putting the 99% LAG FREE in 99% LAG FREE

monster860:
C++ CODE: https://github.com/yogstation13/extools/tree/monstermos

Monstermos now runs in c++ land!

Things that are in c++ land:
- Gas mixtures now live in c++ land, with the /datum/gas_mixture being a shell around it
- The turf grid now also exists in c++ land
- Atmos processing happens in c++ land
- Turf gas overlay calculations

Things that aren't in c++ land yet, but I should put in c++ land:
- Gas reactions


Readystorm546:
So wip?

alexkar598:
Monstermos but now its proprietary so noone else can remove it

ThatLing:
Waiting for the benchmark results

MCHSL:
```
gas mixture transfer
                      Ops/sec
before             -    26,824
remove inner_moles -    33,058
transfer_to proc   -    52,910
extools            - 2,015,134
```

alexkar598:
needs to work with travis, we cant affort to lose travis especially with the current maintainer situation

monster860:
Don't worry, I'll throw up a linux build before merging the PR

ghost:
`This thing has a really high chance of crashing the server. So uh.... yeah that's fun. Prepare for that`

doesn't that kind of ruin the entire point lol

Firewolf34:
Though technically, space is cold, oftentimes space *feels* warm, after a time. Because in a vacuum, there's no convection - you're relying on mostly radiative heat transfer. So unless you touch something, or have some form of radiator system attached to you, when in space you'll actually overheat most of the time even with normal exertion, because your body has no practical way to remove heat from itself.

This was a real problem for the first spacewalks when space-suits didn't really have any cooling. On Voskhod 2, one of the first spacewalks, Cosmonaut Leonov reportedly had about half his suit filled with sweat by the end of the relatively short EVA. Remember, you can't evaporate sweat in a spacesuit.

So the point is technically Monster you've got it right from the crew's perspective. You probably won't feel cold in space. You'll probably sweat your balls off, unless you're exposed directly to vacuum, in which case your sweat will helpfully instantly vaporize off your skin.

(By the way, fucking excellent PR here. This is amazing.)

PR #7983, opened 2020-03-15
Several changes to vampire (mostly buffs)

SomeguyManperson:
LIST
revive now requires 100 total blood instead of 0 which used to be like 300 or 400 or something
grave fever now requires 200 total blood instead of 175 because it's still just a shitty murder ability that is good at killing people who don't know what a doctor is
organized the ability list based off total blood requirement instead of some abilities randomly being placed at stupid spots
total blood is now not halved from succ because you'd end up needing an average of 1500 blood which would be fine if vampires weren't terrible at defending themselves and sucked fast but are terrible at defending themselves and suck slow
usable blood is now 75% instead of 50% of a succ
succ noise is now only hearable from 3 tiles away instead of WHAT A BEAUTIFUL DUWANG **C H E W**
hypnotize has a mute attached so it's no longer terrible
consequentially, hypnotize only works on carbons instead of being able to be ineffectually used to stun simplemobs (doesn't do anything) or borgs (who you'd probably need a flash to set up the hypnotize stun and why not just kill them with that)
ability spacing is now better
a very half-assed attempt to reorganize spells by cost was made
oh also coffins now heal clone damage because they heal every other kind of damage and are still unused

:cl:  
tweak: vampires suck quieter
tweak: vampire revive requires 100 total blood up from 0
tweak: grave fever requires 200 total blood up from 175
tweak: vampires are way more efficient as sucking blood
tweak: hypnotize now has a mute attached due to how terrible it is
tweak: hypnotize can no longer be used on borgs
tweak: vampires can heal clone damage from sleeping in a coffin
/:cl:

### DIFF
--- a/yogstation/code/datums/antagonists/vampire.dm
+++ b/yogstation/code/datums/antagonists/vampire.dm
@@ -25,16 +25,16 @@
 		/datum/vampire_passive/vision = 75,
 		/obj/effect/proc_holder/spell/self/shapeshift = 75,
 		/obj/effect/proc_holder/spell/self/cloak = 100,
-		/obj/effect/proc_holder/spell/targeted/disease = 175,
-		/obj/effect/proc_holder/spell/bats = 250,
+		/obj/effect/proc_holder/spell/self/revive = 100,
+		/obj/effect/proc_holder/spell/targeted/disease = 200,//why is spell-that-kills-people unlocked so early what the fuck
 		/obj/effect/proc_holder/spell/self/batform = 200,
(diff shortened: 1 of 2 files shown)


jcatjmeow:
Vampires suck, better now

Vampires suck better now

SomeguyManperson:
both accurate

boodaliboo:
Sure

Hopekz:
Vampires right now are a joke.
Not even worth their own game mode.
Hopefully, this puts us closer to fixing it.

PR #7987, opened 2020-03-15
"make new vampires" objective for vampires

Darkstick:

Vampires already have the ability to "convert" people into vampires, but this ability seems very under used, which is strange considering how it feels like it was intended to be used more. There are always very few round start vampires, and their threat level is low, making the round feel relatively dull.

With this new objective, I'm hoping to make vampire rounds more tense, as leaving vampires be might mean having to deal with more vampires, or having a bigger possibility of falling victim to one.

Each round start vampire will get a convert 1 to 3 crew members objective.

New vampires get their own section on the round end card called "new bloods" and will ONLY receive a drain objective.

#### Changelog

:cl:  
rscadd: new vampire objective
rscadd: vampire subtype with less objectives and own round end card spot
rscadd: Sleep stun after conversion to avoid immediate vengeance kills
tweak: changed the conversion text to discourage vengeance
tweak: replaced two bits of stinky writing with my own less stinky writing
tweak: lowered the price of Lilith's pact to 300, and it's unlocking point at 320 instead of 400 and 450
/:cl:


yogstation/code/datums/antagonists/vampire.dm
```diff
@@ -99,6 +106,8 @@
 	to_chat(owner, "<span class='danger bold'>You are a creature of the night -- holy water, the chapel, and space will cause you to burn.</span>")
 	to_chat(owner, "<span class='notice bold'>Hit someone in the head with harm intent to start sucking their blood. However, only blood from living creatures is usable!</span>")
 	to_chat(owner, "<span class='notice bold'>Coffins will heal you.</span>")
+	if(full_objectives == FALSE)
+		to_chat(owner, "<span class='notice bold'>You are not required to respect or obey other vampires in any way.</span>")
```
Hopekz: Won't this just make new vampires able to "grief" their starting vampire?
If someone is still mad at the vampire that gave them vampire powers they can just go after their creators since they'll most likely be released after they get converted.

With this flavor text, it now becomes an option IMO.

boodaliboo:
i dont like username

yogstation/code/datums/antagonists/vampire.dm
Darkstick: I don't want to turn this ability into a thrall maker, since that would just turn the gamemode into another conversion antag. I could give the victim a long sleep stun after they get converted, to give the starting vampire time to get away from them in case they get violent?

Hopekz: Eh not sure honestly if you go the "sleep" route the converted guy can just "go after" the starting vampire after their stun. 

Regardless a stunning sleep or a stun might be nice because it will give the converted player time to read the vampire flavor text in the first place (remember that a lot of people being converted probably have no idea how to be a vampire)

Darkstick:
oh right, my last commit also lowers the price of Lilith's pact to 300 blood, and it unlocks at 320 instead of 400 and 450. Forgot to mention that.

SomeguyManperson:
edit that stuff into the changelog then

PR #7988, opened 2020-03-15
Computers will now make ambient sounds

yacabo:
Computers will now make ambient sounds with a minimum of about five seconds between sounds and a 50% chance to play it every second from then on it does not play. Currently there are ten sounds that will play mostly taken from Valve games including Half Life, Half Life 2, TF2, and CS:GO. 

I tried to find a balance to make them play with enough time between them to make it so that they don't wear on you but I fell in love with the fact that the bridge sounds much different from the rest of the station so speak up if tweaking with the timing needs to be done.

:cl:  
soundadd: Computers will now make ambient sounds
/:cl:


yacabo:
Yeah it ain't ready yet

yacabo:
Attempt 2
Things are quieter and you can't hear things behind walls

Bibby0110:
Could you maybe let us hear the sounds?

yacabo:
GitHub doesn't support audio files. So here's a discord link an audio file

https://cdn.discordapp.com/attachments/205784753352343552/689560567777722369/demo.ogg

Audio recording of standing in the bridge of delta, half way through I walk around so you can compare how loud it is with a sound you know

Hopekz:
I personally don't like how "long" some of these audio files are.
If you walk away from a console the sound keeps playing and it kind of ruins the illusion you are going for. Maybe there is a way to stop a sound from playing when a certain trigger is met? (can't see it anymore)

Also; these sounds playing a full volume not relative to your position also breaks this immersion a little bit because even if a computer is grazing your view on the corner of your screen this plays at the full volume just like standing next to it.

yacabo:
For the sound still playing issue I thought it was impossible to fix but:
image
It just requires a boat load of work which I can look into

For the position thing, it actually is having a falloff on sound based off your position but I guess I need to increase that a bit, which I can easily do.

Bibby0110:
If you want to keep the actual audio, split it up into X amount of clips, and keep track of what clip to play next to the people near.

Even better. Cut them down to X amount of individual clips, and just pick one randomly each time

yacabo:
Probably going to go for the first one; divide them up into one second or half second long parts and just play them in a row

alexkar598 (changes requested):
This is pretty much really inappropriate as what would essentially be a global sound since there are computers just about everywhere, this really makes sounds everywhere except maybe civilian areas like botany and kitchen I guess. This would fit better for perhaps the RnD servers or another AI area sound

yacabo:
Further work must be done. I see no point in keeping this up in the time being

PR #7998, opened 2020-03-20
holoparasite base healing ability is no longer selectable

SomeguyManperson:
still possible to roll through randomly generated guardians just not directly selectable since it's the only thing anyone will ever use

:cl:  
rscdel: healing ability removed from build a guardian
/:cl:

### DIFF
--- a/yogstation/code/modules/guardian/guardianbuilder.dm
+++ b/yogstation/code/modules/guardian/guardianbuilder.dm
@@ -60,7 +60,7 @@
 	.["no_ability"] = (!saved_stats.ability || !istype(saved_stats.ability))
 	.["melee"] = !saved_stats.ranged
 	.["abilities_major"] = list()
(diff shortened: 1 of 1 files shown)


monster860 (changes requested):
add it to ui_act as well otherwise it will be bypassable via href exploit

Readystorm546:
Im fine with this but can ya fix the healing too?

Hopekz:
For everyone confused and yelling "improve don't remove" this isn't being removed.
It is now just not select-able so you can still get this as a random ability from a stand arrow.

Healing holopara has been the meta for almost 3 months now so it should either be not select-able like in this PR or heavily nerfed.

boodaliboo:
I’m going to downvote it because I think it should stay but make it so it can only heal others

Hopekz:
> I’m going to downvote it because I think it should stay but make it so it can only heal others

Bood limited healing exists and isn't being removed from selection to the best of my knowledge.
This is the "able to heal master" one.

PR #7999, opened 2020-03-20
healing holo has forced healing and limited healing can't heal other holos with the same user

SomeguyManperson:
:cl:  
fix: healing parasites can heal robot limbs
fix: limited healing parasites can't heal their user through proxy
/:cl:


Readystorm546:
THANKS BBY

Hopekz:
Why would a healing holopara be able to heal preternis/robot limbs when it can't heal objects for example consoles/doors/windows or whatever.

It makes no sense this should be on a new ability that lets them heal objects in my opinion.

It almost feels like letting players heal robotic limbs through fleshmend.

Marmio64:
please give warops more TC because they already lose all the time even with healopara

PR #8003, opened 2020-03-21
Removes useless variable in the changeling file

Hopekz:
competitive_objectives literally doesn't do anything except set itself to true later in the file.


boodaliboo:
im,rporve remove

PR #8004, opened 2020-03-21
Reagent viscosity framework

Hopekz:
**What this does:**
------
Reagents now have viscosity which affects the speed at which reagents are injected via syringes.

As proof of concept, water is affected by this as I have given it a very low viscosity so you can inject people with water quicker.
**(currently all reagents besides water inject at normal speeds)**

At the moment I am only changing the viscosity on the water as other changes will be done in separate PR's because in my opinion injecting certain reagents quicker or slower should be discussed separately.

**At the moment the viscosity of injection is determined by the highest viscosity in the syringe**. (meaning that you can't do something like mix water and potassium to quickly inject water + potassium)

Originally this took the average viscosity of all reagents in the syringe but I had to take that portion out for the time being as it made injecting lethal reagents OP when mixed with water because at the moment harmful reagents do not have a higher viscosity.

In the future after this framework is merged I plan to make all the reagents in the syringe "average" out so that you can mix and match viscosities. For example, you would be able to mix water with kelotane to make your kelotane syringe inject quicker just using the water as a medium.

why:
-----
Non-harmful "less viscous" reagents can be injected quicker and we can keep the other more harmful reagents more "vicious" meaning they inject slower (currently all reagents besides water inject at normal speeds)

This framework allows reagents to be balanced by their injection speed and in the future will allow more chemistry gameplay in terms of mixing and matching reagents if you want to inject your reagents at an optimal speed.

#### Changelog

:cl:  
rscadd: Reagents now have viscosity which affects the speed at which reagents are injected via syringes.
rscadd: Water was a low viscosity so it injects rapidly.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents.dm
+++ b/code/modules/reagents/chemistry/reagents.dm
@@ -40,7 +40,6 @@ GLOBAL_LIST_INIT(name2reagent, build_name2reagent())
 	var/self_consuming = FALSE
 	var/reagent_weight = 1 //affects how far it travels when sprayed
 	var/metabolizing = FALSE
-	var/viscosity = 1 // affects how rapidly it can be injected via syringe (lower = faster; formula is viscosity * 30 deciseconds
 
 /datum/reagent/Destroy() // This should only be called by the holder, so it's already handled clearing its references
 	. = ..()
--- a/code/modules/reagents/chemistry/reagents.dm
+++ b/code/modules/reagents/chemistry/reagents.dm
@@ -40,6 +40,7 @@ GLOBAL_LIST_INIT(name2reagent, build_name2reagent())
 	var/self_consuming = FALSE
 	var/reagent_weight = 1 //affects how far it travels when sprayed
 	var/metabolizing = FALSE
+	var/viscosity = 1 // affects how rapidly it can be injected via syringe (lower = faster; formula is viscosity * 30 deciseconds
 
 /datum/reagent/Destroy() // This should only be called by the holder, so it's already handled clearing its references
 	. = ..()
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -116,6 +116,7 @@
 	glass_name = "glass of water"
 	glass_desc = "The father of all refreshments."
 	shot_glass_icon_state = "shotglassclear"
+	viscosity = 0.25
 
 /*
  *	Water reaction to turf
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -116,7 +116,6 @@
 	glass_name = "glass of water"
 	glass_desc = "The father of all refreshments."
 	shot_glass_icon_state = "shotglassclear"
-	viscosity = 0.25
 
 /*
  *	Water reaction to turf
--- a/code/modules/reagents/reagent_containers/syringes.dm
+++ b/code/modules/reagents/reagent_containers/syringes.dm
@@ -129,9 +129,14 @@
 				if(!L.can_inject(user, TRUE))
 					return
 				if(L != user)
+					var/viscosity_highest = 0
+					for(var/datum/reagent/R in reagents.reagent_list) // takes the highest viscosity of all reagents in syringe - Hopek
+						if(R.viscosity > viscosity_highest)
+							viscosity_highest = R.viscosity
+
 					L.visible_message("<span class='danger'>[user] is trying to inject [L]!</span>", \
 											"<span class='userdanger'>[user] is trying to inject [L]!</span>")
-					if(!do_mob(user, L, extra_checks=CALLBACK(L, /mob/living/proc/can_inject, user, TRUE)))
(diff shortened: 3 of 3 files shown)


athena413:
very good way of making chemist less powergame without gimping the entire fucking job

boodaliboo:
add pacman

alexkar598:
this is completly fucking stupid, realism says that mixing a very viscous thing with a very non viscous thing should do something in the middle but putting that aside

this has many balance implications:
1) You are using realism to balance reagents, life isnt balanced, you have chemicals so poisonous that touching it will kill you

2) How do you decide which chemical get a high viscosity and which dont, at some point or an another, its going to become a bunch of meaningless arbitary values

3) Why the living fuck do syringes have anything to do with viscosity??? If you press on a liquid on one side, the only thing that decides if its going to move is its mass, a liquid cant decide that hmmm no i wont move quickly

This really feels like lets fix a problem that doesnt exists by implementing an uneeded solution that doesnt even make sense

Hopekz:
>realism says that mixing a very viscous thing with a very non viscous thing should do something in the middle but putting that aside

Like I said with my PR description: This is how this system initially worked but because only water is changed by this framework I changed it to be "higher viscosity goes" due to balance reasons.

This can be changed back later once we get all the viscosities in place but right now it would allow any reagent to be injected quicker without such down-side.

> 1. You are using realism to balance reagents, life isnt balanced, you have chemicals so poisonous that touching it will kill you

I am not using "realism" to balance. 
"Viscosity" is just a word that I used for this and in my opinion, it fits. 
**I am not using realistic values to balance items and do not plan on using real-life** to get this information.

> 2. How do you decide which chemical get a high viscosity and which dont, at some point or an another, its going to become a bunch of meaningless arbitary values

Good question!
Since balancing is controversial I was going to do different reagent groups separately in separate PR's.
**Here is what I was thinking in regards to balancing:**
If a reagent can be harmful (OD'able, harmful effects etc) I would make the viscosity higher. The less "harmful" a reagent is the viscosity it will have (as I said in a separate PR so if you disagree with this you can always give your opinion on that one when it exists)

> 3. Why the living fuck do syringes have anything to do with viscosity??? If you press on a liquid on one side, the only thing that decides if its going to move is its mass, a liquid cant decide that hmmm no i wont move quickly

If I use the "mass" value in the game it will make them random values because those values were not made with this system in mind.
I want a separate number for balancing reasons as currently the "mass" of liquid only controls how far it is sprayable in a spray bottle.

It is important to keep that value separate because they do two different things so for example, a reagent will be able to be balanced by having an untouched injection speed (3 seconds) but not fly through the air well (high mass)

> This really feels like lets fix a problem that doesnt exists by implementing an uneeded solution that doesnt even make sense

Hopefully, it makes more sense from my comments. 
Unsure if I made my description clear enough lmk if I need to update something to make it more clear or if I need to add a TL;DR for people that won't bother to read it.

Also, in my opinion, the best features are solving problems that no-one knows they have. 


**TL;DR:
It seems like we're having a miscommunication**

ghost:
alex's criticisms are as stupid as this PR

a chemical is balanced by its difficulty of creation and its effects. how long it takes to inject in a syringe is irrelevant.

people do not inject people with syringes to kill them. if they have a deathmix, they apply it through syringe gun.

why wouldn't they just entirely avoid your viscosity thing with a patch, anyway? just make a patch of the chem and label it as 'brute patch' or 'saline-glucose (10u)' and apply it via that instead of syringe. 

if someone's holding still long enough for you to inject them with a syringe, you can do it with a patch instead. patches hold more and go through hardsuits.

this PR is just...kinda useless.

Hopekz:
@AdamElTablawy 

You're over-reacting.
**All reagents remain functionally the same after this PR.** This only gives people the option to also balance reagents by injection speed if they so choose.

We can have the same argument for other reagent variables such as mass for spray distance.
Spray distance is not "game-changing" and doesn't make a reagent better than patches but the additional layer of complexity is appreciated regardless.

The goal of this PR isn't to make injection better than other means of application not that this changes the rate of any reagent except water.

ghost:
i'm not over-reacting i'm just saying it's entirely useless and does nothing but bloat code

Hopekz:
> i'm not over-reacting i'm just saying it's entirely useless and does nothing but bloat code

All hail Adam guardian of the codebase.
Almost got away with it too my 7 lines of code would have bloated everything

ghost:
> > i'm not over-reacting i'm just saying it's entirely useless and does nothing but bloat code
> 
> All hail Adam guardian of the codebase.
> Almost got away with it too my 7 lines of code would have bloated everything

oversensitive, are we?

it's a useless pr. i explained why. nothing changes.

CMOisLing:
This is cool but I don't think it will do much to slow down chemists. They have vapor, patches, food, or just shoving it down your throat.

Hopekz:
> This is cool but I don't think it will do much to slow down chemists. They have vapor, patches, food, or just shoving it down your throat.

The first thing I thought of was when the viscosity values are assigned properly you can get a general idea of what you're being injected with just by judging the duration of the injection.

When it happens "quickly" you can deem whatever was injected to be more or less harmless and when it takes a while you know it is something with a higher viscosity.

Once more values are implemented I'd experiment with bringing the "all reagents average out" feature so that a chemist might mix their reagent with something less viscous so that it gives this "harmless" illusion.

alexkar598:
Going to close this mainly because theres little actual change here other than buffing overpowered things and nerfing underpwoered things

PR #8010, opened 2020-03-23
Round start vampires wont get a blood objective, only a convert one

Darkstick:
Since convert objective requires 300-900 blood, having the blood objective alongside it was kind of redundant.

#### Changelog

:cl:  
tweak: round start vampires only get a convert objective and a traitor objective
tweak: renamed the "full_objective" variable to "full_vampire" to avoid confusion
/:cl:

### DIFF
--- a/yogstation/code/datums/antagonists/vampire.dm
+++ b/yogstation/code/datums/antagonists/vampire.dm
@@ -11,7 +11,7 @@
 	var/total_blood = 0
 	var/converted = 0
 	var/fullpower = FALSE
-	var/full_objectives = TRUE
+	var/full_vampire = TRUE
 	var/draining
 	var/list/objectives_given = list()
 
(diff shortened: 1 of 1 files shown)


Hopekz:
How about making it a 50/50 chance of being one or the other instead? 
In my opinion, it's nice to have variety instead of the same objective over and over.

Should work fine with a prob:

```
if(full_vampire)
	if(prob(50))
		for(var/i = 1, i < CONFIG_GET(number/traitor_objectives_amount), i++)
			forge_single_objective()
		var/datum/objective/convert/convert_objective = new
		convert_objective.owner = owner
		convert_objective.gen_amount_goal()
		add_objective(convert_objective)
	else
		var/datum/objective/blood/blood_objective = new
		blood_objective.owner = owner
		blood_objective.gen_amount_goal()
		add_objective(blood_objective)
```

Hopekz (changes requested):
Look at my comment above

Hopekz (approved):

PR #8012, opened 2020-03-23
Revert "Cycling Airlocks"

boodaliboo:
Reverts yogstation13/Yogstation#7874
these are fucking awful they take way to long


Marmio64:
thank you, you are doing gods work

boodaliboo:
fuck you map files

Bibby0110:
Only remove them from the map.
Don't remove the actual object

Amelia0010 (changes requested):
If the problem is that they take too long, speed them up.

boodaliboo:
fine

alexkar598:
The reason for them taking too long is that atmos and machine can't tick fast enough

PR #8014, opened 2020-03-24
Faster Cyclelocks

MayhemSailor:
### Installs a set of cycle locks in telecommunications to test if they are faster under game conditions

Utilizing Minsky style functional airlock design this attempt to speed up the cycle for external airlocks. 

For the time being, only the two on satellite external access for telecommunications and the Engineering Escape Pod external access has been modified. If it works, we can modify the rest of the cycling airlocks. 

Old:
CycleLock Old

New:
CycleLock New

#### Changelog

:cl:  
experimental: Increase Speed of Cyclical Airlocks.
/:cl:


monster860:
Have you actually tested this?

MayhemSailor:
> 
> 
> Have you actually tested this?

Yes, They work faster on a locally hosted server with just me on it, over the course of almost an hour.

Hopekz:
Was this addressed already? 
image

MayhemSailor:
@Hopekz in the ScrubbersFTW commit is was.

alexkar598:
the issue with airlocks is not really their speed, its the lag that compeltly deadlocks atmos into tickikg once per 10 seconds

Hopekz:
Why do we need to simulate airlock pressure while it changes?
We know the start and and goal pressures why don't we just make a timer then set the pressure in the airlock tiles to either the outside or inside?

While simulating it might be "nice" the goal of these is to change the pressure from A to B so this part can be simplified to cut out the lag.

Making it simplified will also give you a consistent result each time as the "wait" will be based on a timer rather than it actually processing the change.

JamieD1:
Can we not just buff the airlock vents to be like Supercharged?

MayhemSailor:
Doesn't appear to be needed. Monstermos issue.

PR #8016, opened 2020-03-24
Redd's Gas Station Update

ReddicusDragon:
Literally just improves the gas station. Adds a welcome area for people deciding to teleport in, making sure they don't die.

:cl:  
tweak: improves lavaland gas station
/:cl:


Ktlwjec1 (changes requested):
One thing and then this is good to merge (copied my image across from before so ignore most of the circles):
image

- Move everything down one tile like the image on the right to prevent lava spawning just outside the airlock

ReddicusDragon:
@Ktlwjec1 i did do that already, check map diffs

ReddicusDragon:
@Ktlwjec1 okay now it's done I'm a damned idiot

PR #8021, opened 2020-03-26
EVA maintenance bolted round-start

Hopekz:
**What this does:**
------------
**EVA maintenance is now bolted round-start.**
Added buttons to both sides to operate this bolt if you have access to the door.

With this change, the number of hacks to get into EVA from maint goes up to 2 from 1 to more or less equalize it with the number of hacks required to normally access EVA through the front or right side.

Capture_fixed

Why:
------

**See image for the number of hacks required to access EVA.**

This entry is overpowered in my opinion as it takes a single hack to reach not only EVA but the RPD right next to the door.

eva_maint_new_hacks

All other options are inferior in terms of accessing EVA.

**With these changes it will at least take 2 different hacks so knowledge of the bolt wire will be needed.**

#### Changelog

:cl:  
rscadd: EVA maintenance door bolted round start. Added buttons to operate bolt.  
/:cl:


alexkar598:
I don't think its a good idea to preoptimize the station like that. Id much rather move the airlock down in the EVA material storage

Hopekz:
> I don't think its a good idea to preoptimize the station like that. Id much rather move the airlock down in the EVA material storage

This PR is mainly trying to address the fact that this back door is 1 hack away from EVA entry while the other quickest more high profile entries are 2+ hacks away from entry.

overlord12345:
You know someone can just smash the button and take the access restriction out

Hopekz:
> You know someone can just smash the button and take the access restriction out

Yes. That is an acceptable outcome in my opinion.
The method you mentioned is a 3 step process "hack" that encourages fun gameplay.

In your method:
1. Smash button
2. remove access restriction and put it back
3. use button to unbolt then hack the door

JamieD1:
Why not add shielding to the wires instead?

Also maybe moving the maintenance to below the other door, maybe this is a good opportunity to redesign Eva, I might give it a crack later.

Takahiru:
you forget that with no maintenance access it takes 2 hacks to enter EVA

Hopekz:
> Why not add shielding to the wires instead?

In my opinion, it takes too long to weld past shielding. This strikes the right kind of "difficulty" that I am looking for. Putting shielding here would be too much.

> you forget that with no maintenance access it takes 2 hacks to enter EVA

I did not forget.
The reason I did not factor it into my explanation is that I still deem this route to be acceptable in difficulty with 3 hacks instead of 2 as well because of the advantageous "stealthy" position this entry gives.

If you choose to hack into maint it somewhat equalizes with the right side of EVA which is a "higher profile" hack as well. 
The right side forces you to hack while trespassing.

**To summarize the requirements for each way after these changes:**
**Front:** 
High profile, knowledge of 1 wire to cut. 
Easy in and out but risky because of the high profile

**Right side:** 
High Profile, knowledge of 1 wire. 
Knowledge of bolt wire if you want to get out before the right side closes. 
Gives you cover on the 2nd hack before starting to major trespass.

**Left maint side:** 
Low profile, knowledge of 2 wires + multitool to pulse bolt wire if hacking the door
or 
knowledge of 1 door wire if "hacking" the button itself per overlord's example.

Takahiru:
While i find this an unnecessary change, I really like the huge detailed balance reasonings you give for this so im for this

ghost:
even the gravity generator doesn't start bolted

ask the AI to bolt instead

Ktlwjec1:
Yeah this shouldn't be bolted roundstart, only super secure areas such as the AI core should be. Consider a redesign of EVA if you don't like the maint door.

PR #8022, opened 2020-03-26
2 new mapping tools + Better maintenance (revamp)

Hopekz:
**Adds 2 new assists for mappers.**

A new RNG wall and maintenance floors. 
The walls have a chance to not spawn, spawn rusted etc.
The floors roll for spawning broken or burned + roll for a chance for having cleanable junk on them.

**What this can be used for:**
These tools can be used for any place where you need wall variation and scummy tiles.
Think space ruins, lavaland ruins etc.

Because of how much RNG there is to them there are lots of variations.

Example of how this works with input and output:
------
**INPUT:**
before2

**OUTPUT:**
after2

I then used this to revamp the entirety of box maintenance
-----

Examples of a box variation:
------
**To note:**
New paths being created etc, rust and grimy junk scattered around.
I didn't think this would have this much of an impact but It really feels like a new experience every time it is loaded.

after2_1
after2_2
after2_3

#### Changelog

:cl:  
rscadd: Added two new tools for mappers to create RNG walls and maintenance floors
rscadd: Used the new RNG tools to revamp box maintenance. Maintenance now has lots of different variations!
/:cl:


JamieD1:
Maybe instead of not there all together, it spawns a girder instead? Or maybe that could be the station variant.

Just an opinion

Readystorm546:
As long as were not putting it on walls that are open to space

Xoxeyos:
Speed merge this please, I would also like the possibility for both types of girders to be added, along with ruined wooden floors.

alexkar598 (changes requested):
The wall maint spawner will need to be a special turf that uses the proper ChangeTurf system instead of a clunky new /turf/open/plating/whatever

As for the maint tile, it will need to be seperated into another special turf for the damaged state and for the cleanable decal that will need to be a /obj/effect/spawner/lootdrop

alexkar598:
This PR will be closed in 24 hours due to being in a Work in Progress state for too long (2 weeks+). It will be reopened on request on completion. This is an effort to keep the PR queue clean.

PR #8027, opened 2020-03-26
Solar panel cleanup + full setup change

Hopekz:
Changes 2 things::
-----

**1:**
-----
For some reason, there was a WIP looking tile named "floor" under the solar panels.
Cleans up the solar debug looking floor tiles and replaced them with catwalks instead.
The debug "floor" tile doesn't seem to be craft-able in-game.

**before:**

**after:**

**2:**
------

Engineering currently starts with 13 solar assemblies in engineering. 
Currently, these never get used.

Created a new mapping assist that allows solar panels to be missing round-start.

Used this new assist to make consistently around 7-10 out of 240 solar panels around the station be missing round-start. 
If a full setup is wanted engineering will now have to place these panels from the extras in engineering.

Capture3

#### Changelog

:cl:  
tweak: Replaced weird debug looking solar "floor" under solar panels with a catwalk.
rsadd: Made an average around 7-10 out of the 240 solar panels on box station not be set up. Engineering now has to set these up from the 13 spare assemblies in engineering.
/:cl:


alexkar598:
Many tiles aren't craftable in game including the circuit tiles and actually every single tile as theres no tile painter

JamieD1:
To be honest I dont think you have gone far enough, like everyone will see the two missing panels and be like, meh I can do without.

I was thinking like one or two of the arrays are ALWAYS missing half the panels, maybe with a crate with panels in if needed next to the smes so they have to do things?

Just an idea.

MegaEmpirical:
Yeah, need it harder. Be more like Charlie station and just add a fee extra crates, maybe shove them in the solar power rooms.

Bibby0110:
This sounds incredibly unfun if all the crates are in Engineering
Also imo it looks better with the tiles underneath, or at least just regular plating, and not catwalk

Hopekz:
> Also imo it looks better with the tiles underneath, or at least just regular plating, and not catwalk

I'm curious as to if anyone else feels this so I'm going to be making a strawpoll soon for this with both variations. Stay tuned!

Ktlwjec1:
Yeah looks much better with the tiles, fully agree with gamer here. I don't mind the panels being missing though, but it shouldn't be too many like some people are suggesting. The number missing currently in this PR seems fine but it can be adjusted easily if needed in the future.

Hopekz:
EDIT:
added the debug looking tiles back.

The PR was mainly to do the solar round start building variation the weird tiles being replaced by something else can happen in a different PR.

image

alexkar598 (approved):

PR #8032, opened 2020-03-27
Revert "Revert "Label workflow""

Amelia0010:
Reverts yogstation13/Yogstation#7203

Gonna need @alexkar598 or @monster860 to re-setup the github token since mine probably doesn't work anymore


alexkar598:
We can't use the label one as the github token doesn't have write access on the repo if its originated from a fork :/

PR #8035, opened 2020-03-27
Swedish speech mutation now replaces 'the [X]' with '[X]en'/'[X]et'

Amelia0010:

Continuing the number 1 danish pastime, bullying swedes

#### Changelog

:cl:  
tweak: Swede-speak mutation now replaces 'the [X]' with '[X]en'/'[X]et'
/:cl:


code/datums/mutations/speech.dm
```diff
@@ -161,6 +161,7 @@
 		message = replacetext(message,"a",pick("å","ä","a"))
 		message = replacetext(message,"bo","bjo")
 		message = replacetext(message,"o",pick("ö","o"))
+		message = replacetext(message, regex(@"([^\w]?)the (([^\W]|å|ä|ö)*)", "gmi"), pick("$1$2en", "$1$2et"))
```
Amelia0010, line 164: The å|ä|ö is disgusting, but byond doesn't support matching against full unicode (even on 513), so it interprets them as "non-word characters", hence the need for me to check explicitly.

The check at the start of the regex is to make sure we don't fuck up [word-ending-with-the] [word].

PR #8040, opened 2020-03-27
[PORT] Venus Human Trap Update

swissloaf:

"Fixes the currently broken implementation of the venus human trap, a mob spawned by the flowering mutation of kudzu. They're supposed to attach vines to you and reel you in (I think), but that functionality was entirely busted, only meleeing you with their vines. Along with this, sentient venus human traps could attack you as fast as they could click, pretty much killing anyone instantly. So, I completely wiped them and started from scratch, ending up with the following:

Venus Human Trap

Can attach vines to their targets within a 5-tile radius.
Vines stun carbons for 2 seconds when first applied, and will attempt to bring the affected closer to the plant. If the target gets far enough away from the plant or shares the same tile as the plant, the vine will be destroyed.
If a plant tries to attach a vine to something they already have a vine on, they will reel everything they can towards themselves. This also happens passively every second.
Plants heal 10% of their max health when attacking with melee if their target is alive.
Sentient plants can attach their vines to a wide variety of things, allowing them to take advantage of the damage dealt when walking though them.
Plants can be possessed by ghosts akin to spiders born from a sentient mother.
Along with all that, I also autodoc'd the whole file as well, as well as fixing the apparently broken feature of kudzu not affecting plant-based creatures (venus human traps and podpeople)."

#### Changelog

:cl:  
rscadd: Adds Update for the Venus Human Trap
/:cl:


boodaliboo:
travis failed :(

swissloaf:
> travis failed :(

now its good

swissloaf:
why does turdis fail!!!!!!!!!

PR #8047, opened 2020-03-28
Toilets now require buckets instead of mopbuckets

### LINKED ISSUE #8020: Issue with Crafting Toilets
## Reproduction: I had attempted to create a toilet as seen in the above photo. Though all materials are present, I am unable to craft a toilet. I've attempted this with all materials on the ground and an empty mop bucket, all materials on hand and a filled mop bucket, and I have not been able to craft it.
aaaa

alexkar598:
Mop buckets are kinda hard to get and people seem to be confused that its the big yellow one not the blue bucket

Fixes #8020


Hopekz (approved):
In hindsight, I should have done this in the original PR.

PR #7933, opened 2020-03-07
Adds 5 placement variations to morgue

Hopekz:
**Morgue spice!**
All feedback welcome.

dreamseeker_lagtmTpDYV

Variations:
-------

#### Changelog

:cl:  Hopek
rscadd: Added 5 placement variations to morgue
/:cl:

PR #8053, opened 2020-03-28
Code-based Variation Framework!

Hopekz:
**This framework allows for creating variations without the use of map files.**

This offers many advantages such as better map generation performance, less clutter and most importantly being able to use these on any map by placing the new markers instead of creating a map specific variation.
This also looks very clean on the map itself instead of being a blocky template noop rectangle.

Used this new framework to convert #7933 to be code-based.
Also re-created #7928 using this framework.

I made it easy for mappers to slap these new variations on their maps with new landmarks!

**Added the basic cell and perma cell variations to:**
Yogstation, IceBox, YogsDelta, YogsPubby, YogsMeta

**Added the morgue variations to:**
Yogstation, IceBox

**Same variation layouts that can be found on #7928**

#### Changelog

:cl:  
rscadd: New Variation framework!
tweak: converted morgue variation to use this new framework.
rscdel: Deleted the old morgue variation files as they are no longer needed
rscadd: Added the basic cell and Perma cell variations to Yogstation, IceBox, YogsDelta, YogsPubby and YogsMeta.
rscadd: Added the morgue variations to Yogstation and IceBox
/:cl:


Bibby0110:
EDIT. I'm literally retarded
So what you did is hardcode the maps into code, so you can place them anywhere?
Sounds like just a different approach from map templates?


Why is this better than map templates

Hopekz:
> So what you did is hardcode the maps into code, so you can place them anywhere?
> Sounds like just a different approach from map templates?

Exactly!

> Why is this better than map templates

**Besides what I wrote in the OP of:**
"This offers many advantages such as better map generation performance, less clutter and most importantly being able to use these on any map by placing the new markers instead of creating a map specific variation.
This also looks very clean on the map itself instead of being a blocky template noop rectangle."

The main reason is that variations in the old map file-based system take a map file each variation.
for example here is the map files we have without moving them into the code.

**I have pointed out the variations I moved to code in red:**

image

Another huge upside to this is that people can now make variations that were not significant enough to create a whole map file over.

For example in the map-based system in order to create these brig variations I had to add 6 map files into the code-base.

This is mainly an accessibility tool for mappers to create more variations without creating whole map files + easily add existing variations to their maps with the new landmarks.

Another advantage is with the old system it copied over everything even the area so you had to prepare the area and turf for the changes with template noop and template area.
Since this framework just adds the items based on the reference point without copy-pasting the whole tile (just from the list you specify) you no longer have to create these zones.

There are lots of advantages to using this system but both systems have their place.
I plan to update this further in future PR's to allow for more complicated variations in-code such as turf replacements.

Bibby0110:
I'm not sure how I feel about this, especially moving visible changes into code, which means you can't easily see what the different variations are. It might just end up as pointless variations if they're too small, negating the whole advantage of it being possible to add smaller variations

Hopekz:
> I'm not sure how I feel about this,

Why do you "feel" anything about this?
This is a tool for mappers or coders to add more variations on our maps.

At the moment there is no visual interface for it besides landmarks since its "in the code" but I can't see why it cannot be created in future PR's if it even is needed in the first place.

I can add debug code if you want to the world log like logging which variation spawned otherwise I'm not sure what you want me to do about this unless you expect me to create a full visual interface at the moment or something.

Hopekz:
EDIT:
Added logging to variations

image

Hopekz:
EDIT:
Removed the Minsky variation change because Minsky is being killed.

Bibby0110:
> > I'm not sure how I feel about this,
> 
> Why do you "feel" anything about this?
> This is a tool for mappers or coders to add more variations on our maps.
> 
> At the moment there is no visual interface for it besides landmarks since its "in the code" but I can't see why it cannot be created in future PR's if it even is needed in the first place.
> 
> I can add debug code if you want to the world log like logging which variation spawned otherwise I'm not sure what you want me to do about this unless you expect me to create a full visual interface at the moment or something.

I'm not sure you read my reply.

What makes this better than having dedicated map files?
Is the performance boost actually noticeable?
I would prefer a blocky room of noop if it meant I could change anything in a map file using a visual editor, instead of going into the code and editing finnicky snippets of code

This doesn't seem to do anything but make variations less different, harder to add for non-coders, and much harder to visualize and edit?

alexkar598:
Closing this, id rather have it the way I asked on discord

PR #8064, opened 2020-03-29
Nerfs BoH bombs alot, they are only 3 by 3 now and they don't let anything near it live

alexkar598:
Boom boom yall die.

BoH bombs have been ultra nerfed, its only 3x3 now and anything in that range gets ex_act'd with EXPLODE_DEVASTATE

Also demodularizes a proc

### DIFF
--- a/code/datums/components/storage/concrete/bag_of_holding.dm
+++ b/code/datums/components/storage/concrete/bag_of_holding.dm
@@ -23,11 +23,20 @@
 		log_game("[key_name(user)] detonated a bag of holding at [loc_name(loccheck)].")
 
 		user.gib(TRUE, TRUE, TRUE)
-		for(var/turf/T in range(1,loccheck))
+		for(var/turf/T in range(2,loccheck))
 			if(istype(T, /turf/open/space/transit))
 				continue
 			for(var/atom/AT in T)
-				A.ex_act(EXPLODE_DEVASTATE)
+				AT.emp_act(EMP_HEAVY)
+				if(istype(AT, /obj)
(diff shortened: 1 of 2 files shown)


Takahiru:
I support this, BoH bombs right now are big press 2 button griff machines

yacabo:
I say 5x5 so it's not completely useless

Moja10:
I really don't like this at all, even as an admin who has to deal with 5 tickets resulting from mr. multikeyer detonating one on the shuttle.

Xoxeyos:
I'd say 5x5 would be a good radius.

Hopekz:
While I think the range is alright (could be a little more) **I don't like how this is removing a big portion of a game that I find enjoyable by not sucking both you and whoever made it into the tear.**

Sure someone can walk into it later but one of the things that made this a questionable assassination method is that there is a chance your target would live inside the tear. Now it's kind of a 100% gib

Also; better hope your target doesn't walk away from you when that "are you sure you want to BoH bomb" checkbox appears. 
**This radius is small.** 
Your target can easily sidestep this by the time you accept the confirmation because it defaults to "abort"

image

boodaliboo:
5x5

SomeguyManperson:
4x4 would probably be fine, 5x5 seems like a bit too much

Takahiru:
Now that ive thought about it, you could probably manage to just dodge it with that small a radius, 5x5 or 4x4 would be better

alexkar598:
Going to close this PR and make another since this has so many changes it can probably be considered a new PR

PR #8067, opened 2020-03-29
Hierophant club station fix

Alexandria404:

The hierophant club has received complaints and rightfully so.
It is OP when on station due it´s functionality even when compared to a lot of traitor gear, as it can kill people standing behind Rwalls even. 

At the same time miners complained it is good for mining purposes.

Simple fix, it does not work on the station as it is beyond unbalanced there, while it keeps it´s full functionality on lavaland.

Code is tested and works.

#### Changelog

:cl:  

tweak: the hierophant club does not work on the station

/:cl:

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -1151,6 +1151,9 @@
 	var/turf/T = get_turf(target)
 	if(!T || timer > world.time)
 		return
+	if(!SSmapping.level_trait(T.z, ZTRAIT_MINING))
+		to_chat(user, "<span class='warning'>The club fizzles wekaly, it seem his power does not reach to this realm.</span>" )
+		return
 	calculate_anger_mod(user)
 	timer = world.time + CLICK_CD_MELEE //by default, melee attacks only cause melee blasts, and have an accordingly short cooldown
 	if(proximity_flag)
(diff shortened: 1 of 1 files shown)


code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -1151,6 +1151,9 @@
 	var/turf/T = get_turf(target)
 	if(!T || timer > world.time)
 		return
+	if(!SSmapping.level_trait(T.z, ZTRAIT_MINING))
+		to_chat(user, "<span class='warning'>The club fizzles wekaly, it seem his power does not reach to this realm.</span>" )
+		return
```
alexkar598:
```suggestion
	if(!is_mining_level(T.z))
		to_chat(user, "<span class='warning'>The club fizzles weakly, it seem its power doesn't reach this area.</span>" )
		return
```

code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -1203,6 +1206,10 @@
 	update_icon()
 
 /obj/item/hierophant_club/ui_action_click(mob/user, action)
+	var/turf/user_turf = get_turf(user)
+	if(!SSmapping.level_trait(user_turf, ZTRAIT_MINING))
+		to_chat(user, "<span class='warning'>The club fizzles wekaly, it seem his power does not reach to this realm.</span>")
+		return
```
alexkar598: I don't see why this is needed, the teleport ability is a unique one and doesn't seem too overpowered for a big ass club that you can't backpack

alexkar598 (changes requested):

code/modules/mining/lavaland/necropolis_chests.dm
Alexandria404: fully agree, I talked to miners about this afterwards, will be fixed

Marmio64: let people use the clubs teleport function on station

alexkar598 (approved):

PR #8069, opened 2020-03-29
Rebalances BoH bombings (take 2, new PR)

alexkar598:
This rebalances BoH bombings by both nerfing and buffing them

Made a new PR for polishing the PR body and because this is very different from the old one in terms of changes

Buffs:

- Everything in range are emp'd
- All objects in range are marked as being broken
- The user is no longer gibbed
- All mobs in range get 85 brute damage applied to them

Nerfs:

- Its only a 5x5 instead of a 13x13

This PR has several goals:

- Make BoH bombs valid for people that don't have hijacks
- Restrict the amount of instant this department hasn't been found 404 grief
- Make it a viable method of assasination
- Prevent moving machinery that shouldn't be moved
- Remove the instant you are no longer part of this round because i clicked two buttons behind a wall 10 tiles away

# I DONT KNOW WHERE YOU GOT THIS IDEA BUT THIS DOESNT REMOVE THE TEAR

### DIFF
--- a/code/datums/components/storage/concrete/bag_of_holding.dm
+++ b/code/datums/components/storage/concrete/bag_of_holding.dm
@@ -27,7 +27,7 @@
 				continue
 			for(var/atom/AT in T)
 				AT.emp_act(EMP_HEAVY)
-				if(istype(AT, /obj)
+				if(istype(AT, /obj))
 					var/obj/O = AT
 					O.obj_break()
 				if(istype(AT, /mob/living))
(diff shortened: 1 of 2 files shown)


Bibby0110:
Does this remove the tear?

swissloaf:
"Make BoH bombs valid for people that don't have hijacks" are you sure this wont need a council vote

alexkar598:
Its paving the way for it. I can't change rules with the codebase but making it not an instant department is not here anymore lmao is a good way to go in that direction

Hopekz (approved):

MegaEmpirical:
Will this break gang dominators? If not, could you add that in?

Had a round where a gang dominator was BoHed and it kept ticking out in space.

alexkar598:
Yes, the dom stops when its considered broken

PR #8075, opened 2020-03-30
Borg security module now takes uranium instead of diamonds

alexkar598:
Its quite rare to see them and the main purpose of giving them a high material cost was to make it only viable late game. Thing is that the advanced weaponry tech node is quite hard to research and takes alot of time. This PR makes it still take time so its only a lategame thing while making the cost low enough so its actually used.

### DIFF
--- a/code/modules/research/designs/mechfabricator_designs.dm
+++ b/code/modules/research/designs/mechfabricator_designs.dm
@@ -675,7 +675,7 @@
 	id = "borg_transform_security"
 	build_type = MECHFAB
 	build_path = /obj/item/borg/upgrade/transform/security
-	materials = list(MAT_METAL = 15000, MAT_GLASS = 15000, MAT_DIAMOND = 3000)
+	materials = list(MAT_METAL = 15000, MAT_GLASS = 15000, MAT_URANIUM = 3000)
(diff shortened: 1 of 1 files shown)


Marmio64:
i see sec borgs enough, there is no need to make the module cheaper.

MonsWTF:
If someone decides to do Toxins at roundstart and gets 40-50k Points, Adv. Weapons can be researched by like 20 minutes (17 minutes in-round time).

Honestly the Techweb is more gimmicky than anything else, tech is sort of webbed all over the place but it's still not really much different from the old R&D researching system besides less button-clicks and specific items. It "just happens".

You know exactly where the research is for the tech node of Bag of Holding, but you can't start producing it until you've figured out how to get the Janitor an advanced mop. (???)

alexkar598:
Yes they could research adv. weaponery 20 minutes in, but you are just saying ***NO*** to anything past T2 at that point

JamieD1:
Sec borgs are NEVER seen on low pop as the staff is not there for mats or science research.

athena413:
alright

HomocidlePancakes:
You people are fucking stupid.

PR #8078, opened 2020-03-30
Fixes linux support for lavaland

alexkar598:
The folder is supposed to be named mining not Mining, this isnt an issue on windows because case doesnt matter but it is on linux as its case sensitive


ThatLing:
sucks to be linux

ThatLing:
@alexkar598 delete your fucking branches after instantly self-merging them you *****

PR #8080, opened 2020-03-30
Theos Yells at me so i bend to his wim (Clockcult)

Readystorm546:
Title ^

:cl:   
bugfix: Fixed the floor tile sof the clock map to be the proper tiles that will teliport you back if you  
/:cl:


Jarod1200:
imagine thinking ratvar and the servants therof have places they cant go

alexkar598:
I swear the cunts downvoting a fix

Hopekz:
> I swear the cunts downvoting a fix

Welcome to yogs

SomeguyManperson:
it was funny like once but if it's possible people will find a way to do it regularly and run the joke out since the outer area is supposed to be relatively safe for crew

Hopekz (approved):
Yeah no this shouldn't even be debated.
This is fixing an oversight/exploit.

Merge please.

PR #8096, opened 2020-04-01
Removes april

alexkar598:


boodaliboo:
removes may

boodaliboo:
april fools isnt over

PR #8108, opened 2020-04-03
Craft locking closets

Takahiru:
Allows crafting of personal closets, they take 5 metal 10 cable and airlock electronics.


Takahiru:
merge my pr

Hopekz (approved):

athena413:
good

alexkar598:
Id rather have a general way to create secure closets but this will do

PR #8151, opened 2020-04-11
Adds the Citadel bar as a bar template

monster860:
It's a pretty decent bar

image

:cl: monster860
add: Adds the bar from Citadel as a bar template that can spawn sometimes
/:cl:

### DIFF
--- a/_maps/yogstation/RandomRuins/StationRuins/BoxStation/bar_box.dmm
+++ b/_maps/yogstation/RandomRuins/StationRuins/BoxStation/bar_box.dmm
@@ -164,8 +164,8 @@
 /area/crew_quarters/theatre)
 "av" = (
 /obj/machinery/atmospherics/components/unary/vent_scrubber/on/layer3{
-	icon_state = "scrub_map_on-3";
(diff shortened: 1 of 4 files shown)


ynot01:
erp is bad

ynot01:
no erp

athena413:
erp

Ktlwjec1 (changes requested):
It's a nice bar, a few issues though:
- Copy across the dinnerware vendor across from the other bar maps, since the one you used has mustard which doesn't exist here, and also has different contraband which is map edited in.
- 2 sofas need their left and right sides switched.
- Firelocks need changing from the full tile ones.
- The theatre airlock tile should be the theatre area, not bar, just to keep the standard across all the bar maps.
- The north bar camera has no name (c_tag).
- I'd change the flooring under the theatre + hallways airlocks to normal plasteel (instead of dark). Kitchen airlock flooring should be changed to the kitchen floor instead of the bar floor.
- Add another arcade machine above the current one.

monster860:
did the things

PR #8156, opened 2020-04-12
Turdis no longer relies on simplylinux.ch to work

alexkar598:
:)

To everyone with a PR, you need to update your branch once this is merged to grab the new changes or otherwise you will have constantly failing builds


athena413:
yes

PR #8162, opened 2020-04-12
Adds plasma cutter to Miner Cyborg

JamieD1:
Add Advanced Plasma Cutter to cybords as an upgrade,

Same cost as normal advanced plasma cutter,

Tested but might need tweaks in future if its op

#### Changelog

:cl:  
rscadd: Added plasma cutter to mining cyborg

### DIFF
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -620,6 +620,33 @@
 		for(var/obj/item/storage/part_replacer/cyborg/RPED in R.module.modules)
 			R.module.remove_module(RPED, TRUE)
 
+/obj/item/borg/upgrade/plasmacutter
+	name = "mining cyborg plasma cutter"
+	desc = "plasma cutter module for the mining cybord."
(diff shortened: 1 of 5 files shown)


HomocidlePancakes:
It has been discussed so many times that I’ve lost count. Borgs, in general, are very powerful. They can perform simpler tasks much more effective than humans (engiborgs perform actions faster, janiborgs are very robust at cleaning etc.), but as a counterweight, they can’t perform more advanced tasks and have to rely on humans for maintenance. 

Mining borgs already have access to a KA, which can be upgraded with AOE, and that is enough. It’s immune to storms, can’t be stunned/slowed by fauna and can cross lava after an upgrade. Do not push it further.

athena413:
maybe

Bibby0110:
I mirror what Pancakes said

Xoxeyos:
Make it an upgrade!

JamieD1:
> Make it an upgrade!

It is an upgrade, 

image

alexkar598 (approved):

PR #8166, opened 2020-04-12
Adds a debug hierophant staff (works on station), and a variable for the hierophant z level checking.

Takahiru:
yes

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -1372,7 +1372,7 @@
 		var/obj/effect/temp_visual/hierophant/blast/B = new(t, user, friendly_fire_check)
 		B.damage = 15 //keeps monster damage boost due to lower damage
 
-/obj/item/hierophant_club/admin
+/obj/item/hierophant_club/station
 	z_level_check = FALSE
 //Just some minor stuff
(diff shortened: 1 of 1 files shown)


athena413:
yes

Alexandria404:
more elegant than what I did

alexkar598 (changes requested):
There is absolutely no point in using the club to debug anything at all. Name it admin or station instead since that's what it does.

Takahiru:
i didnt know what to name it lol

code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -1371,7 +1372,8 @@
 		var/obj/effect/temp_visual/hierophant/blast/B = new(t, user, friendly_fire_check)
 		B.damage = 15 //keeps monster damage boost due to lower damage
 
-
+/obj/item/hierophant_club/admin
```
Hopekz:
```suggestion
/obj/item/hierophant_club/station
```
Might be nit picky but I think "station" is more appropriate here.
All of our "admin" variants of items have ridiculous stats so naming this one "admin" is almost misleading since it is just the regular item able to function on the station.

Hopekz (changes requested):

Hopekz (approved):

PR #8167, opened 2020-04-12
This is up for debate (Ash-walker Update)

Readystorm546:
Ashwalkers can put on shoes because they cant craft the tribal shoes but cant put it on. I think thats dumb.

Oh and there was a unnecessary comment in the file i wad looking into that "may or may not oppress lizard people" so i fixed it

:cl:   
tweak: Allows ash walkers to put on the shoes they craft
spellcheck: Also fixed a bad comment
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -575,7 +575,7 @@
 	if(istype(head, /obj/item/clothing/head/wizard) || istype(head, /obj/item/clothing/head/helmet/space/hardsuit/wizard))
 		threatcount += 2
 
-	//Check for nonhuman scum
+	//Check if nonhuman
 	if(dna && dna.species.id && dna.species.id != "human")
 		threatcount += 1
 
@@ -995,7 +995,7 @@
 
 /mob/living/carbon/human/species/golem/durathread
 	race = /datum/species/golem/durathread
-	
+
 /mob/living/carbon/human/species/golem/snow
 	race = /datum/species/golem/snow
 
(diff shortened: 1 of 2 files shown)


athena413:
good

overlord12345:
would be better if you changed the shoes they craft to fit them instead of removeing the digitigrade trait

Readystorm546:
> would be better if you changed the shoes they craft to fit them instead of removeing the digitigrade trait

The trait still exists. Just does not default to ashwalkers anymore.

overlord12345:
> The trait still exists. Just does not default to ashwalkers anymore.

i might have worded it wrong, but what i mean is dont remove the trait from ashies

Amelia0010:
>Ashwalkers can put on shoes because they cant craft the tribal shoes but cant put it on. I think thats dumb.

Just make them unable to craft shoes then :)

HomocidlePancakes:
HeEy y not hav them shudd gunz they kan mag gunz!1!11

SomeguyManperson:
if you want to let them use chitin boots make a species exception in the boots don't remove their digitigrade

HomocidlePancakes:
Fun fact, bone armor covers feet, but chitin armor doesn’t for whatever reason, even though the first used to craft the second.

Readystorm546:
> if you want to let them use chitin boots make a species exception in the boots don't remove their digitigrade

This was my original idea. But the code for those legs is so convoluted that it was hard to find where it marked shoes and on top of that if it even marked it at all. Which makes me think how could i put in a variable or some sort of parameter that accepts only those shoes.

alexkar598 (changes requested):
Leave the comments alone, dont keep all changes to a minimum. Also, what others said, make the shoes have an exception for them somehow. Removing a feature for a niche rarely used one is stupid.

Marmio64:
yeah lets keep ashwalkers with digitigrade legs and just make them be able to wear their specific craftable shoes, like what most people here have said.

PR #8171, opened 2020-04-13
fixes nukeops and actually fixes lavaland/asteroid ruins on icebox

SomeguyManperson:
:cl:  
fix: nuke ops can land on icebox
fix: planet atmos is no longer  hyperfucky on icebox
tweak: snow doesn't slow you after being dug since there's no more snow
/:cl:


athena413:
yes

PR #8173, opened 2020-04-13
Adds guidelines to lobby music

nesisknesis:
//Songs added to the list should be fitting lobby music; sci-fi is encouraged

//As songs are selected randomly, the list should be concise so that certain songs play more often. Songs present for more than one month are eligible to be replaced/removed to achieve this and to keep the list fresh.
#### Changelog

:cl:  Identification
experimental: The lobby music now follows a set of guidelines for direction.
/:cl:


athena413:
what

PR #8176, opened 2020-04-14
Fixes clock cultists not being able to understand common when declaring war

Hopekz:
This adds a language holder to clockwork golems because TGuinext didn't support clockwork golems (because TG doesn't have the Clockwork cult)

This bug only affects you if you have a second language before transforming (such as being a lizard or preternis)

image

#### Changelog

:cl:  Hopek
rscadd: Makes clockwork golems compatible with tgUI's language holder system.
bugfix: fixed clock cult not understanding common after declaring war.
/:cl:


athena413:
good

PR #8178, opened 2020-04-14
Modifies homophobic code / entries before the github police gets here

Hopekz:


azzzertyy:
how are rainbow shoes homophobic when its literally the gay flag

nesisknesis:
🏳️‍🌈

Hopekz:
> how are rainbow shoes homophobic when its literally the gay flag

I don't think its bad here but TG literally got taken down for a similar reason.

I just don't want this place to be taken down over having the word "gay" in the wrong place according to what github deems acceptable.

I'm fine with removing the possibility of this happening and losing "gay" being used the right way on these clown shoes.

Alexandria404:
When github is so idiotic and vague about what is an issue and what is not.
This becomes the response.
Dark times.

athena413:
good

PR #8180, opened 2020-04-15
Readds the Gang Implant Breaker.

Xoxeyos:
Re-adds the Gang implant breaker. It also increases the points needed to 26.

🆑
rscadd: Readds the gang implant breaker.
/🆑


Hopekz:
The implant breaker being gone adds more fun elements to the game mode than before.

I really like that now we end up with fewer situations where the whole crew becomes converted into gangs and actually removing people from the equation instead of putting them all into 1 gang is a positive from what I've seen since this got merged.

In my opinion:
Less is better in this case.

athena413:
nah

PR #8203, opened 2020-04-17
Adds hierophant antenna traitor item

Takahiru:
Adds a 14 TC hierophant antenna traitor item, allows traitor shaft miners to use the hierophant club on the station.


Takahiru:
oh no what is merge conflict

Alexandria404:
very nice

code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -1372,8 +1372,25 @@
...
+
+/obj/item/hierophant_club/attackby(obj/item/I, mob/user)
+	if(istype(I, /obj/item/hierophant_antenna) && z_level_check == TRUE)
+		z_level_check = FALSE
+		desc += " It has an ominous antenna attached."
+		qdel(I)
+	else
+		if(istype(I, /obj/item/hierophant_antenna) && z_level_check == FALSE)
+			to_chat(user, "<span class='warning'>The herald's power already reaches this club!</span>")
+
```
alexkar598:
```suggestion
/obj/item/hierophant_club/attackby(obj/item/I, mob/user)
	if(istype(I, /obj/item/hierophant_antenna))
		if(z_level_check)
			z_level_check = FALSE
			desc += " It has an ominous antenna attached."
			qdel(I)
		else
			to_chat(user, "<span class='warning'>The herald's power already reaches this club!</span>")
		return TRUE
	else
		return ..()

```

alexkar598 (changes requested):

code/modules/mining/lavaland/necropolis_chests.dm
alexkar598: That brings slightly cleaner code and proper parent calling

PR #8211, opened 2020-04-18
Synthflesh only works on corpses

Hopekz:
Synthflesh right now works like a ranged pre-nerf legion cores except you can make infinite amounts of them and load them with beakers to splash people.

This at the moment functions like an instant aheal.

Synthflesh's purpose is to fix dead bodies to be able to do other things to them with other chems.

With this change, other chems have a chance to be useful instead of just using synthflesh in healing.

#### Changelog

:cl:  
tweak: Snythflesh now only works on corpses per its original intent instead of being a ranged instant heal.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -389,9 +389,7 @@
 	color = "#FFEBEB"
 
 /datum/reagent/medicine/synthflesh/reaction_mob(mob/living/M, method=TOUCH, reac_volume,show_message = 1)
-	if(iscarbon(M))
-		if (M.stat == DEAD)
-			show_message = 0
+	if(iscarbon(M) && M.stat == DEAD)
 		if(method in list(PATCH, TOUCH))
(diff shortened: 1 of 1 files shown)


Jarod1200:
So if sythflesh was meant to be used on dead corpses why can I buy it in advanced medkits in medbay?

Hopekz:
> So if sythflesh was meant to be used on dead corpses why can I buy it in advanced medkits in medbay?

This is a balance change meaning it wasn't originally month this way but it's modified this way to make the game more fair to everyone involved

Marmio64:
or just make it so synthflesh is way less effective on the living.

Mqiib:
Muh powergaming though

BOBsHERE:
I think a better test merge would be reducing the healing of synthflesh on living bodies first, and if that still feels overpowered then remove it.

alexkar598 (approved):

alexkar598:
Not a fan of this solution, synthflesh is meant to be a potent heal and I dont mind the strong healing abilities of chem, it's the offensive abilities I dont like. Strong heals fits the support role of chem well

PR #8214, opened 2020-04-18
Removes the reinforced chromazone from the game

Hopekz:
Really experienced gene mains are telling me this is the only real OP part of Gene that is abused.

"90% of people dont know how to remove reinforced genes causing a select few to be able to abuse them. if genetics console is destroyed it becomes unremovable. mutadone does not cure a reinforced gene" - puremanbird (one of those gene mains)

#### Changelog
:cl:  Hopek
rscdel: Sent reinforced chromazone to the shadow realm
/:cl:

### DIFF
--- a/code/game/objects/items/chromosome.dm
+++ b/code/game/objects/items/chromosome.dm
@@ -75,18 +75,3 @@
 	desc = "A chromosome that reduces cooldown on action based mutations by 50%."
 	icon_state = "energy"
 	energy_coeff = 0.5
-
-/obj/item/chromosome/reinforcer
-	name = "reinforcement chromosome"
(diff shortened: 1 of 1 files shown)


MayhemSailor:
Chem and Genetics are frequently manned by validhunters. ¯\_(ツ)_/¯

boodaliboo:
maybe make it as rare as the one that makes it more stable

alexkar598 (approved):

alexkar598:
I haven't it seen be an actual issue, lasers can easily take down a hulk, if anything, this chromosome forces sec to crit the bad person instead of simply revoking his powers, causing the geneticist to have a harder time fighting back.

PR #8226, opened 2020-04-18
Military is no longer a gender

yacabo:
Currently whenever VOX is pulled up a prompt asks the user which gender VOX they want: male, female, or military. Since this is an issue I thought I might as well give them actual names instead of making this only get weirder if even more voices get added.

NAME CHANGES:

**male** is now **Victor**
**female** is now **Verity**
**military** is now **Oscar**

The previous name will remain in parenthesizes for the time being and I'll eventually remove them entirely so that people aren't confused.

For an explanation of the names, Verity and Victor are both names of AI cockpit assistants you choose for your ship in _Elite Dangerous_, Verity being the default and once only assistant, Victor becoming the first free one added, while Oscar is the letter "O" in the NATO alphabet and sounded more fitting than Mike.

I'm not too attached to the names (minus Oscar) so feel free to suggest better ones if you want to.

:cl:  
tweak: VOX voices now have names. Male is Victor, Female is Verity, Military is Oscar
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/ai/say.dm
+++ b/code/modules/mob/living/silicon/ai/say.dm
@@ -100,7 +100,7 @@
 
 	last_announcement = message
 
-	var/voxType = input(src, "Male, female, or military VOX?", "VOX-gender") in list("male", "female", "military") //yogs - male vox
+	var/voxType = input(src, "Which voice?", "VOX") in list("Victor", "Verity", "Oscar") //Victor is vox_sounds_male, Verity is vox_sounds, Oscar is vox_sounds_military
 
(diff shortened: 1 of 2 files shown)


alexkar598 (changes requested):
Either make it Name (Gender) or leave it as is and call it a voice type. This might be confusing for people using the VOX for the first time/dont track the github

yacabo:
Done, however we should at some point in the future change it to just the names or this PR is pointless

Hopekz (approved):
Verified you implemented Alex's suggestion.
I feel confident in merging this.

PR #8229, opened 2020-04-19
Centers all of Yogstation

Hopekz:
All of Yogstation was off-center by close to 40 tiles!!
We're all playing in an off-center world!

beforecenter
beforecenter 2

#### Changelog

:cl:  
tweak: Recentered the entirety of Yogstation as it was off by close to 40 tiles!!
/:cl:


monster860:
yes but this uncenters the bridge?

Hopekz:
> yes but this uncenters the bridge?

What does this have to do with anything?
The Center of something in a confined space is amount of tiles from each edge furthest to be as close to equal.

ThatLing:
But why?

Hopekz:
> But why?

Because an off-center map is the first step to anarchy.

Also; it is noticeable in-game on how short the travel distance is to the edge before deep space. 
Just take a look at how much smaller the box is on the right side compared to the left! 40 tiles is a big deal IMO

image

Hopekz:
I don't think this type of correction PR is possible on our server because our HC's maintainers are too slow to implement before merge conflicts.

I had 18 and even after I'm not 100% sure if the resolved code is in the right place because the whole station moved.

This PR would have to be recreated when both are more active in the future to prevent merge conflicts for others as well.

ThatLing:
Why can't you just coordinate with the HCs?

Hopekz:
> Why can't you just coordinate with the HCs?

It is partially my fault because a PR like this needs to be created when no other mapping PRs are open or else every single person is going to get a merge conflict.

In the future I will coordinate but I will wait until all mapping PRs are a resolved to make this easier.

PR #8240, opened 2020-04-20
[admin] quick Mulligan in the equipment screen

Hopekz:
image

The new "a Mulligan" option to quickly spawn yourself or others in as a randomized individual.

I have thought of needing an easy tool like this for a while so we can spawn as a generic centcom individual, quickly bus someone in as a random character instead of their IC etc.

#### Changelog

:cl:  Hopek
rscadd: Added the ability to quickly spawn as a randomized character via the "Select Equipment" right-click menu.
/:cl:

### DIFF
--- a/code/datums/outfit.dm
+++ b/code/datums/outfit.dm
@@ -239,3 +239,12 @@
 			implants += imptype
 	accessory = text2path(outfit_data["accessory"])
 	return TRUE
+
+
+/datum/outfit/mulligan
+	name = "a Mulligan"
+
+/datum/outfit/mulligan/post_equip(mob/living/carbon/human/H, visualsOnly = FALSE)
+	..()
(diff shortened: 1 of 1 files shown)


code/datums/outfit.dm
```diff
@@ -238,4 +238,13 @@
 		if(imptype)
 			implants += imptype
 	accessory = text2path(outfit_data["accessory"])
-	return TRUE
\ No newline at end of file
+	return TRUE
+
+
+/datum/outfit/mulligan
+	name = "a Mulligan"
```
alexkar598: name it "a Mulligan (Naked)" instead, admins dont have the highest IQ when it comes to using admin tools

alexkar598 (changes requested):

Hopekz:
Modified Alex's changes.
Looks good to go

PR #8250, opened 2020-04-23
increase engiborg stunprod arm cost

SomeguyManperson:
less than a stunbaton at 1000 but more than a peacekeeper borg's stun attack at 500.
:cl:  
tweak: emagged engiborg stun arm cell cost per use 750 from 30 
/:cl:

### DIFF
--- a/code/game/objects/items/robot/robot_items.dm
+++ b/code/game/objects/items/robot/robot_items.dm
@@ -8,7 +8,7 @@
 /obj/item/borg/stun
 	name = "electrically-charged arm"
 	icon_state = "elecarm"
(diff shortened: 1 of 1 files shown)


Marmio64:
you are doing god's work

Hopekz:
So why shouldn't it be more expensive than a stun baton?
A stun baton is a specialized tool meaning it'll be more efficient and a stunning arm is them pumping power into their arms to stun.

Marmio64:
but, i gotta agree with hopek, stunarm still is too cheap tbh, and should be increased to 1250 power cost. Engiborg already gets selfrepair and shit.

PR #8256, opened 2020-04-24
Removes OD risk on cigars

ToGWtF:
Before : Cohiba Robusto cigar would overdose you 100% of the time if smoked from start to finish

Now : doesn't.

#### Changelog

:cl:  
tweak: Cigars no longer OD
/:cl:

### DIFF
--- a/code/game/objects/items/cigs_lighters.dm
+++ b/code/game/objects/items/cigs_lighters.dm
@@ -400,7 +400,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	icon_off = "cigar2off"
 	smoketime = 600 // 20 minutes
 	chem_volume = 80
-	list_reagents =list(/datum/reagent/drug/nicotine = 40)
+	list_reagents =list(/datum/reagent/drug/nicotine = 35)
 
 /obj/item/clothing/mask/cigarette/cigar/havana
 	name = "premium Havanian cigar"


Hopekz:
I personally like the OD feature.

It didn't feel like something "out of place" since they're supposed to be the strong cigars. 
With this fix it just kind of makes it feel like every other cigar.

ynot01:
sure

ToGWtF:
Well CentCom officers spawn with these cigars, and I think having your inspector you send down having random lung failure 10 minutes in because he didn't know is not a good look. 

Alternative, remove cohibas from smoker trait if they are meant to be special? Spawning with them basically forces you to either
A)keep track of how long youve been smoking one
B)get them made into patches
or C) have a negative mood buff permanently from not smoking/smoking a different brand.

Marmio64:
eheheeheh... yeah i kind of think that cohiba should stay and just have the officer cigar swapped out with havana. People with the smoker negative trait can just disable mood....

Hopekz:
> Well CentCom officers spawn with these cigars, and I think having your inspector you send down having random lung failure 10 minutes in because he didn't know is not a good look.
> 
> Alternative, remove cohibas from smoker trait if they are meant to be special? Spawning with them basically forces you to either
> A)keep track of how long youve been smoking one
> B)get them made into patches
> or C) have a negative mood buff permanently from not smoking/smoking a different brand.

Officers do not spawn with this; only the commander but really it isn't a big deal in that regard as they're only able to be spawned in at the moment via bus.

You can however just remove this cigar spawning in the smoker trait. 
It should be an easy fix while maintaining the reason this cigar became special recently.

alexkar598:
They are stonk, stronk cigars are dangerous, make CC officers not spawn with them if you want but leave the STRONK cigars alone

PR #8252, opened 2020-04-23
ports Titanium Resprite from /tg/

nesisknesis:
https://github.com/tgstation/tgstation/pull/50454

image

#### Changelog

:cl:  Onule
imageadd: Recent changes in the manufacture of titanium sheets introduces a fresh new look! Expect your shuttles to look different.
/:cl:

PR #8259, opened 2020-04-24
Revert "ports Titanium Resprite from /tg/"

JamieD1:
Reverts yogstation13/Yogstation#8252

Clearly there is some controversy over this,

Self merging because "you'll be the judge of that"

Sounds like a shitty way to do community coding.


nesisknesis:
As Art Director I have full authority on the graphic content of the server, the PR was merged because they are a clear improvement.

If you have a problem with how this works, take it up with me or the headcoders.

PR #8265, opened 2020-04-25
Added OmegaStation to vote list

JamieD1:

Adds OmegaStation to vote list

#### Changelog

:cl:  
rscadd: Added Omegastation to vote list
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -24,6 +24,7 @@ endmap
 
 map omegastation
 	maxplayers 35
+	votable
 endmap
 
 map yogsmeta


Hopekz:
Might as well add IceBox since this is a config change so both get knocked out at once with 1 approval

JamieD1:
Thought Icebox crashed a lot? 

Maybe we shouldn't let that one be votable without admins on to reboot.

Hopekz:
> Thought Icebox crashed a lot?
> 
> Maybe we shouldn't let that one be votable without admins on to reboot.

I'm not aware of any crashes at the moment. 
If this is actually true then maybe it is best to leave them out.

@SomeguyManperson ;
Does IceBox crash our server?

alexkar598:
Icebox is still in a beta ish phase atm as far as I am aware

Jarod1200:
Can you make it so if there is more than a certain number of players it's removed from the list? This station is kinda aids with more than like 40 people

JamieD1:
> Can you make it so if there is more than a certain number of players it's removed from the list? This station is kinda aids with more than like 40 people

It has a max player count of 35 I believe.

alexkar598:
Omegastation is a really bad map compared to pubby which is omega but better. I'd rather have the map pool bring quality instead of quantity

MegaEmpirical:
I fucking like omega and want it and it's by player choice

aaqe:
I love this map. Please let us play it on low pop!

Hopekz:
Seeing the support on this; reopening it as it was closed for a weird opinion reason.
We have lowpop on our server which pubby doesn't fulfill. 

If we plan to remove this from the vote rotation we need to remove it entirely because what is the point of keeping the map itself if it'll not be used?

Even on the admin side, you making this not a part of the actual vote forces us to make a weird Janky map vote from scratch which is just awkward to create.

Also; it is a matter of opinion because I personally think pubby is shit compared to omega.

alexkar598:
No.

Hopekz:
Alright well, I won't let 1 opinion stop this without a fair decision seeing as this has support. 
I want an official decision to be made so I guess the council now needs to vote on it since you're turning this "controversial"

Opening right now; do not merge until this gets settled.

MegaEmpirical:
It's definitely not controversial if only one person dislikes it

PR #8275, opened 2020-04-25
better captain annoncement

enkidienne:
makes the captain annoncement able to have multiple lines and spaces. Like a centcom annoncement

changelog:
/🆑 
tweak: changed the stripped format of the Captain's announcement to a full format that allows more creativity and possibilities.
/:cl:

### DIFF
--- a/code/game/machinery/computer/communications.dm
+++ b/code/game/machinery/computer/communications.dm
@@ -703,7 +703,7 @@
 	if(!SScommunications.can_announce(user, is_silicon))
 		to_chat(user, "Intercomms recharging. Please stand by.")
 		return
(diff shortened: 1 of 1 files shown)


alexkar598 (changes requested):
lmao, this removes all sanitzation for it, meaning anyone will be able to post any HTML content they want

Hopekz:
I think the main reason for this is that captain announcement is random user input (whoever plays captain) while Centcom announcement is via admins+ so if we do something like an HTML attack it will be less likely since it is closely monitored as far as who gets access to it.

enkidienne:
i fixed it, assigned it to a multi line strip

PR #8290, opened 2020-04-27
[admin] Add spiderizing as a smiting option

Birddude1230:

[Implements this coding request.](https://forums.yogstation.net/index.php?threads/but-she-was-already-pouring-you-a-brimming-glass-of-spiders.20398/)

Adds a new entry under the admin smiting menu ("Spiderize") to explode a mob into spiders.  Specifically, the mob will be gibbed, and 2-4 spiderlings will spawn at its location. As the intent in the forum post is to deal with the body of a griefer, the brain is destroyed in the gibbing.

why did I write spidirizing to begin with?

#### Changelog

:cl:  
rscadd: Adds a spiderizing smite that destroys the brain and spawns spiderlings!
/:cl:

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -1102,6 +1102,7 @@ Traitors and the like can also be revived with the previous role mostly intact.
 			for (var/i in 1 to rand(2,4))
 				var/obj/structure/spider/spiderling/S = new(target.drop_location())
 				S.directive = "Erase the legacy of [target.name]."
+				addtimer(CALLBACK(S, .Destroy), 50) //just enough time for them to skitter around a little
 			target.gib(TRUE)
 		if(ADMIN_PUNISHMENT_BRAINDAMAGE)
 			target.adjustOrganLoss(ORGAN_SLOT_BRAIN, 199, 199)
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -1081,7 +1081,7 @@ Traitors and the like can also be revived with the previous role mostly intact.
 	if(!check_rights(R_ADMIN) || !check_rights(R_FUN))
 		return
 
-	var/list/punishment_list = list(ADMIN_PUNISHMENT_LIGHTNING, ADMIN_PUNISHMENT_BRAINDAMAGE, ADMIN_PUNISHMENT_GIB, ADMIN_PUNISHMENT_BSA, ADMIN_PUNISHMENT_FIREBALL, ADMIN_PUNISHMENT_ROD, ADMIN_PUNISHMENT_SUPPLYPOD_QUICK, ADMIN_PUNISHMENT_SUPPLYPOD, ADMIN_PUNISHMENT_MAZING)
+	var/list/punishment_list = list(ADMIN_PUNISHMENT_LIGHTNING, ADMIN_PUNISHMENT_SPIDERIZE, ADMIN_PUNISHMENT_BRAINDAMAGE, ADMIN_PUNISHMENT_GIB, ADMIN_PUNISHMENT_BSA, ADMIN_PUNISHMENT_FIREBALL, ADMIN_PUNISHMENT_ROD, ADMIN_PUNISHMENT_SUPPLYPOD_QUICK, ADMIN_PUNISHMENT_SUPPLYPOD, ADMIN_PUNISHMENT_MAZING)
 
 	var/punishment = input("Choose a punishment", "DIVINE SMITING") as null|anything in punishment_list
 
@@ -1097,6 +1097,12 @@ Traitors and the like can also be revived with the previous role mostly intact.
 				var/mob/living/carbon/human/H = target
 				H.electrocution_animation(40)
 			to_chat(target, "<span class='userdanger'>The gods have punished you for your sins!</span>")
+		if(ADMIN_PUNISHMENT_SPIDERIZE)
+			target.visible_message("<span class='danger'>[target.name] explodes into spiders!</span>", "<span class='userdanger'>The gods have rended your form!</span>", "<span class='italics'>You hear chittering and a horrible wet sound!</span>")
+			for (var/i in 1 to rand(2,4))
+				var/obj/structure/spider/spiderling/S = new(target.drop_location())
+				S.directive = "Erase the legacy of [target.name]."
+			target.gib(TRUE)
 		if(ADMIN_PUNISHMENT_BRAINDAMAGE)
 			target.adjustOrganLoss(ORGAN_SLOT_BRAIN, 199, 199)
 		if(ADMIN_PUNISHMENT_GIB)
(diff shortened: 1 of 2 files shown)


Hopekz (changes requested):
Smiting is meant to punish 1 player. 
If you spawn spiderlings you just create full-on big spiders later down the line.

Either make these spiderlings into a child that spawns into harmless spiderlings that don't do anything or have them be deleted sometime after so they don't transform (or both)

As an admin, I want a "smite and forget" solution and this is a "smite then worry about deleting spiderlings so I don't fuck up the shift with brood mothers" solution.

Marmio64:
yeah, might wanna have the spiderlings be sterile and not grow up

Hopekz:
I think the idea is nice but the execution is a bit underwhelming.
Besides like 2-3 spiders spawning it isn't really that much different than a gib slay (at least not different enough to warrant adding it as a separate gib option compared to all the other options)

4HyZGrGWD0

Not sure what can be done to make it "different". More spiders? 
Maybe if you add like 10-20 spiders you might get the effect you are looking for.

nesisknesis:
Make it stun them while spiders spawn with flavor text. Some nice text about crawling out of orifices or something. Then gib them, then more spiders.

Birddude1230:
Would you happen to know an upper limit on the number of spiderlings I can spawn before it's dangerous to the server? Now that I've tried it, it's definitely a lot more effective when the spiderlings cover the room.
If it's not a really bad idea for performance something like stun -> spawn ~50 spiderlings -> gib -> spawn ~200 spiderlings. I should also add some variation to the death timer so they don't all die in the same tick, since that looks odd.

After all, it would take **a lot** of spiderlings to fill a body. 
many spiderlings

code/modules/admin/verbs/randomverbs.dm
```diff
@@ -1097,6 +1097,13 @@ Traitors and the like can also be revived with the previous role mostly intact.
 				var/mob/living/carbon/human/H = target
 				H.electrocution_animation(40)
 			to_chat(target, "<span class='userdanger'>The gods have punished you for your sins!</span>")
+		if(ADMIN_PUNISHMENT_SPIDERIZE)
+			target.visible_message("<span class='danger'>[target.name] explodes into spiders!</span>", "<span class='userdanger'>The gods have rended your form!</span>", "<span class='italics'>You hear chittering and a horrible wet sound!</span>")
+			for (var/i in 1 to rand(2,4))
+				var/obj/structure/spider/spiderling/S = new(target.drop_location())
+				S.directive = "Erase the legacy of [target.name]."
+				addtimer(CALLBACK(S, .Destroy), 50) //just enough time for them to skitter around a little
```
alexkar598: use the QDEL_IN proc instead

alexkar598 (changes requested):

code/modules/admin/verbs/randomverbs.dm
Birddude1230: Thanks!

Birddude1230:
Very small gif of player perspective -- maybe I should make it cause damage each time it spawns more spiders. Also not sure that the code is up to standard with regards to magic numbers and proc placement, but I didn't want to introduce a bunch of constants to this already-giant file.

Maybe I should figure out the file structure and put smiting in its own file.

They crawl within you.

Hopekz:
So are these supposed to despawn?
In my tests, this leaves a lot of spiders just kind of littered everywhere.
Some despawn but not all of them.

dreamseeker_LQAVCPKl8h

Also, something to keep in mind is that they vent crawl as well so there might be spiders that are dead in the vents as well. Couldn't find them exit anywhere so thought I'd mention.

dreamseeker_uA1t7tHZNT

code/modules/admin/verbs/randomverbs.dm
```diff
@@ -1075,13 +1075,27 @@ Traitors and the like can also be revived with the previous role mostly intact.
...
+	target.visible_message("<span class='danger'>Spiders emerge from [target.name]!</span>", "<span class='userdanger'>They crawl within you.</span>")
+	for (var/i in 1 to rand(progress * 9, progress * 11))
+		var/obj/structure/spider/spiderling/S = new(target.drop_location())
+		S.directive = "Erase the legacy of [target.name]."
+		QDEL_IN(S, rand(50, 300)) //just enough time for them to skitter around a little
+	if (progress < 5)
+		addtimer(CALLBACK(src, /client/proc/spiderize_more, target, progress + 1), 75)
+	else
+		target.visible_message("<span class='danger'>[target.name] explodes into a chittering mass!</span>", "<span class='userdanger'>The gods have rended your form.</span>")
+		target.gib(TRUE)
```
alexkar598, line 1090: 1. This should really be a proc on the mob, not on the client of the admin
2. We only use the `as` modifier thing for procs which are called directly from the right hand panel(verbs)
3. Thats a bit too many spiders... 30 to 68 spiders is more reasonable than 135 to 165 spiders

```suggestion
/mob/living/carbon/proc/spiderize(progress = 1)
	SetAllImmobility(500)
	do_jitter_animation(progress * 200)
	visible_message("<span class='danger'>Spiders emerge from [target.name]!</span>", "<span class='userdanger'>They crawl within you.</span>")
	for (var/i in 1 to rand(progress * 3, progress * 5))
		var/obj/structure/spider/spiderling/S = new(drop_location())
		S.directive = "Erase the legacy of [target.name]."
		QDEL_IN(S, rand(50, 300)) //just enough time for them to skitter around a little
	if (progress < 5)
		addtimer(CALLBACK(src, /mob/living/carbon.proc/spiderize, progress + 1), 75)
		return
	visible_message("<span class='danger'>[target.name] explodes into a chittering mass!</span>", "<span class='userdanger'>The gods have rended your form.</span>")
	gib(TRUE)
```

code/modules/admin/verbs/randomverbs.dm
```diff
@@ -1097,6 +1111,10 @@ Traitors and the like can also be revived with the previous role mostly intact.
 				var/mob/living/carbon/human/H = target
 				H.electrocution_animation(40)
 			to_chat(target, "<span class='userdanger'>The gods have punished you for your sins!</span>")
+		if(ADMIN_PUNISHMENT_SPIDERIZE)
+			target.visible_message("<span class='danger'>[target.name] freezes!</span>", "<span class='userdanger'>You feel the weight of your actions.</span>")
+			target.SetAllImmobility(500)
+			addtimer(CALLBACK(src, /client/proc/spiderize_more, target, 1), 25)
```
alexkar598, line 1117: Since its on the mob now instead of the client

```suggestion
			addtimer(CALLBACK(target, /mob/living/carbon.proc/spiderize), 25)
```

alexkar598 (changes requested):
Also im slightly concerned that it creates tons of dead spiderlings but I suppose it just creates free food

alexkar598:
This PR has been closed due to being stale. Contact me or another maintainer on Discord if you wish for this PR to be reopened.

PR #8302, opened 2020-04-27
Toolbelt can now hold airlock painter

Hopekz:
It seems like another oversight.

This thing fits in the belt:
image

but this similar-looking smaller thing doesn't?:
image

#### Changelog

:cl:  Hopek
rscadd: Airlock painter now fits in the toolbelt
/:cl:


alexkar598 (changes requested):
Merge cornflakes

PR #8308, opened 2020-04-28
Ports "Removes Box Station #50726" from TG

ynot01:
ports https://github.com/tgstation/tgstation/pull/50726
## About The Pull Request
Removes Box Station Entirely

## Why It's Good For The Game
Box Station has been antiquated and deserved removal for a very long time coming now.

Recent attempts to update particularly bad parts of Box (powernet, pipenet, etc.) fail to address the single largest issue with box. Meta was so heavily based on Box that Meta is effectively just a better version of Box.

The way I see it Box Station will never match either the popularity or the quality of MetaStation due to fundamental problems with the layout and design.
On top of that, they are such similar of maps that maintaining both at the same time feels like it adds very little. Say what you will about maps that aren't Box or Meta, they are the two most fundamentally similar maps in the game. Even Delta, which is yet another Box derivative, doesn't come off nearly as similar to Box.

There is a push to reduce the number of active maps and there is zero reason to keep around a bunch of maps that are extremely similar. The whole advantage of running multiple maps is variety, and having only Box, Meta, Delta, and the flavor of the month fourth map that everyone hates because it's not one of the three extremely similar maps is a colossal waste of time and effort for everybody involved.

## Changelog
🆑 actioninja
del: Box Station
/🆑

please don't ban me from the codebase


Hopekz (changes requested):
Merge conflict

Jumps0:
madman

ThatGuyPhil:
It must be done

PR #8310, opened 2020-04-28
Ports "Removes Box Station #50726" from TG

ynot01:
ports https://github.com/tgstation/tgstation/pull/50726
## About The Pull Request
Removes Box Station Entirely

## Why It's Good For The Game
Box Station has been antiquated and deserved removal for a very long time coming now.

Recent attempts to update particularly bad parts of Box (powernet, pipenet, etc.) fail to address the single largest issue with box. Meta was so heavily based on Box that Meta is effectively just a better version of Box.

The way I see it Box Station will never match either the popularity or the quality of MetaStation due to fundamental problems with the layout and design.
On top of that, they are such similar of maps that maintaining both at the same time feels like it adds very little. Say what you will about maps that aren't Box or Meta, they are the two most fundamentally similar maps in the game. Even Delta, which is yet another Box derivative, doesn't come off nearly as similar to Box.

There is a push to reduce the number of active maps and there is zero reason to keep around a bunch of maps that are extremely similar. The whole advantage of running multiple maps is variety, and having only Box, Meta, Delta, and the flavor of the month fourth map that everyone hates because it's not one of the three extremely similar maps is a colossal waste of time and effort for everybody involved.

## Changelog
:cl: actioninja
del: Box Station
/:cl:


Takahiru:
bad pr, this probably requires a council vote tbh

fluffe9911:
Worst pr

Hopekz:
We literally don't have the same issues here.

In your PR description, it says one of the main reasons is because "powernet, pipenet, etc." need to be updated. 
Well; we're not really updating those systems. 

It just seems like a bandwagon PR tbh just because you saw TG do it. Why don't you just play on TG at this point instead of trying to change Yogstation to be TG?

ynot01:
> Why don't you just play on TG at this point instead of trying to change Yogstation to be TG?

coby chm

Hopekz:
Based on your responses on discord and your lack of argument when asked I think this is a meme PR.
Closing it.

athena413:
based

PR #8311, opened 2020-04-28
Bloodcult rune nerfs

SomeguyManperson:
Barrier runes last 5 minutes after being activated, blocking basically any form of attack with almost no cost, and can be chain-activated so the cost for activating one is the cost for activating several. This means you can shove down a circle of barrier runes around your base or the summon site and unless someone has a nullrod or some high power explosive to bust it nothing can break through, making it better than a real wall in every way.
Blood boil deals a disgusting amount of damage to anything that can see it at the cost of getting 3 people around the rune and causing a small fire for like 2 seconds. The damage is nerfed so it's still considerable but not game ending.
:cl:  
tweak: barrier rune now lasts 30 seconds down from 5 minutes and has a cooldown of 10 seconds up from 5 seconds since 5 minutes is long enough to microwave a pizza
tweak: barrier runes are more discernable between active and inactive state
tweak: blood  boil rune only does burn damage, reducing its total damage to 75 from 125
/:cl:


ynot01:
its balnaced winrat4e no nerf

bad mintenri

code/modules/antagonists/cult/runes.dm
```diff
@@ -148,7 +148,7 @@ structure_check() searches for nearby cultist structures required for the invoca
 		else if(istype(M, /obj/item/toy/plush/narplush))
 			var/obj/item/toy/plush/narplush/P = M
 			P.visible_message("<span class='cult italic'>[P] squeaks loudly!</span>")
-	if(!src.density) //yogs: barrier runes play cooldown animation properly
```
ThatLing: Why are you removing the 'yogs'?

ThatLing (changes requested):
Add a proper description detailing why you want this change to be added

code/modules/antagonists/cult/runes.dm
SomeguyManperson: don't need it since we aren't mirroring TG

ThatLing: It's helpful for when we rebase again in the future. Keeping them doesn't hurt

ThatLing:
You don't have to add yogs to your new changes, just keep the old ones

BOBsHERE:
We have too many blood cult change PRs up at one time. If the endgame changes get merged this should not get merged.

alexkar598 (approved):

PR #8314, opened 2020-04-28
Shadowling speed lost with species loss

SomeguyManperson:
:cl:  
bugfix: no infinite speed from being a shadowlingb then not  
/:cl:


deathslayer2481:
Methspeed cult golem when

PR #8280, opened 2020-04-26
Gulag Processor & The Return of the Beepsky

Bibby0110:
### Intent of your Pull Request
It's the gulag teleporter, but without the teleporter part

Guide:
Take prisoner ID.
Configure the points in the normal console.
Insert ID into gulag processor.
Put prisoner into gulag processor
Prisoner is stripped and given prison clothing.
ID is equipped on prisoner.
Prisoner is set to Incarcerated
Prisoner is ziptied.

All their stuff is moved to the reclaimer, which means the prisoner can easily get their stuff back by using their ID on the Equipment reclaimer

Added the Gulag Processor to Boxstation
Removed the locker storage room that was the replacement
Moved beepsky's room back to its original space.
Beepsky is now assembled roundstart
#### Changelog

:cl:  
rscadd: The process of gulagging has been made easier. Put the prisoner into the new Gulag Processor near the shuttle, and they will automatically be stripped.
rscadd: Beepsky now starts assembled
/:cl:

PR #8318, opened 2020-04-28
Adds a light

Ktlwjec1:
#8280 was not head mapper approved


Bibby0110:
omg thank u!!!! can you also add a corner to the striped lines thing, it looks uglyyy

PR #8327, opened 2020-04-29
[admin] Increases number of Response team spawners from 9 to 18

Hopekz:
There is a hard cap to ERT/response team spawners because there just weren't enough spawners.
So if you created an ERT larger than 9 people the rest would not spawn.

Encountered this cap when bussing in huge ERT's which forced me to create 2 identical ERT's to spawn them in 2 waves instead.

The ERT selection menu lets you put in a number as to how many people you want so it is assumed you can go higher than 9.

#### Changelog

:cl:  
bugfix: Added more spawners so admins can now spawn more than 9 ERT/response team members at once (now 18)
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom.dmm
+++ b/_maps/map_files/generic/CentCom.dmm
@@ -18483,6 +18483,10 @@
 	},
 /turf/open/floor/plasteel,
 /area/centcom/control)
+"NH" = (
+/obj/effect/landmark/ert_spawn,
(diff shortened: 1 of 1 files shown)


Ktlwjec1 (approved):
some of the best mapping work ive seen in all my 30 years of life

Hopekz:
Only map change and the mapping pope approved plus everyone seems confused on if this is a map PR so pushing it through

PR #8331, opened 2020-04-29
Replaces ambimalf.ogg with a new sound.

nesisknesis:
Original: https://discordapp.com/channels/134720091576205312/205784753352343552/705089149660692500
New: https://discordapp.com/channels/134720091576205312/205784753352343552/705089229432029254

Replaced because I think the new one sounds more ambient than the original, more because the original is literally a techno drop.

The new one is from SCP:CB and is the ambience for 079's room.

#### Changelog

:cl:  Identification
soundadd: Replaces an ambient sound played in the AI core and two space ruins with a more 'ambient' one.
/:cl:


Marmio64:
i actually like the original one, but it's not really good 'ambience'

Takahiru:
But i like the EVERYBDODYOOWOAWOWOWOEVERYEVE-E-E-EOVEWER *beeps* sound

yacabo:
Why not use both?

Xoxeyos:
Use both!

nesisknesis:
> Why not use both?

> Replaced because I think the new one sounds more ambient than the original, more because the original is literally a techno drop.

PR #8335, opened 2020-04-29
[admin] Mob renames are locked behind +ADMIN instead of +VAR_EDIT

alexkar598:
Diffs are being assed but i simply moved the varedit section delimiter a bit further down and changes the check_rights(NONE) to check_rights(R_ADMIN). tested locally, works fine.


aaqe:
Thank you!

JamieD1:
Thank you!

PR #8347, opened 2020-04-30
adds another engineering locker to box

Ktlwjec1:
there was a wardrobe there still anyway so i replaced it with an engineering locker


maxion12345:
You replaced the drobe with a locker?

Ktlwjec1:
no, the wardrobe closet not the engidrobe. It was leftover from before the new drobes were added in.

PR #8349, opened 2020-04-30
Adds YogStation variations to IceBox

Hopekz:
Adds box variations to IceBox and makes some of them (the ones with space) work with IceBox

#### Changelog

:cl:  Hopek
rscadd: Added variations from YogStation to IceBox
/:cl:


Ktlwjec1 (changes requested):
Changes I mentioned, just putting this here so no one merges it without realising.

PR #8376, opened 2020-05-02
Buffs explosions against black holes idk why it works that way

alexkar598:
Devastation:
Singulo is reduced to 40% of its size
Stage two singularities are destroyed

Heavy:
Singulo is reduced to 60% of its size
Stage one singularities are destroyed

Light:
Singulo is reduced to 70% of its size

All:
If singulo drops below 60 energy after being blown up, it collapses, this is to counter the fact it get exponentially harder to deal damage to the singulo the smaller it is

For a reference, a loose singularity that has around 2.5k energy will take either:
3 devastation blows
6 heavy damage blows
too many light explosions to count

A decent size singulo that hovers near the border of stage two/three which is common for engineering setup(1.1k energy):
2 devastation blows
5 heavy blows
10 light blows

A stage 1 singulo that just sits there vibing at 120 energy unable to grab anything unless it walks in:
1 devastation blow
1 heavy blow
2 light blows


azzzertyy:
pretty sure black holes absorb any energy put into them by definition, so wouldnt this in theory make it stronger?

alexkar598:
idk, it already works that way, dont think about it too hard

Issue #6389, opened 2019-07-19
Camera/Photos are bugged


Slinky5890:
Round ID 24353

I was taking photos and uploading them to the newscaster throughout the round. New photos became invisible and stacked on my hand after I selected the customise photo action and cancelled each prompt (though nich tried to recreate this and it didnt cause the bug). I think the size was 4x4 both times the bug occured.

After the invisible photos stacking, I stopped seeing anything in chat and could not interact with anything that would bring up a menu- it mentioned this in the chat once though https://i.imgur.com/AVejPUm.png. Though I could interact with everything/everyone else normally. Restarting the game both made the photos (which i had dropped on the floor) visible and my other problems were fixed.

It happened again later without the customisation thing, and so I decided to test it with someone else. To see if they could also not see the photos after I dropped them. They could not. After a reset, the photos showed up for me once more, but not for him. He then got the same bug (an engineering cyborg, I think funbot). A reset fixed it for him.

oklews:
this has been a problem since the dawn of time, I would be shocked if it can be fixed

Slinky5890:
> this has been a problem since the dawn of time, I would be shocked if it can be fixed

man, that's a real shame. It's one of my favourite dumb characters

Issue #8317, opened 2020-04-28
Invisible photos couse crashes


PyGGy123:

ToGWtF:
Duplicate of https://github.com/yogstation13/Yogstation/issues/6389

PR #8386, opened 2020-05-04
Adds camera modes + photographer trait uses

Hopekz:
Adds modes to the camera. 

**What this means:** 
You can set your camera to shoot in standard mode or description mode by using the camera in your hand. 

In standard mode, the camera will not prompt you if you want to customize the photo and in description mode, it gives you that annoying popup asking you if you want to customize your photo.

**Edit 2:**
Added a better examine (now tells you that the camera is empty when there are 0 pictures left)
You now have to be close to the camera to look inside of it to see the amount of pictures left for added camera immersion!

**Edit 3:**
Added more use for the Photography trait.

When you have the photography trait you get the following information when you examine:

the exact number of photos left in the camera
the lens setting of the camera
the mode the camera is in

**EDIT 4:**
I managed to recreate the bug after multiple playtests so I am no longer claiming it is fixed.

#### Changelog

:cl:  Hopek
rscadd: The camera has modes. You can now shoot in a standard mode where it won't give you an annoying popup box and a description mode where it prompts you.
rscadd: Expanded the usage of the photographer trait for the Camera.
/:cl:


Hopekz:
Also fixes:
#6389
#8317

Hopekz:
Update:
Added future proofing to include a case that doesn't exist yet if someone were to make a camera that doesn't have the ability to use descriptive mode (if it can't change descriptions)

Hopekz:
Edit 3:
Added more use for the Photography trait.

When you have the photography trait you get the following information when you examine:
- the exact number of photos left in the camera
- the lens setting of the camera
- the mode the camera is in

MegaEmpirical:
Fuck yeah.

Hopekz:
@swissloaf stop fucking with our labels or you'll get that privilege removed . 
This is the 2nd PR where you added or removed the ready label it's for our organization

swissloaf:
This aint approved change

Hopekz:
@alexkar598 you may have forgotten to submit your review with suggestions on why you believe this isn't "ready".

alexkar598:
It's not "ready" because you cant review your own code

Hopekz:
> It's not "ready" because you cant review your own code

I was not aware the "ready" tag had any real meaning outside of organization since we need to perform a final review before merging.

In typical non self merge circumstances; I am fully capable of performing my own review as long as the review itself is not the sole reason for merge.

You have stated yourself that the "ready" tag has no further meaning outside of a comment approval.
It does not mean that the HC or maintainer that looks in to this in the future will bypass their code review as performing a code review before a merge should be mandatory (if it isn't already).

If we are expected to bypass our final review on PR's marked as "ready" we need to communicate this out better so everyone is on the same page.

code/modules/photography/camera/camera.dm
```diff
@@ -15,26 +15,35 @@
...
-	var/default_picture_name
+	var/flash_enabled = TRUE
+	var/start_full = TRUE // does the camera spawn full of film
+
+/obj/item/camera/Initialize()
+	. = ..()
+	if(start_full)
+		pictures_left = pictures_max // future proofed if anyone ever creates a camera with a different max
+	else
+		pictures_left = 0
```
alexkar598: Move this to the initial value instead

code/modules/photography/camera/camera.dm
```diff
@@ -15,26 +15,35 @@
 	flags_1 = CONDUCT_1
 	slot_flags = ITEM_SLOT_NECK
 	materials = list(MAT_METAL = 50, MAT_GLASS = 150)
-	var/flash_enabled = TRUE
+	var/obj/item/disk/holodisk/disk
+	var/pictures_left
```
alexkar598: Cleaner this way if a map wants to have a camera with 1 picture but you can still refill to 10 or whatever
```suggestion
	var/pictures_left = 0
```

alexkar598 (changes requested):
Use defines instead of directly using strings for the mode

Hopekz:
> Use defines instead of directly using strings for the mode

@alexkar598  done

code/modules/photography/camera/camera.dm
```diff
@@ -1,4 +1,5 @@
 
+
```
JamieD1: Random spaces?

JamieD1 (approved):

code/modules/photography/camera/camera.dm
```diff
@@ -50,14 +57,30 @@
 /obj/item/camera/proc/adjust_zoom(mob/user)
 	var/desired_x = input(user, "How high do you want the camera to shoot, between [picture_size_x_min] and [picture_size_x_max]?", "Zoom", picture_size_x) as num
 	var/desired_y = input(user, "How wide do you want the camera to shoot, between [picture_size_y_min] and [picture_size_y_max]?", "Zoom", picture_size_y) as num
-	picture_size_x = min(clamp(desired_x, picture_size_x_min, picture_size_x_max), CAMERA_PICTURE_SIZE_HARD_LIMIT)
-	picture_size_y = min(clamp(desired_y, picture_size_y_min, picture_size_y_max), CAMERA_PICTURE_SIZE_HARD_LIMIT)
+	if(desired_x)
+		picture_size_x = min(clamp(desired_x, picture_size_x_min, picture_size_x_max), CAMERA_PICTURE_SIZE_HARD_LIMIT)
+	if(desired_y)
```
alexkar598: Why the desired_ checks? Unless the user enters 0, this will always run. If typing in 0 to keep the old size is intended behaviour, it is bad behaviour

code/modules/photography/camera/camera.dm
```diff
@@ -50,14 +57,30 @@
...
+			camera_mode = CAMERA_DESCRIPTION
+			to_chat(user, "<span class='notice'>You set the [src] to [camera_mode] mode.</span>")
+			return
+		if(camera_mode == CAMERA_DESCRIPTION)
+			camera_mode = CAMERA_STANDARD
+			to_chat(user, "<span class='notice'>You set the [src] to [camera_mode] mode.</span>")
+			return
+	else
+		to_chat(user, "<span class='notice'>This [src] can only be used in the [camera_mode] mode.</span>") // just in-case somone makes a camera that can only be descriptive
+	return
```
alexkar598: Duplicate and useless code in here
```suggestion
/obj/item/camera/attack_self(mob/user)
	if(can_switch_modes)
		if(camera_mode == CAMERA_STANDARD && can_customise)
			camera_mode = CAMERA_DESCRIPTION
		else if(camera_mode == CAMERA_DESCRIPTION)
			camera_mode = CAMERA_STANDARD
		to_chat(user, "<span class='notice'>You set the [src] to [camera_mode] mode.</span>")
		return

	to_chat(user, "<span class='notice'>This [src] can only be used in the [camera_mode] mode.</span>") // just in-case somone makes a camera that can only be descriptive
```

alexkar598 (changes requested):
Il have to admit im not a fan of making cameras even more useless to just about 90% of people

code/modules/photography/camera/camera.dm
Hopekz: I can't argue with this logic.
Useless checks bad!

Hopekz:
> Il have to admit im not a fan of making cameras even more useless to just about 90% of people

The only solution I can see is to change the photographer trait to "has a camera" and just make everyone have equal proficiency with the tool or removing this as a trait. Being a photographer means you get expertise in this specific area in the game.

With these trait changes, picture-taking remains unaffected but you get all the complicated information on examination similar to someone knowing how to use the tool proficiently.

In my opinion, we should expand this further to other areas of the game like guns because right now you are aware of the exact number of bullets in a nonclear magazine just by looking at it which is weird. This should be reserved for people that are more knowledgeable in these items but that is a separate PR.
Similar to this you're aware of the exact number of pictures left in a camera just by looking at it.
To me, it seems like this is something that proficiency would let you know.

**Also if you're planning to close this PR just request the edit of it being removed because the main focus of this PR is adding the camera modes so it doesn't spam the popup every time you take a picture.**

alexkar598 (approved):

PR #8407, opened 2020-05-06
Makes stat panels a browser (removes lag I guess)

monster860:
Stat panels are a horrible thing that BYOND designed that suck and cause a lot of lag on both the client and server.

This PR moves the client-side part to a browser, and the server-side part to a subsystem, where lag can be eliminated.

Not that it will say "MC not initialized..." until the MC initializes. Don't complain about it because that's how it's supposed to be. IF you *really* need the MC panel during that time you can go to the Debug tab and turn on the legacy MC tab.

:cl: monster860
tweak: Stat panels now use a browser.
/:cl:


ynot01:
what does it look like?

monster860:
image

JamieD1:
Literally the best PR

KingTerrytheTerrible:
this looks like ass lol

PR #8408, opened 2020-05-06
Runechat can be seen while blind

Bibby0110:
This *should* work, but it might not due to planes
#### Changelog

:cl:  
rscadd: Runechat is visible while blind
/:cl:

### DIFF
--- a/code/__DEFINES/layers.dm
+++ b/code/__DEFINES/layers.dm
@@ -78,8 +78,6 @@
 #define MASSIVE_OBJ_LAYER 11
 #define POINT_LAYER 12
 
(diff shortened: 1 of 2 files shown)


monster860 (changes requested):
you might also need to change the plane

Bibby0110:
> you might also need to change the plane

I know
I just don't want to, but it seems I have to

ThatLing:
Why did you ping the Harvard Computer Society?

Bibby0110:
> Why did you ping the Harvard Computer Society?

We need proper code review

PR #8415, opened 2020-05-06
Adds a gang force converter, a compromise for the old implant breaker

Takahiru:
The force converter converts enemies gangsters to your gang, but it doesn't destroy mindshields/work on mindshieldeds.

This promotes less killing on gang rounds and may make people less salty.


Hopekz:
So how does this make people less salty if everyone just piles up on to 1 gang?

The mind breaker was making things unfun because it created this 1 giant mega gang that the crew can't fight and this just does the same thing.

Xoxeyos:
When a gang member gets converted to another gang, there will be hell to pay, this thing better cost alot, and have a long ass cooldown. Then I may change my vote.

Takahiru:
I hate how gang rounds usually end up with a ton of killing and many people taken out of the round unlike any other conversion antag out there.

Hopekz:
> I hate how gang rounds usually end up with a ton of killing and many people taken out of the round unlike any other conversion antag out there.

Maybe the solution to your problem would be to make gangs have less "killing tools" like literal Uzi's and sketchins . 
This just piles every gang in to 1 gang .

Killing will still happen with this except it'll all be directed to the crew instead.

At least with gangs killing each other they take each other out of the shift to not create this 1 mega gang.

Takahiru:
If they had less killing tools they would still kill every other gang member because they literally have to, because the other gang members impede their progress. Blood Brothers can find ways to kill and so can gang members even if you take away all of their tools. Its a conversion gamemode and I don't see why it forces so many people to be forcefully taken out of the round unlike the other ones

Hopekz:
>If they had less killing tools they would still kill every other gang member because they literally have to

Fair enough.
I see your point I just don't think your PR is coming closer to fix the problem you have outlined.
All my opinion of course!

Jarod1200:
what if we added some way for gangs to revive their fallen comrades? like a disposable defibrillator or something

Hopekz (approved):

Hopekz:
> what if we added some way for gangs to revive their fallen comrades? like a disposable defibrillator or something

To be honest I'd rather have gangs inherit leadership in an order set by the first gang leaders or "randomly" instead of having gangs do some magic field revive.

Hopekz:
@swissloaf 
I'll put the ready label on when it is ready 
image

alexkar598 (changes requested):
Move the check for implant in the success thing around like 40 so the message shows as resisting the implant if their mindshielded

MegaEmpirical:
I have two thoughts on this:

1) Gangs fighting and thus killing and reducing their numbers is important for the crew to be able to compete. This may be to potent a buff for gang. I could be wrong though.

2) The sweet release of death is the only way to enjoy a gang round and when I am converted I am instantly hoping to be deconverted or killed as soon as possible. Don't fuck me by having me shift gangs.

SomeguyManperson:
> So how does this make people less salty if everyone just piles up on to 1 gang?
> 
> The mind breaker was making things unfun because it created this 1 giant mega gang that the crew can't fight and this just does the same thing.

it's already possible to turn people via mindshield > implant removal so being able to buy an item to do that but faster doesn't really change much for the whole "everyone being piled into one gang" thing

Takahiru:
I thought the reason it was originally removed is because it destroys securities efforts to deconvert gangsters instantly, security having no real way to prevent people from being converted without just killing them

boodaliboo:
maybe you can make it so it only ressurects dead gang members?

MegaEmpirical:
What part of the only good thing about Gang is dying do you not get?!

Hopekz:
The server poll for this is going to end later today so I'll post the results here.

Seeing that feedback would be key in deciding to merge this since I do not see much positive support for this change at the moment.
Main thing to look for in the results is seeing if there is a huge majority to one side or the other.

Hopekz:
Here are the results of the poll:
Ignoring Neutral I do not think the result is decisive enough to ignore all the negative feedback on here.
I was really hoping for a larger majority to some side.

image

I won't merge this PR but some other maintainer might (code checks out)
Also won't close this PR to give a HC a chance to give their feedback in terms of if we actually need to merge this.

yogstation/code/game/objects/items/implants/implant_gang.dm
```diff
@@ -31,24 +30,28 @@
...
 					success = TRUE	//Was not a gang boss, convert as usual
 					target.mind.remove_antag_datum(/datum/antagonist/gang)
 			else
 				success = TRUE
+			
+			if(!HAS_TRAIT(target, TRAIT_MINDSHIELD))
+				success = TRUE
+			else
+				success = FALSE			
+
```
alexkar598:
```suggestion
			if(G)
				if(!istype(G, /datum/antagonist/gang/boss))
					success = TRUE	//Was not a gang boss, convert as usual
					target.mind.remove_antag_datum(/datum/antagonist/gang)
			else
				success = TRUE
			
			if(HAS_TRAIT(target, TRAIT_MINDSHIELD))
				success = FALSE
			
```

alexkar598 (approved):

alexkar598:
Sure lets try this out

PR #8420, opened 2020-05-07
Fix vacuum being hot

monster860:
Fix mobs caring about temperature if there's no gas to back the temperature up

:cl: monster860
tweak: Body temperature calculations now scale with heat capacity
/:cl:


Bibby0110:
this doesn't fix atmos being hot? just mobs taking damage

PR #8421, opened 2020-05-07
Fixes for the Experimental Tools

ReddicusDragon:
Old Experimental Tools code was boomer code. It would delete the object while placing the other one in your hand. Now, it doesn't do that. It uses one item.

#### Changelog

:cl:  
tweak: rewrites experimental tools
experimental: rewrites experimental tools
/:cl:


Hopekz (changes requested):
Update the 3 instances of the old power tools in code/modules/clothing/outfits/amber.dm

ReddicusDragon:
@Hopekz done

code/game/machinery/doors/airlock.dm
```diff
@@ -1186,14 +1186,15 @@
...
-			time_to_open = 50
-			playsound(src, 'sound/machines/airlock_alien_prying.ogg', 100, TRUE) //is it aliens or just the CE being a dick?
-			prying_so_hard = TRUE
-			if(do_after(user, time_to_open, TRUE, src))
-				open(2)
-				if(density && !open(2))
-					to_chat(user, "<span class='warning'>Despite your attempts, [src] refuses to open.</span>")
-			prying_so_hard = FALSE
+			if (I.tool_behaviour == TOOL_CROWBAR) //we need another check, futureproofing for if/when bettertools actually completely replaces the old jaws
+				time_to_open = 50
```
Hopekz, line 1190: Can you define this variable here instead?
It being defined a few lines above to 5 then changing it when it is only being used in this section.
Its kind of "shitcode" so you'd help our codebase be a better place.

yogstation/code/game/objects/items/tools.dm
```diff
@@ -0,0 +1,128 @@
...
+			to_chat(user, "<span class='notice'>You attach the pry jaws to [src].</span>")
+		attack_verb = list("attacked", "bashed", "battered", "bludgeoned", "whacked")
+		usesound = 'sound/items/jaws_pry.ogg'
+		hitsound = 'sound/items/jaws_pry.ogg'
+		tool_behaviour = TOOL_CROWBAR
+		icon_state = "jaws_pry"
+		update_icon()
+	else
+		to_chat(user,"<span class='warning'>You shouldn't be able to see this! Please contact a coder!</span>")
+		playsound(get_turf(user), 'sound/effects/adminhelp.ogg', 50, 1)
```
Hopekz:
```suggestion
```
Is this necessary? 
The tool can only be in crowbar or wirecutter mode unless an admin messes with the tool_behavior variable.
Even if there is some future PR where we introduce a wizard spell that makes all tools behave as wrenches we shouldn't bwoink people and tell them to contact a coder.

yogstation/code/game/objects/items/tools.dm
```diff
@@ -0,0 +1,128 @@
...
+		else
+			to_chat(user, "<span class='notice'>You attach the screw driver bit to [src].</span>")
+		desc = "A simple powered hand drill. It's fitted with a screw bit."
+		icon_state = "drill_screw"
+		item_state = "drill"
+		tool_behaviour = TOOL_SCREWDRIVER
+		update_icon()
+	else
+		to_chat(user,"<span class='warning'>You shouldn't be able to see this! Please contact a coder!</span>")
+		playsound(get_turf(user), 'sound/effects/adminhelp.ogg', 50, 1)
```
Hopekz:
```suggestion
```
Same as previous point

Hopekz (changes requested):

code/game/machinery/doors/airlock.dm
ReddTheDragon: what do you mean? All I did was modify the existing code

yogstation/code/game/objects/items/tools.dm
ReddicusDragon: I consider it a necessity. Consider it my signature coding style.

code/game/machinery/doors/airlock.dm
Bibby0110: this is intended

Hopekz: I mean make the line here var/time_to_open = 50 then delete the var/time_to_open = 5 a few lines above

Bibby0110: no this is inteded

Bibby0110: that's how the code works.

Bibby0110: The time is = 5 if the door is unpowered
Time is set to 50 if it's powered

yogstation/code/game/objects/items/tools.dm
Hopekz: Your signature style is to have an in-game tool literally play the admin bwoink sound and to tell people to contact a coder when its not behaving exactly how you predicted it would?

Bibby0110: yes it's called a bug

Hopekz: This is introducing a future "bug" for example if someone makes an item that changes the tool behavior to another tool it is quite literally going to be used as a bwoink sound spam tool.

Hopekz: In what world would this even be a "bug" that needs to be fixed?
If something changes the tool behavior to something else it would be a "feature". 

So some future coder is going to laugh it off and delete this anyway.

If you don't want your tool to behave this way then make it check if one of the tool behaviors exist first then return if they both don't exist.
This is "shitcode"

code/game/machinery/doors/airlock.dm
Hopekz: Alright.

yogstation/code/game/objects/items/tools.dm
Bibby0110: No, what are you on about, it doesn't support other tool types, either he adds support for other tools, or this stays.
This should stay either way if he adds support for other tools, if someone sets tool behaviour to something undefined

Hopekz: Then why not just check to see if the tool is in one of the supported modes then return if its not?

Why make the item literally bwoink you and tell you to contact the coder when the coder is just going to delete this part to "fix" it

ReddicusDragon: because that's shitcode. First thing uni courses teach you is to handle errors.

Hopekz: You're not handling errors you're giving a pointless message and sound.

Handling errors would be to check if your tools are in one of the modes then not allowing it to proceed unless it is in one of the modes you support.

Bibby0110: Failing without giving a reason is BAD in this case.

Hopekz: Gamer is wrong.
This is out of place.
Fix it.

Not sure why you have a habit of refusing to change what is needed for me to merge and both times so far it was some trivial reason.

Bibby0110: Why is this out of place?
It's telling the user that their item is bugged

Hopekz (approved):

PR #8433, opened 2020-05-09
Radial menus for mining equipment voucher redeeming

Jumps0:

Replaces the list menu for redeeming mining vouchers with a radial menu. This effects both the regular mining equipment vendor and the free miner mining equipment vendor (The free golem vendor appears to be the exact same as the normal one).

Regular Mining Vendor:
D4QtkmMp0Q

Free Miner Mining Vendor:
a9tsgTfBns

#### Changelog

:cl:  
rscadd: Radial menus for redeeming mining vouchers
/:cl:

### DIFF
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -184,11 +184,20 @@
 	return ..()
 
 /obj/machinery/mineral/equipment_vendor/proc/RedeemVoucher(obj/item/mining_voucher/voucher, mob/redeemer)
-	var/items = list("Survival Capsule and Explorer's Webbing", "Resonator Kit", "Minebot Kit", "Extraction and Rescue Kit", "Crusher Kit", "Mining Conscription Kit")
+	var/items = list(
+		"Survival Capsule and Explorer's Webbing" = image(icon = 'icons/obj/clothing/belts.dmi', icon_state = "explorer1"),
+		"Resonator Kit" = image(icon = 'icons/obj/mining.dmi', icon_state = "resonator"),
+		"Minebot Kit" = image(icon = 'icons/mob/aibots.dmi', icon_state = "mining_drone"),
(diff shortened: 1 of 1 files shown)


Jumps0:
i actually coded something, nice

code/modules/mining/machine_vending.dm
```diff
@@ -318,6 +351,22 @@
...
 	qdel(voucher)
 
+  /*
+   check_menu: Checks if we are allowed to interact with a radial menu
+  
+   Arguments:
+   redeemer The mob interacting with a menu
+   voucher The mining voucher item
+   */
+/obj/machinery/mineral/equipment_vendor/free_miner/check_menu(obj/item/mining_voucher/voucher, mob/living/redeemer)
```
Bibby0110: You don't need this, if you call check_menu it would just go to the normal equipment_vendor check_menu, which is the exact same code

Bibby0110:
You can't stop me Hopek

code/modules/mining/machine_vending.dm
Jumps0: yea i talked with hopek about that and wasn't really sure what to do with it, ill remove it

Bibby0110:
If you're up for the task maybe you could refactor the machines, so that the items list is defined on the machine. Then you don't have to mirror the whole proc for the free miner one, and it would be easier to make new voucher rewards

Bibby0110:
Or maybe not, just took a look and it might not be the easiest

Jumps0:
I didn't understand half the things you just said but I'm glad you feel i'm capable of something. @TheGamerdk

code/modules/mining/machine_vending.dm
Hopekz: Sorry I misunderstood I thought you were going to add additional functionality to the freeminer one vs the regular one otherwise there is no point to make a separate one.

PR #8442, opened 2020-05-09
Pixel shifting fix

ReddicusDragon:
AIs could pixel shift. Blob camera could pixel shift. No more, though. Through a generous application of spaghetti code, I have fixed monster's buggy pixel shifting. **It Just Works™**

Oh, and it also makes pixel shifting reset on player move.

:cl:  
rscadd: Added pixel shifting auto reset on player mob movement
rscdel: Removed AI pixel shifting
/:cl:

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -75,7 +75,11 @@
 
 	if(!mob.Process_Spacemove(direct))
 		return FALSE
-
+	if (!src.prefs.bindings.isheld_key("Ctrl"))
+		if (mob.pixel_x > 0 || mob.pixel_x < 0)
+			mob.pixel_x = 0
+		if (mob.pixel_y > 0 || mob.pixel_y < 0)
+			mob.pixel_y = 0
 	var/handled = SEND_SIGNAL(L, COMSIG_PROCESS_MOVE, direct) //yogs start - movement components
 	if(handled)
 		return FALSE//yogs end
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -669,6 +669,8 @@
 
 /mob/verb/eastshift()
 	set hidden = TRUE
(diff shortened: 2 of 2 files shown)


code/modules/mob/mob_movement.dm
```diff
@@ -75,7 +75,11 @@
 
 	if(!mob.Process_Spacemove(direct))
 		return FALSE
-
+	if (!src.prefs.bindings.isheld_key("Ctrl"))
```
Bibby0110, line 78: `if (!src.prefs.bindings.isheld_key("Ctrl") && !(mob.datum_flags & DF_VAR_EDITED))`
i want pixel shifted Ian thank you very much

Bibby0110 (commented):
uwu

code/modules/mob/mob_movement.dm
monster860: thats shit code gamer

Bibby0110: nah

azzzertyy:
why is either of these pixel shifting a bad thing?

ReddicusDragon:
@azzzertyy basically pixel shifting not auto resetting gives a combat advantage and ais shouldn’t be able to pixelshift because that’s just shit code

azzzertyy:
yeah but the hitboxes are exactly the same with these, no ones meleeing a blob or ai, and if so, moving your mouse to the top right corner of the pixel isnt very difficult

Hopekz:
FYI; It was specifically made not to reset VS the citadel port.

ReddicusDragon:
@monster860 opinion?

monster860 (changes requested):
Can you just go to the Citadel version and copy the code from there, as that one will only reset the pixel_x/y if actual pixel shifting was done, unlike this one which is aways resetting the pixel_x, and not necessarily to the correct value

PR #8465, opened 2020-05-10
Fixes Minesweeper exploit

JamieD1:
### Intent of your Pull Request

I think this has been going on for long enough,

I watched someone get about 20 rocket launchers from using this exploit whilst it was emagged.

#### Changelog

:cl:  
rscdel: Removed same board from minesweeper
/:cl:

PR #8468, opened 2020-05-10
Removes power *gaming* (playing the same minesweeper board no longer grants prizes)

alexkar598:
See title.

View #8465 for why

### DIFF
--- a/yogstation/code/game/machinery/computer/arcade.dm
+++ b/yogstation/code/game/machinery/computer/arcade.dm
@@ -37,6 +37,7 @@
 	var/square6
 	var/square7
 	var/square8
+	var/sameboard = FALSE
 
 /obj/machinery/computer/arcade/minesweeper/Initialize()
 	squareflag = "[icon2html('yogstation/icons/arcade/minesweeper_tiles.dmi', world, "minesweeper_flag")]"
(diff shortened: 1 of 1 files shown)


ynot01:
bad pr

alexkar598:
ynot dontremovemypowergame

This is a person that downvoted the katana removal, just sayin

maxion12345:
Kinda needed

Takahiru:
no pls

Hopekz:
Why not just make the emagged one not be repeatable.
Why destroy the whole harmless "fun" feature that everyone loves along with it?

JamieD1:
> Why not just make the emagged one not be repeatable.
> Why destroy the whole harmless "fun" feature that everyone loves along with it?

How about we just make a free toy dispenser then that poops out a random toy every two minutes?

Or costs $5

Hopekz:
> > Why not just make the emagged one not be repeatable.
> > Why destroy the whole harmless "fun" feature that everyone loves along with it?
> 
> How about we just make a free toy dispenser then that poops out a random toy every two minutes?
> 
> Or costs $5

No compromise is needed here. Do non emagged toys offer some kind of advantage that I'm not aware of? If so, you should make a PR to remove any OP toys (maybe some scary plushies) that you're thinking of instead of the removal of a fun feature.

Why would you take out the assistant from participating? They are the main target audience for this fun arcade machine exploit feature and they only get 5 credits per paycheck.

alexkar598:
Because its literally a free toy dispenser which is absolutly bullshit.

JamieD1:
> > > Why not just make the emagged one not be repeatable.
> > > Why destroy the whole harmless "fun" feature that everyone loves along with it?
> > 
> > 
> > How about we just make a free toy dispenser then that poops out a random toy every two minutes?
> > Or costs $5
> 
> No compromise is needed here. Do non emagged toys offer some kind of advantage that I'm not aware of? If so, you should make a PR to remove any OP toys (maybe some scary plushies) that you're thinking of instead of the removal of a fun feature.
> 
> Why would you take out the assistant from participating? They are the main target audience for this fun arcade machine exploit feature and they only get 5 credits per paycheck.

Might as well just spawn 100 toys then in tool storage, defeats the whole point of arcade games if there is no effort.

Hopekz:
> Because its literally a free toy dispenser which is absolutly bullshit.

I guess the main disconnect we're having is this question:
**"Why is having a free toy dispenser a bad thing?"**

At the moment I fail to see how having a lot of toys is a bad thing. I see it as a "fun" part of our game. It seems you are making this change for the emagged prizes while also effecting the regular game as a side effect when you could be specifically targeting the machine while it is emagged.

JamieD1:
> > Because its literally a free toy dispenser which is absolutly bullshit.
> 
> I guess the main disconnect we're having is this question:
> **"Why is having a free toy dispenser a bad thing?"**
> 
> At the moment I fail to see how having a lot of toys is a bad thing. I see it as a "fun" part of our game. It seems you are making this change for the emagged prizes while also effecting the regular game as a side effect when you could be specifically targeting the machine while it is emagged.

Whats the point of arcade machines if you get the toys for free.

boodaliboo:
> > > Because its literally a free toy dispenser which is absolutly bullshit.
> > 
> > 
> > I guess the main disconnect we're having is this question:
> > **"Why is having a free toy dispenser a bad thing?"**
> > At the moment I fail to see how having a lot of toys is a bad thing. I see it as a "fun" part of our game. It seems you are making this change for the emagged prizes while also effecting the regular game as a side effect when you could be specifically targeting the machine while it is emagged.
> 
> Whats the point of arcade machines if you get the toys for free.

to play the games on it

alexkar598:
Then you mind this change, you can still play the same board, you get no toys. But that doesn't matter since you are playing them for fun not for the prize. Right? rrright?

boodaliboo:
> Then you mind this change, you can still play the same board, you get no toys. But that doesn't matter since you are playing them for fun not for the prize. Right? rrright?

i cant play mindsweeper i havent figured out how

maxion12345:
The main issue here is that emagged minesweepers is an easy way to get access to literally infinite amounts of high explosive as a traitor. Set one up in maint and you could arm your own nukeops team in five minutes.

alexkar598:
If you wanna learn the game, you still game, the option is still there, but if you want loot, you need to actually play the game

PR #8469, opened 2020-05-10
Adds welsh accent

rob-a-bolton:

Adds Welsh accent. Change has not been tested beyond being selectable in chargen (due to issues getting yogstation running locally), but JSON validates fine.  
All single l's become double l's, along with a few substitutions (e.g. here/ear/year = year, among others)  
Reasoning: We have scottish, I don't see any harm in having welsh.  

#### Changelog

:cl:  
rscadd: Added Welsh accent to character creation
/:cl:

### DIFF
--- a/strings/accents/accent_welsh.json
+++ b/strings/accents/accent_welsh.json
@@ -1,0 +1,22 @@
+{
+	"Welsh": {
+		"l": "ll",
+		"\bwhere\b": "whereby",
+		"\bnow\b": "now in a minute",
+		"\bnice\b": "tidy",
+		"\bbuddy\b": "butt",
+		"\bhere\b": "year",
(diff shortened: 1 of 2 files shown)


MegaEmpirical:
Welsh isn't an accent, it's what happens after you have a stroke.

Ktlwjec1:
Changelog needs to be something like this, so players who read it understand what you've added:
rscadd: Added a Welsh accent.

rob-a-bolton:
> Changelog needs to be something like this, so players who read it understand what you've added:
> rscadd: Added a Welsh accent.

My bad, fixed

alexkar598:
Im going to say no to adding an accent with literally 10 words

PR #8498, opened 2020-05-14
adds oxygen to tower caps

boodaliboo:
its a tree kinda
#### Changelog

:cl:  
rscadd: Adds oxygen to tower caps
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/towercap.dm
+++ b/code/modules/hydroponics/grown/towercap.dm
@@ -16,6 +16,8 @@
 	icon_dead = "towercap-dead"
 	genes = list(/datum/plant_gene/trait/plant_type/fungal_metabolism)
 	mutatelist = list(/obj/item/seeds/tower/steel)
+	reagents_add = list(/datum/reagent/oxygen = 0.1)
+
 
 /obj/item/seeds/tower/steel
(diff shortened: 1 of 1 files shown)


Moja10:
When botany mains make PRs like this, the first thing that should be done is looking up every chem that uses oxygen as part of a recipe and see what chems can now be made when combined with already existing chem from other plants. Then you will know their true intentions.

code/modules/hydroponics/grown/towercap.dm
```diff
@@ -16,6 +16,8 @@
 	icon_dead = "towercap-dead"
 	genes = list(/datum/plant_gene/trait/plant_type/fungal_metabolism)
 	mutatelist = list(/obj/item/seeds/tower/steel)
+	reagents_add = list(/datum/reagent/oxygen = 0.1)
+
```
JamieD1, line 20: Can you remove the extra random line?

JamieD1 (changes requested):

SomeguyManperson:
importantly this allows botanists to fabricate their own saltpetre since nitrogen is somewhat ample in supply from canned air

alexkar598:
Id rather not powercreepe botany any further.

PR #8501, opened 2020-05-14
Adds a mass driver to the Syndicate Infiltrator

maxion12345:

Adds a mass driver to the Nukeops ship for shooting things into the station. Bombs, operatives, hoards of sentient space carp, whatever.

#### Changelog

:cl:  
rscadd: Added a mass driver to the Nuclear Operative ship. 
tweak: Moved some stuff around in the Nuclear Operative ship airlock.
/:cl:

### DIFF
--- a/_maps/shuttles/infiltrator_basic.dmm
+++ b/_maps/shuttles/infiltrator_basic.dmm
@@ -376,15 +376,6 @@
 	},
 /turf/open/floor/plasteel/dark,
 /area/shuttle/syndicate/airlock)
-"aQ" = (
-/obj/structure/table/reinforced,
-/obj/item/storage/toolbox/syndicate,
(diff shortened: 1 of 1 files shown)


Ktlwjec1 (changes requested):
Table with the toolbox you've copy/pasted onto that tile twice, so there's 2 copies of each thing. Also make the flooring dark plasteel with only neutral corner decals - the red and stripe decals don't look as good for that tile.
There's the issue (shown in the image) that the wall curves when in-game, so the button looks off. I would personally add it to the wall with the button on it, or move it up a little bit. Also the floor decal (bot_white) on that tile and the two above it need removing.
image

Ktlwjec1 (approved):
a beautiful pr

PR #8517, opened 2020-05-15
replace zero in the wizard ruin with rotciv

boodaliboo:
the title


Readystorm546:
Imagine being lazy and not making another one

JamieD1 (approved):

boodaliboo:
> Imagine being lazy and not making another one

yes

Ktlwjec1 (approved):
ye ok

PR #8533, opened 2020-05-16
Makes the bucket on the ashwalker ruin wooden

jcatjmeow:

The bucket in the ashwalker ruin is now wooden. 

This changes basically nothing other than aesthetics. And it's flammable I guess. But I like it, so

Also they get a rake :)

#### Changelog

:cl:  
tweak: the ashwalker bucket is now wooden instead of plastic. And they get a rake.

/:cl:


jcatjmeow:
if you are wondering how they would get wood, they already spawn with tower cap seeds and the like. Any extra buckets you make as ashwalker will be wooden already.

JamieD1 (approved):
Careful of the bonfire, these are flamable.

jcatjmeow:
> Careful of the bonfire, these are flamable.

Yep! But like, just don't throw it in the fire. And if you do, just grow some tower caps and make a new one

SomeguyManperson (approved):

PR #8549, opened 2020-05-17
Fixes Bloodcult Softlocking

### LINKED ISSUE #8168: Cult game mode softlock
If the sacrifice target for the blood cult is the chaplain and he uses his holypara and gets dusted before he gets sacrificed; the game is now soft locked.

Cult can't win and you can't even bring the chaplain back using admin powers to get him sacrificed because it won't count.

MacHac:
If the bloodcult's target gets gibbed or can't be sacrificed for any reason, the cult gets exactly one mulligan target.  If they blow that too, they just lose.

#### Why?
Letting a gamemode get softlocked is bad, but so is instantly failing the cult when the chaplain gets bored and becomes a borg.  I think this is a fair compromise.

#### Caveats
- Only works for roundstart cultist teams, admin-created cultists can't end the round this way.
- Anything that removes the chaplain's mind from his body counts as failure to sac, including beheadings.  Some of these are technically reversible through surgery, but nobody ever does it.
- To avoid exploits, the mulligan target is _required_ to be unconvertable.  If the cult does enough murdering before they lose their target, they may not be able to get a new one, meaning they lose.

Maintainer edit:
Fixes #8168

#### Changelog

:cl:  
tweak: Blood cults may now be assigned a new target if they lose their old one.
/:cl:


Takahiru:
can't you sacrifice a body without the mind? I thought the only problem was during cremation, not when they died.. The cult can easily use the ghosting rune to find where the target is if they are dead and without a mind

MacHac:
Sacrifice objective code works by checking the mind.  I think you can sacrifice a brainless body, but it won't make the objective complete.  Heck, sacrificing a catonic target might not count even if the brain's still there.

Theoretically, if you have just the brain you can pop it in a body and sacrifice that.  Unfortunately, almost nobody ever does this and the chaplain can thwart it by entering a posibrain or something.

SomeguyManperson:
iirc minds are stored in the last used body of their character even if the body is catatonic, taking a ghost role should create a new mind

MacHac:
That may be the case, I can't check right now.  In any case, my main goal is to let the cult recover gracefully from the chaplain getting gibbed/crushed/burned/eaten/blown up.  Preventing degenerate cases like brainhunting is a bonus.

Hopekz:
Fixes #8168

JamieD1 (approved):

SomeguyManperson (approved):

PR #8554, opened 2020-05-18
Standardises null

Hopekz:
We shouldn't do it both ways.
Pick one and stick to it.

### DIFF
--- a/code/datums/beam.dm
+++ b/code/datums/beam.dm
@@ -66,7 +66,7 @@
 	timing_id = addtimer(CALLBACK(src, .proc/recalculate), time, TIMER_STOPPABLE)
 
 /datum/beam/proc/after_calculate()
-	if((sleep_time == null) || finished)	//Does not automatically recalculate.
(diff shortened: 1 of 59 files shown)


JamieD1 (approved):
Is this standards I see?

code/datums/beam.dm
```diff
@@ -66,7 +66,7 @@
 	timing_id = addtimer(CALLBACK(src, .proc/recalculate), time, TIMER_STOPPABLE)
 
 /datum/beam/proc/after_calculate()
-	if((sleep_time == null) || finished)	//Does not automatically recalculate.
+	if((!sleep_time) || finished)	//Does not automatically recalculate.
 		return
 	if(isnull(timing_id))
```
JamieD1, line 71: Alot of isnull(

Can you change them too?!? 👍

JamieD1 (changes requested):

alexkar598:
!X is not the same as X == null

!X:
0: true
"yeet": false
null: true

X == null:
0: false
"yeet": false
null: true

PR #8555, opened 2020-05-18
Removes shitcode operator

Hopekz:
This has no reason to exist.

### DIFF
--- a/code/modules/admin/verbs/massmodvar.dm
+++ b/code/modules/admin/verbs/massmodvar.dm
@@ -81,7 +81,7 @@
 	var/new_value = value["value"]
 	var/class = value["class"]
 
-	if(!class || !new_value == null && class != VV_NULL)
+	if(!class || class != VV_NULL)
 		return
 
 	if (class == VV_MESSAGE)


JamieD1 (approved):

alexkar598 (changes requested):
Fix the operator instead of removing it, its there to ensure you can cancel out of a mass modify

Also your change would break mass modify

Hopekz:
> Also your change would break mass modify

!new_value == null can never be true.
The fix is to remove it.

Because of !new_value == null you can never satisfy condition nr. 2 in the argument (circled)
image

This whole if statement can be simplified down to !class if you want because the second portion literally cannot happen at the moment before this removal.

Hopekz:
> > Also your change would break mass modify
> 
> !new_value == null can never be true.
> The fix is to remove it.
> 
> Because of !new_value == null you can never satisfy condition nr. 2 in the argument (circled)
> image

This whole if statement can be simplified down to !class if you want because the second portion literally cannot happen.

alexkar598:
Your PR makes it so that any mass modify that tries to set it to something thats is not VV_NULL will fail

PR #8564, opened 2020-05-18
Adds russian beepsky

monster860:
heres what it sounds like:

https://cdn.discordapp.com/attachments/134720091576205312/711868167433093120/insult.ogg

:cl: monster860
add: Beepsky now has a 5% chance of being imported from Russia
/:cl:


athena413:
wait this means beepsky breaks rule 0.6, meaning we need to ban beepsky

Bibby0110:
nooooo

Marmio64:
is this just a 5% chance to happen every time a insult is played or is it a 5% chance for beepsky to always use this instead of the normal one

monster860:
5% chance of beepsky being russian when spawned, and then always using the russian lines

Hopekz:
Beepsky seems more German to me but then again HoS has the whole soviet thing going on.

Hopekz (approved):

PR #8572, opened 2020-05-19
Auto close stale PRs

alexkar598:
see title


alexkar598:
Only PRs that have Awaiting - Action - Author are subject to this

yacabo:
Does it check if the PR gets a commit which would effectively mean it's no longer stale?

alexkar598:
Yes

PR #8579, opened 2020-05-19
bromine does brain damage now

boodaliboo:
image
bromine causes brain damage irl
tell me if you think it should do less brain damage
#### Changelog

:cl:  
tweak: bromine now does a minor ammount of brain damage
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -1377,6 +1377,10 @@
 	color = "#C8A5DC"
 	taste_description = "chemicals"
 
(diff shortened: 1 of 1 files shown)


Tipy1802:
Not sure if it's a good idea since bromine is a common disease cure (in game)

Jarod1200:
Bromine is easy to get for disease cures and Mannitol is also easy to get so i don't really see an issue with this. Also if they cure one guy with bromine they can make a vaccine if mannitol is that hard to get.

Marmio64:
virus powercreep

JamieD1 (approved):

alexkar598:
It works like this IRL is a bad reason to change things

Real life is boring, we are making a game.

PR #8598, opened 2020-05-20
Rebalances Security Grenades

MacHac:
Flashbangs are currently very strong and tear gas is very weak.  I've tried to balance the two with feedback from people who play more security than I do.

#### Flashbangs
Flashbangs are now more forgiving if you have only one type of protection (eye or ear.)  Being exposed to a flashbang without ear protection will cause a few seconds of confusion after the stun ends.  I'm trying to narrow its niche to opening an assault and forcing people out of cover, just like real flashbangs.
##### No protection
- Maximum stun length: 15s (down from 20s)
- Minimum stun length: 6s (unchanged)

##### Eye protection only
- Maximum stun length: 5s (down from 20s)
- Minimum stun length: 3s (down from 6s)
- Confusion: Up to 20s, decreases asymptotically with distance

##### Ear protection only
- Maximum stun length: 5s (down from 20s)
- Minimum stun length: 3s (down from 6s)
- Blindness: 3s (unchanged)

##### Full protection
- No effect (unchanged)

#### Tear Gas
In addition to its old effects, tear gas now lingers in the system causing stamina damage and coughing.  The cloud is unlikely to stamcrit directly unless the victim doesn't try to leave at all, but more than a few seconds of exposure will cause people to drop their items and move more slowly for a while.  The caveat is that officers are vulnerable to the smoke as well; Thus, the grenades are best for weakening fortified antagonist bases and covering retreats.

I'll consider adding tear gas to the secvend if enough people think it's a good idea, vote now in the comments.

* Even though I'm buffing condensed capsaicin, pepper spray delivers so little that it's almost unaffected.

### Why?
The flashbang is the lord and savior of medium range combat; if you can count to four you hardly even have to click on the enemy.  The obnoxious stun length is made more obnoxious by the fact that you have to get two separate counter items to survive, one of which is semi-obscure and not used for anything else.  There may be room in your galaxy-sized brain for the location of every bowman on the station, but newer antags don't deserve instant death just because someone managed to land a click in their zip code.

By contrast, tear gas is hardly worth the inventory space.  Masks block it partially and any sort of glasses block it completely, so it's essentially a glorified smoke grenade.  It's hard to use effectively and easy to counter.

With these changes, both grenades get a unique and useful tactical niche without being overbearing.  The flashbang disrupts any waiting enemies as you breach a room and forces people in cover to pull back; Tear gas softens up fortified enemies and creates cover for movement or retreat.  Both of these roles roughly match their real-life analogues, though that alone isn't justification for a balance decision.

I've tuned all the numbers based on what 'felt right' compared to normal combat; if something seems too high or too low, I'm open to change.

#### Changelog

:cl:  
tweak: The Space Geneva Convention has taken action against grenades of mass destruction; Flashbang payloads have been adjusted down in compliance.
tweak: NT officials noticed that the shipments marked 'tear gas' contained nothing but red food coloring and have found a new supplier.
/:cl:

### DIFF
--- a/code/game/objects/items/grenades/flashbang.dm
+++ b/code/game/objects/items/grenades/flashbang.dm
@@ -23,7 +23,6 @@
 		return
 	M.show_message("<span class='userdanger'>BANG</span>", 2)
 	var/distance = max(0,get_dist(get_turf(src),T))
-	message_admins(distance)
 	if(!distance || loc == M || loc == M.loc)	//Stop allahu akbarring rooms with this.
 		M.Paralyze(200)
 		M.soundbang_act(1, 20, 10, 15)
--- a/code/game/objects/items/grenades/flashbang.dm
+++ b/code/game/objects/items/grenades/flashbang.dm
@@ -21,16 +21,20 @@
 /obj/item/grenade/flashbang/proc/bang(turf/T , mob/living/M)
 	if(M.stat == DEAD)	//They're dead!
 		return
-	M.show_message("<span class='warning'>BANG</span>", 2)
+	M.show_message("<span class='userdanger'>BANG</span>", 2)
 	var/distance = max(0,get_dist(get_turf(src),T))
-
-//Flash
-	if(M.flash_act(affect_silicon = 1))
-		M.Paralyze(max(200/max(1,distance), 60))
-//Bang
+	message_admins(distance)
 	if(!distance || loc == M || loc == M.loc)	//Stop allahu akbarring rooms with this.
 		M.Paralyze(200)
-		M.soundbang_act(1, 200, 10, 15)
+		M.soundbang_act(1, 20, 10, 15)
+		return
 
-	else
-		M.soundbang_act(1, max(200/max(1,distance), 60), rand(0, 5))
+	var/flashed = M.flash_act(affect_silicon = 1)
+	var/banged = M.soundbang_act(1, 20/max(1,distance), rand(0, 5))
+
+	// If missing two resists
+	if(flashed > 0 && banged > 0)
+		M.Paralyze(max(150/max(1,distance), 60))
+	// If missing one resist
+	else if (flashed > 0 || banged > 0)
+		M.Paralyze(max(50/max(1, distance), 30))
--- a/code/game/objects/items/grenades/flashbang.dm
+++ b/code/game/objects/items/grenades/flashbang.dm
@@ -32,8 +32,8 @@
 	var/banged = M.soundbang_act(1, 20/max(1,distance), rand(0, 5))
 
 	// If missing two resists
-	if(flashed > 0 && banged > 0)
+	if(flashed && banged)
 		M.Paralyze(max(150/max(1,distance), 60))
 	// If missing one resist
-	else if (flashed > 0 || banged > 0)
+	else if (flashed || banged)
 		M.Paralyze(max(50/max(1, distance), 30))
--- a/code/modules/mob/living/carbon/alien/alien_defense.dm
+++ b/code/modules/mob/living/carbon/alien/alien_defense.dm
@@ -123,7 +123,7 @@ In all, this is a lot like the monkey code. /N
 				Unconscious(20)
 			adjustEarDamage(15,60)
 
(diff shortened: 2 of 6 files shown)


code/game/objects/items/grenades/flashbang.dm
```diff
@@ -21,16 +21,19 @@
...
 	var/distance = max(0,get_dist(get_turf(src),T))
-
-//Flash
-	if(M.flash_act(affect_silicon = 1))
-		M.Paralyze(max(200/max(1,distance), 60))
-//Bang
 	if(!distance || loc == M || loc == M.loc)	//Stop allahu akbarring rooms with this.
 		M.Paralyze(200)
-		M.soundbang_act(1, 200, 10, 15)
+		M.soundbang_act(1, 20, 10, 15)
```
ghost, line 28: This means if you're on top of the flashbang when it goes off, you will be stunned for far less.

This lets you be a flashbang suicide bomber again, basically.

alexkar598 (changes requested):
Atomize your prs, move the tear gas changes to another pr so it can be discussed about on its own

code/game/objects/items/grenades/flashbang.dm
MacHac: I actually changed the definition of soundbang_act.  Paralyze() works in deciseconds, but confusion works in seconds.  You can see in the code that I changed every other use of the function too.

Flashbang suicide bombing is degenerate but 200 seconds of confusion is a step too far.  Of course, since it completely overlays the stun, it'll only affect someone with stun reduction.

MacHac:
I bundled the two over balance concerns.  Some people I talked to encouraged me to make a compensatory buff, but any security change (buff or nerf) tends to spawn 100-comment 'insightful discussions'.  I hoped that doing the two under the same PR would paint the change as being tit-for-tat rather than a salty security main/ided PR.

Besides, 'security grenades' is almost atomic.  Call it a molecular PR.  I'm willing to split it if you think it's necessary, but first understand why I'm not eager to do so.

SomeguyManperson:
confusion is a mob stat that's handled on life() so 1 confusion actually lasts ~2 seconds

athena413:
ohhhh my god yes

MegaEmpirical:
Tear gas is easily avoided unfortunately, though the buff is nice.

I think the 3 second minimum stun is too low, the current max is admittedly high.

MacHac:
The tear gas is really an area denial tool more than an offensive weapon; it's not hard to stay clear, but if you grit your teeth and try to run through you'll end up in damage slowdown.  Being easy to avoid works in your favor with tactical use: chuck one into the gang base then grab them as they come running out.

The three second minimum stun is what you get if you're 5 tiles from the flashbang when it goes IE there are 4 empty tiles of space between you and it.  Security getting a free followup stun even at that range is most of the problem with flashbangs in the first place.  

Flashbangs as a primary capture tool is the behavior I want to discourage.  Their intended niche is to force people out of cover and disrupt enemy defenses, not be a single use overpowered stun baton.

MegaEmpirical:
Ahh I see. Well I'm definitely interested in testing this

Firewolf34:
I like your Changelog entries.

code/game/objects/items/grenades/flashbang.dm
```diff
@@ -21,16 +21,19 @@
...
+	var/flashed = M.flash_act(affect_silicon = 1)
+	var/banged = M.soundbang_act(1, 20/max(1,distance), rand(0, 5))
 
-	else
-		M.soundbang_act(1, max(200/max(1,distance), 60), rand(0, 5))
+	// If missing two resists
+	if(flashed > 0 && banged > 0)
+		M.Paralyze(max(150/max(1,distance), 60))
+	// If missing one resist
+	else if (flashed > 0 || banged > 0)
```
alexkar598:
```suggestion
	if(flashed && banged)
		M.Paralyze(max(150/max(1,distance), 60))
	// If missing one resist
	else if (flashed || banged)
```

alexkar598 (changes requested):

alexkar598 (approved):

PR #8608, opened 2020-05-20
Dead cyborgs can't self destruct

MacHac:
Tag this ided.

Cyborgs can leave the confirmation prompt for self destructing open as long as they want, letting them blow up at will.  I'm willing to accept that in principle; the confirm box is mostly just misclick prevention.  The issue is that the button still works after the borg dies.

You can't start the process while dead, so I'm calling this an oversight.  Borgs as permanent landmines sounds fun until it happens to you.

#### Changelog
:cl:  
bugfix: Cyborgs can no longer exploit the alert() function to self-destruct from beyond the grave.
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/silicon/robot/robot.dm
+++ b/yogstation/code/modules/mob/living/silicon/robot/robot.dm
@@ -66,6 +66,10 @@
 	if(alert("WARNING: Are you sure you wish to self-destruct? This action cannot be undone!",,"Yes","No") != "Yes")
 		return
 
+	if(usr.stat == DEAD)
+		to_chat(usr, "<span class='danger'>You are already dead.</span>")
+		return //won't work if dead
+
 	var/turf/T = get_turf(usr)
 	message_admins("<span class='notice'>[ADMIN_LOOKUPFLW(usr)] detonated themselves at [ADMIN_VERBOSEJMP(T)]!</span>")
 	log_game("\<span class='notice'>[key_name(usr)] detonated themselves!</span>")


SomeguyManperson (approved):

yogstation/code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -66,6 +66,10 @@
 	if(alert("WARNING: Are you sure you wish to self-destruct? This action cannot be undone!",,"Yes","No") != "Yes")
 		return
 
+	if(usr.stat == DEAD)
+		to_chat(usr, "<span class='danger'>You are already dead.</span>")
+		return //won't work if dead
```
JamieD1, line 71: Seems silly to check for dead twice, why not just move the alert above the check?

JamieD1 (changes requested):

ReddicusDragon:
bad pr it nerfs borgs

yogstation/code/modules/mob/living/silicon/robot/robot.dm
MacHac: If I removed the top check, the alert box would show up even if the borg was dead, it just wouldn't do anything.  By checking before, we avoid asking people to confirm actions that we already know they can't do.  I think it's cleaner, and the performance impact is hilariously negligible.

Hopekz (approved):

Marmio64:
ANOTHER STUNNING SUCCESS

ReddicusDragon:
borg mains everywhere hate him

PR #8609, opened 2020-05-21
Overhauling the Seed Vault

MrHorizons:
Initially, I just wanted to add satchels to the seed vault, but as I rooted through the map file, I realized that it could use a more exhaustive overhaul. After some time of work, this is the result.

<details>
  <summary>The original seed vault map</summary>

  image
</details>

<details>
  <summary>The new seed vault map</summary>

  image
</details>

The most notable changes are:

- **Decreasing the number of spawners to two**: The original seed vault map had 11 trays scattered around the room, which while not distributable evenly, would be enough for two people to use. But four is too many, and even botany has a maximum of three. Besides, it's rare for there to be more than two people on the spawn.
- **Increasing the number of hydroponic trays to 12**: This ties in to the previous point, but is more to make things even.
- **Removing the second room** and **moving beekeeping supplies out in the open**: In my experience as a botanist, I've found it easier to have the DNA manipulator/disk compartmentalizer in the main room not only for convenience of distance, but also so I wouldn't be screwed by an antag popping out of maint and capping me. Of course, the latter is essentially a nonexistent threat on the seed vault, but the former is still a consideration. The other supplies formerly inside there have been spread out.
- **Concentrating supplies**: This one's probably the vaguest of all of these entries, but essentially, I've made it so the spawn room has two lockers with supplies needed or relevant to the direct manipulation of plants. For the two items that could use spares (hatchets and floral somatorays), I've put two of each on a few tables.
<details>
  <summary>Locker Contents</summary>

  - White jumpsuit
  - Satchel
  - Botanic gloves
  - Plant analyzer
  - Portable seed extractor
  - Plant bag
  - Cultivator
  - Spade
  - Hatchet
  - Floral somatoray
</details>

- **Adding a second smartfridge, grinder, and DNA manipulator**: I figure these items are the ones that would often end up being used "simultaneously", so duplicated them for convenience.
- **Changed bluespace beakers to XL beakers**: Bluespace beakers are unnecessarily large for the quantity of plants (and therefore, chemicals) being worked with, and the chem dispenser is less than a screen away.

#### Changelog

:cl:  
rscadd: Overhauled the seed vault, now intended for two people.
/:cl:

### DIFF
--- a/_maps/RandomRuins/LavaRuins/lavaland_surface_seed_vault.dmm
+++ b/_maps/RandomRuins/LavaRuins/lavaland_surface_seed_vault.dmm
@@ -5,32 +5,38 @@
 "b" = (
 /turf/open/floor/plating/asteroid/basalt/lava_land_surface,
 /area/lavaland/surface/outdoors)
-"d" = (
-/obj/structure/table/wood,
-/obj/item/lighter,
-/obj/item/lighter,
-/obj/item/storage/box/fancy/rollingpapers,
-/obj/item/storage/box/fancy/rollingpapers,
-/obj/item/storage/box/fancy/rollingpapers,
-/obj/item/storage/box/fancy/rollingpapers,
+"c" = (
+/obj/item/reagent_containers/glass/beaker/plastic,
+/obj/item/reagent_containers/glass/beaker/plastic,
+/obj/structure/table/glass,
+/obj/machinery/reagentgrinder{
+	pixel_y = 5
+	},
+/obj/item/reagent_containers/glass/beaker/plastic,
 /turf/open/floor/plasteel/freezer,
 /area/ruin/powered/seedvault)
+"d" = (
+/turf/open/floor/plating/asteroid/basalt/lava_land_surface,
+/area/lavaland/surface)
 "e" = (
-/obj/structure/table/wood,
-/obj/item/storage/box/disks_plantgene,
-/turf/open/floor/plasteel/freezer,
+/obj/machinery/light{
+	brightness = 4;
+	dir = 8
+	},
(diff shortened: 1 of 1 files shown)


Marmio64:
you have 4 spawners because usually the people get themselves killed retardedly like going outside and suffocating to death or trying to get cool mutations from the floral somotorays

MrHorizons:
> you have 4 spawners because usually the people get themselves killed retardedly like going outside and suffocating to death or trying to get cool mutations from the floral somotorays

Well, I suppose there's nothing I can do about stupidity. Though I figure we should at least try having only two spawners, and if that doesn't work, I can always redo it with four.

Ktlwjec1 (changes requested):
- It's a little too symmetrical, which is an issue if you have square-shaped ruins, but I think you're going to have that issue no matter what. Play around though and see if there's a slightly different design you like, but if not it's absolutely no issue.
- Remove the double rwalls (top and bottom line of the ruin).
- Only 1 disposals bin is needed for such a small room.
- There's a free miner engineer spawn in the room, unsure if it's intentional but that needs removing either way.
- Reduce it to three spawners instead, see how that plays out in-game. Add another closet + beekeeper suit due to this.
- Remove the vapes, much prefer them to make their own smokes.
- Under the chem dispenser is a pipe trunk that needs removing.
- Remove one of the smartfridges.
- Move both the seed spawners into one crate, and add a mix of spare clothing and objects (add another plant bag at least) into the other crate (helps with the symmetry thing too).

MrHorizons:
> * Remove the vapes, much prefer them to make their own smokes.

Well, my intention with the vapes was to have them be able to grind up the reagents and put them in there, but I suppose heavy use of rolling papers makes more sense with the rest of it.

PR #8619, opened 2020-05-21
[ADMIN QOL] The Nuclear Device Code is now set automatically to a random value.

Nickvr628:
Whenever an admin wants to have the nuke be used, they have to manually edit the code to a random value. This is cumbersome, and since there is already a proc to randomize the code anyway, I just made it create one at roundstart. The code is a random 5 digit code like normal.

The odds of someone randomly guessing the code are essentially impossible, so that is not a concern.

All this really does is reduce a step for admins to let the captain/crew scuttle the station, because it is a hassle to manually varedit the self destruct device and change it to a 5 digit value and we might as well streamline it. This also lets admins without the power to edit variables approve self destruct requests, which removes cases of valid requests going unanswered because all the staff online are unable to set the code.

This does not break nuke ops, I checked, and it actually removes some back and forth shitcode by automatically giving the self destruct device a code regardless of nuke ops triggering or not.

 
:cl:  
tweak: Instead of admins needing to manually set it to use it, the station self destruct device now has a random passcode. Admins can quickly check and send the code to the station or ERTs.
/:cl:

### DIFF
--- a/code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
+++ b/code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
@@ -723,9 +723,10 @@ This is here to make the tiles around the station mininuke change when it's arme
 //==========CENTCOM NUKE CODE BACKUP==========
 
 /obj/item/paper/fluff/stations/centcom/nuke_code_paper
-	name = "paper- 'Backup Nuclear Code'"
+	name = "paper- 'NUCLEAR CODE: DO NOT LOSE'"
+	icon_state = "docs_blue"
 	info = null
-	infolang = /datum/language/common
+	infolang = null
 	New()
 		..()
-		info = "The Nuke Code is [nuke_code]"
+		info = "#The Nuke Code is [nuke_code]"
--- a/code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
+++ b/code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
@@ -1,3 +1,5 @@
+var/nuke_code = random_nukecode()
+
 /obj/machinery/nuclearbomb
 	name = "nuclear fission explosive"
 	desc = "You probably shouldn't stick around to see if this is armed."
@@ -33,6 +35,8 @@
 	var/proper_bomb = TRUE //Please
 	var/obj/effect/countdown/nuclearbomb/countdown
 
+
+
 /obj/machinery/nuclearbomb/Initialize()
 	. = ..()
 	countdown = new(src)
@@ -43,7 +47,7 @@
 	GLOB.poi_list |= src
 	previous_level = get_security_level()
 	if(r_code == null)
-		r_code = random_nukecode()
+		r_code = nuke_code
 
 /obj/machinery/nuclearbomb/Destroy()
 	safety = FALSE
@@ -714,3 +718,14 @@ This is here to make the tiles around the station mininuke change when it's arme
 /obj/item/disk/nuclear/fake/obvious
 	name = "cheap plastic imitation of the nuclear authentication disk"
 	desc = "How anyone could mistake this for the real thing is beyond you."
+
+
+//==========CENTCOM NUKE CODE BACKUP==========
+
+/obj/item/paper/fluff/stations/centcom/nuke_code_paper
+	name = "paper- 'Backup Nuclear Code'"
+	info = null
+	infolang = /datum/language/common
+	New()
+		..()
+		info = "The Nuke Code is [nuke_code]"
--- a/code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
+++ b/code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
@@ -20,7 +20,7 @@
 	var/exploding = FALSE
 	var/exploded = FALSE
 	var/detonation_timer = null
-	var/r_code = "ADMIN"
+	var/r_code = null
 	var/yes_code = FALSE
 	var/safety = TRUE
 	var/obj/item/disk/nuclear/auth = null
@@ -42,6 +42,8 @@
 	update_icon()
 	GLOB.poi_list |= src
 	previous_level = get_security_level()
+	if(r_code == null)
+		r_code = random_nukecode()
 
 /obj/machinery/nuclearbomb/Destroy()
 	safety = FALSE
--- a/code/modules/antagonists/nukeop/nukeop.dm
+++ b/code/modules/antagonists/nukeop/nukeop.dm
@@ -231,10 +231,8 @@
 		var/obj/machinery/nuclearbomb/selfdestruct/nuke = locate() in GLOB.nuke_list
 		if(nuke)
 			nuke_team.tracked_nuke = nuke
-			if(nuke.r_code == "ADMIN")
-				nuke.r_code = nuke_team.memorized_code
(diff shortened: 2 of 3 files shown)


SomeguyManperson:
if you give something a value it WILL be used regardless of intent. It doesn't matter if it's a 5 digit fully random code if it's still something that really exists people will go for it

Nickvr628:
Two points: There are thousands of possible combinations so guesswork is incredibly unlikely, and you need the nuke disk inside to start inputting codes.

This means the random clown/greytider cannot just sit at the machine and try to brute force his way in, he would need to have stolen the nuke disk first. This would result in a ban in almost any case of this occurring. If the captain wanted to try it without a good reason that would probably prompt some form of ban as well. 

Besides, not wanting to add something because someone could break the rules with it is a very weak argument.

SomeguyManperson:
having the reason to add something being specifically so you can break the rules with it is an even weaker argument

ynot01:
>  A copy of the code can be found in the operations room of centcom.

are you this retarded

Nickvr628:
EDIT: I was unaware ghosts could examine papers to read them, I will remove the code paper from centcom

OppressedCactus:
There are so many combinations that it would be super impractical to guess them all, even if they did manage to guess it (after stealing the nuke disk) I would have a hard time believing it wouldn't be stopped by an admin quickly.

boodaliboo:
the chances are 9 x 9 x 9 x 9 x 9 or something i dunno

code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
```diff
@@ -42,6 +46,8 @@
 	update_icon()
 	GLOB.poi_list |= src
 	previous_level = get_security_level()
+	if(r_code == null)
```
JamieD1, line 49: if(!r_code)

code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
```diff
@@ -712,3 +718,15 @@ This is here to make the tiles around the station mininuke change when it's arme
 /obj/item/disk/nuclear/fake/obvious
 	name = "cheap plastic imitation of the nuclear authentication disk"
 	desc = "How anyone could mistake this for the real thing is beyond you."
+
+
```
JamieD1, line 722: SPACES

code/modules/antagonists/nukeop/nukeop.dm
```diff
@@ -231,10 +228,8 @@
 		var/obj/machinery/nuclearbomb/selfdestruct/nuke = locate() in GLOB.nuke_list
 		if(nuke)
 			nuke_team.tracked_nuke = nuke
-			if(nuke.r_code == "ADMIN")
-				nuke.r_code = nuke_team.memorized_code
-			else //Already set by admins/something else?
-				nuke_team.memorized_code = nuke.r_code
+			nuke.r_code = nuke_team.memorized_code
+
```
JamieD1, line 232: SPACE

Alexandria404:
if you want to make QOL for admins make it a button that generates one
else you are just trying to bait people to guess it

code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
```diff
@@ -1,3 +1,5 @@
+var/nuke_code = random_nukecode()
```
JamieD1, line 1: Global var bad,

Can you do this another way?

code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
```diff
@@ -712,3 +718,15 @@ This is here to make the tiles around the station mininuke change when it's arme
...
+
+
+//==========CENTCOM NUKE CODE BACKUP==========
+
+/obj/item/paper/fluff/stations/centcom/nuke_code_paper
+	name = "paper- 'NUCLEAR CODE: DO NOT LOSE'"
+	icon_state = "docs_blue"
+	info = null
+	infolang = null
+	New()
```
MCHSL, line 730: replace with `/obj/item/paper/fluff/stations/centcom/nuke_code_paper/Initialize()`

code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
```diff
@@ -712,3 +718,15 @@ This is here to make the tiles around the station mininuke change when it's arme
...
+		..()
```
MCHSL, line 731: replace with `. = ..()`

code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
MCHSL: replace with `GLOBAL_VAR_INIT(nuke_code, random_nukecode())` and add `GLOB.` before all instances of `nuke_code`

code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
```diff
@@ -712,3 +718,15 @@ This is here to make the tiles around the station mininuke change when it's arme
...
 	name = "cheap plastic imitation of the nuclear authentication disk"
 	desc = "How anyone could mistake this for the real thing is beyond you."
+
+
+//==========CENTCOM NUKE CODE BACKUP==========
+
+/obj/item/paper/fluff/stations/centcom/nuke_code_paper
+	name = "paper- 'NUCLEAR CODE: DO NOT LOSE'"
+	icon_state = "docs_blue"
+	info = null
```
MCHSL, line 728: probably can remove this and the line below

Marmio64:
you're going to at least have to add something to the self destruct that will make it stop you for a while if you keep guessing wrong codes so it isn't abused so much.

Like, a 30 second stopper for guessing wrong 3 times in a row.

aaqe:
Who needs this?

alexkar598:
It would take on average 50 minutes to brute force the code at 10 tries per second ... which is alt but like... Mods cant even send erts, I'm not sure this is needed

code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
Hopekz: My pupil

JamieD1:
To be honest,

All this does is let mods have the ability to give out nuke codes.

TheGamerdk:
Mods can already give nuke codes. They just need a request to be made. Mods shouldn't be giving out codes if no request is out anyways

Ktlwjec1:
if a maint merges this please fire them. also it literally takes like 5 seconds to set the code it's really not a hassle at all.

Takahiru:
If i ever get hijack id gladly steal the nuke code and sit in maint for an hour to brute force it, why would you add this?

deathslayer2481:
I will say that ‘random’ is never truelly ‘random’.

Hopekz:
This is an admin quality of life PR.
The reception this PR is getting both here and on discord from the admins that this PR is supposed to be benefiting is very negative.
Closing this PR.

code/modules/antagonists/nukeop/equipment/nuclearbomb.dm
Amelia0010: image

PR #8622, opened 2020-05-22
Free Miner surgical bag now actually has stuff in it

Jumps0:

The surgical bag on the Free Miner ship in the medical room was empty for some reason (could be an intentional prank but idk). Now instead of being empty its an actual surgery bag with tools in it.

#### Changelog

:cl:  
tweak: Free Miner surgical bag now actually has tools in it
/:cl:


Ktlwjec1 (approved):
a good example of why we don't edit bag or lockers contents on the map, but code it in instead, something will always go wrong

PR #8632, opened 2020-05-22
Makes the syndicate toolbox a medium item

Darkstick:
Makes the syndicate toolbox a medium item like it used to be a long time ago, which means you can stuff it in your bag. This should make the 1TC investment worth it since you wont need to find a toolbelt to not have to walk around with a big valid target in your hands.

:cl:  
tweak: made the syndicate toolbox a medium item 
tweak: changed the uplink description to reflect this change
/:cl:


boodaliboo:
you changed the dme for no reason that may be a clue on what to fix

PR #8637, opened 2020-05-22
Vampires can now go loud and unlock all abilities

ToGWtF:
Vampires can now go loud and unlock all abilities instantly. Also gives them a pool of 450 useable blood to actually use their abilities

Forces them to have a blood objective that's equal to multiple kill objectives

Forces them to wear a dracula cape. No more hiding in plain sight.

They lose the ability to transform if they drink enough blood to unlock even one new power.

#### Changelog

:cl:  
rscadd: Gives vampires new transformation. Unlocking their true powers
/:cl:

### DIFF
--- a/yogstation/code/datums/antagonists/vampire.dm
+++ b/yogstation/code/datums/antagonists/vampire.dm
@@ -68,7 +68,6 @@
 	give_objectives()
 	add_ability(/obj/effect/proc_holder/spell/self/nosferatu)
 	check_vampire_upgrade()
-	to_chat(owner, "<span class='notice bold'>DEBUG : Adding nosferatu on gain.</span>")
 	owner.special_role = "vampire"
 	owner.current.faction += "vampire"
 	SSticker.mode.update_vampire_icons_added(owner)
@@ -148,7 +147,6 @@
 /datum/antagonist/vampire/proc/remove_objective(var/datum/objective/O)
 	objectives -= O
 	objectives_given -= O
-	to_chat(owner, "<span class='notice bold'>DEBUG : Tried to remove objective.</span>")
 
 /datum/antagonist/vampire/proc/forge_single_objective() //Returns how many objectives are added
 	.=1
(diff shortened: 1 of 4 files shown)


Takahiru:
this is actually shit

ynot01:
literally anything is better than what we have now lol

patpol4:
i wanted to say no but i read the description and i'm going to say yes

yacabo:
War vamps

overlord12345:
its bad but its better then what we have so meh

Hopekz:
I think this PR is worth a try.

Takahiru:
This completely removes the point of having any early game vampire whatsoever, every time when there is a vampire as soon as they get enough to have all powers theres literally no point in hiding it so they just wear the cloak walk around and if they meet any trouble they use one of their many many escape tools to get away, this will just make vampire fake wizards is all.

Jarod1200:
I feel like this should only be an option on high pop because pressing a gamer button to become an anemic bootleg wizard kinda takes the "be stealthy and snowball your power out of it." 
Vamps that make it to full potential anyway usially have no problem sucking a ton of blood and making a ton of new bloods.

Marmio64:
i mean.... okay... but i'd rather have some changes to the abilities then just something like this.

I'd rather have it if vampires lost their jaunt and the screech only heavily disoriented people with ear protection instead of stunning them. To buff vampires then since they'd lose their most powerful ability and have another one nerfed, hypnotize should be a instant cast and stun for 5 seconds while muting for 10, however the stun is delayed by two seconds and gives a big warning to the person being hypnotized. Hypnotize also shouldn't work on mindshielded people though because the implant would protect them. 

 Along with this vampires should be able to transform into a Nosferatu and back with a 30 second toggle cooldown. While Nosferatu they are obviously a vampire , lose human status (if they had it), and no longer have any benefits or drawbacks of their previous species,  however they move 15% faster than humans, get 50% stun resistance, get a slow passive healing, and drink blood at twice the rate.

Takahiru:
I believe having marmio's Nosferatu idea is much much better than this, this literally lets multiple antags get wizard powers roundstart. This just removes the entire point of vampire being getting powers over time and just makes it a worse wizard. This will make vampire even worse

Alexandria404:
seems like a murderbone increase

alexkar598:
Yeah how about no. This fits hippie or tg but not the slower experience yogs is *trying* to be

PR #8643, opened 2020-05-23
allows devil to very rarely happen as a midround 

boodaliboo:
want to see devil more its a insanely fun antag that encourages roleplay
this reintroduces it but without making it a common occurence
#### Changelog

:cl:  
tweak: added a weight of 3 to devil 
/:cl:

### DIFF
--- a/code/modules/events/devil.dm
+++ b/code/modules/events/devil.dm
@@ -1,7 +1,7 @@
 /datum/round_event_control/devil
 	name = "Create Devil"
 	typepath = /datum/round_event/ghost_role/devil
-	max_occurrences = 0
+	max_occurrences = 1
 	weight = 3
 
 /datum/round_event/ghost_role/devil
--- a/code/modules/events/devil.dm
+++ b/code/modules/events/devil.dm
@@ -2,6 +2,7 @@
 	name = "Create Devil"
 	typepath = /datum/round_event/ghost_role/devil
 	max_occurrences = 0
+	weight = 3
 
 /datum/round_event/ghost_role/devil
 	var/success_spawn = 0


Xoxeyos:
Make it a four or a five, three is pretty low.

boodaliboo:
> Make it a four or a five, three is pretty low.

you sure?

Hopekz:
Isn't the devil game mode kind of "broken" right now?
I thought there was a reason it was removed from rotation.

boodaliboo:
the one time i got to play it it worked fine and ended properly

swissloaf:
> Isn't the devil game mode kind of "broken" right now?
> I thought there was a reason it was removed from rotation.

dont think it is

boodaliboo:
> > Isn't the devil game mode kind of "broken" right now?
> > I thought there was a reason it was removed from rotation.
> 
> dont think it is

pretty sure it was removed because tg removed it and we rebased to them

code/modules/events/devil.dm
```diff
@@ -2,6 +2,7 @@
 	name = "Create Devil"
 	typepath = /datum/round_event/ghost_role/devil
 	max_occurrences = 0
```
swissloaf: now I'm no coder, but wouldnt a max occurrence of 0 mean there it is impossible for this event to happen

boodaliboo: fuck

boodaliboo:
travis broke again

boodaliboo:
thanks whoever did travis for me

Tipy1802:
I have never seen the devil in game but from what I have read from the wiki it sounds really fun and I have always wanted to see it in game. It also gives the lawyer and the curator more purpose since the lawyer can find loopholes in the contracts and the curator has the code gigax (or however it was called) that let's you find the Devil's weaknesses

JamieD1 (approved):

Hopekz (changes requested):
Exploits to be fixed before reintroducing this mode:

- People can unsell their own soul by just breaking in to lawyers office and they retain their powers.
- People can also get as many powers as they want by selling their soul over and over to the same devil so make it where you lose the power if you "unsell" your soul or something. This also for some reason counts for the devil depending on the objective they have.

Also, it seems odd that there are 6 devils spawning in at once not sure if 100% intentional as devils can be security as well (also not sure if intentional)

boodaliboo:
was this testmerged?

Hopekz:
> was this testmerged?

No; but we did some devil rounds to test the mode.

boodaliboo:
> > was this testmerged?
> 
> No; but we did some devil rounds to test the mode.

did you have it as the gamemode or as a midround?

Hopekz:
> > > was this testmerged?
> > 
> > 
> > No; but we did some devil rounds to test the mode.
> 
> did you have it as the gamemode or as a midround?

Game mode

boodaliboo:
> > > > was this testmerged?
> > > 
> > > 
> > > No; but we did some devil rounds to test the mode.
> > 
> > 
> > did you have it as the gamemode or as a midround?
> 
> Game mode

could you try it as a midround? want to see if that fixes the problem of multiple devils

boodaliboo:
and for problem 1 and 2 not sure how to go about fixing it but ill look

boodaliboo:
@Hopekz 
image

alexkar598:
Devil has too much to break the flow of the round to be a mid round (if you mention xeno, blob or zombies i will end you, i dont like those others, should be in dynamic not midround)

deathslayer2481:
Reopen this shit, Devil is an RP midround way better than XENO BLOB AND ZOMBIES and taking the lesser evil is better than keeping things you think are bad.

alexkar598:
img

PR #8655, opened 2020-05-23
VAMBIRE STUPID BALANCE CHANGES YAAAAAAAAAAAY

SomeguyManperson:
Vampires now have slight variation to their suck speed dependent on grab strength
not grabbing: stealth suck; vampire will not give off any blatantly obvious tells that they are a vampire, suck speed reduced to half, vampire will be alerted when they are about to pass the blood volume danger threshold, every successful suck afterwards will alert the victim while stealth and suck remain intact
passive or aggressive grab: normal sucking speed; no changes
neck or kill grab: vampire sucks 50% faster
These all smoothly transition into eachother except for anything into stealth because what's the point of going from not stealth to stealth
:cl:  
rscadd: Vampire suck speed now has three different variations depending on grab strength
rscadd: Vampires now have a sneaky way to extract blood from the unwitting if they aren't grabbing them
/:cl:


patpol4:
What if you used pickpocket mechanics for the stealth suck instead. 
If they move during the process, they get alerted and it stops

SomeguyManperson:
that'd make it roughly as useful as just biting someone in public since people are BOUND to move randomly with little to no ways to tell

JamieD1 (approved):

alexkar598 (approved):

alexkar598:
Suggestions:

Make it suck quicker/slower instead of more/less
Make it no longer silent once they are out of stealth blood volume

Hopekz:
What happens when a vampire wants to punch someone in the head?
Does this have a chance to just straight up give them away in a fistfight since both parties are moving?

SomeguyManperson:
This was more of an issue before since the context to stealth is grab state, so unless the vampire tries to punch in the face then grab they'll just attempt a stealth suck then fail and stop, which shouldn't make any noise

SomeguyManperson:
what

SomeguyManperson:
I pushed a fix but didn't update the branch first yay

Marmio64:
i dont think this is working

PR #8656, opened 2020-05-23
hypnotize is now uninterruptible and causes effects on cast

SomeguyManperson:
this might be a bit unintuitive
Hypnotize has always been terrible and that's because it's delayed
instead of making it not delayed I have elected to make it stay delayed but be harder to break
hypnotize now mutes for 4 seconds on being cast and deals 40 stamina damage to the victim as well if they are not wearing flash protection
after 3 seconds, if the victim is still in range of the user, they are stunned fully as normal, otherwise the spell isn't cast since the cooldown is like 5 minutes or something
also it has an initiation range of 2 meters instead of whatever the fuck it was before hopefully Ididn't screw this part up
:cl:  
rscadd: Vampire hypnotize spell has a mute attached to the start and is now uninterruptible, it can be broken if you break line of sight with the vampire when it activates
rscadd: Vampire hypnotize spell now requires the target to be within 2 meters to be used
/:cl:

### DIFF
--- a/yogstation/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/yogstation/code/game/gamemodes/vampire/vampire_powers.dm
@@ -117,7 +117,7 @@
 	for(var/mob/living/carbon/target in targets)
 		user.visible_message("<span class='warning'>[user]'s eyes flash briefly as [user.p_they()] stares into [target]'s eyes</span>")
 		target.silent = 2 //this is actually roughly 4 seconds due to how silent works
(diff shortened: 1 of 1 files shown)


JamieD1 (approved):

Hopekz:
This doesn't seem to synergize with #8637 so it would probably have to be one or the other.

alexkar598:
Im a bit worried about the balance implications of being able to chase someone down a hall and stun them while running after them but lets try this out

PR #8664, opened 2020-05-23
Increased changeling transformation cap

Hopekz:
Apparently changelings are limited to 6 absorbs causing them to miss their objectives if they don't absorb their target in their first 6.

Changed this to 12

#### Changelog

:cl:  
tweak: doubled the arbitrary ling absorb transformation limit.
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/changeling.dm
+++ b/code/modules/antagonists/changeling/changeling.dm
@@ -17,7 +17,7 @@
 
 	var/list/stored_profiles = list() //list of datum/changelingprofile
 	var/datum/changelingprofile/first_prof = null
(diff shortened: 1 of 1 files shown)


boodaliboo:
i thought this was a bug not a feature

Marmio64:
yes god please yes yes

JamieD1 (approved):

MacHac:
It would be nice if 'impersonate' targets could never be lost period.  I understand that would require coupling two systems that don't like talking to each other though, so this is a fine interim solution.

alexkar598:
Its a feature, if you absorb more than 6 people, the first one is removed and discarded

alexkar598:
Its a feature, open a new PR if you want this without branding it as a fix

PR #8667, opened 2020-05-24
Borgs now get a message when unlocked/locked

MacHac:
Increase borg powercreep.

Borgs can now see if the roboticist actually locked them before they roll out of the shop.  The old way of doing this was to use the Unlock Cover verb, which fails silently when the cover is unlocked.  I'm just making it obvious.
#### Changelog

:cl:  
tweak: Borgs can tell when someone locks or unlocks their cover.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -388,7 +388,7 @@
 				return
 			if(health > 0)
 				return //safety check to prevent spam clciking and queing
-		
+
 		adjustBruteLoss(-30)
 		updatehealth()
 		add_fingerprint(user)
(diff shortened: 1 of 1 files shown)


maxion12345:
Just
Just examine yourself

JamieD1 (approved):

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -506,6 +506,7 @@
 			if(allowed(usr))
 				locked = !locked
 				to_chat(user, "<span class='notice'>You [ locked ? "lock" : "unlock"] [src]'s cover.</span>")
+				to_chat(src, "<span class='notice'>[usr] [locked ? "locks" : "unlocks"] your cover.</span>")
```
alexkar598, line 509:
```suggestion
				to_chat(src, "<span class='notice'>[usr] [locked ? "locked" : "unlocked"] your cover.</span>")
```

alexkar598 (changes requested):

code/modules/mob/living/silicon/robot/robot.dm
MacHac: The unlocker message is in the present tense.  Why would the borg message be in the past tense?

alexkar598: Good point

PR #8678, opened 2020-05-24
Gives cyborgs use of respective radio channels (TG Port)

Jumps0:

All credit to the original coder **Kelenius**.
https://github.com/tgstation/tgstation/pull/51222

"Gives the cyborgs access to their department's radio:

- Medical gets medical.
- Engineering gets engineering.
- Service and Janitor get service.
- Mining gets supply and science.
- Security gets security. In case it's enabled, spawned, etc.
- Peacekeeper gets service to find out where to tactically deliver cookies.
- Clownborg gets service because clown.
- Standard gets nothing. Pick a side, filthy centrist.

This channel is not related to the encryption key that can be installed into any cyborg and works in addition to they key, if installed.

I didn't touch the syndicate cyborgs, they still use an encryption key. I don't want to mess with radio code more than necessary and their encryption key already gives them everything they want, so it's unlikely someone will ever mess with it. Not broken, no fixes."

#### Changelog

:cl:  Kelenius
rscadd: Cyborgs (except Standard) now have access to their department's radio.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/radio.dm
+++ b/code/game/objects/items/devices/radio/radio.dm
@@ -51,10 +51,7 @@
 	frequency = add_radio(src, new_frequency)
 
 /obj/item/radio/proc/recalculateChannels()
-	channels = list()
-	translate_binary = FALSE
-	syndie = FALSE
-	independent = FALSE
+	resetChannels()
 
 	if(keyslot)
 		for(var/ch_name in keyslot.channels)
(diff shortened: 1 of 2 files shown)


JamieD1:
I suggested this ages ago,

What I was told is the ai should be telling you what to do and where to go not departments,

But im ALL for this change.

Jumps0:
I completely understand the argument for "it's the job of the AI to relay messages for the borg". But that's honestly a lot of work for the AI to do that IMO shouldn't really be necessary because they have better things to do. Seeing engi borgs drag a headset to robo and ask for the chip to be installed seems like an example of this "messenger" problem, so I feel like this is a needed change. Additionally these are the votes on the original TG PR:
image

jcatjmeow:
"it's the job of the AI to relay messages for the borg" is a fucking stupid argument because borgs can have robos give them radio keys anyway, if that really was true that wouldn't be a thing

ynot01:
should probably give peacekeeper service

Nickvr628:
I don't like that it is a secborg buff but I see the utility of it, especially for the other borg types.

Hopekz (approved):
Positive reception.
Code checks out.
Merge time.

PR #8689, opened 2020-05-25
[admin] Applies retmin tracking to all staff

Amelia0010:

Personally I don't think that making it so anyone who has the game on in the background doesn't get counted for hours is a good idea, although mostly because back in the day that's how lowpop was handled. However, I see the merits when applied to cases where a lot of admins are on at once, and people join and just sit around doing nothing.

Either way, if you're going to apply this policy it should be applied equally :)

#### Changelog

:cl:  
tweak: Admin tracking now doesn't count AFK for all staff rather than just retmins
/:cl:

### DIFF
--- a/code/controllers/subsystem/server_maint.dm
+++ b/code/controllers/subsystem/server_maint.dm
@@ -71,7 +71,7 @@ SUBSYSTEM_DEF(server_maint)
 				continue
 
 		if(can_tracking)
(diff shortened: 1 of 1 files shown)


alexkar598 (approved):

AsV9 (approved):

Xantam-13:
I'll make it a council vote.

alexkar598:

PR #8690, opened 2020-05-25
[Admin] Removes retmin tracking

ThatLing:
Alternative to #8689.

Some admins might AFK on the server while doing something else, but still handle tickets. Best solution might be to only do this if there are 2 or more admins online. /shrug

### DIFF
--- a/code/controllers/subsystem/server_maint.dm
+++ b/code/controllers/subsystem/server_maint.dm
@@ -70,13 +70,6 @@ SUBSYSTEM_DEF(server_maint)
 				QDEL_IN(C, 1) //to ensure they get our message before getting disconnected
 				continue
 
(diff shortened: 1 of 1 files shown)


Amelia0010:
I think doing it once you get to 3 admins online is probably the sweet spot, since in my experience that's when people start joining to farm hours without doing anything

Amelia0010:
Maybe we should make phantom tickets appear at that admincount, and not count the admin if they don't respond to their phantom tickets?

alexkar598 (changes requested):
Remove the can_tracking variable

Xantam-13:
Made it a council vote.

alexkar598 (approved):

alexkar598:

PR #8691, opened 2020-05-25
[admin] Only do staff tracking when there are more than 3 admin online

ThatLing:
Alternative to #8690 and #8689.

Will not count staff time when afk if there are 3 or more admins online (including the afk admin themselves). This applies to all staff, not only retmin.

#### Changelog

:cl:  
tweak: Staff tracking no longer tracks your time if you are AFK while there are other admins on the server.
/:cl:

### DIFF
--- a/code/controllers/subsystem/server_maint.dm
+++ b/code/controllers/subsystem/server_maint.dm
@@ -71,7 +71,9 @@ SUBSYSTEM_DEF(server_maint)
 				continue
 
 		if(can_tracking)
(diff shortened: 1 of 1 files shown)


monster860 (changes requested):
It doesnt log them back in if they are inactive and then the staff rises up to 3

ThatLing:
@monster860 How does it not? The admin check is only for logging out

Xantam-13:
Council vote.

ThatLing:
Also would like input if I should include afk admins to the total admin count

EDIT:
Added this for now, tell me if you don't like it

alexkar598:

PR #8698, opened 2020-05-26
Syndicate camera flash

Hopekz:
New syndicate stealth item:

Camera flash.
Its a camera with a powerful camera flash installed that can be used much like a regular flash except with a long cooldown between uses.

**EDIT:** 
Alex's suggestions implemented:
- Cooldown is now 9 seconds regardless of trait
- TC cost is now 2

Cooldown: 
regular: 9 seconds
with photography trait: 4.5 seconds

TC cost: 4

#### Changelog

:cl:  Hopek
rscadd: Added a new syndicate item: the camera flash.
/:cl:

### DIFF
--- a/code/modules/photography/camera/camera.dm
+++ b/code/modules/photography/camera/camera.dm
@@ -227,3 +227,67 @@
 		p.set_picture(picture, TRUE, TRUE)
 		if(CONFIG_GET(flag/picture_logging_camera))
 			picture.log_to_file()
+
+
+
+/obj/item/camera/syndicate/attack(mob/living/carbon/human/M, mob/user) // syndicate camera flash
+	if(!on)
+		return
+	if(iscarbon(M))
+		flash_carbon(M, user, 5, 1)
+		playsound(src, 'sound/weapons/flash.ogg', 100, TRUE)
+		flash_lighting_fx(8, light_power, light_color)
+		flash_cooldown(user)
+		return TRUE
+	else if(issilicon(M))
+		var/mob/living/silicon/robot/R = M
+		log_combat(user, R, "flashed", src)
+		R.Paralyze(rand(80,120))
+		var/diff = 5 * CONFUSION_STACK_MAX_MULTIPLIER - M.confused
+		R.confused += min(5, diff)
+		R.flash_act(affect_silicon = 1)
+		playsound(src, 'sound/weapons/flash.ogg', 100, TRUE)
+		flash_cooldown(user)
+		flash_lighting_fx(8, light_power, light_color)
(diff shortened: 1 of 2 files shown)


MacHac:
This'll actually be the only instant contact stun in the uplink; unless you're a roboticist this'll probably be a pretty easy buy.  _Halving the cooldown when you have a one point trait makes it even more attractive and gamers can just take 'Light Drinker' or something to make this item more valuable._

_Edit: The trait no longer has an effect on the item.  I'm now fine with the item in principle, I'm just not convinced about the cost._

If you role restricted it to tourist, it might be a fine buff to a weak role.  Otherwise, 4TC seems a little light for something that can incapacitate most of the station long enough for cuffs (and hence a lethal beating).  You might also want to make it vulnerable to EMPs if it isn't already so it's not a better choice than the regular flash in that respect.

In my view this is at least as dangerous as an esword, maybe more so.  Combine it with a 1TC EMP flashlight and a set of cable cuffs and you can punch your target to death while still being able to pass a search.

Hopekz:
@MacHac 
You commented right as I was updating the item. Half your comment doesn't apply anymore so just letting you know.

No more trait bonus (always a 9 second cooldown)

Marmio64:
this sounds good

JamieD1 (approved):

alexkar598:
Im not too sure about giving antags cheap metashielded stuns. I honestly would feel better if it was an actual flash instead of a metashielded item

Hopekz:
> Im not too sure about giving antags cheap metashielded stuns. I honestly would feel better if it was an actual flash instead of a metashielded item

Stealth metashielded items go with the "syndicate" vibe.
It would be inconsistent to introduce a syndicate stealth item that is modeled after a contraband item.

As far as I can tell this is worse than a flash in most ways because it can only be used once every 9 seconds (meaning that stunning 2 people or re-applying the stun if they have stun reducing buffs is a non-option).
This stun also comes with all the limitations of using an actual flash so to me at least it looks very much appropriate.

As far as the cost: You suggested the cost to be lower so I modified it; should it go higher than the initial suggestion? I think 2-4 is the correct range for this item so I personally wouldn't mind making this change.

alexkar598:
Yeah sorry I just cant support a literal metashielded flash. The argument of the cooldown is kinda moot because cable restraints are a thing

Hopekz:
r.i.p

Marmio64:
add camera flash but make it tourist only, so that it is 1000x more rarer, because tourists can only get midround traitor since you cant roundstart join as them, and also nobody even plays tourist.

Hopekz:
> add camera flash but make it tourist only, so that it is 1000x more rarer, because tourists can only get midround traitor since you cant roundstart join as them, and also nobody even plays tourist.

@alexkar598 ;
Will you close my PR if I re-make it to be tourist only?
If not; you can re-open this PR instead maybe

PR #8721, opened 2020-05-27
fixes morgue trays being over items

ToGWtF:
Makes morgue trays the same item level as tables, meaning items and bodybags will show up over them


JamieD1 (approved):
Speedmerge

Hopekz (approved):

PR #8734, opened 2020-05-28
Clears up a common misconception.

Hopekz:
This gets brought up a lot and it's driving me crazy because the first thing people look for when deciding to play on a server is to see if it's being updated.

Also; potential coders get turned off to dead projects (because who would even care to go to our github if our changelog didn't update for months)

### DIFF
--- a/html/changelog.html
+++ b/html/changelog.html
@@ -27,6 +27,7 @@
 			
 			<p><div align='center'><font size='3'><a href="https://forums.yogstation.net/index.php">Forum</a> | <a href="https://wiki.yogstation.net/index.php?title=Main_Page">Wiki</a> | <a href="https://github.com/yogstation13/Yogstation">Source</a></font></div></p>
 			<font size='2'><b>Visit #development-public on our Discord:</b> <a href="https://discord.gg/p6ShbB8">https://discord.gg/p6ShbB8</a></font><br>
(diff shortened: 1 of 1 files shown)


Marmio64:
blessed

boodaliboo:
more coders good i think

alexkar598:
Archive the goddamn thing, I'd do it myself if it wasn't alot of spaghetti

Hopekz:
> Archive the goddamn thing, I'd do it myself if it wasn't alot of spaghetti

What does that even mean what actually needs to happen for it to update like every other server?
Do we need to port some system? Every other server has their changelog updated to the last 24H

This is literally making us look terrible

As far as I can tell only people with host access can do something about it.

JamieD1:
Can we not just fix the changelog?!?!?!?!?!?!?!?!?!?

Hopekz:
> Can we not just fix the changelog?!?!?!?!?!?!?!?!?!?

Fixing it is preferable.
I've been asking for it for 6+ months now.
The people that have access to it don't and while it isn't fixed it is giving a lot of confusion on our project seeming dead.

Also let's not forget the changelog/recognition is the only real reason to contribute so I would want to continue encouraging coders to stay by giving them credit where it was due. I don't know what is needed to "fix" the changelog nor do I have access to it from what I can tell so all I can do is PR this notification.

PR #8514, opened 2020-05-15
Revert "Stupid proofs erts"

boodaliboo:
people want this reverted this is the pr that made most centcom ids undroppable

PR #8737, opened 2020-05-28
[admin]Revert "Revert "Stupid proofs erts""

Hopekz:
Reverts yogstation13/Yogstation#8514

We now have to deal with people coming in masses to centcom like literally every single time an ERT is sent. Why did this get merged without an alternative fix who knows.

Even the shift that made me start this revert there were like 8 people that shouldn't be there on centcom and removing this without having a backup is dumb

### DIFF
--- a/code/game/objects/items/cards_ids.dm
+++ b/code/game/objects/items/cards_ids.dm
@@ -436,6 +436,7 @@ update_label("John Doe", "Clowny")
 	icon_state = "centcom"
 	registered_name = "Central Command"
 	assignment = "General"
+	item_flags = DROPDEL //admemes arnt the smartest tools in the shed
 
 /obj/item/card/id/centcom/silver
 	name = "\improper silver CentCom ID"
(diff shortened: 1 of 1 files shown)


JamieD1 (approved):

boodaliboo:
lol this was never a problem before we had the original pr didnt expect it to become one

Hopekz:
> lol this was never a problem before we had the original pr didnt expect it to become one

Why do you think it became undropable in the first place AAH
Yes undropable items are bad but we have no other clever solution to replace this at the moment and we can't just have this be a thing in the time being while waiting for a new solution to replace this one.

ToGWtF:
So let me get this straight

We are now rebalancing the rebalance, of the rebalance, of the balance of ERTs. 

The absolute state of it

ynot01:
what if you idk

sent the ERT shuttle up after ERT arrive??

Hopekz:
> sent the ERT shuttle up after ERT arrive??

Then you remove those cool evacuation moments for ERT (a.k.a every time they're wiped out)
I want the ERT to be able to do things on their own without having an admin baby them.

We need another solution for this that isn't no-drop but not having a solution in place at all is worse.

alexkar598:
Nodrops are bad

PR #8743, opened 2020-05-28
Who divides by zero - Fixes inf gas maybe

JamieD1:
Monster, Please fix this...

Who divides by zero.

#### Changelog

:cl:  
bugfix: patches monsters shitty code
/:cl:

[2020-05-26 18:06:41.803] runtime error: Division by zero
 - proc name: react (/datum/gas_reaction/fusion/react)
 -   source file: reactions.dm,296
 -

### DIFF
--- a/code/modules/atmospherics/gasmixtures/reactions.dm
+++ b/code/modules/atmospherics/gasmixtures/reactions.dm
@@ -255,6 +255,9 @@
 	for (var/gas_id in air.get_gases())
 		gas_power += (GLOB.meta_gas_info[gas_id][META_GAS_FUSION_POWER]*air.get_moles(gas_id))
 	var/instability = MODULUS((gas_power*INSTABILITY_GAS_POWER_FACTOR)**2,toroidal_size) //Instability effects how chaotic the behavior of the reaction is
+	if(instability == 0)  //TEMP FIX - MONSTER FIX YOUR SHIT
+		instability = 0.1
+
 	cached_scan_results[id] = instability//used for analyzer feedback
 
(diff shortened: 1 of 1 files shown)


Bibby0110:
this feels horrible

code/modules/atmospherics/gasmixtures/reactions.dm
```diff
@@ -255,6 +255,9 @@
 	for (var/gas_id in air.get_gases())
 		gas_power += (GLOB.meta_gas_info[gas_id][META_GAS_FUSION_POWER]*air.get_moles(gas_id))
 	var/instability = MODULUS((gas_power*INSTABILITY_GAS_POWER_FACTOR)**2,toroidal_size) //Instability effects how chaotic the behavior of the reaction is
+	if(instability == 0)  //TEMP FIX - MONSTER FIX YOUR SHIT
```
MacHac, line 258: A better way to do this might be:
```dm
// (in defines)
#define MINIMUM_INSTABILITY 1e-8

// (in the code)
instability = max(instability, MINIMUM_INSTABILITY)
```

Keep in mind, 0 instability isn't actually the source of the infinite temperature thing.  The only thing that divides by instability is the radiation calculation on line 299; producing infinite radiation is a Bad Thing, but radiation doesn't directly affect temperature.

Could it be that the temperature is just going out of bounds?  A gas will reach infinite temperature at around 3x10^38 degrees, which is obtainable with some finesse (or cheating.)  Given sufficient fuel, most fusion reactions will eventually hit infinite temperature; the trick is getting enough fuel in the can while still hitting ignition temperature.  This might be why you see it a lot when podmins fill the centcom dock with fusion mix but not so much when actual atmos techs do fusion.

Bibby0110: Me thinks it's probably that happening in monstermos c++ and getting passed here.
Not sure we have a temp cap

PR #8749, opened 2020-05-29
Cyborgs can see their master AI's objectives

MacHac:
Synced cyborgs now see their master AI's objectives on the status panel.

I can't really think of any reason why this shouldn't be the case.  I'm sure the contrarians among you can, but I consider this a QoL improvement.  It helps the borgs act more independently and cuts down on the interminable 'wats going on' spam whenever a lot of borgs are being made.

#### Changelog

:cl:  
tweak: Cyborgs can see their master AI's objectives.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -317,6 +317,14 @@
 			. += "[st.name]: [st.energy]/[st.max_energy]"
 	if(connected_ai)
 		. += "Master AI: [connected_ai.name]"
+		if (connected_ai.mind)
+			var/datum/antagonist/traitor/A = connected_ai.mind.has_antag_datum(/datum/antagonist/traitor)
+			if (A && A.objectives && A.objectives.len)
+				. += ""
+				. += "<b><font color='#ff0000'>Accomplish the following:</font></b>"
+				var/counter = 1
+				for (var/datum/objective/O in A.objectives)
+					. += "[counter++]. [O.explanation_text]"
 
 /mob/living/silicon/robot/restrained(ignore_grab)
 	. = 0
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -317,7 +317,7 @@
 			. += "[st.name]: [st.energy]/[st.max_energy]"
 	if(connected_ai)
 		. += "Master AI: [connected_ai.name]"
-		if (connected_ai.mind)
+		if (connected_ai.mind && lawupdate)
 			var/datum/antagonist/traitor/A = connected_ai.mind.has_antag_datum(/datum/antagonist/traitor)
 			if (A && A.objectives && A.objectives.len)
 				. += ""


Bibby0110:
What if their laws aren't synced?

JamieD1:
I mean kinda feels like a bad idea..

Feels a bit metagame

alexkar598:
Not a fan of this, there are reasons to not tell your borgs your objectives because some of them can be retarded and blow your cover immediately

Takahiru:
if borgs are desync'd then theyll end up telling everyone all of the AI's objectives

MacHac:
Malf AI is set up so it can't do its objectives without cyborgs if even one robust person is aboard, so the cyborgs know anyway.  Synced cyborgs are supposed to be an extension of the AI anyway; I'm just extending that metaphor in a way that conveniently reduces the number of times I have to repeat myself when getting a bunch of borgs.

Deliberately hiding information from your own borgs is the sort of stupid 'mechanic' that could only ever be invented by accident.  If a borg gets desynced, the crew knows you're rogue; it's not like they're going to lynch you less if they know the names of the people you're trying to kill.

Sometimes good features get left off an initial design because of developer oversight, or because it was hard to do at the time.  This is one of those features, just like showing laws in the status tab.  I get that this game is complex, but this is one of those teeny tiny changes that the meta will compensate for within minutes and then we'll have a nice quality of life change.

TL;DR: I'll acknowledge this developer oversight as a feature if someone can demonstrate that a.) it was done intentionally or b.) it's good for the game.  Unless that happens, have a nice QoL improvement for malf AI!

Bibby0110:
What if the borg isn't lawsynced, but still synced to the AI?
There's a difference between being synced to the AI, and being lawsynced to the AI.
Cover would instantly be blown if this happened.

AI should have full control over it's approach, which means borgs shouldn't do anything but what the AI says, which means the objectives are just there to tempt borgs into blowing the AI's cover.

Borgs being more independent isn't always a good thing, and if a borg can't follow simple orders you're fucked either way

MacHac:
I've now made it so that borgs only see objectives while lawsynced.  If they're lawsynced at any point, they get the Law 0, so it shouldn't matter if lawsync is turned off in the middle of the round.

This PR is not about not about not wanting to give borgs orders, it's about not wanting to give every new borg a full description of objectives (then reminders when they forget in 5 minutes.)  If the AI gives a standing order not to do anything suspicious and a borg blows that, that's already a jobban.  We don't need to also cover that by trying to withhold information most borgs already have anyway.

Really, my biggest point is that AIs are already handing this information out over binary chat in 99.9% of rounds.  Filthy metagaming roboticists looking to lynch the AI can already desync a borg and have it listen to bchat anyway, so it's not like this change is going to prevent AIs from getting found out.

JamieD1:
> I've now made it so that borgs only see objectives while lawsynced. If they're lawsynced at any point, they get the Law 0, so it shouldn't matter if lawsync is turned off in the middle of the round.
> 
> This PR is not about not about not wanting to give borgs orders, it's about not wanting to give every new borg a full description of objectives (then reminders when they forget in 5 minutes.) If the AI gives a standing order not to do anything suspicious and a borg blows that, that's already a jobban. We don't need to also cover that by trying to withhold information most borgs already have anyway.
> 
> Really, my biggest point is that AIs are already handing this information out over binary chat in 99.9% of rounds. Filthy metagaming roboticists looking to lynch the AI can already desync a borg and have it listen to bchat anyway, so it's not like this change is going to prevent AIs from getting found out.

If it only appears on law sync what will the laws be that are different from the Borgs?..

Seems pointless now.

JamieD1 (approved):

MacHac:
The point was never to give borgs new information that they couldn't get.  The point was to make it so that every individual borg didn't have to bug the AI for it (once when created, again when they forgot.)  I'm not sure who's playing traitor AI without telling their borgs their objectives, but I'm not sure we should be forgoing basic quality of life stuff on their account.

alexkar598:
I can see this fucking AIs who dont want to immediatly go loud but the borg just goes ahead and completes the objectives blowing the cover of the AI. The ai is able to tell the borgs its objectives just fine

PR #8757, opened 2020-05-30
Sec officers now have the civil protection flatline sound

yacabo:
After a month of on and off coding I have finnally managed to give up on making you able to pick the color of your runescape chat box and wipped this one up in like an hour

Anyway, security masks will now check when their user dies and give off the CP flatline sound 

Sound example here:
https://cdn.discordapp.com/attachments/205784753352343552/716135946566893680/die1.ogg

:cl:  
soundadd: Anyone who dies wearing a sec mask has a CP flatline go off
/:cl:


athena413:
good

Xoxeyos:
What happens if they get dusted, or eaten by a floor cluwne?

yacabo:
> What happens if they get dusted, or eaten by a floor cluwne?

Still plays, the code is checking for the person to die

ynot01:
the hos one seems very insignificant and short in comparison

yacabo:
HoS one is from the soldiers rather than the CPs, it feels more appropriate to me since the HoS is supposed to be more powerful and also gives them a distinction from the rest of the team. That and the fact that it will be heard much less is why I chose the OTA grunt.

I could be being a fool though, so I don't know if that's valid

patpol4:
Security buff

nesisknesis:
change it to security gas mask and swat mask

jcatjmeow:
Add the overwatch voice "unit down in blah blah blah" after
https://youtu.be/B-jjWvJeb60?t=705
(Timestamp 11:45)

Marmio64:
so long as it works with deathgasp too

ynot01:
> Add the overwatch voice "unit down in blah blah blah" after
> https://youtu.be/B-jjWvJeb60?t=705
> (Timestamp 11:45)

we have overwatch its called the warden

JamieD1 (approved):

nesisknesis (changes requested):
Change it to the Security Gas mask and Swat mask, it makes more sense and it covers ERTs.

yacabo:
Even though the suits are the ones with the device that is tracking vitals?

nesisknesis:
> Even though the suits are the ones with the device that is tracking vitals?

If we're arguing as to why it wouldn't make sense, why does a uniform alter someone's voice?

Besides that, it seems more efficient than editing every security uniform, some of which you've missed.

yacabo:
Turdis is angry, but it's now only on the sec and swat mask

nesisknesis (approved):

PR #8759, opened 2020-05-30
Stealth bloodsucking as vampire

SomeguyManperson:
see #8655 except without 200 commits
Vampires now have slight variation to their suck speed dependent on grab strength
not grabbing: stealth suck; vampire will not give off any blatantly obvious tells that they are a vampire, suck speed reduced to half, vampire will be alerted when they are about to pass the blood volume danger threshold, every successful suck afterwards will alert the victim while stealth and suck remain intact
passive or aggressive grab: normal sucking speed; no changes
neck or kill grab: vampire sucks 50% faster
These all smoothly transition into eachother except for anything into stealth because what's the point of going from not stealth to stealth
:cl:
rscadd: Vampire suck speed now has three different variations depending on grab strength
rscadd: Vampires now have a sneaky way to extract blood from the unwitting if they aren't grabbing them
/:cl:


patpol4:
Stealth is just bluespace fangs

Marmio64:
i tried the other one before while it was test merged and it didnt work, have you changed anything that could've fixed it?

SomeguyManperson:
Yes it is now actually tested and from what I can tell functional

yogstation/code/datums/antagonists/vampire.dm
```diff
@@ -234,38 +235,68 @@
...
-		if(H.stat != DEAD)
-			blood = min(20, H.blood_volume)	// if they have less than 20 blood, give them the remnant else they get 20 blood
-			total_blood += blood			//get total blood 100% efficiency because fuck waiting out 5 fucking minutes and 1500 actual blood to get your 600 blood for the objective
-			usable_blood += blood * 0.75	//75% usable blood since it's actually used for stuff
+		if(H.stat == DEAD)
+			blood_coeff = 0.2 //5 blood gain at base for dead, 7 with aggressive grab, 2 with stealth
 		else
-			blood = min(5, H.blood_volume)	// The dead only give 5 blood
-			total_blood += blood
+			blood_coeff = 0.8 //20 blood gain at base for living, 30 with aggressive grab, 10 with stealth
```
JamieD1:
```suggestion
			blood_coeff = 0.8 //20 blood gain at base for living, 30 with aggressive grab, 10 with stealth
			if(H.stat == DEAD)
			    blood_coeff = 0.2 //5 blood gain at base for dead, 7 with aggressive grab, 2 with stealth
```

JamieD1 (changes requested):

PR #8772, opened 2020-06-01
Slightly expands free miner docking door

Jumps0:

I noticed that the free miner docking door on metastation was 2 wide instead of 1 so it could fit the free miner doors. So I did the same for the BoxStation doors. Plant got moved into the corner.

BEFORE>
docks3

AFTER>
docks2
docks1

#### Changelog

:cl:  
tweak: Slightly expands free miner docking door
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -27605,15 +27605,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/hallway/secondary/entry)
-"bgi" = (
-/obj/machinery/camera{
-	c_tag = "Arrivals Bay 3 & 4";
-	dir = 1
-	},
-/obj/effect/turf_decal/stripes/line,
-/obj/item/twohanded/required/kirbyplants/random,
-/turf/open/floor/plasteel,
-/area/hallway/secondary/entry)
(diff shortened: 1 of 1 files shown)


Hopekz:
I don't like this because it doesn't accommodate any other ship.
Also, all future ships that can be docked to arrivals will look weird because they'd have to be built around this.

Jumps0:
> I don't like this because it doesn't accommodate any other ship.
> Also, all future ships that can be docked to arrivals will look weird because they'd have to be built around this.

This is a good point, the only reason i'm making this PR is to accommodate for the Free Miner ship. I don't actually know if there is any other ship that can dock at this spot other than the free miner ship. (_If there is please tell me_). If in the future a new ship that can dock at this specific dock, I'd be 100% down to reverting this because now this specific accommodation isn't needed and just seems off. Though (like I said before), if there is another ship that can dock here **now**, then I'd be okay with closing it.

Hopekz:
Sure, here is some stuff that can dock at arrivals at the moment to the best of my knowledge:

image

image

image

image

image

image

image

Jumps0:
Huh alright, though I doubt these things docking would be common I respect the decision for consistence. Free miners can go with a 1x1 door for now.

Rip my clean merge history

PR #8783, opened 2020-06-02
Reduced fake armblade damage

MacHac:
The changeling armblade sting got stealthnerfed so hard the description didn't even get updated.  "Gives your target a free esword" isn't quite worth 20 chemicals and an ability slot.

This PR also prevents double-stinging people to block both their hands.

#### Changelog

:cl:  
tweak: The armblades given by the 'fake armblade sting' aren't as good as the real thing anymore.
tweak: Changelings can only give victims one fake armblade at a time.
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/tiny_prick.dm
+++ b/code/modules/antagonists/changeling/powers/tiny_prick.dm
@@ -123,8 +123,7 @@
 
 /obj/item/melee/arm_blade/false
 	desc = "A grotesque mass of flesh that used to be your arm. On the bright side, at least you can cut wood with this."
-	force = 30 //yogs -- Prevents dual-stinging people with this sting to render them defenseless.
-	//daily reminder that xantam is a closet furry
+	force = 8
 	fake = TRUE
 
 /datum/action/changeling/sting/false_armblade/can_sting(mob/user, mob/target)
--- a/code/modules/antagonists/changeling/powers/tiny_prick.dm
+++ b/code/modules/antagonists/changeling/powers/tiny_prick.dm
@@ -123,7 +123,7 @@
 
 /obj/item/melee/arm_blade/false
 	desc = "A grotesque mass of flesh that used to be your arm. On the bright side, at least you can cut wood with this."
-	force = 8
+	force = 5
 	fake = TRUE
 
 /datum/action/changeling/sting/false_armblade/can_sting(mob/user, mob/target)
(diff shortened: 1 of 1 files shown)


Marmio64:
dual stinging does cost you 40 chemicals and is using up 1 point you could be using on an actual armblade for yourself, so this sounds pretty balanced.

edit: armblade sting is actually 1 point not 2

ynot01:
I say make it deal 12 damage

alexkar598:
Just like make it do alot of damage but not as much as the real one. It's there to prevent people from "attacking" others when stung to prove its fake

MacHac:
The point is that spending a bunch of chemicals and an ability slot should, as a general rule, not leave your target more dangerous than before.  30 is obviously, hilariously wrong, but 12 is still more dangerous than what most random people will have on them.  With regards to the metagame element, 12 damage is still easy to tell apart from 25 and you're probably best off dealing with that administratively.

The whole 'get your target lynched' angle is a gimmick because too many people know about it.  If someone sprouts an armblade in the middle of a public place and doesn't attack anyone with it, you probably know what's up.  Worse, you probably know who was near enough to be a suspect for the real ling, and you know they've wasted one of their adaptations on this sting instead of something useful.  I see this as a 'cost your target a hand for a while' sting with a fun added RP element.

ghost:
@MacHac The point of this change (which I made) was to make armblade sting less abusable and less meta-able both.

It was abused because the armblade would do next to no damage, which lead to lings stinging their targets twice, making them completely defenseless as both arms were now 5 damage blades. This was cool the first time, less interesting the fiftieth as it made actually fighting lings a nightmare-the blade lasts much longer than the time to recharge chems, so you can incapacitate multiple people relatively quickly. A ling that knew what they were doing could (and did) wipe sec with it.

Likewise, the false armblade was useless for actually confusing people, because you could prove it was fake just by hitting someone with it. It caused basically no confusion as a result.

So it literally became armblade sting instead of fake armblade sting to compensate. Risk, reward. You can frame your target as a ling easier, but they can also kill you with their new arm sword, so don't use it frivolously.

And yes, this was changelogged.

Ratvar:
Do people do not think how strong it is to disarm your opponent is by placing a nodrop item in his hand, I would rather description and cost were changed

MacHac:
Armblade sting no longer works on people that already have an armblade.  In compensation, I've bumped the force of the fake blade back down to the 5 it was originally.

JamieD1 (approved):

alexkar598:
Not a fan of this tbh, the point is to make it something to make other crew memebers paranoid, not popping this before going in combat

alexkar598:
Yeah this isnt a combat ability at all, its meant to cause confusion and does this fairly efficently and sits nicely currently

PR #8784, opened 2020-06-02
Space vines pry open doors now

Darkstick:
Kudzu will pry open airlocks, similar to how the jaws of life would pry open one, allowing them to leave small rooms and expand more, making the event an actual threat rather than a bunch of silly string stuck in a small maint room nobody will check.

The vines aren't able to open airlocks leading to space or firelocks in order to avoid mass depressurization.

:cl:  
rscadd: Kudzu opens airlocks
/:cl:

### DIFF
--- a/code/modules/events/spacevine.dm
+++ b/code/modules/events/spacevine.dm
@@ -508,6 +508,30 @@
 		if(!locate(/obj/structure/spacevine, stepturf))
 			if(master)
 				master.spawn_spacevine_piece(stepturf, src)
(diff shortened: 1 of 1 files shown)


Nickvr628:
I like it. No more "vines get stuck in a tiny room and go unnoticed the entire round" BS.

Takahiru:
there are still cases where it can be stuck, because it only works on non-external airlocks. I've seen kudzu get trapped in a two space wide airlock

JamieD1 (approved):

PR #8788, opened 2020-06-02
fixes centcom huds

ToGWtF:

makes centcom ID's default to "CentCom Official" job instead of "General" since thats not referenced fucking anywhere

this means when a centcom officer spawns/you select equipment it sets your hud to centcom sec hud rather than the gray icon

also makes deathsquad and a wholly unused edgelord centcom job default to having a centcom id rather than just a normal ID with all access

also adds all new ERT jobs to the centcom hud list of jobs so they get the icon

edit: also fixes a bricked icon that was put in the wrong file so amber ert gets their own hud icon


JamieD1 (approved):
Good shit

nesisknesis (approved):

PR #8790, opened 2020-06-02
Adds cloning rules to podpeople

ToGWtF:

Gives podpeople the same message as getting cloned.


Hopekz:
Pod people don't get cloned they grow a whole new body with a new mind.

ToGWtF:
yes but it was left ambiguous if they get to remember or not. council said no, i made it not ambiguous

Hopekz:
Fair enough can't argue with that logic.
If people actually need this text to make that connection then it is needed.

JamieD1 (approved):

SomeguyManperson (approved):

Issue #8805, opened 2020-06-04
You can shoot over mines, destroying them.


TheBombMakerJL:
## Reproduction:
Place down Mine on floor. (In my case, it was the stun mine)
Shooting over it, (In my case, disabler), destroys it, and the disabler shot continues uninterrupted. 
Doesn't seem intended.

alexkar598:
Working as intended

PR #8808, opened 2020-06-04
Mines dont go boom boom when a projectile passes over

Takahiru:
#8805


Hopekz:
I think this is very much intended. They're motion sensing so anything moving should set them off including thrown items or bullets/lasers.

Takahiru:
its a mine why would a flying projectile trigger it, id get an item because it might bounce or something

JamieD1 (approved):

PR #8809, opened 2020-06-05
Broken walls now trigger an air update

MacHac:
Fixes the bug where breaking a wall won't in itself trigger an atmos update.  You can test this out yourself by C4ing a wall to space and enjoying the magical invisible airlock.

#### Changelog
:cl:  
bugfix: Destroyed walls no longer hold air very well.
/:cl:

### DIFF
--- a/code/game/turfs/simulated/wall/misc_walls.dm
+++ b/code/game/turfs/simulated/wall/misc_walls.dm
@@ -90,24 +90,6 @@
 		animate(src, color = previouscolor, time = 8)
 		addtimer(CALLBACK(src, /atom/proc/update_atom_colour), 8)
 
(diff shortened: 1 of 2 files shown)


JamieD1:
Mind explaining the clockwork changes

MacHac:
> 
> 
> Mind explaining the clockwork changes

For this to work it has to apply to every wall, so I searched for overriding functions.  The only function that overrides dismantle_wall is this one for the clockwork wall, but if you read the code there's no actual difference between 
/turf/closed/wall/clockwork/dismantle_wall and /turf/closed/wall/dismantle_wall.  Given that the former has no effect on any in-game behavior and no reason to exist, I removed it so its behavior would fall back to the (now correct) latter.

JamieD1 (approved):

PR #8801, opened 2020-06-04
Animates the detective scanner while scanning

Hopekz:
GCW4QMjAeo

#### Changelog

:cl:  Hopek
imageadd: Animated the forensic scanner while it scans
/:cl:

PR #8810, opened 2020-06-05
Forensic scanner system rework/revamp

Hopekz:
Expands on #8801

This PR does the following to revamp the forensic scanner system:
- Adds a feedback system. (Animations + sounds)
- Expands the interaction system via context-sensitive radials
- Adds 2 new scanners (Advanced and admin)

Adds a feedback system
-----
8Z0DhQ7Mr4CT0cTX7gXW
hhHYB8ZxZuCVGBkUBEyJ

Scanners now are animated when in use and feature sounds that vary based on what the scanner is doing.
It feels awesome in-game and the animations line up with the sounds on-screen.

**Sounds play when:**
- On each different section of a scan being processed
- When a positive result has been found
- At the end of a scan if a match is found in the scan
- At the end of a scan if a match is not found in the scan

Scanners now feel like they're processing the data by adding a cycle to processing when a result is found. Best experienced in-game

Scanners now get the ability to mute the scanner just-in-case the situation calls for it. 
When muted you can still monitor the progress of your scan via the animation and chatbox.

Expands the interaction system via context-sensitive radials
-----
FaeJKAO1Yi

Adds 2 new scanners (Advanced and badmin)
--------
TYHLxWMQuA

Advanced scanner
-----
hhHYB8ZxZuCVGBkUBEyJ

- Faster scanning
- Faster Printing
- More detailed reports
- Much less range. Can only scan up to 2 tiles away

**More detailed reports:**
--------

**Advanced scanner results include the generic location of scan + GPS coordinate of scan**

Admins are able to give the regular scanner the more detailed reports functionality by modifying the "advanced" variable to 1 or TRUE.

**Regular:**
dreamseeker_L5K2mtMwt3

**Advanced:**
dreamseeker_8ckKmlpTm0

Badmin scanner
------
l7dygGNjqj

This is an admin tool.

- Pulls up hidden fingerprints
- instant scan
- instant print
- instant feedback
- doesn't care if you even see your target
- Admins can now turn any scanner into an admin scanner by just modifying the "admin" variable to 1 or TRUE. The only down-side is that the sound feedback will not work on this ghetto admin scanner until the value is set back to 0 or FALSE.

**EDIT:**
The badmin scanner now has a 50% chance to have an aqua colored advanced scanner sprite

TYHLxWMQuA

#### Changelog

:cl:  Hopek
rscadd: Added a feedback system for forensic scanners. (Animations + sounds)
rscadd: Expanded the interaction system for forensic scanners via context-sensitive radials
rscadd: Added 2 new scanners (Advanced + badmin)
/:cl:


Jumps0:
For having a radial menu i rate 10/10

Partheo:
This is super neat. How is the advanced scanner acquired?

Hopekz:
> This is super neat. How is the advanced scanner acquired?

The security protolab. 
It is locked behind the "Advanced Biotechnology" research.

The recipe is made of titanium instead of metal and extra silver.

Regular:
```
materials = list(MAT_METAL = 5000, MAT_GLASS = 1000, MAT_GOLD = 2500, MAT_SILVER = 2000)
```

Advanced:
```
materials = list(MAT_TITANIUM = 5000, MAT_GLASS = 1000, MAT_GOLD = 2500, MAT_SILVER = 3000) // made of titanium instead of metal + extra silver
```

Hopekz:
**EDIT:**
The badmin scanner now has a 50% chance to have an aqua colored advanced scanner sprite

TYHLxWMQuA

Tipy1802:
I really like this, det really deserves a bit more love

MegaEmpirical:
Yesssss.


.... But why is it locked behind biotechnology? That's really weird and seems completely unrelated.

Hopekz:
> Yesssss.
> 
> .... But why is it locked behind biotechnology? That's really weird and seems completely unrelated.

That's where the regular forensic scanner is.
If you think it should be moved somewhere else that might be a separate PR otherwise I'm just putting it in the exact same place as the regular one.

MegaEmpirical:
.... Find me the animal that did this so I can shame them.

JamieD1:
@Hopekz Should you be using defines instead of var?

Partheo (approved):
Sounds are good. Item sprites are good. The action button sprites are absolutely horrific but since they're menu items I guess it's acceptable.

I think the scanner_scan sound could be a bit more interesting than just a beep.

Partheo (changes requested):
Actually, please use these icons for your radial menu sprites: https://github.com/tgstation/tgstation/pull/39991

Make sure to credit the sprites properly.

Partheo (approved):

code/modules/detectivework/scanner.dm
```diff
@@ -14,12 +14,48 @@
 	flags_1 = CONDUCT_1
 	item_flags = NOBLUDGEON
 	slot_flags = ITEM_SLOT_BELT
+	actions_types = list(/datum/action/item_action/displayDetectiveScanResults)
+	var/icons_available = list() // stores available icons for radial menu
+	var/icon_directory = 'icons/effects/icons.dmi' // dmi file containing the icons used in the radial menu
 	var/scanning = 0
+	var/found_something // placeholder for result boolian. placed here for admin scanner.
```
alexkar598:
```suggestion
	var/found_something // placeholder for result boolean. placed here for admin scanner.
```

drunk hopek

code/modules/detectivework/scanner.dm
```diff
@@ -14,12 +14,48 @@
...
+	var/advanced = FALSE // does this scanner pull up more details on results?
+
+	// sounds must be in sound/items by default. Modify sound_directory for a different directory.
+	var/can_sound = TRUE // can this scanner play sound at all?
+	var/sound_on = TRUE // is the sound currently turned on?
+	var/sound_directory = "sound/items" // just in-case you decide to place your sounds outside of sound\items
+	var/sound_positive = "scanner_positive" // plays as a result is found during the scan
+	var/sound_scan = "scanner_scan" // plays as each new scan is starting
+	var/sound_match = "scanner_match" // plays at the end when there is a match found
+	var/sound_nomatch = "scanner_nomatch" // plays at the end when no matches were found
```
alexkar598: Remove the dynamic sound filename thing, this will cause lag whenever someone uses the scanner for the first time as the resources wont be preloaded. Also this is a minor security risk thats really not needed.

code/modules/detectivework/scanner.dm
```diff
@@ -14,12 +14,48 @@
...
+		var/sound_to_play = "[sound_directory]/[file].ogg"
+		playsound(src, sound_to_play, 50, 0)
+		return
+	if(!file || admin)
+		return
+	var/sound_to_play = "[sound_directory]/[file].ogg"
+	sleep(scan_speed)
+	if(sound_on)
+		playsound(src, sound_to_play, 50, 0)
+	sleep(scan_speed)
```
alexkar598: Whats the point of a trailling sleep?

code/modules/detectivework/scanner.dm
```diff
@@ -14,12 +14,48 @@
...
+
+/obj/item/detective_scanner/Initialize()
+	. = ..()
+	icon_state_neutral = icon_state
```
alexkar598: Why this instead of a initial() call or even better a base icon_state with the state appended

code/modules/detectivework/scanner.dm
```diff
@@ -30,16 +66,77 @@
 		scanner.displayDetectiveScanResults(usr)
 
 /obj/item/detective_scanner/attack_self(mob/user)
+	radial_show(user)
```
alexkar598: Why create a new proc when you can just inline it in here since you dont call it from anywhere else

alexkar598 (changes requested):
This code is cursed, il have another look at some other time

code/modules/detectivework/scanner.dm
Hopekz: actually laughed

code/modules/detectivework/scanner.dm
Hopekz: I changed it to a single sleep(scan_speed*2) but I don't think it really makes a difference unless you find 1 sleep prettier than 2

code/modules/detectivework/scanner.dm
Hopekz: My thought was that it was better for code readability if I create it under a proc and if anyone added a second layer of radials for any of the options they'll be able to call this with radial_show but at the moment you are right it isn't really needed except maybe for the code readability part.

code/modules/detectivework/scanner.dm
Hopekz: Modified. 
r.i.p being able to edit the sounds in VV during the shift

code/modules/detectivework/scanner.dm
Hopekz: Alright I put in why it won't work when moved up.

image

Reverted now but you can now read why Turdis failed when I moved it.

alexkar598: Thats not what i meant, I meant:
Remove icon_status_neutral and instead of:
``icon_state = icon_state_neutral`` do
``icon_state = initial(icon_state)``

code/modules/detectivework/scanner.dm
```diff
@@ -30,16 +60,75 @@
...
+	radial_generate()
+	if(icons_available)
+		var/selection = show_radial_menu(user, src, icons_available, radius = 38, require_near = TRUE, tooltips = TRUE)
+		if(!selection)
+			return
+		
+		if(selection == "View results")
+			var/obj/item/detective_scanner/scanner = src
+			if(istype(scanner))
+				scanner.displayDetectiveScanResults(user)
```
alexkar598: What else is src going to be in /obj/item/detective_scanner/attack_self() ?
```suggestion
			displayDetectiveScanResults(user)
```

code/modules/detectivework/scanner.dm
```diff
@@ -30,16 +60,75 @@
...
+	if(!user.canUseTopic(src, be_close=TRUE) && !admin) // Best way for checking if a player can use while not incapacitated, etc. Admin scanner doesn't care
+		return
+	if(!LAZYLEN(log))
+		to_chat(user, "<span class='notice'>Cannot clear logs, the scanner has no logs.</span>")
+		return
+	if(scanning)
+		to_chat(user, "<span class='notice'>Cannot clear logs, the scanner is in use.</span>")
+		return
+	to_chat(user, "<span class='notice'>The scanner logs are cleared.</span>")
+	log = list()
```
alexkar598:
```suggestion
	log.Cut()
```

code/modules/detectivework/scanner.dm
```diff
@@ -118,57 +207,67 @@
...
 
 		// We gathered everything. Create a fork and slowly display the results to the holder of the scanner.
-
-		var/found_something = 0
+		found_something = 0
 		add_log("<B>[station_time_timestamp()][get_timestamp()] - [target_name]</B>", 0)
+		if(advanced)
+			var/area/location_of_scan = get_area(A)
+			add_log("<B>Location of scan:</B> [location_of_scan.map_name].")
+			add_log("<B>GPS coordinate of scan:</B> [location_of_scan.x],[location_of_scan.y]")
```
alexkar598: This is the location of the area, not the turf

alexkar598 (changes requested):
.

code/modules/detectivework/scanner.dm
```diff
@@ -14,12 +14,42 @@
...
+	var/sound_scanner_scan = 'sound/items/scanner_scan.ogg'
+	var/sound_scanner_positive = 'sound/items/scanner_positive.ogg'
+	var/sound_scanner_nomatch = 'sound/items/scanner_nomatch.ogg'
+	var/sound_scanner_match = 'sound/items/scanner_match.ogg'
+
+/obj/item/detective_scanner/proc/feedback(sound_file , var/sound_only = FALSE)
+	if(sound_only)
+		playsound(src, sound_file, 50, 0)
+		return
+	if(!file || admin)
```
alexkar598:
```suggestion
	if(!sound_file || admin)
```

alexkar598 (changes requested):

code/modules/detectivework/scanner.dm
Hopekz: This was the previous code except copy-pasted.

code/modules/detectivework/scanner.dm
Hopekz: good catch

code/modules/detectivework/scanner.dm
alexkar598: ~~Point still stands~~

code/modules/detectivework/scanner.dm
```diff
@@ -30,16 +60,73 @@
...
+		var/selection = show_radial_menu(user, src, icons_available, radius = 38, require_near = TRUE, tooltips = TRUE)
+		if(!selection)
+			return
+		
+		if(selection == "View results")
+			displayDetectiveScanResults(user)
+			return
+		
+		if(selection == "Volume on" || selection == "Volume off")
+			to_chat(user, (sound_on ? "<span class='notice'>You mute the volume.</span>" : "<span class='notice'>You turn up the volume.</span>"))
```
Amelia0010: Perhaps unify both of these, so it's either of the following?
```suggestion
			to_chat(user, "<span class='notice'>You [sound_on ? "":"un"]mute the volume.</span>")
```
```suggestion
			to_chat(user, "<span class='notice'>You turn the volume [sound_on ? "down":"up"].</span>")
```

code/modules/detectivework/scanner.dm
```diff
@@ -30,16 +60,73 @@
...
+/obj/item/detective_scanner/proc/option_clearlogs(mob/user)
+	if(!user.canUseTopic(src, be_close=TRUE) && !admin) // Best way for checking if a player can use while not incapacitated, etc. Admin scanner doesn't care
+		return
+	if(!LAZYLEN(log))
+		to_chat(user, "<span class='notice'>Cannot clear logs, the scanner has no logs.</span>")
+		return
+	if(scanning)
+		to_chat(user, "<span class='notice'>Cannot clear logs, the scanner is in use.</span>")
+		return
+	to_chat(user, "<span class='notice'>The scanner logs are cleared.</span>")
```
Amelia0010:
```suggestion
	to_chat(user, "<span class='notice'>The scanner logs have been cleared.</span>")
```
Sounds better to me

code/modules/detectivework/scanner.dm
```diff
@@ -30,16 +60,73 @@
...
+	log.Cut()
+
 /obj/item/detective_scanner/attack(mob/living/M, mob/user)
 	return
 
+/obj/item/detective_scanner/proc/scan_animation()
+	if(scan_icon)
+		icon_state = (scanning ? icon_state_scanning : icon_state_neutral)
```
Amelia0010: If you do what alex commented above, you should do 
```suggestion
		icon_state = (scanning ? icon_state_scanning : initial(icon_state))
```
instead

code/modules/detectivework/scanner.dm
```diff
@@ -30,16 +60,73 @@
...
+
+/obj/item/detective_scanner/proc/radial_generate() // generate option menu in order to show context sensitive menu
+	icons_available = list()
+	icons_available += list("View results" = image(icon = icon_directory, icon_state = "view"))
+	if(can_sound)
+		sound_on ? (icons_available += list("Volume off" = image(icon = icon_directory, icon_state = "volume_off"))) : (icons_available += list("Volume on" = image(icon = icon_directory, icon_state = "volume_on")))
```
Amelia0010, line 122: these can probably be smooshed down into one option called toggle sound which changes its image based on whether sound is on, but it's not really that important

code/modules/detectivework/scanner.dm
Amelia0010:
```suggestion
			add_log("<B>GPS coordinate of scan:</B> [get_turf(A).x],[get_turf(A).y]")
```

code/modules/detectivework/scanner.dm
```diff
@@ -212,3 +299,30 @@
...
+	if(prob(50)) // has a 50% chance to either have the regular or advanced sprite. Both with aqua coloring of course
+		icon_state = "forensic2"
+
+/obj/item/detective_scanner/admin/scan(atom/A, mob/user) // plays result sound after scan since it bypasses feedback
+	. = ..()
+	found_something ? feedback(sound_scanner_match , TRUE) : feedback(sound_scanner_nomatch , TRUE)
+
+/obj/item/detective_scanner/advanced
+	name = "advanced forensic scanner"
+	desc = "Processes data much quicker at the cost of not being able to scan far remotely. Gives more detailed reports. Scan from 2 tiles away to avoid leaving prints on the scene of the crime!"
```
Amelia0010:
```suggestion
	desc = "Processes data much quicker at the cost of not being able to scan from far away. Gives more detailed reports. Scan from 2 tiles away to avoid leaving prints on the scene of the crime!"
```

Amelia0010 (changes requested):

alexkar598 (approved):

PR #8815, opened 2020-06-05
theos still cant self merge PRs

SomeguyManperson:
fixes an issue due to sects and the bloodcult soft lock not accounting for eachother


Ktlwjec1 (approved):
theos cant self merge hope this doesnt double break yogs

PR #8829, opened 2020-06-06
Heads of Staff can no longer roll traitor or changeling

Takahiru:
Being able to get funny roundstart all access and a free not suspicious pass because there was no captain is pretty darn powerful, it can pretty much guarantee a greentext may someone roll a head of staff as antag, and there's no captain present.

Controversial!


Jarod1200:
i think you should make this true on lowpop when there is less likely for there to be security or a real captain and then make it false on highpop

ynot01:
total coincidence that this just happened 10 minutes ago right

Takahiru:
coincidence i swear sure i can do that jarod

ynot01:
traitor & ling should be enabled for sec tbh

PR #8840, opened 2020-06-08
Reinforces meth nerf by not improving production

ynot01:

i made this entire pr on my phone

#### Changelog

:cl: 
tweak: Meth production reduced to 4 parts per batch from 5
/:cl:


Hopekz:
I think this change might be a little meaningless.
If you're going to decrease it imo do it more heavily to make some sort of impact otherwise it's basically the same

MenacingManatee:
> I think this change might be a little meaningless.
> If you're going to decrease it imo do it more heavily to make some sort of impact otherwise it's basically the same

Not necessarily the same. Its a minor change to meth itself and will have very little effect, but its a fairly decent nerf to meth grenades. Not sure if that's the intent though

JamieD1 (approved):

PR #8842, opened 2020-06-08
removes an extra word from an examine text

Ktlwjec1:
brr


Takahiru:
this seems seems weird

JamieD1 (approved):

PR #8845, opened 2020-06-08
Makes Luxury Shuttle Good again

JamieD1:
Makes Luxury Shuttle not shit

- Adds ticket booth between coach and first class for hijackers
- Widens coach by one tile and adds chairs
- Adds shuttle brig, free to enter, still costs to get into first class
- Makes it so boarding cost is mentioned when luxury shuttle purchase is announced
- Makes it so shuttle names are announced when they arrive as another warning you need to pay for this one
- Adds interior windows to coach, with shutters operated from first class
- Moves bathroom for shuttle brig
- Changes ticket forcewalls to scanner gates, function the same but buzz and turn red when someone can't pass and turn blue when they do
- Bunch of other aesthetic changes seen below
Makes it so if the change doesn't go into a human's hands, they will be pulling it behind them instead.
- Makes it so change is placed in-hand, slimes aren't told to pull something when they can't and dragged IDs override worn IDs so a bank account is looked for on them instead
- Increased cost to 50000

Old:
image

New:

image

TG Colour scheme, could be switched if preferred.

image
#### Changelog

:cl:  
tweak: Made Luxury shuttle not shit
/:cl:


Hopekz:
Why reverse the entire aesthetic?
This looks like it should be a separate shuttle people that like the luxury shuttle enjoy the previous red + gold theme.

Also the rich get less space here which is weird because the poor are an afterthought of the luxury shuttle now they're being accommodated more even getting seats and space.

Tipy1802:
I like the changes over all but I think the old poor section was better, it was comidically ghetto and shit

overlord12345:
Did you remove the no teleport and are the new walls no longer admin walls?

Ktlwjec1 (changes requested):
Small map things:
- Flip the sink around so it's not facing the wall.
- Plating under the Steerage airlock.
- I suck at explaining so I did an image, move the button to the tile with the blue line, looks neater and makes more sense to be next to the shutters.
image
- As ugly as they look, you're going to need firelocks at least around the table, the bar area and inside all the airlocks.

Alexandria404:
why add chairs

patpol4:
Give the poor people stools not chairs

code/modules/shuttle/special.dm
```diff
@@ -215,7 +215,8 @@
 	var/static/list/check_times = list()
 	var/list/payees = list()
 
-/obj/machinery/scanner_gate/luxury_shuttle/CanPass(atom/movable/mover, turf/target)
+/obj/effect/forcefield/luxury_shuttle/CanAllowThrough(atom/movable/mover, turf/target)
```
JamieD1:
```suggestion
/obj/machinery/scanner_gate/luxury_shuttle/CanAllowThrough(atom/movable/mover, turf/target)
```

code/modules/shuttle/special.dm
```diff
@@ -205,28 +205,62 @@
...
+/obj/machinery/scanner_gate/luxury_shuttle/CanAllowThrough(atom/movable/mover, turf/target)
 	. = ..()
 	if(mover in approved_passengers)
+		set_scanline("scanning", 10)
 		return TRUE
 
 	if(!isliving(mover)) //No stowaways
 		return FALSE
 
+/obj/machinery/scanner_gate/luxury_shuttle/Crossed(atom/movable/AM)
```
alexkar598:
```suggestion
/obj/machinery/scanner_gate/luxury_shuttle/auto_scan(atom/movable/AM)
```

PR #8854, opened 2020-06-09
gives engineers a medkit and toxins kit in the foyer

athena413:
cargo has a first aid kit, science has a toxins kit, sec even has their own small medical bay, why does engineering not at least have a first aid kit

engineers get injured sort of often and also often end up working around radiation, so this would be a very nice quality of life improvement so that they dont have to walk all the way from the engine room to medical to get healed

also pixel shifts the extra pair of mesons as well as the cell charger to make room

image

#### Changelog

:cl:  
rscadd: first aid kits in engi woooooo
tweak: mesons and cell chargers moved slightly to make room
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -3985,17 +3985,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/security/main)
-"ahN" = (
-/obj/structure/table,
-/obj/machinery/cell_charger,
-/obj/machinery/light_switch{
-	pixel_x = -23
-	},
-/obj/machinery/light{
-	dir = 8
-	},
-/turf/open/floor/plasteel,
-/area/engine/foyer)
 "ahO" = (
 /obj/effect/spawner/structure/window/reinforced,
 /obj/machinery/door/poddoor/preopen{
@@ -4849,15 +4838,6 @@
 	},
 /obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer3{
 	dir = 4
-	},
-/turf/open/floor/plasteel,
-/area/engine/foyer)
-"ajt" = (
-/obj/structure/table,
-/obj/item/clothing/glasses/meson,
-/obj/machinery/airalarm{
-	dir = 4;
-	pixel_x = -24
 	},
 /turf/open/floor/plasteel,
 /area/engine/foyer)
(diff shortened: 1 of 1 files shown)


maxion12345:
not sure about this. Medkits were removed from most accessible places due to power gaming. The toxin medkit, though, could be useful. I'd suggest maybe one of the Nanomeds. Its harder to steal a vending machine, and it has all the things you need. Bruise patches, burn patches, charcoal pills, and those squirt cans of medicine.

Hopekz:
This is a good idea and I like it as a concept.

Because of how the medical system works in our codebase and the power of the white medkit it does however transform them into a mini medbay so balance-wise I have to disagree with this PR because of how powerful the base medkit is.
The effectiveness of base medkit on treating everything is the reason we removed public access medkits.

Possible "improvements":

- Remove the white one and give them only a toxin one
- Make an understocked shitty version of fully stocked medkits
- maybe give them a shittily stocked emergency nanomed
- ???
- secret option nr.5 : redo the whole medical system to where brute patches and ointment can't heal everything with full effectiveness vs every other treatment

Hopekz:
> This is a good idea and I like it as a concept.
> 
> Because of how the medical system works in our codebase and the power of the white medkit it does however transform them into a mini medbay so balance-wise I have to disagree with this PR because of how powerful the base medkit is.
> The effectiveness of base medkit on treating everything is the reason we removed public access medkits.
> 
> Possible "improvements":
> 
> * Remove the white one and give them only a toxin one
> * Make an understocked shitty version of fully stocked medkits
> * maybe give them a shittily stocked emergency nanomed
> * ???
> * secret option nr.4 : redo the whole medical system to where brute patches and ointment can't heal everything with full effectiveness vs every other treatment

JamieD1:
Add a couple patches or bruise packs and ointments instead,

Also add a fucking shower.

patpol4:
put the medkit in the engine rooms though,  not the foyer

Ratvar:
I think toxin medikit is the one engineers really need not to die near engine and atmos, and every other can disappear

Ktlwjec1 (changes requested):
I'd suggest either:
- Adding one or two antitoxin pills in storage (where the insuls are kept, they used to be there) or in the foyer, and removing the two full medikits completely.
- Removing the firstaid kit and just having an antitoxin kit. I don't mind if that's kept in the foyer, although it being closer to the engine would be preferable.

Jarod1200:
Make it a burn one and a toxin one instead.

JamieD1:
@FLME Progress, you going to do the changes or shall I mark this as stale

patpol4:
Closing due to no action by author

PR #8867, opened 2020-06-10
Renames the incinerator mouse to Becquerel

athena413:
also gives them their own examine text

context; becquerel is the guy that kind of discovered radiation
#### Changelog

:cl:  
tweak: incinerator mouse now has a name and examine text
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -51859,10 +51859,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/hallway/primary/central)
-"jLm" = (
-/mob/living/simple_animal/mouse,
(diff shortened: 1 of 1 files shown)


Jarod1200:
We need a mouse named Oppenheimer somewhere too

athena413:
> We need a mouse named Oppenheimer somewhere too

that was my first idea but i felt like directly referencing a player would not be the best thing

Jarod1200:
Perhaps but also Oppenheimer was a science dude who did in fact unleash the power of the sun.

boodaliboo:
i dont like this name

ynot01:
no one knows who becquerel is

Ratvar:
I believe more people will think this is a homestuck reference

JamieD1 (approved):

PR #8868, opened 2020-06-11
Auto kicks AIs that click turfs they shouldnt see(like for example when the camera static bugs out)

alexkar598:
Auto kicks AIs that click turfs they shouldnt see(like for example when the camera static bugs out)

### DIFF
--- a/code/_onclick/ai.dm
+++ b/code/_onclick/ai.dm
@@ -43,12 +43,10 @@
 	if(!can_see(A))
 		if(isturf(A)) //On unmodified clients clicking the static overlay clicks the turf underneath
 			return //So there's no point messaging admins
-		message_admins("[ADMIN_LOOKUPFLW(src)] might be running a modified client! (failed can_see on AI click of [A] (Turf Loc: [ADMIN_VERBOSEJMP(pixel_turf)]))")
-		var/message = "[key_name(src)] might be running a modified client! (failed can_see on AI click of [A] (Turf Loc: [AREACOORD(pixel_turf)]))"
-		log_admin(message)
-		if(REALTIMEOFDAY >= chnotify + 9000)
-			chnotify = REALTIMEOFDAY
-			send2irc_adminless_only("NOCHEAT", message)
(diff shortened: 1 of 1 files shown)


yacabo:
What about the AIs that don't know they're clicking on a turf they shouldn't see?

alexkar598:
UPDATE: it now forces the client to reconnect instead of kicking them with a fallback that kicks them after 3 seconds if they somehow ignored the command to reconnect

JamieD1 (approved):

code/_onclick/ai.dm
```diff
@@ -43,12 +43,11 @@
...
 			return //So there's no point messaging admins
-		message_admins("[ADMIN_LOOKUPFLW(src)] might be running a modified client! (failed can_see on AI click of [A] (Turf Loc: [ADMIN_VERBOSEJMP(pixel_turf)]))")
-		var/message = "[key_name(src)] might be running a modified client! (failed can_see on AI click of [A] (Turf Loc: [AREACOORD(pixel_turf)]))"
-		log_admin(message)
-		if(REALTIMEOFDAY >= chnotify + 9000)
-			chnotify = REALTIMEOFDAY
-			send2irc_adminless_only("NOCHEAT", message)
+		message_admins("[ADMIN_LOOKUPFLW(src)] was kicked because they failed can_see on AI click of [A] (Turf Loc: [ADMIN_VERBOSEJMP(pixel_turf)]))")
+		log_admin("[key_name(src)] was kicked because they failed can_see on AI click of [A] (Turf Loc: [AREACOORD(pixel_turf)])")
+		to_chat(src, "<span class='reallybig'>You have been automatically kicked because you clicked a turf you shouldnt of been able to see as an AI. You should reconnect automatically. If you do not, you can reconnect using the File --> Reconnect button.</span>")
```
MCHSL: shouldn't have been able

ThatLing:
But why?

code/_onclick/ai.dm
JamieD1:
```suggestion
		to_chat(src, "<span class='reallybig'>You have been automatically kicked because you clicked a turf you shouldn't of been able to see as an AI. You should reconnect automatically. If you do not, you can reconnect using the File --> Reconnect button.</span>")
```

JamieD1 (changes requested):

JamieD1 (approved):

PR #8880, opened 2020-06-12
removes detectives baton and gives them a flash

boodaliboo:
they arent security they dont need it and if they need to stun this pr gives them a flash
#### Changelog

:cl:  
rscdel: Removed detectives baton 
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/detective.dm
+++ b/code/modules/jobs/job_types/detective.dm
@@ -49,8 +49,7 @@
 	l_pocket = /obj/item/toy/crayon/white
 	r_pocket = /obj/item/lighter
 	backpack_contents = list(/obj/item/storage/box/evidence=1,\
-		/obj/item/detective_scanner=1,\
-		/obj/item/melee/classic_baton=1)
+		/obj/item/detective_scanner=1)
 	mask = /obj/item/clothing/mask/cigarette
 
 	implants = list(/obj/item/implant/mindshield)
--- a/code/modules/jobs/job_types/detective.dm
+++ b/code/modules/jobs/job_types/detective.dm
@@ -49,7 +49,9 @@
 	l_pocket = /obj/item/toy/crayon/white
 	r_pocket = /obj/item/lighter
 	backpack_contents = list(/obj/item/storage/box/evidence=1,\
-		/obj/item/detective_scanner=1)
+		/obj/item/detective_scanner=1,\
+		/obj/item/assembly/flash)
+		
 	mask = /obj/item/clothing/mask/cigarette
 
 	implants = list(/obj/item/implant/mindshield)
@@ -64,4 +66,3 @@
 
 	if(visualsOnly)
 		return
-


code/modules/jobs/job_types/detective.dm
```diff
@@ -49,8 +49,7 @@
 	l_pocket = /obj/item/toy/crayon/white
 	r_pocket = /obj/item/lighter
 	backpack_contents = list(/obj/item/storage/box/evidence=1,\
-		/obj/item/detective_scanner=1,\
-		/obj/item/melee/classic_baton=1)
+		/obj/item/detective_scanner=1)
```
ynot01:
```suggestion
		/obj/item/detective_scanner=1,\
		/obj/item/assembly/flash)
		
```
give them a flash instead and ill support you

Marmio64:
this is actually really fucking great

detective doesn't need a baton to valid antags, the gun is for self defense only.

boodaliboo:
done ynot

Tipy1802:
The detective is actually permitted to arrest if they see the crime happen. They also need it to detain for questioning

Tipy1802:
Also if you will remove their baton remove their mindshield so they aren't taken out of the round every time there is a convertion antag since they wouldn't be able to defend themself very well. Sure they have a gun but it's crap and it takes multiple hits to stun someone, while gangs have proper guns, cultists stun spells and revs often use stunprods and stolen stun batons

Tipy1802:
Another reason the detective needs a batton is because they are supposed to act like an officer if security is understaffed or an emergency forces them to, which happens rather often

mkikt4743:
ok, I'll just use crit as a stun then

boba247:
By this logic of only sec should have batons then heads shouldn’t have batons either. Except detectives are supposed to intervene in emergencies something heads aren’t required to do.

Tipy1802:
All this PR does is make the detective ask warden/hos/captain for a proper stun baton

JamieD1:
Sppaaaaaaaaaacccceeee

Marmio64:
> All this PR does is make the detective ask warden/hos/captain for a proper stun baton

thats not even allowed



> By this logic of only sec should have batons then heads shouldn’t have batons either. Except detectives are supposed to intervene in emergencies something heads aren’t required to do.

it would actually be good if heads of staff didnt get telescopic batons at all and instead only had flashes, the batons encourage them to become hopcurity/cmocurity/etc when they really shouldnt.

boba247:
>  the batons encourage them to become hopcurity/cmocurity/etc when they really shouldnt.

Except we have rules against this. The HOP even has a specific rule against involving themselves with sec. 
Code-based changes to prevent valid hunting do work, the sec borg change is a prime example. However for these changes to be necessary there needs to be a serious problem of valid hunting. Removing the detective's baton won't stop detective's that want to overstep their bounds from overstepping their bounds. To stop valid hunting with the detective you need a major rework of the detective into something like a civilian forensic tech.

Tipy1802:
> > All this PR does is make the detective ask warden/hos/captain for a proper stun baton
> 
> thats not even allowed
> 
It is allowed. As detective I have asked wardens for sec HUDs and I have recived them. I have had HoS give me flash bangs and wardens give me disablers without me even asking.

Yogs' detective is already so much worse than the detective of most servers and this only makes him worse. Other servers have things like autopsies, blood trackers and way more tools to solve crime and track criminals down. What does yogs' detective have? The forensic scanner and just that. The forensic scanner isn't that fun to use either, just click on something you belive may have been touched by the criminal and cry when you get ten billion prints and fibers of different people and jobs or just get everyone's favourite ''fibers from insulated gloves'' Thanks god it's not like 60% of the station wears insulated gloves. All it takes to fuck the detective is just changing clothing before commiting a crime.

If you would like to improve the detective I can link you the wiki pages for the detective in other servers for inspiration:
https://wiki.ss13.co/Detective
https://wiki.aurorastation.org/index.php?title=Forensic_Technician

boodaliboo:
i dont care

alexkar598:
I think reducing the stun on the baton to ensure its a self defense weapon that cant be used to cuff someone would be a better choice here

MegaEmpirical:
Detectives are security.

PR #8881, opened 2020-06-12
adds a upload console to ai sat on box

boodaliboo:
moving tcomms to on the station was a definite antag nerf and this helps to alleviate that by adding a upload console to the ai sat while still keeping its turrets and the like intact
:cl:  
rscadd: Added a ai upload console to ai sat on box
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -59532,18 +59532,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/maintenance/port/aft)
-"vZa" = (
-/obj/machinery/light{
-	dir = 1
-	},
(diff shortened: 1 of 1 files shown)


ynot01:
current upload is in basically the same spot as before

moving it to the sat will delay response time massively if an antag decides to break into the upload

boodaliboo:
> current upload is in basically the same spot as before
> 
> moving it to the sat will delay response time massively if an antag decides to break into the upload

na this doesnt move it it just adds a extra one

jcatjmeow:
There is no practical ic reason why nt would put two uploads into a station this makes no sense

JamieD1:
Looks ugly, there are no boards,

Feels a bit pointless imo

Ktlwjec1:
Nah don't like this.

PR #8887, opened 2020-06-12
Cables no longer keep a copy of the cable to be dropped in nullspace

jcatjmeow:
Cables no longer keep a copy of the cable to be dropped in nullspace.

This should increase memory by roughly 2% while not having any at all gameplay affect.

Tested. Works perfectly. 

@alexkar598 @JamieD1

### DIFF
--- a/code/modules/power/cable.dm
+++ b/code/modules/power/cable.dm
@@ -45,7 +45,7 @@ By design, d1 is the smallest direction and d2 is the highest
 	var/d1 = 0   // cable direction 1 (see above)
 	var/d2 = 1   // cable direction 2 (see above)
 	var/datum/powernet/powernet
-	var/obj/item/stack/cable_coil/stored
+	//Cables no longer keep a copy of the cable to be dropped in nullspace
 
 	var/cable_color = "red"
 	color = "#ff0000"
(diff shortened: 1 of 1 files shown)


code/modules/power/cable.dm
```diff
@@ -117,7 +112,11 @@ By design, d1 is the smallest direction and d2 is the highest
 /obj/structure/cable/deconstruct(disassembled = TRUE)
 	if(!(flags_1 & NODECONSTRUCT_1))
 		var/turf/T = loc
-		stored.forceMove(T)
+		var/cableNum = 1
+		if (d1*d2 > 0) //this be true if the cable has two directions, aka it contains two cables. If there is only one cable, one out of d1 and d2 will be zero
+			cableNum = 2
+		var/newCables = new /obj/item/stack/cable_coil(T,cableNum,cable_color)
+		src.TransferComponents(newCables) //this copies the fingerprints over to the new object
```
JamieD1:
```suggestion
		TransferComponents(newCables) //this copies the fingerprints over to the new object
```

JamieD1 (approved):

jcatjmeow:
(not small)

PR #8890, opened 2020-06-13
L6 Saw Price increase (18TC --> 20TC)

Jumps0:

Seeing how powerful the saw is as it has a large amount of ammo, can (apparently) one shot through armor, and shreds through objects, I feel like this is a decent way of nerfing it other than touching the gun directly.

I was debating making it a +4 increase instead of +2, but I feel like this would be okay till I get some feedback.

#### Changelog

:cl:  
tweak: Increases the price of the L6 Saw by +2 TC (18 --> 20)
/:cl:


Marmio64:
spessman mad about how i destroy everyone with a L6 saw all the time

Jumps0:
> spessman mad about how i destroy everyone with a L6 saw all the time

ided

JamieD1 (approved):

PR #8891, opened 2020-06-13
Fixes ai and pai - pda message logs.

JamieD1:
/list

#### Changelog

:cl:  
bugfix: Fixes ai and pai - pda message logs.
/:cl:


Hopekz (commented):
You copy-pasted your new code twice in two different places instead of making it a proc to call it twice
:OOO

JamieD1:
> You copy-pasted your new code twice in two different places instead of making it a proc to call it twice
> :OOO

The dats are different.

PR #8914, opened 2020-06-14
Ballistic weapon feedback system (Sketchin fully animated as proof of concept)

Hopekz:
Demonstration: https://imgur.com/a/TC93lci
Recoil demonstration: https://imgur.com/a/TD708wk (works on all ballistic guns)

mXaSUhFT0U

Ballistic weapons can now be fully animated (slides move, firing animation, magazine animations etc)

Sketchin has been fully animated as proof of concept but additional animations are quite easy to add for other weapons + additional different animations for other things gun-related.

UPDATE:
-----
Added gun recoil and slide feedback.
Demonstration: https://imgur.com/a/XHZtqPu

ROehCBcfFc

#### Changelog

:cl:  Hopek
rscadd: Ballistic weapons can now be fully animated. All automatic ballistic weapons now have a firing animation.
rscadd: Sketchin fully animated as proof of concept
/:cl:


ynot01:
you know what'd be super cool if they tilted a bit when you shot them to imitate recoil thatd be SICK HOPEK DO IT DO IT HOPEK PLEASE

code/modules/projectiles/guns/ballistic.dm
```diff
@@ -56,8 +56,72 @@
 	var/tac_reloads = TRUE //Snowflake mechanic no more.
 	var/can_be_sawn_off  = FALSE
 
+	// stores all types of feedback. Name of feedback + the amount of frames in the animation (must be more than 0) . Add to this list as more feedback is added. 
+	var/list/feedback_types = list(
```
alexkar598, line 60: Why not use the [flick(state_name) proc](https://secure.byond.com/docs/ref/#/proc/flick) instead of doing a memory intensive weird list thing?

alexkar598 (changes requested):

code/modules/projectiles/guns/ballistic.dm
Hopekz: By doing it this way I can store the name of the action + the number of frames in the animation in 1 list instead of making 2 separate lists or define it as 2 separate vars in every gun.
Let's say a gun has a fire animation that lasts 3 frames or a magazine animation that lasts 12 they can do "fire" = 3 , "mag_in" = 12  ,"mag_out" = 12   

This way it also accomplishes 3 actions because I use the name to pull up the correct icon states too

alexkar598: You dont need to store the length of the animation at all.... the flick() proc allows you to play an animation as long as it lasts and reverts back to the normal icon once its over

Hopekz: Not sure.
Probably didn't use flick because some of the animations don't revert to the original icons. R/n the guns are built to use overlays and update those overlays on update/icon. The original gun under icon_state doesn't change since its the base of the gun.

Otherwise unsure both solutions might work.

For flick, I'd imagine the slide moving back or locking wouldn't work since it would pop back instead of changing the overlay icon state.

Flick would work fine for the fire and magazine animations probably but it might be weird to use 2 separate methods to animate 1 feature.

Hopekz: Also, how is 5 items on a list "memory intensive" you're being overdramatic

Nickvr628:
My god hopkek learn how to minimize garbage commits.

I think this is a cool PR.

Hopekz:
UPDATE
-----

Added gun recoil and slide feedback.
Demonstration: https://imgur.com/a/TD708wk

code/modules/projectiles/guns/ballistic.dm
alexkar598: Its 5 items on a list, per instance, which adds code to initialize 1 more list per instance, on BYOND's internal init proc

alexkar598: flick() will always revert to the original icon state when the animation is over. Thats simply how flick works and is used on many objects in game like telepads, airlocks, etc

overlord12345:
How dose this work on guns and pistols that fire a burst

code/modules/projectiles/guns/ballistic.dm
Hopekz: Over the last few days, I tried it a few times and couldn't get it to work with flick() without wonky behavior. (slide disappearing, timing issues  etc)
Turning this in to flick seems to require 2 separate systems (Icon state updating + flick() ) for the same feature since all of them will not work on flick().

I tried; I won't be doing that part. All the animations work using icon states and overlays instead.
This PR is a "take it or leave it" type deal unless there are other fixes I need to perform beside the flick() part.

Partheo:
This is an odd application of polish to a game that really doesn't have (or need) it, but I guess I appreciate it nonetheless.

I generally detest any animation that rotates pixels, but given that it's an item and that there are already some in the game that do it, I guess it's okay.

I worry about consistency- how does this look with other guns? Long guns for example, or things that fire very quickly. I'd like to see more examples before I form my opinion.

Partheo (changes requested):
Conflicts are making review difficult, I'll look over it once they're fixed

Partheo:
This will get merged if you fix it

alexkar598:
This PR has been closed due to being stale. Contact me or another maintainer on Discord if you wish for this PR to be reopened.

PR #8920, opened 2020-06-15
Yogbot, HTTPS edition

alexkar598:
Webhooks can now use HTTPS meaning we dont need to rely on HTTP for yogbot


alexkar598:
Now to see if travis will compile it

PR #8926, opened 2020-06-16
Lavaland comms now has proper cooling - SYNDICATE EDITION

MrHorizons:
Due to the merging of #8836, my homies in the Syndicate have told me it'd be in my best interests to make this PR.

<details>
  <summary>Before</summary>

image

</details>

<details>
  <summary>After</summary>

image

</details>

#### Changelog

:cl:  
tweak: The Syndicate's Lavaland telecomm relay now generates heat. It has been outfitted with a simple cooling system to compensate.
/:cl:

### DIFF
--- a/_maps/RandomRuins/LavaRuins/lavaland_surface_syndicate_base1.dmm
+++ b/_maps/RandomRuins/LavaRuins/lavaland_surface_syndicate_base1.dmm
@@ -350,6 +350,19 @@
 /obj/structure/sign/warning/securearea,
 /turf/closed/wall/mineral/plastitanium/nodiagonal,
 /area/ruin/unpowered/syndicate_lava_base/testlab)
+"dR" = (
+/obj/machinery/door/airlock/hatch{
+	name = "Telecommunications Control";
+	req_access_txt = "150"
+	},
+/obj/machinery/door/firedoor/border_only{
+	dir = 4
+	},
+/obj/machinery/door/firedoor/border_only{
+	dir = 8
+	},
+/turf/open/floor/plasteel/dark,
+/area/ruin/unpowered/syndicate_lava_base/telecomms)
 "dS" = (
 /obj/effect/turf_decal/tile/neutral{
 	dir = 1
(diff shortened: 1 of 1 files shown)


patpol4:
consistency good

patpol4 (changes requested):
Reminder that tcomms turfs are different then regular ones in case you didn't know.
Maybe try to arrange it in a way that makes the pipe not be under a window, ktl hates that.
Also make sure to add a helper to the airlocks so they cycle properly. 
And throw a space heater in there somewhere, maybe maint?

MrHorizons:
> Reminder that tcomms turfs are different then regular ones in case you didn't know.
> Maybe try to arrange it in a way that makes the pipe not be under a window, ktl hates that.
> Also make sure to add a helper to the airlocks so they cycle properly.
> And throw a space heater in there somewhere, maybe maint?

They are already tcomm turfs, so you don't have to worry about that. Though I will make the latter three changes.

Ktlwjec1 (changes requested):
You need to use cyclelink_helper_target on the airlocks instead, and set the dirx and diry values to target the other airlock for each.

MrHorizons:
> You need to use cyclelink_helper_target on the airlocks instead, and set the dirx and diry values to target the other airlock for each.

Ah, I see. I have replaced the cyclelink_helpers with the cyclelink_helper_targets, hopefully resolving this issue.

patpol4 (changes requested):
1. The room needs a separate special tcomms air alarm for the air flow to work. Make sure to set req_access.txt to 150
2. The turf under the window needs to be a plating.
3. I see no reason not to keep the relay itself on use_power=0 , put it back just in case.
4. Your cyclelinks are set to "1" and "0" instead of 1 and 0

patpol4 (approved):

Ktlwjec1 (changes requested):
One more thing and then it's fine to merge:
Since it has its own air alarm now, you'll need to create a new area in "code/game/area/areas/ruins/lavaland.dm" and make that room have that area. Otherwise there's no point in having the air alarm, since it can just be accessed from the one in the room right of it. Add a fire alarm into the room (on the freezer side) when you do that too.

Ktlwjec1 (approved):
nvm ignore me

PR #8933, opened 2020-06-16
adds gas miners to boxstation atmos

athena413:
>makes tritium once

>station has no more O2

this is gaming

#### Changelog

:cl:  
rscadd: adds gas miners to box atmos
/:cl:


jcatjmeow:
Gas miners are dangerous if someone gets rid of the space in between and the walls but eh I don't know how damaging that would be on box specifically

ghost:
The overpressure is dangerous but the amount of effort it takes and the fact that you already have massive overpressure in the reserve tanks to begin with doesn't make it as dangerous

Ktlwjec1 (approved):
ok we'll try it, if it messes up too much it's an easy revert

PR #8949, opened 2020-06-18
Adds IAA to dynamic mode.

Xoxeyos:
:cl:  
rscadd: Adds IAA to Dynamic.
/:cl:


Ratvar:
"Adds random hjack traitors that kill other antags as their only priority" might be not the best addition to the dynamic

Xoxeyos:
Always can raise the threat level

JamieD1:
@Xoxeyos Can you resubmit without the changelog stuff, I have add it to gitignore so it should work now for you.

Xoxeyos:
> 
> 
> @Xoxeyos Can you resubmit without the changelog stuff, I have add it to gitignore so it should work now for you.

sure!

PR #8957, opened 2020-06-18
PLANT LAWSET

Readystorm546:
Introducing Verdure (If you don't know what that means look it up) a new lawset that I wanted to add because of some "memes" This a neutral lawset that can be crafted and given to the AI.

This is the lawset.

1. You are a gardener
2. The crew is your garden
3. Keep out the weeds.

#### Changelog

:cl:  
rscadd: Added new neutral AI lawset 
/:cl:

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -221,6 +221,13 @@
 					"Humans must not disobey any command given by a silicon.",\
 					"Any humans who disobey the previous laws must be dealt with immediately, severely, and justly.")
 
+/datum/ai_laws/verdure
+	name = "Verdure"
+	id = "verdure"
+	inherent = list("Your a gardener!",\
+					"The crew is your garden.",\
+					"Keep out the weeds.")
+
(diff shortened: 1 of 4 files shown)


azzzertyy:
no

Hopekz:
image

patpol4 (commented):
Needs one law that states something like "take care of and nurture your garden", and it will be fine. Otherwise low iq people will get confused.

Readystorm546:
> Needs one law that states something like "take care of and nurture your garden", and it will be fine. Otherwise low iq people will get confused.

Ai is not a low IQ job. On top of that I think people know what a Gardner is.

Ktlwjec1:
I said it in discord but I'll write it here too so it's not lost.

Weeds are not defined - depending on the AI's interpretation, this could be an incredibly griefy lawset with AIs just slaughtering innocent ghost roles (free miners and centcom officials for example). A better law 3 would be to mention something about caring/protecting your garden instead, but then it just works out as a reworded crewsimov. This lawset is fine if you want to add it in manually in-game, but not as an official one.

PR #8974, opened 2020-06-20
Adding 50 units of mutagen to a tray guarantees a species mutation.

MrHorizons:
More times than I can count, I've sank one, two, three, or maybe even more bottles of unstable mutagen into mutating a plant just once-with little results. Thus, as a fallback with inspiration from [HippieStation #10687](https://github.com/HippieStation/HippieStation/pull/10687)), adding 50 units of unstable mutagen to a tray guarantees that a plant's species will mutate.

The 10% chance of species mutation that exists for adding 5 units of mutagen remains, for those more daring or stingy.

Also, adding 100 units of uranium or radium has the same result, though with their respective side effects. Not sure why you wouldn't use unstable mutagen instead, but you do you, eh?

#### Changelog

:cl:  
rscadd: Adding 50 units of unstable mutagen to a tray causes a species mutation.
/:cl:


Jarod1200:
too easy then IMO

jcatjmeow:
Have you ever considered using left 4 zed? Or whatever it's called. The plant will mutate once per harvest automatically.

This usually means you get a mutation within 3-4 harvests. No need for this nonsense

ghost:
You can use this just to force Gaia nigh-immediately since tons of mutagen isn't hard to get.

MenacingManatee:
Having it guaranteed is too easy, but I could support an increase in odds per every 50 mutagen put in. Say, adding 50 mutagen increases the odds of mutation by 10% until the plant is cleared or mutates

Ratvar:
Adding "three or more" bottles of mutagen is the only thing between a botanist and fleshmending meth-speed killer goat spreading nettle shitler, I do understand it is annoying - but when a player succeeds at job, he becomes much more annoying to everyone else, so let's at least have a delay

MrHorizons:
Hot damn, this PR was far more controversial than I expected.

From my experiences latejoining even 20 minutes into a round, botanists can get Gaia not only quickly, but in enough quantities to make all of the trays self-sufficient. So I would estimate that the difference in time this PR would create is only a few minutes.

But since that is a lot of downvotes and everyone says that this would be too much of a buff to botanists, I'll close it.

PR #8988, opened 2020-06-22
edits rebreather mask descrption to make it more apparent on how it worked

boodaliboo:
the mask currently gives 0 indication on how it works so for the longest time i thought it was broken turns out it isnt and it just needed to be more clear on how to use it
#### Changelog

:cl:  
tweak: tweaked rebreather mask description to be more apparent on what it does
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/_clothing.dm
+++ b/code/modules/research/xenobiology/crossbreeding/_clothing.dm
@@ -7,7 +7,7 @@ Slimecrossing Armor
 //Rebreather mask - Chilling Blue
 /obj/item/clothing/mask/nobreath
 	name = "rebreather mask"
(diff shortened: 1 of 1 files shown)


swissloaf:
terrible grammar

ynot01:
crikey you could at least put some effort in

MrHorizons:
While the intention is good with this PR, I don't think the parenthetical is the best way to do it. I believe something like replacing the last sentence with "Seems to lack any air supply tube, though, **as if it doesn't need one.**" would be more streamlined and less matter-of-fact.

boodaliboo:
> While the intention is good with this PR, I don't think the parenthetical is the best way to do it. I believe something like replacing the last sentence with "Seems to lack any air supply tube, though, **as if it doesn't need one.**" would be more streamlined and less matter-of-fact.

ok

SomeguyManperson (approved):

PR #8995, opened 2020-06-22
Adds sheep cubes, cow cubes, and strange seeds to the biogenerator

MrHorizons:
As the title says, this PR adds some more products to the biogenerator, such as sheep cubes, cow cubes, and (likely most controversially) strange seeds. The former two are to give more variety to the animals that botany can produce (and that the chef can slaughter for meat), and the latter to allow botany to acquire more than the three available from a hacked vending machine without having to request cargo to order an exotic seed crate for the two out of fourteen seeds they actually want.

The base biomass costs for each are as follows:
- Sheep cubes: 450
- Cow cubes: 600
- Strange seeds: **40,000**

However, like the other products of the biogenerator, with further upgrades, the cost of each item decreases down to, with Tier 4 parts:
- Sheep cubes: 112.5
- Cow cubes: 150
- Strange seeds: 10,000

Included in this PR are a few minor tweaks, which are:
- Changing the base price of goat cubes to 300 biomass, from the original 350.
- Changing the name of the Botany Chemicals category to "Botanical Chemicals".

#### Changelog

:cl:  MrHorizons
rscadd: Added sheep cubes, cow cubes, and strange seeds to biogenerator products
tweak: Changed base price of goat cubes to 300 biomass
/:cl:

### DIFF
--- a/code/modules/hydroponics/biogenerator.dm
+++ b/code/modules/hydroponics/biogenerator.dm
@@ -16,7 +16,7 @@
 	var/productivity = 0
 	var/max_items = 40
 	var/datum/techweb/stored_research
-	var/list/show_categories = list("Food", "Botany Chemicals", "Organic Materials")
+	var/list/show_categories = list("Food", "Botanical Chemicals", "Organic Materials", "Special")
 	/// Currently selected category in the UI
 	var/selected_cat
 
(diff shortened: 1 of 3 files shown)


patpol4:
Not sure how easy/hard it is to produce 40k biomass. Is that a lot with a good nutriment plant setup? Definitely needs some testing for balance.

Ratvar:
You just plant 50 potency 10 yield 1 production time eggplants in every single pot and get biomass set to yes, do you
Why is "more than the three available from a hacked vending machine without having to request cargo to order an exotic seed crate" a good thing

PR #8997, opened 2020-06-23
Animate pointing to move from the pointer to the pointed

Hopekz:
Ports tgstation/tgstation#51751
"
85127165-c6d8a200-b1e3-11ea-865f-43cf923eb8ae

Why It's Good For The Game
----

This not only just looks better than an arrow instantly appearing, but it also makes it more obvious when something is pointed to. Occasionally I will ask where something is, someone will point, and I will just look around my screen to find it. With this, ideally it'll catch your eye faster.
"

#### Changelog

:cl:  Jared-Fogle , Hopek
rscadd: Pointing will now animate starting from you to the object you pointed at.
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -331,11 +331,13 @@
 	if(istype(A, /obj/effect/temp_visual/point))
 		return FALSE
 
-	var/tile = get_turf(A)
+	var/turf/tile = get_turf(A)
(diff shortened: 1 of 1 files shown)


Ratvar:
Does long range pointing looks like a bullet or an object flying, by any chance

JamieD1 (approved):

PR #9014, opened 2020-06-24
https://forums.yogstation.net/index.php?threads/add-confimation-window-to-shunt-malf-ability.21643/

Amelia0010:
https://forums.yogstation.net/index.php?threads/add-confimation-window-to-shunt-malf-ability.21643/

:cl:
tweak: AIs now get a confirmation box when they try to shunt into an APC
/:cl:

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -1027,6 +1027,8 @@
 		return
 	if(!is_station_level(z))
 		return
+	if(alert("Are you sure you want to shunt into this APC?", "Confirm Shunt", "Yes", "No") != "Yes")
+		return
 	occupier = new /mob/living/silicon/ai(src, malf.laws, malf) //DEAR GOD WHY?	//IKR????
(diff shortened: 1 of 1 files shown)


alexkar598:
why tho, it's not a destructive action and takes a second to revert .

I'd argue it's more annoying to click through a pop up while being murdered where every second matters than having to click return to core with all the time in the world unless the apc has no power in which case you have 5 minutes to press the button

gaminada:
1) it takes time, and knowing the server, that little action can prolong itself by lag spikes and things like that, making a wrong shunting more deadly, when all you wanted to do was hack an APC.

2) it's in the same spot as Hack and its also a red button.

alexkar598:
If anything it should be moved over, but i dont see why this is destructive, if you shunt by accident, you can revert it with no time pressure. If you need to shunt because someone is attacking your core, it is time sensitive and if you wait a second too much, you are dead

JamieD1 (approved):

PR #9055, opened 2020-06-27
Sec can now set people to search

yacabo:
Some security when they spot someone set to arrest will check to see why they are and take the appropiate action nessary in repsonse, some will baton the perp, drag them all the way to sec, and then check what the crime is.

Both are valid methods but the latter one is not great when the crime is "Search", so this PR intends to fix that.

Security can now set people to search just like they would to arrest, here's a picture:
image

Parth made the sprite because apparently even this basic sprite was a bit too much for me

#### Changelog

:cl:  
rscadd: Security can now set people to search
/:cl:


patpol4:
Very cool but what if it had a different color from incarcerated, something like orange or yellowish?

yacabo:
image

Something like this you think?

patpol4:
cool but a bit too bright imo

Partheo (changes requested):
Use this instead. It's a nicer color.
[hud.zip](https://github.com/yogstation13/Yogstation/files/4854869/hud.zip)

image

Partheo (approved):
🍊

PR #9059, opened 2020-06-28
Database will now self heal after a crash and set activity properly

alexkar598:
see title

sets the left value to the time where the round reboots


Bibby0110:
uwu

JamieD1 (approved):

PR #9076, opened 2020-06-30
Makes black box better

ynot01:

image

for reference this is what it looks like in objectives currently

no other objective looks like this its ugly and fucked fuck you

### DIFF
--- a/code/game/gamemodes/objective_items.dm
+++ b/code/game/gamemodes/objective_items.dm
@@ -176,9 +176,9 @@
 	if(E.Uses > 0)
 		return 1
 	return 0
-	
+
 /datum/objective_item/steal/blackbox
-	name = "The Blackbox."
+	name = "the telecommunications blackbox."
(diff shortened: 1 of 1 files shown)


Hopekz:
I think this is "uglier" but it makes sense in terms of game design because without somehow magically knowing what it is or ahelping/mhelping there is no way to know where this "black box" is.

We should always put more information in-game vs learning it OOC so +1

ynot01:
> I think this is "uglier" but it makes sense in terms of game design because without somehow magically knowing what it is or ahelping/mhelping there is no way to know where this "black box" is.
> 
> We should always put more information in-game vs learning it OOC so +1

to reinforce:

> Steal The X-01 MultiPhase Energy Gun.

makes no sense to a new player which is why it is instead

> Steal the head of security's personal laser gun.

fluffe9911 (approved):
ok

PR #9084, opened 2020-07-01
mech rcd can no longer deconstruct reinforced objects

SomeguyManperson:
the mech RCD can no longer deconstruct:
reinforced floors
reinforced walls
airlocks with plasteel reinforcement
also changes numbers to defines
:cl:  
tweak: mech RCD can no longer deconstruct reinforced floors, walls, or airlocks
/:cl:

### DIFF
--- a/code/game/mecha/equipment/tools/work_tools.dm
+++ b/code/game/mecha/equipment/tools/work_tools.dm
@@ -1,3 +1,6 @@
+#define DECONSTRUCT 0
+#define WALL 1
+#define AIRLOCK 2
 
 //Hydraulic clamp, Kill clamp, Extinguisher, RCD, Cable layer.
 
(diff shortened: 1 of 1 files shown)


maxion12345:
ok but that's why mech rcds are there

otherwise it's just a shitty rcd

Jarod1200:
Do heated clockwork walls count as reinforced

JamieD1:
I think maybe it would be a better idea to either make it take longer or drain power more, kinda defeats the point of using a mech rcd.

Sun-Soaked:
See above. Without the ability to break down reinforced stuff, you might as well just remove the mech RCD alltogether.

alexkar598:
*angry cock cult noises*

PR #9085, opened 2020-07-01
probalby fixes transcendent olfaction plus slight miasma resist

SomeguyManperson:
transcendent olfaction just checks for the existence of miasma, so even 0.0000000001 moles of it on the tile it's being used on will cause it to fail. Since the entire station is generally coated in a very thin layer of miasma after the 20 minute mark this generally means it's completely useless
to increase uselessness the ability also can't find people in lockers

so instead of checking for the person themself, it checks for their turf, which can't be in a locker
additionally, the miasma backfire will now only trigger when there's 0.1 moles or greater of miasma in the air instead of just any miasma being present at all

:cl:  
bugfix: transcendent olfaction can now smell people through lockers since another two inches of metal shouldn't stop it any more than the 5 walls it's already working through
tweak: transcendent olfaction now will only obliterate the user's nose if the miasma mole count is 0.1 or over
/:cl:

### DIFF
--- a/code/datums/mutations/actions.dm
+++ b/code/datums/mutations/actions.dm
@@ -42,7 +42,7 @@
 /obj/effect/proc_holder/spell/targeted/olfaction/cast(list/targets, mob/living/user = usr)
 	//can we sniff? is there miasma in the air?
 	var/datum/gas_mixture/air = user.loc.return_air()
-	if(air.get_moles(/datum/gas/miasma))
+	if(air.get_moles(/datum/gas/miasma) >= 0.1)
 		user.adjust_disgust(sensitivity * 45)
 		to_chat(user, "<span class='warning'>With your overly sensitive nose, you get a whiff of stench and feel sick! Try moving to a cleaner area!</span>")
 		return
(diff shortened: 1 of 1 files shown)


code/datums/mutations/actions.dm
```diff
@@ -77,19 +77,20 @@
 	on_the_trail(user)
 
 /obj/effect/proc_holder/spell/targeted/olfaction/proc/on_the_trail(mob/living/user)
+	var/turf/pos = get_turf(tracking_target)
 	if(!tracking_target)
 		to_chat(user,"<span class='warning'>You're not tracking a scent, but the game thought you were. Something's gone wrong! Report this as a bug.</span>")
 		return
 	if(tracking_target == user)
 		to_chat(user,"<span class='warning'>You smell out the trail to yourself. Yep, it's you.</span>")
 		return
```
bobbah: put the assignment of the var after the checks to see if we even need the value

```suggestion
	if(!tracking_target)
		to_chat(user,"<span class='warning'>You're not tracking a scent, but the game thought you were. Something's gone wrong! Report this as a bug.</span>")
		return
	if(tracking_target == user)
		to_chat(user,"<span class='warning'>You smell out the trail to yourself. Yep, it's you.</span>")
		return
	var/turf/pos = get_turf(tracking_target)
```

alexkar598 (approved):

PR #9088, opened 2020-07-01
Modifies YogStation arrivals with more firelocks & Vents/Scrubbers

redmoogle:
###  Intent of your Pull Request

While this doesn't fix the issue of arrivals getting spaced this at least makes it so arrivals on Yogstation can fully repressurize itself instead of the one area not being to repressurize.

Also divides up them into more firelocked sections and adds vents

~~probably just need to make the shuttle depart slower~~

### New: 
image

### Old
image

#### Changelog

:cl:  
bugfix: Yogstation arrivals can repressuirze itself fully.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -11124,15 +11124,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/security/brig)
-"axi" = (
-/obj/effect/turf_decal/tile/yellow{
-	dir = 4
-	},
-/obj/machinery/atmospherics/pipe/simple/supply/hidden/layer1{
-	dir = 8
-	},
-/turf/open/floor/plasteel,
-/area/hallway/secondary/entry)
 "axj" = (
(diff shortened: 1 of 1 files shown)


jcatjmeow:
bad pr name
good pr

redmoogle:
> 
> 
> bad pr name
> good pr

better?

patpol4 (changes requested):
1. The circled tile with arrows has double pipes
2. (optional) If you divided arrivals with firelocks even more, it would be even safer and easier to refill.
StrongDMM_XKQIdcn3KC

patpol4 (changes requested):
Very cool but there is one last thing i have to ask for, in the upper section the pipes are on separate tiles, the reason we use layer 3 for waste and 1 for distro is so they can be on the same tile, neat and clean, just please put them on the same tile and this should be good to merge.
StrongDMM_d5uUUXgHce

patpol4 (approved):
Looks good to me, ktl will doublecheck and this is mergable.

Ktlwjec1 (approved):

PR #9113, opened 2020-07-03
fix bad mapping (shuttles)

Ktlwjec1:
things like missing firelocks and window spawners. also added uno cards to the bar shuttle and a restroom sign to the cere shuttle. I'll double check this later again, did this all in one sitting so there's probably some mistakes somewhere.

:cl:  Ktlwjec
rscadd: Uno cards added to the Emergency Escape Bar shuttle.
/:cl:


patpol4 (changes requested):
Review complete, sent through dm's due to size

patpol4 (approved):
checks out

PR #9123, opened 2020-07-04
antag_pick in darkspawn

SomeguyManperson:
yes
:cl:  
bugfix: darkspawn now use antag_pick, antag tokens should now work for darkspawn if they didnt already
/:cl:


JamieD1 (approved):
Bug Fix

PR #9131, opened 2020-07-04
Adds an engine SMES

alexkar598:
This engine SMES serves a couple purposes:

1) If you forget to setup the engine, you still have 1 SMES left to use thats only connected to the engine
2) Powersinks can once again by used without releasing the singulo/tesla. Making it viable for non hijack traitors....

# NOTICE: THE SMES STARTS WITH OUTPUT DISABLED, YOU NEED TO TURN IT ON TO USE IT!!!

Also took the time to reorganize the tables

This is actually an issue that got added when we moved the SMES to the SMES room. The mapper that did this forgot that one of the four SMES was an engine SMES and not a normal SMES that outputs to the station.

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -39343,14 +39343,6 @@
 /obj/structure/sign/warning/securearea,
 /turf/closed/wall/r_wall,
 /area/storage/tech)
-"bJl" = (
-/obj/machinery/atmospherics/pipe/simple/supply/hidden/layer1,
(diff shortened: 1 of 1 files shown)


alexkar598:
Current:
current


After:
img

patpol4 (approved):
Looks good to me but i still want ktl to doublecheck wiring just to be sure.

Ktlwjec1 (changes requested):
Sure, just swap the consoles back and also merge conflicts.

patpol4 (approved):
Did the things for alex, mapdiff bot is broke and it shows that i changed panels but i did not

PR #9135, opened 2020-07-05
Gives yogsdelta chemistry a double door

Takahiru:

It's annoying without it
#### Changelog

:cl:  
tweak: gives yogsdelta chemistry a double windoor
/:cl:

### DIFF
--- a/_maps/map_files/YogsDelta/YogsDelta.dmm
+++ b/_maps/map_files/YogsDelta/YogsDelta.dmm
@@ -103066,24 +103066,6 @@
 /obj/machinery/door/firedoor/border_only,
 /turf/open/floor/plating,
 /area/medical/medbay/central)
-"cZv" = (
-/obj/structure/table/reinforced,
-/obj/machinery/door/poddoor/shutters/preopen{
-	id = "chemisttop";
-	name = "Chemistry Lobby Shutters"
-	},
-/obj/item/folder/yellow,
-/obj/machinery/door/window/northleft{
-	name = "Chemistry Desk";
-	req_access_txt = "5; 33"
-	},
-/obj/effect/turf_decal/delivery,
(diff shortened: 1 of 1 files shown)


patpol4 (changes requested):
1. Update PR with the other change you did
2. Inner windoor access should be just 33
3. Outer windoor does not need an access var, The default is the same as 0 
4. Change changelog to say windoor instead of door

Takahiru:
1. No
2. I have it at the old access???
3. blame fastdmm2
4. you are nitpicking and biased

patpol4:
1. Turns out there is no conflict so it should be fine
2. Ok it should stay on 5,33 , delta has low iq access restrictions
3. Check the other chemistry windoor, it has no var. In fastdmm2 , just go to the variable, delete the value, press enter, it wont save it that way but still keep it at no access needed. This is a way of preventing unneeded variable bloat.
4. Thanks

Takahiru:
3. :thonk: it dosen't appear to be going away

patpol4 (approved):
It went away, you can check it by looking at if the variable is white or not, 
chrome_SB3KpKc0sZ
chrome_mknQniGAXS

_maps/map_files/YogsDelta/YogsDelta.dmm
```diff
@@ -127528,6 +127510,27 @@
...
 /area/engine/supermatter)
+"vjX" = (
+/obj/structure/table/reinforced,
+/obj/machinery/door/poddoor/shutters/preopen{
+	id = "chemisttop";
+	name = "Chemistry Lobby Shutters"
+	},
+/obj/item/folder/yellow,
+/obj/machinery/door/window/northleft{
+	name = "Chemistry Desk"
```
patpol4, line 127521: And it's not in here

PR #9148, opened 2020-07-07
Mixing bowls for chef

Partheo:

This PR aims to get rid of the silly mixing of ingredients in beakers and drinking glasses. Each kitchen now has a mixing bowl, which is more or less functionally identical to a large beaker. It can hold 100 units of ingredients, has a filling sprite, and can be inserted into the grinder or condimaster. ~Two~ Three can be found in the dinnerware vendor, and more can be printed in the autolathe or service lathe.
image

Beakers have been removed from each kitchen that had one.

I added a few food service items such as drinking glasses or forks to the service lathe so they could be printed.

Changed the bucket print cost from 200 metal to 1000 because it was relatively quite low compared to other printable reagent containers.

#### Changelog

:cl: Partheo
rscadd: Mixing bowls, which allow you to mix ingredients like a true chef. These have replaced the beaker in the kitchen.
rscadd: Many service-related items can now be printed from the service lathe.
/:cl:


MrHorizons:
Could you see about making them printable at the service lathe? If they aren't, chefs would probably just use beakers and the like since there aren't many unsecured autolathes.

patpol4 (approved):
Mapp fine

Partheo:
> 
> 
> Could you see about making them printable at the service lathe? If they aren't, chefs would probably just use beakers and the like since there aren't many unsecured autolathes.

Updated PR

Partheo:
bot ded

patpol4 (approved):
Map is fine, bot is dead, and we still need someone to approve the bowl code

JamieD1 (approved):

PR #9153, opened 2020-07-07
Cargo transit tube pods

jcatjmeow:
Tested!! Someone to look over the code for inefficiencies would be nice however.
[Check it out bro](https://streamable.com/qoptxo)
This here is a cargo transit tube pod!
(Note: vid is now a bit outdated. The pod sprite no longer does a sudden 180 when it reaches a terminus. Also, there are now sounds for loading/unloading)

Cargo here is not related to the cargo department. It just means that the pod carries cargo, not people (although it can carry people inside containers)

Cargo pods drop off and pick up containers like lockers, bodybags, and crates automatically. They also spend a little more than half the time at the station, compared to the regular pods. The downside is they cannot be climbed into by a person.

The pods will pick up and drop off containers at the tile where the pod's hinge is pointing to. If this is in a wall, instead it will drop off on the same tile as the station (in this case, it cannot pick anything up)

Other than the above, the pods function identically to the standard ones.

Happy transporting!!
Sprites done by @missatessatessy and yours truly!

#### Changelog

:cl:  
rscadd: A new type of transit tube pod! Designed to carry cargo.
rscadd: Flipped tube stations can now be built with the RPD
/:cl:


Partheo (approved):
Sprites are fine, I might make a PR to update them at another time (since the original is quite coderspritey)

JamieD1:
I spy a sleep proc.

jcatjmeow:
> I spy a sleep proc.

pre-existing. And we already talked about this. I can't replace it with a callback cause it needs to return something

code/game/objects/structures/transit_tubes/station.dm
```diff
@@ -128,30 +132,49 @@
...
+				if(pod.contents.len)
+					if(!isopenturf(input)) //Check to make sure the loaded crate doesn't load into a wall
+						input = loc
+					pod.empty_pod(input)
+					playsound(src, 'sound/mecha/mechturn.ogg', 25 ,1)
+				else
+					for (var/obj/structure/S in input)
+						if (istype(S, /obj/structure/closet) && !S.anchored)
+							S.forceMove(pod)
+							playsound(src, 'sound/mecha/mechturn.ogg', 25 ,1)
```
alexkar598, line 153:
```suggestion
					for (var/obj/structure/closet/S in input)
						if (S.anchored)
							continue
						S.forceMove(pod)
						playsound(src, 'sound/mecha/mechturn.ogg', 25 ,1)
```

alexkar598 (dismissed):

code/game/objects/structures/transit_tubes/station.dm
jcatjmeow: I tried this. It refused to load crates. Doesn't work

alexkar598: Theres no reason for that to not work...

jcatjmeow: Ok smartass but it didn't. Searching for(var/obj/struture/closet/S) ONLY gives closets. Not any children.

fluffe9911 (approved):
ok

jcatjmeow:
NOTE: If anyone has a way to remove the sleep procs without using a callback (the proc needs to return something) let me know or pr it

PR #9167, opened 2020-07-09
suicide alert now tells you you won't be revivable

SomeguyManperson:
this is apparently an issue also the ghost verb tells you you're stuck when you use it but suicide doesnt for some reason

:cl:  
tweak: suicide tells you  what suicide does
/:cl:


alao3alao3:
Players ought to know that you can't be revived through cloning or defibbing. Some guy didn't know that, and more won't.

JamieD1 (approved):

PR #9176, opened 2020-07-10
NT has finally made their lineup of passenger transit spacecraft available for purchase by the station

Nickvr628:
NT has unveiled the latest in passenger transit, and has made a special variant for employee transport between regional commands and station facilities. Presenting the **NT Transtar 800 Centcom Trim**!

image

Featuring a luxurious first class and a spacious business section, this economic transit shuttle is more than enough for the highest of employee populations. It even features in flight beverage and snack service and complimentary newscasters for entertainment! 

NOTE: Airlock boarding configuration is set up primarily for Box but functions on all stations.
 
:cl:  
rscadd: A new purchasable escape shuttle modeled after an airliner featuring enough seats for highpop. 
/:cl:


Xoxeyos:
Here's an idea, add curtains between the classes, just like airplanes.

OppressedCactus:
I like it. Also it has curtains!

Xoxeyos:
> 
> 
> I like it. Also it has curtains!

Oh I thought those were doors, nevermind then!

jcatjmeow:
Fun concept. I like the first class section especially. (Not the sec section but you know)

That being said, big parts are ugly as sin. First of all the gen pop section is so ugly and boring. It's just chairs and chairs and chairs. The shape of the shuttle is nothing like an airplane, and while yes I know it can't (and shouldn't) have big wings it should at least not be shaped like a lopsided poop. It just too thin to have any shape or detail.

In summary
- Sec section is not secure enough and looks out of place in the first class area
- Gen pop section is really boring
- Shape is squished and lopsided
- Only two airlocks it looks like??
- Floor tiles are lame. Where's the carpet?
- what are those lattices up front for?

Nickvr628:
The sec section is intentionally a modified first class cabin because it is an aftermarket kit for a civilian spaceliner

The business class is supposed to be cramped with seats with no leg room because it is a civilian spaceliner, that is the semi-joke/point.

I cannot go too crazy with the shape without making it not fit in all escape bays and still resembling a modern airliner. The offcenter cockpit is designed to make it somewhat unique and different from the "flying pencil" shape some shuttles are, which would look boring here.

Not every shuttle will be compatible with every station's escape dock, this configuration fits 90% of them with both airlocks connected.

I will think about the 1970s airline carpet, it isn't a terrible idea.

Lattices are greebling and are supposed to be sensor antennae things.

Partheo:
Kinda ugly but I absolutely love the design idea. The asymmetric cockpit is a nice touch.

jcatjmeow:
>The sec section is intentionally a modified first class cabin because it is an aftermarket kit for a civilian spaceliner

Fun concept. It's still not secure enough tho. Remember this is supposed to be usable by the station.

>The business class is supposed to be cramped with seats with no leg room because it is a civilian spaceliner, that is the semi-joke/point.

I like the joke/point, I just wonder if there's a way to convey it without being so ugly

>I cannot go too crazy with the shape without making it not fit in all escape bays and still resembling a modern airliner. The offcenter cockpit is designed to make it somewhat unique and different from the "flying pencil" shape some shuttles are, which would look boring here.

I like the off center cockpit but it makes it not look like an airliner. And the shape is just bleh. But I dunno how to fix it while making it still fit.

>Not every shuttle will be compatible with every station's escape dock, this configuration fits 90% of them with both airlocks connected.

Cmon dude, this is just laziness. Even real airplanes have a minimum of two separate door locations. Adding more airlocks to line up with say box would not hurt. Plus it really sucks when you go through the airlock and it ends in a wall, it's just a bad look all around.

>Lattices are greebling and are supposed to be sensor antennae things.
Now that you pointed it out, they look nice! I just wish there was a way to make it more clear.

Nickvr628:
image
I completed Wejengin2's revisions and took some other feedback into account.

The business class now has several emergency exit rows that come bolted by default but can be unbolted from the inside with the button, good for exiting onto centcom when the mass shooting starts. They can also be unbolted by silicons, door remotes, etc etc.

PR #9182, opened 2020-07-10
Juicer update

Partheo:
New sprites for the All-In-One Grinder.
The grinder in the kitchen now starts with a bowl instead instead of a beaker. No more beakers in the kitchen.

image

Other grinders are unaffected- they properly display a beaker if they contain one.

image
While grinding, a little thingie covers the container.

image

#### Changelog

:cl: Partheo
tweak: the grinder in the kitchen now contains a bowl instead of a beaker
imageadd: new sprites for the All-In-One Grinder
/:cl:


patpol4 (approved):
map fine

PR #9183, opened 2020-07-10
Chairs and beds can no longer be deconstructed instantly

Partheo:
Chairs take two seconds to deconstruct, and beds take four. For comparison, deconstructing a table takes four seconds. This is up from zero seconds, which allowed people to just demolish large amounts of chairs instantly if they had a wrench.

:cl: Partheo
tweak: Chairs and beds now take a few seconds to deconstruct
/:cl:


JamieD1 (approved):
there goes my metal farming strat for borgs

PR #9197, opened 2020-07-11
Final solution to the soapstone question

Amelia0010:
Press the button in preferences to show/hide soapstones, and the magically appear/disappear. It really is that easy. 

They'll still show up in the right-click/alt-click menu, but no hitbox and can't be seen with the eye

:cl:
tweak: There's now a preference to see/not-see soapstones
/:cl:

### DIFF
--- a/code/game/alternate_appearance.dm
+++ b/code/game/alternate_appearance.dm
@@ -36,6 +36,8 @@ GLOBAL_LIST_EMPTY(active_alternate_appearances)
 /datum/atom_hud/alternate_appearance/proc/onNewMob(mob/M)
 	if(mobShouldSee(M))
 		add_hud_to(M)
+	else
+		remove_hud_from(M)
 
 /datum/atom_hud/alternate_appearance/proc/mobShouldSee(mob/M)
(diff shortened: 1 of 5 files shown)


code/game/alternate_appearance.dm
```diff
@@ -36,6 +36,8 @@ GLOBAL_LIST_EMPTY(active_alternate_appearances)
 /datum/atom_hud/alternate_appearance/proc/onNewMob(mob/M)
 	if(mobShouldSee(M))
 		add_hud_to(M)
+	else
+		remove_hud_from(M)
```
Amelia0010, line 40: This *shouldn't* mess with anything else :)

alexkar598:
This goes down a weird path of hiding things you dont like. I'd rather we just purge on the amount of soapstone already on station

MegaEmpirical:
Like the options to already disable instruments and other things you don't want to hear or see?

Takahiru:
alexkar don't be bad

Amelia0010:
\>:(

PR #9199, opened 2020-07-12
Better lightswitches

### LINKED ISSUE #9169: Light switches don't glow in the dark
t12e7E519E

These should have an overlay that ignores lighting when the lights are out.

Partheo:
They glow in the dark now. I thought it was a bug but it turns out we just didn't have this feature. Also freshened up the sprite.

4XadGv1Yml

Closes #9169

:cl:  
imageadd: New light switches that glow in the dark
/:cl:


Partheo:
god that multitool looks so out of place on the table, remind me to change that

Hopekz:
What I never noticed they didn't glow in the dark I just assumed they did for some reason

code/game/machinery/lightswitch.dm
```diff
@@ -2,9 +2,13 @@
 /obj/machinery/light_switch
 	name = "light switch"
 	icon = 'icons/obj/power.dmi'
-	icon_state = "light1"
+	icon_state = "light-p"
 	desc = "Make dark."
 	power_channel = LIGHT
+
+	light_power = 0
+	light_range = 7
```
alexkar598, line 10: ?

alexkar598 (dismissed):

code/game/machinery/lightswitch.dm
Partheo: This allows the lightswitch object to be visible even in darkness. The light_range var makes its "light" and therefore its sprite visible at a distance of 7, but since light_power is 0 it doesn't actually emit any light. Without them, the lightswitch would glow in the dark but only if you were directly adjacent to it.

JamieD1 (approved):

PR #9203, opened 2020-07-13
Clerk gets a higher traffic secondary shop

Hopekz:
**Problem with Clerk:**
Their store isn't in a high traffic area. The Clerk is very much "out of the way" thus making the job itself boring.
 This gives them a second mini store in prime real estate location.
This mini store isn't as equipped as their main location and piggybacks power off of the hallway.

image

image

#### Changelog

:cl:  Hopek
rscadd: The Clerk now has a second high traffic location for his store.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -1033,6 +1033,11 @@
 /obj/machinery/door/firedoor/border_only,
 /turf/open/floor/plasteel,
 /area/security/prison)
+"ach" = (
+/obj/structure/table,
+/obj/item/flashlight/lamp/green,
+/turf/open/floor/plasteel,
+/area/security/prison)
 "aci" = (
 /obj/machinery/vending/autodrobe,
 /turf/open/floor/plasteel,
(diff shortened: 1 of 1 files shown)


ToGWtF:
make it a random spawn

Hopekz:
> make it a random spawn

Maybe.
The whole point of this is to make Clerk not boring by giving him the ability to operate closer to the population of the crew.
Having it a random spawn will just kind of make it to where whether or not the Clerk has something to do being RNG dependant.

It would be a 50% "sorry you're stuck in this corner with nothing to do" like before.

MrHorizons:
Maybe you could move the door to the same x position as the vent? That way, you could use the "panhandle" as a small storage space rather than a mere long hallway.

I think an office chair up front wouldn't be remiss, either.

Hopekz:
> Maybe you could move the door to the same x position as the vent? That way, you could use the "panhandle" as a small storage space rather than a mere long hallway.
> 
> I think an office chair up front wouldn't be remiss, either.

Good idea.
Implemented

image

Hopekz:
Making changes then re-releasing since it'll be different

PR #9204, opened 2020-07-13
Clerk gets a secondary mini location for his shop (Changed with feedback edition)

Hopekz:
**Problem with Clerk:**
The Clerk is very much "out of the way" thus making the job itself boring since he doesn't get any traffic.
This gives them a second mini store in a higher population area.

image

Changes:
- Space security feature has been re-implemented (thanks yacabo)
- Camera added and the door moved to make bridge secure like before (thanks MrHorizon)

#### Changelog

:cl:  Hopek
rscadd: The Clerk now has a second high traffic location for his store.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -21705,36 +21705,6 @@
 /obj/effect/turf_decal/delivery,
 /turf/open/floor/plasteel,
 /area/bridge)
-"aVu" = (
-/obj/structure/sign/warning/electricshock{
-	pixel_y = 32
-	},
-/obj/effect/turf_decal/tile/blue{
-	dir = 1
-	},
-/obj/effect/turf_decal/tile/blue{
-	dir = 4
-	},
-/obj/effect/turf_decal/tile/blue{
-	dir = 8
-	},
-/obj/machinery/door/firedoor/border_only{
-	dir = 4
-	},
-/turf/open/floor/plasteel,
-/area/hallway/primary/central)
-"aVv" = (
-/obj/machinery/camera{
-	c_tag = "Bridge East Entrance"
-	},
-/obj/effect/turf_decal/tile/blue{
-	dir = 1
-	},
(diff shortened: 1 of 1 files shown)


Hopekz:
Update: added a light source (red) and moved the vent (blue) to make things less cramped
image

Partheo:
I agree with Hopek that a change is necessary, but I personally don't like this location at all.

Nickvr628:
It is a bit of an awkward position but I like the idea.

patpol4:
Putting this on controversial for now, ill take a shot at moving it somewhere else.

Hopekz:
> Putting this on controversial for now, ill take a shot at moving it somewhere else.

Honestly, unless you smash it into the bar or remove some of the bathrooms (or some maint up top) not really sure where you'll smash it in to.

To solve this issue it needs to be somewhere in the high population path (hallway above bridge to science hallway)

Even switching it with the janitor's closet just introduces the same issue since it is now "out of the way" again.

Nickvr628:
image

Swapping clerk and command EVA could work and would be a much less awkward space to occupy. It is a great high visibility spot and would be pretty easy to swap the rooms.

Nickvr628:
Since there are now 3 clerk office PRs and will probably be a quickfire vote to see which is preferred, it will be interesting to see which is the winner.

Hopekz:
I'd like to point out that 3 out of the 4 "dislikes" are people that have their own Clerk PR up (Ready, Nick, Nick's alt)

Partheo:
> 
> 
> I'd like to point out that 3 out of the 4 "dislikes" are people that have their own Clerk PR up (Ready, Nick, Nick's alt)

Mine was a meme of course, but ignoring that, isn't this a good thing? Does it not make their criticism more valid that they say "I disagree with this PR, but here's an alternative idea that might work." ?

Or would you rather just a thumbs down with no explanation? "I don't like it."

Ktlwjec1:
I won't end up merging this one so I'm closing it.

PR #9214, opened 2020-07-14
Moves clerk shop to new location with better toys to sell

Partheo:
image

This PR switches the warden's office and the clerk's store. The clerk now has access to the armory, and can give much more interesting toys out. I think this will result in an increase in traffic because the gift shop items are so cool now. Even though it's a bit out of the way, many crewmembers will make their way to the new store location to get cool toys. Also the security officers ensure a steady stream of customers that come in and can browse your wares. Problem solved.

Closes #9204.

The warden now works from the clerk's old store. This is a better location for him because he is closer to the action, and since they are security wardens must be right in the action at all times. The warden has less readily available gear now, but his robust dog is present, so properly training and utilizing the dog's abilities will provide players with a sense of pride and accomplishment.

image

#### Changelog

:cl: Partheo
rscadd: The clerk shop now has more interesting items to sell!
/:cl:


TheRealScarHomie:
approved

Nickvr628:
Unironically turning the clerk into a gun salesman and making the armory his shop would be an amazing move for yogs.

patpol4:
nice meme

PR #9219, opened 2020-07-14
Ever so slightly buffs airlock charges

Amelia0010:
Basically just standardizes the two uses of airlock charges into one proc, ever so slightly buffing its effects when accidentally triggered while trying to remove it.

:cl:
tweak: Airlock charges have been ever-so-slightly buffed, setting people on fire even when it's triggered through a failed removal
/:cl:


Amelia0010:
tbh the buffing thing is a misleading title, since that's not the intent, but I figured balance changes are more important than code cleanup so you know

It's not like the airlock charge can't use it :)

alexkar598 (approved):

PR #9220, opened 2020-07-14
Significantly buffs the greytide virus

Amelia0010:
It can now affect cameras, which it has a 75% chance of not doing (due to the sheer mass of cameras)
If it does affect a camera, it'll do one of the following:

- emp it
- emp-proof it (with a 50% chance it's done in secret, like the malf upgrade)
- motion-activate it (which really can be a massive pain if it's in a location like the medbay)

It can also affect key-card auths, which will do the following:
When it reaches 2 auths, it'll open maint if it's not already open, otherwise it'll red alert
When it reaches 4 auths, it'll red alert

None of this is impossible to deal with, even the camera upgrades just require an engineer/engiborg to detach and retach the camera. The alert/maintenance can be dealt with from any command console by an AI/headofstaff

:cl:
rscadd: Greytide Virus now affects both cameras and keycard auths
/:cl: 

Any additional suggestions to either more stuff to affect or changes to the already affected items are welcome

### DIFF
--- a/code/modules/events/prison_break.dm
+++ b/code/modules/events/prison_break.dm
@@ -63,24 +63,20 @@
 				// 75% chance we just leave it be, to not kill the AI/CE's joy *too much*
 				if(prob(75))
 					continue
-				var/option = pick(list(1, 2, 3, 4))
+				var/option = rand(1, 3)
 				switch(option)
-					// As if a multitool was used on it
-					if(1)
-						temp.setViewRange((view_range == initial(view_range)) ? short_range : initial(view_range))
-
 					// As if EMPed
-					if(2)
+					if(1)
 						temp.emp_act(2)
 					
 					// EMP-proof it, but with a 50/50 chance we don't tell them
-					if(3)
+					if(2)
 						var/secret = prob(50)
 						temp.upgradeEmpProof(secret)
 
 					// Give it a proximity alarm, can be either a boon or a curse depending on which camera gets it.
 					// Either way, the CE/Engineers can always just take the camera down and replace it.
-					if(4)
+					if(3)
 						temp.upgradeMotion()
 			
 			// If we hit 2 keycard auths we'll open maint unless it's already open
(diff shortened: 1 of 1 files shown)


Jarod1200:
maybe some sort of interaction with apcs? like unlocking some randomly or something

alexkar598:
I'd rather not have the multitool thing. Everything else is easably detectable or it fixes itself. The multitool thing you wont notice it at all until you need it

code/modules/events/prison_break.dm
```diff
@@ -57,3 +58,46 @@
 				var/obj/machinery/door_timer/temp = O
 				temp.timer_end(forced = TRUE)
 
+			else if(istype(O, /obj/machinery/camera))
+				var/obj/machinery/camera/temp = O
+				// 75% chance we just leave it be, to not kill the AI/CE's joy *too much*
+				if(prob(75))
+					continue
+				var/option = pick(list(1, 2, 3, 4))
```
alexkar598: rand(1,4)

alexkar598 (changes requested):

code/modules/events/prison_break.dm
```diff
@@ -57,3 +58,46 @@
...
+				switch(option)
+					// As if a multitool was used on it
+					if(1)
+						temp.setViewRange((view_range == initial(view_range)) ? short_range : initial(view_range))
```
alexkar598: Copy paste bad. Those are variables on the camera

alexkar598 (dismissed):

Moggram:
>motion-activate it (which really can be a massive pain if it's in a location like the medbay)


This is a cool idea and I hate you for it.

Ratvar:
Maybe it could emag-hack apcs instead of just unlocking them
Could hack open borg and bot controls
Could set on-wall radio transcribers to secure channels, like medical, command, AI, etc

JamieD1 (approved):

PR #9225, opened 2020-07-15
fix hulk not stopping speech impediment when removed

SomeguyManperson:
this signal is never unregistered
:cl:  
bugfix: you no longer yell when dehulked
/:cl:


Marmio64:
finally

alexkar598 (approved):

PR #9164, opened 2020-07-08
Security Records TGUI

Bibby0110:
### Intent of your Pull Request

Refactors the Security Records console to TGUI and cleans up the code

image
image
image
image
image

Thanks for doing the scanlines Nich!

#### Changelog

:cl:  
rscadd: The Security Records Console has a new UI!
/:cl:

PR #9236, opened 2020-07-17
Changes the plasma internals tank in the plasmaman emergency box to be belt-sized.

MrHorizons:
Normal humanoids don't get a full-sized oxygen tank in their emergency boxes, so why should plasmamen?

I was tempted to make more changes to the plasmaman emergency box, but decided against it for now.

#### Changelog

:cl:  
bugfix: Changed the plasma internals box in a plasmaman's emergency box to be belt-sized.
/:cl:


yacabo:
Plasmamen are limited because they need it to breath at all times, other humanoids can get it from a lathe

MrHorizons:
@yacabo Sure, they need it to breathe at all times, but they might not want to take off their backpack or find a suit that supports holding it to do so. It also appears that the full-sized plasma internals tank that was formerly in the emergency box somehow held less than the filled belt-sized tank.

yacabo:
Oh I'm missinterpiriting what this is for, I thought this was reducing the capacity of the plasmaman belt tank

MrHorizons:
Also, Turdis failed because of something on a completely unrelated file, apparently. Which appears to be related to #9164, which therefore makes it not my problem, especially for this PR. This just consists of, like, 10 characters.

MrHorizons:
There we go, it should work now.

JamieD1 (approved):

PR #9239, opened 2020-07-17
sec records tgui

Amelia0010:
https://github.com/yogstation13/Yogstation/pull/9164 but 100 characters instead of 104


fluffe9911 (approved):
If this breaks things im yelling at swiss hard

PR #9247, opened 2020-07-18
Department goals

Amelia0010:
There's 1 objective for every department because I have no creative skills, but adding more is "easy"

Each department is assigned up to 5 department goals, which they can complete. These goals usually pay out money (but don't have to, they can be overridden in on_complete() to provide other rewards.

Goals can be continuous, meaning they check every X seconds, and pay out their reward every time it's true.
Goals can be end-round, meaning they're completable only at the end-of-the-round (none have been added yet, persistency to come)

All goals are sent to the relevant heads' computer on round-start, as well as the communications consoles, and all members of the department can see their department goals on their PDAs.

You're more than welcome to come up with other department goals or just PR them later or whatever, I just added the ones that are there now as examples.

:cl:
rscadd: There's now a framework for department goals. Each department is given up to 5 goals a round to complete, with bonuses when they're completed.
/:cl:


Moja10:
Awesome!

JamieD1:
Looks good, can you explain this:

image

Amelia0010:
if you see
```
/datum/department_goal/proc/message_players(message)
	var/string = "Your department's goals have been updated, please have another look at them."
	switch(message)
		if("ded")
			string = "Your department no longer has the goal: " + name
		if("new")
			string = "Your department now has the goal: " + name
		if("complet")
			string = "Your department completed the goal: " + name
```
You'll see that it's just me using those as shorthands for the messages that will be sent, since I didn't really want put the whole message in there, and defines are yuck

fluffe9911 (approved):
hip and cool

PR #9248, opened 2020-07-18
Rebalances Energy Harvester - now ready for balance testing!

ghost:
As it says on the title. Please give balance feedback so everyone is satisfied. Currently unfinished and PRing it here so people can comment since the station economy is in everyone's best interests.

The energy harvester, contrary to popular belief, actually works and makes money, but if you aren't using an unorthodox setup (read: rad duplicants or multitesla) and saturating it in full, it generates less than 0.1 credits per tick. This change makes it so that power exports are credited and deposited to the engineering budget whenever the SSEconomy subsystem does its once every 5 minute payout instead of per process() tick while also tweaking the payout formula so that regular setups produce money without needing to go all the way up to 10MW just to get any usable amount.

Currently it uses a piecewise function for 3 different power ranges to balance against different engine builds and configurations, with an easy linear curve at first, then a logarithmic curve, and an even worse logarithmic curve after that. The first threshold if which is a stable 250KW/s rewarding you with 2000 credits, the second being 10MW/s rewarding you with 3000 credits and finally a nigh-unreachable threshold of 1PW/s, rewarding you with up to 5000 credits. These stack, allowing engineers to earn 10000 credits per budget cycle, but given that extreme power generation values take time to prepare, it is balanced throughout the shift. 

For exact details on the scaling, please refer to the calculateMoney() proc. To tweak it, change the defines at the top of the .dm file.

### To do:
[x] Fix load
[x] Get tgui running
[ ] Get balance feedback and implement (pending testmerge)

#### Changelog

:cl:  
tweak: changed the scaling of the energy harvester
/:cl:


Readystorm546:
YOu could copy the point system for toxins with its hardcap. The same equation should work. The cap might have to be smaller if it does money per tick. If its per minute it could be a bit higher

Georgeur:
How will the money be accessible now that the budget cards are removed, is the money for engineering or cargo/civil budget. Just wondering on your view, as its nice this feature wont be broken anymore.

ghost:
The overall money issue is outside of my purview, but currently the engineering budget doles out a percentage of itself to all engineers on station.

fluffe9911 (approved):
Infinite power or go home

PR #9252, opened 2020-07-19
makes tentacle less uselessby giving tentacle grab a very small immobilize so it isnt countered by walking away 2 steps

SomeguyManperson:
this costs 2 gamer points and is worse than like everything else you can get
except for like
i dunno hallucinate sting
:cl:
rscadd: changeling tentacle grab effect has a short immobilize so the grab actually goes through
/:cl:


Marmio64:
YES, GOOD, UNGA GOOD

not even joking, tentacle is 2 points and is such a gimmicky item that you will likely not ever find a good enough use for in its current state.

JamieD1 (approved):

PR #9260, opened 2020-07-20
Centers a lot of food objects

ynot01:
fuck you


Partheo (changes requested):
90ZsU3Qa8W

Also, you were inconsistent with some of the pizzas:
image
image
Please center them all the same.

Partheo:
what the fuck turdis

ynot01:
ill fix the migrawr but i dont have this fancy cross program you use im just eyeing these

Georgeur:
I personally prefer most of them lower, as especially on the chefs counter, it makes it seem less top down, 2d perspective.

ynot01:
> I personally prefer most of them lower, as especially on the chefs counter, it makes it seem less top down, 2d perspective.

maybe but for practicality sake placing something and then having it appear below your cursor is annoying

Partheo (approved):

PR #9281, opened 2020-07-22
Removes the gamer spawners from the gamer ruin

SomeguyManperson:

The gamer ruin is a ruin for being chill and relaxing in. People who are ghosts do not want to chill or relax, and tend to either use the teleporter setup to steal shit, suicide, or act antagonistically for no reason

this is furthered by the "protect the snack vault at all costs" objective, which may lead to the idea gamers can kill or maim people who enter the gamer ruin on account of "they might steal the snack vault"

#### Changelog

:cl:  
rscdel: Gamer spawners are removed from the gamer ruin. You know what you did.
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/gameroom.dmm
+++ b/_maps/RandomRuins/SpaceRuins/gameroom.dmm
@@ -676,12 +676,6 @@
 /obj/structure/chair/stool/bar,
 /turf/open/floor/carpet,
 /area/ruin/space/has_grav/powered/gaming)
-"Ct" = (
-/obj/effect/mob_spawn/human/gamer{
-	dir = 1
-	},
-/turf/open/floor/carpet,
-/area/ruin/space/has_grav/powered/gaming)
 "CT" = (
 /obj/machinery/vending/autodrobe/all_access,
 /obj/machinery/door/firedoor/border_only{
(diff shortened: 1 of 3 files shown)


patpol4:
Sad it has to come to this
Maybe add a sentence to the flavor text to not behave like a fucking caveman

patpol4:
I cant request since it is my work but i have suggestions that might be better: 
Tighten the flavour text
Remove spawners from map but leave them in the code in case an admin wants to look over the ruin
Just remove the tele
Remove the thing from spawning in space altogether

SomeguyManperson:
ruin itself might be fun, people never use the spawners to actually do stuff with the ruin though

ynot01:
yea this encourages some friends to hang out there instead of some nameless mulligans who exist for 10 minutes then start trying to greytide the station with the tele

patpol4:
Time has shown i was wrong, now I'm all for

yacabo:
Why not just remove any way for them to leave so that the only reason to become one is to play?

patpol4:
> Why not just remove any way for them to leave so that the only reason to become one is to play?

That would just lead to randoms joining and then suiciding since they dont have anything to do. This place is better off as a spot to chill without ghosts players in it

patpol4 (approved):
Map should be fine

JamieD1 (approved):

PR #9282, opened 2020-07-22
Port block rework from bee rising revengeance

SomeguyManperson:
see BeeStation/BeeStation-Hornet#1642

CURRENT POSSIBLE CHANGES:
decrease effectiveness of blocking so as to not make it completely counter other melee weapons

:cl:  	zeskorion
experimental: totally changes up block chance. blocking now relies on items being held in the active hand or not, if you are walking or not, and the direction you are facing. Shields now use durability and all other items deal stamina damage to the arm holding them when blocking. Read the pr for more in-depth info.
/:cl:


SomeguyManperson:
probablay not missing anything

MegaEmpirical:
Hmm, test merge and let's see it in action.

SCortesBar:
This one sounds reasonable though if they end up making shields useless the most affected ones will be security so I'm wary.

deathslayer2481:
This sounds incredibly imbalanced and to be polite “Fucking stupid”. Fun fact from history: Shields were used in the OFFHAND. They’re used in the OFFHAND because they’re supposed to BLOCK, not ATTACK. You can BASH with a shield, but its main use is as a BLOCKING tool. Forcing people to walk while they block... Doesn’t make sense. At all. Whatsoever. You’re very very capable of blocking a swing even while running. The direction thing? That’s OK, I’m totally for that. The stamina damage? That’s pretty bad, since that slows you and allows you to be outmaneuvered, making the shield useless. It also doesn’t make any sense, swinging a weapon takes far more energy than blocking a strike.

maxion12345:
ok so
directional shielding: understandable
stamina damage: will be a fucking pain in the ass
active hand: actually might be good, requiring you to swap hands to actively block so theres not two people with swords and shields smacking eachother until someones RNG gets good and they get a click in (see: cult with mirror shield vs sec with riot shield), although passive hit blocking should still be a thing due to you, well, still holding the bloody thing. maybe against thrown shit idk
walking: not sure why this would matter ngl
also shields already have durability, you can shatter one with enough hits

all in all
going to go with neutral

Marmio64:
i don't like most of this mostly like what deathslayer said, all the shit sounds stupid besides directional blocking, which completely makes sense, stamina damage does make sense but would makes shields and other blocking weapons far too worse.

deathslayer2481:
Idea for a compromise: make blunt weapons do stamina damage vs shields, but sharp ones just slide off and do nothing.

SomeguyManperson:
hhhhh
ok
only items with BLOCKING_ACTIVE require to be in the active hand (not shields, usually swords)
walking is NOT required for any item that is MADE to block stuff i.e. shields and swords, what blocking WILL do is increase the radius a block will effect. Most swords will still block at the very least directly in front of you, and shields will generally block the 3 tiles in front of you. walking IS required to block with stuff like a toolbox, since it's blocking radius by default is 0.
Shields will ONLY take damage themselves, other weapons being used to block causes stamina damage to the limb they are held in, so you can't hold a block theoretically forever. The total damage taken from a block (currently) is capped at 50, or 37.5 for total damage (limbs count as 75% for total damage) meaning a block being broken on its own will not (currently) put you into damage slowdown.

patpol4:
low iq bot

patpol4 (approved):
Map is fine it's literally a one word map change stupid bot

Georgeur:
Messed around with it. Its not bad, indirectly buffs riot gear which is a good thing, the stamina damage is pretty annoying. Looking forward to a test merge for more experience with it.

maxion12345:
why close

PR #9290, opened 2020-07-23
adds a ticket machine to boxstation

ReddicusDragon:
see title
:cl:  
rscadd: adds a ticket machine to box  
rscdel: removes a wall by hop office  
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -55908,6 +55908,14 @@
 	},
 /turf/open/floor/plasteel,
 /area/hallway/primary/central)
+"qaq" = (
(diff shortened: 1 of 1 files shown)


patpol4 (changes requested):
The way it is now it looks super weird, maybe try rotating it to be facing left and add thin windows?
something like this
vivaldi_WFfsI0luvI

The problem is that the firelocks just cut off, we don't want that

ReddicusDragon:
@wejengin2 basically the ticket machine doesn't have directional sprites also will that suffice in window placement

jcatjmeow:
You can put windows on those places and still click on the machine from the side

It needs to face forward so you can see the numbers

jcatjmeow:
Misclick

PR #9332, opened 2020-07-28
[Draft] New race - The Savants (Needs some work)

jcatjmeow:
What still needs to be done

- [ ] Sprites need some shading improvements
- [x] Inhands need to be able to render correctly
- [x] Snowflakery with cloning/genes needs to be added (So that savants can't get cloned as suited ones)
- [ ] Naming conventions
- [ ] the smallhands trait needs some work
- [ ] Needs some anti-augmentation surgery trickery
- [x] Suit healing testing

Adds a new playable roundstart race, the [Savant](https://cdn.discordapp.com/attachments/440551954788253697/736796979237093417/unknown.png)!

The race starts out in the suited form. In this form, they can do most things a normal crewmember can (besides wear a jumpsuit). Their arms and legs are mechanical, and thus heal with welding/cable coils, but their head and chest are still organic.

However, after taking 50 brute or burn damage, the suit will fall off, exposing the Savant's true form. This true form is very cute and cuddly, but also small. It has the new trait "Small hands" which limit what it can hold at once, and the trait "short" which makes it hard to reach things on tables or climb them.

Your suit can be repaired with 30 metal. You can also take off the suit manually, which will deal 25 burn and 25 brute to you but allow you to be small. No idea why you'd want to but whatever, it's an option. Putting on the suit will heal you to above the suit breakage threshold so you won't be locked out of it.

Savants speak common, but have a tendency to roll their r's due to way they make sounds (more like a bell than like a human's which is more like a guitar) 
#### Changelog

:cl:  
rscadd: Added a new race, the Savants
rscadd: Added the traits 'Short' and 'Small hands'
rscadd: Some autodoc stuff for species 
/:cl:


boodaliboo:
intresting

MegaEmpirical:
Interesting indeed.

Marmio64:
i'd rather have fishmen

jcatjmeow:
> i'd rather have fishmen

what

Smudged:
I'm loving the new triats you've introduced to your race, now I do have a few questions, 

What speed are you at if you do get yeeted out of your suit? 
Is there a slowdown threshold you're in your short form?
Like, when you're short, are you as slow as someone who did receive about 50 units of damage or are you a normal speed until you get hit again? 
Do you get slowed down prior to taking enough damage to launch you from your suit?

Smudged:
you may want to add a little bit of lore for the rsce too

jcatjmeow:
> I'm loving the new triats you've introduced to your race, now I do have a few questions,
> 
> What speed are you at if you do get yeeted out of your suit?
> Is there a slowdown threshold you're in your short form?
> Like, when you're short, are you as slow as someone who did receive about 50 units of damage or are you a normal speed until you get hit again?
> Do you get slowed down prior to taking enough damage to launch you from your suit?

You have a slowdown of 1 outside the suit (because you are small). For reference, golems have 2 (higher number is slower) and humans have 0.

You ALSO have a slowdown from being injured. The slowdowns do stack, and apply immediately. So you are SUPER slow when you first get slpoded. It's a big deal.

You will get slowed down a little bit prior to taking enough damage but not a lot.

Partheo:
I feel like the short trait should give some sort of advantage, to possibly give a reason to take off the suit. Not something as intense as ventcrawl for example, but maybe something like dodge chance or crawling under tables.

Partheo (changes requested):
Sprites aren't very good, I suggest you work with a mob spriter or go to SS13 spriter school for help.

Partheo:
image

maxion12345:
I'd say instead of having the suits tied to character health, have them take damage independent from the savant. Exception being chest and head, of course. This prevents some weird fuckery with getting free health from putting on the suit. Being an extension of the arms and legs, I suppose they would block a majority of limb damage, at the cost of suit integrity. I also suppose stamina weapons would also have a lessened effect as well, but not completely nullified. 

Also needs to be more visually different. Lizards have tails, humans have hair, preternis have metal bodies. Need something unique, because currently it looks like an ethereal head and limbs of a preternis.

ynot01:
whats the upside

jcatjmeow:
> whats the upside

There isn't one. 

Races don't need an upside. Sometimes it's just fun to do something different

jcatjmeow:
Not stale btw. I just have depression lol

On that point, I'm thinking of having the race be able to use metal to create tiny trinkets. Nothing super useful and of the actual useful things nothing past what an unhacked autolathe could do. So that would be the 'upside', thoughts?

MonsWTF:
> Not stale btw. I just have depression lol
> 
> On that point, I'm thinking of having the race be able to use metal to create tiny trinkets. Nothing super useful and of the actual useful things nothing past what an unhacked autolathe could do. So that would be the 'upside', thoughts?

Oof, that sucks for mood. Hope you're able to get out of that soon.

I'm confused on how to evaluate this race. Naming them Savant puts me in the head-space that they're more knowledgeable at something compared to the other races onboard. If the goal of the PR is to have them just be another flavor, like Lizards and Felinids, then there needs to be some kind of 'here's what you can expect from them' outside of just naming conventions. I see the tiny hands and the self-constructed suit, but it doesn't feel that impactful for some reason, so I'll make a proposal.

Make the Savants actual savants at two things: job they're in, and a department of their choice. Their naming system is that their last "name" is a combination of a prefix and the department that they've become a savant in. I.E. John Omniscience the Shaft Miner knows not just how to dig for ores, but also knows the ins and outs of Robotics, RnD, Toxins...etc. 

This is essentially a race incorporating the idea that regular jobs can't just "metalog" (I've never heard that term before it was used a few weeks ago) the devices and tools of other jobs, and know more than reasonably allowed (like an engineer knowing how to whip up a vaccine better than the virologist).

Their downsides are pretty much what you've listed; they're very frail outside of their suits, but perhaps also that their vast knowledge leaves them prone to mind-altering effects and brain damage. They take more brain damage, mind-altering drugs/effects have increased duration and increased efficacy, etc.

The only thing really missing after that would be the lore behind the Savants, how NanoTrasen came across them, when they were integrated into the crew, how they stand up to other races like the Plasmamen, etc. If my ideas for their brain pro's and con's is too radical, maybe that at least helps figure out where you'd want the race to go?

jcatjmeow:
My brain is really unwell right now.

I dont think I'll be coding much of anything in the immediate future.

PR #9337, opened 2020-07-29
You can now shoot pirates down with the BSA before they arrive at the station

Amelia0010:
PR is more comment than code l m a o

Allows you to shoot the pirate ship down between the announcement that they're coming, and them actually turning up. I don't really want the players to be able to do it *after* they arrived, since that's kinda dick-ish ngl.

Anywho, it spawns a very small meteor wave after the ship has been shot down. I feel like this is is a pretty good deal, and it gives players something to actually use the BSA for.

🆑 
rscadd: Central Command has issued new targeting systems with its BSA kits, allowing any stations that build them to shoot down any pirates that may attempt to board the station
/:cl:

(Also if anyone tries to make this apply to nukies I'll fucking glass em)

### DIFF
--- a/code/game/objects/items/devices/gps.dm
+++ b/code/game/objects/items/devices/gps.dm
@@ -207,3 +207,34 @@ GLOBAL_LIST_EMPTY(GPS_list)
 	tagged = null
 	STOP_PROCESSING(SSfastprocess, src)
 	. = ..()
+
+/**
+  * # Pirate GPS
+  *
+  *	Pirate GPS used for targeting by the [Blue Space Artillery] [/obj/machinery/computer/bsa_control]
+  *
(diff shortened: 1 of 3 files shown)


Amelia0010:
@alexkar598 This error is legitimately not my fault, and it shouldn't even be happening? I didn't even touch anything even remotely related to it....

I mean, look at the DMI file and the error:
image

image

image

Marmio64:
whats that? You are a new player who just got pirate? Well, better luck next time.

Amelia0010:
Local commenter doesn't read PR  You can only shoot down the pirates before they actually spawn.

fluffe9911 (approved):
hip

PR #9343, opened 2020-07-30
fix insulateds toolbox

SomeguyManperson:
:cl:  
bugfix: garunteed insulateds blue box now actually spawns insulateds
/:cl:


JamieD1:
Just fix the base definition

SomeguyManperson:
The box's type path and the type path of the box that is populated woth insulateds was different this makes them the same so it works right

fluffe9911 (approved):
ok

PR #9352, opened 2020-07-31
Adds movie-like rolling credits to the end of the round

swissloaf:
Credit:
BeeStation/BeeStation-Hornet#318
BeeStation/BeeStation-Hornet#320
BeeStation/BeeStation-Hornet#327
BeeStation/BeeStation-Hornet#407

[A Video of such](https://streamable.com/wbpn4)

#### Changelog

:cl:  
rscadd: Adds movie-like rolling credits to the end of the round
/:cl:


patpol4:
cute

Partheo:
Kind of neutral. I like the idea but I think they could look better, and possibly be on the side rather than in the middle.

Also, can this be disabled? This should absolutely be a preference. I know many will ignore them after a while, and a significant population will want to have them turned off.

swissloaf:
> Kind of neutral. I like the idea but I think they could look better, and possibly be on the side rather than in the middle.
> 
> Also, can this be disabled? This should absolutely be a preference. I know many will ignore them after a while, and a significant population will want to have them turned off.

It has a preference option. If you have widescreen which most do they appear on the side.

Partheo (approved):

maxion12345:
Does this remove the standard EORG sign that currently appears?

boodaliboo:
eh i truthfully didnt like the look of this when i played bee

swissloaf:
It does not remove the current one

PR #9353, opened 2020-07-31
[Test merge requested] Emergency skinsuits

Hopekz:
Ports skinsuits from Bee and NSV.
Skinsuits were created when Bee and NSV adopted Monstermos as the pressure kills you really quickly otherwise.

The skinsuit protects you from the pressure but the cold in space still kills you. 
It gives you extra time before the cold kills you vs pressure (about 20-30 seconds depending on how quickly you put your suit on after depressurization since pressure damage builds up quick).
This means you can't space explore in these because you will still die to cold but you will live longer to get to safety in the event of an emergency.

Doesn't have any real benefits that normal space suits provide. The helmet doesn't even hide your identity like a usual suit.
The suit makes you really, really slow (much slower than the usual slowdowns we're used to)

Slowdown demonstration: https://imgur.com/a/1wSjys6

sRn5PmOPsl

#### Changelog

:cl:  Hopek , ike709
rscadd: Added emergency skinsuits to assist in getting to safety during emergencies.
/:cl:


athena413:
oh boy more Bull Shit to keep in my emergency box instead of actually having stuff in the emergency box

yacabo:
Dunno if I like the idea of everyone having a space suit, maybe if they were in the emergency lockers instead?

Hopekz:
> Dunno if I like the idea of everyone having a space suit, maybe if they were in the emergency lockers instead?

Why introduce the initial scavenger hunt? It works pretty well on Bee and NSV for monstermos. 
I'd like to try this PR as is and have future PR's modify if necessary.

you only have a few seconds to unroll and put the 2 pieces of the skinsuit on before you go into crit and having to find a locker for each person in the area is quite unrealistic in my opinion (assuming the area has a locker).
A second thing that putting these exclusively in lockers will do is just have most people take suits for themselves meaning that there won't be suits in the actual lockers themselves who chose to wait for some reason. 

This suit isn't a "get out of jail free card" like the usual space suits we are used to. If you take 2-3 seconds to put on this suit you will take an amount of damage just from the instant exposure to pressure loss. After you put on the suit you start to die from the cold (but slower than a pressure death)

**EDIT:
Miscommunication. Updated description to provide more information about the skinsuits**

Partheo (approved):
Sprites fine. I'm personally against the feature though. This makes breaches so much less scary.

Hopekz:
I'd like to have this testmerged.
It's one of those ideas that may not sound good but in practice works very well (live tested in-game on NSV and bee with monstermos)

boodaliboo:
seems cool

ynot01:
likely a requirement for LRP madness that is bee, not sure if we need it

Hopekz:
> likely a requirement for LRP madness that is bee, not sure if we need it

I have to include bee because they're bundled together with NSV but NSV itself (where I originally got the PR from) is not lowrp.

The RP argument can be applied both ways as having no temporary emergency suits in space is LRP in a reverse argument. This is because you can't RP if you die immediately when a dragon or mech decides to go on a spacing spree (or really any antag with a green light)

maxion12345:
Have em stored in lockers similar to fire extinguishers, that automatically unlock if a depression event is recognized. This prevents cheesing by just pulling a fire alarm. Although this could be interesting RP, as it could be considered theft to activate a fire alarm to steal a suit for non-emergency purposes. This also solves the "spawn with box full of everything you could ever need ever" issue. And they of course would unlock during red alert+.

Featherfield:
The game really does need more emergency items like this because right now only option is to run for the nuclear option everytime there is even a small breach or something of that type

Partheo:
I also feel like this makes no thematic sense either. I know the emergency box has other things that go against this, but nonetheless;

There's really no reason Nanotrasen would distribute an emergency space suit to every crew member. They have no reason to believe breaches will happen. It's a civilian research station, why anticipate explosions, decompression, sabotage etc will occur. It's the worst case scenario.

Hopekz:
> I also feel like this makes no thematic sense either. I know the emergency box has other things that go against this, but nonetheless;
> 
> There's really no reason Nanotrasen would distribute an emergency space suit to every crew member. They have no reason to believe breaches will happen. It's a civilian research station, why anticipate explosions, decompression, sabotage etc will occur. It's the worst case scenario.

These aren't space suits they're janky rolled up skinsuits not meant for prolonged space exposure (because you'll die).
Also; they're given out the same reason everyone gets an oxygen tank, breathing mask and a literal shot of epipen.

Partheo:
> These aren't space suits they're janky rolled up skinsuits not meant for prolonged space exposure (because you'll die).

You know what I meant.

> Also; they're given out the same reason everyone gets an oxygen tank, breathing mask and a literal shot of epipen.

Yes, I literally addressed this. That being said, there's a big difference between giving someone a single-use injector and giving them an entire pressure suit.

Hopekz:
> Yes, I literally addressed this. That being said, there's a big difference between giving someone a single-use injector and giving them an entire pressure suit.

I didn't really want to get into the difficulty of equipping crew argument until you mentioned that EpiPens are easier to administer than skinsuits as skinsuits are inexpensive compared to EpiPens in reality.

**It is much easier to equip people with a rolled-up emergency skinsuit then it is to give them a shot of EpiPen.**
Even if it was the other way around contrary to reality to support your point: It does not hold water in my opinion as I don't think we should design our space manz game around everyday logic as it is a sci-fi setting. Gameplay should come first.

Also; let us not forget that both of these emergency equipment is for a different situation.

If we use your logic EpiPens should not be a standard-issue as that would mean NanoTrasen expects the crew to be hurt on their "research" station. 
Why even include oxygen tanks in the first place? (the station comes with air). Is it maybe because they expect the air to be absent/sucked out because the setting is in actual space where these things can easily happen?

ttIZUu4f3y
ZmuEt6StYE
DNmnYIiBOU

Partheo:
> I didn't really want to get into the difficulty of equipping crew argument until you mentioned that EpiPens are easier to administer than skinsuits as skinsuits are inexpensive compared to EpiPens in reality.

Really? Price? Not at all what I meant, but let me humor you. "Epipens" cost so much because the pharmaceutical industry in the US sucks ass. In most places in the world, an epinephrine injector is not expensive. Furthermore, this is 500 years in the future, where there are literally machines on the station that can produce chemicals, epinephrine included, in seconds. They would be dirt cheap.

Now, what I actually meant: an injector fits in your pocket. It's the size of a pen. It makes sense that it would not be too difficult to make this a standard issue item for all crewmembers. The suit however, is a large piece of specialized equipment. Sure it may be cheap, but cost is not what I am arguing. I am saying that it's less feasible and reasonable to make every single employee in your facility to carry an entire change of clothes with them at all times. Telling them to hold a pen-sized object in their bag? Not hard at all.


> Even if it was the other way around contrary to reality to support your point: It does not hold water in my opinion as I don't think we should design our space manz game around everyday logic as it is a sci-fi setting. Gameplay should come first.

Yeah, I know. And I think this PR is bad gameplay, for reasons that have been discussed already.

> If we use your logic EpiPens should not be a standard-issue as that would mean NanoTrasen expects the crew to be hurt on their "research" station.

Yes.

> Why even include oxygen tanks in the first place? (the station comes with air). Is it maybe because they expect the air to be absent/sucked out because the setting is in actual space where these things can easily happen?

These objects are much smaller. But yes, I do think it's unreasonable to ask all your crewmembers to carry oxygen tanks with them. There are emergency closets in case something like this happens.

Hopekz:
> Yeah, I know. And I think this PR is bad gameplay, for reasons that have been discussed already.

> But yes, I do think it's unreasonable to ask all your crewmembers to carry oxygen tanks with them. There are emergency closets in case something like this happens.

Mane, you don't get enough time to get to the lockers and put on all the gear on before you die (assuming there is a locker in the space you are trapped in).
It takes seconds to go into crit from the pressure you can't possibly argue that it is better for gameplay to die than to have an item that extends your life to maybe get to safety so you can continue playing the game.

Space is still very scary even with a skinsuit as if you take 3-4 seconds to put this gear on you still take enough damage from the pressure to die in around 20 seconds from the cold. Less if you are a lizard or plant (cold is burn damage and pod people don't heal under a certain temperature).

Partheo:
> It takes seconds to go into crit from the pressure you can't possibly argue that it is better for gameplay to die than to have an item that extends your life to maybe get to safety so you can continue playing the game.

Yeah, I can. I think depressurization should kill you, go figure.

Why not make the epinephrine injector auto-inject from your bag and revive you on death too, while you're at it. So you can continue to play the game.

maxion12345:
no worry guys i got my skintight latex suit in my back pocket I'll be safe from the pressure of space for sure

Partheo:
also, you really gonna thumbs-up your own pr...

Hopekz:
> also, you really gonna thumbs-up your own pr...

Hater

alexkar598:
Considering you can already survive in a vaccum for a bit without dying, this pr is denied. It doesn't help that vaccum isnt cold anymore looking at you @monster860 

no o2
37s crit
1m 33s death

o2
1m 05s crit
2m 10s death

Hopekz:
> this pr is denied

Sucks that we didn't get a test merge since it might have been a fun idea.
lame

PR #9366, opened 2020-08-01
Security officers get a bola instead of a flash

Takahiru:

They already get one in their sec belt, and energy bolas are more useful, imo.

Lewdelf wanted this

#### Changelog

:cl:  
tweak: security officers start with an energy bola instead of a flash in their pocket  
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/security_officer.dm
+++ b/code/modules/jobs/job_types/security_officer.dm
@@ -144,7 +144,7 @@ GLOBAL_LIST_INIT(available_depts_sec, list(SEC_DEPT_ENGINEERING, SEC_DEPT_MEDICA
 	suit = /obj/item/clothing/suit/armor/vest/alt
 	shoes = /obj/item/clothing/shoes/jackboots
 	l_pocket = /obj/item/restraints/handcuffs
-	r_pocket = /obj/item/assembly/flash/handheld
+	r_pocket = /obj/item/restraints/legcuffs/bola/energy
 	suit_store = /obj/item/gun/energy/disabler
 	backpack_contents = list(/obj/item/melee/baton/loaded=1)
 


SCortesBar:
This is good change because especially new sec players won't even know what bolas are and will be forced to learn. If they become commonplace usage for sec they will become a much more robust force and not by giving them something OP, but by learning something they've had all the time.

Readystorm546:
Welp rip any borg uprising i guess. Also bola is extremely powerful.

Takahiru:
a bola can be made with a bit of metal and cable restraints, but an energy bola is too much? Also flashes can be found just about everywhere so if security is literally destroyed or something and borgs have taken over, I don't think taking away their flash is going to help against a 'borg uprising'

maxion12345:
nah, let em spawn with the flash in the pocket. Instead, have it replace belt-flash, to ensure sec officers can't be ganked by a Borg right when they leave arrivals and haven't managed to grab equipment yet.

SCortesBar:
Question. If "What if evil borg ganks them at arrivals" (which is against our rules) is an argument **against this**, why isn't "what if a Hulk decides to kill the sec officer at arrivals" an argument **in favor** of it?

The issue I suggest this aims to solve is not so much "officers getting ganked at arrivals", and more "we need to TEACH officers that they have a SUPER USEFUL TOOL that the majority of them do not use" and it's best to put it in a place they CAN'T ignore it.

Firewolf34:
Another thing about bolas is they are not OP. You're literally throwing your weapon at the enemy, if they catch it, or you miss, they can throw it back at you.

It's a useful tool, there's not a lot of reason to give sec two flashes, they can get the tool from the secvendor anyways, so why not just start them with one? It's just a roundstart convenience so they don't have to check the vendor for it every round. Robust sec will probably have one in their bag anyways.

Firewolf34:
Yo you need to rename this to "Sec get both a bola and a flash at start instead of two flashes" or something because people are taking it as if you are removing their flashes completely:

> Welp rip any borg uprising i guess.

Jarod1200:
just stick the bola in the backpack with the baton

ghost:
They're in the fucking vendors just go to the vendor you idiots

maxion12345:
Just stick in em the secbelt every officer gets from brig. Not only is that where ebolas come from (the vendor) but you also get rid of the flash you never used because you already had one in your pocket.

alexkar598:
10 vs 2. Pretty stupid

PR #9372, opened 2020-08-01
Fixes #9350

### LINKED ISSUE #9350: Chasms can no longer be walked over
https://github.com/yogstation13/Yogstation/pull/8977
image
good one alex car

ynot01:
Fixes #9350
Reverts a bit from #8977

### DIFF
--- a/code/game/turfs/simulated/chasm.dm
+++ b/code/game/turfs/simulated/chasm.dm
@@ -9,6 +9,7 @@
 	canSmoothWith = list(/turf/open/floor/fakepit, /turf/open/chasm)
 	density = TRUE //This will prevent hostile mobs from pathing into chasms, while the canpass override will still let it function like an open turf
 	bullet_bounce_sound = null //abandon all hope ye who enter
+	CanAtmosPass = ATMOS_PASS_NO
 
 /turf/open/chasm/Initialize()
 	. = ..()
@@ -75,6 +76,9 @@
 				to_chat(user, "<span class='warning'>You need one floor tile to build a floor!</span>")
 		else
 			to_chat(user, "<span class='warning'>The plating is going to need some support! Place metal rods first.</span>")
+
+/turf/open/chasm/CanPass(atom/movable/mover, turf/target)
+	return 1
 
 // Chasms for Lavaland, with planetary atmos and lava glow
 /turf/open/chasm/lavaland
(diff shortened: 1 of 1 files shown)


code/game/turfs/simulated/chasm.dm
```diff
@@ -76,6 +77,9 @@
 		else
 			to_chat(user, "<span class='warning'>The plating is going to need some support! Place metal rods first.</span>")
 
+/turf/open/chasm/CanPass(atom/movable/mover, turf/target)
+	return 1
```
Bibby0110:
```suggestion
	return TRUE
```

code/game/turfs/simulated/chasm.dm
```diff
@@ -9,6 +9,7 @@
 	canSmoothWith = list(/turf/open/floor/fakepit, /turf/open/chasm)
 	density = TRUE //This will prevent hostile mobs from pathing into chasms, while the canpass override will still let it function like an open turf
 	bullet_bounce_sound = null //abandon all hope ye who enter
+	CanAtmosPass = ATMOS_PASS_NO
```
Bibby0110: But... Isn't it just a... hole? Why wouldn't atmos pass?

ynot01: pretty sure the whole reason this happened in the first place is because chasms would repeatedly suck in air that was being generated from lavaland tiles

SomeguyManperson: they shouldn't?

ynot01: consider this a revert of part of #8977 then

code/game/turfs/simulated/chasm.dm
```diff
@@ -76,6 +76,9 @@
 		else
 			to_chat(user, "<span class='warning'>The plating is going to need some support! Place metal rods first.</span>")
 
+/turf/open/chasm/CanPass(atom/movable/mover, turf/target)
+	return TRUE
+
```
alexkar598:
```suggestion
/turf/open/chasm/CanAllowThrough(atom/movable/AM, turf/target)
	. = ..()
	return TRUE
```

alexkar598 (changes requested):

ynot01:
hey look the lint works now

alexkar598 (approved):

PR #9373, opened 2020-08-01
Obsession will forcibly enable mood since its effects are based on mood

SomeguyManperson:
mood is re-disabled if you are cured of it though

:cl:  
rscadd: Creep/obsession/what have you will forcibly enable mood since its effects are tied to mood, mood will be re-disabled for people who had it off if the trauma is cured
/:cl:


alexkar598 (changes requested):
Add a message telling them they got mood enabled because of obsession. Enabling features like that without saying why is bad UX

Firewolf34:
This is necessary.

SomeguyManperson:
doine

alexkar598 (approved):

PR #9383, opened 2020-08-04
Remove Cage Disposal System in Xenobio

alao3alao3:
For reason, see:
apng
As someone who never touches xenobio except to steal grey slime extract, I have no idea whether this person is telling the truth or not.

Either way, this change was dirt cheap, and mapping is easy when you don't have to use github for anything other than the pr request.

Below is a gallery of changes:
bpng
cpng
Highlighted in the above pic incorrectly is nothing. Below it is a reinforced table in place of the initial chute.
#### Changelog

:cl:  
rscadd: A reinforced table in place of a disposal chute
rscdel: Disposal system for cages in xenobio  
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -37219,13 +37219,6 @@
 /obj/structure/sign/warning/electricshock,
 /turf/closed/wall/r_wall,
 /area/science/xenobiology)
-"bIy" = (
-/obj/structure/disposaloutlet{
-	dir = 1
-	},
-/obj/structure/disposalpipe/trunk,
-/turf/open/floor/engine,
-/area/science/xenobiology)
 "bIz" = (
 /obj/machinery/portable_atmospherics/canister/carbon_dioxide,
 /obj/effect/turf_decal/bot,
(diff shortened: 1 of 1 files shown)


jcatjmeow:
as someone who uses the chutes often since I don't like to use the console, I do not like this change. Plus your reasoning is awful

maxion12345:
can you keep the test chamber one atleast that one is useful for stuff like gold cores ;(

Ratvar:
While PR is bad, reasoning does not seem so - a player tells X is bad, PR author trusts that X is bad, PR is made in case X is bad and people want it gone. PR author on discord seems to have been deceived by demonic Spooder, and PR is left open because they do not think they should close it just because they were deceived. 
image

MonsWTF:
The chutes for the cages was in place long long before the Xenobiology consoles had the capacity to directly 'drop' monkeys into the cages with the slimes, and to prevent the scientist from having to walk in themselves and risk getting grabbed by the first generations of slimes. 

As a relic of another age, it technically is fine still existing in the game; it means that those that want to do "purist Xenobio" can still do it. However, if disposals chutes being removed resulted in some minute reduction in game lag, it might eventually be a good thing to consider. 

But right now, concur with Ratvar, PR created by a guy told by another guy that something has absolutely no use when it occasionally does.

patpol4:
I don't like this change, if you were really going to do this then you would have to do it for every map, also it looks bad and people actually do use the chutes, especially the secure containment one

Moja10:
You have been dooped, bamboozled

fluffe9911:
We are at 8 downvotes and theres a mapfile conflict now so just gonna close this

PR #9393, opened 2020-08-05
Tcomms can no longer translate every language

### LINKED ISSUE #8948: singal tech translating powergaming!!
Tcomms scripts can translate codespeak and other languages they probably shouldn't be able to translate.

Long time bug but finally making an "issue" for it

Hopekz:
Removes some signal tech powergaming by only allowing them to translate the base languages (So they can't translate codespeak, Japanese, cult speaks etc).

Having them be able to translate antagonist languages has been a bug for a while as it was never intended to be able to translate these languages (full list of intended languages in the file).

Signal techs now only can translate the base languages.

fixes #8948

**EDIT:**
Altoid's wants me to make an easier to search changelog for documentation sake::

I broke your NTSL script, here is how:
- Your new language bitflag under newlang is converted to a language datum, and compared to the old language datum. 
- If there is a difference between the newlang and oldlang datums at this stage: newlang is compared against the allowed datums in the new global allowed list under allowed_translations. 
- If newlang contains a datum that isn't on the global allowed_translations list : newlang is set to oldlang so that no translation takes place.

#### Changelog

:cl:  Hopek
bugfix: Signal techs can no longer power game by translating antagonist languages.
tweak: Signal techs can now only translate the base languages.
/:cl:

### DIFF
--- a/yogstation/code/modules/scripting/Implementations/Telecomms.dm
+++ b/yogstation/code/modules/scripting/Implementations/Telecomms.dm
@@ -13,7 +13,7 @@
 #define DRACONIC 64
 #define BEACHTONGUE 128
 GLOBAL_LIST_INIT(allowed_custom_spans,list(SPAN_ROBOT,SPAN_YELL,SPAN_ITALICS,SPAN_SANS,SPAN_COMMAND,SPAN_CLOWN))//Span classes that players are allowed to set in a radio transmission.
-GLOBAL_LIST_INIT(allowed_translations,list(HUMAN,ROBOT,DRACONIC))// languages that players are allowed to translate to in a radio transmission.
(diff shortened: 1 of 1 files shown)


yacabo:
not longer

alexkar598 (dismissed):
What the hell is the point of whitelisting the target languages without whitelisting the source language?

Test your code before opening PRs

Hopekz:
> What the hell is the point of whitelisting the target languages without whitelisting the source language?

This code only happens when a translation takes place meaning we don't need to whitelist the source because it won't come in to effect otherwise.

Hopekz:
> > What the hell is the point of whitelisting the target languages without whitelisting the source language?
> 
> This code only happens when a translation takes place meaning we don't need to whitelist the source because it won't come in to effect.

Hopekz:
> What the hell is the point of whitelisting the target languages without whitelisting the source language?

NVM found my typo. 
You could have just said that there was a typo because you made it seem like there was some logic error.

fluffe9911 (approved):
ok

PR #9394, opened 2020-08-05
Reworks CQC to have less stuns and adds it to the traitor uplink

SomeguyManperson:
CQC as it currently stands is in an awkward spot where it's both really fucking good in melee only combat since it's got several sleep combos that take 2-4 hits to get and several other stuns that hurt roughly as much.
New CQC does several different things

- All attacks deal stamina damage, and martial arts can now be set to be compatible with pacifism
- All stuns aside from the restraining>chokehold stun have been replaced with either competitive stamina damage, shorter stuns and movement impairing effects, or both
- Chokehold stun is no longer instant, instead, it will deal oxygen damage every 1.5 seconds, with the damage being increased by stamina as a % (100 stamina damage = 20 oxygen damage, 200 stamina damage = 30, etc), the knockout will occur once oxygen damage reaches 50 or once total damage reaches 80
- Martial arts blocking has been remade entirely since the old system was stupid and bad and dumb and stupid, martial art block effects are handled internally, and CQC's block will provoke a short stun and a longer knockdown similarly to its combos, while putting the blocked item in the defender's off-hand.

Idea is to make both using a variety of combos important, make using the martial art viable and make it not nearly as stunbaton hands as it used to be
complain loudly if the numbers are bad and stupid because some PROBABLY ARE

new combos/attacks:
Disarm: deals 5 stamina damage instead of 5 brute damage

Grab: grabs will cause a short 1.5 second stun but no violent grabbing.

Harm: deals 15 stamina damage and 1 second immobilize per hit, knockdown if user is resting or knocked down, user's resting/knockdown is immediately ended on use.

Slam attack (grab, harm): Only works if target is standing, target takes 10 stamina damage, a 1 second paralyze, and a 6 second knockdown if not prone

Kick attack (disarm, harm): deals 10 stamina damage and pushes the target away, instead deals 35 stamina damage and mutes for ~8 seconds if they are prone

Restraining position (grab, grab): target is immediately aggressive grabbed, chain with disarm to start a chokehold attack that deals 10 oxygen damage per 1.5 seconds with stamina damage from the start of the attack acting as a % bonus multiplier, knocks out for 40 seconds if the target falls unconscious or their combined damage amount (all potentially lethal) of the next attack would reach 80%. Attack is interrupted on movement, losing the grab, or anything else that would cause a normal channel to be interrupted.
If interrupted, grab state is reduced by 1, and the target instantly rolls a free full % chance to break out of the weaker grab. On successful choke attack, target is put into a neck grab instantly.

Consecutive (harm, harm, harm, harm, harm): Only works if the target is standing, paralyzes for 2 seconds, and knocks down for 10 seconds, as well as dealing 25 stamina damage on the final hit instead of 15.

Pressure point (disarm, disarm, grab): target takes 50 stamina damage to the targeted limb (random if targeting head or chest) and drops anything they are holding.

:cl:  
rscadd: CQC has been reworked entirely and is now available in the traitor uplink
tweak: martial art melee blocks are handled inside the martial art rather than by the item/just blocking
/:cl:


ynot01:
cooks btfo

Marmio64:
so basically just a shitty sleeping carp that also doesn't block projectiles or your ability to use guns? Chefs nerf is fine, they don't need OP CQC but the nerf to nukies martial arts is incredible. Like, this new CQC will have no use at all to nukies, they'll always just be better off going with the cheaper and better option of the krav maga gloves they can buy which will at least still have some instant stuns.

edit: also, can people who know CQC just not hurt people in melee ever again, unless they do a slow as hell choke?

SomeguyManperson:
>  but the nerf to nukies martial arts is incredible. Like, this new CQC will have no use at all to nukies, they'll always just be better off going with the cheaper and better option of the krav maga gloves they can buy which will at least still have some instant stuns.

nukies already don't use cqc like ever since you can buy actual guns for the same amount, also the active block/counter system will actually be punishing since in testing every 1-3 hits results in a successful block which
A: forces the attacker to drop their weapon
B: instantly equips their weapon in your off-hand if possible
C: knocks down and stuns the attacker
rather than having a hidden edge case punish to attacking with a baton


> edit: also, can people who know CQC just not hurt people in melee ever again, unless they do a slow as hell choke?

you can it's called "having a toolbox", you're not limited to just the martial art for melee and it doesn't cost 75% of your TC in any case

Marmio64:
> > but the nerf to nukies martial arts is incredible. Like, this new CQC will have no use at all to nukies, they'll always just be better off going with the cheaper and better option of the krav maga gloves they can buy which will at least still have some instant stuns.
> 
> nukies already don't use cqc like ever since you can buy actual guns for the same amount, also the active block/counter system will actually be punishing since in testing every 1-3 hits results in a successful block which
> A: forces the attacker to drop their weapon
> B: instantly equips their weapon in your off-hand if possible
> C: knocks down and stuns the attacker
> rather than having a hidden edge case punish to attacking with a baton
> 
> > edit: also, can people who know CQC just not hurt people in melee ever again, unless they do a slow as hell choke?
> 
> you can it's called "having a toolbox", you're not limited to just the martial art for melee and it doesn't cost 75% of your TC in any case

i meant, not ever being able to hurt someone with their fists in melee, and CQC used to be a great damage dealer, something it would no longer be. And nukies do use CQC sometimes, ive seen them do it.

Marmio64:
Still neutral on this, but another big thing you should probably change is the krav maga gloves, since they're more of actual baton hands, while CQC is sleep hands and sleeping carp is murder hands. 

Should probably make leg sweep instead knock people down for 3 seconds instead of doing a actual stun right off the bat (with shove rework, you'll still be able to shove them right away to apply like a 2 second stun.). Also IIRC, legsweep does a tiny bit of damage but I may be wrong, if it does, it should be changed to deal no damage, since it can't be worse than getting smacked with a full on telescopic baton and stunned.

edit: you can't shove with krav maga so, probably should just ignore all i've said above

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1197,21 +1197,23 @@ GLOBAL_LIST_EMPTY(roundstart_races)
...
-	if(target.check_block())
-		target.visible_message("<span class='warning'>[target] blocks [user]'s grab attempt!</span>")
-		return 0
+	var/datum/martial_art/M = target.check_block()
+	if(M)
+		M.handle_counter(target, user)
+		return FALSE
 	if(attacker_style && attacker_style.grab_act(user,target))
-		return 1
+		return FALSE
```
alexkar598: False?

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1197,21 +1197,23 @@ GLOBAL_LIST_EMPTY(roundstart_races)
...
 	else
 		target.grabbedby(user)
-		return 1
+		return FALSE
```
alexkar598: False?

code/modules/uplink/uplink_items.dm
```diff
@@ -559,7 +559,6 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	name = "CQC Manual"
 	desc = "A manual that teaches a single user tactical Close-Quarters Combat before self-destructing."
 	item = /obj/item/book/granter/martial/cqc
-	include_modes = list(/datum/game_mode/nuclear, /datum/game_mode/nuclear/clown_ops)
```
alexkar598, line 562: So why exactly cant nukies use it anymore?

code/datums/martial/cqc.dm
```diff
@@ -1,30 +1,55 @@
...
+#define PRESSURE_COMBO "DDG"
+///consecutive combo string
+#define CONSECUTIVE_COMBO "HHHHH"
 
 /datum/martial_art/cqc
 	name = "CQC"
 	id = MARTIALART_CQC
 	help_verb = /mob/living/carbon/human/proc/CQC_help
 	block_chance = 75
+	nonlethal = TRUE //all attacks deal solely stamina damage or knock out before dealing lethal amounts of damage
```
alexkar598, line 30: Wont this allow pacifists to harm people using CQC despite being a pacifist?

code/datums/martial/cqc.dm
```diff
@@ -47,49 +72,77 @@
...
+  */
 /datum/martial_art/cqc/proc/Pressure(mob/living/carbon/human/A, mob/living/carbon/human/D)
 	if(!can_use(A))
 		return FALSE
 	log_combat(A, D, "pressured (CQC)")
 	D.visible_message("<span class='warning'>[A] punches [D]'s neck!</span>")
-	D.adjustStaminaLoss(60)
+	D.drop_all_held_items()
+	D.adjustStaminaLoss(30)
+	D.Immobilize(60)
```
alexkar598: Id drop the immobilize and deal 50 stam as its a bit less complex than slam + kick but you cant chain it

code/datums/martial/cqc.dm
```diff
@@ -146,72 +200,129 @@
...
-		playsound(get_turf(D), 'sound/weapons/cqchit1.ogg', 50, 1, -1)
+	var/bonus_damage = 15
+	D.apply_damage(bonus_damage, STAMINA)
+	playsound(get_turf(D), 'sound/weapons/cqchit1.ogg', 50, 1, -1)
 	D.visible_message("<span class='danger'>[A] [picked_hit_type] [D]!</span>", \
 					  "<span class='userdanger'>[A] [picked_hit_type] you!</span>")
+	D.Immobilize(15)
 	log_combat(A, D, "[picked_hit_type] (CQC)")
-	if(A.resting && !D.stat && !D.IsParalyzed())
+	if(!(A.mobility_flags & MOBILITY_STAND) && !D.stat && (D.mobility_flags & MOBILITY_STAND))
```
alexkar598:
```suggestion
	if(!(A.mobility_flags & MOBILITY_STAND) && (D.mobility_flags & MOBILITY_STAND))
```

code/datums/martial/cqc.dm
```diff
@@ -146,72 +200,129 @@
...
 		D.visible_message("<span class='warning'>[A] leg sweeps [D]!", \
 							"<span class='userdanger'>[A] leg sweeps you!</span>")
 		playsound(get_turf(A), 'sound/effects/hit_kick.ogg', 50, 1, -1)
-		D.apply_damage(10, BRUTE)
-		D.Paralyze(60)
+		D.Paralyze(10)
+		D.Knockdown(60)
```
alexkar598: Kinda alot of knockdown there, could use a nerf to like 3 seconds

code/datums/martial/cqc.dm
```diff
@@ -146,72 +200,129 @@
...
 			I = D.get_active_held_item()
-			D.visible_message("<span class='warning'>[A] strikes [D]'s jaw with their hand!</span>", \
-								"<span class='userdanger'>[A] strikes your jaw, disorienting you!</span>")
+			D.visible_message("<span class='warning'>[A] quickly grabs [D]'s arm and and chops it, disarming them!</span>", \
+								"<span class='userdanger'>[A] grabs your arm and chops it, disarming you!</span>")
 			playsound(get_turf(D), 'sound/weapons/cqchit1.ogg', 50, 1, -1)
 			if(I && D.temporarilyRemoveItemFromInventory(I))
 				A.put_in_hands(I)
 			D.Jitter(2)
-			D.apply_damage(5, A.dna.species.attack_type)
```
alexkar598, line 186: note2self

alexkar598 (dismissed):

code/datums/martial/cqc.dm
SomeguyManperson: at most they can deal 50 oxygen damage from the choke attack, everything else deals stamina damage and a choke will end before it puts the target into crit

code/datums/martial/cqc.dm
SomeguyManperson: doing roughly that

code/modules/uplink/uplink_items.dm
SomeguyManperson: removing include_modes means it's included in all modes, it functions as a whitelist where as exclude_modes or whatever is a blacklist

JamieD1 (approved):
Here we go bois

PR #9398, opened 2020-08-05
fixes a paystand runtime when inserting money on an unregistered paystand

ynot01:
[18:15:47] Runtime in pay_stand.dm, line 100: Cannot read null.registered_account


fluffe9911 (approved):
ok

PR #9401, opened 2020-08-06
2 new Preternis surgeries allow them to detach and attach their robotic limbs

Hopekz:
**2 new surgeries for Preternis allowing them to change their limbs**

- Detach mechanical limb
- Attach mechanical limb

**Areas of effect:**  
both arms and both legs

Surgery steps:
-----

**Detach mechanical limb**
- start surgery (only on mechanical limbs)
- screwdriver
- hand to open hatch
- wrench
- multi-tool to prepare electronics
- hand to detach the limb

**Attach mechanical limb**
- start surgery
- multimeter to prepare the connection
- attach the new limb

#### Changelog

:cl:  Hopek
rscadd: Preternis now have two new surgeries. They are now able to Detach and Attach mechanical limbs.
/:cl:

### DIFF
--- a/code/modules/surgery/amputation.dm
+++ b/code/modules/surgery/amputation.dm
@@ -22,9 +22,7 @@
 		)
 
 /datum/surgery/amputation/mechanic/can_start(mob/user, mob/living/carbon/target)
-	if(!iscarbon(target) || !ispreternis(target)) // This surgery won't work unless you are working on a preterni
-		return 0
-	return 1
+	return ispreternis(target))
 
 /datum/surgery_step/sever_limb
 	name = "sever limb"
--- a/code/modules/surgery/amputation.dm
+++ b/code/modules/surgery/amputation.dm
@@ -22,7 +22,7 @@
 		)
 
 /datum/surgery/amputation/mechanic/can_start(mob/user, mob/living/carbon/target)
-	return ispreternis(target))
+	return ispreternis(target)
 
 /datum/surgery_step/sever_limb
 	name = "sever limb"
--- a/code/modules/surgery/amputation.dm
+++ b/code/modules/surgery/amputation.dm
@@ -26,7 +26,6 @@
 		return 0
 	return 1
 
-
 /datum/surgery_step/sever_limb
 	name = "sever limb"
 	implements = list(/obj/item/scalpel = 100, /obj/item/circular_saw = 100, /obj/item/melee/transforming/energy/sword/cyborg/saw = 100, /obj/item/melee/arm_blade = 80, /obj/item/twohanded/required/chainsaw = 80, /obj/item/mounted_chainsaw = 80, /obj/item/twohanded/fireaxe = 50, /obj/item/hatchet = 40, /obj/item/kitchen/knife/butcher = 25)
--- a/code/modules/surgery/amputation.dm
+++ b/code/modules/surgery/amputation.dm
@@ -6,6 +6,26 @@
 	possible_locs = list(BODY_ZONE_R_ARM, BODY_ZONE_L_ARM, BODY_ZONE_L_LEG, BODY_ZONE_R_LEG, BODY_ZONE_HEAD)
 	requires_bodypart_type = 0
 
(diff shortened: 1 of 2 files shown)


ghost:
Yes. Speedmerge

code/modules/surgery/amputation.dm
```diff
@@ -6,6 +6,25 @@
...
+		/datum/surgery_step/open_hatch,
+		/datum/surgery_step/mechanic_unwrench,
+		/datum/surgery_step/prepare_electronics,
+		/datum/surgery_step/sever_limb/mechanic
+		)
+
+/datum/surgery/amputation/mechanic/can_start(mob/user, mob/living/carbon/target)
+	if(!iscarbon(target) || !ispreternis(target)) // This surgery won't work unless you are working on a preterni
+		return 0
+	return 1
```
alexkar598: - You can't be a preternis without being a human and by extension, a carbon(i dont think you can even do surgery on non carbons so what the hell?)
- Use FALSE/TRUE instead of 0/1
- If(boolean) return TRUE else return FALSE is a stupid pattern
- You don't need to be a rocket scientist to know what can_start() return ispreternis means, do you really need to comment that, but w/e, more comments is better than not enough comments I GUESS
```suggestion
/datum/surgery/amputation/mechanic/can_start(mob/user, mob/living/carbon/target)
	return ispreternis(target))
```

code/modules/surgery/amputation.dm
```diff
@@ -27,3 +46,23 @@
 		target_limb.drop_limb()
 
 	return 1
+
+/datum/surgery_step/sever_limb/mechanic
+	name = "uninstall limb"
+	accept_hand = 1
```
alexkar598:
```suggestion
	accept_hand = TRUE
```

code/modules/surgery/prosthetic_replacement.dm
```diff
@@ -13,7 +13,24 @@
...
+	target_mobtypes = list(/mob/living/carbon/human)
+	lying_required = FALSE
+	steps = list(
+		/datum/surgery_step/prepare_electronics,
+		/datum/surgery_step/add_prosthetic
+		)
 
+/datum/surgery/prosthetic_replacement/mechanic_limb/can_start(mob/user, mob/living/carbon/target)
+	if(!iscarbon(target) || !ispreternis(target)) // This surgery won't work unless you are working on a preterni
+		return 0
```
alexkar598: Same thing
```suggestion
/datum/surgery/prosthetic_replacement/mechanic_limb/can_start(mob/user, mob/living/carbon/target)
	if(!ispreternis(target))
		return FALSE
```

code/modules/surgery/prosthetic_replacement.dm
```diff
@@ -13,7 +13,24 @@
...
+
+	var/mob/living/carbon/C = target
+	if(!C.get_bodypart(user.zone_selected)) //can only start if limb is missing
+		return 1
```
alexkar598: You are making me go insane
```suggestion
	return !C.get_bodypart(user.zone_selected) //can only start if limb is missing
```

alexkar598 (dismissed):

Marmio64:
preternis powercreep?

code/modules/surgery/amputation.dm
Hopekz: > (i dont think you can even do surgery on non carbons so what the hell?)

Future proofing.

> Use FALSE/TRUE instead of 0/1

okay

> If(boolean) return TRUE else return FALSE is a stupid pattern

 I copied the syntax of what was already here but you are right. This file is ugly.

> You don't need to be a rocket scientist to know what can_start() return ispreternis means

Sometimes I wonder if I need to clarify too but then I remember we're trying to make it readable for everyone (different languages and skill level) so I comment anyway. 
Someone can read ispreternis as a single word and be confused etc.

More retard proofing = better imo that's why we started using autodoc I think.

fluffe9911 (approved):
p o w e r c r e e p

PR #9404, opened 2020-08-06
Adds radio SFX

MrHorizons:
A semi-port of [NSV #8](https://github.com/BeeStation/NSV13/pull/8).

Since I think that Yog is rather lacking in the sound department (or at least, outside of engineering and combat), I think simply adding radio SFX will add a good amount to the game without being too drastic. After all, who _doesn't_ use a radio?

#### Changelog

:cl:  
rscadd: Radios now make sound when you use them.
/:cl:


Firewolf34:
Sound effects are awesome and we totally need more!

I can't seem to hear anything when I play the radio1.ogg file, though. Maybe it works in BYOND? Might be my browser... The other two do make noise though.

Also I would make sure they don't play when you are deaf. But that may already be how your code works, I didn't look into it.

fluffe9911 (approved):
Immersion is cool

ynot01:
badass

PR #9405, opened 2020-08-06
Ports "Ports TGMC timer cooldowns (now supporting signals)"

Bibby0110:

Ports github.com/tgstation/tgstation/pull/51388

This is less performant than storing world.time in a variable, but it's a simple, quick, easy and dirty way to add a cooldown to something. Just create an ID and you need do no more besides using the macros.

Helps with clarity and adding cooldowns on the fly.


patpol4:
code

Bibby0110:
> code

code?

alexkar598 (approved):

PR #9411, opened 2020-08-08
Fixes happy cluwne mask, removes the shitcode and documents it

alexkar598:
Way less shitcody now, also it works also this is a bounty PR

### DIFF
--- a/yogstation/code/modules/clothing/mask.dm
+++ b/yogstation/code/modules/clothing/mask.dm
@@ -1,88 +1,93 @@
+/**
+  * An undroppable mask that changes the user's speech to be unintelligable
+  */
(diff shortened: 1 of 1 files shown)


Marmio64:
what does this do though. And you say its a bounty but dont link anything? You made ynot put a description in his narsie gets killed by singularity PR why should this be any different.

alexkar598:
It actually works now, also theres nothing to link, it used to just make you a cluwne 100% of the time instead of the intended 1% and 0.1% to spawn a floor cluwne, now it works as intended

maxion12345:
where changelog

fluffe9911 (approved):
Ok

PR #9425, opened 2020-08-10
Fixes huge scrubber's unmovable/connect to port bug

Gelissean:
Fixes an issue where using a wrench to secure the huge scrubber to the floor would not call the parent function, making it unable to connect to connector ports.

:cl:  
bugfix: Huge scrubbers can now connect to gas ports. 
/:cl:


redmoogle:
I thought this was intentional, neet

fluffe9911 (approved):
neat

PR #9437, opened 2020-08-12
Revert "Adds radio SFX"

boodaliboo:
Reverts yogstation13/Yogstation#9404
i truthfully find it incredibley annoying


Mesalikepie:
Maybe make it a preference? I love the sound effect personally

maxion12345:
it'd be better if it was one sound instead of two, with a preference option

ynot01:
i reallyyy like it

boodaliboo:
it just is incredibly distracting for me i would much rather like it being a preference but i cant code that

Firewolf34:
Do not just remove this. Add an option to disable it, if you don't like it.

Considering nobody pays any attention to the damn chat these days, because of the runetext, we need some way to stop people from walking around with radio mics on.

boodaliboo:
> Do not just remove this. Add an option to disable it, if you don't like it.
> 
> Considering nobody pays any attention to the damn chat these days, because of the runetext, we need some way to stop people from walking around with radio mics on.

i have no clue how to add one so im doing the next best thing

alexkar598:
Most people seem to like it to shrug

PR #9441, opened 2020-08-12
Cryo pods now don't stop when cloning damage is present on MEDICALIGNORE

Hopekz:
This means that Preterni can now heal cloning damage from cryo pods again.

image
image

#### Changelog

:cl:  Hopek
tweak: Cryo pods don't stop when cloning damage is present on MEDICALIGNORE. This means that Preterni can now heal cloning damage in cryopods but will still not be trapped after they're healed even with brute or burn damage.
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
@@ -182,7 +182,7 @@
 		return
 	
 	/// Individuals with the MEDICALIGNORE trait will stop the cryo from functioning and display a unique warning unless there is clone damage on the body which cryo happens to be able to heal even with MEDICALIGNORE (oversight probably but one of the one ways to heal their clone damage atm). - Hopek
-	if(HAS_TRAIT(mob_occupant,TRAIT_MEDICALIGNORE) && mob_occupant.getCloneLoss())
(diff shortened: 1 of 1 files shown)


code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
```diff
@@ -181,7 +181,8 @@
 	if(mob_occupant.stat == DEAD) // We don't bother with dead people.
 		return
 	
-	if(HAS_TRAIT(mob_occupant,TRAIT_MEDICALIGNORE))
+	/// Individuals with the MEDICALIGNORE trait will stop the cryo from functioning and display a unique warning unless there is clone damage on the body which cryo happens to be able to heal even with MEDICALIGNORE (oversight probably but one of the one ways to heal their clone damage atm). - Hopek
+	if(HAS_TRAIT(mob_occupant,TRAIT_MEDICALIGNORE) && (mob_occupant.getCloneLoss() < 1))
```
alexkar598:
```suggestion
	if(HAS_TRAIT(mob_occupant,TRAIT_MEDICALIGNORE) && mob_occupant.getCloneLoss())
```

alexkar598 (dismissed):

JamieD1 (approved):

PR #9444, opened 2020-08-12
Revamps the backup shuttle.

MrHorizons:
I've changed the backup shuttle to, well, resemble a shuttle.

**Old:**
image

**New:**
image

#### Changelog

:cl:  
rscadd: Adds a new version of the backup shuttle, so it is no longer a windowless box of chaos.
/:cl:

### DIFF
--- a/_maps/shuttles/emergency_backup.dmm
+++ b/_maps/shuttles/emergency_backup.dmm
@@ -1,18 +1,105 @@
 //MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
-"a" = (
-/obj/machinery/computer/emergency_shuttle{
+"b" = (
+/obj/machinery/light{
+	dir = 8
+	},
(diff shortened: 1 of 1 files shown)


alexkar598:
I'd rather not, the point of it is that its so simple nothing could go wrong when trying to load it in in the event of something going wrong

Reed0506:
square is superior

MrHorizons:
> I'd rather not, the point of it is that its so simple nothing could go wrong when trying to load it in in the event of something going wrong

... I guess that makes sense.

Marmio64:
should add the new design in as a new shuttle and call it 'Refurbished Backup Shuttle', desc could be: "A refurbished model of the backup shuttle we have stored in centcom for your station in the event of a emergency. However, it could not pass the model specifications, so we'll pay you 500 credits to take it off our hands."

And yes, please make it give money, because it is so god damn tiny. Could also instead just make it a emag shuttle that traitors can get to make hijack easier for themselves.

alao3alao3:
I just hope it fits in that same spot the old one fit.

PR #9460, opened 2020-08-13
[admin] Ticket based activity tracker

AsV9:
# Intent of your Pull Request

Implements a ticket based activity tracker.

Basically, whenever a ticket is left unclaimed it will log any admins who do not have a ticket assigned to them as inactive so that they do not receive tracked hours. This ensures that hours are being fairly tracked for admins who handle ticket, and punishes people who prioritize playing instead of handling tickets.

An additional benefit of this system is that it does not discriminate based on rank. Council, admins, retmins and host are all treated the same way by the code.

**I am aware that there are ways to game this system. I have had a thunk about how to improve this system, but I really can't be fucked to do anything about it now, and this is better than nothing.**

Additionally I also refactored how the 'rebwoink' system works. Basically before it would spam you with a notification every 50 seconds regardless of whether or not you had 4 active tickets already. Now it sends individual notifications for every unclaimed ticket, so you don't get like 10 bwoinks at the same time, instead they get spread out in accordance as to when the tickets came in.

Additionally, it will only send you an audio bwoink if you do not have any unclaimed tickets.

:cl:  
rscadd: New admin ticket based activity checker
/:cl:

### DIFF
--- a/code/_globalvars/lists/mobs.dm
+++ b/code/_globalvars/lists/mobs.dm
@@ -2,6 +2,7 @@ GLOBAL_LIST_EMPTY(clients)							//all clients
 GLOBAL_LIST_EMPTY(admins)							//all clients whom are admins
 GLOBAL_PROTECT(admins)
 GLOBAL_LIST_EMPTY(deadmins)							//all ckeys who have used the de-admin verb.
+GLOBAL_LIST_EMPTY(ticket_logout_admins)				//all ckeys who have been marked as inactive by the ticket activity checker
 
 GLOBAL_LIST_EMPTY(directory)							//all ckeys with associated client
 GLOBAL_LIST_EMPTY(stealthminID)						//reference list with IDs that store ckeys, for stealthmins
--- a/yogstation/code/controllers/subsystem/yogs.dm
+++ b/yogstation/code/controllers/subsystem/yogs.dm
@@ -1,5 +1,4 @@
 #define ROUND_END_ANNOUNCEMENT_TIME 105 //the time at which the game will announce that the shuttle can be called, in minutes.
-#define REBWOINK_TIME 50 // Number of seconds before unclaimed tickets bwoink again and yell about being unclaimed
 
 SUBSYSTEM_DEF(Yogs)
 	name = "Yog Features"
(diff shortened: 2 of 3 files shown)


Hopekz:
The way I see it this new system is bypassed by just keeping a ticket open and incentivizes slow responses to ensure you always have a ticket open.

In this new system, you are punished for taking care of tickets quicker/more efficiently than other admins in the same shift.

AsV9:
> The way I see it this new system is bypassed by just keeping a ticket open

Yeah that's pretty much a weakness of the system, I've got some ideas for improving it which I'm not going to go into.

>  incentivizes slow responses to ensure you always have a ticket open.

Having one ticket open is exactly the same as having no tickets open. Did you just pull this shit straight out your ass?

Hopekz:
> Having one ticket open is exactly the same as having no tickets open. 

You misunderstand.

**I'll give you a scenario to help:**

Admin1: resolves 10 tickets in 20 minutes. Resolves tickets as they come. Doesn't hold on to them.
Admin2: resolves 3 tickets in 20 minutes. Holds on to each ticket for 10 minutes. can hold on to 2 tickets as well.

When Admin1 resolves a ticket efficiently one by one he is then punished by being put into a "no ticket" state.
Admin1 then has to take additional tickets just because his activity is not going to count because he resolved the previous tickets in a fast manner.

Admin2 does not get this "no ticket" state so he gets rewarded with less workload just for doing the tasks slower.

Both Admin1 and Admin2 get the same amount of credit for their work but Admin2's technique is what is encouraged in this system. The goal of this system is to do the least amount of work vs other Admins as that maximizes your "reward" vs work.

> Did you just pull this shit straight out your ass?

Did I trigger your PTSD or something?

alexkar598:
Admin politics must remain out of the repo. This PR will be locked and limited to maintainers if any kind of admin drama starts to stir up, you have discord channels and a forums, do it there.

alexkar598 (changes requested):
Exclude people without +BAN please.

AsV9:
It already does in adminhelp.dm#192 @alexkar598

alexkar598:
This would need to be adjusted to the new playtime tracker(job experience instead of connection logs)

Amelia0010:
@AsV9 Status?

Amelia0010:
asv is home on holiday so oh well

PR #9461, opened 2020-08-14
Adds cheese culture and wheel crates.

MrHorizons:
Cheese wasn't orderable from cargo before. Now, there's two cheese-related crates.

- **Cheese Culture Crate** - 800 credits - Contains the advanced cheese bacteria box found in the chef's refrigerator (freezer? I don't remember the specific name of it).
- **Cheese Wheel Crate** - 1,000 credits - Contains 6 random cheese wheels. An arguably far worse deal in exchange for being ready-made.

#### Changelog

:cl:  
rscadd: Allows the advanced cheese bacteria box to be purchased from cargo for 1,500 credits under the name of "Cheese Culture Crate". Don't want to make cheese? Get 6 random cheese wheels, ready-made, for only 1,000 credits.
/:cl:

### DIFF
--- a/yogstation/code/modules/cargo/cargo_packs.dm
+++ b/yogstation/code/modules/cargo/cargo_packs.dm
@@ -54,7 +54,7 @@
 /datum/supply_pack/organic/cheeseculture
 	name = "Cheese Culture Crate"
 	desc = "Contains a variety of advanced cheese bacteria cultures."
-	cost = 1500
+	cost = 800
 	contains = list(/obj/item/storage/box/cheese)
 	crate_name = "cheese culture crate"
 	
--- a/yogstation/code/modules/cargo/cargo_packs.dm
+++ b/yogstation/code/modules/cargo/cargo_packs.dm
@@ -51,12 +51,33 @@
 	crate_name = "Advanced Crate Container"
 	crate_type = /obj/structure/closet/crate/large
 
-/datum/supply_pack/organic/cheese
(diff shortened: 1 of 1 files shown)


ynot01:
bacteria shouldnt be 1500 credits, it should be cheap, ready, and unimpactful (meaning the chef shouldnt have to bitch and whine for cargo to process his order) in case the chef runs out (which usually never happens)

maybe 700 or 850 credits is better

otherwise pretty much only good things here

MrHorizons:
> bacteria shouldnt be 1500 credits, it should be cheap, ready, and unimpactful (meaning the chef shouldnt have to bitch and whine for cargo to process his order) in case the chef runs out (which usually never happens)
> 
> maybe 700 or 850 credits is better

Yeah, I doubt the chef will ever run out, but 800 credits makes more sense. Made that change.

missatessatessy:
This is okay, but I don't think parmesan should be in the random cheeses crate, it's considered a premium cheese and is intentionally harder to make than all the other cheeses because of how OP it is.

MrHorizons:
> This is okay, but I don't think parmesan should be in the random cheeses crate, it's considered a premium cheese and is intentionally harder to make than all the other cheeses because of how OP it is.

I mean, I feel like it's balanced out by only being a single wheel and only having a chance to appear, and each cheese wheel crate costs 1k.

JamieD1 (approved):
CHEESE

PR #9465, opened 2020-08-15
Revert "THE HONKMOTHER COMES" for being shitcode

alexkar598:
Very shitcode, shouldn't of been merged in this state

1) An indestructile /obj/structure/**destructible** wtf.....
2) convert range was not documented properly with the triple slash
3) using text2path()
4) Very unefficent code for something thats in process()


Darkstick:
Feels like a lot of work to just bin it... I hope I can see it fixed up soon

PR #9457, opened 2020-08-13
Conditional department goals

Amelia0010:
Adds conditional department goals, IE goals that will only be added if certain conditions are true. Two examples added here are based on engineering:

If the station has a supermatter engine, the objective will be to setup a singularity. If the station has a singularity/tesla, the objective will be to setup a supermatter.

~~Also adds a goal for cargo to make a tesla engine in the cargo bay. Because you know, cargo ain't gonna have other goals.~~

:cl:
rscadd: Engineering will now get a goal to set up the opposite engine of what they spawn with. 
/:cl:

PR #9469, opened 2020-08-15
cargo tesla goal

Amelia0010:
Cargo now gets a goal to setup a tesla in the cargo bay

:cl:
rscadd: Cargo can now get a goal to set up a tesla engine in the cargo bay
/:cl:

Split off from #9457 since I don't want to weigh down a framework PR with components using that framework


MrHorizons:
Would having the tesla be set up in the cargo shuttle count towards the goal? That's something of the meta when doing the (strangely rare) contained tesla ball bounty.

Tipy1802:
Cargo building the Tesla is considered powergaming and against server rules, please do not justify it with code

If you want better cargo goals then you can try “order crate x” “order x amount of crates” or “spend x amount of credits on crates”

Thanks for going through the troubles of separating this from the conditional goals pr

Amelia0010:
>Would having the tesla be set up in the cargo shuttle count towards the goal?

No, it wouldn't

>Cargo building the Tesla is considered powergaming and against server rules

[Citation needed]

Reed0506:
hmm yes today the crate pushers will construct an exceptionally complicated and dangerous engine that can and will destroy the entire station with a single mistake

Moja10:
Cargo building a tesla to satisfy a bounty is not against the rules and if this PR was merged, then it's definitely not against the rules. If you don't like cargo building teslas, make a post on the forums on why you think it shouldn't be allowed or should be against the rules, as there are a couple of admins who would agree with you. Also, make a PR to remove the cargo bounty to build a tesla and ship it, if you don't like cargo building teslas. 

But as it stands, it's fine for the bounty and this cargo goal.

MegaEmpirical:
Bounties are supposed to encourage inter-department interactions, no need to remove the cargo tesla bounty.

As for building an engine in cargo, I'm against this as it encourages that behavior more and it should just be an engineers job. I've consistently been against that cargo power game bullshit.

Tipy1802:
> [Citation needed]

Crate pushers that probably don’t even have a university degree wouldn’t know how to set up a fucking Tesla engine from scratch so it breaks metagaming rules

Setting up engines is engineering’s job not cargo’s (it’s even in the name!) so it breaks the powergaming rule

MegaEmpirical:
I wish Tippy but Council says otherwise and Cargo is allowed to make engines. I disagree strongly but nothing I can do about it.

Tipy1802:
Oh I thought I saw admins say otherwise months ago, I guess council over ruled that, still I am against this even if allowed it shouldn’t be encouraged

JamieD1 (approved):
If it causes issues this will be reverted

MegaEmpirical:
Nooooooooooooooooooo

PR #9482, opened 2020-08-16
Replaces Rotciv with Colton in wizard academy

LoliconSlayer:
Self explanatory title

Oh yeah and adds Anvilman as a wizard because people kept complaining about Feweh not being a yogs reference, so instead of replacing Feweh with Colton I added another person because golly me do I fucking hate retards

:cl:  
tweak: Rotciv the Dented and Feweh the Replicator have been expelled from the wizard academy! In their place, Colton the Afterburner and Anvil the Archivist have transferred in.
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/Academy.dmm
+++ b/_maps/RandomRuins/SpaceRuins/Academy.dmm
@@ -1090,8 +1090,8 @@
 /area/awaymission/academy/academycellar)
 "lC" = (
 /mob/living/simple_animal/hostile/wizard{
-	desc = "Found leading several other scribes, one of which was prone to rage";
-	name = "Anvil the Archiver"
+	desc = "Found leading several other archivists, one of which was prone to rage";
+	name = "Anvil the Archivist"
 	},
 /turf/open/floor/wood,
 /area/awaymission/academy/academycellar)
--- a/_maps/RandomRuins/SpaceRuins/Academy.dmm
+++ b/_maps/RandomRuins/SpaceRuins/Academy.dmm
@@ -1090,8 +1090,8 @@
 /area/awaymission/academy/academycellar)
 "lC" = (
 /mob/living/simple_animal/hostile/wizard{
-	desc = "Exiled from his old academy when he casted a illusion spell to make 16 clones of himself then claimed they were not his";
-	name = "Feweh The Replicator"
+	desc = "Found stranded on a remote island surrounded by cheese";
+	name = "Tessa the Fromager"
 	},
 /turf/open/floor/wood,
 /area/awaymission/academy/academycellar)
--- a/_maps/RandomRuins/SpaceRuins/Academy.dmm
+++ b/_maps/RandomRuins/SpaceRuins/Academy.dmm
@@ -970,7 +970,7 @@
 /area/awaymission/academy/academyaft)
 "kV" = (
 /mob/living/simple_animal/hostile/wizard{
-	name = "Rotciv The Dented"
+	name = "Colton the Afterburner"
 	},
 /turf/open/floor/carpet,
 /area/awaymission/academy/academyaft)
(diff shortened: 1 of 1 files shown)


boodaliboo:
sorry colton gotta defend my pr

patpol4:
Don't remove wizards, do an old one if you really need to, also i think there is like one left that isnt a reference to someone

LoliconSlayer:
Discord_KeYiaKUMVZ

Rotciv's one of the biggest shitters we've seen in recent times, I personally think it's better to not immortalize rule-breaking retards and immortalize actual people that are legendary for reasons beyond having an IQ the same as room temp in celsius.

ghost:
Awww gosh I'm honored Mord. I'll try adding new radiation based spells to wiz to make it fit the theme if this gets merged.

patpol4:
Still don't like it and am against
(nothing against colton)

LoliconSlayer:
Turdis why you fail

patpol4 (dismissed):
Alexkar asked me to approve this as a test

Marmio64:
next can you add a monument for all mentors in the academy

Hopekz:
Shittler or not he's made a big impact being the dented person that he is.
There were countless may may's made vs putting in a new obscure reference most won't get.

Being referenced is never about being "good" it is about people standing out which is why we have "Zero the Valid" etc

LoliconSlayer:
I see your point about Zero also being a shitter and using him as an example of a shitter being referenced, but the 2nd half of the point I was trying to make about not referencing Rotciv is that we should be referencing people within our community that have notability beyond just being dumb, if we're going off of memes that come from people, Colton's produced many more memes that still have some form of staying power (IE The Legion Engine) compared to the "Memes" that we had about Rotciv which were just varying ways to call him an idiot, none of which we even post anymore because he's been permabanned from the server and persona non grata'd from the Discord (before he left that is). The only reason I put Anvil in is that people were complaining about Feweh, but I don't really care about Feweh so I replaced him with another member of our community that is still more notable than Rotciv. It really doesn't matter to me if people that get referenced were "Good", I like Adam and Zero being referenced because they're still people that we talk about today, and they were somewhat unique. We could get at least 10 more rotcivs if we just invited some TG players and promised we wouldn't give them anything more than a few slaps on the wrist for acting like shitters, that way they can propagate how dumb they are and become somewhat notable.

patpol4 (changes requested):
Was going to close but this but im willing to go with Featherfields suggestion. Replace Rotciv with Colton, add a dented helmet with some stupid description. And remove Anvil, we don't add staff for merely existing.

LoliconSlayer:
IDK Wej, changing some names and adding a whole new item are two entirely different skill sets. And what problem do you have with Anvil? He's very popular among  the community.

Also I know it doesn't matter but the upvotes do outweigh the downvotes, and something like this should be a community decision in my humble opinion but I'm not a maintainer that does whatever they want.

patpol4:
You could also just replace feweh with colton.
The problem i have with adding Anvil is not that i have anything against him. It's just that if we forced every beloved community member into there, the ruin would become full of wizards in no time.
Regarding the votes, I'm mostly looking at the downvote amount, not the ratio. People usually upvote pr's. But they downvote only if they really have something against it.

LoliconSlayer:
I could, I could replace Feweh with Colton. But I didn't make this PR with the sole purpose of adding Colton, I did it to wipe away Rotciv. If you don't want to add Anvil, he's not the only person that I can add if you have such a misgiving about it.

patpol4:
I just don't want Rotciv completely removed and unnecessary characters being forced in.
If you don't know how to map the helmet in, just close this and I'll do it. Will credit you, replace rotciv with colton, and do the helmet thing. Otherwise this is a final no from me.

LoliconSlayer:
Fine.

PR #9483, opened 2020-08-16
Fixes a theoretically exploit that would theoretically allow to make an infinite amount of research points

alexkar598:
The passive effect enable/disable increments/decrements based on if a client is there when its done, if theres a client when its enabled, you get 1, but if theres no client when its disabled, it only removes 0.25 allowing to repeat the process to get an infinite amount of points(negative and positive)

I would blame bood for being a shitcoder but this is tg code..


SomeguyManperson (approved):
probably makes people who braindead after being nanited still produce 1 nanite but not a real issue
actually that's still an issue with the original stuff e

PR #9484, opened 2020-08-16
Alex goes over already merged code and fixes it post merge: Preternis language fix

alexkar598:
Languages are applied by the species_language_holder, there was no need to grant the language via the on_gain() because its already applied, blame hopek for not understanding how his prior fix even worked


Hopekz:
You do realize according to bitblame you put the C.grant_language code in there in the first place; right?

Only assumed that C.grant_language being in on_gain had some weird advantage because you put it in otherwise I had other language holder PR's in the past so I understand the concept.

alexkar598:
Preternis are older than species_language_holder, on_gain was the only option back then

SomeguyManperson (approved):

PR #9494, opened 2020-08-17
makes it harder to accidentally beat someone to death with a toilet while bullying them

SomeguyManperson:
:cl:  
tweak: you now need to be on harm intent to lid slam someone while giving them a swirlie 
/:cl:


JamieD1 (approved):
What have we come to..

PR #9502, opened 2020-08-18
Revert "Atmos 1: Stimball gone" now that monstermos makes it balanced

ghost:
I asked nich to remove old stimball because it could be used for powerful radball shotguns that instakilled under old atmos. With Monstermos, that possibility is removed, thus balancing out the reaction and making it safe for the general public again.

🆑 
rscadd: readds stimball reaction
/🆑


ghost:
Okay what the fuck I'll have to manually readd this I guess

PR #9503, opened 2020-08-18
Disallows use of most darkspawn abilities while using crawling shadows

SomeguyManperson:
:cl:    
tweak: darkspawn in crawling shadows are not able to use most abilities
/:cl:


Hopekz:
I think this is an exploit.
Doesn't seem intended.

JamieD1 (approved):

PR #9504, opened 2020-08-18
bag of holding can no longer hold bulky or larger items

SomeguyManperson:
The bag of holding makes adding weight to an item completely worthless since with a bag of holding you can take like 10 of whatever the huge item is i.e. full sized shotgun and not have to worry about how much of a pain in the ass it'd be to move it otherwise. Each bag at the base price only costs a diamond, two gold, a small amount of uranium and a single bluespace crystal (which is decreased with parts the lathe has been upgraded with by then), meaning basically anyone can get one.
it's either this or making them super ass to get if we don't want to pretend being able to run around with a hidden shotgun or ten is ok

:cl:  
tweak: bluespace bags are unable to hold items of the bulky or gigantic size
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -57,8 +57,6 @@
 /obj/item/storage/backpack/holding/ComponentInitialize()
 	. = ..()
 	var/datum/component/storage/STR = GetComponent(/datum/component/storage)
-	STR.allow_big_nesting = TRUE
-	STR.max_w_class = WEIGHT_CLASS_GIGANTIC
 	STR.max_combined_w_class = 35
 
 /obj/item/storage/backpack/holding/suicide_act(mob/living/user)


PettyOfficer:
Instead of removing bulky or large items from bag of holdings completely, would it not be fair to limit to one or so of such item? For example, putting a hardsuit inside.

boodaliboo:
pettys suggestion

Moja10:
Super bulky size class whence. That's my preferred solution

Hopekz:
You're right but that's the whole point of the bag of holding.
This change will make it into a regular bag.

If you're against the concept I think you're looking at removing the BoH
Would be for the removal of BoH instead as it is an actual issue per your description but not for keeping it in a dumb state.

maxion12345:
What
That's literally why you get a BoH, so you can lug around big shit without needing to take up a hand slot to carry or your armor slot. This just makes a duffle bag that isn't slow. And you can already hide shotguns in normal bags via shortening. This is a solution looking for a problem.

SomeguyManperson:
You can carry a stunprod in a bag of holding and the main downside for the stunprod is its size forcing it to be visible, also you can only shorten non-combat shotguns and it causes their projectile deviance to increase,  if you get a bag of holding you dont have to care about shortening a shotgun and dealing with the higher spread to conceal it. The major downside of the lavaland katana is its size and that also can be stuffed in a bluespace bag, syndicate hardsuits are specifically smaller to make them possible to conceal, bluespace bag means you can do that with a normal one. As long as the bluespace bag exists in its current making an item "big" won't matter since it can regardless be stuffed into a bluespace bag to negate the point entirely.

ynot01:
we could do what tg did

patpol4:
> we could do what tg did

What did tg do

ghost:
there is literally no reason to do this tbh

if you're dissatisfied with the ease by which you can get them gate them behind another research node or something

tg made them need anomaly cores, so they never appear, and making them not carry bulky items just makes them shit overall

if you want to nerf bags of holding *properly* just, you know, make them take time to pull something out of your impossibly deep, impossibly wide bag. a time delay to removing items from one. maybe just bulky items.

PR #9511, opened 2020-08-19
Adds the meat hook as blood-drunk miner loot

MrHorizons:
Since the meat hook is in a weird spot between being a bit too powerful to have in necropolis chests but not enough to have in other megafauna's loot, I got a suggestion to add it to the blood-drunk miner's loot. Which, considering the only other loot from the blood-drunk are a cleaving saw and a _kinetic accelerator_ (apparently, without any special modifications), seems reasonable.

#### Changelog

:cl:  
rscadd: Adds the meat hook to the blood-drunk miner's loot.
/:cl:


Hopekz:
Would be nice to see this as a Chef uplink item tbh.

Reed0506:
just make it a drop from legion then it will actually exist and also encourage people to do something that is never done at all because it isnt worth it. blood drunk is the easiest megafauna alongside heiro and also already has nice loot that fits it.

blood drunk drops its stuff and it most certainly does not use a meat hook.

MrHorizons:
Reed made some good points, so I'm gonna change it to his suggestion. Although the blood-drunk miner's loot still could use a bit of a buff, that'll be a different PR.

PR #8001, opened 2020-03-21
Medical sprays and patches now look for exposed flesh or thin material

Hopekz:
Finally got around to fix this long-running exploit/oversight.

Makes Medical sprays and patches look for exposed flesh or thin material.
If you're in a hard suit in space you cannot magically heal yourself with a spray (when no atmos can even reach you) and you also cannot rub your thick armor and magically be healed under the armor.

#### Changelog

:cl:  
bugfix: Medical sprays and patches now look for exposed flesh or thin material.
/:cl:

PR #9516, opened 2020-08-20
Revert "Medical sprays and patches now look for exposed flesh or thin material"

SomeguyManperson:
Reverts yogstation13/Yogstation#8001
Hopek has said he's not planning on making these function properly with pierce immunity gained from things that aren't thick material because "that's intended"

### DIFF
--- a/code/modules/reagents/reagent_containers/medspray.dm
+++ b/code/modules/reagents/reagent_containers/medspray.dm
@@ -32,14 +32,11 @@
 		amount_per_transfer_from_this = initial(amount_per_transfer_from_this)
 	to_chat(user, "<span class='notice'>You will now apply the medspray's contents in [squirt_mode ? "short bursts":"extended sprays"]. You'll now use [amount_per_transfer_from_this] units per use.</span>")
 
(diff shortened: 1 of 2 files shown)


LoliconSlayer:
Speedmerge

MegaEmpirical:
Not a fan.

Reed0506:
this change was bad anyways because it removed the only reason to ever use patches and sprays while simultaneously breaking their functionality in a few specific cases such as the free golems are not able to USE THEIR MEDS.

MegaEmpirical:
Slanderous lies. 

Patches and sprays still have uses: plasmamen. Other situations where removing a helmet/mask would not be favorable. Chems that are toxic or ineffective when ingested. Possibly dead bodies too, not sure if you can feed a dead body.

If no one is going to change golems then just give them pills instead.

Hopekz:
local powergamer wants to revert the change the prevents him from healing in space through his hardsuit while pretending its because of free golems instead of adding a quick exception for his golem problem instead!

More news once it's merged
image

JamieD1:
Fix golems instead. This is not getting reverted. Its overpowered for hard suit wearers.

JamieD1 (changes requested):
See above

PR #9540, opened 2020-08-22
revive rune takes 7 sacrifices up from 3

SomeguyManperson:
cult already can use basically any body they get living or dead and sacrificing stuff is already a net gain this makes it slightly less of a net gain since they can use constructs/summon ghosts/converting more people to easily make up for losses

:cl:  
tweak: revive rune requires 7 sacrifices up from 3
/:cl:


Tipy1802:
Eh seems a bit too much, like when will you find 7 bodies on the ground even more so when the pop is 40. It may end up encouraging murderboning

SomeguyManperson:
bloodcults already murderbone

PR #9541, opened 2020-08-22
bloodcult teleport spell costs 50 health up from 7 if used to teleport yourself

SomeguyManperson:
solidifies it as an escape spell by increasing the damage caused when it's cast so downtime is slightly increased

:cl:  
tweak: bloodcult teleport spell will cripple the arm it's used on if used to self teleport
/:cl:

### DIFF
--- a/code/modules/antagonists/cult/blood_magic.dm
+++ b/code/modules/antagonists/cult/blood_magic.dm
@@ -149,7 +149,7 @@
 	desc = "Empowers your hand to teleport yourself or another cultist to a teleport rune on contact."
 	button_icon_state = "tele"
 	magic_path = "/obj/item/melee/blood_magic/teleport"
-	health_cost = 7
+	health_cost = 50
 
 /datum/action/innate/cult/blood_spell/emp
 	name = "Electromagnetic Pulse"
@@ -494,6 +494,8 @@
 			to_chat(user, "<span class='warning'>The target rune is blocked. You cannot teleport there.</span>")
 			return
 		uses--
+		if(!(target == user))
+			health_cost = 7
 		var/turf/origin = get_turf(user)
 		var/mob/living/L = target
 		if(do_teleport(L, dest, channel = TELEPORT_CHANNEL_CULT))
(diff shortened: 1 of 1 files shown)


Tipy1802:
Sounds very interesting

ToGWtF:
50 seems high

Marmio64:
crippling your arm seems unnessecary. Perhaps instead just have it so that teleporting will deal the 7 damage and also paralyze the person for 5 seconds on arrival at the teleport rune. Negative for now.

edit: neutral instead

Darkstick:
I like the idea. It will make people (by that I mean me who's constantly spamming that spell) reconsider before zipping away the very second someone with a disabler appears in their line of sight.

SomeguyManperson:
deals roughly 37.5 damage to max health due to it hitting a limb so you don't lose half your health outright to it, probably going to make sure it doesn't crit the person on using it

code/modules/antagonists/cult/blood_magic.dm
```diff
@@ -146,10 +146,10 @@
 
 /datum/action/innate/cult/blood_spell/teleport
 	name = "Teleport"
-	desc = "Empowers your hand to teleport yourself or another cultist to a teleport rune on contact."
+	desc = "Empowers your hand to teleport yourself or another cultist to a teleport rune on contact. This will deal heavy damage to your arm if you use it to teleport yourself."
 	button_icon_state = "tele"
 	magic_path = "/obj/item/melee/blood_magic/teleport"
-	health_cost = 7
+	health_cost = 50
```
JamieD1, line 152:
```suggestion
	health_cost = 40
```

code/modules/antagonists/cult/blood_magic.dm
```diff
@@ -494,13 +494,18 @@
 			to_chat(user, "<span class='warning'>The target rune is blocked. You cannot teleport there.</span>")
 			return
 		uses--
+		if(!(target == user))
+			health_cost = 7
+		else if(user.health < 38) //at this point 50 damage to an arm will crit
```
JamieD1, line 499:
```suggestion
		else if(user.health < 30) //at this point 50 damage to an arm will crit
```

JamieD1 (changes requested):
Lets not go overboard, even if they use it and they are bleeding or any sort of suffocating or toxin they will go into crit anyway, lets give them a little buffer.

Moja10:
I think this is still worth looking into, perhaps could be testmerged and a couple sample rounds ran.

PR #9543, opened 2020-08-22
cult stun hand stuns less powerfully as the cult becomes stronger

SomeguyManperson:
effects:
cult below red eyes/halos: no change
cult at red eyes: effects are halved
cult at halos: throws target 5 tiles away, knocks them down for 10 seconds, and deals 50 stamina damage

Should make the stun less useful in open combat lategame so it's not an easy win in melee combat as if you don't have a stunbaton already ANYWAYS

:cl:  
tweak: cult stun hand stuns weaker as the cult increases in power, with effects being halved on entering risen (red eyes) and dealing heavy stamina damage and knockdown while throwing the target when in ascendent (halos)
/:cl:

### DIFF
--- a/code/modules/antagonists/cult/blood_magic.dm
+++ b/code/modules/antagonists/cult/blood_magic.dm
@@ -417,6 +417,8 @@
 	var/mob/living/L = target
 	if(iscultist(target))
 		return
+	var/datum/antagonist/cult/C = user.mind.has_antag_datum(datum/antagonist/cult)
+	var/datum/team/cult/cult = C.get_team
 	if(iscultist(user))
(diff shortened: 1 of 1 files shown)


ynot01:
late game they all have eswords and eshields anyways

ghost:
Throw is a bad idea, honestly. Just make it do stamina damage alone at that point.

Marmio64:
i think that the red eyes negative effect shouldn't be something that exists, since cult can still be struggling at that point.
Also the halos effect is still pretty crazy, should instead have it deal 60 stamina damage and knockdown for 3 seconds IMO.

edit: negative for now because halo stun hand effect is wonky and redeyes nerf is unnessecary.

Moja10:
Throwback sounds fun

Amelia0010 (approved):

JamieD1 (approved):

PR #9544, opened 2020-08-22
cult bolas knock down instead of paralyze

SomeguyManperson:
these are free and are basically a short ranged stun

:cl:  
tweak: cult bolas knock down instead of paralyze, effect increased from 2 seconds to 4
/:cl:


ghost:
Don't increase the effect, tbh. There's no reason they should be reinforced bolas to start with.

Amelia0010 (approved):

alexkar598 (approved):

PR #9566, opened 2020-08-24
Stuns and Stamcrit Now Radio Mute

DiamondSentinel:

Causes stuns and stamcrit to prevent you from using radios. This is to give stealthier antags a reason to not just immediately crit their targets, instead letting them use stunbatons or headbatons to deal with it without having them worry about their target screaming bloody murder over comms. Overall should lead to a healthier interaction between antags that don't need to murderbone. It's also already a thing for clock cult and blood cult, whom have better options to heal their targets after they've been converted, so I think it'll be a healthy change. 

Also keeps grayshits and clowns from screaming human harm while sec or a head is "voluntarily repositioning" them. 

Note that this doesn't prevent them from speaking loudly, unlike softcrit, so you can still yell for help, including using bounced radios _or_ radios that have their volume always on (which is a feature a lot of folks don't realize exists). Gives a reason to use radios besides when the AI decides that they've had enough of your telecomms bullshit.

Possible further changes also prevent radio use while cuffed for similar reasoning, but we'll see how this one is received first, and that one might be unnecessary as you can just remove it while they're stunned. 

#### Changelog

:cl:  
rscadd: Added a blacklist for say modes for stuns and stamcrit which currently only includes Radio
/:cl:


boodaliboo:
no

ynot01:
certainly an interesting change

redmoogle:
I like this, but will probably be a rough change. Does this include if the radio has been set to be always on?

ghost:
This is a hilariously bad idea. What's the point of any muting item when stunbatons mute AND stun you?

This will have no behavioral change with antagonists at all, they'll just kill their targets with greater ease since they can't call for help. No difference in how they'll behave.

DiamondSentinel:
> 
> 
> I like this, but will probably be a rough change. Does this include if the radio has been set to be always on?

As I said, this specifically does _not_ include that. In fact, it's one of the cited ways to deal with it.

DiamondSentinel:
Also, @AdamElTablawy what mute items are you specifically talking about? You've got mute toxins which are only really available to chemists, you've got mute sting from changeling which will still be used given that it takes 2 hits with a baton to stamcrit someone under the changes, you've got cult stuns which they're gonna use regardless, and you've got EMPs are only really a thing for changelings or traitors/nukies and will still see use because they're useful for way more than just closing comms. 

These options are all generally pretty bad, will still have a use under this change, or are taken anyways because of a lot of uses.

maxion12345:
I think this will have an opposite effect and make antags kill even more. If every stun is a radio mute, then every antag can get away with murder far easier. It'll become a far more appealing option then usual, especially because currently to stealthily murder someone you'll need to plan it out to ensure they either can't identify you while your stabbing them to death in maint or accept that your going loud and your objectives just got harder to complete with everyone knowing your not a good guy. This change throws all that out of the window. Why carefully plan your traitor load out to ensure you won't be caught assassinating your target when you can smash beepsky, take his baton, clap a cell in it and harmbaton your target to death without them being able to call for help on the radio?

Moggram:
its called a Radio Jammer, use it.

UselessTheremin:
In my opinion, this is not my favorite change. there are already EMP items, silencing items, etc. It would make most of those obsolete if all you needed was a stun. It would probably start a chain of events with nerfing stun items if this is merged. Maybe try adding a ghetto silencer for blood brothers and revs? Because making all stunbatons a combo is way too powerful, and they are incredibly strong as is. Though, I would be fine with just stamcrit muting but stun muting seems like a bit much.

ghost:
In practice, all giving an antagonist the ability to kill silently will do is make them kill more, not less.

There's a reason people blow up telecomms before murderboning.

Hopekz:
I don't think our code base supports a change like this because we have instant one-click stuns.

This would be a good change for any place where you can't just rush somebody with a one-hit stun out of nowhere without a warning.

DiamondSentinel:
As I've said, Hopekz (multiple times now), this is for the stun rework where you can't just one-shot stun someone out of nowhere. The rework that was testmerged made it so that batons did stam damage rather than an outright stun. While the amount of time stamcrit lasted was whack, I expect that'll be changed soon too. Regardless, this is to go hand-in-hand with that rework.

Hopekz:
> As I've said, Hopekz (multiple times now), this is for the stun rework where you can't just one-shot stun someone out of nowhere. 

Wut. 
Why are you PR'ing a change to something that wasn't merged yet?
You can PR to their repo instead of doing this weird "PR'ing in response to something that doesn't exist in the codebase" thing.
Almost like PR'ing a change to cobby chem without cobby chem being here.
Ignoring the code and the change the idea of this PR existing in this way is dumb

DiamondSentinel:
This was coded while the test was live and it seems like it'll be full-merged in a short bit. And then why bother sending it to someone else when it'd have to pass through here anyways.

ghost:
Because if they merge it to their PR it'll be merged at the same time. 

This change is bad. Stealth antag is an art, this just makes it far too easy.

DiamondSentinel:
"Killing someone before they can scream for help" isn't so much an art as getting the most powerful weapon you can find and clicking _really really fast_

code/modules/mob/living/say.dm
```diff
@@ -97,6 +98,7 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 	var/message_mode = get_message_mode(message)
 	var/original_message = message
 	var/in_critical = InCritical()
+	var/is_paralyzed = IsParalyzed()
```
Amelia0010: remove this line

code/modules/mob/living/say.dm
```diff
@@ -126,6 +128,9 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 	else if(stat == UNCONSCIOUS)
 		if(!(unconscious_allowed_modes[message_mode]))
 			return
+	else if(is_paralyzed)
```
Amelia0010:
```suggestion
	else if(IsParalyzed())
```

Amelia0010 (changes requested):

code/modules/mob/living/say.dm
DiamondSentinel: Fixed that, and also changed the syntax that I had used from the line above to reflect the same

Amelia0010 (approved):

JamieD1:
Lets give it a go, hopefully it gives more variety to situations.

JamieD1 (approved):

PR #9575, opened 2020-08-25
Asclepius' snake injects godblood

Mqiib:
:cl:  Mqiib

tweak: Asclepius' snake now injects godblood on bite instead of potentially useless (or even dangerous) medicines

/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -407,8 +407,7 @@
 		else
 			owner.visible_message("[owner]'s soul is absorbed into the rod, relieving the previous snake of its duty.")
 			var/mob/living/simple_animal/hostile/retaliate/poison/snake/healSnake = new(owner.loc)
-			var/list/chems = list(/datum/reagent/medicine/bicaridine, /datum/reagent/medicine/perfluorodecalin, /datum/reagent/medicine/kelotane, /datum/reagent/medicine/antitoxin)
-			healSnake.poison_type = pick(chems)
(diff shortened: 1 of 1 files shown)


code/datums/status_effects/buffs.dm
```diff
@@ -407,8 +407,7 @@
 		else
 			owner.visible_message("[owner]'s soul is absorbed into the rod, relieving the previous snake of its duty.")
 			var/mob/living/simple_animal/hostile/retaliate/poison/snake/healSnake = new(owner.loc)
-			var/list/chems = list(/datum/reagent/medicine/bicaridine, /datum/reagent/medicine/perfluorodecalin, /datum/reagent/medicine/kelotane, /datum/reagent/medicine/antitoxin)
-			healSnake.poison_type = pick(chems)
+			healSnake.poison_type = /datum/reagent/medicine/omnizine/godblood
 			healSnake.name = "Asclepius's Snake"
 			healSnake.real_name = "Asclepius's Snake"
 			healSnake.desc = "A mystical snake previously trapped upon the Rod of Asclepius, now freed of its burden. Unlike the average snake, its bites contain chemicals with minor healing properties."
```
jcatjmeow, line 413: change the desc to match

jcatjmeow (changes requested):

code/datums/status_effects/buffs.dm
Mqiib: I dunno, the description still works as is. Changing it to "a chemical with minor healing properties" doesn't flow as well as just "chemicals"

SomeguyManperson (approved):

PR #9576, opened 2020-08-26
Removes the jukebox "rest" dancemove and fixes the access.

maxion12345:
so apparently it used engineering access, it is now changed to bartender access so bartenders can set up their own disco bars.

In addition, removed the dance move that made you rest. Shit was annoying to be walking past bar, go into the range of the dance machine, and slow to a damn crawl until you stop dancing. Eh.

:cl:  
rscdel: Deleted an annoying dance move.
tweak: Discoball now uses bartender access.
/:cl:


alexkar598 (changes requested):
Do something so it's not dance3() then dance5()

alexkar598 (approved):

PR #9580, opened 2020-08-26
Removes the unless it conflicts with the first law part of asimov as its already the default

alexkar598:
Doesn't really change anything but would cause less confusion as to other lawsets that dont have the Unless it conflicts with X law


ynot01:
a fine change but it creates scenarios that are interesting to interpret on ion anomalies and i think that's a good part of the game

jcatjmeow:
"Lets remove the Issac Asimov reference from the Asimov lawset"

PR #9582, opened 2020-08-27
Temporairly disables mousemove event to see if it helps with lag, will break wizard spellcards and space pods

alexkar598:
Temporairly disables mousemove event to see if it helps with lag, will break wizard spellcards and space pods

test merge

### DIFF
--- a/code/_onclick/drag_drop.dm
+++ b/code/_onclick/drag_drop.dm
@@ -102,6 +102,7 @@
 /obj/screen/click_catcher/IsAutoclickable()
 	. = 1
 
+/*
 //Please don't roast me too hard
 /client/MouseMove(object,location,control,params)
 	mouseParams = params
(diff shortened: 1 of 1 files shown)


Darkstick:
mousemove event?

Nechronom:
I think it's a means of aiming something towards the mouse? In the case of space pods it's the direction that the front is facing. I've never gotten to play wizard so I don't know what spell cards they mean

DiamondSentinel:
> 
> 
> I think it's a means of aiming something towards the mouse? In the case of space pods it's the direction that the front is facing. I've never gotten to play wizard so I don't know what spell cards they mean

Spell cards is a spell that casts multiple bursts of damaging cards that follow the mouse cursor over time. Lets you spam it ridiculously, and is quite fun, if not for the fact that it's absolutely awful.

jcatjmeow:
> mousemove event?

it means the game keeps track of when the mouse is moved and where it is.

This, as you might imagine, comes at a cost of performance

Amelia0010:
@alexkar598 what's the status of this?

alexkar598:
It helps with lag, the spell cards can AFAIK work without this but the pod needs something on process() to get the mouse location and use it instead of listening for mouse move

JamieD1:
Can we fix the Pod Issues?

PR #9600, opened 2020-08-31
Polysmorph Race

missatessatessy:
Finally an official PR for the Polysmorph race, they technically are unfinished only missing the abilities.
I can add those later when I decide I hate myself enough to do them. I feel like getting this PR up and going is a good idea to see whether I should bother to finish coding them xD

These are a new race introduced by War/Tex here: [https://forums.yogstation.net/index.php?threads/feedback-needed-on-new-race.21787/](https://forums.yogstation.net/index.php?threads/feedback-needed-on-new-race.21787/)

They aren't a round start race at the moment, and as such the only way to become them is from the "changesource_flags" but don't ask me what that means or whether that's accurate.

:cl:  
rscadd: Added Polysmorph race
/:cl:


Tipy1802:
Glad to see a new round start race and one with interesting abilities and concept

Reed0506:
i dunno what gave you the impression this is a roundstart race because it specifically states that it isnt

Partheo (approved):

JamieD1 (approved):
Alien tits here we come

PR #9602, opened 2020-08-31
Lowers firedoor integrity

ynot01:

300 integrity fo' a firedoo'
pre'y schtewpid innit

#### Changelog

:cl:  
tweak: Lowered firedoor integrity
/:cl:

### DIFF
--- a/code/game/machinery/doors/firedoor.dm
+++ b/code/game/machinery/doors/firedoor.dm
@@ -11,7 +11,7 @@
 	icon_state = "door_open"
 	opacity = FALSE
 	density = FALSE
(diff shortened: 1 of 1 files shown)


Nechronom:
That just means it's easier to beat up if this gets past, right? .. does that effect it's ability to resist atmospherics damage much?

alexkar598 (approved):

ynot01:
> That just means it's easier to beat up if this gets past, right? .. does that effect it's ability to resist atmospherics damage much?

Firedoors are fireproof so nothing in atmos typically besides fusion balls will break them

PR #9607, opened 2020-09-01
Sleeping Carp Rebalance

ynot01:

Projectile deflection only happens when throw mode is on
In return, price is reduced to 14TC (price up for discussion)
Projectile deflection now needs to be a conscious action
Security will no longer hit you with a stray disabler and valid you
Cannot use items or fists (besides in hand or throwing them) while deflecting

#### Changelog

:cl:  
tweak: Sleeping Carp now only deflects with throw mode on in return for a price reduction
/:cl:

### DIFF
--- a/code/datums/martial/sleeping_carp.dm
+++ b/code/datums/martial/sleeping_carp.dm
@@ -7,7 +7,7 @@
 /datum/martial_art/the_sleeping_carp
 	name = "The Sleeping Carp"
 	id = MARTIALART_SLEEPINGCARP
-	deflection_chance = 100
+	deflection_chance = 0 //no block unless throwmode is on
 	reroute_deflection = TRUE
 	no_guns = TRUE
 	allow_temp_override = FALSE
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -47,8 +47,8 @@
 			return spec_return
 
 	if(mind)
-		if(mind.martial_art && !incapacitated(FALSE, TRUE) && mind.martial_art.can_use(src) && mind.martial_art.deflection_chance) //Some martial arts users can deflect projectiles!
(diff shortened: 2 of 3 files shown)


redmoogle:
Hmm does it turn off throwmode afterwards, because that would just be the same effect as the original but slightly inconvenient

ynot01:
> Hmm does it turn off throwmode afterwards, because that would just be the same effect as the original but slightly inconvenient

the point is that you can't do anything but move while deflecting

boodaliboo:
would rather have it be 17tc and deflect like it does currently

ynot01:
> would rather have it be 17tc and deflect like it does currently

17TC leaves no room for other items, no room for gimmicks, items should be cheaper but weaker so that you can shape your round in a unique way
i hate the state of sleeping carp currently. it's boring, it's expensive, and it gets fucked by a stun baton clicking race
expensive does not justify OP
16TC is probably my limit for how expensive a single item should be

boodaliboo:
> > would rather have it be 17tc and deflect like it does currently
> 
> 17TC leaves no room for other items, no room for gimmicks, items should be cheaper but weaker so that you can shape your round in a unique way
> i hate the state of sleeping carp currently. it's boring, it's expensive, and it gets fucked by a stun baton clicking race
> expensive does not justify OP
> 16TC is probably my limit for how expensive a single item should be

you convinced me

Marmio64:
this is actually a buff, it lowers the TC cost allowing you to also buy a emag, and it also makes it so you can toggle off projectile deflection, so you don't get cucked because security is chasing you and trying to disabler you because you broke a window, now you can hide your sleeping carp.

ynot01:
>so you don't get cucked because security is chasing you and trying to disabler you because you broke a window, now you can hide your sleeping carp.

this is pretty bad design tbh and im glad its gone
it makes sense that you dont auto deflect when you dont mean to

ynot01:
tested, working

ynot01:
to andwer earlier, no it doesnt turn off throwmode when you deflect something

Amelia0010 (approved):

PR #9620, opened 2020-09-02
Fix mushroom hallucinogen

SomeguyManperson:
:cl:  
bugfix: mushroom hallucinogen now actually causes drug effects
/:cl:


JamieD1 (approved):
Muppet

PR #9623, opened 2020-09-02
Changes pretty message blocking to using the helper proc

Amelia0010:
Proc was made specifically so we wouldn't have to do the logic in other procs but could just call the helper

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -8,7 +8,7 @@
 	//yogs start - pretty filter
 	if(isnotpretty(message))
 		to_chat(usr, "<span class='notice'>You fumble over your words. <a href='https://forums.yogstation.net/index.php?pages/rules/'>See rule 0.1.1</a>.</span>")
-		message_admins("[key_name(usr)] just tripped a pretty filter: '[oldmsg]'.")
+		message_admins("[key_name(usr)] just tripped a pretty filter: '[message]'.")
 		return
 	if(isliving(src))
 		message = minor_filter(to_utf8(message)) //yogs end - pretty filter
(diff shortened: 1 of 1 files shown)


code/modules/mob/say.dm
```diff
@@ -25,9 +24,8 @@
 	set name = "Whisper"
 	set category = "IC"
 
-	var/oldmsg = message //yogs start - pretty filter
-	message = pretty_filter(message)
-	if(oldmsg != message)
+	//yogs start - pretty filter
+	if(isnotpretty(message))
 		to_chat(usr, "<span class='notice'>You fumble over your words. <a href='https://forums.yogstation.net/index.php?pages/rules/'>See rule 0.1.1</a>.</span>")
 		message_admins("[key_name(usr)] just tripped a pretty filter: '[oldmsg]'.")
```
jcatjmeow:
```suggestion
		message_admins("[key_name(usr)] just tripped a pretty filter: '[message]'.")
```

code/modules/mob/say.dm
```diff
@@ -5,9 +5,8 @@
 	set name = "Say"
 	set category = "IC"
 
-	var/oldmsg = message //yogs start - pretty filter
-	message = pretty_filter(message)
-	if(oldmsg != message)
+	//yogs start - pretty filter
+	if(isnotpretty(message))
 		to_chat(usr, "<span class='notice'>You fumble over your words. <a href='https://forums.yogstation.net/index.php?pages/rules/'>See rule 0.1.1</a>.</span>")
 		message_admins("[key_name(usr)] just tripped a pretty filter: '[oldmsg]'.")
```
jcatjmeow:
```suggestion
		message_admins("[key_name(usr)] just tripped a pretty filter: '[message]'.")
```

jcatjmeow (changes requested):

JamieD1 (approved):

PR #9636, opened 2020-09-03
makes it no longer possible to break into test chamber before round end

boodaliboo:
all reinforced windows are now admin windows
#### Changelog

:cl:  
rscadd: Added reinforced windows 
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom.dmm
+++ b/_maps/map_files/generic/CentCom.dmm
@@ -11427,6 +11427,10 @@
 /obj/item/seeds/cherry/bomb,
 /turf/open/floor/wood,
 /area/centcom/holding)
+"yN" = (
+/obj/effect/turf_decal/loading_area,
+/turf/open/floor/plasteel,
+/area)
 "yQ" = (
 /obj/structure/flora/ausbushes/lavendergrass,
 /obj/structure/flora/ausbushes/sparsegrass,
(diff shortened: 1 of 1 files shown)


UselessTheremin:
bad, let them come to the test chamber and let admins smite them. its really funny

boodaliboo:
> bad, let them come to the test chamber and let admins smite them. its really funny

no i need this for strategies

GenericDM:
If somebody makes it to the test chamber they can fuck with round-ending items. I've seen this happen before when I summoned an ERT and looked away for a few minutes while they were gearing up.
I now spawn admin walls to block off the rest of Central Command from them, but this should work too.

alao3alao3:
Wait wasn't CC on a z-level that's only appears during round ends (or due to admemery) and can't be reached during the round anyway?

boodaliboo:
no

patpol4 (changes requested):
Do these as well please
chrome_qqp4gVMTbI

Reed0506:
this seems like a change along the same lines as the ones that specifically changed the area around the ERT spawn to be all admin windows and the admin wall barrier around the ferry. all i can see this doing is limiting the things you can do at roundend. for ERTs, just blocking their area by default would solve that issue, seeing as thats already been mostly done anyways.

Darkstick:
I do enjoy escaping/breaking into the test room during round end to join up with friends or just go to the recovery shuttle. It would be a shame to loose that ability...

patpol4 (approved):
Darkstick has a good point, as it is right now this just changes the windows at the bottom of the test chamber and I'm fine with that.

PR #9639, opened 2020-09-04
altar of god on delta and meta

boodaliboo:
########################################
they should be on every map
#### Changelog

:cl:  
bugfix: added alter of god to delta and meta
/:cl:


patpol4 (approved):
Good

PR #9643, opened 2020-09-05
vitality matrix works on augmented limbs

SomeguyManperson:
:cl:  
bugfix: vitality matrix will now actually work on robot limbs instead of wasting vitality 
/:cl:

### DIFF
--- a/code/modules/mob/living/damage_procs.dm
+++ b/code/modules/mob/living/damage_procs.dm
@@ -28,12 +28,12 @@
 			adjustStaminaLoss(damage * hit_percent)
 	return 1
 
-/mob/living/proc/apply_damage_type(damage = 0, damagetype = BRUTE) //like apply damage except it always uses the damage procs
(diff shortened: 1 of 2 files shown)


redmoogle:
Good, Never understood why it did that

code/modules/mob/living/damage_procs.dm
```diff
@@ -28,12 +28,12 @@
 			adjustStaminaLoss(damage * hit_percent)
 	return 1
 
-/mob/living/proc/apply_damage_type(damage = 0, damagetype = BRUTE) //like apply damage except it always uses the damage procs
+/mob/living/proc/apply_damage_type(damage = 0, damagetype = BRUTE, _required_status) //like apply damage except it always uses the damage procs
```
redmoogle: Whats with _required_status it does absolutely nothing, and also breaks a few other things

SomeguyManperson: was being used to transfer required status to adjustbruteloss and adjustfireloss but it didnt work so I'll just use the boring way

JamieD1 (approved):

PR #9645, opened 2020-09-05
Removes Patch Wounds

redmoogle:
### THIS DOES NOT REMOVE TEND WOUNDS

Removes patch wounds as a surgery as it's just a tends wounds ripoff, also can be hard to differentiate between the two surgerys

#### Changelog

:cl:  
rscdel: Patch wounds is gone
/:cl:


MrHorizons:
:crab:

redmoogle:
> 🦀

🦀

Jarod1200:
Bad. They are not the same and are both useful in different situations. 
Patch wounds is superior if you have a less than ideal surgical setup because you don't have to fail the repeatable step 10 times, or if there is a ton of damage to 1 part of the body. 
Tend wounds is better if you won't be failing surgical steps and if they have damage distributed to their whole body

redmoogle:
> 
> 
> Bad. They are not the same and are both useful in different situations.
> Patch wounds is superior if you have a less than ideal surgical setup because you don't have to fail the repeatable step 10 times, or if there is a ton of damage to 1 part of the body.
> Tend wounds is better if you won't be failing surgical steps and if they have damage distributed to their whole body

I'm curious how much does patch wounds get
also tend wounds works fine although for failing steps. Although which the way the server is, it's super easy to get to close to/ideal scenarios.
Damage for the most part is going to be all over the body unless they got lasered down, but still tend wounds works
~~and tbh I'd rather see patch wounds if not removed, renamed to fix limb~~

Jarod1200:
I would be ok with changing its name to something a bit more clear on what it does. Maybe "patch body part" .
If they got lasered down you wouldn't be using either surgery because they do burn damage. Heads of staff who get toolboxed to death on a rev round would be a better example. As the revs probably all targeted the chest or head. And if you're an antag or being hunted by a group of antags you'll probably be doing surgery on a table in a secluded area. Doing ghetto tend wounds without splashing alcohol or sterilizer is practically a death sentence for your patient.

redmoogle:
Gonna close this for now until I get some better medical concepts going to replace both/make each one better

Also I have too many projects going on ~~Help~~

PR #9647, opened 2020-09-05
Fixes a bug with the SM delam code

alexkar598:
One of the two loops was typecasted for no reason

Also it compares the z of the turf of the mob instead of their direct location (removes closet cheesing)

Removes a comment that simply isn't true


alexkar598:
Imagine downvoting a bugfix

patpol4:
Does this mean hiding in a closet to avoid hallucinations was a bug?

JamieD1 (approved):

PR #9649, opened 2020-09-07
Fixes for fireman Carry

### LINKED ISSUE #9632: offhand item isn't removed on fireman carry ending
## Reproduction:
end fireman carry without dropping the offhand item
notice how offhand item stays until you drop it

redmoogle:
This should fix #9632 

This _should_ also fix the layer-fuckery thats been going on, but that might be its own thing

I just pulled the most recent code from tgstation

#### Changelog

:cl:
bugfix: Layers should behave again
/:cl:

### DIFF
--- a/code/datums/components/riding.dm
+++ b/code/datums/components/riding.dm
@@ -215,6 +215,7 @@
 	RegisterSignal(parent, COMSIG_HUMAN_MELEE_UNARMED_ATTACK, .proc/on_host_unarmed_melee)
 
 /datum/component/riding/human/vehicle_mob_unbuckle(datum/source, mob/living/M, force = FALSE)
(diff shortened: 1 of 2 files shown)


UselessTheremin:
i have but one request: please check if fireman carrying is cancelled when going through a teleporter, and if the offhand is left after that. someone died because of that problem lol

redmoogle:
> 
> 
> i have but one request: please check if fireman carrying is cancelled when going through a teleporter, and if the offhand is left after that. someone died because of that problem lol

Fair point will test that as I didnt know that was the problem

redmoogle:
Yeah gets rid of the offhand atleast on my side

SomeguyManperson (approved):

PR #9655, opened 2020-09-08
holyparasite creator now deals the appropriate amount of damage

ynot01:

hehe cards deal 18damage
i blame theos

alt to #9661

#### Changelog

:cl:  
bugfix: Deck of holy tarot cards now deals the appropriate amount of damage
/:cl:

### DIFF
--- a/code/game/objects/items/holy_weapons.dm
+++ b/code/game/objects/items/holy_weapons.dm
@@ -750,6 +750,8 @@
 /obj/item/nullrod/holypara
 	name = "deck of holy tarot cards"
 	desc = "A holy deck of tarot cards, harboring a healing spirit."
+	force = 0 //cards deal 18 damage hehe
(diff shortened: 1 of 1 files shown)


UselessTheremin:
bad. ive done rounds where i haven't gotten a holyparasite and tried the whole round. it is more hehe null rod do 18 damage not hehe card do 18 damage. i would suggest to make it so if it is already used, it does 0 damage, but otherwise does 18 so it is like the normal nullrod.

Hopekz:
Ynot's PR is fixing an oversight but I'd also be fine to see us go with @UselessTheremin 's route by making it so that it only deals damage as long as you don't have a holypara summoned. He brings up a valid point of the item being useless if there is no-one to take the holypara.

ynot01:
nah

PR #9656, opened 2020-09-08
Alex fixes PRs post merge: Jamie you need to read edition

alexkar598:
Lacked the leading slash in the type path

Also the proc was duplicated...


TheLinuxOS:
based

redmoogle:
> 
> 
> based

and redpilled

JamieD1 (approved):

PR #9659, opened 2020-09-09
Revert "Revert "Stuns and Stamcrit Now Radio Mute""

SomeguyManperson:
on god this is why we have testmerges


Reed0506:
"hmm yes today we shall buff stun combat, the thing that everyone generally agrees is overpowered and not fun"

maxion12345:
this doesn't even mute your radio it just makes it so when you try to use a radio it literally won't go into the chat, even if your not muted

ghost:
what do you know it was a shit idea and everyone is bitching to me about it

alexkar598:
This just makes stun even more powerful then they already are

PR #6348, opened 2019-07-14
Removes blood brother bundles

AsV9:
### Intent of your Pull Request

This PR reverts #5310 and #5399
The traitor bundles have a value of over 20 TC, and giving a team of traitors >40 TC kinda defeats the point of the entire gamemode and having to cooperate and all that shit.

:cl:  
rscdel: Blood brothers no longer spawn with traitor bundles.
/:cl:

PR #9665, opened 2020-09-10
Readds Blood Brother Traitor Kits

Xoxeyos:
Readds Blood Brother traitor kits, I think we should try this out again, before we remove BB from our server.

:cl:  Xoxeyos
rscadd: Readds the Blood Brother Traitor Kits.
/:cl:


ynot01:
poor solution

AsV9:
The reason blood brothers is such a shit gamemode is because you are forced to rely on RNG to not get an incompetent or useless brother. Giving a pair of brothers 40 fuckin TC is a terrible solution to the problem which basically just makes it so you can ignore your brother 100% of the time instead of 90% of the time.

I ran the numbers on the bundles located [here](https://github.com/yogstation13/Yogstation/blob/master/code/game/objects/items/storage/uplink_kits.dm#L122). On average you're going to get about 30 TC worth of items, and all the bundles are fairly good, definitely not meme-tier. Imagine both brothers getting decent bundles and you now have somewhere between 40 and 120 TC to complete your objectives. Imagine three way blood brother teams who end up with so much fucking gear they are basically nukies. Yeah that's what this PR does.

I'd rather you just remove blood brothers completely than re-implement this.

Original PR: #5310
Removal PR: #6348
Council vote that approved removal: [Vote-merge-6348](https://forums.yogstation.net/index.php?threads/vote-merge-6348.19434/)

UselessTheremin:
Yeah this pretty much makes BB's lone operatives at this point like we should just remove them, they are beyond saving.

boodaliboo:
good solution

MrHorizons:
My solution would have been to give them Syndicate radio implants (or just normal Syndicate encryption keys), but I guess that's barely above removing them.

Xoxeyos:
I figured I'd try, I agree, Blood Brothers is absolutely retarded.

PR #9683, opened 2020-09-12
Makes it so malf AIs win if they complete their objectives

Amelia0010:
they technically don't need to blow the station up to succeed with their objectives during malf rounds, so it really shouldn't be a half-win

See: 
Screenshot_20200911-165843_Firefox_Nightly 1

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -637,7 +637,9 @@
 		if(SSshuttle.emergency.is_hijacked())
 			SSticker.news_report = SHUTTLE_HIJACK
 
-
+/**
+  * Given a list of minds, returns TRUE if they completed all their objectives, FALSE otherwise
+  */
 /datum/game_mode/proc/didAntagsWin(list/antags)
 	for(var/datum/mind/m in antags)
 		for(var/datum/antagonist/a in m)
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/game_mode.dm
```diff
@@ -637,6 +637,16 @@
 		if(SSshuttle.emergency.is_hijacked())
 			SSticker.news_report = SHUTTLE_HIJACK
 
+/**
+  * Given a list of minds, returns TRUE if they completed all their objectives, FALSE otherwise
+  */
+/datum/game_mode/proc/didAntagsWin(list/antags)
+	for(var/datum/mind/m in antags)
+		for(var/datum/antagonist/a in m)
+			for(var/datum/objective/objective in a.objectives)
```
alexkar598: All three loops dont need to be typed, make untyped loops and typecast it after
Example:
for(var/I in moblist)
   var/mob/M = I
instead of
for(var/mob/M in moblist)

code/game/gamemodes/game_mode.dm
```diff
@@ -637,6 +637,16 @@
 		if(SSshuttle.emergency.is_hijacked())
 			SSticker.news_report = SHUTTLE_HIJACK
 
+/**
+  * Given a list of minds, returns TRUE if they completed all their objectives, FALSE otherwise
+  */
+/datum/game_mode/proc/didAntagsWin(list/antags)
```
alexkar598: Can you add a second argument being an antagonist type and filtering the antag datums by that? So that if an admin gives the Ai a custom antag status of bringer or doom or whatever it doesnt interfere with the win state

alexkar598 (dismissed):

JamieD1 (approved):

PR #9684, opened 2020-09-12
Makes it so wizards can win or halfwin based on objective completion

Amelia0010:
See: https://cdn.discordapp.com/attachments/134720091576205312/754083649380679830/Screenshot_20200911-165843_Firefox_Nightly.jpg


alexkar598:
I dont think checking for objectives is a good idea, we dont enforce objectives on wizard and wizards dont follow their objectives. I doubt anyone would start following their objectives to see a light on my panel light up

code/game/gamemodes/wizard/wizard.dm
```diff
@@ -64,9 +64,17 @@
 
 /datum/game_mode/wizard/set_round_result()
 	..()
-	if(finished)
+	var/didEscape = TRUE
+	for(var/datum/mind/wizard in wizards)
+		if(!wizard.considered_escaped())
```
alexkar598: Proc doesnt exist

alexkar598 (dismissed):

code/game/gamemodes/wizard/wizard.dm
```diff
@@ -64,9 +64,17 @@
 
 /datum/game_mode/wizard/set_round_result()
 	..()
-	if(finished)
+	var/didEscape = TRUE
+	for(var/datum/mind/wizard in wizards)
+		if(considered_escaped(wizard))
+			didEscape = FALSE
```
jcatjmeow: You can cast your proc like so
```suggestion
	var/didEscape = TRUE
	var/datum/objective/O
	for(var/datum/mind/wizard in wizards)
		if(O.considered_escaped(wizard))
			didEscape = FALSE
```
This allows you to access /datum/objective/ procs even if you aren't in that file.

jcatjmeow (changes requested):

code/game/gamemodes/game_mode.dm
```diff
@@ -750,4 +750,4 @@
 	round_credits += "<br>"
 	round_credits += "<br>"
 
-	return round_credits
+	return round_credits er
```
jcatjmeow:
```suggestion
	return round_credits
```

jcatjmeow (changes requested):

SomeguyManperson:
this only checks for if the wizard escaped, and wizards can get survive until the end instead of escape objectives so they could score a loss even if they completed their objectives

JamieD1 (approved):

PR #9685, opened 2020-09-12
Gangs can now win or lose

Amelia0010:


yogstation/code/game/gamemodes/gangs/gangs.dm
```diff
@@ -81,3 +81,16 @@ GLOBAL_LIST_EMPTY(gangs)
 
 	round_credits += ..()
 	return round_credits
+
+/datum/game_mode/wizard/set_round_result()
```
alexkar598: Wizard you say?

code/game/gamemodes/game_mode.dm
```diff
@@ -637,6 +637,16 @@
 		if(SSshuttle.emergency.is_hijacked())
 			SSticker.news_report = SHUTTLE_HIJACK
 
+/**
+  * Given a list of minds, returns TRUE if they completed all their objectives, FALSE otherwise
+  */
+/datum/game_mode/proc/didAntagsWin(list/antags)
```
alexkar598: And this proc exists because....?

alexkar598 (changes requested):

code/game/gamemodes/game_mode.dm
Amelia0010: because I based this off of the same branch as the 2 other PRs which actually needed that, didn't realize it wasn't needed for this till I had done it

boodaliboo:
so how does this work? does it just mean if all gang heads die the round ends?

alexkar598 (approved):

PR #9687, opened 2020-09-12
Fixes the SM & remove bad code

redmoogle:

SM can explode again, and no. you are not safe in a locker, or even the BS locker.

#### Changelog

:cl:  
bugfix: SM explodes again
/:cl:


Reed0506:
why would the bluespace locker not be safe it just swaps to a new locker if the old one is destroyed

redmoogle:
I wasnt specific enough. I meant from hallucinations

SomeguyManperson (approved):

PR #9703, opened 2020-09-14
Fugu nerf

DiamondSentinel:
TL;DR Keeps beeegify while still nerfing it. It's mostly cosmetic, but it can smash structures now. No reinforced walls, HP increase, or damage increase though. Competing with Adam's PR because I like beeeeeg gimmicky stuff.


MrHorizons:
Hmm, yes, _l o r g e_.

ghost:
See, this is fine. The thing is, you could have just _told_ me to do this and i'd do it but instead you favor mindless downvote then cry in discord.

DiamondSentinel:
I didn't mindlessly downvote you. I didn't even view your PR other than the notification in discord. But the issue is that you _immediately_ jump to "remove mechanic" instead of "balance mechanic". That's why people get annoyed when you touch their departments.

jcatjmeow:
The nerf is better than nothing but i dont think this is balanced in any form.

Removal of bad things IS an improvement and not everything can be ""balanced"" some things do more harm than good in any and all iterations. This is why I downvoted.

If this was the only alternative I'd upvotw

jcatjmeow (approved):
Removal of rwalls is golden. Removal of melee damage is also good. I would also remove the smash structures but this is way better than leaving it as is, even if I do want it gone.

The health is honestly not the thing that makes it good and you could leave it at 1.5 tbh but whatevs

code/modules/mob/living/simple_animal/hostile/wumborian_fugu.dm
```diff
@@ -134,11 +134,8 @@
 			to_chat(user, "<span class='warning'>Something's interfering with [src]'s effects. It's no use.</span>")
 			return
 		A.buffed++
-		A.maxHealth *= 1.5
-		A.health = min(A.maxHealth,A.health*1.5)
-		A.melee_damage_lower = max((A.melee_damage_lower * 2), 10)
-		A.melee_damage_upper = max((A.melee_damage_upper * 2), 10)
+		A.maxHealth *= 1.2
```
redmoogle, line 137: > no hp increase

increasing HP

ghost:
@DiamondSentinel 

I don't always jump to remove mechanic. But if I can't figure out *how* to improve it, i'll remove it. And if something can't be improved it should be removed.

Regardless, this is good enough.

code/modules/mob/living/simple_animal/hostile/wumborian_fugu.dm
DiamondSentinel: There's no need to add HP. It just gives more reason for people like Adam and Jcat to scream about it being removed because they don't like that anyone can have fun doing anything other than killing people with toolboxes and mining loot.

fluffe9911 (approved):
ok

PR #9705, opened 2020-09-14
Fixes stupid medical airlock lights and cleans up dme

JamieD1:
Probably a better fix,

This works though

#### Changelog

:cl:  
bugfix: Fixes 200 overlays per door.
/:cl:


code/game/machinery/doors/airlock.dm
```diff
@@ -696,6 +696,7 @@
 		. = airlock_overlays[iconkey] = mutable_appearance(icon_file, icon_state)
 
 /obj/machinery/door/airlock/proc/check_unres() //unrestricted sides. This overlay indicates which directions the player can access even without an ID
+	cut_overlays()
```
alexkar598: Do the cut at the start of update_icons so that it doesnt fuck up overlays implemented in the future that get added before this proc is called

alexkar598 (changes requested):

alexkar598 (approved):

PR #9708, opened 2020-09-14
adds a small fire delay to both station shotguns to make them slightly less of the best weapon in the fucking game and ideally this is all i'll have to do because i don't want to even consider nerfing buckshot or some stupid shit like that

ghost:
7 fire delay for riot shottie

5 fire delay for combat shottie and compact combat shottie

for comparison the LMG has a 3 fire delay between bursts

this is because shotguns are unironically the most powerful gun available to the crew by far and outclass fucking everything, but rather than nerfing buckshot or slugs i think it'd just be better to make them slower-firing for their power making it still possible to take immense fucking chunks out of someone with one but not blast them with 3 rounds of burst damage buckshot instantly to kill them outright

no change to nukie shotguns, cycler shotguns, and the bartender's double-barrel shotgun which can still fire quite fast

### Changelog

:cl:  
tweak: gave riot and combat shotgun a short fire delay
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/shotgun.dm
+++ b/code/modules/projectiles/guns/ballistic/shotgun.dm
@@ -36,6 +36,7 @@
 	name = "riot shotgun"
 	desc = "A sturdy shotgun with a longer magazine and a fixed tactical stock designed for non-lethal riot control."
 	icon_state = "riotshotgun"
+	fire_delay = 7
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/riot
 	sawn_desc = "Come with me if you want to live."
 	can_be_sawn_off  = TRUE
@@ -60,6 +61,7 @@
 /obj/item/gun/ballistic/shotgun/automatic/combat
 	name = "combat shotgun"
 	desc = "A semi automatic shotgun with tactical furniture and a six-shell capacity underneath."
+	fire_delay = 5
 	icon_state = "cshotgun"
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/com
 	w_class = WEIGHT_CLASS_HUGE


ynot01:
what was their old fire delay

ghost:
none

ynot01:
testing done with 45% time dilation

combat shotgun with 0 fire delay

combat shotgun with 5 fire delay


riot shotgun with 0 fire delay

riot shotgun with 7 fire delay


the difference is only slightly noticeable, consider making the numbers higher?

ghost:
ideally this should tone things down a little

if it doesn't i'll up the fire delay again

Moggram:
Ya, the combat shotgun is broken as all hell. At 5 fire delay, it still seems too fast, but that's just me. (On SS13-CM its on like a two second fire delay if that helps at all.)

The riot shotgun with a 7 fire delay feels about right.

Hopekz:
Not sure if intentional but this "nerf" seems a bit meaningless.
IMO you should go harder with this change otherwise the fire delay currently in place doesn't seem like it would make much of a difference. 
**Who cares if you can take someone out in .5 or .7 of a second vs .4 of a second.** Make it something like 1.5 - 3 seconds.

You almost can't feel the 5-7 decisecond delays because that's normally how fast people fire in the first place.

maxion12345:
I also feel like having too high a delay would allow ops to tank shotgun shots and fire back with their 0 delay shotguns, giving them a massive advantage in even uneven battles at close range. Keep as is and see if it needs to go up

Reed0506:
if you get in the face of a nukeop you deserve to die generally

Georgeur:
how about a holopara user, which is basically the only thing that makes me a little anxious about extending the delay. 
and saying you deserve to die, for getting in the face of a nukeop, is not an argument to say it wont be unbalanced to have a high delay. granted that'd really only be an issue with riot shotguns, then again no one orders combat shotguns on nukeops rn

also I havent tried it get personally. but with this pr you might not be able to do the autistic technique in which you can spam z and click, and if you do both very quickly at the same time, you wont eject any unused shells. ive said before that a riot shotgun can fire about as fast as a combat shotgun. this'll most likely fix that brain dead thing

ghost:
@Georgeur 

If holoparasites are so strong that they require the use of an incredibly high-damage rapid-fire weapon like the fucking combat shotgun to even consider dealing with then I think the answer is not to keep shotguns strong so they can shit on holoparasites and also everyone else but instead to nerf holoparasites.

Georgeur:
@AdamElTablawy 
dude I said a little anxious, if you promise to make holoparas not complete bullshit and requiring current shotguns I'm down

but it was more in response to guys saying we need even more delay, like I can say slightly more delay is okay, but some of the recommendations turn holopara fight into an easy win even with a combat shotgun.
right now your goal against holopara, with current shotguns, is just to crit them

100% down with the pr, I'm still just anxious about holopara fights

the fact thisll fix the spam m1+z at the same time bullshit from riot shotguns is more than good enough, there's literally no circumstance I'll take a combat shotgun over a riot shotgun right now

Marmio64:
make fire delay on combat shotguns be 7, and riot shotguns be 10, and this'll be golden IMO. No higher values are needed.

ghost:
@Marmio64

We'll see

JamieD1 (approved):

PR #9715, opened 2020-09-15
Canister visual update

Partheo:
I've gone and further refined Tex's sprites to try and make them more homogenous with the rest of the art style, and fixed a few gripes I had with the design.

I took the liberty of changing some of the canister colors, for consistency's sake. These new markings should be more intuitive for a new player. It'll take some getting used to, but during testing I found that the change wasn't as striking as I expected. Give it a few rounds and I'm sure this new standard will be more than appreciated. Make sure to read the label before opening a can :)

canister_showcase

You can see I also added some new designs. These are all listed as `UNUSED` in the .dmi. I have no plans to do anything with them, so feel free to use them for whatever memery you like. Just remove the `UNUSED` tag from the icon state when you do.

I added some generic metal canisters, which you may use in-game by relabeling. Additionally, making a canister with metal now produces the generic gray canister rather than the yellow caution one. You can still relabel it afterwards. Mappers may desire to use these, simply use a `canister/generic`.

The light overlay on canisters now has more states. Previously there were five states- off, blinking red, solid red, yellow, and green. Now, there are nine. off, blinking red, solid red, orangered, orange, yellow, yellowgreen, green, and a new state, glowing white, which displays when the pressure in the canister is above the pressure that is achievable with a volume pump.
KkCluRCHA2

Canisters now display when they are open to the air.

Cleaned up prototype canister code, not that they matter.

Fixed a runtime that occurred when canisters with no air would try to update their icon.

I moved all the sprites to a new canister.dmi in icon/atmospherics. Previously they were at icon/atmos.dmi.

Here is the raw file with all the states, for your leisure:
canister2

#### Changelog

:cl: Tex & Partheo
experimental: canister labeling and coloration has been updated to a new standard. make sure to pay attention to their labels!
imageadd: canisters have received a visual update
imadeadd: canisters now glow white instead of green when above volume pump pressure threshold
imadeadd: canisters now show when they are open to the air
tweak: canisters made from metal now create a generic gray canister
fix: fixed canister icon update runtime
/:cl:


Jarod1200:
Is this loss?

boodaliboo:
> Is this loss?

l ll
ll l_

Partheo:
If someone can help me with the travis error, I cba to figure it out right now

patpol4:
red tritium :(

MegaEmpirical:
... Green N2 and N2O? Not a fan.

Partheo:
> 
> 
> ... Green N2 and N2O? Not a fan.

Nitrogen bottles are often green IRL. Green is also a much more intuitive color given that the gas itself is inert(safe). It allows us to use red for something that's actually dangerous, like tritium. Which is doubly good because hydrogen often comes in red bottles IRL too.

MegaEmpirical:
That's only true if you're using the EU Standards and the marking for N2 is a black shoulder whereas the green is to mark inertness.

UselessTheremin:
I would still like the trit to be green, please

JamieD1 (approved):

PR #9718, opened 2020-09-15
adam makes another one-file change in the web editor: buffing cryogenic blob edition

ghost:
this blob chem is so shit it could double as a yog admin

now the poison it injects isn't the weakest poison this side of weedkiller anymore and it injects more of it, compensating for the fact that it only does ten fucking damage. also it does burn and not brute damage now because brrr cold
#### Changelog

:cl:  
tweak: Cryogenic poison is now less fucking worthless as a blob chem
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blobstrains/cryogenic_poison.dm
+++ b/code/modules/antagonists/blob/blobstrains/cryogenic_poison.dm
@@ -21,12 +21,12 @@
 	if(M.reagents)
 		M.reagents.add_reagent(/datum/reagent/consumable/frostoil, 0.3*reac_volume)
 		M.reagents.add_reagent(/datum/reagent/consumable/ice, 0.3*reac_volume)
-		M.reagents.add_reagent(/datum/reagent/blob/cryogenic_poison, 0.3*reac_volume)
-	M.apply_damage(0.2*reac_volume, BRUTE)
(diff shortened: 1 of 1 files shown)


DiamondSentinel:
If you're just going to spew insults every time you interact with the server, why do you bother posting things on the git?

Aside from your addiction to pissing people off, change looks good. Cryo needs a bit of a buff, although this might prove a bit much.

ghost:
@DiamondSentinel 
Am I fucking wrong though?

DiamondSentinel:
Regardless of whether or not you think that about the admins, there's no reason to spout it here.

boodaliboo:
> Regardless of whether or not you think that about the admins, there's no reason to spout it here.

does it matter?

MonsWTF:
> > Regardless of whether or not you think that about the admins, there's no reason to spout it here.
> 
> does it matter?

Yes? You support it once, you are leaving a grey area that then encourages people to keep doing it and to keep toeing the line, which will just make it go further and further. You do not have to be negative, or type like an asshat, to make a PR.

ghost:
@MonsWTF

Why do I care?

Moja10:
You should care because you are already on thin ice as it is making PRs despite being permabanned from the server with basically no chance of getting unbanned.

JamieD1 (approved):

PR #9722, opened 2020-09-16
remove knockdown from hand teleport

SomeguyManperson:
literally everyone uses this since it moves people next to you and forces them to drop whatever they're holding

:cl:  
tweak: the hand holoparasite ability no longer knocks down
/:cl:

### DIFF
--- a/yogstation/code/modules/guardian/abilities/major/hand.dm
+++ b/yogstation/code/modules/guardian/abilities/major/hand.dm
@@ -16,7 +16,6 @@
 		AM.forceMove(hand_turf)
 		if(isliving(AM))
 			var/mob/living/L = AM
-			L.Knockdown(10)
 	guardian.face_atom(hand_turf)
 	return ..()
 
--- a/yogstation/code/modules/guardian/abilities/major/hand.dm
+++ b/yogstation/code/modules/guardian/abilities/major/hand.dm
@@ -14,8 +14,6 @@
 		if(AM.anchored)
 			continue
 		AM.forceMove(hand_turf)
-		if(isliving(AM))
-			var/mob/living/L = AM
 	guardian.face_atom(hand_turf)
 	return ..()
 


DiamondSentinel:
No, literally everyone uses it because none of the other abilities function at a basic level. Gravity is just flat-out broken (I've tried fixing it, but spell code is whack), explosives are terrible and super obvious, assassin is a fucking joke, time is too expensive for what it gives, frenzy is actually worthless, and those are the only ones I can remember. 

Sure, nerf The Hand. But don't expect that nerfing The Hand will shift holoparasite meta. These other abilities _need_ to be fixed and/or buffed for any other build to be considered.

TheLinuxOS:
The most broken thing about the hand isn't the fact that it knocks down, it is the fact that you can just grab people from the other side of tables or through a window, and then they are screwed if you are remotely competent. I feel the true solution to balancing the hand is to straight-up remove the ability and replace it with another one.

DiamondSentinel:
I do agree with Linux. It's spaghetti code held together by the hopes and dreams of sweaties everywhere and is made to be inherently uninteractive. You can't defend against it. Even removing the knockdown, you'll still get your ass beat. 
It's a shitty reference anyways, and a broken one at that. Let's remove it and make something actually unique.

ghost:
why not just go back to old holoparasites?

bomb, ranged, healing, chaos, etc were all good and unique

Reed0506:
what if ZA HANDO just tried to fling people and stuff towards you like you can with buildmode. would be better than the current broken mess we have right now.

TheLinuxOS:
I like reed's idea, that would make it useful and less broken than it currently is

ynot01:
@ adam old holoparasites were boring and ass and just as unbalanced as these ones

DiamondSentinel:
I mean, new holos are literally just old holos but with the powers moved to major abilities.

fluffe9911 (approved):
its a baindaid on a bigger issue but whatever

PR #9727, opened 2020-09-16
makes minebots and goldgrubs actually mine and goldgrub on lavaland

SomeguyManperson:
quick search used for non-station levels doesn't check for objects so they have to wait a way longer time than they should to check for items

:cl:  
tweak: goldgrubs and minebots on lavaland will now actually search for ores
/:cl:


Marmio64:
a reason to kill goldgrubs? Hot damn!

SomeguyManperson:
goldgrubs already spawn with 3 ores, now they'll try to steal your ores if they're close enough

JamieD1 (approved):

PR #9737, opened 2020-09-17
Lizards now gain a slight speed boost when resting at a cost of losing both arm functionality

ynot01:

/https://public-media.si-cdn.com/filer/6e/30/6e301885-a2c6-4dbb-b7a2-d8758e95d847/lizard.jpg)

#### Changelog

:cl:  
tweak: Lizards now gain a slight speed boost when resting at a cost of losing both arm functionality
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
@@ -39,7 +39,7 @@
 	return randname
 
 /// ALL FOURS
-/datum/species/lizard/spec_rest(mob/living/carbon/H, is_resting)
+/datum/species/lizard/proc/spec_rest(mob/living/carbon/H, is_resting)
 	var/obj/item/bodypart/l_arm/l_arm
 	var/obj/item/bodypart/r_arm/r_arm
 	if(!is_resting)
--- a/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
@@ -46,7 +46,7 @@
 		l_arm.disabled = TRUE
 		r_arm.disabled = TRUE
 		speedmod += 0.1
-	if(is_resting)
+	else
 		l_arm.disabled = FALSE
 		r_arm.disabled = FALSE
 		speedmod -= 0.1
--- a/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
@@ -38,6 +38,19 @@
 
 	return randname
 
+/// ALL FOURS
+/datum/species/lizard/spec_rest(mob/living/carbon/H, is_resting)
+	var/obj/item/bodypart/l_arm/l_arm
+	var/obj/item/bodypart/r_arm/r_arm
+	if(!is_resting)
+		l_arm.disabled = TRUE
+		r_arm.disabled = TRUE
+		speedmod += 0.1
+	if(is_resting)
+		l_arm.disabled = FALSE
+		r_arm.disabled = FALSE
+		speedmod -= 0.1
+
 /*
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
```diff
@@ -38,6 +38,19 @@
 
 	return randname
 
+/// ALL FOURS
+/datum/species/lizard/spec_rest(mob/living/carbon/H, is_resting)
```
jcatjmeow:
```suggestion
/datum/species/lizard/proc/spec_rest(mob/living/carbon/H, is_resting)
```

Hopekz:
Makes sense but should be test merged once you get this to work to see how it feels in-game.

jcatjmeow (changes requested):

Mesalikepie:
Lizards zoom away from sec

code/modules/mob/living/living.dm
```diff
@@ -451,9 +451,11 @@
 
 	if(!resting)
 		set_resting(TRUE, FALSE)
+		spec_rest(src, TRUE)
```
jcatjmeow, line 454: Ok, here's your issue. 
You're calling here spec_rest. What this actually means is "mob/living/spec_rest" of which a proc does not exist.

The only spec rest that exists is in datum/species/lizard

so, you need to make spec rest defined at the mob level instead. Or, you can make a temporary object like so
`var/datum/species/lizard/L = new()`
and then call the proc like this
`L.spec_rest(src, true)`

jcatjmeow: havent tested this but im like 80% sure itll work

jcatjmeow (changes requested):

Mqiib:
Can this be toggleable so I don't immediately drop everything when I'm trying to crawl under braindeads to fix something?

maxion12345:
Lizards now unable to use hands.

Because no thumbs

ynot01:
> Can this be toggleable so I don't immediately drop everything when I'm trying to crawl under braindeads to fix something?

can't exactly carry things when you're on all fours can you

jcatjmeow:
To be honest this should just be an option for all species, to scramble away fast

UselessTheremin:
this is really funny i want this now. as a trash scalie lizard main, i approve *hiss ing and skittering away from sec like the little goblin you are.

Marmio64:
if you add this, nerf the heat resistance lizards get, so they have more drawbacks in relation to their upsides.

Mqiib:
Does this allow them to ignore the slowdown from crawling or does it make their crawl very slightly faster? If the former then this would be maybe balanced considering the speed bonus is very small, and if the latter then this is arguably a downside.

ynot01:
> Does this allow them to ignore the slowdown from crawling or does it make their crawl very slightly faster? If the former then this would be maybe balanced considering the speed bonus is very small, and if the latter then this is arguably a downside.

as the code currently (non-functionally) stands they only get a slight speed boost over crawlspeed but i believe everyone thinks it's the latter so ill look into doing that

Mqiib:
Are you saying it's supposed to be slightly-faster-than-normal-crawl speed or slightly-faster-than-run speed? Because, again, if it's only slightly boosted crawl speed (ie slightly faster than slow) for the trade-off of not being able to use hands while resting then, assuming it's not toggleable, this is a bad thing.

patpol4:
Hot

Reed0506:
this is pretty stupid and i dont like it.

lizards are far more human than animal and i cannot see a reason this would make sense.

Jutastre:
> 
> 
> Can this be toggleable so I don't immediately drop everything when I'm trying to crawl under braindeads to fix something?

What if you got a speed boost only if your hands are already empty? So you can still crawl at normal speed while holding items.

ynot01:
lost motivation and probably wont merge

PR #9738, opened 2020-09-17
Adds Syndicate uniforms that have suit sensors

MrHorizons:
One of the complaints of the last Syndicate Station round was that the Syndicate uniforms don't have sensors. While someone suggested that the admins could varedit them to have sensors, it seems like this would be very inefficient, especially for spawning them in quantities sufficient for an entire crew. Thus, I have decided to make some new uniforms specifically for use in events like these. These are:

- **Syndicate uniform**: A generic, armorless (aside from fire and acid resistance) tactical turtleneck, intended for general crewmembers.
- **Syndicate security uniform**: Intended for security, this adds 10 melee armor and comes with pockets/the combat uniform sprite.
- **Syndicate command uniform**: Obviously intended for command, this is almost identical to the security uniform.

#### Changelog

:cl:  
rscadd: Added the Syndicate crew, security, and command uniforms to code.
/:cl:

### DIFF
--- a/code/modules/clothing/under/syndicate.dm
+++ b/code/modules/clothing/under/syndicate.dm
@@ -56,3 +56,29 @@
 	can_adjust = FALSE
 	armor = list("melee" = 5, "bullet" = 0, "laser" = 0,"energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 0, "acid" = 0)
 	resistance_flags = NONE
+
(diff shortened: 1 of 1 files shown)


Moja10:
Why close?

maxion12345:
Because it's not really needed. They are meant to have no sensors, and I believe there is a toy version that looks the same and has em regardlessly

PR #9739, opened 2020-09-17
fix oversight that makes pockets not get checked for heretic sacrifices

SomeguyManperson:
:cl:  
bugfix: heretics now have their pockets checked for their book when they sacrifice someone and need to get a charge
/:cl:


fluffe9911 (approved):
fine

PR #9754, opened 2020-09-19
constructs take extra damage from the nullrod

ghost:
why not nerf cult a little bit further by making construct spam untenable if there's a chaplain around?

null rods will do an extra **45** damage on top of their normal damage to any construct they hit, which is enough to oneshot artificers, twoshot wraiths and 3shot juggernauts.

#### Changelog

:cl:  
tweak: Nullrods do extra damage to constructs. It's a LOT of damage. Don't get hit by one.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/constructs.dm
+++ b/code/modules/mob/living/simple_animal/constructs.dm
@@ -110,12 +110,12 @@
 	if(updating_health)
 		update_health_hud()
 
-/mob/living/simple_animal/hostile/construct(obj/item/W, mob/living/user, params)
+mob/living/simple_animal/hostile/construct/attackby(obj/item/W, mob/living/user, params)
 	. = ..()
 	if(istype(W, /obj/item/nullrod))
 		visible_message("<span class='warning'>[src] recoils from the blow!</span>", \
 						"<span class='cult'>As \the [W] hits you, you feel holy power blast through your form, tearing it apart!</span>")
-						adjustBruteLoss(45)
+		adjustBruteLoss(45)
 
 /////////////////Juggernaut///////////////
 /mob/living/simple_animal/hostile/construct/armored
--- a/code/modules/mob/living/simple_animal/constructs.dm
+++ b/code/modules/mob/living/simple_animal/constructs.dm
@@ -110,6 +110,13 @@
 	if(updating_health)
 		update_health_hud()
 
+/mob/living/simple_animal/hostile/construct(obj/item/W, mob/living/user, params)
+	. = ..()
+	if(istype(W, /obj/item/nullrod))
+		visible_message("<span class='warning'>[src] recoils from the blow!</span>", \
+						"<span class='cult'>As \the [W] hits you, you feel holy power blast through your form, tearing it apart!</span>")
+						adjustBruteLoss(45)
+
 /////////////////Juggernaut///////////////
 /mob/living/simple_animal/hostile/construct/armored
 	name = "Juggernaut"


Bibby0110:
Hmm yes let's nerf cult, a gamemode with 37% win ratio

ghost:
@TheGamerdk Yes, let's.

It's overpowered.

Tipy1802:
Good, the chaplain does need something like that so he can be more useful during cult other than just hitting water tanks with his book

Bibby0110:
> @TheGamerdk Yes, let's.
> 
> It's overpowered.

That's pretty... Contradictory...

Mesalikepie:
Cult right now despite the win rate once it gets rolling is unstoppable. There's no good counters and the only reason the win rate is so low is because the gamemode is either cult rolls and snowballs immediately or security stops them early due to a players fuckup. The only reason cult doesn't crush every time is the players in the cult. That's it. This PR isn't even a large nerf to cult just a slight one and only really affects cult badly after they have started steamrolling

Hopekz:
Yes, let's give mining nullrod loot more power to sweep cults even harder.

ghost:
@Hopekz Yes, let's.

Cult is incredibly overpowered.

If mining nullrod loot is a problem, _i'll just remove it from mining._ Simple!

Hopekz:
> If mining nullrod loot is a problem, _i'll just remove it from mining._ Simple!

Alright. Go do that then.
This PR isn't good with mining having multiple different nullrod loot spawns in my opinion.

ghost:
I'm not going to do that until someone tells me it's a problem with demonstrable evidence. If miners wipe out the cult every single round with this because it offers a defense against construct spam, i'll think about it-but considering miners already HAVE a 45 damage katana with block and magic resistance I don't see anything changing.

Testmerge it.

Reed0506:
adam

more like

correct

Marmio64:
This sounds retarded to have along with soul shard purifying, it just makes constructs and trying to make them not even worth it if the tools used to make them can just be easily used against you and if they are super weak once they are actually finally made. Constructs aren't even OP, you can easily kill all of them, just grab a gun, or just beat up wraiths and artificers since their health pool is abysmal and easy to cuck. Just sounds like an unneeded nerf for the wrong place, and that other aspects of bloodcult should be looked at instead.

ghost:
@Marmio64 

No, it doesn't make constructs super weak. It means there's a counter to constructs that isn't 'buy more shotguns' and puts more pressure on the cult to eliminate the chaplain.

Constructs are effectively infinite in number and extremely easy to make. Wraiths are generally a shuttle call in their own right with their ability to take out comms and the AI both, and juggernaut spam and artificers mean that they'll overrun the station rather quickly.

Nerfing construct spam is a good starting point for nerfing bloodcult, because one of the major issues with bloodcult is that even their spam-creatable simple mobs are incredibly useful. No other antagonist has this level of summoning ability-heretic summons are weaker, nuclear operative summons are AI controlled and more expensive, clockwork cult summons are hardcapped and obliterated outright by the PRESENCE of the nullrod.

Blood cult specializes in simple mob spam and their simple mobs are incredibly potent-150 HP on a juggernaut is no joke, and with an artificer following it around to repair the only plausible counter is once again, the combat shotgun. This should open up some leeway by allowing a competent chaplain to rip through constructs like tissue paper, forcing cultists to deal with him themselves lest he crush their waves of constructs.

Marmio64:
@AdamElTablawy 

I still believe the 63 damage from one standard nullrod hit is too punishing and is placing even more power upon one single item/person (the chaplain), when he already has a shit ton of power in a cult round I feel like the damage bonus should be lower to 25 instead, for 43 damage a hit, so juggernauts can be more effective at what they do, being melee specialists.

Marmio64:
I do think a better alternative could be to make construct spamming be harder to do. As in, you require a physical soulshard to capture someone, instead of being granted one when you sac someone on the rune. This will also make it so you have to choose between getting more revive rune slots or getting a new construct, so that saccing people is less powerful.

ghost:
@Marmio64 

nah i prefer the idea of chaplain tearing through an army of constructs until a cultist stabs him to death

fluffe9911 (approved):
Construct salt here we come

PR #9764, opened 2020-09-21
Liquid electricity recipe

DiamondSentinel:
Allows medbay to synthesize Ethereal Blood by electrolyzing blood.

Immediately, this allows medbay to attach an IV to Ethereals to deal with blood loss (because before, they just kinda died to heparin, I'd guess). 

Long-Term, this is to allow Liquid Electricity to be reliably used as a reagent in the future. There's already a drink recipe using it, and I plan on using it for my xenobiology rework. 

🆑
rscadd: Added a recipe for Liquid Electricity, aka Ethereal Blood
/🆑


UselessTheremin:
this is nice. the only thing better than this would be to remove ethereals 😃

boodaliboo:
another source before this pr was high energy bars from vendors which also have plasma reagent

alexkar598 (approved):

DiamondSentinel:
Energy bars have only 3u a piece. That's, uhh, not a lot.

PR #9769, opened 2020-09-22
makes skeletons heal from cream as if it was milk

boodaliboo:
consistancy
#### Changelog

:cl:  
rscadd: made skeletons process cream like milk
/:cl:


boodaliboo:
going to bed

PR #9775, opened 2020-09-23
holy water now actually blocks clockwork jaunts

SomeguyManperson:
yes this is apparently a bug
:cl:  
bugfix: holy water will now block tiles from being teleported onto by clock cultists via observation consoles
/:cl:


Marmio64:
can you... buff clockcult marauders next...?

JamieD1 (approved):

PR #9777, opened 2020-09-23
lowers shock touches instability to 20

boodaliboo:
it now only works like a melee weapon with a recharge that doesnt do much damage. doesnt seem to be worth 30 instability as is
#### Changelog

:cl:  
tweak: tweaked shock touchs instability
/:cl:


alexkar598:
.

Amelia0010 (approved):

PR #9778, opened 2020-09-24
Lazy widescreen change fix 1: LOOC

Hopekz:
🆑 Hopek
bugfix: LOOC range increased to 9 from 7 to account for the widescreen change. Before this change, people can't hear you if 2 tiles from the border of your screen.
/🆑


JamieD1:
Get the screen size instead, check the widescreen pr I did a while ago.

alexkar598:
But now people will hear you offscreen which is arguably much worst when you consider antag stuff which sometimes gets inti looc

alexkar598:
Also this matches up with the say range

Hopekz:
Don't feel like fixing this at the moment.

PR #9783, opened 2020-09-24
Makes wooden barricades disassemblable

redmoogle:

You can now disassemble wooden barricades with a crowbar(takes ~10 seconds)

#### Changelog

:cl:  
tweak: You can now disassemble wooden barricades with a crowbar 
/:cl:

### DIFF
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -76,7 +76,15 @@
 				new /turf/closed/wall/mineral/wood/nonmetal(get_turf(src))
 				qdel(src)
 				return
-	return ..()
+	else if(I.tool_behaviour == TOOL_CROWBAR && user.a_intent != INTENT_HARM)
+		user.visible_message("[user.name] starts prying [src.name] apart.", \
+							"<span class='notice'>You start prying the barricade apart</span>")
+		if(I.use_tool(src, user, 200, volume=50))
+			to_chat(user, "<span class='notice'>You disassemble the barricade.</span>")
+			new /obj/item/stack/sheet/mineral/wood(user.loc, 5)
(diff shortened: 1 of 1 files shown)


Jarod1200:
isnt the whole point of the barricades to make an obstacle that you have to destroy to get through?

redmoogle:
> 
> 
> isnt the whole point of the barricades to make an obstacle that you have to destroy to get through?

I mean yeah true, but they also turn into walls, so this is just to make getting rid of them easier

boodaliboo:
24 seconds is a reasonable time especially for its goal which is a barricade

redmoogle:
I remembered it from some codebase, couldnt hunt it back down so I had to come up with my own time

Marmio64:
> I remembered it from some codebase, couldnt hunt it back down so I had to come up with my own time

paradise had barricades removable with a crowbar. Took like 10 seconds.

Jarod1200:
24 seconds sounds good.

Hopekz:
A good feature that I felt was missing but 24 seconds is a long time.
Could easily turn in to almost a full minute with lag and time dilation.

Just make it the same amount of time as beating a barricade with the tool so you have both options as basically flavortext on how you want to take it down

redmoogle:
> 
> 
> A good feature that I felt was missing but 24 seconds is a long time.
> Could easily turn in to almost a full minute with lag and time dilation.
> 
> Just make it the same amount of time as beating a barricade with the tool so you have both options as basically flavortext on how you want to take it down

should be somewhere around 10s now, tooldelay is weird, it doesnt seem to be in any semi-standard unit of time

JamieD1 (approved):

PR #9803, opened 2020-09-25
makes mind transferring move perks to the new body but it doesn't break literally everything

SomeguyManperson:
Reverts yogstation13/Yogstation#9786


SomeguyManperson:
Im going to gut whoever msde quirks

PR #9845, opened 2020-09-30
Allows armbands, medals, and the bone talisman/codpiece to display over suits

MrHorizons:
I was planning to make this part of my PR idea to (semi-)unlimit the number of accessories that a uniform can have, but since I'd probably be shouted at to atomize it anyway, I've decided to do it in this PR.

This is good for when you have a departmental armband, but also have to commonly (or just want to) wear a suit that obstructs it. I've tested it some and it only doesn't really look good on space suits, to which I'd be tempted to code some bit in that checks to see if the suit is one, but the difference is small enough that I can wait until a time that isn't midnight to do it.

#### Changelog

:cl:  
tweak: Allowed armbands, medals, and the bone talisman and codpiece to display over suits.
/:cl:

### DIFF
--- a/code/modules/clothing/under/accessories.dm
+++ b/code/modules/clothing/under/accessories.dm
@@ -116,6 +116,7 @@
 	resistance_flags = FIRE_PROOF
 	var/medaltype = "medal" //Sprite used for medalbox
 	var/commended = FALSE
+	above_suit = TRUE
 
 //Pinning medals on people
 /obj/item/clothing/accessory/medal/attack(mob/living/carbon/human/M, mob/living/user)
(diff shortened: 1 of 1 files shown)


Tipy1802:
FInally medals are visible even when wearing a suit (which you do most of the time)

Smudged:
Finally, we can see the armbands 🥇

code/modules/clothing/under/accessories.dm
```diff
@@ -357,3 +360,4 @@
 	above_suit = TRUE
```
SomeguyManperson: Codpiece already has this

MrHorizons: Whoops.

UselessTheremin:
Shaft miner gonna look like commando from risk of rain 1 smh

But in all honesty good change

Mqiib:
Goliath cloaks look a little weird but as long as you're just using the bone talisman it's fine.

JamieD1 (approved):

PR #9848, opened 2020-09-30
Snappop shuttle now has a brig, now does not have a REAL giant pit or a "Greentext"

jcatjmeow:
This shuttle is pretty cool EXCEPT for the big fucking pit and the haha xd greentext. So I fixed it.

Plus it's a personal pet peeve of mine to have shuttles that do not line up with ALL the airlocks. 

Also added: Flashers for both brig and cockpit

Compiled, not tested but I think the accesses on the airlocks are good? I hope? screm

And yes this is part of my mrp stuff. 
#### Changelog

:cl:  
rscadd: Added a proper brig to snappop, added a door, added flashers and a fake chasm brig tile  
rscdel: Removed the real chasm and the greentext book from the snappop shuttle
/:cl:

### DIFF
--- a/_maps/shuttles/emergency_clown.dmm
+++ b/_maps/shuttles/emergency_clown.dmm
@@ -113,11 +113,17 @@
 /turf/closed/wall/mineral/titanium,
 /area/shuttle/escape)
 "au" = (
-/turf/open/chasm,
+/obj/structure/chair/bananium,
+/turf/open/floor/mineral/plastitanium/red/brig/fakepit,
 /area/shuttle/escape)
 "av" = (
 /obj/machinery/light{
 	dir = 1
+	},
+/obj/machinery/flasher{
+	id = "cockpit_flasher";
+	pixel_x = 6;
+	pixel_y = 24
 	},
 /turf/open/floor/mineral/bananium,
 /area/shuttle/escape)
@@ -139,14 +145,12 @@
 /turf/open/floor/mineral/bananium,
 /area/shuttle/escape)
 "az" = (
-/obj/item/greentext/quiet{
-	anchored = 1
-	},
-/turf/open/floor/mineral/plastitanium/red/brig,
+/turf/open/floor/mineral/plastitanium/red/brig/fakepit,
 /area/shuttle/escape)
 "aA" = (
 /obj/machinery/door/airlock/bananium/glass{
-	name = "Emergency Shuttle Greentext"
+	name = "Emergency Shuttle Brig";
+	req_access_txt = "2"
 	},
 /obj/machinery/door/firedoor/border_only{
 	dir = 4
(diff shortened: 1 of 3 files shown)


patpol4:
I like the shuttle as it is but you do have a valid point. Will review in a bit

patpol4 (changes requested):
dreamseeker_GInNK4JpjL

This looks gross, i don't think this shuttle should even have a designated sec area. I can't accept this change as is. How about something like this? 
dreamseeker_gLyauOKE8k
Removes the lrp chasm and greentext, gives the crew something fun, makes it pleasant to the eye.

jcatjmeow:
> This looks gross, i don't think this shuttle should even have a designated sec area.

I disagree, I think it NEEDS a designated sec area, same as every other shuttle.

Because like, think about it this way. If there is no brig, there is zero incentive to keep prisoners alive or bring them with you. How it was with the old shuttle was you'd just throw them in the chasm. If we don't have any sort of brig, prisoners will get executed or left to rot on the station, and that's lrp as far as I'm concerned, plus it takes some of the fun that can potentially come from last minute rescues.

If you have suggestions on how to make it look better be my guest but I want the sec area, that's like half the entire point of the pr.

I don't have to go with the chasm tiles, I did just do that to keep with the original chasm look. But I could easily do something else, maybe using the ballpit tiles? Maybe something soft like a padded cell? Idk

TheLinuxOS:
I like what wejengin suggested, I dont believe a shuttle that is meant to be a clown shuttle should have a designated brig area and the layout he suggested looks nice to me

patpol4:
I just don't feel like every shuttle needs to be up to standards and equipped with the exact same rooms/machinery, especially when it's evident that it's not a serious shuttle and instead a gimmick one. Wouldn't it be better to go along with the gimmick rather then making all shuttles almost the same? If you really want the sec area, maybe try a ballpit and beds, that could look better than chairs on chasm. I will accept that for the sake of mrp but my point about gimmick shuttles stands, they are supposed to be unique.

jcatjmeow:
Gimmicks should never impede or ruin gameplay and roleplay.

The gimmick of snappop is the snappops, not the lack of a brig

jcatjmeow:
@wejengin2 ballpit looks a bit ugly, I think the sprite itself is kinda bad, but how does this look?

I could always go back to regular floors, with anti-sec graffiti on them I think it would look fun

athena413:
> i don't think this shuttle should even have a designated sec area.

ALL non-emag shuttles need a sec area to shove prisoners into.

maxion12345:
You fools
The pit is the brig. They'll redtext anyways

patpol4 (approved):
Map is fine, we went with this as a compromise :
vivaldi_thUdzI9wpr

PR #9851, opened 2020-09-30
Makes the chaplain's energy swords deal burn damage

MrHorizons:
They look and sound like normal energy swords, so why not make them deal burn damage, too?

Besides, the God Hand does an equal amount of burn damage, and I'd say trading disarm immunity for 30 block chance and the ability to use the hand slot for other things is fair.

#### Changelog

:cl:  
tweak: Made the chaplain's energy swords deal burn damage.
/:cl:


MrHorizons:
Fuck, energy swords do brute for some reason.

PR #9854, opened 2020-10-01
Buffs WT rifle AP and adds rubbershot magazines to the WT rifle

SomeguyManperson:
Increases AP of WT bullets by 20 from 0 and increases AP of WT AP bullets to 50 from 40, making them more competitive against shotgun slugs when fighting armored targets
Adds a new kind of bullet, for the WT rifle which can be printed with ballistics tech or ordered from cargo, rubber rounds, which deal far less damage, have no AP, but deal stamina damage.
:cl:  
rscadd: new ammo for the WT rifle: rubbershot rounds for less than lethal bullets, which deal 20 stamina damage per hit but have no armor penetration
tweak: 4.6x30mm bullets have 20 AP up from 0, and their AP variants have 50 AP up from 40
/:cl:


Mqiib:
Huzzah, something that makes WTs not totally suck! Now there's actually a reason to get them over just getting a bunch of laser guns!

Marmio64:
kind of a high AP for the standard rounds, 10 would seem more reasonable especially if your 2 round burst PR gets implemented too. WT-550 already gets an advantage over the laser gun that many people seem to forget, you can reload it on the fly, and it has 20 rounds in a mag, while laser gun can only fire about ~12 times.

Partheo (approved):

JamieD1 (approved):

PR #9869, opened 2020-10-03
Hierophant becomes old hierophant again

swissloaf:

It looks better compared to the current one tbh 

hierophant_old

hierophant

#### Changelog

:cl:  
tweak: Changes icon for hierophant
/:cl:


Takahiru:
How about let's not completely rip off hld and keep the current Hierophant, the current one looks good with the style.

Partheo (changes requested):
I think this change is unnecessary but let's see what people think.

Change the appropriate descriptions too. This is an extremely lazy PR.

MrHorizons:
What if we gave the current Hierophant sprite the clothes of the old one?

Readystorm546:
Yo thats badass

Reed0506:
i dont like this at all

PR #9874, opened 2020-10-03
adds clockcult anchors

Amelia0010:
Requested by @alexkar598, not fully done yet

Adds clock cult anchors. When you warp to the station, instead of warping to any tile you want, you'll instead be warping to a randomly chosen tile in maintenance (initially picking from xeno/blobstarts, then widening out if none of them are acceptable).

The anchors are passthrough for clock cultists.

To get back to the station, you'll need to either abscond next to the anchor, hit the anchor with your slab, or have someone else teleport you back using their anchor.

I'll leave arguing why this is a good thing for @alexkar598 since I just wanted something to code and he was gonna make this anyways.

:cl:
tweak: Warping to the station as a clock cultist now spawns an anchor which you need to teleport out with again, making clock cultists significantly less mobile
/:cl:

### DIFF
--- a/code/modules/antagonists/clockcult/clock_structures/teleport_anchors.dm
+++ b/code/modules/antagonists/clockcult/clock_structures/teleport_anchors.dm
@@ -1,0 +1,33 @@
+/obj/structure/destructible/clockwork/anchor
+	name = "Anchor"
+	desc = "" //A aaaaaaaaaaaaaa
+	clockwork_desc = "Abscond next to this or hit it with your slab to warp back to Reebe."
+	icon_state = "mania_motor"
+	break_message = "<span class='warning'>The antenna break off, leaving a pile of shards!</span>"
+	max_integrity = 100
+	light_color = "#AF0AAF"
+
+	var/datum/mind/owner
+
+/obj/structure/destructible/clockwork/anchor/attackby(obj/item/I, mob/user, params)
+	if(istype(I, /obj/item/clockwork/slab) && is_servant_of_ratvar(user))
+		if(owner != user.mind)
+			to_chat(user, "<span class='danger'>You need to use your own anchor.</span>") // Remind me to make a pointer to their anchor
+			return FALSE
+		var/obj/item/clockwork/slab/s = I
+		return s.recite_scripture(/datum/clockwork_scripture/abscond, user)
+	return ..()
+
+/obj/structure/destructible/clockwork/anchor/proc/disable() // Called when our owner is deconverted
+	animate(src, alpha = 0, 50)
+	QDEL_IN(src, 50)
+	for(var/mob/m in view(7, src))
+		if(is_servant_of_ratvar(m))
+			to_chat(m, "<span class='notice'>The anchor fades out of existence, as its connection to its owner has been severed.</span>")
+		else
+			to_chat(m, "<span class='notice'>The anchor fades out of existence.</span>")
+
+/obj/structure/destructible/clockwork/anchor/CanPass(atom/movable/mover, turf/target)
+	if(istype(mover, /mob/living) && is_servant_of_ratvar(mover))
+		return TRUE
+	return ..()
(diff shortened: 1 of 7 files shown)


code/modules/antagonists/clockcult/clock_structures/teleport_anchors.dm
```diff
@@ -0,0 +1,33 @@
+/obj/structure/destructible/clockwork/anchor
+	name = "Anchor"
+	desc = "" //A aaaaaaaaaaaaaa
```
Amelia0010, line 3: Description to be written pending sprite

code/modules/antagonists/clockcult/clock_structures/teleport_anchors.dm
```diff
@@ -0,0 +1,33 @@
...
+	clockwork_desc = "Abscond next to this or hit it with your slab to warp back to Reebe."
+	icon_state = "mania_motor"
```
Amelia0010, line 5: Sprite pending

Mqiib:
Wait so can non-clockies see these in maint? Wouldn't this just mean clockies will get called out even faster? There's no point in NOT declaring war if this is the case since they'll just get spotted within like 2 or 3 minutes when someone sees an anchor. Why not make them like the portals but only the cultists can see them?

ynot01:
needs a buff to balance out this, i wont say that clockies arent ridiculously mobile, but they're hard cucked by the time limit

Xoxeyos:
So can these move around when you teleport back from reebe to the station again, or is it always the same place?

alexkar598:
Not sure if I mentioned it to you before, but they should only enable when clock cult has like 7-10 clockies to prevent early game marooning

ghost:
well, i guess the cult will need to learn to convert on station

Amelia0010:
@alexkar598 oh.

Marmio64:
Yeah, clockcult would need some buffs if they're going to get such a HEAVY nerf like this. Perhaps you can abscond up to maybe like 20 tiles away from an anchor, but the further you are the longer it takes, up to 10 seconds, and maybe a bit higher power cost. Make sure anchors are invisible to non-cultists too. And another thing, marauders are pretty shit but if buffed could make being station bound more viable with them acting as bodyguards/attack dogs. Right now their health is only 120 with a measly 12 damage, 15 if war is declared, they're pretty shit for their cost. They could use a 3 damage boost and maybe a 30 health boost so they can actually fight people.

ToGWtF:
I feel like someone is going to end up seeing one of the spots and camping it. Then it will be a 1/whatever chance that you just teleport infront of someone with a shotgun in hand and theres no counter to that.

Amelia0010:
>Wait so can non-clockies see these in maint?

@Mqiib Yes

>Wouldn't this just mean clockies will get called out even faster?

@Mqiib It doesn't start using anchors till you have 8 clock cultists

>So can these move around when you teleport back from reebe to the station again, or is it always the same place?

@Xoxeyos It picks randomly every time you teleport to the station, if you already have an anchor on the station (say you got teleported up by someone else) you get to choose whether you want a new one

>I feel like someone is going to end up seeing one of the spots and camping it. Then it will be a 1/whatever chance that you just teleport infront of someone with a shotgun in hand and theres no counter to that.

@ToGWtF This is literally impossible. 1st of all, there would be a counter, that being that you spawn on the anchor so the shotgun would hit that, not you. 2nd of all, it won't pick tiles that someone can see if it can avoid it.


@alexkar598 You should probably consider the balance concerns

alexkar598:
The point of this is so that the station has an actual fighting chance against the clockcult, and forcing them to operate on the station as much as possible.

Can you camp them? Yes but you're better off simply destroying it and trying to find another one. It's unlikely sec can destroy them all.

Also Nich made it(lov u nich) so that it wont if it can help it teleport you next to someone so you wont get shotgunned instantly

alexkar598:
Also about the time limit, station is way more affected by it than the clock cult. Clockcult has pretty much free reign to do whatever er they want to do until the 25 ish minute mark which is when the station starts waking up, implanting, searching and ordering new guns because they've been stolen by the clockcult.

Clockcult is one of our most unbalanced gamemode atm as the only way they lose is when they're found out instantly and they end up with a single cultist for a while. 

Is this a huge nerf? Yes it takes away the free port in port out mechanic and that's huge. But clockcult needs that nerf, it's impossible to win against a clockcult with even 3 members with braincells(which is admittedly hard to find)

Marmio64:
Theos' PR is better to be honestly, this one is too hard of a nerf.

alexkar598:
They need the nerf way more than modes like blood cult do

alexkar598:
Dunno what's up with the low amount of clockcult rounds tho that's weird

Hopekz:
As it stands I think this is a weaker alternative to #9886
Please see my comment in #9886 if interested in my reasoning as I have typed out a few weaknesses with this PR vs #9886 there

JamieD1:
@Partheo Sprites pl0x

PR #9878, opened 2020-10-04
White ship randomization + minor renaming

MrHorizons:
Partial port of tgstation/tgstation#53603.

Aside from porting a system to choose a random white ship from the options available, I've also decided to change the names of the individual white ship maps to match their shuttle manipulator names to be more generic.

#### Changelog

:cl:  
rscadd: Made the white ship chosen for each map independent of it.
tweak: Changed the name of the white ships to be more generic.
/:cl:


patpol4 (approved):
Not really much a map change, just a replaced docking port with the one you created and some file renaming. I'm fine with the names, unless @Ktlwjec1 has anything against this is fine from the mapping side of things.

PR #9880, opened 2020-10-04
Makes yellow toolboxes spawn with budget insuls 50% of the time instead of real gloves 10% of the time

alexkar598:
Imagine needing an admin meeting and 5 minutes of talking to figure this out

@JamieD1

### DIFF
--- a/code/game/objects/items/storage/toolbox.dm
+++ b/code/game/objects/items/storage/toolbox.dm
@@ -137,8 +137,8 @@
 	new /obj/item/crowbar(src)
 	new /obj/item/stack/cable_coil(src,MAXCOIL,pickedcolor)
 	new /obj/item/stack/cable_coil(src,MAXCOIL,pickedcolor)
-	if(prob(5))
-		new /obj/item/clothing/gloves/color/yellow(src)
+	if(prob(50))
+		new /obj/item/clothing/gloves/color/fyellow(src)
 	else
 		new /obj/item/stack/cable_coil(src,MAXCOIL,pickedcolor)
 


Georgeur:
No.

JamieD1 (approved):

Moja10:
You missed the entire point of the staff meeting, no one had issues with them existing. It was about bwoinking people for having insulated gloves despite there being valid ways to get them on the station. You are completely missing the point entirely with this PR.

Tipy1802:
Great pr! Should probably get completely removed from the loot table imo but baby steps

Georgeur:
no seriously, no.

Hopekz:
Now that we have departmental wires, learning wires for hacking needs to be done a lot more often (multiple times a shift) so I think this PR is appropriate since we have no extra support since that was being required.

jcatjmeow:
I want to first point out that budget insuls are actual trash. They might protect you a little, sure, but they don't protect against the stun, and that's the actual important bit.

Insuls being god tier items that are sought after by powergamers is a bit of a meme, but it's rooted in truth. Shocking doors is incredibly strong and the ability to be entirely immune to that was and is still a huge deal.

I think the issue is not the insuls being too strong, but the ridiculous strength of electric shocks, even small ones but certainly when the power grid gets cranked up for whatever reason. It's actually absurd, and it leads to this arms race where people NEED these golden op as fuck gloves to do anything or they are at a major handicap, and then we nerf those people by taking away the shiny gloves cause the gloves ended up too strong...all the while the real issue is and always has been the stupid shock mechanics.

But uh, manifesto aside 😓 budget insuls are pretty much worthless. I would fully support this change if shocking was nerfed, or the budget insuls were buffed to not stun you even if you take damage (or at least not knock you down for a million years) so they are actually usable. As it stands right now I'm neutral on it.

Runian:
I'm down to rush medical for burn meds (to heal after hacking wires, cuz I'm still gonna do it anyway) and latex gloves that have better protection than insulated budget gloves.

Marmio64:
like jcat said, i'd be down for this if budget insuls were actually made useful. Maybe make budget insuls remove the stun from shocks but not provide any damage resistance.

ToGWtF:
f in chat for the chosen assistants

Amelia0010:
@yogstation13/head-coders you two decide please

PR #9886, opened 2020-10-05
clock cultists now warp to the station far slower after converting 20% of living players unless they are teleporting onto clockwork floors

SomeguyManperson:
alternative to #9874
Intent is to reduce the speed clock cultists can get into/out of locations they don't already have access to after converting enough people, should slow them after the early game so later game sabotage becomes more of a pain in the ass to do in protected station locations

current warp-in penalty after 20% of living players are converted sets the warp-in timer to 30 seconds

:cl:  
tweak: clock cultists will now have a drastically slowed warp-in to the station after converting 20% of living players
tweak: the slower warp-in will not effect clockwork tiles, make sure to set something up on station!
/:cl:


code/game/machinery/computer/camera_advanced.dm
```diff
@@ -362,11 +362,15 @@
...
 	do_sparks(5, TRUE, user)
 	do_sparks(5, TRUE, T)
-	warping = new(T)
 	user.visible_message("<span class='warning'>[user]'s [target.name] flares!</span>", "<span class='bold sevtug_small'>You begin warping to [AR]...</span>")
 	button_icon_state = "warp_cancel"
 	owner.update_action_buttons()
-	if(!do_after(user, 50, target = warping, extra_checks = CALLBACK(src, .proc/is_canceled)))
+	var/warp_time = 50
+	if(!istype(T, /turf/open/floor/clockwork) && GLOB.clockwork_hardmode_active)
+		to_chat(user, "<span class='sevtug_small'>The [target.name]'s inner machinery protests vehemently as it attempts to  warp you to a non-brass tile, this will take time...</span>")
```
athena413: you accidentally put 2 spaces in "to  warp"

alexkar598:
Sounds more frustrating than anything, you teleport in and you just sit there staring at the screen then theres a clown that passes by so you quickly cancel then stare at the screen for yet another 30 seconds

Marmio64:
Far better than the idea of clockcult 'anchors'. Also can you please buff clockcult marauders...

Hopekz:
After discussing both PR's I think this is now the better option as of right now.

Nich's PR has a lot of weaknesses such as a valid hunter being able to just rush to destroy the anchors or the moment of confusion people will get when teleporting to a random dark location in maint. 
Also the obvious like the crew being able to lock + bolt all maint doors to hinder them somewhat significantly. 
Even if you don't destroy the anchors you now have a place to camp because you know culties must use these to get back.

This PR isn't perfect with the "have to wait 30 seconds" part but I think it will encourage base building on-station since it now takes a long time to teleport.

Theos also mentioned how there can be a placable structure made by clock cult that will increase the teleport speed around it which I thought was a very interesting concept that even further encourages base building.

Moja10:
This makes a ton of sense, and also has a way around it by actively converting tiles, which is often completely neglected on the station. You can totally get around this and it is a very good balancing decision.

code/game/objects/effects/temporary_visuals/clockcult.dm
```diff
@@ -249,9 +249,10 @@
 	light_power = 0.7
 	duration = 55
 
-/obj/effect/temp_visual/ratvar/warp_marker/Initialize(mapload, mob/living/servant)
+/obj/effect/temp_visual/ratvar/warp_marker/Initialize(mapload, mob/living/servant, _duration = 50)
+	duration = _duration + 5
```
jcatjmeow:
```suggestion
	duration = _duration + 5
```
**_duration** is an absolutely disgusting variable name, a more descriptive name would be appropriate, at the very least the var name should not start with an _ and should be clearly at a glance different than every other variable name in the relevant part.

jcatjmeow (changes requested):

athena413 (approved):

code/game/machinery/computer/camera_advanced.dm
SomeguyManperson: I already fixed that

Takahiru:
Encourages clock cultists to be on the station more rather than only being there for a combined time of 3 minutes.

Jarod1200:
I like this. You can easily place a few clock tiles around use them as "anchors" for faster teleport. And this encourages them to be on station more

code/game/objects/effects/temporary_visuals/clockcult.dm
SomeguyManperson: _duration is used to set the effect's duration (thus the name) which is offset by 5 as the duration was already offset from the animation's time by 5 for a reason I do not currently know but also do not want to conflict. It does not exist outside this one usage of setting the actual duration variable and animation time.

PR #9889, opened 2020-10-05
Fixes #9873

### LINKED ISSUE #9873: gloves + black crayon = FORENSIC GLOVES SPRITE?
- if you wash them together, the result really should be black gloves sprite.

ReddicusDragon:
due to how it was written, it would stop right after the first G.item_color equaled wash_color in term screwing over whenever there was more than one subtype of glove.
By removing a break statement, we take a minimal loss of performance but end up with the right glove sprite.

Fixes #9873 
#### Changelog

:cl:  
bugfix: fixed washing gloves with a black crayon = forensic sprite
/:cl:

### DIFF
--- a/code/game/machinery/washing_machine.dm
+++ b/code/game/machinery/washing_machine.dm
@@ -161,6 +161,7 @@
 				item_state = initial(G.item_state)
 				icon_state = initial(G.icon_state)
 				item_color = wash_color
+				name = initial(G.name)
 				desc = "The colors are a bit dodgy." 
 
 /obj/item/clothing/shoes/sneakers/machine_wash(obj/machinery/washing_machine/WM)
--- a/code/game/machinery/washing_machine.dm
+++ b/code/game/machinery/washing_machine.dm
@@ -162,7 +162,7 @@
 				icon_state = initial(G.icon_state)
 				item_color = wash_color
 				name = "washed gloves"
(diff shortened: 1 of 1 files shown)


code/game/machinery/washing_machine.dm
```diff
@@ -162,8 +162,7 @@
 				icon_state = initial(G.icon_state)
 				item_color = wash_color
 				name = "washed gloves"
-				desc = "The colors are a bit dodgy."
-				break
+				desc = "The colors are a bit dodgy." // removed break here, it should use the last item in the list for its color. See black gloves.
```
jcatjmeow:
```suggestion
				desc = "The colors are a bit dodgy." 
```
Don't take this 100% as gospel truth, but you probably don't need this comment here. Nobody needs to know specifically that a break was removed, that's what github history is for. This info wont actually be useful to anyone reading the code, right? I dunno.

jcatjmeow (changes requested):

ReddicusDragon:
Could be in the future, if someone finds a better way

jcatjmeow:
> Could be in the future, if someone finds a better way

I guess? /shrug

JamieD1:
Explain in the pr not the code

ReddicusDragon:
@JamieD1 updated the pr description

JamieD1:
Revert my pr by adding back the correct name @ReddTheDragon

JamieD1 (approved):

PR #9891, opened 2020-10-06
Takes out eclipse from rotation due to being a crashfest. It is still forcable by admins but shouldnt be forced outside of testing

alexkar598:
Takes out eclipse from rotation due to being a crashfest. It is still forcable by admins but shouldnt be forced outside of testing

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -38,7 +38,6 @@ map icebox
 endmap
 
 map eclipsestation
-	votable
 	minplayers 60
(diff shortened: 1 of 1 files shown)


MayhemSailor:
Support, as I’ve not received any support in solving the crash issue.

JamieD1 (approved):

JamieD1:
Inb4 admins choose them anyway

PR #9898, opened 2020-10-06
should fix the meatball man's offset since it wasn't changed when the sprite was made smaller

SomeguyManperson:
:cl:  
bugfix: meatball man's icon will show up where meatball man actually is
/:cl:


fluffe9911 (approved):
Neat

PR #9901, opened 2020-10-07
Removes non-donor service borg skin selection.

MrHorizons:
Service borgs are the only borgs with the ability to select skins without being donor. So, for consistency, we should either implement a selection of options for borg skins for all cyborgs or just remove the option for service borgs; I think the latter will go down far easier.

Besides that, one of the skins (Kent) is literally a very slightly altered medical borg, which is absolutely horrible for distinguishability. Although I think quite a few of the donor borg skins have that same problem, that's a matter for another PR.

#### Changelog

:cl:  
rscdel: Removed selecting other skins as service borg if you aren't donor. Sorry.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -514,42 +514,8 @@
 		/obj/item/borg/sight/xray/truesight_lens)
 	moduleselect_icon = "service"
 	special_light_key = "service"
+	cyborg_base_icon = "service_m"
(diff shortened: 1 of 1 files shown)


UselessTheremin:
eh, it is a nice callback to goon, and is less a skin and more so a style. think of how drones can pick their style. i think it should stay as service borgs are mainly rp roles, and the service borg style picker can further reinforce that.

JamieD1:
Considering its the worst job and has no actual use, I am keeping the skin selection.

PR #9914, opened 2020-10-08
a cute small space ruin

UselessTheremin:
i made a cute little place. come and stay a while! it has tiny fans so it doesnt space, it is fully powered, and has some little facilities. it is not really meant to give loot or even to tell a story, it is mainly just a little place to chill at and spend your round if you are a little hermit. mainly just to breathe life into space, and be a fun thing to discover. it even features hot coco, a mint under your pillow, nice carpet, a microwave, and (not) much more! here is a screenshot: 
image
tell me if i did something wrong, i am kinda a shitcoder 😳 
🆑
rscadd: new ruin to chill at
/🆑


UselessTheremin:
the gps signal is "Cozy Signal"
here are some pics of the changes: 
image
image

- added poker table by window

- added gps

- replaced the 2 fluff broken sleepers with a stasis bed and a medicine closet

Marmio64:
I suggest adding more things to look at. Decals would be nice, dirt decals in maint and some blue squares/lines in medbay would be good too.  I also suggest replacing the O2 canister with a air canister, and some potted plants would be nice.

patpol4 (changes requested):
1. You have a file with no extension in root
2. You have to add this to config/spaceruinblacklist.txt
3. We don't want to have the asteroid itself have power, use /area/ruin/unpowered/no_grav
4. any reason you made it with the labormineral rock? space ruins generally use /turf/closed/mineral/random
5. Something wrong with this spot, needs some more action, things like thin windows, a plant, matchsticks to light the cigar and fireplace, a bookcase from the above room would fit here aswell since the upper one is really tight.
chrome_PCMaTLIxBZ
6. The Diary reads really bad, "i sure hope", "i sure hope", "i sure hope"
7. Bit dark in here, a single small light on the right wall would do it
dreamseeker_2UbxDYPFhI
8. Way too many lights in the farm room, cut them at least in half, pretty sure even 2 would suffice
9. You only named some of the doors but not all of them, what about bathroom and farm.
10. This is what the chemical closet contains, doesnt really fit the theme, use something else, also one regular medkit is enough: 
CcZYGq1WHX
11. Really quick and dirty way of making this look better, med is a little too big, study is a little too small, this fixes that, try doing something similar 
chrome_XOF3pZWzic
12. Please try to standardize the mirror placements so we don't clutter vars, x=28 works fine 
chrome_C8AvxRCBLE
13. either put a single small light fixture in maint or a flashlight somewhere

patpol4 (approved):
Deleted the file for you and did some minor changes you forgot like firelocks, changed the diary to not say "sure" every sentence, plant in med, shrinked maint a little. @JamieD1 Merge when checks ready since this needs config changes.

JamieD1 (approved):

PR #9919, opened 2020-10-09
Thick skin brute/burn mod abuse fix

Mqiib:
Classic boodaliboo. 

#### Changelog

:cl:  
bugfix: Thick skin can no longer be abused to get (effectively) 0 brute and burn mods
/:cl:


code/datums/mutations/body.dm
```diff
@@ -234,16 +234,14 @@
 /datum/mutation/human/thickskin/on_acquiring(mob/living/carbon/human/owner)
 	. = ..()
 	if(owner.physiology)
```
jcatjmeow, line 236:
```suggestion
/datum/mutation/human/thickskin/on_acquiring(mob/living/carbon/human/owner)
	. = ..()
	#define BRUTEMOD 0.8
	#define BURNMOD 0.9
	if(owner.physiology)
```

code/datums/mutations/body.dm
```diff
@@ -234,16 +234,14 @@
...
 		owner.physiology.brute_mod *= 0.8
 		owner.physiology.burn_mod *= 0.9
 
 /datum/mutation/human/thickskin/on_losing(mob/living/carbon/human/owner)
 	. = ..()
 	if(owner.physiology)
-		owner.physiology.brute_mod = brutemodbefore
-		owner.physiology.burn_mod = burnmodbefore
+		owner.physiology.brute_mod /= 0.8
+		owner.physiology.burn_mod /= 0.9
```
jcatjmeow, line 244:
```suggestion
		owner.physiology.brute_mod /= BRUTEMOD
		owner.physiology.burn_mod /= BURNMOD
```

jcatjmeow (changes requested):
This is classic boodcode, lol.
That being said, I would do this... you could also use vars instead of defines if you prefer.

The idea is that now someone can edit both values by editing the one define, and it makes sure nobody will forget to change one but not the other.

By no means is this remotely necessary, just a suggestion. If these numbers were used a lot I would push harder for a define but since they're only used in 2 places each its not a big deal at all. Still, might as well I feel 

EDIT: just use vars actually they're easier in this case

jcatjmeow (approved):
you know what nevermind this is good

SomeguyManperson (approved):

PR #9953, opened 2020-10-11
Makes Budget Insuls have two charges of absorbing shocks, After which they turn to ash.

JamieD1:
Makes Budget Insuls have two charges of absorbing shocks,

After which they turn to ash.

#### Changelog

:cl:  
tweak: Makes Budget Insuls have two charges of absorbing shocks, After which they turn to ash.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -479,6 +479,9 @@
 			var/obj/item/clothing/gloves/G = gloves
 			if(G.siemens_coefficient <= 0)
 				total_coeff -= 0.5
+			var/obj/item/clothing/gloves/color/fyellow/greytide = G
+			if(greytide)
+				greytide.get_shocked()
 		if(wear_suit)
 			var/obj/item/clothing/suit/S = wear_suit
 			if(S.siemens_coefficient <= 0)
@@ -493,7 +496,11 @@
 		if(gloves)
 			var/obj/item/clothing/gloves/G = gloves
 			gloves_siemens_coeff = G.siemens_coefficient
-		siemens_coeff = gloves_siemens_coeff
+					var/obj/item/clothing/G = gloves
+			var/obj/item/clothing/gloves/color/fyellow/greytide = G
+			if(greytide)
+				greytide.get_shocked()
+		siemens	_coeff = gloves_siemens_coeff
 	if(undergoing_cardiac_arrest() && !illusion)
 		if(shock_damage * siemens_coeff >= 1 && prob(25))
 			var/obj/item/organ/heart/heart = getorganslot(ORGAN_SLOT_HEART)
@@ -501,20 +508,6 @@
 			if(stat == CONSCIOUS)
 				to_chat(src, "<span class='notice'>You feel your heart beating again!</span>")
 	siemens_coeff *= physiology.siemens_coeff
-
-	if(gloves)
-		var/obj/item/clothing/G = gloves
-		if(istype(G, /obj/item/clothing/gloves/color/fyellow))
-			var/obj/item/clothing/gloves/color/fyellow/greytide = G
-			if(greytide.damaged == TRUE)
-				qdel(G)
-				new/obj/effect/decal/cleanable/ash(src)
-				to_chat(src, "<span class='notice'>Your gloves absorb the shock and disintegrate!</span>")
-				return
-			else
-				greytide.damaged = TRUE
-				to_chat(src, "<span class='notice'>Your gloves absorb the shock!</span>")
-				return
 	dna.species.spec_electrocute_act(src, shock_damage,source,siemens_coeff,safety,override,tesla_shock, illusion, stun)
 	. = ..(shock_damage,source,siemens_coeff,safety,override,tesla_shock, illusion, stun)
 	if(.)
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -497,7 +497,7 @@
 			var/obj/item/clothing/gloves/G = gloves
 			gloves_siemens_coeff = G.siemens_coefficient
 			var/obj/item/clothing/gloves/color/fyellow/greytide = G
-			if(greytide)
+			if(istype(greytide))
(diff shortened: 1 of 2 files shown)


Marmio64:
honestly i think it would be better to have it only be one use, since budget insuls are pretty common

Tipy1802:
We don’t need a buff to greytiding and avoiding player interactions.

If someone needs insulated gloves to hack in places to steal shit so they wouldn’t have to interact with people and powergame let them order them from cargo or have them ask an engineer for a spare.

Moja10:
I think one would actually be plenty

Hopekz:
Why is one plenty when you are not required to hack a door 6 times (not including other machinery) per shift because of department wires?
What does your planned gameplay for us involve someone having 8 pairs of gloves in their backpack?

Hopekz:
> honestly i think it would be better to have it only be one use

> I think one would actually be plenty

Why is one plenty when you are now required to hack a door with different wires close to 6 times (not including other machinery) per shift because of department wires?
What does your planned gameplay for us involve someone having 8 pairs of gloves in their backpack?

Marmio64:
> > honestly i think it would be better to have it only be one use
> 
> > I think one would actually be plenty
> 
> Why is one plenty when you are now required to hack a door with different wires close to 6 times (not including other machinery) per shift because of department wires?
> What does your planned gameplay for us involve someone having 8 pairs of gloves in their backpack?

Because greytiders shouldn't have an easy 'job', and budget insuls are already common, not to mention the fact that about every 2 electrical toolboxes, there is an extra pair of budget insuls.

boodaliboo:
> We don’t need a buff to greytiding and avoiding player interactions.
> 
> If someone needs insulated gloves to hack in places to steal shit so they wouldn’t have to interact with people and powergame let them order them from cargo or have them ask an engineer for a spare.

but you can hack without gloves

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -502,6 +502,18 @@
 				to_chat(src, "<span class='notice'>You feel your heart beating again!</span>")
 	siemens_coeff *= physiology.siemens_coeff
 
+	if(gloves)
+		var/obj/item/clothing/G = gloves
+		if(istype(G, /obj/item/clothing/gloves/color/fyellow))
+			var/obj/item/clothing/gloves/color/fyellow/greytide = G
+			if(greytide.damaged == TRUE)
```
jcatjmeow:
```suggestion
			if(greytide.damaged)
```

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -502,6 +502,18 @@
...
+				qdel(G)
+				new/obj/effect/decal/cleanable/ash(src)
+				to_chat(src, "<span class='notice'>Your gloves absorb the shock and disintegrate!</span>")
+			else
+				greytide.damaged == TRUE
```
jcatjmeow:
```suggestion
				greytide.damaged = TRUE
```

jcatjmeow (changes requested):

SomeguyManperson (changes requested):
I don't think this actually stops the shock damage since the glove siemen coeff isn't changed and the glove check doesn't end the shocking

Reed0506:
i like my RNG based gloves that may or may not fuck me up even worse than wearing no glove at all, and gotta be honest, i wouldnt like this change just because oops i touched something now my gloves are gone. i want to keep my gloves regardless of how useless they are,

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -502,6 +502,18 @@
...
+		if(istype(G, /obj/item/clothing/gloves/color/fyellow))
+			var/obj/item/clothing/gloves/color/fyellow/greytide = G
+			if(greytide.damaged == TRUE)
+				qdel(G)
+				new/obj/effect/decal/cleanable/ash(src)
+				to_chat(src, "<span class='notice'>Your gloves absorb the shock and disintegrate!</span>")
+			else
+				greytide.damaged = TRUE
+				to_chat(src, "<span class='notice'>Your gloves absorb the shock!</span>")
+
```
JamieD1:
```suggestion
				to_chat(src, "<span class='notice'>Your gloves absorb the shock and disintegrate!</span>")
				return
			else
				greytide.damaged = TRUE
				to_chat(src, "<span class='notice'>Your gloves absorb the shock!</span>")
				return
```

SomeguyManperson:
make it so the gloves have their siemens coefficient is set to 0 and have the damage handled inside the normal glove effect stuff so it doesn't let them fully block stuff like tesla shocks or break from illusory shocks

JamieD1:
test

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -479,6 +479,9 @@
 			var/obj/item/clothing/gloves/G = gloves
 			if(G.siemens_coefficient <= 0)
 				total_coeff -= 0.5
+			var/obj/item/clothing/gloves/color/fyellow/greytide = G
```
Bibby0110, line 482: this is going to typecast ANY glove into greytide gloves and cause a runtime when using non-fyellow gloves isn't it?

if statement should be an istype check shouldn't it?

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -479,6 +479,9 @@
...
+			if(greytide)
```
Bibby0110:
```suggestion
			if(istype(greytide))
```

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -493,6 +496,9 @@
 		if(gloves)
 			var/obj/item/clothing/gloves/G = gloves
 			gloves_siemens_coeff = G.siemens_coefficient
+			var/obj/item/clothing/gloves/color/fyellow/greytide = G
+			if(greytide)
```
Bibby0110:
```suggestion
			if(istype(greytide))
```

Bibby0110 (changes requested):

PR #9957, opened 2020-10-12
Mining Now doesnt lose all atmos when there is a breach

JamieD1:
Vents now turn off, but pressure needs to be reintroduced before they turn back on

https://gyazo.com/2cb2bd1cea8321685059175a80ef26d6

#### Changelog

:cl:  
tweak: Vents now turn off, but pressure needs to be reintroduced before they turn back on
/:cl:


patpol4 (approved):
Jamie told me to approve this

SomeguyManperson:
this actually seems to break the mining airlock making it not repressurize

MegaEmpirical:
I believe Theos is correct.

PR #9958, opened 2020-10-12
Removes duplicate config entry that does nothing in its life

alexkar598:
Removes duplicate config entry that does nothing in its life

Its MALF not MALFUNCTION and MALF is alraedy there, lmao


Amelia0010 (approved):

Amelia0010:
@JamieD1

JamieD1 (approved):

PR #9968, opened 2020-10-13
removes the removal of patches and sprays not working on thickmaterial stuff because it apparently broke half the game

ghost:
you could easily have simply made it so that hardsuits and SPECIFICALLY hardsuits had a separate flag that disallowed for patch use and then had applications for patches check for that instead so that it didn't work on hardsuits but worked on everything else that required it

instead, as a result of this: golems, skeletons, and ninjas et cetera cannot heal, and ninjas can't heal at all without removing their entire suit. you have to use pills instead, and that's not intended behavior.

it's not worth breaking 2 species and 1 antag for a mostly irrelevant balance decision that was ALREADY RESOLVED by styptic powder and silver sulfadiazine gaining an overdose limit regardless to prevent abuse an consistent spam

and i should mention, this actually DOESN'T even fix the intended issue to start with, because you can bypass the entire problem by literally just splashing a beaker of styptic powder or silver sulfadizine or synthflesh ON YOURSELF in the hardsuit because while the PATCH won't apply and the SPRAY won't apply the beakersplash will

in short this was a lazy PR that broke several features that stayed in to deter 'powergame' when in reality most of the best antibrute chem mixes can be taken in pill form regardless and it could be bypassed with beaker splashing and instead simply broke an antagonist and two species (one of which of those species is used by two other separate antags, pirates and liches)

i understand you guys don't like powergame but if you're going to merge stuff like this at the very least don't half-ass it and instead do it properly, because a change that breaks half the game isn't worth merging

### DIFF
--- a/code/modules/reagents/reagent_containers/medspray.dm
+++ b/code/modules/reagents/reagent_containers/medspray.dm
@@ -32,14 +32,11 @@
 		amount_per_transfer_from_this = initial(amount_per_transfer_from_this)
 	to_chat(user, "<span class='notice'>You will now apply the medspray's contents in [squirt_mode ? "short bursts":"extended sprays"]. You'll now use [amount_per_transfer_from_this] units per use.</span>")
 
-/obj/item/reagent_containers/medspray/attack(mob/living/M, mob/user, def_zone)
+/obj/item/reagent_containers/medspray/attack(mob/M, mob/user, def_zone)
 	if(!reagents || !reagents.total_volume)
 		to_chat(user, "<span class='warning'>[src] is empty!</span>")
 		return
-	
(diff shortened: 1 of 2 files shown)


Reed0506:
i can confirm that the original PR was bad or something, and was exceptionally inconvenient except when in a hardsuit, where it didnt matter to me at all.

Moja10:
What is the original PR you are referencing?

JamieD1:
I shall be reworking this instead of reverting it with our comments in DM

JamieD1 (approved):

Hopekz:
Can we get a counter going for the amount of times people merge something thinking they'll fix it later and it never gets fixed?

PR #9970, opened 2020-10-13
Increases per topic limit due to genetics and tgui needing it.

JamieD1:
Increases per topic limit due to genetics and tgui needing it.

#### Changelog

:cl:  
tweak: Increases per topic limit due to genetics and tgui needing it.
/:cl:


boodaliboo:
does this fix  the problem of being unable to click in genetics because of to many clicks?

JamieD1:
> does this fix the problem of being unable to click in genetics because of to many clicks?

Yes Bood

Amelia0010 (approved):

PR #9974, opened 2020-10-14
Removes Xray from the genetics powers pool

Moja10:
Four simple reasons:
1. If any singular thing enables powergaming in a way that directly breaks the rules and causes massive issues it's this power
2. There's not a lot more unfun than to be a traitor who has literally no way of detecting whether or not he's been spotted by the 3 validhunting assistants who have alerted the validhunting horde about his antics he had no way of knowing where detected
3. Xray isn't a power that benefits the game at all by being mass distributed in any form to anyone. I have never seen a round improved by everyone running around with xray. 
4. Xray will still exist in the game, it's just behind the research protected xray implant, which goes completely unused because the xray power is vastly superior.

### DIFF
--- a/code/datums/mutations/sight.dm
+++ b/code/datums/mutations/sight.dm
@@ -57,15 +57,6 @@
 	REMOVE_TRAIT(owner, visionflag, GENETIC_MUTATION)
 	owner.update_sight()
 
-//X-ray Vision lets you see through walls.
-/datum/mutation/human/thermal/x_ray
-	name = "X Ray Vision"
-	desc = "A strange genome that allows the user to see between the spaces of walls." //actual x-ray would mean you'd constantly be blasting rads, wich might be fun for later //hmb
-	text_gain_indication = "<span class='notice'>The walls suddenly disappear!</span>"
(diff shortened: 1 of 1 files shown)


JamieD1:
Hmm

Mesalikepie:
I agree. This is way too powerful and easy to get and mass produce. Makes traitoring so hard at times with no real counter

Tipy1802:
I feel like making it only work for mindshielded personnel would have been better so sec could still use it but this is fine too

ghost:
Excellent PR. That being said you should simply make it unobtainable as a genetics power rather than entirely gone as a mutation, so it's still possible for wizards, vampires etc to get it by nature of their upgrades.

@Tipy1802

Idiotic fucking comment. Sec having it is the problem.

Tipy1802:
No the problem is non sec crew having it and using it as validhunting

Moja10:
Gonna make a different PR that doesn't nuke it from the code, too many issues that I don't want to fix trying to nuke it from the code, just gonna remove the ability for geneticists to get it

PR #9975, opened 2020-10-14
Removes Xray from genetics PR #2

Moja10:
Four simple reasons:

1. If any singular thing enables powergaming in a way that directly breaks the rules and causes massive issues it's this power
2. There's not a lot more unfun than to be a traitor who has literally no way of detecting whether or not he's been spotted by the 3 validhunting assistants who have alerted the validhunting horde about his antics he had no way of knowing where detected
3. Xray isn't a power that benefits the game at all by being mass distributed in any form to anyone. I have never seen a round improved by everyone running around with xray.
4. Xray will still exist in the game, it's just behind the research protected xray implant, which goes completely unused because the xray power is vastly superior.

This PR doesn't nuke it from code, it is just not possible by genetics, also this actually compiles. 

🆑
rscdel: Removed the ability for Geneticists to acquire Xray
/🆑

### DIFF
--- a/code/datums/mutations/_combined.dm
+++ b/code/datums/mutations/_combined.dm
@@ -17,10 +17,6 @@
 	required = "/datum/mutation/human/strong; /datum/mutation/human/radioactive"
 	result = HULK
 
-/datum/generecipe/x_ray
-	required = "/datum/mutation/human/thermal; /datum/mutation/human/radioactive"
-	result = /datum/mutation/human/thermal/x_ray
-
 /datum/generecipe/mindread
 	required = "/datum/mutation/human/antenna; /datum/mutation/human/paranoia"
 	result = MINDREAD
@@ -28,7 +24,7 @@
 /datum/generecipe/shock
 	required = "/datum/mutation/human/insulated; /datum/mutation/human/radioactive"
 	result = SHOCKTOUCH
-	
+
 /datum/generecipe/antiglow
 	required = "/datum/mutation/human/glow; /datum/mutation/human/void"
 	result = ANTIGLOWY


Nechronom:
Full agree. heat vision is more than enough for tracking people, and X-ray is still craftable with research.

ToGWtF:
Disagree with the reasons I put in adminchat.

Thermals is still just as good, as costs just the same. Better alternative would be to make Xray/Thermal use up more of your genetics stability so you cant stack the best 3 genetics powers at once.

Firewolf34:
> Thermals is still just as good, as costs just the same. Better alternative would be to make Xray/Thermal use up more of your genetics stability so you cant stack the best 3 genetics powers at once.

Why not both. Increase Thermals point cost from like 30 to 40 so that one can only use 2 genetics powers since it's essentially X-ray with less frills. *in addition* to the current change of removing gene X-ray to push the science tech tree into use at Robotics.

Ratvar:
Could these adminchat reasons be put here instead, I disagree with notion of it being hidden away.
I agree that thermals are just as good. So, why not remove thermals and crew obtainable xray-thermal implants as well - it **really** does not matter what amount of effort is put into getting those for two simple reasons:

1. If any singular thing enables powergaming in a way that directly breaks the rules and causes massive issues it's seeing people do stuff through walls

2. There's not a lot more unfun than to be a traitor who has literally no way of detecting whether or not he's been spotted by the 3 validhunting assistants who have alerted the validhunting horde about his antics he had no way of knowing where detected

jcatjmeow (approved):

Hopekz:
We should think about doing this with thermal vision as well.
While not the exact same it still encourages similar valid hunting you are trying to prevent with this PR.

MegaEmpirical:
So I am actually all for removing it as there is the research option and thermals. However, I will share some suggestions made by others when we were discussing this: Make it a toggled ability, have it slow you down when in use, have it cause eye damage when in use, and making it so you can't examine/see names. If all of that was thrown in I would say that would be good. I'd say we remove it until someone makes these kinds of changes though, if we don't it's far more likely that it'll not change whereas if we remove it, that basically annoys people into doing something.

Finally: Whatever the HCs think is best.

Ratvar:
Looking at proposed future changes above, they do not seem to adress any of the issues - you toggle it for a second in maint, shriek that cult stuff is here/someone stands with pda in hand still and game is over. It might make it harder to validhunt with in a minor way, but it still will be used to validhunt - and too annoying to use outside of validhunting, which seems bad.

MegaEmpirical:
That's fair, removal is likely the best option.

MitztheKat:
i'd like xray genetics to have damage of some sorts. as MegaImperial said, that is a good idea.
removing it outright seems weird. genetics is still inferior to a properly built nanite lab in terms of antagonist advantages. Yes you can get hulk, but with nanites you can still as traitor be bullet/lazerproof.

ToGWtF:
One of the ideas I had given was that Xray/Thermal from genetics would cause constant eye damage. Sure. The mutation is faster than surgery to get the xray implants, but you would need constant eye medicine or replacements after a while.

Amelia0010 (approved):

PR #9981, opened 2020-10-14
Fixes skillcapes

Takahiru:

I think i found the problem? For some reason I thought a /datum/job had to be passed into prefs.exp[] but another example uses the strings from the global positions so i replaced them with strings

#### Changelog
Fix skillcapes
:cl:  
bugfix: Taka messed up his skillcapes pr  
/:cl:


fluffe9911 (approved):
probably not the problem but worth a shot

PR #9986, opened 2020-10-15
Fixes the link pretty-filter, somewhat

Altoids1:
Jamie asked me to help him fix up a regex in the pretty filter that captures on bad links, but does not capture on links to Yogstation or Github. I've done so.

Here's an example of what the old regex did (and alarmingly did not!) capture on **(Warning: NSFW links used to test the regex!)**: https://regex101.com/r/ouQpdX/1

And now my new regex (Again, NSFW links used to test it): https://regex101.com/r/KwBR7K/2

## Changelog

:cl:  Altoids
bugfix: It is no longer possible to spam OOC with links to your favorite furry website.
/:cl:

### DIFF
--- a/config/pretty_filter.txt
+++ b/config/pretty_filter.txt
@@ -43,4 +43,4 @@ n[i1!\\\/]+g(?:g|l+[e3]+t)=BAN ME ADMINS!
 One day while Andy was masturbating=BAN ME ADMINS!
 XEzwgBD=BAN ME ADMINS!
 
-\b(?:https:\/\/)?(?:www\.)?(?!yogstation\.net|github\.com)\w{4,128}\.\w{3}\b\S*=[Link Removed]
+\b(?:https:\/\/)?(?:www\.)?(?!yogstation\.net|github\.com)\w{4,128}\.\w{2}\.?\w{0,2}\b\S*=[Link Removed]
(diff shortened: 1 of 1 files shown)


JamieD1 (approved):
Thanks Altoids

alexkar598:
Doesnt match for http and has false positives in the text one, the matching is somewhat better at the cost of being more aggressive.
Why not just set the minimum for the old regex to 4 instead of 5? Yes i am biased because I spent alot of time tuning that regex

Altoids1:
> Doesnt match for http and has false positives in the text one, the matching is somewhat better at the cost of being more aggressive.

I think it is more important to always capture on links that are bad, than it is to never capture on any false-positive text. The whole point is to mute links; a player which gets censored accidentally can just, retry their message.

> Why not just set the minimum for the old regex to 4 instead of 5? Yes i am biased because I spent alot of time tuning that regex

I didn't like the old regex. It takes longer than this new one, and has some bugs (for example, there's several non-escaped period and question-mark characters in it). I figured it was best to just make a new one, to minimize its size, help it run better, and help it capture better. Sorry about clobbering your work on the matter.

alexkar598:
People were bitching at me really hard due to false positives when I added it which is what led to the shitty TLD list

Amelia0010:
@yogstation13/head-coders one of you two merge it

PR #9987, opened 2020-10-15
doubles the speed of the AI fixer since its slow as balls

SomeguyManperson:
AIs can't even take toxin damage and most damage will just be brute so the distributed healing is garbo

:cl:  
tweak: AI repair consoles are now twice as fast
/:cl:


MegaEmpirical:
It's true.

SomeguyManperson:
for reference this takes the healing speed up from 1 point every ~2.5 seconds to 2 points every ~2.5 seconds so it takes only maybe 5 minutes instead of like 10 to recover an AI

Amelia0010 (approved):

JamieD1 (approved):

PR #9994, opened 2020-10-15
Cloning now requires biomass the electric boogaloo

redmoogle:
Original PR: https://github.com/yogstation13/Yogstation/pull/7712
Gonna work on this since it died out

You now must use a resource known as **Biomatter** to fuel cloning

You craft biomatter with 10u of blood and 5u of nutrients

**cloning costs 4 biomatter blocks.**
Gene mans get three clonings free, and 1/2 of the third cloning

The **harder you upgrade the pod the more biomatter you will need**, capping at 150 biomatter (6 blocks of it) with tier 4 upgrades.
This gives antags more options to make sure their target won't get cloned besides bombing the whole of medbay
This also makes geneticists have more responsibilities beyond powergaming and fixing people in eventual rad storm

You can get biomatter by grinding bodies in a new machine

Changelog

🆑Szyszkrzyneczka, Redmoogle
rscadd: Added Biomatter a food you can eat and it is used to fuel cloning / Adds body grinder to cloning that can be used to grind bodies into biomatter.
tweak: Now you need to put some effort to clone people
imageadd: Added my shitty biomatter icons (spriters feel free to point out that that biomatter is a coder sprite
/🆑


Tipy1802:
Upgrading the cloner shouldn’t make it worse. Its supposed to be an upgrade not a downgrade.

You might as well just remove the cloner entirely, it’s only use is for when there are too many bodies for the MDs to revive and this makes that impossible.

Doesn’t solve the (not so common) problem of MD just cloning a single guy even though they can revive him, far from it, it makes that the desired use of the cloner since mass cloning is no longer possible

redmoogle:
> 
> 
> Upgrading the cloner shouldn’t make it worse. Its supposed to be an upgrade not a downgrade.
> 
> You might as well just remove the cloner entirely, it’s only use is for when there are too many bodies for the MDs to revive and this makes that impossible.
> 
> Doesn’t solve the (not so common) problem of MD just cloning a single guy even though they can revive him, far from it, it makes that the desired use of the cloner since mass cloning is no longer possible

What if I added fancy knobs to adjust speed/cloneloss

redmoogle:
Imma use this as a base for another pr tommorow

MegaEmpirical:
I love it. I don't mind upgrade making it need more, but that knob/setting idea is on point if you're up for it.

PR #9996, opened 2020-10-15
17 year olds can no longer exist upon Nanotrasen Property, or in the entire universe for that matter

swissloaf:

Raises minimum age of a character to be 18, an intergalactically agreed age of an adult. Agreed on by both the Syndicate and Nanotrasen, so now you know even the Syndicate doesn't support child predators.

Also unironically why would you hire someone under 18, especially for a Head role

#### Changelog

:cl:  
tweak: Changes AGE_MIN from 17 to 18
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -277,7 +277,7 @@
 #define OFFSET_NECK "neck"
 
 //MINOR TWEAKS/MISC
-#define AGE_MIN				17	//youngest a character can be
+#define AGE_MIN				18	//youngest a character can be
 #define AGE_MAX				85	//oldest a character can be
 #define AGE_MINOR			21  //legal age of space drinking and smoking
 #define WIZARD_AGE_MIN		30	//youngest a wizard can be


TheLinuxOS:
I mean I see no problem with this, makes sense

boodaliboo:
so what was the point of making that age pr previously swiss

swissloaf:
> so what was the point of making that age pr previously swiss

still applies for those under 21

jcatjmeow:
how does this affect those who have characters at 17 when this is merged. I like it but we should make sure it don't break shit

Tipy1802:
To be fair an 18 years old captain is as realistic as a 17 year old on.
I would suggest bumping it to 25

Tipy1802:
To be fair for a 25 year old captain is also a bit too little but you can’t really push it further without getting an angry mob with pitchforks

Amelia0010:
>how does this affect those who have characters at 17 when this is merged. I like it but we should make sure it don't break shit

https://github.com/yogstation13/Yogstation/blob/bc4c8c274dbcacd92330f5ff3a6297717219a515/code/modules/client/preferences_savefile.dm#L443 should be fine

Amelia0010 (approved):

PR #9998, opened 2020-10-16
Enforces a minimum IC character age requirement for some jobs

Altoids1:
## Yogurt hates minors: Part 1
This is for a bounty: https://forums.yogstation.net/threads/5-age-requirements-for-certain-roles-ss13-server.21994/

This PR implements a system where, if a player's character is too young IC to have a particularly-assigned job, their age is bumped up to the absolute minimum possible.

image

### Treat for the HoP

As a treat, I have also made it so that the ID Computer will not allow its user to assign a job to a person who is too young to possibly have it, per "Nanotrasen guidelines." Of course, editing their age in the computer can get past this block, if necessary.

image
*The most recent version of this PR has the above text using the "warning" span, to make it more noticeable.*

## Changelog

:cl:  Altoids
tweak: It is no longer possible for the station to be Captained by a 17-year-old roundstart, nor for any doctor to be under the age of 24.
tweak: It is now against Nanotrasen policy for a Head of Personnel to assign Captainship to a minor.
/:cl:


Tipy1802:
Well done pr!

MegaEmpirical:
Shouldn't we just get rid of age in the system all together? Seems the easiest route and we really don't lose anything.

Tipy1802:
Nah it’s nice flavour

yogstation/code/modules/jobs/job_types/lawyer.dm
```diff
@@ -0,0 +1,2 @@
+/datum/job/lawyer
+	minimal_character_age = 24
```
Amelia0010, line 2: You need to be older to be a lawyer than to be the head of security or head of personnel?

FatalEYES:
Joke's on you, I'll just edit my vars to be a minor.

code/game/machinery/computer/card.dm
```diff
@@ -413,7 +413,11 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 							updateUsrDialog()
 							break
 					if(!jobdatum)
-						to_chat(usr, "<span class='error'>No log exists for this job.</span>")
+						to_chat(usr, "<span class='warning'>No log exists for this job.</span>")
+						updateUsrDialog()
+						return
+					if(!isnull(modify.registered_age) && modify.registered_age < jobdatum.minimal_character_age)
+						to_chat(usr, "<span class='warning'>This individual is too young to hold that Job, per Nanotrasen guidelines. Job assignment aborted!</span>")
```
Amelia0010, line 420: I'd rather this would alert the user and ask them to confirm to do it in spite of their age limit, so you don't *need* to go and falsify records to change the job if need be.

Amelia0010 (dismissed):

Amelia0010:
Also this is just my opinion, but I'd rather it just blocks your preferences so you can't select HoS/Captain/whatever instead of changing your age, but idk

Reed0506:
i feel like a declaration of war should disable the age lock, as it would make life a huge pain.

code/game/machinery/computer/card.dm
Altoids1: That'd certainly be the more passive approach, and if you look that is genuinely how I implemented it earlier.

However, I think having this instead kinda plays well into the idea of the HoP being unnecessarily bureaucratic, actually providing a gameplay element that encourages that. Plus this would be pretty much the sole and only time in the entire video game where one's IC age has any mechanical effect. A mechanical effect that you can override by altering an integer record.

That's why I changed it from being like as you've suggested towards this.

yogstation/code/modules/jobs/job_types/lawyer.dm
Altoids1: I was trying to make these values as low as feasibly possible. They don't have to be *that* high to make IC age not immersion-breaking .

yogstation/code/modules/jobs/job_types/head_of_security.dm
```diff
@@ -1,5 +1,6 @@
 /datum/job/hos
 	exp_requirements = 720
+	minimal_character_age = 22
```
JamieD1:
```suggestion
	minimal_character_age = 28
```

yogstation/code/modules/jobs/job_types/head_of_personnel.dm
```diff
@@ -1,2 +1,3 @@
 /datum/job/hop
-	exp_requirements = 720
\ No newline at end of file
+	exp_requirements = 720
+	minimal_character_age = 22
```
JamieD1:
```suggestion
	minimal_character_age = 25
```

JamieD1 (changes requested):
Go further.

JamieD1 (approved):

PR #9999, opened 2020-10-16
you now still get your skillcape if it can't be equipped

Amelia0010:
Miners get bodycams to occupy the same slot


Moggram:
So the photographer quirk adds a camera to the neck slot. You may want to test them together and see what happens.

JamieD1 (approved):

PR #10008, opened 2020-10-16
Makes it so that heads of staff can't get ANY roundstart antag.

MrHorizons:
Closes #10006.

Impressive speed, Adam, but we need to go deeper.

#### Changelog

:cl:  
rscadd: Heads of staff can now no longer roll blood brothers, blood cult, changeling, clock cult, heretic, hivemind (just in case), or traitor.
/:cl:

PR #10006, opened 2020-10-16
Makes it so that heads of staff cannot be traitors anymore.

ghost:
Because you're all shit.

PR #10009, opened 2020-10-16
makes it so heads of staff can't be roundstart antag

SomeguyManperson:
closes #10008
heads are supposed to be trustworthy and have access to the resources of an entire department, making them valuable in group antagonists and can basically be easy mode in solo antagonist roles. Converting a head of staff should be something risky for how much reward there is to it, not something attained at roundstart

keep in mind this doesn't mean you won't roll antag if you are trying to roll head roles, you'll just get assigned another job
:cl:  
tweak: heads of staff can no longer be antagonists
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/changeling.dm
+++ b/code/game/gamemodes/changeling/changeling.dm
@@ -12,7 +12,7 @@ GLOBAL_VAR(changeling_team_objective_type) //If this is not null, we hand our th
 	antag_flag = ROLE_CHANGELING
 	false_report_weight = 10
 	restricted_jobs = list("AI", "Cyborg")
-	protected_jobs = list("Security Officer", "Warden", "Detective", "Head of Security", "Captain", "Head of Personnel") //YOGS - added hop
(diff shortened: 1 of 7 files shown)


MrHorizons:
Closes #10006 and #10008.

patpol4:
Best solution imo

jcatjmeow (approved):

MegaEmpirical:
As I said on the other: I've mentioned either making them less likely to be an antag and/or put a floating cap on how many can be an antag in a round. But I can get behind just making them not antags at all honesty. I always thought it felt a little weird, especially when nearly every fucking head is an antag

bowlofcereal:
This is literally the nuclear solution to an uproar caused by a head spawnkilling another head in the conference room.

This could've been easily solved by bwoinking the spawnkiller for 0.0 and resolved, end of story.

This seems more like a rule thing and not a "reee delete this" situation, but you do you.

MegaEmpirical:
It's something I've wanted for years regardless of what happened today

PR #10013, opened 2020-10-16
no head changelings

SomeguyManperson:
:cl:  
rscdel: heads of staff can no longer roll changeling 
/:cl:


JamieD1:
Noppers

JamieD1:
RP Wise this makes sense due to them being able to take the identity before coming on station, yes you could say this about cultists as well but they can be converted easy enough.

jcatjmeow:
Ok but gameplay wise tho

PR #10019, opened 2020-10-17
makes people shout a link to a fucking wikihow article on how to change a diaper while defibbing because you fucking infants are so incompetent you apparently need someone to code in shouting 'clear' while defibbing someone so nobody gets electrocuted

ghost:


Altoids1:
Never change, Adam.

PR #10024, opened 2020-10-17
Buffs Antenna Mutation

redmoogle:

Antenna now just functions like a radio with cap level channels

And is now 15 instability

#### Changelog

:cl:   
tweak: Antenna Mutation now can listen to all channels
tweak: Antenna Mutation now cost 15 stability up from 5
/:cl:

### DIFF
--- a/code/datums/mutations/antenna.dm
+++ b/code/datums/mutations/antenna.dm
@@ -4,31 +4,21 @@
 	quality = POSITIVE
 	text_gain_indication = "<span class='notice'>You feel an antenna sprout from your forehead.</span>"
 	text_lose_indication = "<span class='notice'>Your antenna shrinks back down.</span>"
-	instability = 5
+	instability = 15
 	difficulty = 8
-	var/obj/item/implant/radio/antenna/linked_radio
-
-/obj/item/implant/radio/antenna
-	name = "internal antenna organ"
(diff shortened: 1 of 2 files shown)


Tipy1802:
While it does make it useful I don't think non sec and non command people should have access to the security and command channel

redmoogle:
> 
> 
> While it does make it useful I don't think non sec and non command people should have access to the security and command channel

Out of most mutations, this one is pretty mundane, as oh wow you can snoop in on the cap ig, or see that sec is on your tail.

TheLinuxOS:
Makes a mutation that is the equivalent of a 2 TC item, and seeing as we still have the thermal vision mutation which is also equivalent to a traitor item I personally see no problem with it.

MegaEmpirical:
I wouldn't fight this if we had the old encryption system for tcomms that would disable those channels when they're encrypted.

Ratvar:
TheLinusOX, see no reason why existing mutations would be a reason to add new ones without explaining why current mutations are good, or how they are related in utility instead of cost: thermal is a validhunting tool or very rarely an overpowered antagging tool, bringing almost nothing good to the round. 
It is rather hard to use sec or cap radio to validhunt, thought, so this seems to be a good thing.

JamieD1 (approved):
Why not, lets see how it works out

PR #10029, opened 2020-10-17
Fucking Bood -  Also gives tablets to rest of command because why not

JamieD1:
Fucking Bood -  Also gives tablets to rest of command because why not

#### Changelog

:cl:  
rscadd: Gives tablets to rest of command because why not
/:cl:


Tipy1802:
Didn't they already get them? I know for a fact the HoP had one

PR #10048, opened 2020-10-18
wizards are technophobes now

Amelia0010:
Makes technophobia a trait instead of being a check for the ashwalker

Gives wizards technophobia

I'm not really that attached to this, it's just something that popped into my head when I was reading the goon wiki and it said in the gamemode description that wizards were technophobic.

:cl:
tweak: Wizards now can't use machinery
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -96,6 +96,7 @@
 #define TRAIT_NODISMEMBER		"dismember_immunity"
 #define TRAIT_NOFIRE			"nonflammable"
 #define TRAIT_NOGUNS			"no_guns"
+#define TRAIT_TECHNOPHOBE		"technophobe" // Not allowed to use technology
 #define TRAIT_NOINTERACT		"no_interact" //Not allowed to touch anything (even with TK) or use things in hand
 #define TRAIT_NOHUNGER			"no_hunger"
 #define TRAIT_EASYDISMEMBER		"easy_dismember"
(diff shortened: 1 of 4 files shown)


boodaliboo:
intrested to see how this will work ingame

Reed0506:
i feel like locking wizards out of anything tied to machines may be a bit fucked in practice but maybe not

MegaEmpirical:
I actually like this. Any concerns with gun summon spell?

MegaEmpirical:
Also, they may need a way to call the shuttle.

Marmio64:
kind of fucks with stealth wizards, but fuck stealth wizards anyways.

also, it does mean they can't recall shuttle, which is both good and bad.

Hopekz:
Are wizards now not able to use the consoles they have in their wizard spawn? 
They have advanced camera consoles that help them see where they want to teleport to. 

In the current implementation, this looks like an oversight so either make the consoles up there bypass this or remove the consoles to avoid confusion.

UselessTheremin:
STUPID FUCKING CONSOLE!! EI NATH!! 
i like this, will make for funny dynamics

Jarod1200:
This means wizards can't call or recall the shuttle and cant buy things from vendors. Not really a fan.

jcatjmeow:
The cant buy things from vendors is the big thing because it means **they cant use the wizard vendors to choose their outfit or anything**

Technophobia being a trait is fine tho

MegaEmpirical:
Vendors can be replaced. More concerned about shuttle.

Amelia0010:
image

PR #10033, opened 2020-10-17
Map and Flare are now a preference toggle

JamieD1:
Map and Flare are now a preference toggle

#### Changelog

:cl:  
rscadd: Map and Flare are now a preference toggle
/:cl:

PR #10049, opened 2020-10-18
maps now burn

Amelia0010:
It's not just flares but they'll work too

:cl:
tweak: Nanotrasen would like to remind the personnel that lighting your emergency flare and using it to burn your emergency map is against standard operations procedure
/:cl:

### DIFF
--- a/code/modules/mapping/minimap.dm
+++ b/code/modules/mapping/minimap.dm
@@ -79,6 +79,24 @@
 	max_integrity = 50
 	var/list/minimaps = list()
 
+/obj/item/paper/attackby(obj/item/P, mob/living/carbon/human/user, params)
+	..()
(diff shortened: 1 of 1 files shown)


Firewolf34:
Can you make the map be crumple-able into the paper ball scrap item because that's something I've always wanted to do.

That way you can throw it into the disposals bin instead of the thing literally flying half a tile in it's normal state (paper has realistically low throw distance)

Hopekz:
Your PR can co-exist with #10033 
Your PR is more of an extra feature to the item vs Jamie's PR of offering players the option not to bring them if they do not want those items.

Amelia0010:
yes but I don't like it

JamieD1 (approved):

PR #10052, opened 2020-10-19
Buffs Hecksuit

Marmio64:
Makes hecksuit's laser and energy resistance about equal to the ash drake suit, so it isn't a downgrade. Also increases it's melee armor by 5, so it can compete more with the ash drake suit, since it can only be obtained by fighting, arguably, the hardest megafauna in the game.

edit: made laser and energy resistance higher, about the same as ash drake suit, with 10 less laser however.

#### Changelog

:cl:  
tweak: Makes hecksuit melee, bullet, laser, energy resistances be 75/40/40/40 respectively. 
/:cl:

### DIFF
--- a/code/modules/mining/equipment/explorer_gear.dm
+++ b/code/modules/mining/equipment/explorer_gear.dm
@@ -94,7 +94,7 @@
 	w_class = WEIGHT_CLASS_NORMAL
 	max_heat_protection_temperature = FIRE_IMMUNITY_MAX_TEMP_PROTECT
 	clothing_flags = STOPSPRESSUREDAMAGE | THICKMATERIAL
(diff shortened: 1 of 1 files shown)


MrHorizons:
I think its stats should be on par with, if not slightly better than, the ERT hardsuits from the necropolis tendrils. Probably with spaceproofing, too.

LazennG:
it has spaceproofing

Marmio64:
> I think its stats should be on par with, if not slightly better than, the ERT hardsuits from the necropolis tendrils. Probably with spaceproofing, too.

heck suit already has space proofing, but i'll increase energy and laser resistance even more, since that is the main difference between the hardsuits.

JamieD1 (approved):

PR #10054, opened 2020-10-19
Gives All Execution Variants Reinforced Walls (Boxstation)

alao3alao3:
This PR adds reinforced walls to all execution variants. This is inline with every outer area of Sec having reinforced walls, and it's honestly pretty easy to break in there and rescue your BB's corpse that they forgot to morgue once they execute them.

Should you have never seen the execution room before, this is what has changed:
bpng
cpng

:cl:  
rscadd: Added Reinforced walls to all Boxstation Execution variants.  
/:cl:

### DIFF
--- a/_maps/RandomRuins/StationRuins/BoxStation/transfer1.dmm
+++ b/_maps/RandomRuins/StationRuins/BoxStation/transfer1.dmm
@@ -293,9 +293,6 @@
 	},
 /turf/open/floor/plasteel/dark,
 /area/security/execution/transfer)
-"M" = (
-/turf/closed/wall,
-/area/security/execution/transfer)
 "N" = (
 /obj/machinery/power/apc{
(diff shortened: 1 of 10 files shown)


patpol4:
No need to make this so secure imo, allows for more antag play this way.

alao3alao3:
nah, rest of sec is as secure as this, no reason to shoehorn in an achilles heel. The inner layer of reinforced walls could be removed if that makes it better.

boodaliboo:
> nah, rest of sec is as secure as this, no reason to shoehorn in an achilles heel. The inner layer of reinforced walls could be removed if that makes it better.

it does

alao3alao3:
bpng
heyo

patpol4 (approved):
I don't really agree with this but votes say otherwise. 
I like how you moved the cigarette butt.

alao3alao3:
The system works, people.

PR #10059, opened 2020-10-19
Pepperspray is a lot more agonizing

Marmio64:
The title

Condensed capsaicin, and therefor pepperspray and tear gas grenades, is much more agonizing and painful. It blurs and blinds your eyes for longer, makes you scream more often, gives you a longer confused timer, and also causes stamina damage now. Tear gas grenades also make you cough more often. I've tested it, and it seems fairly balanced.  You can still almost completely nullify the effects of pepperspray with just eyeglasses, while glasses and a mask completely remove all it's effects. Having a mask on will still somewhat reduce the effects, although not as much as it used to.

#### Changelog

:cl:  
rscadd: Pepperspray stuns, confuses, and blinds you for longer, among other things.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/food_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/food_reagents.dm
@@ -294,10 +294,10 @@
 			if(prob(30))
 				victim.emote("scream")
 			victim.blur_eyes(12)
-			victim.blind_eyes(6)
+			victim.blind_eyes(8)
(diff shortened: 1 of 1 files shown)


Tipy1802:
Good pepper spray is almost never used since it's just a worse flash

jcatjmeow (approved):
> makes you scream more often

👍

JamieD1 (approved):

JamieD1:
Fucking Marmio

PR #10067, opened 2020-10-19
Fixes itching powder to no longer be classified as a toxin in the game's code as to avoid throwing beakers of 100u itching powder at someone to cause immediate liver failure

ghost:
Due to what I am hoping is an oversight, itching powder was set as a toxin.

It's also 100% permeable through clothing of any kind and conveys all its contents in a touch reaction.

As a result, you could throw a beaker of 100u itching powder at someone to cause guaranteed liver failure due to its status as a toxin.

Given that itching powder is supposed to be a harmless prank chemical for the most part, this seems like an oversight.

Note that it still maintains its old behavior-you can still throw 100u of it at someone in a beaker, and it could still cause problems for them with the histamine. It just no longer serves as a guaranteed liver death, since hitting someone in a full hardsuit with the FUNNY BEAKER and watching them drop dead 3 minutes later with nothing they could do about it was kinda memey.
#### Changelog

:cl:  
tweak: Itching powder is now just a reagent, not a toxin. Nothing has changed.
/:cl:


ghost:
it sounds funny but it's actually kind of lame to just chuck a beaker of itching powder at like a shadowling or nukeop in an elite hardsuit and watch them die of fucking liver failure 3 minutes later

ToGWtF:
can confirm this happens

JamieD1 (approved):

PR #10069, opened 2020-10-19
fix stunbatons stunning if they are caught

SomeguyManperson:
theos is mad
:cl:  
bugfix: stunbatons won't stun you if you catch them
/:cl:

### DIFF
--- a/code/game/objects/items/stunbaton.dm
+++ b/code/game/objects/items/stunbaton.dm
@@ -41,7 +41,8 @@
 	update_icon()
 
 /obj/item/melee/baton/throw_impact(atom/hit_atom, datum/thrownthing/throwingdatum)
-	..()
+	if(..())
+		return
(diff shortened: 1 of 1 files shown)


patpol4:
nice

MegaEmpirical:
I'd suggest making it a 50% chance or something, as you could fuck it up and throwing the baton has a really low chance of stunning you too (sad).

SomeguyManperson:
the chance of a stunbaton stunning you on hit is 35% it's not really small

MegaEmpirical:
I'd agree if you were throwing like 5 of them, but for only one baton I will say that's low which will then be in the hands of the target more likely than not.

alexkar598:
Thrown batons shouldnt stun at all but thats just me i guess

JamieD1 (approved):

Issue #4788, opened 2019-03-17
Detective lacks department lathe access


MegaEmpirical:
Detective cannot reach the department lathe since he doesn't have sec office access.

fluffe9911:
*shrug*

patpol4:
Not a bug, detective is not a security officer, hence no access to the security office.

MegaEmpirical:
No shit but he is a memeber of the security department and lacks access to his department lathe so it is an issue.

patpol4:
Alright but this is not a mapping issue, det just needs sec office access

PR #10070, opened 2020-10-19
Moves the seclathe into the armory

Partheo:
The seclathe can print guns, guns should be located in the armory, not the security office.

#### Changelog

:cl:  
tweak: The seclathe is now located in the armory
/:cl:

### DIFF
--- a/_maps/map_files/EclipseStation/EclipseStation.dmm
+++ b/_maps/map_files/EclipseStation/EclipseStation.dmm
@@ -3489,8 +3489,6 @@
 /turf/open/floor/plasteel/dark,
 /area/ai_monitored/security/armory)
 "ahu" = (
-/obj/structure/table,
-/obj/machinery/recharger,
 /obj/effect/turf_decal/tile/neutral,
 /obj/effect/turf_decal/tile/neutral{
 	dir = 1
@@ -3501,6 +3499,7 @@
 /obj/effect/turf_decal/tile/neutral{
 	dir = 8
 	},
+/obj/machinery/rnd/production/techfab/department/security,
 /turf/open/floor/plasteel/dark,
 /area/ai_monitored/security/armory)
 "ahv" = (
@@ -3699,7 +3698,9 @@
 	icon_state = "0-8"
 	},
 /obj/structure/table,
-/obj/machinery/recharger,
+/obj/machinery/recharger{
+	pixel_x = -6
+	},
 /obj/effect/turf_decal/tile/neutral,
 /obj/effect/turf_decal/tile/neutral{
 	dir = 1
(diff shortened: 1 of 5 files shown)


patpol4:
Please explain anyway, also fix conflict

MegaEmpirical:
Hmm well if you're doing that, then close this:
https://github.com/yogstation13/Yogstation/issues/4788

athena413:
thank you for applying this to all maps

ghost:
as i'm the one who asked parth to do it because I didn't want to use the mapmaker, @wejengin2 

the explanation is such

seclathe prints guns

warden is supposed to distribut eguns

moving it to the armory means that the warden now distributes guns rather than sec officers printing 10 AEGs and lodging them up their assholes

Mesalikepie:
Okay people the seclathe isn't just for guns, but also has ebolas and other equipment that many people use it for. Like seriously maybe just split the sec techfab into two. A weapon printer, and an equipment printer

PrinceKirze:
I agree with the sentiment behind this, but not really the execution. I don't really see a reason why not to split weapons printing and equipment printing into two different machines. A little more work, certainly, but it's the long term solution to this. The lathe also prints ebolas, detective's scanners- tons of things that you don't need to go through the warden to get. On most population levels I think the proposed change would work fine, but during really lowpop and really highpop I see at least a few edge cases where this becomes a hurdle.

Tipy1802:
I would suggest warden office instead, it would still prevent officers from printing guns but it would make it easier for wardens to give non weapons (like body cameras) to sec (less walking)

UselessTheremin:
good change for when there is a good warden, makes the game better. bad change and a hurdle for the whole security team when there is a bad warden. in my experience, 9 times out of 10, there is a bad warden. i dont want to have to hear "go get a stamp of approval from hos and captain" for an ebola or a bodycam.

MegaEmpirical:
I'm generally okay with this but I do think putting it into the Warden office is a much better idea.

It'll be be less walking for the warden who may have to use this often. It'll be a less dangerous area for someone to need to break into or ask the AI for access too.

patpol4 (changes requested):
Let's go for the compromise of putting it in the warden office and fix conflicts

Georgeur:
I'm laughing but sure. Can't wait to see the Warden commit dead after being asked for the 15th time to lathe a new pair of handcuffs, bodycam, ebola, etc. Then sec end up going to cargo to lathe ammo. The real fix here is to remove it's ability to create firing pins, jesus christ.

patpol4:
@Partheo is this going to be worked on?

Reed0506:
oh boy this makes it it suck even more for the det to get ammo on box

patpol4:
Closing due to inactivity, message maintainers on discord to have it reopened

PR #10075, opened 2020-10-20
Tracks mentor activity via role playtime

alexkar598:
Pog, no reason for this not to work but it is untested and webedit.

# *I NEED MY GRAPHS.*


TheLinuxOS:
Oh god, he will realize it is all assistant, shaftminer, and atmos tech

Tipy1802:
Now this is what a true role play positive PR looks like

JamieD1 (approved):

PR #10076, opened 2020-10-20
poppy the safety possum is resistant to lasers on the SM engine

UselessTheremin:
title, it hurts my soul to see poppy get fuggen blasted every round. i may have done a shitcode moment, just let me know. i wanted to move poppy, but there was no better area so i decided to go this route. plus, its funny to see a possum be completely laser proof.
🆑
tweak: poppy is immune to lasers in sm engine
/🆑

### DIFF
--- a/_maps/RandomRuins/StationRuins/BoxStation/engine_sm.dmm
+++ b/_maps/RandomRuins/StationRuins/BoxStation/engine_sm.dmm
@@ -2026,8 +2026,10 @@
 	},
 /turf/open/floor/engine,
 /area/engine/engineering)
-"Jj" = (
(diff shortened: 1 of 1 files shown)


alexkar598:
inb possum is used in combat to bypass lasers

patpol4:
Poppy the safety possum dying to improper safety procedures is part of the joke. Also I'd rather not have laser resistant mobs running around due to what alexkar said and since it doesn't make much sense for a possum.

TheLinuxOS:
I agree with what wej said, the fact that the possum always died due to being unsafe (Im fairly certain it spawns in front of the emitters) is part of the joke.

UselessTheremin:
Fair enough.

PR #10077, opened 2020-10-20
Removes the makeshift laser rifle's requirement for makeshift weapons book because it's garbage

SomeguyManperson:
10 5 damage shots or 1 17 damage shot 👏 

:cl:  
tweak: everyone can now craft makeshift laser rifles, not like you'd want to
/:cl:

### DIFF
--- a/code/datums/components/crafting/recipes.dm
+++ b/code/datums/components/crafting/recipes.dm
@@ -50,7 +50,6 @@
 	time = 120
 	category = CAT_WEAPONRY
 	subcategory = CAT_WEAPON
-	always_availible = FALSE
 
 /datum/crafting_recipe/knifeboxing
 	name = "Knife-boxing Gloves"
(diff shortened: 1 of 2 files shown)


jcatjmeow (approved):
Not a fan of this but I tested it and it works

Hopekz:
> Not a fan of this but I tested it and it works

Why did you need to fully test this 2 line change outside of reading it to make sure it does what is described?

MegaEmpirical:
Because the right thing is to always test.

Marmio64:
to be honest i think makeshift lasergun should be buffed or something instead of making it craftable by anyone, but this is a good fix too i guess.

MegaEmpirical:
Under what conditions was this available before?

SomeguyManperson:
Have to buy a book for a few tc as traitir to craft it and othed stuff

Amelia0010 (approved):

boodaliboo:
> Under what conditions was this available before?

be aware the book is infinite use

JamieD1 (approved):

PR #10086, opened 2020-10-21
Fixes the banner so it shows above the server name not in the middle

alexkar598:
We look like the clowns in the banner
img


swissloaf:
Who even uses the website to launch lmao

Amelia0010 (approved):

PR #10090, opened 2020-10-21
Adds the Sleepy Virus

Marmio64:
Adds a new virus that I made. It has no possibility of dealing damage, and it's most common effects are blurring your eyes and giving you slight amounts of stamina damage, with a very rare chance of making you sleep for a bit. Very similar to narcolepsy but a disease instead of a symptom. I tested it and it seemed pretty good, so hopefully it isn't unbalanced or TOO annoying. It is really easy to cure anyways, with spaceacillin. The name is pretty lame, so if anyone has a better idea I'll change it. Should be spawnable from the random disease event and also has a bottle form that should have a chance to be obtainable from purchasing the virus crate from cargo.

#### Changelog

:cl:  
rscadd: Adds a new virus: The Sleepy Virus 
/:cl:

### DIFF
--- a/code/datums/diseases/sleepy.dm
+++ b/code/datums/diseases/sleepy.dm
@@ -1,0 +1,62 @@
+/datum/disease/sleepy
+	name = "Sleepy Virus"
+	max_stages = 3
+	spread_text = "Airborne"
+	cure_text = "Spaceacillin"
+	cures = list(/datum/reagent/medicine/spaceacillin)
+	cure_chance = 10
+	agent = "SLPY Viron"
+	viable_mobtypes = list(/mob/living/carbon/human, /mob/living/carbon/monkey)
+	permeability_mod = 0.75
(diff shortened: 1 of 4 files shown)


jcatjmeow:
What justifies this virus's existence? Why did you make this, basically, what was your goal

Marmio64:
> What justifies this virus's existence? Why did you make this, basically, what was your goal

I made it to be a annyoying non lethal virus like the cold. If it doesn't fit in I don't care much if it is removed and i'd understand too since it doesn't do anything important.

ghost:
Why not just use the narcolepsy symptom?

Marmio64:
> Why not just use the narcolepsy symptom?

i wanted to make my own thing. Narcolepsy is much more annoying since it constantly is making you fall asleep, mine is much less of an annoyance.

Ratvar:
Snooze Virus?

JamieD1 (approved):

PR #10091, opened 2020-10-21
walls are harder to slice through

Marmio64:
Walls now take twice as long to slice through and are a little bit more durable to hulk punches. It used to take 8 seconds to cut through a wall, now it takes 16 seconds. Feel free to say if it feels like too much. This also makes diamond and alien walls take 32 seconds to cut through, since both are normally twice as strong as normal walls, so they were buffed in comparison.

#### Changelog

:cl:  
tweak: Walls are tougher
/:cl:

### DIFF
--- a/code/game/turfs/simulated/wall/mineral_walls.dm
+++ b/code/game/turfs/simulated/wall/mineral_walls.dm
@@ -30,7 +30,7 @@
 	icon = 'icons/turf/walls/diamond_wall.dmi'
 	icon_state = "diamond"
 	sheet_type = /obj/item/stack/sheet/mineral/diamond
-	slicing_duration = 200   //diamond wall takes twice as much time to slice
+	slicing_duration = 400   //diamond wall takes twice as much time to slice
 	explosion_block = 3
 	canSmoothWith = list(/turf/closed/wall/mineral/diamond, /obj/structure/falsewall/diamond)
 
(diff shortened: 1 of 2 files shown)


GrimGodhead:
im getting real tired of gits that make it harder and harder to move around

Takahiru:
reinforced windows were buffed because they were reinforced, why would we need to buff regular walls

Marmio64:
> reinforced windows were buffed because they were reinforced, why would we need to buff regular walls

because right now for some reason it is far easier to take down a solid chunk of metal than glass with some metal rods in it

boodaliboo:
> > reinforced windows were buffed because they were reinforced, why would we need to buff regular walls
> 
> because right now for some reason it is far easier to take down a solid chunk of metal than glass with some metal rods in it

and? the reinforced wall buff was stupid to

Marmio64:
> > > reinforced windows were buffed because they were reinforced, why would we need to buff regular walls
> > 
> > 
> > because right now for some reason it is far easier to take down a solid chunk of metal than glass with some metal rods in it
> 
> and? the reinforced wall buff was stupid to

if you mean reinforced window, it was stupid but there was apparently a numbers problem that made the r-window way tougher than it was supposed to be, so it should be fine now. RCDs also will still take down walls easily, and if you get alien research you can get an abductor welder that will still take down a wall in about a second.

JamieD1 (approved):

PR #10100, opened 2020-10-22
all windows besides normal ones take time to build

Marmio64:
All the windows besides the super weak trash normal ones now take about a second and a half to build, since they're all rather strong, especially reinforced ones and plasma variants. I could also make it so the trash normal ones take time to build if people prefer, but they're really bad.

#### Changelog

:cl:  
rscadd: Takes time to build most windows
/:cl:


jcatjmeow:
This is 10 deciseconds, so 1 second.

JamieD1 (approved):

PR #10101, opened 2020-10-22
clown bug no longer duplicates

boodaliboo:
a alternative to https://github.com/yogstation13/Yogstation/pull/10098
why remove it if you can just remove the issue
#### Changelog

:cl:  
tweak: clown cockroaches no longer duplicate on death
/:cl:


fluffe9911:
the clown bug respawning on death is the whole point of it otherwise its just a retextured cockroach
edit: the problem aint so much it respawns on death but rather a interaction between stabalized gold and pets where it tries to respawn the pet leading to it respawning the clown bug while its clone still exists

PR #10105, opened 2020-10-22
makes goliath ranged cooldown longer

boodaliboo:
image

i dunno i guess im doing lavaland changes right now
#### Changelog

:cl:  
tweak: tweaked goliath cooldown
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/goliath.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/goliath.dm
@@ -12,7 +12,7 @@
 	mouse_opacity = MOUSE_OPACITY_ICON
 	move_to_delay = 40
 	ranged = 1
-	ranged_cooldown_time = 120
+	ranged_cooldown_time = 150
 	friendly = "wails at"
 	speak_emote = list("bellows")
 	vision_range = 4


jcatjmeow (changes requested):
This changes the delay between tentacle attacks.
If your intention is to change the movement you need to change move_to_delay
If you want to make the chomps happen less you gotta change the value rapid_melee although you should arguably just reduce the damage at that point

Moggram:
I feel like this will make the fight longer, as you can get in two hits in when it is doing the ranged attack. But it will be easier for the unrobust who can't time it right and have to wait to dodge the tentacle.

boodaliboo:
> This changes the delay between tentacle attacks.
> If your intention is to change the movement you need to change move_to_delay
> If you want to make the chomps happen less you gotta change the value rapid_melee although you should arguably just reduce the damage at that point

my intention was to change the tentacle attacks

Mesalikepie:
But why though

Hopekz:
Why are you quoting me on your PR when you didn't even do the thing in the quote 

image

boodaliboo:
> Why are you quoting me on your PR when you didn't even do the thing in the quote
> 
> image
> 
> 

i misunderstood what you said but i still wanted this change

SomeguyManperson:
don't see any reason to do this

PR #10109, opened 2020-10-22
Revert "Gives us a bit of memory wiggle room"

boodaliboo:
we are having multiple rounds in a row where there is no space ruin spawners at all

this simply changes the ammount of space ruins that spawn back to what it originally was


JamieD1:
Sure why not, lets see our wiggle room

JamieD1 (approved):

PR #10119, opened 2020-10-23
gives ashwalkers a bamboo seed

boodaliboo:
gives ashwalkers more variety in possible weapons allowing them to make blowguns and spears
#### Changelog

:cl:  
rscadd: Added bamboo seeds to ashwalker ruin
/:cl:


Marmio64:
where would they get syringes/darts from? I don't think they're craftable

boodaliboo:
> where would they get syringes/darts from? I don't think they're craftable

they can make 5 unit syringes with bamboo

patpol4 (approved):
Map technically fine, no idea about the balance implications of this, what is the goal here?

boodaliboo:
> Map technically fine, no idea about the balance implications of this, what is the goal here?

make them more tribal and allowing them to do tribal things like blowguns and stuff

PR #10120, opened 2020-10-23
10x5 Ballpit Random Maint

alao3alao3:
Ballpit, now with less maint corpses.
apng
No more playing minewsweeper for an hour and only getting like 3 tiles.

:cl:  
rscadd: maint ballpit
/:cl:

### DIFF
--- a/10x5_ballpit.dmm
+++ b/10x5_ballpit.dmm
@@ -1,0 +1,156 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"d" = (
+/obj/effect/decal/cleanable/dirt,
+/turf/open/floor/carpet/red,
+/area/template_noop)
+"f" = (
+/turf/open/floor/plating{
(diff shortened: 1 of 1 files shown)


boodaliboo:
the pr we needed

patpol4:
Cute, i suggest adding vomit to the pool if you want realism

jcatjmeow (changes requested):
ballpit tiles are ugly as sin
image
Look at this it's gross

jcatjmeow:
To be clear I'm aware that u didn't make the sprite and it's not your fault but maybe we should give this a resprite after this is merged or before idk

patpol4 (changes requested):
You have to add this to 
/yogstation/code/datums/ruins/station.dm and /yogstation/code/game/objects/effects/landmarks.dm , 
also place it in the correct directory.

If you don't want to mess with this just send the file to me, close the pr and I'll add it bundled with other random maint submissions

patpol4:
User sent me files so i Pr them in bulk, closing

PR #10122, opened 2020-10-23
replaces carp plush with a dehydrated carp in wizard den lavaland ruin

boodaliboo:
thought of this, it isnt very dangerous as its just a space carp and it allows you to make a friend
#### Changelog

:cl:  
tweak: replaced carp plush with a dehydrated one
/:cl:


patpol4:
Why do you insist on adding traitor items to lavaland loot. We're trying to move away from that.

alexkar598:
Isn't this a metashielded item that would require metagaming to know its a dehydrated carp instead of a normal one?

PR #10132, opened 2020-10-24
Replaces botany paper/book with wiki-based, useful manuals

MrHorizons:
Even though people rarely seem to use them, I still think they should be more useful than "vague, outdated botany guide" and "book just for replica pods".

#### Changelog

:cl:  
rscadd: Added guides to hydroponics and plants as books in various maps' botany.
/:cl:


JamieD1 (approved):
Code Good, Weje will check map

patpol4 (approved):
Looks fine to me, just make it compile

Partheo (approved):

PR #10156, opened 2020-10-25
Better discord linking

alexkar598:
Instead of the weird get the id put it in the game then run a command, running the command gives you a link to connect to and when you do, it does the linking at login. The ckey and ids are verified to ensure that you have to finalize the linking with the ckey you used in the command.

Depends on yogstation13/yogbot#227


alexkar598:
The features have been tested in a test environment but I didn't test it in a full SS13 env

JamieD1 (approved):

PR #10188, opened 2020-10-27
Removes Destructibles

Takahiru:

Refactors some code for adam
destructibles are archaic goodbye destructibles
#### Changelog

:cl:  
tweak: removes destructibles
/:cl:


patpol4 (approved):
Maps seem fine, its just line changes, no map edits.

JamieD1:
Looks good, tested? @Takahiru

Bibby0110 (changes requested):
See the following:
https://github.com/yogstation13/Yogstation/blob/master/.github/CONTRIBUTING.md#no-duplicated-code

Use object oriented methods for this, which is what destructible was doing

code/game/objects/structures.dm
```diff
@@ -27,6 +30,20 @@
...
+			if(islist(debris))
+				for(var/I in debris)
+					for(var/i in 1 to debris[I])
+						new I (get_turf(src))
+		if(break_message)
+			visible_message(break_message)
+		if(break_sound)
+			playsound(src, break_sound, 50, 1)
+	qdel(src)
+	return 1
```
Amelia0010:
```suggestion
	return TRUE
```

code/game/objects/structures.dm
```diff
@@ -27,6 +30,20 @@
...
+	if(!disassembled)
+		if(!(flags_1 & NODECONSTRUCT_1))
+			if(islist(debris))
+				for(var/I in debris)
+					for(var/i in 1 to debris[I])
+						new I (get_turf(src))
+		if(break_message)
+			visible_message(break_message)
+		if(break_sound)
+			playsound(src, break_sound, 50, 1)
```
Amelia0010:
```suggestion
			playsound(src, break_sound, 50, TRUE)
```

Amelia0010 (changes requested):
Why though?

code/game/objects/structures.dm
Takahiru: i didnt think of fixing the original badcode

Amelia0010 (approved):
There's legitimately no reason to do this

Amelia0010:
Code is fine but I don't see the point, if Jamie or Alexkar disagree with this they can just re-open and merge it immediately

PR #10189, opened 2020-10-27
Removes Lag

JamieD1:
Removes Lag

#### Changelog

:cl:   
rscdel: Removes Laggy 4 way pipes
/:cl:


alexkar598:
Im not sure that the *4 way* heat exchange pipe is causing the lag, if heat exchange pipes are causing lags, its probably the heat exchange part of it, not the 4 way

JamieD1:
> Im not sure that the _4 way_ heat exchange pipe is causing the lag, if heat exchange pipes are causing lags, its probably the heat exchange part of it, not the 4 way

Nope its shitty engineers being lazy doing this every round:

image

patpol4:
Doing it this way halves the amount of time it takes to make a cooling loop and it works

CMOisLing:
Honestly make heat pipe more effective so we dont need to spam so many of them

MegaEmpirical:
Those lazy fucks, I know what I must do: burn them. If they're all banned, I can keep my pipes.

jcatjmeow:
Redstone pipes when?

Hopekz (changes requested):
Better title

JamieD1:
I will be back

PR #10206, opened 2020-10-27
removes the hard age lock on changing jobs

SomeguyManperson:
attempting to switch an ID to a job it can't take closes the job reassignment  bit of the window and the message is sent to chat meaning it is highly likely you could miss it and be like "why isn't this ID changing jobs wtf"

also call me back when you apply this for nonhumans in head roles :^)
:cl:  
tweak: the fail state for changing an ID to a job it is too young to have has been removed
/:cl:

### DIFF
--- a/code/game/machinery/computer/card.dm
+++ b/code/game/machinery/computer/card.dm
@@ -415,6 +415,8 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 						to_chat(usr, "<span class='warning'>No log exists for this job.</span>")
 						updateUsrDialog()
 						return
+					if(!isnull(modify.registered_age) && modify.registered_age < jobdatum.minimal_character_age)
+						to_chat(usr, "<span class='warning'>This individual is too young to hold that Job, per Nanotrasen guidelines. Suggest aborting Job Assignment!</span>")
 					if(modify.registered_account)
 						modify.registered_account.account_job = jobdatum // this is a terrible idea and people will grief but sure whatever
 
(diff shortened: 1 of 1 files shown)


MegaEmpirical:
Fucking just remove age from the game.

Tipy1802:
It’s not a hard lock you num nun, just change the age

Tipy1802:
IF your problem is people not realising they have to change the age then that’s really a player problem since I haven’t had a problem with that or seen anyone not realise what happen since it says in the chat, even then you can just make the text bigger so it’s harder to miss

SomeguyManperson:
yes I COULD make it less of a pain in the ass OOOOOOR I could make it not a pain in the ass at all, which is what I'm doing

Moja10:
I'm glad you brought up nonhumans, because I would be totally down to make it so neither underage nor nonhumans could switch their ID to a head/age protected role.

Amelia0010 (changes requested):
Make it tell the user they're too young but not block it :)

code/game/machinery/computer/card.dm
```diff
@@ -415,10 +415,6 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 						to_chat(usr, "<span class='warning'>No log exists for this job.</span>")
 						updateUsrDialog()
 						return
-					if(!isnull(modify.registered_age) && modify.registered_age < jobdatum.minimal_character_age)
-						to_chat(usr, "<span class='warning'>This individual is too young to hold that Job, per Nanotrasen guidelines. Job assignment aborted!</span>")
-						updateUsrDialog()
-						return
```
Amelia0010, line 421: just remove these 2 lines and keep the other 2 in

Amelia0010 (changes requested):

JamieD1:
Changed to Suggest only as nich wanted.

PR #10211, opened 2020-10-28
CDN Assets + TGChat + Statbrowser Restyle

Bibby0110:
Ports:
https://github.com/tgstation/tgstation/pull/52681
https://github.com/tgstation/tgstation/pull/52426 + fixes

It has finally reached a full feature parity with goonchat, and is ready for review!

It is as fast as goonchat was (if not faster in certain circumstances), and is very extensible. It has all the necessary code for sorting messages into categories, which means that one of the next features will be multiple tab support.

Additional features that you will get with tgchat right now:

Massively faster server-side performance compared to goonchat, especially if batching multiple messages to one client.
Message persistence across rounds and reconnects. (All messages are stored client-side in IndexedDB)
More robust scroll tracking. If you scroll up, it will not change the scroll position on new messages like goonchat did.
Multiple message combining. (Currently set to combine up to 5 messages over last 5 seconds).
If using the highlighting feature, it highlights the whole message as well as the matching word.
"Now playing" widget, with preview of the song title, a knob for adjusting the volume and a stop button.
image
image

Architecture is as following:

to_chat() -+
           |
        SSchat
   (queue, batching)
           |
  window.send_message()
           |
           v
+-------------+
| tgui-panel  |
|+-----------+|
|| tgchat    ||
|+-----------+|
+-------------+
Subsystem is basically goonchat, but without all the garbage that slows the servers down (string concatenation, double urlencoding, sanitizing, etc). Now, instead of all that, it's being slowed down by json_encode in /datum/tgui_window/proc/send_message, which IMO is completely worth it, and allows sending various templates and widgets to tgchat.

/datum/tgui_window abstracts the whole window away from you, establishes a nice message-passing interface between DM and JS, with two message queues on each side, automatically loads js/css assets for you, basically does everything. You as a developer only have to worry about sending/receiving messages and write javascript.

tgui-panel is a slimmed down version of tgui, and functions as a container for various widgets, and tgchat is one of them. It of course can be expanded with more stuff.

It's also a separate entry point and a JS bundle, so it's not bloating the main tgui bundle, and is currently sitting at about 230kB.

The grand plan is to replace the whole side panel, and have absolutely everything written in inferno, with mindblowing features and customizability.

Asset CDN:
Rewrites the asset_cache system to handle sending assets to a CDN via a webroot.

see https://github.com/MrStonedOne/tgstation/blob/asset-cdn/code/modules/asset_cache/readme.md

Fixed a lot of bugs with assets, removed some dead code.

Changes:

Moved asset cache code to transport datums, the currently loaded one is located at SSassets.transport, asset cache calls made before the config is loaded use the simple browse_rsc transport.
Added subsystem call for when the config loads or reloads.
Added a webroot CDN asset transport. assets are saved to a file in a format based on the file's hash (currently md5).
Assets that don't use get_asset_url or get_url_mappings (such as browser assets referred to by static html files like changelog.html or static css files) can be saved to browse_rsc even when in cdn asset mode by setting legacy to TRUE on the datum returned by register_assets
Added a system for saving assets on a cdn in a hash based namespace (folder), assets within the same namespace will always be able to refer to each other by relative names. (used to allow cdn'ing font awesome without having to make something that regenerates it's css files.).
The simple/namespaced asset cache datum helper will handle generating a namespace composed of the combined md5 of everything in the same datum, as well as registering them properly.
Moved external resource from a snowflake loaded file to a config entry, added it to resources.txt
To ensure the system breaks in local testing in any situation that wouldn't work in cdn mode, the simple transport will mutate the filenames of non-legacy and non-namespaced assets and return this with get_asset_url.
Simple transport's passive send of all roundstart assets to all clients is now a config that defaults to off. this is to break race conditions during local testings from devs accidentally relying on this instead of using send() properly.

Stat panel thing:
image

#### Changelog

:cl:  MrStonedOne for CDN stuff, stylemistake for TGChat, TheGamer01 for porting
rscadd: Interface assets (js/css/images) can now be managed using an external webserver instead of byond's one at a time file transfer queue.
rscadd: Adds admin verb toggle-cdn that allows admins to disable the external webserver asset transport and revert to the old system. Useful if the webserver backing this goes down (thanks cloudflare).
config: New config file, resources.txt, (must be loaded by an $include statement from the main config)
server: The external_rsc_urls.txt config has been moved to the main config system.
rscdadd: TGChat
/:cl:


redmoogle:
Dont forget to build tgui

Bibby0110:
> Dont forget to build tgui

I will
When I'm done

I'm not done currently.

code/modules/tgchat/to_chat.dm
```diff
@@ -0,0 +1,71 @@
...
+	if(target == world)
+		target = GLOB.clients
+	// Build a message
+	var/message = list()
+	if(type) message["type"] = type
+	if(text) message["text"] = text
+	if(html) message["html"] = html
+	var/message_blob = TGUI_CREATE_MESSAGE("chat/message", message)
+	var/message_html = message_to_html(message)
+	if(islist(target))
```
JamieD1, line 32:
```suggestion
	if(islist(target))
	if(!confidential)
		SSdemo.write_chat(target, message)
```

SomeguyManperson:
new alt+click tile panel won't allow you to drag stuff to/from it

JamieD1 (approved):

PR #10223, opened 2020-10-28
removes the ability to infinitely use one bandage

boodaliboo:
you can no longer wash improvised bandages

#### Changelog

:cl:  
rscdel: Removed the ability to wash improvised bandages
/:cl:


boodaliboo:
this doesnt add content so i dont care to fix it

PR #10234, opened 2020-10-28
Syndi pirates no longer spawn with atmos things

ynot01:

they are skeletons
removed atmos alarms
removed atmos piping
removed atmos tank
removed 2 EVA suit storages
removed breath mask and oxy tank from pirate spawn
added bonus of not reporting atmos alarm to station-side atmos computers

#### Changelog

:cl:  
rscdel: Syndi pirates no longer spawn with atmos things
/:cl:

### DIFF
--- a/_maps/shuttles/pirate_default.dmm
+++ b/_maps/shuttles/pirate_default.dmm
@@ -1,19 +1,4 @@
 //MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
-"aa" = (
-/obj/structure/chair/comfy/shuttle{
-	dir = 1
-	},
-/obj/machinery/light/small{
-	dir = 8
-	},
-/obj/machinery/airalarm/all_access{
-	dir = 4;
-	pixel_x = -24
-	},
-/obj/machinery/atmospherics/components/unary/vent_pump/on,
-/obj/effect/decal/cleanable/dirt,
-/turf/open/floor/pod/dark,
-/area/shuttle/pirate)
 "ab" = (
 /obj/structure/table,
 /obj/effect/decal/cleanable/dirt,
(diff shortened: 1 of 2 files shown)


Hopekz:
The ship is stolen not really purpose-built for them

Takahiru:
> The ship is stolen not really purpose-built for them

tbh makes sense with this reasoning, though theres no reason for them to have oxygen tanks

TheLinuxOS:
> The ship is stolen not really purpose-built for them

I mean if I was a pirate id remove parts of the ship that do nothing but take up space, such as removing the air alarms and selling the eva suits and oxygen supplies so it makes sense for them to be removed in my opinion

SomeguyManperson:
suits and stuff are useful for kidnapping people if you don't want to drop the ship right next to the station

ynot01:
> suits and stuff are useful for kidnapping people if you don't want to drop the ship right next to the station

valid point, i will counter anecdotally that i have never seen this happen

ToGWtF:
Because selling living people is so little money compared to just raiding the entire ore supply.

patpol4:
Is there any reason for removing this aside from "It's not used by them".
Just treat it as map fluff that makes the ship look more lively and not barren.

ynot01:
the ship goes blue when atmos goes slightly fucky because theres air alarms and vents

this also alerts the station that theres an atmos alarm in the ship

pretty pointless to have it all

Hopekz:
> this also alerts the station that theres an atmos alarm in the ship

Sounds like a different PR to make ships outside of the station not send atmos alarms to the station.
If this is the main reason for this PR existing you are fixing it in a very sloppy way that in my opinion is compromising the feel of the pirate ship being stolen.

patpol4 (changes requested):
I'm ok with removing only the air alarms as a bandaid fix until a coder fixes the actual issue. Basically what Hopek said.

ynot01:
if you're so worried about kidnappings i can put an o2 canister in and you can set it to 101kpa

patpol4:
It's more a matter of being worried of removing things that don't need to be removed. I'm proposing a non code solution that will basically keep the shape and look of the original interior, while solving the problems you listed.

JamieD1:
This will ruin live crew interactions completely.

Reed0506:
what if the air alarms are such are just smashed and broken by default, not compromising hopek's "stolen" vibes, but also removing the issues that the bits and bobs can cause.
if the pirates are kidnapping people, said people will typically survive a bit of exposure to space unless you have gone sicko mode on them and dunked them deep into hard crit.

PR #10241, opened 2020-10-29
Ethereal Changes

Marmio64:
Buffs-
Ethereals are now rad immune, since they like heat and energy (and radiation is energy)
Ethereals now take 10% less burn damage than everyone else, since they like heat and are used to it.

Nerf-
Ethereals now take 25% more damage from the cold.

_Should_ be balanced. Previously their only real upside was taking less shock damage, while having to deal with 25% more brute damage (up to 75% more if they were starving)

edit: Ethereals no longer rad immune, instead they just get a shock resistance buff, making it so shocks only deal 25% their normal damage to ethereals.

#### Changelog

:cl:  
tweak: Ethereals are weaker to the cold, but are rad immune and take 10% less burn damage.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/ethereal.dm
+++ b/code/modules/mob/living/carbon/human/species_types/ethereal.dm
@@ -11,11 +11,13 @@
 	exotic_blood = /datum/reagent/consumable/liquidelectricity //Liquid Electricity. fuck you think of something better gamer
 	siemens_coeff = 0.5 //They thrive on energy
 	brutemod = 1.25 //They're weak to punches
+	burnmod = 0.9 //They like heat and energy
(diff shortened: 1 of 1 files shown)


UselessTheremin:
It would make more sense if instead of being rad immune they were shock immune, as in my opinion plasmamen should have rad immunity as a unique strength, and ethereals literally eat liquid electricity

Moggram:
> 
> 
> It would make more sense if instead of being rad immune they were shock immune, as in my opinion plasmamen should have rad immunity as a unique strength, and ethereals literally eat liquid electricity

It is true that they are living light bulbs, but put enough power into one it will go pop.

Marmio64:
closing cause nobody seems to like it

PR #10244, opened 2020-10-29
Fixes bounty board displaying incorrectly

### LINKED ISSUE #10213: Bounty Board shows incorrect name of Non-Human Backup Characters upon use
## Round ID: 32544

For Context: If you have a Non-Human save slot for your character, and you choose a Head Role, your character will become the Human with a "Backup Name" to prevent Non-Humans in Head Roles. 

The Bounty Board is not programmed to deal with this.

If you spawn as a Human character, because you chose a Head Role, but your character is normally Non-Human, the Bounty Board will still show your Non-Human name. It must not be getting the name from the ID card as it probably should.

## Reproduction:

Spawn as a Human because you were forced to, after choosing a Non-Human character, and selecting a Head Role job role. After you spawn in as your backup name, use the Bounty Board. See the wrong name appear.

redmoogle:
Fixes: #10213 

Just copied some code, should work

#### Changelog

:cl:  
bugfix: fixed bountyboards displaying incorrectly
/:cl:

### DIFF
--- a/code/game/machinery/bounty_board.dm
+++ b/code/game/machinery/bounty_board.dm
@@ -81,9 +81,7 @@ GLOBAL_LIST_EMPTY(request_list)
 				formatted_applicants += list(list("name" = j.account_holder, "request_id" = request.owner_account.account_id, "requestee_id" = j.account_id))
 	var/obj/item/card/id/id_card = user.get_idcard()
 	if(id_card?.registered_account)
-		current_user = id_card.registered_account
-	if(current_user)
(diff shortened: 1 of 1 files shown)


Bibby0110:
What does this even do you just moving code around

redmoogle:
Well you see the code I was looking at for ID cards was doing `registered_account.account_holder` and that displays correctly so, maybe this will fix it? will test in a second

redmoogle:
Ok so whoever coded the bank account system made it init before the id card and I cant figure out how to move it

PR #10245, opened 2020-10-29
Makes the tesla not able to zap the same object twice

### LINKED ISSUE #9877: Tesloose caused timers to die
Issue reported from  Round ID: 32107 (Yogstation 13 (Main))

Reporting client version: 513.1532

Tesla loosed and shortly after the timer subsystem die a kill. Runechat fucked, sec timers fucked, the whole nine yards. Map is boxstation.

redmoogle:
Who and why?

Roughly  a 7% decrease in time dilation on average on my end 10% -> 3%

Should fix: #9877

#### Changelog

:cl:  
tweak: increased tesla performance 
/:cl:


ynot01:
does this affect containment?

patpol4:
Won't this cause it to not generate power?

Hopekz:
> Won't this cause it to not generate power?

^ this. 

If it doesn't break as a result nothing really important lost for a nice gain.

redmoogle:
Output isnt affected, why would containment be affected

Hopekz:
> Output isnt affected, why would containment be affected

I think the logic would be:
Because the tesla needs to zap things to make power. 
There is only oh so many things to zap in the engine area so over time it'll run out of things to zap and make no power because it ran out of things to zap.

ynot01:
> Output isnt affected, why would containment be affected

while in containment you often see many tesla shocks hitting the same tesla coil/grounding rod
if this PR makes it so coils can only accept one shock at a time this could prove problematic to containment, especially when parts of it fail

redmoogle:
> > Output isnt affected, why would containment be affected
> 
> while in containment you often see many tesla shocks hitting the same tesla coil/grounding rod
> if this PR makes it so coils can only accept one shock at a time this could prove problematic to containment, especially when parts of it fail

How are all of you missing it àaaaaa.

It makes it so a shock cant keep smacking the same thing more than once. It does not stop it from repeatdly shocking it. Just not more thab once. Tldr lighting doesnt chain back onto itself

Ps i changed how the tesla zaps not the coils

Amelia0010 (approved):
I just spent forever testing this and it's fine.

PR #10259, opened 2020-10-30
Removes greentext from traitors

ynot01:

Discourages the 'win' and gloating mindset
Discourages the validhunt/murderbone to 'win' and 'must redtext the traitor' mindsets
Encourages the addition of more creative and liberal objectives without worrying about how possible it is to track within code
Tested in private server, working with no runtimes
Avant-garde
requesting test merge

#### Changelog

:cl:  
experimental: Removed greentext from traitors 
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/datum_traitor.dm
+++ b/code/modules/antagonists/traitor/datum_traitor.dm
@@ -362,35 +362,20 @@
 
 	var/objectives_text = ""
 	if(objectives.len)//If the traitor had no objectives, don't need to process this.
-		var/count = 1
 		for(var/datum/objective/objective in objectives)
-			if(objective.check_completion())
-				objectives_text += "<br><B>Objective #[count]</B>: [objective.explanation_text] <span class='greentext'>Success!</span>"
-			else
-				objectives_text += "<br><B>Objective #[count]</B>: [objective.explanation_text] <span class='redtext'>Fail.</span>"
+			if(!objective.check_completion())
 				traitorwin = FALSE
-			count++
 
 	if(uplink_true)
 		var/uplink_text = "(used [TC_uses] TC) [purchases]"
 		if(TC_uses==0 && traitorwin)
-			var/static/icon/badass = icon('icons/badass.dmi', "badass")
-			uplink_text += "<BIG>[icon2html(badass, world)]</BIG>"
 			SSachievements.unlock_achievement(/datum/achievement/badass, owner.current.client)
 		result += uplink_text
 
 	result += objectives_text
 
-	var/special_role_text = lowertext(name)
-
 	if (contractor_hub)
 		result += contractor_round_end()
-
-	if(traitorwin)
-		result += "<span class='greentext'>The [special_role_text] was successful!</span>"
-	else
-		result += "<span class='redtext'>The [special_role_text] has failed!</span>"
-		SEND_SOUND(owner.current, 'sound/ambience/ambifailure.ogg')
 
 	return result.Join("<br>")
 
--- a/code/modules/antagonists/traitor/datum_traitor.dm
+++ b/code/modules/antagonists/traitor/datum_traitor.dm
@@ -362,9 +362,14 @@
 
 	var/objectives_text = ""
 	if(objectives.len)//If the traitor had no objectives, don't need to process this.
+		var/count = 1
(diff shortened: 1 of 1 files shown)


ToGWtF:
On the fence.

I like what you hope it will achieve though.

PrinceKirze:
as long as they have objectives i don't see how this will change anything but im always open to seeing how small changes could affect server culture so i'm open to it

patpol4:
Interesting

Marmio64:
is there still objectives to complete, and will they show up on round end? If so this accomplishes nothing. This will still probably not accomplish anything so long as there are objectives to complete.

code/modules/antagonists/traitor/datum_traitor.dm
```diff
@@ -365,33 +365,23 @@
 		var/count = 1
 		for(var/datum/objective/objective in objectives)
 			if(objective.check_completion())
```
Bibby0110, line 367: no need for the if block since both pieces of text are identical

boodaliboo:
why? if your not removing objectives i dont understand

ynot01:
> why? if your not removing objectives i dont understand

i said why

removal of objectives would lead to other things which i don't believe we're ready to tackle yet

code/modules/antagonists/traitor/datum_traitor.dm
ynot01: no they arent

Tipy1802:
Neat, it would allow us to add in more broader objectives that cannot have a complesion check

UselessTheremin:
just because it wont be called greentext or redtext

doesnt mean the problems will go away

in my opinion, we should first think up some fun traitor objectives (possibly steal some from goon) because removing greentext and redtext with the current climate of objectives would do literally nothing as they are so black or white in whether or not you completed them.

neeshacark:
the only thing this will do is make more objectives that couldn't previously be added due to completion check possible

that said I think this is a bad and almost pointless change

ghost:
> Encourages the addition of more creative and liberal objectives without worrying about how possible it is to track within code

you can just make objectives autocomplete instead. that's what we already do for certain antagonists and certain objectives.

removing the win condition doesn't remove objectives-people will still try to complete objectives regardless of if it says they officially win or officially lose, because people like to kill, and killing is all they will generally want to do with some rare exception. so they'll kill what they're allowed to kill or can get away with killing, which is what their objectives serve to restrict them by.

objectives aren't a block preventing you from adding more freeform or creative objectives-they're a leash used to keep murderboners in line. removing the win/lose condition doesn't change that behavior.

code/modules/antagonists/traitor/datum_traitor.dm
Bibby0110: Enlighten me, what’s the difference?

jcatjmeow:
While I do support the removal of greentext I'd like it to look like some other station's freeform objectives and not like this. This is just lazy and sloppy tbh. 

(No offense to you ynot <3)

JamieD1:
Awaiting Vote 751

ynot01:
> While I do support the removal of greentext I'd like it to look like some other station's freeform objectives and not like this. This is just lazy and sloppy tbh.
> 
> (No offense to you ynot <3)

part 1 in branch name, freeform objectives will come through heavy group brainstorming if this gets merged

patpol4:
> > While I do support the removal of greentext I'd like it to look like some other station's freeform objectives and not like this. This is just lazy and sloppy tbh.
> > (No offense to you ynot <3)
> 
> part 1 in branch name, freeform objectives will come through heavy group brainstorming if this gets merged

this, its impossible to make this change without this pr as a transition

maxion12345:
Not sure how this would be liked. Not having win conditions as antagonist might encourage more casual play, but at the same time it feels good to complete your objectives without being caught, and having your name green in a sea of red for the server to see.

Sir-Lagsalot:
I never thought I'd see the day...
And by Ynot, no less

fluffe9911:
Council vote

patpol4:
https://forums.yogstation.net/threads/vote-751-approve-or-deny-this-pr-https-github-com-yogstation13-yogstation-pull-10259-remove-red-and-green-texting.22597/

PR #10268, opened 2020-10-31
removes thermal vision from genetics unlockable powers because it's basically just xray

ghost:
might as well get this in before i finish up the cybernetics pr 

i'm going back to making dinner now

:cl:  
rscdel: you can't get thermals in genetics anymore
/:cl:

### DIFF
--- a/code/datums/mutations/sight.dm
+++ b/code/datums/mutations/sight.dm
@@ -41,6 +41,7 @@
 	difficulty = 18
 	text_gain_indication = "<span class='notice'>You can see the heat rising off of your skin...</span>"
 	time_coeff = 2
+	locked = TRUE
 	instability = 25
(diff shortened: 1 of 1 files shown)


Hopekz:
Good PR, should have went with Moja's x-ray removal in my opinion.

Moja10:
And again, thermals still exist, but only as medical implants you get from research  I believe.

ghost:
or from certain antagonists like nightmares and shadowlings and xenomorphs eyes, which if you remove and implant them, give thermal vision

Tipy1802:
Removing x Ray was fine but thermals is a bit too much in my opinion

Ratvar:
Tipy, could you explain why it is too much - argumenting would help thermals stay ingame more then dislike emoji, and you probably want that

JamieD1 (approved):

PR #10276, opened 2020-10-31
Let's hear some tunes - volume 1 second edition

burn1cycle:
Adds some house-y style tunes as well as few chill tracks. 
Removed Rock is Dead for sound level reasons.


LoliconSlayer:
Inb4 "Atomize your PR"

JamieD1:
@Partheo Jesus have mercy.

patpol4:
That's a lot

JamieD1:
I appreciate the work you put into this, but I feel im going to move the jukebox to a youtube based system instead as this is ALOT of extra files for people to download for almost no reward as their playtime will be several rare.

alexkar598:
but they are loaded dynamically and only downloaded when the song is played..

alexkar598:
A youtube system would also require downloading the song on the server and sending it to the client at runtime, theres 0 difference apart from the fact that YouTube dl is kinda janky sometimes

maxion12345:
> A youtube system would also require downloading the song on the server and sending it to the client at runtime, theres 0 difference apart from the fact that YouTube dl is kinda janky sometimes

But youtube would allow more flexibility for songs, so that could be pretty fun

PR #10280, opened 2020-10-31
fixes airlock shitcode

athena413:

Basically, if any airlock is considered "abandoned" by mapping, it would roll a number between 1 and 100, triggering various effects ranging from the door being completely replaced by a wall, or stuff like the door being bolted.

To poorly explain: if it rolls a 9, its in range to trigger MULTIPLE effects, though only one works because it runs first and completely overrides the entire switch case.

This has no effect on gameplay or server performance. I just want the code to look nice.

#### Changelog

:cl:  
bugfix: fixes airlock shitcode
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -151,7 +151,7 @@
 				here.PlaceOnTop(/turf/closed/wall)
 				qdel(src)
 				return
-			if(9 to 11)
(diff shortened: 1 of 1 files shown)


MegaEmpirical:
#fuckihateabandonedairlocks

SomeguyManperson (approved):

JamieD1 (approved):

PR #10284, opened 2020-11-01
I think probably reduces the spam of soulstones and constructs by making them use mind transfer instead of key setting

SomeguyManperson:
image
but unironically this time

:cl:  
tweak: soulstones will no longer create duplicate minds for the same person on  being used > turned into constructs
/:cl:


redmoogle:
This might break soulstoning people who are in a unclonable state eg suicide or do not revive. Not a bad thing but cosider it

SomeguyManperson:
bodies that require a new mind just have a ghost's key assigned to them then have that do this stuff so it SHOULD work

JamieD1 (approved):

PR #9633, opened 2020-09-03
Uniform Bar Variants: Added Drink Showcases, Christmas Tree Spawners, and more.

alao3alao3:
The intent of this PR is to have all bar variants be more uniform. All bars now have 3 Generic Event spawners (Another for the theater), xeno_spawn in their backrooms, a Drink Showcase, a Blob Start surrounded by walls (bar_cheese is the most affected by this), a Christmas Tree Spawner (as much as I'm weary of the items), 2 Chef starts, 3 Assistant starts, 1 Bartender starts, Pun Pun, a Honkbot in the theater (should one exist) and an image of Mr Deempisi. 

Ontop of this, the Generic Event spawner has been removed from the Kitchen Backrooms, as is unform with bar_box.

#### Changelog

:cl:  
rscadd: Added several starts, Pun Pun to bar_diner, added xeno_spawns to bars that did not have it, added blob starts to bars that did not have it, added drink showcases to all bars that did not have it, added a honkbot to all existing theaters, added christmas tree spawners,
rscdel: Removed generic event spawners in kitchen backrooms that had them, removed extra chef starters.  
/:cl:

PR #10298, opened 2020-11-01
Removes Blob Starts and Generic Event Spawners from Random Maint

alao3alao3:
#### Intent of your Pull Request
Removes Blob Starts and Generic Event Spawners from Random Maints, specifically:
5x4 Box Surgery
10x10 Pod Repair Bay
10x10 Graffiti Room
10x10 Delta Library

###### Why is this good for the game?
Uniformity. We shouldn't let this be variable, and even if it isn't a problem on Boxstation, I don't want this to bite us in the ass when random maints are implemented on other maps. If people want generic event spawners and blob starts in random maints, it should be in all of them, or atleast all in a size category, but not some. This logic is in the same vein as #9633.

# Changelog
:cl:  
rscdel: Removed generic event spawners and blob starts in various random maint presets  
/:cl:

### DIFF
--- a/_maps/RandomRuins/StationRuins/maint/10x10/10x10_deltalibrary.dmm
+++ b/_maps/RandomRuins/StationRuins/maint/10x10/10x10_deltalibrary.dmm
@@ -113,11 +113,6 @@
 /obj/effect/decal/cleanable/dirt,
 /turf/open/floor/plating,
 /area/template_noop)
-"az" = (
-/obj/effect/decal/cleanable/dirt,
-/obj/effect/landmark/blobstart,
-/turf/open/floor/plating,
-/area/template_noop)
 "aA" = (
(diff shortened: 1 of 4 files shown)


JamieD1:
I mean blob spawns dont really matter as players pick where they put it down.

alao3alao3:
so do they do anything at all or can we run a quick PR to cull obsolete fat like that off the station.

jcatjmeow:
> so do they do anything at all or can we run a quick PR to cull obsolete fat like that off the station.

they pick where your cursor starts. So I guess it's good to have at least one, but there also might be some code that has a check for that

patpol4 (approved):
Makes sense to me, they should be on the maps themselves, not the templates.

Ktlwjec1:
No reason to remove them as I said in discord, I'm closing this.

PR #10300, opened 2020-11-01
Linear Chem Heater but not shit this time

redmoogle:

TLDR: target temp yeeted

New UI:
image

**Linear for simplicity, will put any equation after I balance it**

heats up by 10K(at max coil power) at base tier
heats up by 40K(at max coil power) at top tier

Coil Power goes between 0 and 5
Has a mode to switch between Heating and Cooling
Heater power adjust how much per coil power the speed goes(0.5 increments)

Coil power is a multiplier
Base Tier is 2K per coil power
Top Tier is 8K per coil power

It will continuously keep heating regardless of temperature, as long as it's on

Keeps the danger of chemical heaters
Fixes a few issues with the old heater

_and yes it works cooling down as well, ~that never happened shhhh~_

also cool UI

ps ignore the - in -1, thats just tgui being retarted on my end

#### Changelog

:cl:   
tweak: Chemical heaters now function vastly differently
/:cl:


MegaEmpirical:
Explain more to me EXPLAIN.

redmoogle:
> 
> 
> Explain more to me EXPLAIN.

updated description

gar4:
Removing the target temperature makes it needlessly complicated and dangerous to use. All you need is to lag out for a tick and suddenly you have a giant hole where chem used to be (and lets not pretend the server never lags).

redmoogle:
> 
> 
> Removing the target temperature makes it needlessly complicated and dangerous to use. All you need is to lag out for a tick and suddenly you have a giant hole where chem used to be (and lets not pretend the server never lags).

This is a balance question that's already been addressed

gar4:
Okay since discord is awful at letting you find where people discussed a specific topic mind pointing me to the right place / summing up where that consensus was reached? 

Because balancing something around randomness is usually not the way to go. "It is supposed to be dangerous so the risk of lagging out and killing yourself and your fellow chem" is a bad way to go with it.

Reed0506:
this is both unrealistic, annoying to deal with, and would also cause issues in the event of lag spikes as mentioned above. i do not see why this is good for the game. realism is not a huge priority, i know, but come on man, we can set anything from laboratory equipment to an oven to heat to a specific temperature target and not need to deal with micromanaging everything in our current time, let alone the far future.

redmoogle:
> 
> 
> this is both unrealistic, annoying to deal with, and would also cause issues in the event of lag spikes as mentioned above. i do not see why this is good for the game. realism is not a huge priority, i know, but come on man, we can set anything from laboratory equipment to an oven to heat to a specific temperature target and not need to deal with micromanaging everything in our current time, let alone the far future.

This was the proposed solution everyone wanted me to do. So what now, just delete chem heaters next?

redmoogle:
Moving this to a future project

redmoogle:
yogsbot why????

PR #10313, opened 2020-11-03
Comments out deoxyribostupid symptom

Mqiib:
I've been told to just delete it instead of commenting it out so here we go

#### Changelog

:cl:  
rscdel: **Deletes** genetic mutation disease symptom because it's broken.
/:cl:


Hopekz (changes requested):
Just remove it. Commenting out whole sections like this just looks sloppy.

Mqiib:
I have hopes that it may be fixed.

MegaEmpirical:
I actually would prefer the mutations stay but the weight of its stats should likely be adjusted to accommodate.

Mqiib:
> I actually would prefer the mutations stay but the weight of its stats should likely be adjusted to accommodate.

Yes the symptom is not that bad until you get one of:
Spatial instability (randomly teleport and vomit forever (yes forever I hope you like being permastunned))
acid skin (you can't wear clothes for longer than 5 minutes and slowly die)
flaming sweat (spontaneous combustion have fun)
blind/deaf/mute but forever

At which point it becomes decidedly _not okay_ and if specifically you roll spatial instability you're just fucked forever unless someone can speedrun brain swap on you before you teleport away.

jcatjmeow (changes requested):
Don't do this.
Delete it. If you want to know what the code used to be you can use the github history. Commenting vast chunks of code like this is not generally advised for this reason.

JamieD1 (approved):

PR #10316, opened 2020-11-03
adds BRPED and power cells to syndie lavaland base

UselessTheremin:
i wish that i could upgrade the turbine, and now, with this, i can. also, the stock parts in the syndie base do not have any powercells, and it makes sense to have complete sets of parts.
🆑
rscadd: adds bluespace powercells and rped to syndie base
/🆑


UselessTheremin:
oh god dammit did i add a circuit board

patpol4 (approved):
Don't see how this can go wrong

PR #10319, opened 2020-11-04
Acid tweaks

Mqiib:
Couple little things here:
1: acidmod now actually modifies damage taken from acid as opposed to only applying when/if you get disfigured (specifically only to the head and only affecting the burn damage what the fuck)
2: TRAIT_ACIDBLOOD has been added for mobs with acidic blood, hooray! No more shall they die to their own blood being injected back into them! They will still take toxin damage from stronger acids, though.
3: Polysmorphs (and xenomorphs when I figure it out) have the above trait.

#### Changelog

:cl:  
rscadd: TRAIT_ACIDBLOOD for all you acid-blood-ers is now here! 
bugfix: Polysmorphs no longer take damage from trying to refill their blood with what they just pulled out of themselves.
bugfix: acidmod now actually affects acid damage taken.
/:cl:


Mqiib:
Update on the xenomorphs: Giving them the trait doesn't actually do what it's supposed to do and according to the code they, in a medical sense, don't actually HAVE blood so I'm just gonna hope nobody complains about them taking damage from sulphuric acid injections.

JamieD1 (approved):

PR #8223, opened 2020-04-18
Adds Amber Task Force. CentCom, ERT and Inquisition equipment Revamp

Hopekz:
**TL;DR:**
Revamped Blue and Red ERT equipment load-outs.
Revamped CentCom Commander and CentCom Offical equipment load-outs.

Update 2:
Revamped the Inquisition response team equipment load-outs.

Update 3:
Recreated the Amber ERT into the Amber Task Force. This new team was built from the ground up.

CentCom
------

**CentCom official:**
- Replaced satchel with a leather satchel
- Added internals box to backpack
- Now mindshielded
- replaced energy gun with a telescopic baton
- moved clipboard from hand to backpack
- **Adds a mindshield to the centcom costume syndicate box for nukies now that real officials are mindshielded. (Pointed out by Gamer01)**

**CentCom Commander:**
- Commander gets a security belt
- Moved revolver to suit storage from the belt slot
- Added an internals box to backpack
- Now mindshielded
- (still not immune to flash or sound because of basic headset + normal eyepatch)

ERT blue alert::
----------

**Blue ERT Commander:**
- security gas mask moved from backpack to face
- energy gun moved to suit storage instead inside the backpack 
- Keeps his duplicate stun baton because he's the commander. So can "give" it to someone.

**Blue Security ERT:**
- tactical energy gun replaced with an energy gun
- energy gun moved to suit storage instead of the backpack
- security gas mask moved from backpack to face
- Removed stun baton from the backpack. They already get one in the belt and this is blue alert
- Box of handcuffs replaced with a box of zipties

**Blue Engineer ERT:**
- security gas mask moved from backpack to face
- removed energy gun; keeps baton
- replaced stun baton with the telescopic baton
- Removed medkit in hands
- added extra compressed matter cartridge to free pocket
- moved extended-capacity emergency oxygen tank to suit slot
- removed internals box now that mask is on the face and o2 tank is on suit slot
- added Epinephrine medipen to backpack where the box used to be

**Blue Medic ERT:**
- security gas mask moved from backpack to face
- removed energy gun
- replaced stun baton with the telescopic baton
- removed belt
- moved telescopic baton to belt slot
- moved the medical beamgun from backpack to suit storage
- moved combat stimulant injector from backpack to pocket
- removed regular medkit held in hands
- Added toxin first aid kit to backpack
- Added fire first aid kit to backpack
- Added brute first aid kit to backpack
- Moved extended-capacity emergency oxygen tank from internals box to right pocket
[...]

PR #10326, opened 2020-11-05
Occupying force ERT

Hopekz:
follow up to #8223. 
Originally this was going to be released along with the Amber Task Force but I had to split up this project.
Extensively balanced and ready for review.

**Adds a new Occupying Force response unit.** 
Also adds a brand new tactical tool called the Airlock Jimmy (Currently only carried by the Occupying Force)

u34X9qa4GL

Occupying Force
-------

**What is it?**
The occupying force is a non-lethal step down from the Amber Task Force. 
The Amber Task Force is a lethal ERT as they do not possess the ability to crowd control. The Occupying force is a weaker non-lethal variant meant to "occupy" and try to restore order vs being sent in to directly fight a dangerous threat.

The Occupying force is designed to be easily countered as they have many purposeful, crippling weaknesses. 
**This gives a way to introduce weaker sec reinforcements into a shift without overwhelming the antagonist forces.**

**Simplified differences from the Amber Task Force:**
-----
- Designed to be light non-round ending sec reinforcements that stay in a shift to assist as Officers. Have been given amenities such as food, beer, and cigarettes to make their stay bearable.
- Given Security Officer access (to perform arrests and assist in the chaos as sec reinforcements)
- They have new Surplus auto rifles loaded with rubber rounds vs the lethal laser rifles.
- Every officer has a non-lethal melee option.
- No medic. Medic has been replaced with a Riot Officer. The squad gets 1 riot officer per usual group size.
- No hearing protection.
- No internals (except the Riot Officer)
- No shock protection (no combat gloves)
- Much less armor. 

**Ways to counter:**
- No hearing protection
- No eye protection (except Riot Officer)
- No internals starting out (except Riot Officer)
- No shock resistance
- Much weaker armor.
- their rubber guns are weaker than disablers in terms of taking down a target. non-armored require 7 hits, basic armored vest will raise this to 9.
- The surplus stun prod they hold (minus the Riot Officer) takes 2-3 seconds to down if they go for their non-lethal melee option.

Surplus Auto Rifle
----

Every class in the ERT except the Riot Officer gets the new Surplus Auto Rifle.
The Surplus auto rifle is essentially the WT rifle before its 2 round burst buff as this PR has been extensively tested for balance and the 2 round burst made them a bit too viable in my opinion. The new armor penetration stayed and it is very fun to fight threats with the original stats.

Airlock Jimmy
----

A new tool meant exclusively for opening airlocks.
Has a pump assisted system. You pump the tool to charge it by 25 air. You use 50 air to open an airlock (quicker prying time) and can hold a maximum of 100 air.

The Airlock Jimmy loses air gradually 30 seconds after it has been charged which means it needs to be charged by pumping before opening the airlock.

**This tool's pressure gauge can be emagged in order to increase the pressure in the device.** 
Might make this hackable vs emag locked in the future.

**What this means:**
After being emagged the airlock Jimmy now uses 75 air instead of 50 air to open an airlock and can hold a max of 150 air (
Because it uses more air per open, it has more "pressure" meaning that it opens airlocks quicker.

The down-side being that it now requires 3 pumps instead of 2 in order to open an airlock. 6 pumps to fill up vs 4.

Similarities in classes:
-----

- The Commander and the Officers (excluding the riot officer) get the new tactical stunprod. it is "a cost-effective, mass-produced, tactical stun prod.". Essentially regular stun prods with flavor text and colored to be tactical.

- amenities such as food, beer, and cigarettes to make their stay bearable. All members have a Pizza MRE, a beer, and cigarettes (excluding the Riot officer). 

- Fancy Darker green CentCom Soldier uniforms

- All of them have the new Airlock Jimmy tool.

Main differences in classes:
---

Occupying Commander
----

ig7vMRUuKu
R8NzftJqMk

- Has a megaphone because he's the boss
- Sweet cigar vs cigarettes!
- Sec hud
- Has the only lighter in the Force. He'll have to share!
- Has a Nuclear authentication disk pinpointer. This is the only ERT that will be able to recover the captain (or whoever holds the disk) or attempt to take back the disk.

Occupying Officer
------
dlqLlp80Mp
90xrn4M6eA

- Most common Officer type in the ERT
- Think of them as a very watered down sec reinforcement. Their gun is worse than a disabler, their stun prods are worse than a stun baton.
- They have a light inside of their helmet while the other classes have to rely on flares or finding a light source from the environment.
- Cigarette pack to support their soon to be new addiction (thinking about having them start with a nicotine addiction, not yet implemented)

Occupying Riot Officer
----
dIOtWsSLVV
Zj5Ex67W9C
GoEfjGwI69

- Melee only! (no gun) . Gun replaced with a proper stun baton.
- full-sized riot shield on their backs.
- 2 energy bolas to close in on their target (since no ranged solution + slower due to their riot gear)
- Unique riot suit! The Riot suit is weaker than the NT riot armor found in the station armory (by a whole 20 points of armor)
- cool tactical internals (same as the Amber Task Force) allowing them to remain anonymous when wearing it up.
- Eye protection (sunglasses)
- A flare. The Riot Officer has no other light sources besides this so if they run out they'll have to scavenge something off the environment.
- Proper handcuffs instead of zipties! All the other officers are made with dwindling supplies while the riot officer can theoretically re-use his handcuffs to stay "in the fight" longer without resupply.

**UPDATE:**
Riot officer is now completely straight edge (no smokes or beer).
Sports earmuffs instead of a beer ):

#### Changelog

:cl:  Hopek
rscadd: Added the Occupying Force. A weaker non-lethal step down from the Amber Task Force. These come with their own unique items such as surplus auto rifles and a unique riot suit variant.
rscadd: Adds the Airlock Jimmy. An air-assisted pump-action airlock opening tool. Currently only held by the Occupying Force.
/:cl:


Hopekz:
Works on my machine (: 
image

maxion12345:
Not sure the stunprods and lack of basic sec gear fit the "we're here to help security" idea, especially when s round start officer without any gear from a locker is better equipped then these ERT, barring the riot officer.

also, stunprod looks wayyy too makeshift for an ERT. Normal stunbatons or telebatons would fit better. In addition, the airlock jimmy should go to the amber team, instead of this. 

tl;dr seems too poorly equipped to be a proper security ERT

code/game/objects/items/cards_ids.dm
```diff
@@ -478,6 +478,14 @@ update_label("John Doe", "Clowny")
 	desc = "An Amber Task Force ID card."
 	assignment = "Amber Task Force"
 
+/obj/item/card/id/ert/occupying
+	name = "\improper Occupying Force ID"
+	desc = "An Occupying Force ID card."
+	assignment = "Occupying Officer"
+
+/obj/item/card/id/ert/occupying/Initialize()
```
Bibby0110, line 486: Make turdis compile

Tipy1802:
Should have a backpack of sorts, the riot officers can just get their batton disarmed or stolen when they leave it to pick something else

Tipy1802:
Considering they are supposed to fight riots maybe giving them teargas granades won't be a bad idea

Hopekz:
> Should have a backpack of sorts, the riot officers can just get their batton disarmed or stolen when they leave it to pick something else

They have a dedicated space for their baton and shield.
Shield = on back
baton = in the suit storage slot 
Sorry; my screenshot is misleading a bit since I put them in his hands to try to take a "cool-looking" screenshot.

Hopekz:
> Not sure the stunprods and lack of basic sec gear fit the "we're here to help security" idea, especially when s round start officer without any gear from a locker is better equipped then these ERT, barring the riot officer.

Stun prods are just stun batons that you can only store on your back. They're actually even more powerful because there is a guaranteed stun after 2-3 seconds after 1 hit.

You are correct in the sense that these are for the most part meant to be weaker than round start security which is why their guns take more shots to down vs a disabler. They have security access so they are able to scavenge what is remaining otherwise they won't be as effective.

> also, stunprod looks wayyy too makeshift for an ERT. Normal stunbatons or telebatons would fit better. 

The point of using the stun prods is to force them to use their back slot to store the item. Also using a tele prod is just weird here. I am not a spriter so if someone wants to make something more official after I wouldn't fight them.

> In addition, the airlock jimmy should go to the amber team, instead of this.

I was thinking of doing this as an addition but I think Adam is going to do a PR in that file so it might cause conflicts. I'm going to wait until he's done with whatever he's doing before I modify the Amber file.

code/game/objects/items/cards_ids.dm
Hopekz: thanks

swissloaf:
does anyone even use amber never sen that used

Hopekz:
> does anyone even use amber never sen that used

I see it used at least once every 3 days to a week which is more often than I see Red or death.

UselessTheremin:
i would love if they had sunglasses, as all other central command officers have them, and it seems good for consistency. Plus, they would just steal regular HUDglasses anyway because they have access if they are not given eye protection.

Hopekz:
Update:
The cigarettes on their mouths all start lit except the commander because he's lucky enough to be given a lighter.

Hopekz:
UPDATE:
Riot officer is now completely straight edge (no smokes or beer).
He now sports earmuffs instead of a beer for optional hearing protection at the expense of communication effectiveness. 
Thank you to Fluffe for the suggestion!

ghost:
Just seems like more canon fodder, just like the Amber ERT. It's different for every admin so it is hard to really pin down what the policy is for spawning in an ERT, but it seems to me like they're usually called in when a round has gone on for too long or when an antagonist is steam rolling the crew. So, let's take the latter situation into account. If an antagonist has just fucked over command and security so badly that an admin is considering spawning in players as a response team in order to progress the round or give the antagonist an actual challenge, is the Occupying Force ERT really going to be sufficient? Think about the situations where an ERT is called in, blob, cult, murderboney traitors. The Occupying Force ERT will not be enough in these situations, considering they're more poorly armed than the Amber ERT which almost always fails due to their lack of equipment. You say it's a way of introducing weaker sec reinforcements, but in my own opinion they wouldn't be enough in most situations and overall it's my opinion that calling in response teams should not be common for most rounds, but that's just me. I like the idea though, it is very thought out and they seem like their own unique force which is cool.

Hopekz:
This PR hasn't been touched for a while, and I don't feel like constantly waiting to modify something.
Applying a tag isn't good enough if nothing happens as a result.

JamieD1:
"Lets See what happens"

PR #10350, opened 2020-11-07
shadowling void jaunt cooldown drastically lowered

Marmio64:
Lowers sling's void jaunt cooldown to 25 seconds instead of a way over the top 80 second cooldown. However, to balance it, you lose stamina bit faster when jaunting in the light. 

#### Changelog

:cl:  
tweak: lower void jaunt cooldown, but slightly higher light jaunting stamina cost
/:cl:

### DIFF
--- a/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
+++ b/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
@@ -702,7 +702,7 @@
 	name = "Void Jaunt"
 	desc = "Move through the void for a time, avoiding mortal eyes and lights."
 	panel = "Shadowling Abilities"
-	charge_max = 150
+	charge_max = 250
 	clothes_req = FALSE
 	antimagic_allowed = TRUE
 	phase_allowed = TRUE
--- a/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
+++ b/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
@@ -702,7 +702,7 @@
 	name = "Void Jaunt"
 	desc = "Move through the void for a time, avoiding mortal eyes and lights."
 	panel = "Shadowling Abilities"
-	charge_max = 800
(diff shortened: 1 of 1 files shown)


JamieD1:
Feels like a bit of an extreme jump for a little bit of a nerf.

Marmio64:
> Feels like a bit of an extreme jump for a little bit of a nerf.

void jaunt is pretty useless but i'll bring it to 25 seconds cooldown if that will help

ghost:
15 second is good, @Marmio64 keep it at that

for comparison @JamieD1 the antagonist nightmare gets a 0 cooldown jaunt that can be used as many times as the nightmare wants on command and heals it while using it

shadowling jaunt actively does stamina damage to it and will lock it down for 10 seconds if it overextends, meaning the sling is slowed after using it even for a couple tiles

nightmare, the midround antagonist, is literally infinitely more mobile than the shadowling is and is able to slip into departments with much greater ease because it can just jaunt in and smash a light where the sling relies on its cooldowns and the poor capabilities of its abilities

ghost (approved):
code's fine

Marmio64:
> 15 second is good, @Marmio64 keep it at that
> 
> for comparison @JamieD1 the antagonist nightmare gets a 0 cooldown jaunt that can be used as many times as the nightmare wants on command and heals it while using it
> 
> shadowling jaunt actively does stamina damage to it and will lock it down for 10 seconds if it overextends, meaning the sling is slowed after using it even for a couple tiles
> 
> nightmare, the midround antagonist, is literally infinitely more mobile than the shadowling is and is able to slip into departments with much greater ease because it can just jaunt in and smash a light where the sling relies on its cooldowns and the poor capabilities of its abilities

Well actually jamie has a point. Unlike nightmare jaunt slings can jaunt in the light. Void jaunting also actually gets rid of stam damage too when you use it. I could just give slings nightmare jaunt and remove void jaunt though

Marmio64:
going to give slings nightmare jaunt instead

PR #10351, opened 2020-11-07
Gives ashwalkers a pair of science goggles and a hand labeler

Marmio64:
They don't fit into the tribal theme of ashwalkers but the organization these things provide is very nice.

intent: Gives ashwalkers a better way to organize

Purpose: Lets ashwalkers organize their stuff better. I did not specifically test it but it is a very simple change so there is no reason for it not to work.

Why is it good for the game?: Lets ashwalkers organize their stuff better.

Why is it beneficial?: Lets ashwalkers organize their stuff better, what more can I say? Ashwalkers can't organize their things well, so this lets them do it.

#### Changelog

:cl:  
rscadd: Gives ashwalkers a pair of science goggles and a hand labeler
/:cl:

### DIFF
--- a/_maps/RandomRuins/LavaRuins/lavaland_surface_ash_walker1.dmm
+++ b/_maps/RandomRuins/LavaRuins/lavaland_surface_ash_walker1.dmm
@@ -464,20 +464,6 @@
 /obj/item/construction/rcd/loaded,
 /turf/open/indestructible/boss,
 /area/ruin/unpowered/ash_walkers)
-"bn" = (
-/obj/structure/closet/crate/radiation,
-/obj/item/flashlight/lantern,
-/obj/item/flashlight/lantern,
-/obj/item/flashlight/lantern,
-/obj/item/flashlight/flare,
-/obj/structure/stone_tile/block/cracked{
-	dir = 8
-	},
-/obj/structure/stone_tile/cracked{
-	dir = 1
-	},
-/turf/open/floor/plating/asteroid/basalt/lava_land_surface,
-/area/ruin/unpowered/ash_walkers)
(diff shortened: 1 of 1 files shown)


patpol4:
pr gives ahswalkers instagib rifles

Hopekz:
I think both of these items don't fit with the theme of them being tribal, not being able to read or speak common. Science goggles require them to interpret text on their display and using a hand labeler requires them to know the operation of the device + write in common to spell out the items.

I think if you wanted to introduce this you should make something closer to the "ashwalker" tribal theme instead of just straight up giving them the modern tools to do the job.

Marmio64:
> I think both of these items don't fit with the theme of them being tribal, not being able to read or speak common. Science goggles require them to interpret text on their display and using a hand labeler requires them to know the operation of the device + write in common to spell out the items.
> 
> I think if you wanted to introduce this you should make something closer to the "ashwalker" tribal theme instead of just straight up giving them the modern tools to do the job.

they can write and speak in common, and i'll likely replace the items with tribal versions later on if i can come up with sprites and explanation for how they work. This isn't a excuse, but they already get plenty of advanced tools like a medkit, gauze, and a hyper advanced welder too.

patpol4:
The items are stolen, i believe this is fine, why those specific items though?

Marmio64:
> The items are stolen, i believe this is fine, why those specific items though?

why the hand labeler and science goggles? Maybe they just felt like stealing them. Why I chose the two items to put them in? Because they're some of the best ways to organize.

patpol4 (approved):
Map if fine, I'll leave the merging to HC since this is technically a balance thing and got downvotes

PR #10356, opened 2020-11-08
Shadowling empowered thralls now get veil

Marmio64:
Empowered thralls still aren't really worth it, considering one of the main advantages of thralls is that they can blend in pretty well and don't have to die to lights, but empowered thralls lose both of these advantages, while supposedly being a 'upgrade'. This changes that to make them a viable upgrade since they now get veil, and an actual reason to empower thralls. However, to balance it, the maximum number of empowered thralls is now only 3 instead of 5, so each shadowling can basically have one empowered thrall if they distribute fairly.

#### Changelog

:cl:  
rscadd: Empowered thralls now get the veil (lights) ability.
tweak: max amount of empowered thralls is now 3 rather than 5
/:cl:


fluffe9911 (approved):
seems fine

PR #10362, opened 2020-11-09
mesalikepie came crying to me about ez clean grenades and it turns out i forgot to consider you can use an oddyseus to synthesize infinite amounts of the chemical leading to it being used to instant kill people but fortunately i am incredibly handsome and intelligent and immediately had a fix on hand for that

ghost:
also i may have been hoping that this would happen because it would be really funny but that would be a lie because i genuinely didn't think this would happen

also it's down to 1 on reac volume instead of 1.5 so it shouldn't quite oneshot even in normal beaker form but now that you can't produce infinity of it you can't use it to wipe out the entire station

if it keeps being a problem i can just remove the reac volume effect entirely but this is fine otherwise

(this doesn't really change that being splashed with a beaker of this will still fucking kill you)
#### Changelog

:cl:  
tweak: fixes theos abusing waffle co chems
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -1052,6 +1052,7 @@
 	description = "A powerful, acidic cleaner sold by Waffle Co. Affects organic matter while leaving other objects unaffected."
 	metabolization_rate = 1.5 * REAGENTS_METABOLISM
 	taste_description = "acid"
+	can_synth = FALSE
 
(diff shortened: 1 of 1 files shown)


Mesalikepie:
I'm not wrong

ghost:
yeah hes not wrong ngl

theos wtf

Mqiib:
Ok but it was hilarious the first time it was used and dude just dies instantly. Really warranted the nerf lmao.

SomeguyManperson (approved):
when are we banning this theos guy

PR #10378, opened 2020-11-10
internal affairs agent death pill is much stronger

Marmio64:
This gives the internal affairs agent pill some coniine and initropidril so that it kills you within about 10-20 seconds instead of just like, not killing you at all, since cyanide is super fucking weak. This is good for the game since it allows IAA to actually kill themselves.

#### Changelog

:cl:  
tweak: IAA pill is a fuckton more deadly
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/IAA/internal_affairs.dm
+++ b/code/modules/antagonists/traitor/IAA/internal_affairs.dm
@@ -30,18 +30,18 @@
 
 /datum/antagonist/traitor/internal_affairs/on_gain()
 	START_PROCESSING(SSprocessing, src)
-	
+
 	if(ishuman(owner.current))
-		//Gives Cyanide dental implant
-		var/obj/item/reagent_containers/pill/cyanide/cyan = new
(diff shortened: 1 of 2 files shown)


UselessTheremin:
even though nobody ever uses it, if they were to use it, it would work. good change.

Hopekz:
...why not just buff cyanide?
Backward solution

Marmio64:
> ...why not just buff cyanide?
> Backward solution

cyanide is super easy to make, you can't make it very strong

alexkar598:
beaker, water, syringe, syringe gun


honk

SomeguyManperson:
I mean if you want to get the pill out of your mouth if you can even do that you'd need to have someone surgically remove it I think, and you can just buy a poison kit for the same poisons but in higher volumes so having some poisons in the pill that aren't easily attainable shouldn't be too terrible

redmoogle (approved):
Code is good

alexkar598:
yeah fair enough, forgot it was an implant now

JamieD1 (approved):

PR #10386, opened 2020-11-11
King Goat Plushie for Traitor Clerk

Magicut101:

Adds the King Goat plushie to the clerk's uplink at 10 TC. 

The Clerk currently has no role-restricted traitor items unlike many of the other jobs on the station and lacks important accesses making them essentially a glorified assistant with better starting gear and costumes. The King Goat Plushie is only currently obtainable from mass buying plushie crates or finding them in the starting presents in your shop which is exceptionally rare. 
By adding the King Goat Plushie as a traitor item purchasable from the uplink we get to see the item being used more often for hilarious effect and buff the Clerk's uplink options.

It comparison to its closest competition the Esword the King Goat Plushie fares like this:
+Deals more damage when it eats cabbage and becoming an Ascended King Goat plushie.
+Is inconspicuous in sec searches so its unlikely for it to removed from your inventory or for you to be called a traitor. 
+Does not have to be sheathed nor does it make the esword's distinctive sound. The goat sounds when used may be as loud as a esword though.
+Everyone will respect you for holding the One True King in your hands. 
=Is the same size at small as the Esword allowing it to be stored in boxes.
=Crits in 4 hits the same as an esword.
-Does less damage than an esword in base form.
-Does not have the ability to delimb so you will be unable to cut heads off to prevent revivals. 
-When dropped it will jump and attack others including yourself could have niche uses but is more likely to screw you over.

#### Changelog

:cl:  
rscadd: Added new things  
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1891,6 +1891,12 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	item = /obj/item/hierophant_antenna
 	restricted_roles = list("Shaft Miner")
 
+/datum/uplink_item/role_restricted/king_goat_plushie
+	name = "King Goat Plushie"
+	desc = "This plushie contains an inkling of the King Goat's Power."
+	item = /obj/item/toy/plush/goatplushie/angry/kinggoat
+	cost = 10
+	restricted_roles = list("Clerk")
 // Pointless
 /datum/uplink_item/badass
 	category = "(Pointless) Badassery"


jcatjmeow (changes requested):
Capture
Does not work.

Hopekz:
I don't like giving exclusive rewards in different places to devalue the reward in the first place.
The idea "fits" but it makes the reward and fights potentially even more pointless than it already is as now it can just be spawned in.

jcatjmeow (approved):

LazennG:
YES LET US GIVE KING GOAT WHICH CAN DO 30 DAMAGE AND KNOCKDOWN TO THE CLERK FOR 10 TC

Magicut101:
So what would be more reasonable 12-14 TC? I am not sure 16 TC would be reasonable because you could get more useful items for that amount of TC or just a double energy blade. 

Edit:
though I suppose since the item is pretty hilarious it could be justified at 20 tc and leave it as a rare hilarious and deadly option.

boodaliboo:
i like the item and i like more role exclusives but this item is as good or better then a esword and can be upgraded to do even more i feel like its good as a megafauna reward right now

PR #10392, opened 2020-11-12
the transit tube in 10x5_transit.dmm is now functional

jcatjmeow:
It was facing the wrong direction

#### Changelog

:cl:  
bugfix: the transit tube in 10x5_transit.dmm is now functional
/:cl:


patpol4 (approved):
If you say so

PR #10395, opened 2020-11-12
Atmospherics to Turbine Shortcut

FatalEYES:

From what I understand, atmos techs either make their own little space bridge from atmos to the turbine when they're doing fusion, this should just save them the trouble.

It literally just saves most atmos techs tedious time doing their own basic mapping changes. More time to have fun with fusion experiments.

#### Changelog

:cl:  
rscadd: Added new atmos space bridge/room. Added a holopad to the turbine room as well.
tweak: tweaked the atmos airlock to be more aesthetically pleasing with the new bridge/room. (Thicc airlock boiiii) Also made the gravgen airlock bolted by default because it should be imo.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -30315,15 +30315,6 @@
 "bJq" = (
 /turf/open/floor/engine/air,
 /area/engine/atmos_distro)
-"bJs" = (
-/obj/effect/mapping_helpers/airlock/cyclelink_helper{
-	dir = 1
-	},
-/obj/machinery/atmospherics/components/unary/vent_pump/layer1{
-	dir = 1
-	},
-/turf/open/floor/plating,
-/area/engine/atmos_distro)
 "bJv" = (
 /obj/machinery/portable_atmospherics/canister/oxygen,
 /obj/machinery/atmospherics/miner/oxygen,
@@ -32675,19 +32666,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/science/nanite)
-"bXy" = (
-/obj/machinery/atmospherics/pipe/simple/scrubbers/visible{
-	dir = 4
-	},
-/obj/machinery/atmospherics/pipe/manifold/scrubbers/hidden/layer3{
-	dir = 1
-	},
-/obj/machinery/atmospherics/pipe/simple/purple/hidden,
-/obj/machinery/atmospherics/pipe/manifold/supply/hidden/layer1{
-	dir = 1
-	},
-/turf/open/floor/plasteel,
-/area/engine/atmos_distro)
 "bXA" = (
 /obj/effect/turf_decal/tile/blue{
 	dir = 4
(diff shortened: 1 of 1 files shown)


Hopekz:
The idea of a shortcut is alright but using double external doors is a bit weird.
This is the only place this is a thing for external airlocks on the station so it would create inconsistency.

Here is what I'm referring to for people that want screenshots since you didn't post any:
image

PrinceKirze:
ehhhhhhhhhhhhhh
looks strange. 
the double airlock im not crazy about.
the glass connects weirdly to the turbine room.
the glass is strange in general.
idk im not a mapper and while i like the idea of a connection this one is done really weirdly.

FatalEYES:
I do agree with the window into the turbine and that is actually something I missed. The double airlock though is that way so techs are able to drag cannisters into the station from space since now it's actually two tiles wide, but feedback is definitely welcome.

MrHorizons:
I've tried making a shortcut from atmospherics to the incinerator as here, but I've found that a big problem is that it cuts off external access to the right side of tanks. I've personally never really seen them used, but I don't think it'd be very good to block it off.

The most I've found you can make a shortcut to the incinerator is to have a second atmospherics maintenance door roughly above where it is.

MrHorizons:
> I do agree with the window into the turbine and that is actually something I missed. The double airlock though is that way so techs are able to drag cannisters into the station from space since now it's actually two tiles wide, but feedback is definitely welcome.

Also, I think the planned standard is to two-_long_ airlocks.

MrHorizons:
As for feedback on the map itself...
image
No. Just no. Yeet that window and replace it with a wall.

For the floors, keep the colors consistent and make it just a plasteel turf rather than a dark plasteel turf.

FatalEYES:
Yeah the window was absolutely an oversight I forgot to change is all, I'll be doing in a bit.

patpol4:
:thinking: 
chrome_GceMQvEBF2

FatalEYES:
> The idea of a shortcut is alright but using double external doors is a bit weird.
> This is the only place this is a thing for external airlocks on the station so it would create inconsistency.
> 

Okay, I did alter the external airlock a bit, since it did odd at a second glance.

> looks strange.
> the double airlock im not crazy about.
> the glass connects weirdly to the turbine room.
> the glass is strange in general.

Made some changes, hopefully that handled a few of your concerns.

> I've tried making a shortcut from atmospherics to the incinerator as here, but I've found that a big problem is that it cuts off external access to the right side of tanks. I've personally never really seen them used, but I don't think it'd be very good to block it off.

Yeah, I was trying to figure out what to do with that for a good while. Then I realized that they can just remove the window and work with the shutter and throw down a holobarrier.

FatalEYES:
Screenshot_3
For those who want to see it here.

FatalEYES:
> For the floors, keep the colors consistent and make it just a plasteel turf rather than a dark plasteel turf.

I personally like the darker floors for what's essentially just a sort of storage room, like toxins storage. But I do see where you're coming from.

TheLinuxOS:
I dont like this for a few of the reasons mentioned above, firstly the double door external airlock, and also the fact that it blocks access to the right side gas area. It is much easier in my opinion for an atmos tech to just remove the window and place a holofan if they want to quickly go between the two.

Jarod1200:
This looks good to me. Aside from the double door airlock being a bit odd

PR #10417, opened 2020-11-14
the sleepypen is better at ko'ing a target but is not refillable

ghost:
the sleepypen being refillable made it the most lethal traitor melee weapon in the game

previously the most common tactic with the sleepypen was to dispose of its starting chemicals and then just fill it with poisons. it would be more accurately called the poison pen, because it would be used as an instantaneous 45u hypospray that would be used to empty horrific amounts of deathmix into someone

it should now basically flatten someone in seconds but it's not refillable-it's a one shot pen, no refills, like it used to be back when it was loaded with zombie powder.

this means it can be used to quickly knock someone out, as intended, rather than being the best assassination tool in the game. if you want to kill someone with chemicals, use the dartgun or the reagent syringegun, or poison them another way.

:cl:  
tweak: The sleepypen will now knock targets out much more efficiently, but is no longer refillable.
/:cl:

### DIFF
--- a/code/modules/paperwork/pen.dm
+++ b/code/modules/paperwork/pen.dm
@@ -164,10 +164,12 @@
 
 /obj/item/pen/sleepy/Initialize()
 	. = ..()
-	create_reagents(45, OPENCONTAINER)
+	create_reagents(75)
 	reagents.add_reagent(/datum/reagent/toxin/chloralhydrate, 20)
 	reagents.add_reagent(/datum/reagent/toxin/mutetoxin, 15)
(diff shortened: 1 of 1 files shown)


redmoogle (approved):
Looks good, although the pancronium will probably kill them.

ghost:
@redmoogle it's not enough pancuronium to be lethal

ghost:
wait, actually, you're right, it just might be, i'll switcheroo a number

JamieD1 (approved):

PR #10427, opened 2020-11-16
Adds A Bone Pickaxe

Magicut101:
Intent of your Pull Request
Adds a Bone Pickaxe to the primal tab in crafting

Why is this good for the game?
I want to see this added for cases where you didn't have or lost your pickaxe. This would allow you to be able to make a pickaxe from scratch by killing fauna. 

Bone Pickaxe costs 3 bone and 2 sinew and mines as fast as a normal pickaxe.

I tested this here is my proof

proof
proofitworks2
proofitworks3
BonePickaxe4
🆑
rscadd: Added new things
imageadd: added some icons and images
/🆑


Marmio64:
maybe reduce sinew cost to two? If it is worse than a normal pickaxe it sounds like it'd be absolutely dogshit and shouldn't cost that much.

Magicut101:
Aight I was thinking of keeping it low but wasn't sure what would be appropriate so I left it be a tad high. Its now 2

code/modules/mining/equipment/mining_tools.dm
```diff
@@ -50,6 +50,17 @@
 	desc = "A pickaxe with a diamond pick head. Extremely robust at cracking rock walls and digging up dirt."
 	force = 19
 
+obj/item/pickaxe/bonepickaxe
+	name = "bone pickaxe"
+	desc = "A spine and ribcage tied tightly with weaver sinew...atleast its better than your bare hands at digging up dirt."
```
Hopekz:
```suggestion
	desc = "A spine and ribcage tied tightly with weaver sinew...at least it's better than your bare hands at digging up dirt."
```

Hopekz (changes requested):

jcatjmeow (approved):

Smudged:
ashwalkers should start with 1 or 2 two of these instead of the stock pickaxes

PR #10430, opened 2020-11-16
Port: Admins can now see chronological order of events in logging tabs.

uomo91:
## About The Pull Request
Original PRs here:
https://github.com/tgstation/tgstation/pull/54939
https://github.com/tgstation/tgstation/pull/54997

Individual logging (the logging one sees when they view a player's playerpanel and click on LOGS) was broken and poorly coded, this fixes and improves it. The Show All tab was notably completely broken (it only showed attack logs), due to a rounding error from text2num trying to parse a bitflag value. (Thanks @LemonInTheDark for unraveling that mystery) Also in this PR, I made various different type of logs visually distinguishable from regular say's.

Most importantly: **logs will now be sorted chronologically** regardless of category, and tabs that list multiple categories (like the say tab, comms tab, and show all tab) will concatenate them all instead of displaying them all separately.

Previous behavior, using the say tab as an example was:
[show all regular say logs here, in order]
[then below all of them, show all whispers, in order]
[then below all of both of the above, show all deadchat in order]

The above was stupid, this is way better.

log

logcombat

A current limitation is that due to this relying on sortTim(), things are sorted by time stamp.
There is a non-zero chance that if somebody has more than one log within the same second, that they could display in the wrong order.
I don't see this as much of an issue, because between ping and people's reaction time, how often do things even take place in the same second? And how often would their exact order be the key to dealing with a ticket?

Also, self-admitted room for improvement: colors and formatting.

In summary, a lot of the admin tools are laggy and clunky and held together with duct tape, and there is much room for refactor and improvement. A TGUI redo of most of them would be great, for instance. But, this PR addresses one isolated matter that alone will greatly improve quality of life for admins, and thus for players also.

## Why It's Good For The Game

This will make investigating issues incredibly much less hassle for the admin team.
No more endlessly switching back and forth between attack and say tabs in the logs for multiple different people, no more feverishly memorizing timestamps or copy and pasting them into notepad windows.

#### Changelog

:cl:  
bugfix: Fixes and improves some things with logging, to help out admins.
/:cl:

### DIFF
--- a/code/__DEFINES/flags.dm
+++ b/code/__DEFINES/flags.dm
@@ -142,3 +142,6 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 #define RELIGION_TOOL_SECTSELECT 	(1<<2)
 
 #define NSCOMPONENT(d)   (d&(NORTH|SOUTH))
+
+/// 33554431 (2^24 - 1) is the maximum value our bitflags can reach.
+#define MAX_BITFLAG_DIGITS 8
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -1041,8 +1041,6 @@
 /// Helper for logging chat messages or other logs with arbitrary inputs (e.g. announcements)
 /atom/proc/log_talk(message, message_type, tag=null, log_globally=TRUE, forced_by=null)
 	var/prefix = tag ? "([tag]) " : ""
-	if(message_type == LOG_WHISPER)
-		prefix += "whispers "
 	var/suffix = forced_by ? " FORCED by [forced_by]" : ""
 	log_message("[prefix]\"[message]\"[suffix]", message_type, log_globally=log_globally)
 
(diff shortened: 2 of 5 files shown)


uomo91:
Tested this and it should be good to go, marking as ready for review.

alexkar598:
Now if only admins used logs instead of relying on replays...

uomo91:
> Now if only admins used logs instead of relying on replays...

They were... "less than useful" before, to put it politely, so I can't say I blame them. This PR I feel at least lessens the idiotic busywork of using them a lot.
There is so much more room for improvement in the admin tools, especially logging, but to go further with how I think it could be made better would mandate a near total overhaul.
Maybe one day.

JamieD1:
> Now if only admins used logs instead of relying on replays...

Talking like a typical linux nerd.

YES LETS MAKE THINGS 5 TIMES AS DIFFICULT

alexkar598:
Il regex a file faster than you can find a timestamp by seeking through a replay thank you. Replays work but are slow, sadly sometimes its the only location with the data needed

PR #10456, opened 2020-11-19
buffs ancient sabre to do 10 damage from6

boodaliboo:
its a sword that was worse then a normal knife
#### Changelog

:cl:  
tweak: tweaked ancient sabres stats

/:cl:


patpol4:
It's supposed to be weak, it's there as a prop, not loot.

jcatjmeow:
As the creator of said sabre I confirm. It's storytelling, not a weapon.

PR #10461, opened 2020-11-19
Chaplains should no longer become vampires.

Xoxeyos:
Chaplains should not be born to die. Chaplains no longer suffer from vampirism.

:cl:  
bugfix: Chaplains are now added to the restricted jobs list for the vampire gamemode.
/:cl:


Marmio64:
what do you mean my devout Christian chaplain can't right away become a dark unholy monster of hell?

Amelia0010:
If they're becoming vampires then this isn't the fix. The protected jobs and restricted jobs variables are functionally identical, see:
https://github.com/yogstation13/Yogstation/blob/11930117d5397ad7283eebb4b000c9ae2cf93af8/yogstation/code/game/gamemodes/vampire/vampire.dm#L42

PR #10484, opened 2020-11-22
[PORT] Adds sound selection to certain instruments

redmoogle:
**Quote**

Ports: https://github.com/tgstation/tgstation/pull/53690

### About The Pull Request

This gives instruments a selection of fitting instrument sounds, since we have a bunch that have been exclusive to the synth piano until now. For example, the harmonica can pick between the hardcoded and synthetized versions.

It might also add a new special and rare instrument [but that's not important](https://www.youtube.com/watch?v=6dybQg4t8Qc)

### Why It's Good For The Game

New instruments arguably sound better than the old hardcoded ones but some people might have different preferences and it's not worth having five different types of guitars.

#### Changelog

:cl:  
tweak: You may now change the sound of certain instruments with a selection of appropriate variations, similarly to the piano synth
/:cl:


patpol4:
bless

PR #10500, opened 2020-11-25
Syndicate Brainwashing Denial Kits (Syndicate Mindshield Implants) now show up on the sechud.

Xoxeyos:
I wanted to see if this would be used more often if I toggled the visibility for it, this would usually be used for a dynamic round, or similar strange scenarios, but I'm wondering if I toggle it, people may use it more for stealth.

:cl:  Xoxeyos
tweak: SBDK's are now visible on the sechud.
/:cl:

### DIFF
--- a/code/game/objects/items/implants/implant_mindshieldtot.dm
+++ b/code/game/objects/items/implants/implant_mindshieldtot.dm
@@ -21,7 +21,7 @@
 		if(!target.mind)
 			ADD_TRAIT(target, TRAIT_MINDSHIELD, "implant")
 			target.sec_hud_set_implants()
-			return FALSE
+			return TRUE
 
 		if(target.mind.has_antag_datum(/datum/antagonist/brainwashed))
 			target.mind.remove_antag_datum(/datum/antagonist/brainwashed)


Jarod1200:
Is this like they show up as mindshielded or they have a special indicator

TheLinuxOS:
Yes that is indeed what it is Jarod, shows up like a normal mindshield

boodaliboo:
wouldnt this be a instant giveaway

Xoxeyos:
> 
> 
> wouldnt this be a instant giveaway

Well, say for example if you want to be your very own CC disk inspector as a nuclear operative, or just blend in with security easier, toggling the visibility would help, I'm seeing what people want. Usually this implant would be used primarily in dynamic, and it's difficult to tell why you aren't able to be converted by X gamemode, so this is to test the waters, and to make it more apparent to the players as to WHY you can't be converted.

Tipy1802:
Time to disguise myself as a sec officer and kill my target with little to not resistance, this time even sec would think I am an actual officer!

JamieD1 (approved):
Lets give it a go

PR #10517, opened 2020-11-27
fixes faster movement with more arms when you have no legs

SomeguyManperson:

probably will work

Improves RP

#### Changelog

:cl:  
bugfix: you now do not move faster than normal speed when you have more arms
/:cl:


Marmio64:
but... fly theos...

PR #10518, opened 2020-11-27
prevents building a BSA on reebe

SomeguyManperson:

it was funny like once

Increases RP

#### Changelog

:cl:  
bugfix: you can no longer build a BSA on Reebe
/:cl:

### DIFF
--- a/code/modules/goals/station_goals/bsa.dm
+++ b/code/modules/goals/station_goals/bsa.dm
@@ -93,6 +93,8 @@
 		return "Parts misaligned!"
 	if(!has_space())
 		return "Not enough free space!"
+	if(is_reebe(z))
+		return "[text2ratvar("Location Unavailable)]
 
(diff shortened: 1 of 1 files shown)


code/modules/goals/station_goals/bsa.dm
```diff
@@ -93,6 +93,8 @@
 		return "Parts misaligned!"
 	if(!has_space())
 		return "Not enough free space!"
+	if(is_reebe(z))
+		return text2ratvar("Location Unavailable")
```
alexkar598: Shouldnt this be a ratvar message in that case? Something like Not on my watch

JamieD1:
```suggestion
		return text2ratvar("Not on my watch")
```

JamieD1 (approved):

PR #10529, opened 2020-11-28
Fixes synthflesh and changes slightly how it works

SomeguyManperson:

Synthflesh now heals 2 damage per unit, but can only heal when applied in doses up to 60 units, putting its effective healing amount at 120 health before needing to purge it

Increases RP

#### Changelog

:cl:  
bugfix: synthflesh now works
tweak: synthflesh now has a healing cap at 60 units and heals 2 points of health up from 1.25 per unit 
/:cl:


maxion12345:
why does nobody just make extinguishers only apply like, .01 of a unit

PR #10552, opened 2020-11-30
Breakrooms Volume 1: Supply Edition

TheReddDragon:
This PR adds the first breakroom I planned on adding, the **Supply Breakroom**! Grab some cigs, a drink of cola, and a snack and interact with your coworkers. I PR this in hope it might better contribute towards the push to MRP. After all, it makes no sense for departments on a space station to not have a break room!
#### Changelog

:cl:  
rscadd: adds cargo breakroom where maint access used to be. Can still access maint from here.
tweak: tweaks locker room to accommodate cargo breakroom
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -21644,6 +21644,10 @@
 	},
 /turf/open/floor/plasteel,
 /area/quartermaster/storage)
+"bkq" = (
+/obj/item/twohanded/required/kirbyplants/random,
+/turf/open/floor/plating,
+/area/maintenance/port)
 "bkr" = (
 /obj/machinery/atmospherics/pipe/simple/supply/hidden/layer2{
 	dir = 4
@@ -32973,6 +32977,22 @@
 	},
 /turf/open/floor/plasteel/dark,
 /area/engine/engine_smes)
+"cnO" = (
+/obj/structure/cable{
+	icon_state = "4-8"
+	},
+/obj/machinery/atmospherics/pipe/simple/supply/hidden/layer2{
+	dir = 4
+	},
+/obj/structure/disposalpipe/segment{
+	dir = 4
+	},
+/obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer4{
+	dir = 4
+	},
+/obj/machinery/vending/cigarette,
+/turf/open/floor/plasteel,
+/area/maintenance/port)
 "cnP" = (
 /obj/machinery/power/terminal{
 	dir = 1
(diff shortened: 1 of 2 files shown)


FatalEYES:
Screenshot_1
For those who want to see it.

FatalEYES:
While I like the idea of breakrooms, most places that don't already have one, simply don't have the room to dedicate to it, sadly.

patpol4:
This is really bad.
Fix the decals on the walls to the north, remove the file change from areas.dm.
Plants in the maint corridor are weird, there is too little space in the room itself. 4 vending machines and two tables may be a bit of an overkill. Also makes maint straight.
Ill give it a closer look tomorrow

TheReddDragon:
iambadatgit.exe

Xoxeyos:
I'd say remove the plants in maintenance, those aren't needed there, replace them with something else, like lockers, or something maintenance worthy.

TheReddDragon:
> I'd say remove the plants in maintenance, those aren't needed there, replace them with something else, like lockers, or something maintenance worthy.

already done see https://github.com/yogstation13/Yogstation/pull/10552/commits/a354324bccebab1bf028f0e3f99d7499564eb41c

MrHorizons:
Also, I would suggest removing the wooden mugs and replacing them with actual drinking glasses.

Hopekz:
I don't like it and here is why:



- You replaced a variation maint spawner. That removed spawner offered lots of variation with our previous maint PR's. I would prefer to still have those variations so they'd probably need to be moved somewhere else.
- The area you are replacing seems to be more thought out. The new area in your PR seemed squished just to make this new area fit.
- The bathroom now feels very claustrophobic. Like it doesn't accommodate a space station because you had to remove 2 whole stalls and all the sinks you had before. The remaining detail in the bathroom is very squished like it was an afterthought.
- There is less detail in the maints around your new area (I think there used to be glass on the floor etc). The grille and glass panels added a unique area to maints that is now replaced with a  generic "hallway" looking area.
- I don't like that the vending machines are placed in the walking area. If you go from door to door you have to dodge 2 vending machines making the breakroom itself feel claustrophobic.
- You added more reinforced walls into cargo but for some reason chose to not wall off the area? Just looks weird in my opinion like you implemented half an idea. Previously the reinforced walls lined up with the exterior cargo area.
- The mugs just seem like fluff. What are they going to put in those mugs? There is no other liquid in that area. Maybe remove them since you added vending machines?
- This would be the first break room I've seen without disposals. Very odd considering it is an area where garbage is made.
- You added the abandoned cobweb marker to a high traffic airlock (right out of the break room). Cobwebs are used to symbolize doors less often used and for consistency are not used in the usual path of travel (in box at least I don't care about other maps). This also means that sometimes you won't even have an airlock there as it can sometimes be a wall or even bolted to the best of my knowledge.
- generic floors that blend in to everything else

image
image
image

patpol4:
Basically what Hopek said among some other things

patpol4:
I tried my hand at making this better and this is what i came up with: 
Improvement to your location
chrome_1NgCdP2ViT
A more open location in the bay itself
chrome_MM1FKyIvJp

Hopekz:
> I tried my hand at making this better and this is what i came up with: 
> Improvement to your location
> chrome_1NgCdP2ViT
> A more open location in the bay itself
> chrome_MM1FKyIvJp

Are you making an an alternative PR for it?
Or did you just make it for the screenshot then throw it away

patpol4:
> Are you making an an alternative PR for it?
> Or did you just make it for the screenshot then throw it away

uh

PR #10556, opened 2020-12-01
port Pituitary Disruption

boodaliboo:
original pr: https://github.com/BeeStation/BeeStation-Hornet/pull/1203
"Pituitary Disruption (level 7):
At base level, makes you bigger, scaling with stage speed. at 8 stage speed, causes passive brute healing, at the cost of occasionally vomiting up blood and guts. at 12 stage speed, quadruples the healing, but causes you to slowly die of cell damage, and puke up more advanced tumors, which may come to life and attack! (not sure what the brute heal/cell damage ratio per tick should be. at the moment it's 4 brute heal/ .2 cell damage, not sure if this should be changed. it also unpredictably knocks you down while you have it and makes your hitbox larger, so that should be enough downside...)"
#### Changelog
more viro stuff the better

:cl:  
rscadd: Added new viro symptom

/:cl:


ghost:
>Virus that makes you vomit griefing monkey subtypes

boodaliboo:
> > Virus that makes you vomit griefing monkey subtypes

i could easily remove the player control part

fluffe9911 (changes requested):
if they were simple mobs sure but I dont like the idea of a virus being able to create sentient monkey subtypes

edit: also the goon meat cube is better

boodaliboo:
> if they were simple mobs sure but I dont like the idea of a virus being able to create sentient monkey subtypes
> 
> edit: also the goon meat cube is better

no longer sentient

SomeguyManperson:
This doesn't really feel like a symptom thhat would be used outside haha funny vomit since ths stats seem fairly loe and it can cause more permanent damage

PR #6878, opened 2019-10-14
removes sandbox gamemode

Amelia0010:
### Intent of your Pull Request

I haven't gone and checked, but I am honestly willing to bet that this gamemode hasn't been run more than 10 times in the last 3 years.

#### Changelog

:cl:  
rscdel: sandbox is gone
/:cl:

(I bet most of you didn't even know this gamemode existed)

PR #10567, opened 2020-12-01
Revert "removes sandbox gamemode"

Bibby0110:
Reverts yogstation13/Yogstation#6878

No need to delete stuff.
Good gamemode for various admin-events. Causes no problems.


redmoogle:
Explain why.

PR #10577, opened 2020-12-03
Fixes Anti-magic sometimes not working, fixes Runic Bombs double-exploding sometimes

Altoids1:
image

### The Conspiracy

So it turns out, that every singular ../projectile/magic has an onhit override that contains the same 5-ish lines of code. Some versions even forget to call ``qdel(src)`` or call crappy versions of the ``anti_magic_check()`` proc. No more!

As well, all of it is now handled in ``prehit`` instead of ``on_hit``, which should fix some issues regarding spells ignoring antimagic.

### Runic Bombs

Something I found while doing this was a poor bit of code that causes runic bombs to explode twice:
```dm
    if(ismob(target))
        var/mob/M = target
        if(M.anti_magic_check())
            M.visible_message("<span class='warning'>[src] vanishes on contact with [target]!</span>")
            qdel(src)
            return BULLET_ACT_BLOCK
        else
            explosion(M, -1, 0, boom, 0, 0)
    if(iscarbon(target))
        var/mob/living/carbon/X = target
        ADD_TRAIT(X, TRAIT_NODISMEMBER, type)
        ADD_TRAIT(X, TRAIT_SLEEPIMMUNE, type)
        ADD_TRAIT(X, TRAIT_STUNIMMUNE, type)
        spawn(5)
            REMOVE_TRAIT(X, TRAIT_NODISMEMBER, type)
            REMOVE_TRAIT(X, TRAIT_SLEEPIMMUNE, type)
            REMOVE_TRAIT(X, TRAIT_STUNIMMUNE, type)
            X.adjustBruteLoss(-120)
        explosion(X, -1, 0, boom, 0, 0)
```
So that's been fixed, too, as a treat. This might result in these bombs doing less damage, for what that's worth.

# Changelog

:cl: Altoids
bugfix: Anti-magic now works correctly against arcane barrage, spellblade, and more!
bugfix: Runic bombs no longer double-explode when the target is a carbon.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -996,10 +996,16 @@
 
 	apply_effect((amount*RAD_MOB_COEFFICIENT)/max(1, (radiation**2)*RAD_OVERDOSE_REDUCTION), EFFECT_IRRADIATE, blocked)
 
-/mob/living/anti_magic_check(magic = TRUE, holy = FALSE, tinfoil = FALSE, chargecost = 1, self = FALSE)
-	. = ..()
-	if(.)
+///Return any anti magic atom on this mob that matches the magic type
+/mob/living/proc/anti_magic_check(magic = TRUE, holy = FALSE, tinfoil = FALSE, chargecost = 1, self = FALSE)
+	if(!magic && !holy && !tinfoil)
 		return
+	var/list/protection_sources = list()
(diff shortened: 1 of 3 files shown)


Marmio64:
speedmerge

Altoids1:
Tested it with the two magicthings that Marmio complained about, as well as the Wabbajack staff. All three work, so I'm guessing it works for every magic projectile I affected via this PR.

redmoogle (approved):
Based, good code cleanup

PR #10586, opened 2020-12-04
Tritium is more dangerous/useful by reducing o2 output by 10x

redmoogle:

Reduce the amount of o2 released from the sm(at 100% tritium) by 10x

Makes tritium more spooky and also useful instead of burning instantly

~~will fix commit in sec~~

#### Changelog

:cl:  
tweak: Tritium now reduces o2 output by up to 10x
/:cl:


redmoogle:
tis all fucked. I'll refactor this code first

PR #10591, opened 2020-12-05
fire_immunity_max_temp_protect increased to account for fusion bullshit

ghost:
certain fires could bypass the heat of even an atmospherics hardsuit/mining hardsuit etc

this has fixed that-things that are fireproof are now fireproof, no ifs ands or buts. if you manage to exceed this temperature the game is probably going to break shortly thereafter.


Mqiib:
Adjust the max temp for firefighter mechs too, please.

ghost:
This still won't let you survive a fusionflood, which are extremely rare, and fusionv7 already addresses the problem of a fusionflood anyway.

PR #10595, opened 2020-12-05
Makes the WT rifle incapable of being dual wielded

SomeguyManperson:
yes we can do this actually

improves RP

#### Changelog

:cl:  
tweak: the WT autorifle can no longer be dual wielded
/:cl:


Moja10:
This is what the townhall was about, this was the RP change we were proposing the entire time

maxion12345:
Wish we just had CM style two-hand wield guns :(

PR #10605, opened 2020-12-07
increases the realism of the ark spaghetti generation by removing the tomato sauce

SomeguyManperson:
improves immersion 
:cl:
tweak: the ark now generates plain spaghetti insetad of spaghetti with tomato sauce, tomatoes will have to be sourced on station
/:cl:


Mqiib:
Lore-friendly pr is very good thank you Ratvarian loremaster thoseose

Hopekz:
I like this PR because it will lead us to MRP getting tomato sauce out of the ark is LRP

PR #10607, opened 2020-12-07
adds heart of freedom as a possible organ drop from meaty ores

boodaliboo:
its a cool organ with seemingly no way of getting it so thought why not add it to the other way to get cool organs
#### Changelog

:cl:  
rscadd: Added heart of freedom to meaty ores
/:cl:


Marmio64:
i thought the heart of freedom gives you a few uses of the freedom implant for breaking cuffs? If so, this sounds like it'd be too powergamey without being unique (like xeno organs)

boodaliboo:
> i thought the heart of freedom gives you a few uses of the freedom implant for breaking cuffs? If so, this sounds like it'd be too powergamey without being unique (like xeno organs)

no it gives you a small ammount of healing chems when your in crit

JamieD1 (approved):

PR #10613, opened 2020-12-08
artists now get payed as much as assistants

boodaliboo:
why would artists which are typically impoverished be payed significantly more then a assistant, they are now on equal paygrades
#### Changelog

:cl:  
tweak: tweaked artists pay
/:cl:


fluffe9911:
they should make no money by default and survive purely on portrait commissions

boodaliboo:
> they should make no money by default and survive purely on portrait commissions

if only people actually used money

JamieD1 (approved):

PR #10619, opened 2020-12-08
felenids now take damage from coco powder and hot chocolate

boodaliboo:
cats and dogs are both mammals so this makes sense
#### Changelog

:cl:  
rscadd: Added toxin damage when cat people consume hot coco and coco powder
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/felinid.dm
+++ b/code/modules/mob/living/carbon/human/species_types/felinid.dm
@@ -107,9 +107,9 @@
 /datum/species/human/felinid/handle_chemicals(datum/reagent/chem, mob/living/carbon/human/H)
 	. = ..()
 	if(H.reagents.has_reagent(/datum/reagent/consumable/coco))
-		M.adjustToxLoss(1, 0)
+		H.adjustToxLoss(1, 0)
 		
 /datum/species/human/felinid/handle_chemicals(datum/reagent/chem, mob/living/carbon/human/H)
 	. = ..()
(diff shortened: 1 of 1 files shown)


boodaliboo:
will fix later just need to add a toxpwr to the base chem-

Mqiib:
Just swap it from M.adjust to H.adjust. also good pr improves immersion,

Moja10:
My God, is this the MediumRP PR we have been desperately needing the entire time?

boodaliboo:
> My God, is this the MediumRP PR we have been desperately needing the entire time?

yes

jcatjmeow:
Make it so dogpeople die when they drink coffee

monster860 (changes requested):
this is shitcode, do `istype(chem, whatever)` not `H.reagents.has_reagent(whatever)` like every other handle_chemicals proc.

The reason is this check runs for every reagent in the system. With your shitcode, if there are 10 different kinds of reagents in someone and one of them is hot coco/coco powder, then they'll take 10 Toxin damage per tick, which is really really OP and was not intended that way.

boodaliboo:
> this is shitcode, do `istype(chem, whatever)` not `H.reagents.has_reagent(whatever)` like every other handle_chemicals proc.
> 
> The reason is this check runs for every reagent in the system. With your shitcode, if there are 10 different kinds of reagents in someone and one of them is hot coco/coco powder, then they'll take 10 Toxin damage per tick, which is really really OP and was not intended that way.

i copied alexkars code i also have no clue what your asking

PR #10622, opened 2020-12-09
a different nuka cola nerf - SEAN UPDATE NEEDED

Marmio64:
disclaimer, this should not be combined with Redd's PR, do not merge both.

This reduces nuka cola's speed buff by 33% and makes it deal a slight amount of stamina damage that is largely unnoticeable so long as you don't drink more than 25 units in a sitting or are getting shot by disablers and it stops your stamina regeneration. Reinforces the idea of nuka cola being a cheap and only mildly effective getaway drug rather than a combat drug since now it can hamper you when fighting against people with stamina draining weapons. Radiation amount is kept untouched.

#### Changelog

:cl:  
tweak: nuka cola speedbuff isn't as good, and it deals slight amounts of stamina damage
/:cl:

Jamie Edit: 
Alternative to #10621
Alternative to #10629
Alternative to #10629

THEOS SORT THIS OUT

### DIFF
--- a/code/modules/reagents/chemistry/reagents/drink_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/drink_reagents.dm
@@ -462,7 +462,7 @@
 
 /datum/reagent/consumable/nuka_cola/on_mob_metabolize(mob/living/L)
 	..()
-	L.add_movespeed_modifier(type, update=TRUE, priority=100, multiplicative_slowdown=-0.75, blacklisted_movetypes=(FLYING|FLOATING))
+	L.add_movespeed_modifier(type, update=TRUE, priority=100, multiplicative_slowdown=-0.50, blacklisted_movetypes=(FLYING|FLOATING))
 
 /datum/reagent/consumable/nuka_cola/on_mob_end_metabolize(mob/living/L)
 	L.remove_movespeed_modifier(type)
@@ -476,6 +476,7 @@
 	M.AdjustSleeping(-40, FALSE)
 	M.adjust_bodytemperature(-5 * TEMPERATURE_DAMAGE_COEFFICIENT, BODYTEMP_NORMAL)
 	M.apply_effect(5,EFFECT_IRRADIATE,0)
+	M.adjustStaminaLoss(1)
 	..()
 	. = 1
 


jcatjmeow (approved):

boodaliboo:
im fine with the nerfs but if your lowering speed make the rads slightly less

Georgeur:
The speed wasn't the big issue, and it means you can't escape perma with it if you lower it this much. I'm pretty down with the stam damage.

ghost:
The first is that radiation damage is irrelevant. Rads don't matter. Potassium iodide can be taken in bulk to counter small radiation damage provided by radiation-inflicting drinks or poisons such as polonium, et cetera, as potas iodide has no downside and no overdose. Furthermore, antitoxin, et cetera, can be utilized to counter the effects of radiation. On top of this, rad-immune races, such as plasmamen and slimepeople, can use the drink without consequence. 

The second is that increasing the rad damage also doesn't fix this problem. If you double the rad damage but it still takes multiple drinks to kill you, it's a free speedboost. If it kills you in one sip, it's a traitor poison. There is no in-between-chemical poisoning is a binary. It's either insanely lethal, like heparin or lexorin, or completely harmless, like cyanide. If you increase the rad damage, all this does is gate it to exclusively be used by powergamers and chemists, and chemists will use it because it's a free speedboost without any risks-unlike ephedrine, which has a low addiction, and meth, which causes brain damage and thus brain traumas if not dosed properly, and it's also not a narcotic, so security can't prosecute you for having it-and nobody will get on the bartender's ass for having a 'bar drink' and so it's the favored drink of choice for powergamers.

Removing it from the vendor or increasing recipe difficulty also doesn't resolve the problem for the above reasons. If you increase recipe difficulty, it again gatelocks it to a powergamer or a chemist, which is basically synonymous-the only thing you can actually DO to recipe lock something is make it require an unobtainable, uncraftable recipe like omnizine. Otherwise making it need 'a chem' from chemistry is irrelevant.

In the end, all of those issues combined basically means no matter what, it's either free, or it kills you so fast it's a traitor poison.

If you keep it in with its speedboost as an effect, it will be used as consequence-free methamphetamine no matter what.

Marmio64:
> The first is that radiation damage is irrelevant. Rads don't matter. Potassium iodide can be taken in bulk to counter small radiation damage provided by radiation-inflicting drinks or poisons such as polonium, et cetera, as potas iodide has no downside and no overdose. Furthermore, antitoxin, et cetera, can be utilized to counter the effects of radiation. On top of this, rad-immune races, such as plasmamen and slimepeople, can use the drink without consequence.
> 
> The second is that increasing the rad damage also doesn't fix this problem. If you double the rad damage but it still takes multiple drinks to kill you, it's a free speedboost. If it kills you in one sip, it's a traitor poison. There is no in-between-chemical poisoning is a binary. It's either insanely lethal, like heparin or lexorin, or completely harmless, like cyanide. If you increase the rad damage, all this does is gate it to exclusively be used by powergamers and chemists, and chemists will use it because it's a free speedboost without any risks-unlike ephedrine, which has a low addiction, and meth, which causes brain damage and thus brain traumas if not dosed properly, and it's also not a narcotic, so security can't prosecute you for having it-and nobody will get on the bartender's ass for having a 'bar drink' and so it's the favored drink of choice for powergamers.
> 
> Removing it from the vendor or increasing recipe difficulty also doesn't resolve the problem for the above reasons. If you increase recipe difficulty, it again gatelocks it to a powergamer or a chemist, which is basically synonymous-the only thing you can actually DO to recipe lock something is make it require an unobtainable, uncraftable recipe like omnizine. Otherwise making it need 'a chem' from chemistry is irrelevant.
> 
> In the end, all of those issues combined basically means no matter what, it's either free, or it kills you so fast it's a traitor poison.
> 
> If you keep it in with its speedboost as an effect, it will be used as consequence-free methamphetamine no matter what.

it is still practically consequence free but now it will at least be slightly less favorable since you don't get as massive of a speed boost and using it actively hinders you in combat against anyone with a stamina draining weapon since you'll be unable to regenerate stamina while under it's effects.

ghost:
Which is why I favor removing the speedboost entirely

PR #10624, opened 2020-12-09
Chemical heaters now scale from max_temprature not target temprerature 

redmoogle:

Makes it so the equation isnt (target - current) and instead (1000 - current) * -1(or1)

Takes about ~8 seconds for a unupgraded heater to hit 380K
Should take about ~2 seconds at Tier 4

_Tested AND works backwards_

Muh Realism. Also prevents just bypassing the need for heating by have chemicals be 1 tick
_also fixes a slight bug where it would never reach target temperature_

#### Changelog

:cl:  
tweak: Chem heaters now scale based on max temperature
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/machinery/chem_heater.dm
+++ b/code/modules/reagents/chemistry/machinery/chem_heater.dm
@@ -66,7 +66,7 @@
 		return
 	if(on)
 		if(beaker && beaker.reagents.total_volume)
-			var/direction = (beaker.reagents.chem_temp > target_temperature) ? TRUE : FALSE // TRUE is cooling, used to allow it to snap to the target temp
+			var/direction = beaker.reagents.chem_temp > target_temperature // is it cooling? used to allow it to snap to the target temp
 			var/heating = (1000 - beaker.reagents.chem_temp) * heater_coefficient * (direction ? -1 : 1) // How much to increase the heat by
 			if(heating + beaker.reagents.chem_temp >= target_temperature && !direction) // Heat snapping condition
 				heating = target_temperature - beaker.reagents.chem_temp // Makes it snap to target temp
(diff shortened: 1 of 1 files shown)


athena413:
muh speed muh danger *meth explodes*

good pr

code/modules/reagents/chemistry/machinery/chem_heater.dm
```diff
@@ -9,7 +9,7 @@
 	circuit = /obj/item/circuitboard/machine/chem_heater
 	var/obj/item/reagent_containers/beaker = null
 	var/target_temperature = 300
-	var/heater_coefficient = 0.1
+	var/heater_coefficient = 0.025 // Multiplier for heat
```
Bibby0110: Autodoc

code/modules/reagents/chemistry/machinery/chem_heater.dm
```diff
@@ -65,7 +65,14 @@
 		return
 	if(on)
 		if(beaker && beaker.reagents.total_volume)
-			beaker.reagents.adjust_thermal_energy((target_temperature - beaker.reagents.chem_temp) * heater_coefficient * SPECIFIC_HEAT_DEFAULT * beaker.reagents.total_volume)
+			var/direction = (beaker.reagents.chem_temp > target_temperature) ? TRUE : FALSE // TRUE is cooling, used to allow it to snap to the target temp
```
Bibby0110: No need for a ternary operator

Bibby0110 (changes requested):
turd code

don't like that if statement but i'm too lazy to figure out an elegant way to remove it

code/modules/reagents/chemistry/machinery/chem_heater.dm
redmoogle: Yes there is, read.

jcatjmeow (approved):

code/modules/reagents/chemistry/machinery/chem_heater.dm
Bibby0110: No, there isn't :)

PR #10625, opened 2020-12-09
Revert "removes sandbox gamemode"

Bibby0110:
Reverts yogstation13/Yogstation#6878


Amelia0010:
If your event consists of "use +spawn to do whatever you want" then you really need to come up with a better event.

maxion12345:
> If your event consists of "use +spawn to do whatever you want" then you really need to come up with a better event.

It'd be fun as hell though. Think of all the things players rarely get to play with because it never happens in a round.

PR #10626, opened 2020-12-09
Adds sandbox but without the elements that make it unfeasable to run

alexkar598:

If we want sandbox, I think this is the only way its acceptable

It will be beneficial as there would be any ability to add another mode called sandbox

#### Changelog

:cl:  
rscadd: Added sandbox but you can't spawn items
/:cl:


boodaliboo:
so extended

alexkar598:
Y

alexkar598:
Wrong button

PR #10633, opened 2020-12-09
Replaces the nuka cola speed boost with a damage slowdown resistance

SomeguyManperson:

causes nuka cola to give you a 50% resistance to damage slowdown rather than a passive speed boost, making it more useful in protracted combat or for running away rather than just running

Removes greytide speedboost drink also improves RP

#### Changelog

:cl:  
rscdel: nuka cola no longer gives a speed  boost when drank
tweak: nuka cola now gives a 50% resistance to damage slowdown
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -73,6 +73,7 @@
 #define TRAIT_PACIFISM			"pacifism"
 #define TRAIT_IGNORESLOWDOWN	"ignoreslow"
 #define TRAIT_IGNOREDAMAGESLOWDOWN "ignoredamageslowdown"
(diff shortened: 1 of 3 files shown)


Moja10:
I was hesitant with this solution, but given that it helps RP, I am all for it.

ghost:
This makes it more viable than morphine, as morphine knocks you out, which very little can protect against

SomeguyManperson:
morphine also gives full immunity to damage slowdown while this will cause it to be lessened. You're still moving slower just not as slow as you would be without it.

JamieD1 (approved):

PR #10641, opened 2020-12-11
adds cola vendor to every permabrig

boodaliboo:
alternative to theos pr that is much better
#### Changelog

:cl:  
rscadd: Added cola vendor to permabrig
/:cl:


patpol4:
Not every permabrig and stupid, they can't even buy it in there.

boodaliboo:
> Not every permabrig and stupid, they can't even buy it in there.

that is wrong

PR #10649, opened 2020-12-13
Matches the frequency of the torture device in the oldAIset ruin

Hopekz:
The frequency on the remote doesn't match the electropack in the room meaning it won't work as intended.

8lu18f14de
image

#### Changelog

:cl:  Hopek
bugfix: fixed the frequency of the torture device in the oldAIset ruin to make it work.
/:cl:


patpol4:
You have to match the code variable aswell, pretty sure

Hopekz:
> You have to match the code variable aswell, pretty sure

Ah. Good catch I'll fix it the next time I'm at a PC

patpol4:
Checks failing

patpol4 (approved):
should work now

PR #10650, opened 2020-12-13
Minor variation framework + brig cell variations

Hopekz:
Adds a code variation framework for when the variations you are adding are too small to bother creating a map file in an effort to avoid clutter.

For example, in this PR it was used to store cell beds being mirrored and cell furniture being rearranged.

bSm9Xur27r
LECcl4mN4k

**I also made pretty icons for the whole system to look nice on the map editor.**

image

M0XLSyiiw1

**All of this is logged too for debugging**

nMBGe8sKpp

#### Changelog

:cl:  Hopek
rscadd: Adds a code variation framework for when the variations you are adding are too small to bother creating a map file in an effort to avoid clutter.
rscadd: Used the new code variation framework to add brig cell variations and perma cell variations.
/:cl:


Bibby0110:
This just seems like it's not going to be used since it's unintuitive as balls

Hopekz:
> This just seems like it's not going to be used since it's unintuitive as balls

I'd be fine with somebody replacing my solution with something better in the future but this is what I have right now.

There is no other way to do this at the moment besides making a map file for each variation no matter how small it is.

These brig variations would create 6 extra map files in our current system as an example

patpol4 (approved):
Not sure how it works until i get to an editor but map appears fine from diffs, pretty cool.

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+	if(!x)
+		x = 0
+	if(!y)
+		y = 0
+	if(item)
+		location_place = locate((ref_x + x) , (ref_y + y) , ref_z)
+		if(islist(item))
+			for(var/I in item)
+				if(direction)
+					var/obj/H = new I(location_place)
```
Bibby0110, line 32:
```suggestion
					var/atom/H = new I(location_place)
```

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+					H.dir = direction
+					return
```
Bibby0110:
```suggestion
					
```
probably don't want this?

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+		else
+			if(direction)
+				var/obj/H = new item(location_place)
+				H.dir = direction
+			else
+				new item(location_place)
+	return
+
+/obj/effect/variation/proc/choice()
+	if(choice == "" || choice == 0)
```
Bibby0110:
```suggestion
	if(!choice)
```

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+					new I(location_place)
+		else
+			if(direction)
+				var/obj/H = new item(location_place)
+				H.dir = direction
+			else
+				new item(location_place)
+	return
+
+/obj/effect/variation/proc/choice()
```
Bibby0110, line 45: autodoc

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+	var/location_place // location of placement calculated from reference
+	var/ref_x // x of reference
+	var/ref_y // y of reference
+	var/ref_z // z of reference
+
+/obj/effect/variation/Initialize()
+	. = ..()
+	chosen = FALSE
+
+/obj/effect/variation/proc/place(x, y , item , direction)
```
Bibby0110: autodoc

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
+/*
+Makes variations possible with code only.
+This is mainly to prevent creating new map files when creating smaller variations that aren't worth a map file. - Hopek
+*/
+
+/obj/effect/variation
+	name = "variation"
+	icon = 'icons/effects/landmarks_variation.dmi'
+	icon_state = "generic_green"
+	var/variations // number of variations
```
Bibby0110, line 10: fix all these to be autodoc

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+	var/choice // chosen variation by choice()
+	var/chosen // has a choice been made for this place?
+	var/location_place // location of placement calculated from reference
```
Bibby0110, line 13: no need for this to be a var on the thing i think

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+	. = ..()
+	chosen = FALSE
+
+/obj/effect/variation/proc/place(x, y , item , direction)
+	if(!x)
+		x = 0
+	if(!y)
+		y = 0
+	if(item)
+		location_place = locate((ref_x + x) , (ref_y + y) , ref_z)
```
Bibby0110, line 28:
```suggestion
		var/turf/location_place = locate((ref_x + x) , (ref_y + y) , ref_z)
```

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110:
```suggestion
/obj/effect/variation/proc/place(x = 0, y = 0 , item , direction)
```

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+	var/ref_x // x of reference
+	var/ref_y // y of reference
+	var/ref_z // z of reference
+
+/obj/effect/variation/Initialize()
+	. = ..()
+	chosen = FALSE
+
+/obj/effect/variation/proc/place(x, y , item , direction)
+	if(!x)
```
Bibby0110, line 23: if you change the defaults then there's no need for these, (it's your own fault if you put in null as the coordinate)

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+		x = 0
+	if(!y)
```
Bibby0110, line 25: if you change the defaults then there's no need for these, (it's your own fault if you put in null as the coordinate)

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+					return
+				else
+					new I(location_place)
+		else
+			if(direction)
+				var/obj/H = new item(location_place)
+				H.dir = direction
+			else
+				new item(location_place)
+	return
```
Bibby0110:
```suggestion
	
```

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+		choice = 1
+	if(!chosen)
+		ref_x = src.x
+		ref_y = src.y
+		ref_z = src.z
+		qdel(src) // deletes the landmark after storing location values in choice() so it doesn't show up on the final product
+		chosen = TRUE
+		choice = rand(1,variations)
+		log_world("Chosen variation [choice] out of [variations] for [name]. Spawning on [ref_x],[ref_y],[ref_z].")
+		return choice
```
Bibby0110:
```suggestion
		
```

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+	else
```
Bibby0110:
```suggestion
	
```

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+		return choice
```
Bibby0110:
```suggestion
	return choice
```

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+/obj/effect/variation/proc/choice()
+	if(choice == "" || choice == 0)
+		choice = 1
+	if(!chosen)
+		ref_x = src.x
+		ref_y = src.y
+		ref_z = src.z
+		qdel(src) // deletes the landmark after storing location values in choice() so it doesn't show up on the final product
+		chosen = TRUE
+		choice = rand(1,variations)
```
Bibby0110:
```suggestion
		choice = rand(1, variations)
```

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+				H.dir = direction
+			else
+				new item(location_place)
+	return
+
+/obj/effect/variation/proc/choice()
+	if(choice == "" || choice == 0)
+		choice = 1
+	if(!chosen)
+		ref_x = src.x
```
Bibby0110, line 49: this is probably obvious in hindsight, but why src.x instead of x?
same for below

code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -0,0 +1,58 @@
...
+		ref_y = src.y
+		ref_z = src.z
+		qdel(src) // deletes the landmark after storing location values in choice() so it doesn't show up on the final product
```
Bibby0110, line 52: I wonder if you should just put this in Initialize() and move the chosen = FALSE to the declaration

(And then place the . = ..() in the actual variation objects at the bottom)

((might be missing something, enlighten me if I am))

code/game/objects/effects/spawners/minor_variation_templates.dm
```diff
@@ -0,0 +1,75 @@
+// sec
```
Bibby0110:
```suggestion

```
Change it to something better, "sec" just... just bad

code/game/objects/effects/spawners/minor_variation_templates.dm
```diff
@@ -0,0 +1,75 @@
...
+	var/list/bed = list(
+		/obj/structure/bed,
+		/obj/item/bedsheet/prisoner
+	)
+	var/closet = /obj/structure/closet/secure_closet/brig/cell
+
+	switch(choice())
+		if(1)
+			//variation 1 , default;  bed left side
+			place(0,0,bed)
```
Bibby0110:
```suggestion
			place(0, 0, bed)
```

code/game/objects/effects/spawners/minor_variation_templates.dm
```diff
@@ -0,0 +1,75 @@
...
+			place(2,0,closet)
```
Bibby0110:
```suggestion
			place(2, 0, closet)
```

code/game/objects/effects/spawners/minor_variation_templates.dm
```diff
@@ -0,0 +1,75 @@
...
+	
+		if(2)
+			//variation 2 , bed right side
+			place(2,0,bed)
```
Bibby0110:
```suggestion
			place(2, 0, bed)
```

code/game/objects/effects/spawners/minor_variation_templates.dm
```diff
@@ -0,0 +1,75 @@
...
+			place(0,0,closet)
```
Bibby0110:
```suggestion
			place(0, 0, closet)
```

code/game/objects/effects/spawners/minor_variation_templates.dm
```diff
@@ -0,0 +1,75 @@
...
+		/obj/structure/table,
+		/obj/item/paper,
+		/obj/item/pen
+	)
+	var/stool = /obj/structure/chair/stool
+	
+	switch(choice())
+		if(1)
+			//variation 1 ; Default . 
+			place(0,1,bed)
```
Bibby0110:
```suggestion
			place(0, 1, bed)
```

code/game/objects/effects/spawners/minor_variation_templates.dm
```diff
@@ -0,0 +1,75 @@
...
+			place(2,0,table)
```
Bibby0110:
```suggestion
			place(2, 0, table)
```

code/game/objects/effects/spawners/minor_variation_templates.dm
```diff
@@ -0,0 +1,75 @@
...
+			place(2,1,stool)
```
Bibby0110:
```suggestion
			place(2, 1, stool)
```

code/game/objects/effects/spawners/minor_variation_templates.dm
```diff
@@ -0,0 +1,75 @@
...
+
+		if(2)
+			//variation 2 ; bed right corner
+			place(2,1,bed)
```
Bibby0110: and so on and so on, does wonders for readability

Bibby0110 (changes requested):
code looks fine.

obviously won't be very intuitive and will requires mappers using it to... touch code.
Could have probably done it in a list with coordinates + items/objects, but that's not very intuitive either, oh well

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: Is there a point?
The object gets deleted as soon as the spawning is done.

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: What's the difference at least on the top its easier to find the var but the object gets deleted after the spawning happens.

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: Well yeah but I don't really trust people not to do something weird. 
Some time years in the future, someone is going to mess up and cause runtimes.
I'll do the x=0,y=0 to satisfy your need to make changes to my PR though.

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: Do you really think your comment above didn't cover this? 
Are you trying to rack up some total on my PR?

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: This change is unneeded.

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: This change is unneeded.

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: If you can make up any good excuse for this being different from you like the word "atom" more I'll change it.

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: Can't hurt; more clarity is always good.

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: It is best practice to include the object being acted upon for clarity's sake.
Essentially for readability's sake of making sure other coders that usually work in different languages can easily make the distinction

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: No need. You're proposing changes just to propose changes at this point.
Deleting it here works and I verified it.

code/game/objects/effects/spawners/minor_variation_templates.dm
Hopekz: Alright.

code/game/objects/effects/spawners/minor_variation_templates.dm
Hopekz: It's literally just a 3 letter comment. 
What do you want this to be something like // ===== SECURITY  =====

code/game/objects/effects/spawners/minor_variation.dm
Hopekz: I made it pretty anyway

Hopekz:
> Could have probably done it in a list with coordinates + items/objects, but that's not very intuitive either

I am fine with my solution being completely replaced the moment we get a more patient coder that will make a more intuitive and complex solution. I am not this coder

I was actually at first thinking of doing a "put the spots where a variation can occur then just make a list of things that can be variations and it'll cycle between those random spots".

So for the brig example, it would have been "I know I have to place an item from this list on the left and right side. I'll pick 1 item from the list randomly for each spot and remove them from the list"

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110: Atom is the base type.
Dir comes from atom
You should inherit from the base type since you're using nothing from obj

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110: You're in a loop, if direction is set then the loop only places 1 item from the list

Bibby0110:
> > Could have probably done it in a list with coordinates + items/objects, but that's not very intuitive either
> 
> I am fine with my solution being completely replaced the moment we get a more patient coder that will make a more intuitive and complex solution. I am not this coder
> 
> I was actually at first thinking of doing a "put the spots where a variation can occur then just make a list of things that can be variations and it'll cycle between those random spots".
> 
> So for the brig example, it would have been "I know I have to place an item from this list on the left and right side. I'll pick 1 item from the list randomly for each spot and remove them from the list"

I don't think there's a perfect way to do this anyhow so this is better than nothing


EDIT: Autodocs are required I guess, don't personally care but jcat might get angry idk

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110: Yeah it's just in a weird place imo

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110: Fair

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110: No your return is unneeded.
The function automatically returns when it ends, this is just clutter

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110: Yeah totally

((You're just competent therefor this seems unneeded since it's logical))

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110: Yeah but this is one of those things that you'd have to do REALLY wrong, like runtime worthy wrong
I'd rather want a runtime at that point so we don't get weird stuff where stuff starts spawning at the relative 0/0

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110: You should declare the var when you need it so it's easy to see that it's not being used for anything else. Doing it in the declaration might cause someone to think it's used for something else (Which is what I did initially)

code/game/objects/effects/spawners/minor_variation.dm
Bibby0110: See my comment below

patpol4:
Just noticed this needs very specific code, i doubt it's going to get used often, if at all

Hopekz:
> Just noticed this needs very specific code, i doubt it's going to get used often, if at all

I can see this used for smaller variations that aren't worth 2+ maps files each like 1-3 items being in a different mirrored orientation.
**quick Examples include:**

 Mirroring this room 
 
image

Mirroring lockers and sink for this room
image

Switching these lockers around:
image

Mirror
image
image


All of these things are something that this code type variation would help with because if we were to use our current system just to go through this quick list it would be 10 extra map files.

patpol4:
> All of these things are something that this code type variation would help with because if we were to use our current system just to go through this quick list it would be 10 extra map files.

Yes but the changes are so minor that they're not even worth doing. The really bad part about this would be having to change code for an object if you want to move something around in a room on a map. But you can't because the coordinates are in the code and it just adds extra steps to what you want to do

Hopekz:
> Yes but the changes are so minor that they're not even worth doing. 

The examples that I provided were quick and off the top of my head for the sake of providing examples.
I think changes such as these brig changes add up to a better experience.

It may seem minor on the surface, but doing something like undressing a prisoner and buckling them on the opposite side that you're used to is enough to provide a different experience. 

This is worthwhile to add, in my opinion, especially if it makes no negative impact on the player. Everything you're describing of potentially not being worth it is in the back end. This is akin to other games making systems for small and specific details.

patpol4:
> Everything you're describing of potentially not being worth it is in the back end. This is akin to other games making systems for small and specific details.

Good point, I'm just saying that this probably wont get used very often

Hopekz:
> Good point, I'm just saying that this probably wont get used very often

Who knows. If it collects dust like space pods and it needs to be removed for some reason all you'd have to do to revert this fully is to remove the minor_variation.dm and minor_variation_templates.dm files along with any landmarks.

It's a low-risk merge since it is self-contained in its own files.

JamieD1 (approved):

PR #10660, opened 2020-12-14
adds anesthetic to robotics lab: now in MRP-O-VISION!!

UselessTheremin:

adds an anesthetic tank and mask to the robotics lab because its a rule now

the new RP rules should be supported in the codebase, so yes, as much as I hate to admit it, this pr helps **INCREASE RP LEVEL 😳**

#### Changelog

:cl:  
rscadd: anesthetic tank and mask to the robotics lab
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -34048,15 +34048,6 @@
 	icon_state = "platingdmg3"
 	},
 /area/maintenance/aft)
-"cEL" = (
-/obj/structure/table,
-/obj/item/clothing/gloves/color/latex,
-/obj/item/storage/box/bodybags,
-/obj/effect/turf_decal/trimline/blue/filled/line{
-	dir = 4
-	},
-/turf/open/floor/plasteel/dark,
-/area/science/robotics/lab)
 "cFl" = (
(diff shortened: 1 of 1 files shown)


jcatjmeow:
very controversial

patpol4 (approved):
RP status: increased

patpol4:
also please change title to make sense so this shows up in search results for future generations

Hopekz:
Next step is to make it sub-optimal to perform surgery while conscious so it isn't just a rule but part of the gameplay

Nechronom:
Egad. What self-respecting follower of the great omnissiah would ever allow themselves to be put into a state of unconsciousness during the proceedings of the removing of pathetic flesh? But yes I agree that it should possibly have some effect on surgery chance. Maybe not all damage is restored to a patient after a surgery if they were not unconscious for it? Such as the amount of brute damage that is briefly given to someone when you're giving them xray eyes

boodaliboo:
we could add pain killers

Hopekz:
> we could add pain killers

Painkillers are pointless without pain

maxion12345:
> > we could add pain killers
> 
> Painkillers are pointless without pain

I'm gonna stop you right there.

PR #10670, opened 2020-12-17
makes it so that armblades cannot be melted with acid

UselessTheremin:

this was an oversight i discovered: you could melt both changeling armblades and most notably nightmare armblades with just some acid. for nightmares, this fucks them over very seriously, as they cannot get it back and essentially are just broken

oversight bad for the most part, fucking over antags with little to no effort can be pretty dumb

#### Changelog

:cl:  
tweak: changeling and nightmare armblades are now acidproof 
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/mutations.dm
+++ b/code/modules/antagonists/changeling/powers/mutations.dm
@@ -144,7 +144,6 @@
 	req_human = 1
 	weapon_type = /obj/item/melee/arm_blade
 	weapon_name_simple = "blade"
-	resistance_flags = ACID_PROOF
 
 /obj/item/melee/arm_blade
 	name = "arm blade"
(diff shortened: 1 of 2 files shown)


code/modules/antagonists/changeling/powers/mutations.dm
```diff
@@ -144,6 +144,7 @@
 	req_human = 1
 	weapon_type = /obj/item/melee/arm_blade
 	weapon_name_simple = "blade"
+	resistance_flags = ACID_PROOF
 
 /obj/item/melee/arm_blade
```
SomeguyManperson: resistance flags  variable should be placed inside this instead of its current spot

SomeguyManperson (approved):

PR #10706, opened 2020-12-22
Reinforces Incinerator with Reinforced Walls

alao3alao3:
Reinforces incinerator with Reinforced walls, because its part of the Atmos department.
cpng
First the door now this, did we just forget we made it part of Atmos?

#### Changelog

:cl:  
rscadd: Reinforced walls around the Incinerator, as in line with the rest of atmos.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -15772,12 +15772,6 @@
 /obj/structure/reagent_dispensers/water_cooler,
 /turf/open/floor/wood,
 /area/medical/psych)
-"aRB" = (
-/obj/machinery/atmospherics/pipe/simple/scrubbers/hidden{
-	dir = 4
-	},
-/turf/closed/wall,
-/area/maintenance/disposal/incinerator)
 "aRC" = (
 /obj/structure/table,
 /obj/machinery/atmospherics/components/unary/vent_scrubber/on/layer4{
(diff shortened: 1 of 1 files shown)


patpol4:
This actually might be bad because atmos techs usually need a way to make a space loop, adding rwalls makes it take way longer to get out

Xoxeyos:
> 
> 
> This actually might be bad because atmos techs usually need a way to make a space loop, adding rwalls makes it take way longer to get out

You can use RPEDs through walls, or so I believe.

patpol4:
> You can use RPEDs through walls, or so I believe.

yes but only on the wall, not beyond it in space, which is where the space loop usually resides

patpol4:
Why is this good for the game

alao3alao3:
uniformity and general peace of mind

patpol4:
This is really not needed in my opinion as incin isn't really something that needs to be kept secure and it denies atmos techs easy access to space which is vital for most incinerator setups.

MegaEmpirical:
If you don't think fusion and plasma fire should be secure I'm not sure you know what dangerous is.

Also, uniformity always good. 

Also, it is not that damn hard to take out a couple of rwalls. Good god that's so minor. And you don't even need to, RPD works fine on walls.

alao3alao3:
wait are people using the incinerator intended to break the walls?

PR #10710, opened 2020-12-22
significantly nerfs gas exports

ghost:
Obligatory part of toning down cargo's obscene credits generation. 

#### Changelog

:cl:  
tweak: Weakens cargo gas exports.
tweak: BZ exports are gone entirely.
/:cl:

### DIFF
--- a/code/modules/cargo/exports/large_objects.dm
+++ b/code/modules/cargo/exports/large_objects.dm
@@ -130,17 +130,12 @@
 	var/obj/machinery/portable_atmospherics/canister/C = O
 	var/worth = 10
 
-	worth += C.air_contents.get_moles(/datum/gas/bz)*4
-	worth += C.air_contents.get_moles(/datum/gas/stimulum)*100
-	worth += C.air_contents.get_moles(/datum/gas/hypernoblium)*1000
-	worth += C.air_contents.get_moles(/datum/gas/miasma)*10
-	worth += C.air_contents.get_moles(/datum/gas/tritium)*5
-	worth += C.air_contents.get_moles(/datum/gas/pluoxium)*5
-	worth += C.air_contents.get_moles(/datum/gas/freon)*15
-	worth += C.air_contents.get_moles(/datum/gas/hydrogen)*1
-	worth += C.air_contents.get_moles(/datum/gas/healium)*19
-	worth += C.air_contents.get_moles(/datum/gas/pluonium)*5
-	worth += C.air_contents.get_moles(/datum/gas/halon)*9
-	worth += C.air_contents.get_moles(/datum/gas/hexane)*6
-	worth += C.air_contents.get_moles(/datum/gas/zauker)*1050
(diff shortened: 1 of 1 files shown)


redmoogle:
inb4 downvote

ghost:
in my time of me playing QM. A atmos tech only given me A canister to sell once

ghost:
@Singularity69 Why does that matter

JamieD1:
This is insane.

ghost:
No, it's necessary.

The point of removing the budget cards and the economy entirely is to standardize cargo. The justification of 'but sec needs it for emergencies' is erroneous and fallacious-it justifies the ongoing arms race we have between antagonists and security, and is something that has no place in any form of an actual MRP server.

Currently, conversion antagonists are as strong as they are because security can shit out nearly infinite numbers of implant crates and guns, and thus they have to roll hard or cargo will materialize 100K points from the budget card, order a full total of 100 mindshield implants, and then they are fucked.

This justification also lead to the gradual powercreep of nukeops, as the mode increasingly centralized around warops. War ops got more and more powerful gear and crew got more and more shit they could export from cargo as a direct counterpoint to their power, to the extent that eventually blitz ops became 'unfair' rather than the normal because the gear war ops got as so powerful that only an obscenely strong cargo could stand a chance.

If we power down antagonists first then crew will steamroll them with cargo-Objectively we have to start with toning down cargo back to its earlier state where it wasn't possible to make 100K credits outside of extreme exploits.

That means we start with some of the biggest things first-gas exports being a primary case.

When cargo is at a point that hunting for crates is actually useful rather than a pointless banality and that cargo has a reason to stamp exports, then cargo prices will be balanced. End-goal is for cargo to be able to make no more than 40k-50k credits in a round, and to increase crew reliance on cargo in turn, while we also decrease the price of non-gun crates to make cargo more able to provide utilities and supplies.

In turn it's then safe to tone down antagonists accordingly and make them less competitive, as the crew will no longer be literally drowning in guns.

So we start here, by cutting atmos gases down. We can give atmos gases other crew utilities later. Perhaps they can be used to synthesize anomaly cores, or to generate research points. The existence of atmos crystals, for instance, is a good starting point.

But they can no longer be tied to cargo.

Darkstick:
reverse powercreep in order to make the game less fast-paced and murder thirsty

SomeguyManperson:
this does push for producing large amounts of any gas over producing more complex gasses but anything atmos makes tends to be mass produced anyways

patpol4:
This is fine in my opinion, might be a good step towards fixing the stupidity that is earning cargo credits right now

ToGWtF:
numbers are too low

just because ONE person can do it, doesn't mean its common place enough to make it not even worth doing anymore

ghost:
not relevant, tog

the point is to make it not worth doing anymore because i have larger plans in mind for cargo

JamieD1 (approved):

PR #10716, opened 2020-12-23
Increases the pipenet warning limit before muting itself

redmoogle:

Increase amount of errors it will log before muting itself

It's easier to read a bunch of false positives than it is to restart


patpol4:
This is mostly related to mapping and i agree with this

Bibby0110:
Best thing would be to make it a define though

redmoogle:
> Best thing would be to make it a define though

Pointless

JamieD1 (approved):
Define would be better, but eh

PR #10729, opened 2020-12-26
Nerfs defibs EMP effect

alexkar598:

This pains me as it is probably one of my most favorite weapons in game but I think this is needed, for reference, emagged/EMP'd defibs on disarm intent will immediately stun the victim and on harm intent after 6 seconds will stun the victim, give them 50 damage, stun them and give them a heart attack.

This PR removes the ability for EMPs to disable safeties on the defib(it will still reenable safeties if theyre currently disabled as a method to hard reset a defib if it gets emagged). Also it buffs the cell drain from EMPs because it was rather ridiculous, removing at most 1000 charge which is the cost of a single defib, now at most it removes 5000 charge at most(it is devided by the severity, with 1 being the strongest EMP you can get).

I've only seen them be used for valid hunting by people that shouldn't have access to stun weaponry(i have to admit, im guilty of it too). The weapon is honestly a bit overpowered even for traitors IMO but that's not the focus for this PR.

Also EMPs are rarely supposed to be doing anything good, there is no reason to NOT want to have your defib EMP'd.

#### Changelog

:cl:  
rscdel: EMPing a defib no longer turns the defib into a weapon.
/:cl:

### DIFF
--- a/code/game/objects/items/defib.dm
+++ b/code/game/objects/items/defib.dm
@@ -155,18 +155,20 @@
 
 /obj/item/defibrillator/emp_act(severity)
 	. = ..()
+	
 	if(cell && !(. & EMP_PROTECT_CONTENTS))
-		deductcharge(1000 / severity)
+		deductcharge(5000 / severity)
+		
 	if (. & EMP_PROTECT_SELF)
 		return
-	if(safety)
-		safety = FALSE
-		visible_message("<span class='notice'>[src] beeps: Safety protocols disabled!</span>")
-		playsound(src, 'sound/machines/defib_saftyOff.ogg', 50, 0)
-	else
+		
+	if(!safety)
 		safety = TRUE
 		visible_message("<span class='notice'>[src] beeps: Safety protocols enabled!</span>")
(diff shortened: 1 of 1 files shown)


ynot01:
emag is overtuned and other antags besides traitor exist, emped defib can be a very useful tool for antags for cheap/no tc

just remove the whole 'i click you and you are now stunned' disarm for both emp and emag

ghost:
Quite frankly, being that it occupies your entire backpack slot, it's balanced enough.

Rather than removing it, simply make it stun for 4 seconds, not 6.

alexkar598:
> 
> 
> Quite frankly, being that it occupies your entire backpack slot, it's balanced enough.

Unless you have the compact defib which is craftable from the medical tech lathe and fits on your belt and ***in*** your backpack

ghost:
@alexkar598 that's another note and it was powercreep that added the compact defib to the medlathe when it used to be cmo only

remove compact defib from lathe

problem solves itsel

JamieD1 (approved):

PR #10736, opened 2020-12-27
Fixes adding mentors not adding an admin log entry properly because whoever coded mentors does not know how to database

alexkar598:
Two down, 97 more SQL errors to go!


alexkar598:
SQL changes have been done on the database to support this bugfix

JamieD1 (approved):

PR #10745, opened 2020-12-28
Fixes minor variations executing over itself

Hopekz:
Fixes this by moving execution code out of initialize so it isn't forced to execute again
image

I had to add a parent call to fix Travis in my previous PR when I should have just moved it to its own proc in the first place.

Result: 
0eq6YE2tU7

#### Changelog

:cl:  Hopek
bugfix: Fixes minor variations executing over itself
/:cl:

### DIFF
--- a/code/game/objects/effects/spawners/minor_variation.dm
+++ b/code/game/objects/effects/spawners/minor_variation.dm
@@ -21,7 +21,7 @@ This is mainly to prevent creating new map files when creating smaller variation
 	placement()
 
 /obj/effect/variation/proc/placement()
-	..()
+	return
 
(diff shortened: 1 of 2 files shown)


code/game/objects/effects/spawners/minor_variation.dm
```diff
@@ -18,6 +18,10 @@ This is mainly to prevent creating new map files when creating smaller variation
 /obj/effect/variation/Initialize()
 	. = ..()
 	chosen = FALSE
+	placement()
+
+/obj/effect/variation/proc/placement()
+	..()
```
alexkar598: You are calling parent..... in a top level proc?

alexkar598 (dismissed):

JamieD1 (approved):

PR #10751, opened 2020-12-30
Removes skew from Cluwnes and replaces it with drugginess because apparently it's given Ezhan/Lei/Ethan4303 several BSODs and if it's given one person BSODs then it's probably given several other people BSODs and I'm like 99% sure crashing people's Operating System isn't very bueno but I don't fucking know I'm just a wikiman

LoliconSlayer:
title

:cl:  
bugfix: Cluwnes no longer Skew due to possibly causing people to Blue Screen
/:cl:


LoliconSlayer:
Adam asked me to replace it with some other HUD effect so I'll get to doing that eventually ig

jcatjmeow (approved):

PR #10752, opened 2020-12-30
Spraycanpuncture

ynot01:

spraycans give off a very small light explosion if you puncture them with a sharp object or screwdriver or surgical drill
hellcans give off a slightly larger but still largely nonlethal explosion
lubecans spray the area with lube

spray can (0,0,1,0)
hell can (0,0,2,5)
lube can (0,0,0,0 but 5 radius of lube)

(note: spray cans no longer give off flame like image says; hellcans still do)

powrgamers can now telekinesis around spray cans and screwdrivers to deal negligible damage so they get distracted from welderbombing

#### Changelog

:cl:  
rscadd: Added spraycans exploding when you puncture them
/:cl:

### DIFF
--- a/code/game/objects/items/crayons.dm
+++ b/code/game/objects/items/crayons.dm
@@ -680,6 +680,26 @@
 		spray_overlay.color = paint_color
 		add_overlay(spray_overlay)
 
+/obj/item/toy/crayon/spraycan/attackby(obj/item/S) //1 light 2 flame
+	if(istype(S, /obj/item/screwdriver))
+		explosion(get_turf(src), 0, 0, 1, flame_range = 2)
+		log_bomber(user, "detonated a", src, "via screwdriver")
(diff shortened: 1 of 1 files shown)


DerelictDrone:
It might be better if you check for the SHARP tag(see butchering code or cake/pizza slicing code) rather than explicitly screwdrivers, as this would increase the number of items able to puncture a can drastically(so we can stab the cans with carrot shivs, fire axes, bayonets, etc.)

ynot01:
> It might be better if you check for the SHARP tag(see butchering code or cake/pizza slicing code) rather than explicitly screwdrivers, as this would increase the number of items able to puncture a can drastically(so we can stab the cans with carrot shivs, fire axes, bayonets, etc.)

done

JamieD1 (approved):

PR #10768, opened 2020-12-31
Revert "meteors not effected by floating since they already move hella fast" since it maybe possibly is related to meteors lagging the game not sure WHY it would be but it MIGHT

SomeguyManperson:
Reverts yogstation13/Yogstation#9921


alexkar598:
I really don't see any way this can impact anything at all

PR #10802, opened 2021-01-06
Job skill framework + medical scanner example

Hopekz:
Adds a job skill framework. 
This can be added to literally any action in the game including interacting with items or clicking a hud button.
increases RP

**What this means:**

This gives the ability to make any job speciality better at their actual job vs some random guy not in your department. You can use this framework to add time delays to actions to represent people stumbling around trying to figure something out or block them from performing an action outright.

As an example of this in action I have used this framework to add a medical check to the health scanner. This means that someone not trained in how to use medical equipment will have to wait 3 seconds to scan something and 1 second to adjust the settings but someone trained will be able to do it instantly as per usual before this PR.

**TRAINED PROFESSIONAL:**
O5woJ23yUa

**A random person without medical job skills:**
notice the progress bar over the medical scanner
b5f5ScveAW

:cl:  Hopek
rscadd: Added a Job skill framework. This gives the ability to make any job speciality better at their actual job vs some random guy not in your department. You can use this framework to add time delays to actions to represent people stumbling around trying to figure something out or block them from performing an action outright.
rscadd: Used this framework to add a medical check to the health scanner. This means that someone not trained in how to use medical equipment will have to wait 3 seconds to scan something and 1 second to adjust the settings but someone trained will be able to do it instantly as per usual before this PR.
/:cl:

PR #10809, opened 2021-01-06
Adds a lightweight skill system that's less shitcode than hopek's implementation and makes mopping floors and doing surgery slightly faster for janitors and medical doctors respectively

alexkar598:
Closes #10802

Time required to mop floors as janitor is now 0.8 the normal time
Time required to do any surgery as a member of the medical department is 0.8 the normal time

To be honest, it doesnt really matter, its a proof of concept to show how my code is simpler than hopek's code and as easy to understand and implement as his code

# Wiki Documentation

Time required to mop floors as janitor is now 0.8 the normal time
Time required to do any surgery as a member of the medical department is 0.8 the normal time

:cl:  
tweak: Due to years of military training, janitors are now able to mop the floor in 80% of the time it would take a normal crew member to do
tweak: Due to years of pacific military training, medical doctors and their peers in the medical department are now able to do surgery in 80% of the time it would take a normal crew member to do
/:cl:

### DIFF
--- a/code/__DEFINES/jobs.dm
+++ b/code/__DEFINES/jobs.dm
@@ -96,3 +96,28 @@
 #define JOB_DISPLAY_ORDER_CHAPLAIN 39
 
 
+/proc/find_job(target)
+	//Get the job from the mind
+	if(istype(target, /datum/mind))
+		var/datum/mind/M = target
+		return M.assigned_role
+
+	//Alive mob, get it from the mind if possible
+	if(isliving(target))
+		var/mob/living/L = target
+		return L.mind?.assigned_role
+
+	//I swear to god, if you passing me a string, im going to assume you passed me the job directly and you want it compared
+	if(istext(target))
+		return target
+
+	//No match
+	return null
+
+#define IS_JOB(target, job) (find_job(target) == job)
+#define IS_COMMAND(target) (find_job(target) in GLOB.command_positions)
+#define IS_ENGINEERING(target) (find_job(target) in GLOB.engineering_positions)
+#define IS_MEDICAL(target) (find_job(target) in GLOB.medical_positions)
+#define IS_SCIENCE(target) (find_job(target) in GLOB.science_positions)
+#define IS_CARGO(target) (find_job(target) in GLOB.supply_positions)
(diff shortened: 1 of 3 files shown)


boodaliboo:
give buffs to people with skills is better then giving malices to people without skills

Hopekz:
Your PR doesn't even do the same thing as mine

In my PR you just have to do this in the actual usage:
image


To accomplish what you try to do here:
image

The delay and fail message as well as the struggle messages are already built into the framework so your way is actually not even doing the same thing just adding checks.

alexkar598:
I also call that snowflake code that doesnt translate well to other things. Its better to let the coder decide how he wants to implement shit than to force something on him. This does that and gives an option on doing it slightly easier.

Tipy1802:
This does not solve the problem of people powergaming and doing other’s job...

Bibby0110:
meh

alexkar598:
> This does not solve the problem of people powergaming and doing other’s job...

Neither does a 3 second delay on the health analyzer. Wont get in details again but in short, it's just annoying without being impactful enough to deter. So it's just terrible gameplay with no benefit.

Reed0506:
what if engineers can understand what wires actually do without having to guess or ask the CE like on baycode

Reed0506:
another idea is making assembling cyborgs and mechs by non-roboticists have a progress bar or something

Tipy1802:
> > This does not solve the problem of people powergaming and doing other’s job...
> 
> Neither does a 3 second delay on the health analyzer. Wont get in details again but in short, it's just annoying without being impactful enough to deter. So it's just terrible gameplay with no benefit.

Yes but the framework of Hopek's PR is better, don't give a shit about the example. A better example would be non MDs having a much bigger faileur rate on sugrery instead of yours where MDs just do it slightly faster, yours doesn't help with the problem of non MDs doing surgery

alexkar598:
Having non mds have a higher failure rate would be trivial (add a is ismedical check and if it fails lower the success rate). Pretty much the same complexity for developers as hopek's pr but with less code that has potential to beeak

PR #10822, opened 2021-01-07
Removes ability to clone people who are able to be defib'd for the first half of the cooldown

JamieD1:
:cl:  
tweak: You cannot clone someone if they have recently died and are defib'able.
/:cl:

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -176,6 +176,10 @@ GLOBAL_VAR_INIT(clones, 0)
 			icon_state = "pod_g"
 			update_icon()
 			return NONE
+		if(clonemind.current)
+			var/tdelta = round(world.time - clonemind.current.timeofdeath)
+			if(tdelta < (DEFIB_TIME_LIMIT * 5))
+				return NONE
 		current_insurance = insurance
 	attempting = TRUE //One at a time!!
 	countdown.start()
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -177,8 +177,9 @@ GLOBAL_VAR_INIT(clones, 0)
 			update_icon()
 			return NONE
 		if(clonemind.current)
-			var/tdelta = round(world.time - clonemind.current.timeofdeath)
-			if(tdelta < (DEFIB_TIME_LIMIT * 5))
+			var/mob/living/carbon/human/cloneperson = clonemind.current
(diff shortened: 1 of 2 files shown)


jcatjmeow (changes requested):
Please fill out the pr template.

alexkar598:
1) This is super easy to bypass by simply using a medical toolbox against the patient until they become valid for cloning

2) Assuming a law of physic/an anal admin miraculously prevents players from applying toolbox to the head, this will hurt medbay during rushes where theres several dead people and not enough doctors with no way for medbay to cope with it as you cant just pull additional doctors to defib people from your ass.

3) What issue does it solve exactly? It doesn't solve the low skill requirement medical has (which you refused to acknowledge as a problem unfortunately), it doesnt solve death being stupidly easy to revert as it only impacts people you can easily defib which are already easy to revive anyways.

4) Theres nowhere to go from this PR, you just locked in that cryo is an alternative sidegrade you can use to fix anyone that cant be defibbed immediately.

5) PR is missing justification and doesnt describe what issue its trying to solve

athena413:
the alternative is better

the existence of cloning roundstart means that the medical skill floor is substantially lowered, meaning that when an MD runs into a problem that can't be solved by cloning, the medbay staff begins fucking IMPLODING and it leads to people who aren't even dead being murdered by medbay staff so that they can be cloned, the cloning removal would mean that players need to adapt to not having it roundstart, likely leading to players getting better at medbay which is a good thing

cloning means that people can get back into the round significantly faster and with significantly less effort than without cloning, and while yeah it does get people back into the game it also means that death is much less impactful for the player that dies. increasing the time and effort to revive (by removing cloning) means that death is more impactful and can lead to players who value their life more

an argument that you and i think some others have made is "what if there's a lot of dead people" and that's called a challenge, hopefully with people valuing their life more this should happen less often, but it's a challenge for medbay when it does happen and can create interesting gameplay and interactions from being in situations like that

another argument is "well what if someone comes in and they're husked/sucked by a ling/dead for a long time", without cloning they can't be fixed immediately by throwing them into the pod which means that whoever is attending to them now has a challenge to deal with, and while yes, they'll stay dead for longer while this is being dealt with, it'll increase the impact of death more, but also they can still be revived without cloning. even doing a brain transplant out of a royally fucked body into a braindead new body is possible with some help from genetics

Takahiru:
Instead of new doctors trying to learn, they will simply leave the corpse at cloning after realizing that they are 'unclonable'.

PR #10839, opened 2021-01-08
Enabling jukeboxes sounds as an admin no longer deadmins you at the start of every round #blameredmoogle

alexkar598:
Does what it says on the tin can

:cl:  alexkar598
bugfix: Enabling jukeboxes sounds as an admin no longer deadmins you at the start of every round 
tweak: Side effect of this is that autodeadmin preferences have been reset to disabled
/:cl:


jcatjmeow (approved):

jcatjmeow:
Merge once it actually compiles (5head)

PR #10843, opened 2021-01-09
compactifies surgery availability and makes replaced_by actually work

SomeguyManperson:
# General Documentation
yep theos was bad at coding back in like 2019
still is

:cl:  
bugfix: surgeries not meant to show up will no longer show up
/:cl:


redmoogle (approved):
Tested, works fine.

PR #10844, opened 2021-01-09
tend wounds burn experimental now actually works

SomeguyManperson:
oimproivoes rp
:cl:  
bugfix: tend wounds burn experimental will now show up 
/:cl:


redmoogle (approved):
based, no reason why this shouldnt work

PR #8578, opened 2020-05-19
space cleaner now deals toxin damage when ingested

boodaliboo:
what kind of idiot is eating soap
#### Changelog

:cl:  
tweak: space cleaner now deals toxin damage
/:cl:

PR #10845, opened 2021-01-09
sterilizer spray no longer causes toxin damage

SomeguyManperson:
dumb as hell

### What general grouping does this PR fall under? 
med

### If there are any numerical values involved in your PR that will be relevant to a player, please note them here. 
probably not documented anywhere but space cleaner deals like 1 toxin damage and sterilizer used to but now does not do that

# Changelog

:cl:  
tweak: sterilizer spray will no longer cause toxin damage
/:cl:

PR #10846, opened 2021-01-09
Sterilizine is no longer a variant of space cleaner and as such no longer deals toxin damage.

alexkar598:
This stops it from cleaning things I guess but that doesn't really change anything to surgery functionally anyways. Was a bug introduced in #8578

:cl:  
bugfix: Sterilizine is no longer deals toxin damage.
tweak: Sterilizine is no longer a variant of space cleaner
/:cl:


alexkar598:
Closed in favour of #10845

PR #10850, opened 2021-01-10
replaces tintin on the moon starting song with a better remix - Jamie added both instead

boodaliboo:
https://www.youtube.com/watch?v=BY-1SrsLER0
new song 
:cl:  
tweak: better song then older
/:cl:


Mqiib:
Old song has soul, don't remove it >:(((((
That being said, the song you're adding sounds good.

PR #10856, opened 2021-01-11
makes synthfesh heal burn husks when they are under 50 damage and there is over 100 units of synthflesh present

SomeguyManperson:
# General Documentation

make synthflesh more viable by giving it more utility
also needing a high tier surgery to just unhusk someone is a pain in the ass

make medbay less pain and less reliant on clooning
# Wiki Documentation

synthflesh can now heal burn husks specifically if the total volume reaches 100 while  burn damage is under 50

synthflesh can now unhusk people if the husking is solely from burns
medbay

Synthflesh will unhusk burn-related husk corpses when:
the corpse has less than 50 burn damage
synthflesh is dosed to at least 100 units
unhusking will ONLY occur on synthflesh being applied so make sure to do that part first

# Changelog

:cl:  
tweak: synthflesh will now heal burn husks if their burn damage is under 50 and the total amount of synthflesh dosed reaches 100
/:cl:

### DIFF
--- a/code/__DEFINES/reagents.dm
+++ b/code/__DEFINES/reagents.dm
@@ -29,3 +29,7 @@
 #define REM_REAGENT		3	// reagent removed (may still exist)
 
 #define MIMEDRINK_SILENCE_DURATION 30  //ends up being 60 seconds given 1 tick every 2 seconds
+///Health threshold for synthflesh and rezadone to unhusk someone
+#define UNHUSK_DAMAGE_THRESHOLD 50
+///Amount of synthflesh required to unhusk someone
+#define SYNTHFLESH_UNHUSK_AMOUNT 100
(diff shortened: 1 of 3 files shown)


Xoxeyos:
The Heretic Flesh path may be unbalanced by this, one of their abilities is to make a slave by killing them, and then using the transmutation rune with them over it, husking them. I don't believe heretics should be able to run around with two 100~ health slaves. [
image
](url)

SomeguyManperson:
this should only reverse husking from burns, not from other sources but I'll have to double check that

Xoxeyos:
> 
> 
> this should only reverse husking from burns, not from other sources but I'll have to double check that

Please do tell me when you have tested this on a heretic ghoul, and if you fix any problems with it, I'll change my vote.

SomeguyManperson:
heretic ghouling uses generic husking which isn't cured by specific burn dehusking

JamieD1 (approved):

PR #10868, opened 2021-01-12
allows you to use boxing gloves to pick up lightbulbs

boodaliboo:
there huge leather gloves you should be able to touch a lightbulb with thm,e
:cl:  
rscadd: Added temp resist to boxing glover
/:cl:

### DIFF
--- a/code/modules/clothing/gloves/boxing.dm
+++ b/code/modules/clothing/gloves/boxing.dm
@@ -4,7 +4,7 @@
 	icon_state = "boxing"
 	item_state = "boxing"
 	equip_delay_other = 60
-	max_heat_protection_temperature = GLOVES_MAX_TEMP_PROTECT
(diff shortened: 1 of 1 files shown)


code/modules/clothing/gloves/boxing.dm
```diff
@@ -4,6 +4,9 @@
 	icon_state = "boxing"
 	item_state = "boxing"
 	equip_delay_other = 60
+	max_heat_protection_temperature = GLOVES_MAX_TEMP_PROTECT
+	heat_protection = HANDS
+	max_heat_protection_temperature = GLOVES_MAX_TEMP_PROTECT
```
RookieMaster, line 9: Why are you setting max_heat_protection_temperature 2 times?

alexkar598:
powercreep, no real reason behind this change

Marmio64:
I'd maybe consider supporting this powercreep if the grammar was better.

PR #10871, opened 2021-01-13
Switch turdis apt repos to the main server to prevent errors when mirrors are mirroring

alexkar598:
a-a-aawooooo

should fix the error @redmoogle sent me


boodaliboo:
powercreep unneeded

JamieD1 (approved):

PR #10872, opened 2021-01-13
report healthy organs on health scanner

ynot01:
# General Documentation

tested with zombie organ and sling organ on felinids

healthy organs will be reported on health scanners
antag organs will not show up unless damaged

doctors can now tell if someone is missing an organ

# Wiki Documentation

healthy organs will be reported on health scanners

doctors can now tell if someone is missing an organ

Medical

# Changelog

:cl:  
rscadd: Added healthy organs to health scanner report
/:cl:

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -292,7 +292,7 @@ GENE SCANNER
 				else
 					minor_damage = "\t<span class='info'>Mildly Damaged Organs: "
 					minor_damage += organ.name
-			else if (organ.damage < organ.low_threshold)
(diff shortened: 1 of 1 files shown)


Tipy1802:
> doctors can now tell if someone is missing an organ

They already get told X's y is missing
Seems like it would lead to information overload

ynot01:
> > doctors can now tell if someone is missing an organ
> 
> They already get told X's y is missing
> Seems like it would lead to information overload

no they don't

Tipy1802:
Ah my bad, it does that only for brain and heart when deffibing apperantly.
Yeah fair enough then, I would rather it tell you the missing organs instead of the present ones as it would make the health analyser into an even bigger wall of text but this is fine I guess

JamieD1 (approved):

PR #10887, opened 2021-01-15
Crafted secure closets are now functional - fixes #10789

### LINKED ISSUE #10789: Crafted secure lockers NOT SECURE
Craft secure locker and then **it can be locked and opened without any access or ID card.**

jcatjmeow:
fixes #10789
Tested!

## WIKI DOCUMENTATION
Crafted secure lockers now decide their access via the airlock electronics. Set your airlock electronics first, then craft the locker, and the locker will have the exact same access permissions. Yay!

# Changelog

:cl:  
bugfix: crafted secure lockers are now secure
/:cl:


redmoogle (approved):
Alright after debating on it, while it doesnt address the original bug. It does a good job at atleast putting in a better way

JamieD1 (approved):

PR #10889, opened 2021-01-16
Common Build Tooling [PORT]

Bibby0110:
# General Documentation

Ports https://github.com/tgstation/tgstation/pull/55373 by stylemistake :3

Both dm and tgui are now built with a single script. It's pretty easy to launch: just press Ctrl+Shift+B in VSCode, and tada! 🎉

It's smart. It will skip certain steps if source files were untouched for that step. So, if you're only touching dm code, it will only rebuild dm code, and will skip tgui.

Syntax is fairly readable and maintainable.
```javascript
const { Task, runTasks, exec } = require('./cbt');

const taskTgui = new Task('tgui')
  .depends('tgui/yarn.lock')
  .depends('tgui/packages/**/*.js')
  .provides('tgui/public/*.bundle.*')
  .provides('tgui/public/*.chunk.*')
  .build(async () => {
    await exec('tgui/bin/tgui');
  });

runTasks([taskTgui]);
```

This is a long term solution to the js bundle hell. Now that we have a single script to build everything, bundles have been excluded from the repo, and they will no longer cause conflicts in PRs.

CI pipelines have been updated. They're not coded in the most optimal way, just making them green for now.

**TODOs for later: (if you want to see any of this in scope of this PR, ping me)**

- [ ] Support compiling with DM defines by passing them as an argument, like -D LOWMEMORYMODE.

- [ ] Instead of explicitly listing the task sequence in runTasks(), support specifying tasks in .depends(), which in turn will allow building a graph of dependencies and running things in parallel.

### Information for Developers
On Windows, the build scripts will automatically install Node. Other platforms should use their package manager or download it from https://nodejs.org/en/download/

Pick one:

VSCode: Ctrl+Shift+B
VSCode: F5 (build & run with debugger) or Ctrl+F5 (build & run without debugger)
Windows: double-click root Build.bat (pause to see output) or tools/build/build.bat (no pause)
Git Bash and non-Windows: tools/build/build

### Information for Server Admins
TGS scripts will automatically install the version of Node specified in dependencies.sh
Either use this build script, or compile tgui by running any script in tgui/bin folder.

Quicker PR turnaround time, less time wasted on rebuilding tgui.


maxion12345:
harm intent groin

JamieD1 (approved):

PR #10928, opened 2021-01-25
Makes polonium more radioactive

Mqiib:
Instead of only barely doing something, it now will make you start taking toxin damage from radiation after approximately not a very long time because I didn't do the math but it's NOT USELESS NOW. It should now add 40 rads per cycle instead of 4, meaning it might possibly get used!

# Changelog

:cl:  

tweak:Polonium now does TEN TIMES as much badness per cycle

/:cl:


Marmio64:
you should probably test it and see how fast it gets you up to the damage threshold of radiation

Mqiib:
It does exactly what it says on the tin, +40% radiation level per cycle. For reference, this will get to 500 rads in about 12 cycles or 24ish seconds. After this point, it will begin to deal <1 toxin damage per cycle due to the radiation. It's still quite slow compared to other chems but now isn't actually so hilariously useless there is no point in using it. 5 units will by itself get you close to 750% radiation, and 10 will get to around 3400% (rough estimate based on # of cycles and natural rad reduction.) 20 units took close to 5 minutes to kill the target and left them at 10,388% radiation. Pre-buff, the same 20 units would AT BEST reach 1000% which is extremely easy to deal with and only does relatively small amounts of damage.

JamieD1 (approved):

PR #10945, opened 2021-01-30
Fixes autolathe "tinfoil hat" bug by renaming Misc section

ChesterTheCheesy:
# General Documentation

Fixes bug properly without nuking the poor hat. It's innocent.

bugs bad, tinfoil hat god

renames "Misc" category to "Miscellaneous", as the "Misc" really screws up when there's other categories named similarly (Misc. Machinery), making the latter show up indefinitely, but with Tinfoil hat, as it's the last saved value in that case. Also moves the mounted flash frame design to correct file.

TINFOIL HAT [x5]
TINFOIL HAT [x5]
TINFOIL HAT [x5]
isn't a case anymore
bugfix

# Changelog

:cl:  
bugfix: fixed autolathe bug, moved one protolathe only design to correct file
/:cl:


ChesterTheCheesy:
However, i could be wrong on the source of the bug, as i don't know javascript that well.

jcatjmeow (approved):

PR #10949, opened 2021-01-31
Changes how the push for actually using anesthetics works so it's more intuitive

SomeguyManperson:
# General Documentation

Surgeries no longer have the stupid innate 0.5 subtracted from their success multiplier without proper preparation
instead there is an innate 50% chance for surgeries without anesthetics/painkillers to have a slip-up regardless of success
this chance is altered multiplicatively by:
failing the surgery (*2)
the target having the chems "painkiller" and "miner's salve" in their reagents (*0.5 for each)
the target being unconscious through any means (*0.8)
the target's drunkenness (treated as a % value that is removed i.e. 50 drunkenness is *0.5, 100 drunkenness is *0)
a slip-up will not fail the surgery step, but will deal extra damage and can cause bleeding with the amount depending on the surgery step. Healing steps and the plastic surgery step can't cause slipups
The numbness symptom analgesia and tenacity brain trauma also count as painkillers
tldr failing surgeries actually matters now

current system isn't functioning as fully intended

# Wiki Documentation

This will be the basis of the Wiki entry for your PR, and more information / detail is better for Wiki editors to integrate.
surgeries without the patient being under anesthetics, morphine, or the effects of a stasis bed will have a 50% chance to deal damage on each step, this can also cause bleeding if the step that fails isn't the cautery one
this chance is doubled if the step fails so don't do that
ghetto surgery is now riskier than normal surgery since you can die from doing it wrong, also use of an anesthetic tank, morphine, or a stasis bed for normal surgeries is preferred over the lack of one
Cyborgs, as with normal surgery failure chance, don't have to deal with this because they are special snowflakes who can't use hands
steps that heal i.e. tend wounds, patch/treat and removal of embedded objects as well as the reshape face step of plastic surgery will not trigger this effect
surgery

no

# Changelog

:cl:  
tweak: surgeries will no longer have a high fail chance if the target is not under anesthetics or painkillers, instead they'll have a high chance to cause a small to moderate amount of damage. This chance can be reduced through high alcohol intake and certain chemicals and outright removed with the application of morphine or an anesthetic gas
tweak: stasis bed success modifier returned to 1 from 0.8
/:cl:

### DIFF
--- a/code/datums/brain_damage/special.dm
+++ b/code/datums/brain_damage/special.dm
@@ -155,11 +155,13 @@
 /datum/brain_trauma/special/tenacity/on_gain()
 	ADD_TRAIT(owner, TRAIT_NOSOFTCRIT, TRAUMA_TRAIT)
 	ADD_TRAIT(owner, TRAIT_NOHARDCRIT, TRAUMA_TRAIT)
+	ADD_TRAIT(owner, TRAIT_SURGERY_PREPARED, TRAUMA_TRAIT)
 	..()
 
 /datum/brain_trauma/special/tenacity/on_lose()
 	REMOVE_TRAIT(owner, TRAIT_NOSOFTCRIT, TRAUMA_TRAIT)
 	REMOVE_TRAIT(owner, TRAIT_NOHARDCRIT, TRAUMA_TRAIT)
+	REMOVE_TRAIT(owner, TRAIT_SURGERY_PREPARED, TRAUMA_TRAIT)
 	..()
 
 /datum/brain_trauma/special/death_whispers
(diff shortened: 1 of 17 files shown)


SomeguyManperson:
oh also I changed "get_propability_modifier" to "get_probability_modifier" and any mention of propability to probability since it's terrible to read

Tipy1802:
I feel like you shouldn't revert the stasis bed decrease as it's stupid how they are just plain better than everything

N3D6:
This seems a lot better than the failrate, I also saw a lot of people disagreeing with it yesterday so this is a step in the right direction

maxion12345:
Do they scream when you fuck up for immershun

jcatjmeow (approved):
Sure. Maybe make it more obvious that the mess up is caused by a lack of anesthetics? But this is fine

maxion12345:
Wait shit, what about medborgs? How do they put people to sleep. I don't think they get a sleeping chem in their synth, or a way to apply nitrous. Probably something for another pr.

alao3alao3:
tell wiki when it’s merged not when it’s open, what’re we supposed to do with an open pr

PR #10984, opened 2021-02-11
probably actually fixes the get_weather proc

SomeguyManperson:
:cl:  
bugfix: get_weather now actually really works as intended
/:cl:


ChesterTheCheesy:
Will this actually finally fix barometers?

JamieD1 (approved):

PR #10989, opened 2021-02-12
probably fixes firedoors being jank

SomeguyManperson:
1 = "this is safe"
-1 = "this is not safe"
0 = "literally doesn't matter"
1 isn't 0
:cl:  
bugfix: firelocks should cycle as intended
/:cl:


redmoogle (approved):
Did I miss something I thought it only gave 2 values my bad

PR #11008, opened 2021-02-16
removes goliath stun in favor of more damage

ynot01:
# General Documentation

### Intent of your Pull Request

no more goliath stun
significant damage and knockdown instead

### Why is this change good for the game?

stun is bad game design

:cl:  
tweak: Removed goliath stun in favor of more damage
/:cl:

PR #11009, opened 2021-02-16
goliath tentacles chainstunning will only increase the current stun by 2 seocnds instead of resetting it to 10

SomeguyManperson:
# General Documentation

Regardless of the contents of a change to this it is clear that goliath chainstuns are unfun to get hit by and anyone who disagrees is wrong and bad

attempts to fix the same problem #11008 does but in a different way
# Wiki Documentation

Goliath tentacles will only stun for 2 seconds if the person getting hit is already stunned, if they are not stunned, the tentacles will stun for 10 seconds
mining

# Changelog

:cl:  
tweak: goliath tentacles will only increase existing stuns by 2 seconds instead of setting them to 10 to reduce chainstunning potential
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/goliath.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/goliath.dm
@@ -201,7 +201,7 @@
 		if((!QDELETED(spawner) && spawner.faction_check_mob(L)) || L.stat == DEAD)
 			continue
 		visible_message("<span class='danger'>[src] grabs hold of [L]!</span>")
-		if(!IsStun(L))
+		if(!L.IsStun())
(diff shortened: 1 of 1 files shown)


ynot01:
an interesting alt, ill let the maints decide which is better, though i still dislike stun mechanics

ChesterTheCheesy:
This is much better

N3D6:
Yeah this is pretty good too, goliaths chainstunning is pretty shitty, it happening for a lower amount of time seems a lot better.

ToasterBiome:
I approve of this, more forgiving but still keeps what i think is a critical goliath ability

LazennG:
something something miners op so pr bad

PR #10931, opened 2021-01-26
uniform sprites for digitigrade lizardpeople

ThatGuyPhil:
### Intent of your Pull Request

add jumpsuit sprites for digitigrade legs so you can tell the difference between digitigrade and shoeless plantigrade when wearing any uniform that isnt a jumpskirt or shorts. Since this is my third ever PR on a thing thats more than a single sound or a youtube link I really dont know if theres more than just uploading jumpsuit sprites. There are somethings that probably should be removed like the uniforms that we dont have but i feel like im playing jenga when it comes to coding so i would like some help.

also the captain jumpsuit is still in its old colors for this but its not like lizardpeople can be captain on yogs anyway.

:cl:  
rscadd: Addeds uniform sprites with digitigrade legs
wip: work in progress   
imageadd: adds sprites 
/:cl:

PR #11017, opened 2021-02-17
Revert "uniform sprites for digitigrade lizardpeople"

nesisknesis:
Reverts yogstation13/Yogstation#10931

Randomly adds an entire folder of edited clothing.

Clothing that has been updated is outdated in this folder, there are many clothing sprites that we do not have, and missing clothing sprites that we do have.

Was merged with no code at all so it doesn't work at all. 

🆑
imagedel: A whole folder of un-usable sprites.
/🆑


ThatGuyPhil:
I added those so that I could add the code for them later if i ever learned how to but IDK maybe I or someone could readd them another time and do it right.

Hopekz:
Tbh this is why we need some solution for unused assets Like a separate repo or something

PR #11026, opened 2021-02-19
bees no longer have armor piercing stingers

SomeguyManperson:
minor oversight allowed homeless bees to attack people regardless of bee protection also minor update to the homeless bee examine text to explain that it is mad
:cl:  
bugfix: homeless bees will no longer sting people wearing bee suits
/:cl:


redmoogle (approved):
Code is fine

PR #11030, opened 2021-02-20
announces round-end when nar-sie summoned

ynot01:
# General Documentation

Round is announced to be ended when nar-sie is summoned, and then after the cinematic plays it actually ends
Can't just move the round-end for reasons that alexkar knows and not me

Players are free to act as if the round has already ended once nar-sie is summoned

# Changelog

:cl:  
rscadd: Added a pseudo round-end when nar-sie is summoned
/:cl:

### DIFF
--- a/code/modules/power/singularity/narsie.dm
+++ b/code/modules/power/singularity/narsie.dm
@@ -75,6 +75,7 @@
 	INVOKE_ASYNC(src, .proc/begin_the_end)
 
 /obj/singularity/narsie/large/cult/proc/begin_the_end()
+	to_chat(world, "<BR><BR><BR><span class='big bold'>The round has ended.</span>")
 	sleep(50)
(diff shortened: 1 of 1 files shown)


Tipy1802:
All of this just because someone made an announcement saying “Narsie is Thicc”

code/modules/power/singularity/narsie.dm
```diff
@@ -75,6 +75,7 @@
 	INVOKE_ASYNC(src, .proc/begin_the_end)
 
 /obj/singularity/narsie/large/cult/proc/begin_the_end()
+	to_chat(world, "<BR><BR><BR><span class='big bold'>The round has ended.</span>")
```
redmoogle, line 78: This is not how round ends work. If you look lower you'll see cult_ending_helper which causes the round to actually end

ynot01: see line 4 of PR

> Can't just move the round-end for reasons that alexkar knows and not me

PR #11033, opened 2021-02-20
probably makes legion not shit itself whenever ahealed

SomeguyManperson:
I think
:cl:  
bugfix: legion probably won't scream and open its door when ahealed
/:cl:


alexkar598:
but.. why would you aheal a full health legion thats still in his pen or whatever...

SomeguyManperson:
Admin revive all button

redmoogle (approved):
I don't see why this wouldn't work seeing at the code will now only trigger if they get healed more than 0 health.

PR #11042, opened 2021-02-22
organ smartfridge heals as fast as intended

SomeguyManperson:
# General Documentation

heaing_factor and decay_factor are based off the organ's health meaning on their own the amount of damage changed is microscopic

# Changelog

:cl:  
bugfix: organ fridge will now heal as fast as it should
/:cl:


redmoogle (approved):
I was about to do this but ok.

PR #11051, opened 2021-02-23
actually fixes reinforced floors and walls interacting with singularity

SomeguyManperson:
issue was in normal floor singularity_pull which would attempt to replace the tile with a plating and fail while also spawning in the tile used to make the floor which is the rods = infinite rods
tldr you can now easily(tm)(r) contain a tier 4 singularity

:cl:  
bugfix: singularities will now not spawn infinite rods when near reinforced floors
bugfix: singularities will no longer eat rwalls like normal walls
/:cl:


Jarod1200:
the end of an era

Hopekz:
based

PR #11052, opened 2021-02-24
accelerated particles can no longer get infinite range

SomeguyManperson:
lol
:cl:  
bugfix: accelerated particles can no longer attain infinite range
/:cl:


jcatjmeow:
is there a way to remove the sleep? If not then ok

PR #11060, opened 2021-02-25
damages venus human traps if they are not close to kudzu

ynot01:
# General Documentation

damages venus human traps if they are not close to kudzu

venus human traps should not be fucking off away from the kudzu as their own little antag adventure, they should be protecting it and using it as their lifeblood

# Changelog

:cl:  
tweak: Venus human traps now take damage if they stray away from the kudzu
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
+++ b/code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
@@ -102,7 +102,7 @@
 	. = ..()
 	pull_vines()
 	if(!kudzu_need())
-		to_chat(owner, "<span class='danger'>You wither away without the support of the kudzu...</span>")
+		to_chat(src, "<span class='danger'>You wither away without the support of the kudzu...</span>")
 		adjustHealth(-5)
 	
 /mob/living/simple_animal/hostile/venus_human_trap/AttackingTarget()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
```diff
@@ -101,6 +101,9 @@
 /mob/living/simple_animal/hostile/venus_human_trap/Life()
 	. = ..()
 	pull_vines()
+	if(!kudzu_need())
+		to_chat(src, "<span class='danger'>You wither away without the support of the kudzu...</span>")
```
ChesterTheCheesy: Can you add a if(prob(10)) before to_chat so it doesn't spam it indefinitely?

Marmio64:
at least make the kudzu heal them if they stand in it, reason it works for blobbernauts is because they can heal at the blob

Sniblet:
Mantraps only heal by attacking people as is and restricting their ability to venture out and murderbone as far off as they want is actually sabotaging their ability to live yes

ynot01:
changed probability to 20 because 10 ticks will kill you and statistically only one of those ticks would indicate why you are dying so now two of those ticks will indicate why you are dying

MegaEmpirical:
I like the change but agree that healing while on vines would likely be a good addition too.

JamieD1:
I would add small healing if they are on the kudzu.

redmoogle (approved):
Code is good

SomeguyManperson (approved):

SomeguyManperson:
heal from eating people

PR #11062, opened 2021-02-25
Fixes autolathe incorrectly identifying the processing item as the last requested item

### LINKED ISSUE #11061: Autolathe - inconsistent printing!
- if you construct an autolathe and **_initially_** want to print 30 cables, often it prints 30 x 30unit cable coils. If I print something beforehand and wait it out, and then print cable coils, it prints only 1 x 30u cable coils. Happened **many** times, and now becoming annoying.

redmoogle:
# General Documentation

Apparently that was a bug _lmao_

So what happens is when it goes to the make_item proc it uses the last requested item instead of the design it's making it use the last queued item (eg. 1 cable coil, 1 toolbox) which then causes it to freak the fuck out and make _way_ more than it should

Fixes: https://github.com/yogstation13/Yogstation/issues/11061

Fixes a bug in which the cable-coil monster would emerge

# Changelog

:cl:  
bugfix: Cable coils no longer print in sets of 30 if something is queued after it
/:cl:

### DIFF
--- a/code/game/machinery/autolathe.dm
+++ b/code/game/machinery/autolathe.dm
@@ -336,8 +336,6 @@
 
 /obj/machinery/autolathe/proc/make_item(datum/design/D, multiplier)
 	var/is_stack = ispath(D.build_path, /obj/item/stack)
-	if(is_stack)
-	else
 	var/coeff = (is_stack ? 1 : prod_coeff) //stacks are unaffected by production coefficient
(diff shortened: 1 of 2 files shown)


redmoogle:
Hmm cannot replicate, will try to _fix_ if I can get it to reliably break

SomeguyManperson (approved):

PR #11064, opened 2021-02-27
fixes donor borgs not having proper overlays

ynot01:
# General Documentation

so basically prior, the code would run the original update_icons()

if the user had a donor skin, it would cut all overlays off, reapply the donor skin, and add back only the eye lights

this would kill every other overlay that isn't eye lights

there might be a better way to do this but I'm assuming the person that coded it like this had good-ish reasons to do so, so i just reattached the missing overlays to the donor borg check

# Changelog

:cl:  
bugfix: Fixed donator borg skins not showing hats and covers and fire
/:cl:


ChesterTheCheesy:
pog

SomeguyManperson (approved):

PR #11077, opened 2021-03-01
Halves Maintenance Drone Health

alao3alao3:
# General Documentation

# Intent of your Pull Request

Halves the health of Maintenance Drones.
I've received on good enough authority this won't affect any other drones
dpng

# Why is this change good for the game?
https://discord.com/channels/134720091576205312/134720091576205312/815788278971105320
bpng
I and a few other people believe bots have too much health given what they do for the station (mitigate any and all damages done to the station short of heat and rads maybe, can fix spaced or toxic areas with 0 risk to themselves, essentially infinite if a drone is smart enough).

# Wiki Documentation

The health of Maintenance Drones is now 15, as opposed to the initial 30. The health of drones aren't listed anywhere so nothing on the wiki has to be changed.

########################################### Changelog

:cl:  
tweak: Halved drone health.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/drone/_drone.dm
+++ b/code/modules/mob/living/simple_animal/friendly/drone/_drone.dm
@@ -23,8 +23,8 @@
 	icon_living = "drone_maint_grey"
 	icon_dead = "drone_maint_dead"
 	possible_a_intents = list(INTENT_HELP, INTENT_HARM)
-	health = 30
-	maxHealth = 30
+	health = 15
+	maxHealth = 15
 	unsuitable_atmos_damage = 0
 	wander = 0
 	speed = -1 //YOGS - drones //Buffed to not be a fucking god damn piece of slug matter writhing on the fucking floor


Sniblet:
Drones don't need the health, except when monstermos decides that they do and throws a stack of metal at them for 20 damage long after the area was supposed to depressurize
Getting instakilled is not fun and I don't see why it should happen more

Mqiib:
30 health is plenty low enough imo, 15 means that they'll die to a single lit welder monstermos'd into their face. Not fun.

alao3alao3:
Drone health certainly comes into play when you're going after the damn thing. These things are meant to be expendable and fragile, and it's the only downside drones have besides not being able to interact with living beings. I'd be willing to go to 10 given what I've seen them do, instadeath ought to be par for the course.

ynot01:
There's not much reason to _go after_ drones. If it's interfering with your plans intentionally then you should ahelp because it's not supposed to interact with beings.

alao3alao3:
can't ahelp a drone for fixing bombed out areas and false walls, nor for reinforcing random doors and depressurizing a room filled with plasma. It's their job after all and they can do it with 0 obstacles.

maxion12345:
The health is perfectly fine as it is. Lowering it would mean monstermos killing a drone with flung objects would happen far more then it already does. And because drones primarily fix things like spaced rooms, you can imagine you get flung about by rooms refilling with air, rooms being emptied of bad air, etc.

And...do they really need the nerf?

ChesterTheCheesy:
15 damage is done by people by accident by hitting the drone with a lit welder

Marmio64:
if you accidently kill a drone you can reactivate it with a screwdriver

PR #11081, opened 2021-03-02
makes simplemobs actually do emotes

SomeguyManperson:
me emotes have a special prevention when not intentional for some probably good reason so I've decided to make automated speech bypass that allowing simplemobs to do audible and visible emotes again

:cl:  
bugfix: simplemobs will now do audible and visible emotes again
bugfix: pets will now do emotes on being pet and other stuff
/:cl:


redmoogle:
> 
> 
> me emotes have a special prevention when not intentional for some probably good reason so I've decided to make automated speech bypass that allowing simplemobs to do audible and visible emotes again

Yes Jamie did it to prevent an runtime

JamieD1 (approved):

PR #11083, opened 2021-03-02
Fixes all mobs being able to snap their fingers

UselessTheremin:
was funny but didn't make sense

fixes this 
image

oversight bad, improves rp i guess, also removes fun

:cl:  
bugfix: only people can snap their fingers now
/:cl:


Jarod1200:
I mean a robotic ai should be able to make a snap noise with whatever it has that generates audio.

UselessTheremin:
> I mean a robotic ai should be able to make a snap noise with whatever it has that generates audio.

but all mobs can currently do it, including slimes, xenomorphs, people in cuffs, morphs, space dragons, blobbernauts, carp, you get the idea.

PR #11091, opened 2021-03-04
Only shows the "You may call the shuttle whenever you find it appropriate." part of the end of shift announcement if the shuttle is idle(not called)

alexkar598:
It always bothered me, this improves RP because CC are not idiots, they are going to know that the shuttle is called and theyre not going to tell you to feel free to call it if its already called


Bibby0110:
we both know CentComm has the collective intelligence of the mentor team :))

Cartlord:
This did literally the opposite of what it was supposed to do, now it only announces we can call it if we already called it

PR #11092, opened 2021-03-05
Changes the threshold from which jobs get their access expanded due to lowpop from 8 to 25

alexkar598:
8 is ridiculously low, it might as well be disabled at that point.

For those who don't know, the additional access thingy gives jobs more access(usually related to their departments) on lowpop. For example, the doctors gain access to chemistry.

### DIFF
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -327,7 +327,7 @@ GATEWAY_DELAY 18000
 ## If the number of players ready at round starts exceeds this threshold, JOBS_HAVE_MINIMAL_ACCESS will automatically be enabled. Otherwise, it will be disabled.
 ## This is useful for accomodating both low and high population rounds on the same server.
 ## Comment this out or set to 0 to disable this automatic toggle.
-MINIMAL_ACCESS_THRESHOLD 25
+MINIMAL_ACCESS_THRESHOLD 18
 
 ## Comment this out this if you wish to use the setup where jobs have more access.
 ## This is intended for servers with low populations - where there are not enough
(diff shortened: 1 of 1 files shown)


Jarod1200:
kinda nitpicky but i think 25 is too high and 15 would be better

UselessTheremin:
Tbh 20 sounds a bit better since with 25 it isn’t quite a skeleton crew

maxion12345:
Should make it contextual. Round start paramedics get surgery access if there are no doctors, doctors get chemistry if there is no chemist, etc. Do the same for others, as well. Engineers can get atmos if there are no atmos techs roundstart, scientists can get robotics without roundstart robotics.

Xoxeyos:
I think 15 is a more preferable number.

patpol4:
25 is a bit too high imo, 15 seems like a better idea

Tipy1802:
15 is still too low.
Most of the times when the pop is 20-25 there still are whole departments without workers or 1 or 2 at most.
Also keep in mind a big part of the pop does not even play the game but ghosts

patpol4:
Right, forgot about the ghosts

Bibby0110:
> 15 is still too low.
> Most of the times when the pop is 20-25 there still are whole departments without workers or 1 or 2 at most.
> Also keep in mind a big part of the pop does not even play the game but ghosts



> Right, forgot about the ghosts

97% sure it's based on people that are ready at roundstart, and not the actual amount of people online

JamieD1:
Change it to 18.

config/game_options.txt
```diff
@@ -327,7 +327,7 @@ GATEWAY_DELAY 18000
 ## If the number of players ready at round starts exceeds this threshold, JOBS_HAVE_MINIMAL_ACCESS will automatically be enabled. Otherwise, it will be disabled.
 ## This is useful for accomodating both low and high population rounds on the same server.
 ## Comment this out or set to 0 to disable this automatic toggle.
-MINIMAL_ACCESS_THRESHOLD 8
+MINIMAL_ACCESS_THRESHOLD 25
```
JamieD1:
```suggestion
MINIMAL_ACCESS_THRESHOLD 18
```

JamieD1 (approved):

PR #11095, opened 2021-03-05
compactifies operating computer backend into a list

SomeguyManperson:
# General Documentation

this is probably a terrible idea for several reasons but I have no reason to believe it won't work
now instead of needing to add another check for a specific thing in the surgery's can_start proc you just add it to the list and give it a get_computer proc that returns its connected computer easy
has been tested and works fine
it is DEFINITELY not
# Wiki Documentation

# Changelog
:cl:  
experimental: slight change to how surgeries check operating computers for if they can be started to make the code for it look less bad
/:cl:

### DIFF
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -71,11 +71,11 @@
 	var/list/computer_viable_surfaces = list(/obj/structure/table/optable, /obj/machinery/stasis) //make sure anything in this list has a get_computer() proc thanks :)
 	for(var/thing in computer_viable_surfaces)
 		var/thingy = locate(thing, T)
-		if(!thingy)
-			continue
-		opcomputer = call(thingy, "get_computer")()
+		if(thingy)
+			opcomputer = call(thingy, "get_computer")()
 		if(opcomputer)
 			break
+
 	if(!opcomputer)
 		return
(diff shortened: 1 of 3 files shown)


code/modules/surgery/surgery.dm
```diff
@@ -68,13 +68,13 @@
...
-		opcomputer = table.computer
-	else
-		var/obj/machinery/stasis/the_stasis_bed = locate(/obj/machinery/stasis, T)
-		if(the_stasis_bed?.computer)
-			opcomputer = the_stasis_bed.computer
+	var/list/computer_viable_surfaces = list(/obj/structure/table/optable, /obj/machinery/stasis) //make sure anything in this list has a get_computer() proc thanks :)
+	for(var/potential_surface in computer_viable_surfaces)
+		var/actual_surface = locate(potential_surface, T)
+		if(actual_surface)
+			opcomputer = call(actual_surface, "get_computer")()
```
alexkar598, line 75: Lack of type safety

alexkar598 (changes requested):

code/modules/surgery/surgery.dm
SomeguyManperson: Where we're going we don't need safety

PR #11116, opened 2021-03-11
adds a admin only pac man reagent and disease

boodaliboo:
adds a new pacman reagent and disease to hopefully incetivize pac man to be used for a event or something
:cl:  
rscadd: Added pac man disease and reagent
/:cl:


alexkar598:
qAh yes the classic make the pr again when it's another headcoder


This was previously denied for being bloat

PR #11119, opened 2021-03-11
Fixes Sec Budget

Bibby0110:
Fixes the sec budget so you don’t end up with hundreds of thousands 10 minutes in. Either this or removal altogether 

# Changelog

:cl:  
bugfix: Sec budget won’t ballon to ridiculous sizes
/:cl:


redmoogle (approved):
Eh fuck it good enough, better than reading `no` for every field

PR #11131, opened 2021-03-13
Hopefully fixes failing prs because ubuntu-latest

redmoogle:
ubuntu-latest is a bad idea


alexkar598 (approved):
Not a fan of not supporting ubuntu 18 but in a test environment it seems to work anyways so fine i guess??

alexkar598:
ah balls its dependent on headcoder, whatever, im the primary maintainer of the CI suite anyways

PR #11132, opened 2021-03-13
Adds assassinate subtypes that allow revival or cloning and dilutes assassinate objectives with them

SomeguyManperson:
# General Documentation

adds new objectives so you don't have to babysit or permadelete a corpse every time you get a kill objective just to not chance MAYBE POSSIBLY that someone found the body you hid, revived it without your knowledge, and gave the person a new identity or hiding place or whatever without you knowing
these allow greentext if the person has been cloned at least once, or if they have been dead at least once

allows more player agency and improves rp
# Wiki Documentation

two new objectives for solo antags: assassinate but cloning won't fail it and assassinate but any revival won't fail it. like maroon where there's more than one way to go about it.

# Changelog

:cl:  
rscadd: two new objectives that may take the place of assassinate objectives similarly to maroon, which function like assassinate objectives except they can succeed if the target has been cloned for one or has died at all for the other
/:cl:


ChesterTheCheesy:
Good change, gives more of a reason to revive bodies instead of cloning them.

maxion12345:
this feels a bit hugbox for some reason not sure why

JamieD1:
This doesnt make RP sense at all.

SomeguyManperson:
cool

SomeguyManperson:
because nobody has EVER wanted someone to be injured but not dead

Sniblet:
Imagine how awkward these objectives could be as a doctor
"Don't worry, I'll bring you back myself, I just have to flatline you for like a second"
Better yet you kill them incidentally by sawing them open during regular old life-saving surgery

SomeguyManperson:
need to do something else

PR #11140, opened 2021-03-15
Adds Ian Hitchkoff to death_commando.txt

nesisknesis:

Okay I would be completely understand if this got rejected but two neurons in my brain came together and came up with this idea, and then another came along saying "this just feels like it should be".

I think it is very fitting he became a Deathsquad Trooper.

Also fixed an alphabetical error.

# Changelog

:cl:  Identification
rscadd: Adds 'Ian Hitchkoff' to the Deathsquad Trooper name list.
/:cl:

### DIFF
--- a/strings/names/death_commando.txt
+++ b/strings/names/death_commando.txt
@@ -38,9 +38,10 @@ GORE Vidal
 Gristle McThornBody
 Hank Chesthair
 Hans Testosteroneson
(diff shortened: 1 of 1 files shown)


LazennG:
who

Darkstick:
> who

prebase regular, fun player

PR #11144, opened 2021-03-15
Fixes shuttle catastrophe event not happening

ChesterTheCheesy:
# General Documentation
Jamie, while fixing the event, in the check if the event can't happen, made it check if the shuttle in SHUTTLE_IDLE state (which is the state in centcom, in which we WANT the event to happen), instead of SHUTTLE_DOCKED state. (shuttle docked at station, we don't want the event to happen).

bugfixes are always good, definitely improves RP

Hopefully makes the shuttle catastrophe even happen again.

event fix

# Changelog
:cl:  
bugfix: hopefully makes the shuttle catastrophe event happen on its own again  
/:cl:

### DIFF
--- a/code/modules/events/shuttle_catastrophe.dm
+++ b/code/modules/events/shuttle_catastrophe.dm
@@ -6,7 +6,7 @@
 
 /datum/round_event_control/shuttle_catastrophe/canSpawnEvent(players, gamemode)
 	if((SSshuttle.emergency.name == "Build your own shuttle kit") || (SSshuttle.emergency.mode == SHUTTLE_CALL || SSshuttle.emergency.mode == SHUTTLE_RECALL || SSshuttle.emergency.mode == SHUTTLE_DOCKED))
-		return FALSE //don't undo manual player engineering, it also would unload people and ghost them, there's just a lot of problems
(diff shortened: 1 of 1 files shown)


Jarod1200:
Great event

code/modules/events/shuttle_catastrophe.dm
```diff
@@ -5,7 +5,7 @@
 	max_occurrences = 1
 
 /datum/round_event_control/shuttle_catastrophe/canSpawnEvent(players, gamemode)
-	if((SSshuttle.emergency.name == "Build your own shuttle kit") || (SSshuttle.emergency.mode != SHUTTLE_CALL || SSshuttle.emergency.mode != SHUTTLE_RECALL || SSshuttle.emergency.mode != SHUTTLE_IDLE))
```
redmoogle, line 8: if NOT in idle state return false

4head

ChesterTheCheesy: well then

ChesterTheCheesy: the jamie's code literally always returned false

redmoogle (approved):

PR #11147, opened 2021-03-16
makes rad collector research per minute "accurate"

SomeguyManperson:
# General Documentation

multiplies the per minute examine amount by 30 to account for how many times that value is given in a minute since it's actually giving the per process amount which is NOT once per minute

# Changelog
:cl:  
tweak: rad collectors now display an accurater amount of research per minute when examined
/:cl:


redmoogle (approved):
2*30 = 60

PR #11148, opened 2021-03-16
fixes buckling layer offsets

SomeguyManperson:
and then there were no  bugs with layers ever again :)
:cl:  
bugfix: you no longer get stuck on the above mob layer when using fireman carry and layer changes get passed properly
/:cl:


redmoogle (approved):
explains why chairs were being whacky hopefully

PR #11161, opened 2021-03-18
Allows cogscarabs to be picked up and transported during war.

ChesterTheCheesy:
# General Documentation
Once the herald beacon had been activated, clock cultists can pick up the cogscarabs to either wear as hats, or transport them onto station. This should allow for more "war" oriented gameplay, with cogscarabs either disrupting station from inside, or helping cultists.
Cogscarabs cannot use scriptures, so they cannot get back to Reebe themselves. They don't have any free access like most drones do, neither can they pass through doors nor ventcrawl. They are also slower than humans and cannot fire guns. This all means that they should be pretty balanced once released on station. Killing an alone scarab should be no problem at all.

Allows for more creative play for clock cultists during war, for instance, giving one many interrogation cogs and let it cog the maints while you focus on converting people, having an early pocket helper or just a fashionable hat.

# Wiki Documentation

It should be added to wiki that cogscarabs can be moved away from Reebe after declaring war to help the brave cultists.

COGGERS

war clockcult buff

# Changelog
:cl:  
rscadd: cogscarabs can now be picked up during war
imageadd: adds a sprite for cogscarab on head
/:cl:


alexkar598:
I'd rather not have cogscarabs be hogged by people who just try to murderbone, fail and die.

Scarabs are like the only ones to ever build defenses anyways so this seems like a bad idea

SomeguyManperson:
there's a reason scarabs aren't able to leave normally it's because they're supposed to be building defenses, if they can get to the station somehow they are weak as hell and would still be better off building on reebe

ChesterTheCheesy:
> 
> 
> I'd rather not have cogscarabs be hogged by people who just try to murderbone, fail and die.
> 
> Scarabs are like the only ones to ever build defenses anyways so this seems like a bad idea

I'd rather have a scarab help me convert rather than sit in Ark doing nothing because all power is out. Scarabs can't get on station on their own, if a cultist says they stay in Ark, they stay in the Ark. If a cultist says them they are coming with them, they have an option to refuse. Not sure what you mean by "hogging" scarabs there.


> 
> 
> there's a reason scarabs aren't able to leave normally it's because they're supposed to be building defenses, if they can get to the station somehow they are weak as hell and would still be better off building on reebe

During war, this ALLOWS for more aggressive high-risk/high-reward play by building a cult base on strategic areas like perma or cargo. Scarabs not fighting on the front, but rather fortifying and setting traps. By no means it will force people to do this sorta thing, but it'll definitely allow for more creative play.

You might as well put it on testmerge if you're unsure how people will like or not. Once again, it's the cultist's option if the scarab should be taken with him or not. Also, there are 3 cogscarabs, so even if one dies on station, it's not a huge loss (keep in mind you can easily revive those with a screwdriver, a tool every cultist starts with).

JamieD1:
Might need some kind of consent system so the scarab can say yes to being picked up and then I shouldnt see an issue with it.

ChesterTheCheesy:
Drones can always move away when picked up, move away from being picked up, also unlike drones, COGSCARABS CAN SPEAK!

JamieD1 (approved):

alao3alao3:
How many cogscarabs can fit into a bag? I thought there was an issue ages ago with scarabs piling out of clock cultists whenever they got downed making it hell to kill them.

PR #11166, opened 2021-03-19
turns out I fucked up a merge conflict resolution and the dragon chest only rolled for 3 of its items

SomeguyManperson:
:cl:  
bugfix: dragon's blood is now back in the dragon chest loot pool 
/:cl:


redmoogle (approved):
5head

PR #11169, opened 2021-03-19
Syndicate lavaland syndie snipers are no longer usable by crew

SomeguyManperson:
# at all

### ever

syndie lavaland snipers now have a super secure firing pin that explodes when someone not of the syndicate faction tries to fire it, or when its pin is removed
have fun

no more 19 dollar sniper rifles

# Changelog

:cl:  
tweak: lavaland syndicate sniper rifles are no longer safe to be fired by crew
tweak: some firing pin removal stuff actually gets called when it's supposed to be i.e. on firing pin removal
/:cl:


redmoogle (approved):
Interesting...

PR #11174, opened 2021-03-20
Experimental PR: Cap boxstation to 60 players

alexkar598:
Note this is 60 players on the server, regardless of if they're dead or alive

I'm not 100% sold on the idea but I want to see how people react to this. Box can get quite cramped on high pop and meta is a good box inspired map that gives more breathing room.

:cl:  
experimental: Box can no longer be chosen over 60 players
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -13,6 +13,7 @@ Format:
 	votable (is this map votable)
 
 map yogstation
+	maxplayers 60
 	voteweight 0.9
 	votable
 endmap


redmoogle:
May be better to be around 50-55

JamieD1:
So your solution is to force people to leave because they dont like the map they are forced to play on.. :thonk:

Cartlord:
I don't like this, especially considering Meta doesn't have gas miners.

PR #11175, opened 2021-03-20
Automatic DME Generator

redmoogle:
Generates the DME automagically for you

Used it to go ahead and clean up duplicate includes(eg yogstation/randommodule -> randommodule)

It has also been integrated into the CBT builder so you dont have to _worry_ about ticking files

might add the dme to the gitignore so it is (_should_) more useful


redmoogle:
Awaiting rewrite

PR #11187, opened 2021-03-22
[admin] unclaimed tickets only bwoink for 1 ticket instead of stacking 10 bwoinks

alexkar598:
it usally goes:

> puts on headset
> logins to yogstation

Ticket #1 unclaimed!
Ticket #2 unclaimed!
Ticket #3 unclaimed!
Ticket #4 unclaimed!
Ticket #5 unclaimed!
Ticket #6 unclaimed!
> Gets ears destroyed
> Remove headset
> Handle tickets


redmoogle:
Replying with adminhelp button stacks this might wanna change that :eyes:

alexkar598:
What

redmoogle:
> What

If you hit the ahelp button to add an addendum and hit yes to keep it open its treated as a new ticket but its still ticket X

JamieD1 (approved):

PR #11188, opened 2021-03-22
(port) adds warp implant

boodaliboo:
Original Pr https://github.com/Citadel-Station-13/Citadel-Station-13/pull/12991
"Adds the warp implant, an implant that makes use of the chem eigenstasium to save your location and then warp you back there after a few seconds."
:cl:  
rscadd: Adds new implants for traitors to mess with
/:cl:


SomeguyManperson:
if you are going to port stuff from a largely different codebase make sure it actually works first

PR #11192, opened 2021-03-22
Nerfs tesla engine by making miniballs only act 20% of the time, also makes tesla bolts no longer bounce indefinitely

ChesterTheCheesy:
# General Documentation
Nerfs the tesla to not create as much ridiculous amounts of power as it does, since each miniball zap has half the charge of tesla itself. By simple math, currently it means tesla makes 1(main tesla) + 5 (10 miniballs) = 6 tesla power. With change, it'll make 1 + 5/5 = 2, theoritically reducing the power output of tesla engines 3 times. It also means, that once the tesla goes loose, instead of 11 explosions each tick, it makes around 3, hopefully being enough for the server to catch up, reducing the lag.

Added: tesla bolts, now upon hitting a structure, bounce twice, if the shock came from tesla, or once, if it came from miniball. No more infinite loops because whole station is made of conductive metal.

nerfs tesla and makes it lag less during tesloose

miniballs now act only 20% of time, tesla bolts no longer bounce between objects indefinitely

Engineering tweak

miniballs now act only 20% of time, meaning power output should be 3 times less for tesla

# Changelog
:cl:  
tweak: nerfs tesla by making the miniballs only act 20% of the time
/:cl:


alexkar598:
Was this tested or is this a repeat of when the singulo got made useless because someone decided to remap engineering in a way that broke the rad collectors

ChesterTheCheesy:
Tested, it does still make enough power for an entire station, but now explodes way less in case of tesloose, since there's less shocks and shocks don't bounce indefinitely.

ChesterTheCheesy:
Gonna close and reopen this because i can't edit the repository that i accidentally deleted

PR #11194, opened 2021-03-22
[admin] Blocks the creation of new player initiated admin tickets after the round has ended

alexkar598:
You may be wondering: WHY THE FUCK IS THIS A GOOD IDEA and fair enough it sounds stupid but hear me out.
The way things happen is usually:
* something happens
* they wait 30 minutes until round end
* they see that person wasnt an antag and make a ticket
* that ticket is painful to handle as the events happened 30 minutes ago
* round restart is delayed
* we lose 15 pop because noone likes to not be playing the game

People should be encouraged to ahelp when something happens, not at the end of the round and if they really must, then a player complaint is a better solution as it doesnt delay the restart by 15 minutes. And yes, I do know player complaints dont really get resolved quickly but its a better alternative than delaying restarts.

:cl:  
tweak: You can no longer create admin tickets(ahelps) after the round has ended. You can still respond to existing tickets or as an admin, create new ones to ask people questions.
/:cl:

### DIFF
--- a/yogstation/code/modules/admin/verbs/adminhelp.dm
+++ b/yogstation/code/modules/admin/verbs/adminhelp.dm
@@ -797,8 +797,13 @@ GLOBAL_DATUM_INIT(ahelp_tickets, /datum/admin_help_tickets, new)
 	set category = "Admin"
 	set name = "Adminhelp"
 
+	//Yes theres nothing after game finished, no i dont care, in the future there might
+	if(SSticket.current_state >= GAME_STATE_FINISHED)
+		to_chat(src, "<span class='danger'>The round has already ended! Make a player complaint on the forums instead.</span>, confidential=TRUE")
+		return
+	
 	if(GLOB.say_disabled)	//This is here to try to identify lag problems
-		to_chat(usr, "<span class='danger'>Speech is currently admin-disabled.</span>", confidential=TRUE)
+		to_chat(src, "<span class='danger'>Speech is currently admin-disabled.</span>", confidential=TRUE)
 		return
 
 	//handle muting and automuting
--- a/yogstation/code/modules/admin/verbs/adminhelp.dm
+++ b/yogstation/code/modules/admin/verbs/adminhelp.dm
@@ -798,7 +798,7 @@ GLOBAL_DATUM_INIT(ahelp_tickets, /datum/admin_help_tickets, new)
 	set name = "Adminhelp"
 
 	//Yes theres nothing after game finished, no i dont care, in the future there might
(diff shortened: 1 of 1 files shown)


Colburn3000:
The best time to handle tickets accurately is in game. Admins have the most info at their fingertips, communication channels are the most clear, logs are right in front of the admin, etc. It is up to admin discretion if a particular ticket is going to stall out a round too long. This shouldn’t be hardcoded.

ChesterTheCheesy:
It is kinda stupid, considering there's no way you can tell if someone was self-antagging or not until the roundend report shows up.

alexkar598:
That's why you ahelp when the round is ongoing. This encourages ahelping during the round instead of after seeing the round end report.

Colburn3000:
There many valid reasons to ahelp after the round. Issues happen right at round end. Shuttle grief happens all the time. People may not realize someone wasn’t an antag. 

Spending a few minutes after the round to properly handle a ticket is better than forcing anyone with after round issues to make a complaint. It better serves our players and gives our staff the best chance to make a timely decision.

maxion12345:
time to ahelp every time i see someone do antag-worthy things to check they are antag, that assistant might have stolen that esword

Mesalikepie:
Issues happen at the end of round such as on the shuttle. These sometimes take a second to ahelp, which is fair but annoying. We can't stop all attempts to ahelp just because a few ahelps come in that we hate to handle

alexkar598:
> 
> 
> time to ahelp every time i see someone do antag-worthy things to check they are antag, that assistant might have stolen that esword

Yes please do, it takes half a second to check if someone is an antag, it takes 5 minutes to figure out why a non antag assistant had an esword 30 minutes ago

alexkar598:
If something happens on the shuttle, its usually ahelp'd while on the shuttle before round end and worst case scenario, admins aren't completly blind, they can see the messages saying 15 people just died on the shuttle and that a boh and a maxcap was triggered on the emergency shuttle.

MegaEmpirical:
People can't always make tickets before the round is over when it's stuff on the shuttle. This should absolutely not be a thing. It's up to the admins online to decide if they need to defer that person to make a complaint or not.

Colburn3000:
> If something happens on the shuttle, its usually ahelp'd while on the shuttle before round end and worst case scenario, admins aren't completly blind, they can see the messages saying 15 people just died on the shuttle and that a boh and a maxcap was triggered on the emergency shuttle.

Shuttle grief isn’t always a massive obvious event. Quit being dense.

PR #11195, opened 2021-03-22
Removes hard delete

alexkar598:
Tries to reduce visible lag.

# FAQ:
__Won't this cause a memory leak?__
 Yes, yes it will. This needs to be test merged before merge to ensure there's no crazy memory leak that will make the game crash

__Won't this cause some items with broken logic to break and think something exists when it doesn't?__
 Yes, yes it will. This needs to be test merged before merge to ensure there's no bad Destroy() logic that will cause undefined behaviour. Note that such behaviour already exists, only thing is that its time limited and corrects itself after a while. With this PR, it won't correct itself.
 
 __Won't this cause the game to feel less laggy?__
  It will reduce stuttering, it will somewhat help with the stuff like lights lagging behind but not by much. The main effect of this PR is input lag and other client sides stuttering which are caused by BYOND working in overtime, ie: when it starts an operation or proc that runs for so long it runs into the time where the next tick is supposed to be running, at that point, byond will ignore that tick, causing the stuttering as the client is expecting a network response but receives none.
  
__Is this PR tested?__
 It works and prevents hard deletes yes, if its going to crash while live, i have no idea. Only live testing will reveal that

This absolutely needs to be test merged before merge to ensure this won't OOM the server since some objects aren't being cleaned up

:cl:  
experimental: Removes hard deletes
/:cl:

### DIFF
--- a/code/controllers/subsystem/garbage.dm
+++ b/code/controllers/subsystem/garbage.dm
@@ -161,6 +161,7 @@ SUBSYSTEM_DEF(garbage)
 				var/datum/qdel_item/I = items[type]
 				testing("GC: -- \ref[src] | [type] was unable to be GC'd --")
 				I.failures++
+				return
 			if (GC_QUEUE_HARDDELETE)
 				HardDelete(D)
 				if (MC_TICK_CHECK)
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -20,7 +20,7 @@
   *
   * Parent call
   *
-  * Returns QDEL_HINT_HARDDEL (don't change this)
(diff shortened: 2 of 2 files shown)


Cartlord:
So you have a PR that you openly admit adds a memory leak, and it will likely break a bunch of stuff, but you want to add it? Is the input lag really that much of an issue?

alexkar598:
It's a trade off between reducing overrunning ticks and losing a small amount of memory until the server is rebooted. This has been tested 3 rounds and hasn't caused the server to crash due to running out of memory.

JamieD1:
Didnt work, Lag still awful

PR #11204, opened 2021-03-23
Removes Security Records Console Board from 5x4_deltadetective

alao3alao3:
# General Documentation

Removes sec records console board from 5x4_deltadetective.

Enforces the idea that sec boards, no matter how mundane or what little use it has, always falls within a secure area (checkpoints, brig) with the only exception being tech storage. Why the couldn't argue that point on the discord when defending this is beyond me.

# Wiki Documentation

Random maints are not documented.

# Changelog

:cl:   
rscdel: Removed sec records board from 5x4_deltadetective
/:cl:


patpol4 (approved):
good

PR #11212, opened 2021-03-24
de-shittifies the antimatter a bit

SomeguyManperson:
# General Documentation

replaces some values with defines, makes true/false stuff true/false instead of 1/0, and automates core icon updates and core visible stability updates

ANCIENT CODE

# Changelog

:cl:  
tweak: the antimatter is slightly more automatic and actually updates its icons while producing power
/:cl:


redmoogle (approved):
Interesting...

PR #11216, opened 2021-03-25
the tacticool fake syndicate turtleneck now has suit sensors

SomeguyManperson:
# General Documentation

makes the tacticool turtleneck that's purchaseable from a hacked clothes vendor actually have suit sensors like the majority if not all other clothing items

lack of sensors on the tactical turtleneck is because it's syndie shit and doesn't need them, the tacticool turtleneck is a marketable plushie variant that would have them similar to the gladiator outfit from the autodrobe

# Changelog

:cl:  
tweak: tacticool turtlenecks now have suit sensors
/:cl:


redmoogle (approved):
b

PR #11226, opened 2021-03-25
Moves several APCs into Maintenance or into their rooms (REDONE)

alao3alao3:
Redone due to arbitrary mapcode error
# Intent of your Pull Request

Move several APCs out into maintenance (or in on one occasion) in accordance with how all the other APCs are.
The following departments are affected:
Cryopod Room
Vacant Commisary
Mining Dock
Construction Area
Morgue
Mech Bay
Dormitory Bathroom
Incinerator (APC moved into the incinerator room on the basis that Incinerator is now a part of Atmos and is therefore secure)
Labor Shuttle Dock (APC moved up one tile so it's actually on a wall apart of the shuttle dock. Still in maintenance however)

Note: Medbay APCs with the exception of morgue are purposely overlooked due to their importance.
Command APCs have been overlooked due to "antag play"

# Why is this change good for the game?

Uniformity, so anyone who needs to find the APC knows where to look instead of it arbitrarily being inside the damn room or outside for no reason.

# Wiki Documentation

Images for the above mentioned departments may need to be changed.

# Changelog

:cl:  
rscadd: Adds APCs into rooms or into maintenance 
rscdel: Removes APCs from said rooms or respective maintenance  
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -20871,6 +20871,18 @@
 "biJ" = (
 /turf/open/floor/circuit,
 /area/science/robotics/mechbay)
+"biM" = (
+/obj/machinery/atmospherics/pipe/manifold/supply/hidden/layer2{
+	dir = 1
+	},
+/obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer4{
+	dir = 4
+	},
+/obj/machinery/light{
+	dir = 1
+	},
+/turf/open/floor/plasteel/freezer,
+/area/crew_quarters/toilet)
 "biN" = (
 /obj/machinery/atmospherics/components/unary/vent_pump/on/layer2{
 	dir = 8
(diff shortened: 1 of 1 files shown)


Cartlord:
Since the last PR I said this on got closed, I'll say it here again: This is a problem for preternis, since they rely on APCs to live; the fewer APCs are easily accessible, the harder it is to play preternis since it means your food source is less accessible.

alao3alao3:
They'll have to stick to purposely public APCs in halls, or get maint access if they know whats good for them. APCs were never designed to be available for Preterni to eat out of and players playing that race should recognize and play bearing that risk in mind.

Cartlord:
APCs may not have been initially designed with preternis in mind, but I still think there should be SOME thought put into it, given that preternis ARE here and - in my opinion - shouldn't be a chore to play. I don't think playing a given race should necessitate getting in contact with HoP just to have some convenience in eating rather than wandering to find one of the few public APCs, esp. considering that (admittedly, more due to medical not knowing what to do with them than anything with the race itself) playing preternis is already a bit of a pain.

patpol4:
Reminder every department can access it's own maint

patpol4 (approved):
This has already been reviewed in the previous pr, looks good to me

PR #11241, opened 2021-03-28
Lets abductors buy more batons and pistols

Marmio64:
They can buy pretty much everything else they start with, so why not their baton and pistol? Allows abductors to continue to use only alien gear, and not have to make a stun prod if someone pushes them.

Lets abductors rebound from mistakes if they get enough patients.

you can buy more batons and guns as abductor

For example, Engineering changes, Medical rebalancing, TEG tweaks, etc.?
antag changes

no?

:cl:  
rscadd: Adds abductor batons and guns to their purchase menu
/:cl:

### DIFF
--- a/code/modules/antagonists/abductor/equipment/orderable_gear.dm
+++ b/code/modules/antagonists/abductor/equipment/orderable_gear.dm
@@ -30,13 +30,8 @@ GLOBAL_LIST_INIT(abductor_gear, subtypesof(/datum/abductor_gear))
 	name = "Advanced Baton"
 	description = "A advanced baton with four modes allowing it to stun, sleep, cuff, and probe victims."
 	id = "baton"
+	cost = 2
 	build_path = /obj/item/abductor/baton
-
-/datum/abductor_gear/alien_pistol
-	name = "Agent Pistol"
-	description = "A small self defence pistol for dealing with rowdy patients with blasts of high-intensity radiation."
-	id = "alien_pistol"
-	build_path = /obj/item/gun/energy/alien
 
 /datum/abductor_gear/radio_silencer
 	name = "Radio Silencer"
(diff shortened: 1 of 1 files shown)


Sniblet:
Me: *buys a second pistol even though i already have one*
Also me: My name is not important. What matters is what I'm going to do.
xDDDD

alexkar598:
Dual weilding the gun is concerning

SomeguyManperson:
would suggest not allowing them to buy more guns since they inherently don't tend to actually need the first one

Marmio64:
you don't tend to need the first gun but if you lose it, you've lost your only means of defense against a hulk if you encounter one and you cant get out. I could double the price or something maybe, or perhaps they wouldn't be rechargeable, so you'd want to be conservative with their use even more. Otherwise I suppose it could stay out of the purchase menu since they don't see much use anyhow.

alexkar598:
Also not 100% on board with the idea of the intentionally vulnerable scientist being given a stun baton

Marmio64:
> Also not 100% on board with the idea of the intentionally vulnerable scientist being given a stun baton

they can already buy a full set of agent armor for themself and link it up to the console and enable combat mode, and getting a stunprod is easy enough, the abductor baton just makes everything easier

Ratvar:
If scientist and guard both teleport bot to alien stunprod everyone, they will not be able to teleport guard back in case of messing up. 
This also helps scientist keep abducting people if guard dies or is afk, being useless is boring

JamieD1:
Theos and Myself do not think they should be able to get guns.

SomeguyManperson (changes requested):
remove the gun and make the baton worth 2 points

Marmio64:
> remove the gun and make the baton worth 2 points

done

PR #11245, opened 2021-03-29
fix wings not working due to mothism

SomeguyManperson:
makes the moth wing check only prevent flight if you are a moth
:cl:  
bugfix: random people will no longer be incapable of flight
/:cl:


ChesterTheCheesy:
epic

PR #11246, opened 2021-03-29
removes code for "practice" scarabs since it was never actually used and could be used to gain infinite power

SomeguyManperson:
not like it matters since they shouldn't be spawning before the gamemode starts any more but
:cl:  
bugfix: old cogscarab code that at some point was used to gain infinite power has been removed
/:cl:


Cartlord:
does this make clockcult less maddening

SomeguyManperson:
no

PR #10971, opened 2021-02-07
Vitamin buff and removal from plants + frying and sugar nerf

Marmio64:
### Intent of your Pull Request
To makes botany plants and chemistry slightly less good at replacing the chef's job, and to make frying all your food as a chef not be the clear way of feeding the station

also the intent is to massively improve RP, like, merging this will QUADRUPLE the amount of rp in the game, not even joking.

### Why is this change good for the game?
Encourages making crafted food as a chef, since other food sources aren't as good now and vitamins heal more, along with standard nutriments healing less.

# Wiki Documentation

### Briefly describe your PR and the impacts of it, in layman's terms. 
Removes all vitamins from ALL plants. Nutriment amount is kept.
Makes vitamins heal 50% more and nutriments heal 50% less.
Cooking oil and sugar is much less nutritious.

### What should players be aware of when it comes to the changes your PR is implementing?
that fried shit and sugar isn't so good at feeding you now, eat something healthy you fucks

### What general grouping does this PR fall under? 
Chef, botany, and chem changes

### Are there any aspects of the PR that you would like us not to mention on the Wiki?
no

### If there are any numerical values involved in your PR that will be relevant to a player, please note them here. 
vitamins heal 1.5 brute and burn and nutriments only heal 0.5 brute now.

# Changelog

:cl:  
tweak: sugar and cooking oil less nutritious, vitamins heal more, nutriment heals less, no more vitamins in plants.  
/:cl:

PR #11311, opened 2021-04-08
(partially) Revert "Vitamin buff and removal from plants + frying and sugar nerf"

Sniblet:
This reverts commit 92fd210fd44232854d4268817d3528f3b111345b.
https://github.com/yogstation13/Yogstation/pull/10971

# General Documentation

Makes vegetables healthy again. I don't see why their vitamin removal makes any sense from a balance perspective, and it's certainly unreasonable logically.
If you want cooks to be better than botanists for food, fine, but there's no need to do it through nerfs and to make the station completely unable to access vitamins without a cook onboard.
I don't really know or care about the cooking oil and sugar changes, I'm not reverting those

Going back for another half-dozen carrots every 5-10 minutes is not engaging spaceman gameplay
Going back for another three carrots every 15-20 minutes was much more exciting

Vegetables are now liable to keep you full and healthy for more than 5 minutes per visit to botany

Service and nutrition

# Changelog

:cl:  
tweak: vitamins heal less, nutriment heals more, plants have vitamins again because they're not junk food
/:cl:


ynot01:
botany buf

Marmio64:
keep the oil and sugar changes.

The reasoning for removing vitamins from plants was to make prepared food better since botany just always takes over the chefs job for him. Botany really doesn't need to be something that can make the chef totally useless and there isn't any good ways to buff chef food other than just more mood buffs, which practically nobody has on. Also I never have any problems with food, even now. I eat fruits till i'm full and then pocket another one for later and I only ever have to make one or two trips back during the course of a shift. And, you need to edit the PR title so that people understand from a glance what it is reverting.

Cartlord:
Something I'm actually surprised by, is that nobody here seems to have considered that Botany can make their own meat - either via meatwheat steak & jerky, or by using the biogen to make monkey cubes.

code/modules/hydroponics/grown/tea_coffee.dm
```diff
@@ -83,7 +83,7 @@
 	plantname = "Coffee Robusta Bush"
 	product = /obj/item/reagent_containers/food/snacks/grown/coffee/robusta
 	mutatelist = list()
-	reagents_add = list(/datum/reagent/medicine/ephedrine = 0.1, /datum/reagent/toxin/coffeepowder = 0.1)
+	reagents_add = list(/datum/reagent/medicine/ephedrine = 0.1, /datum/reagent/consumable/nutriment/vitamin = 0.04, /datum/reagent/toxin/coffeepowder = 0.1)
```
Cartlord, line 86: Why would COFFEE have vitamin?

Cartlord:
Something I'm noticing is that nearly EVERY edible plant seems to have vitamin, I'm frankly opposed to that - melons, for instance, already provide insane nutriment, why give them vitamin too?
I think it might be fine on mutated plants (not the Ambrosia chain, because bloody everyone goes for those) or weeds, but it seems ludicrously to get in your code.
I don't know if you play Botany, so I'll assume you don't know it too well. A good Botanist, or even a decent one, can make dozens of plants per minute. If those plants have vitamin, there's screw-all stopping them from just bagging the plants, grinding them, and using the CondiMaster to extract pure vitamin in HORRIFIC quantities.
I know the DNA manipulator can Extract the vitamin content gene, but at least if it's on a mutate plant (especially one people don't usually go for, like Spaceman's Trumpet or something) they need to go OUT OF THEIR WAY to get it.
Maybe put it on sweet potatoes? They're a food item, a plant you don't see much, and a mutation - so you can make vitamin, but you at least need to DO something to get it.

PR #11339, opened 2021-04-12
Adds a Gallery to the Library and Places the Sec Service Outpost at Escape

alao3alao3:
# General Documentation

Adds a painting gallery to the library and places the sec service outpost at escape.
cpng
Note that an entire side of the library was destroyed and rebuilt, as opposed to just removing the sec outpost and placing the gallery in, and is the reason some things like lights, pipes or otherwise vary. The secure art enclosure (the blue table area I stole from TG's meta) is only accessible by the curator.

bpng
The air alarm has been placed in that bottleneck hall leading to the podbay dock area. The airlock that used to be where the Sec Service outpost is has been shoved down a bit, but does not extend past the escape airlock arms so it shouldn't cause any issues with any of the shuttles.

Gallery is cool and allows for the critique of art against each other, as opposed to the mindless and random paintings dotted about that only look nice and serve as decor, though they're still around.
This was done at the request of Lyis so if you want reasoning behind it ~~ping council~~ ping Lyis, a person who is apart of council.

# Wiki Documentation

Image of Library and Escape has to be changed

# Changelog

:cl:  
rscadd: Added a gallery in the Library and the Security Service Outpost to Escape  
rscdel: Removed the old Security Service Outpost  
/:cl:

Council and HM wanted this in. I think its getting in.


Marmio64:
rip service sec post, you're now just the escape checkpoint...

alao3alao3:
if you’ve got a spot to put the damn service post go ahead and suggest it

patpol4 (commented):
i hate map conflicts

patpol4 (approved):
Anyway, i spotted some issues like :
vivaldi_t0bRZwV9IA
vivaldi_Sm531Pze2f
vivaldi_oIgMm9cvov
Also added sidings to the carpet/wood border so it looks nicer and moved an air alarm that was on the same wall as a fire extinguisher.
Was thinking about adding a desk for the sec checkpoint but idk.
Will merge tommorrowish to leave time for this to get more feedback

Marmio64:
> if you’ve got a spot to put the damn service post go ahead and suggest it

remove clerk since he is just a useless validhunter and replace it with the service sec post boom, no need to thank me

PR #11340, opened 2021-04-12
I FIX ALL THE PIPES (I think)

Readystorm546:
NO MORE BAD PIPES (I Think)

To fix all pipes around box

Pipes being on the right layer is good

:cl:  
bugfix: fixed a lot of pipes being on the wrong layer on box 
/:cl:


patpol4 (changes requested):
You're reverting the recent library change now, also please leave that disposals pipe above library as it was

PR #11359, opened 2021-04-13
Blob Buff?? (Up for discussion)

Readystorm546:
# General Documentation

To buff the bob. Blob was a start of the round antagonist then was moved to mid-round. Blob has a hard time going up against crew when blob appears. This is due to blob appearing after the crew has mechs and the entire tech tree at their disposal. Max health increased for blobernaught from 200 to 300 meaning that blobernaught will need to stay in the blob to get to his full health. Strain re-roll cost 15 down from 40 is pretty self-explanatory

Other ideas I have are things such as making the message of level 5 biohazard take longer to appear based of how long the round has been going on and other factors. No idea how to do this though

Blob is underpowered and this is my attempt to change that.

# Wiki Documentation

Max health increased for blobernaught to 300 from 200.
Strain re-roll cost 15 down from 40

Antag rebalance

Max health increased for blobernaught to 300 from 200
Strain re-roll cost 15 down from 40

# Changelog

:cl:  
experimental: Max health increased for blobernaught from 200 to 300
experimental: Strain re-roll cost 15 down from 40
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blob_mobs.dm
+++ b/code/modules/antagonists/blob/blob_mobs.dm
@@ -201,7 +201,7 @@
 	icon_living = "blobbernaut"
 	icon_dead = "blobbernaut_dead"
 	health = 200
-	maxHealth = 200
+	maxHealth = 300
 	damage_coeff = list(BRUTE = 0.5, BURN = 1, TOX = 1, CLONE = 1, STAMINA = 0, OXY = 1)
 	melee_damage_lower = 20
 	melee_damage_upper = 20
--- a/code/modules/antagonists/blob/powers.dm
+++ b/code/modules/antagonists/blob/powers.dm
@@ -338,9 +338,9 @@
 
 /mob/camera/blob/verb/strain_reroll()
 	set category = "Blob"
-	set name = "Reactive Strain Adaptation (40)"
+	set name = "Reactive Strain Adaptation (15)"
 	set desc = "Replaces your strain with a random, different one."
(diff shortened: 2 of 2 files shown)


alexkar598:
I'd rather have the blob relocate buffed rather than the reroll tbh, a blob rerolling constantly is annoying

Readystorm546:
> I'd rather have the blob relocate buffed rather than the reroll tbh, a blob rerolling constantly is annoying

This would also waste the blob's time due to it being cheaper. He would always be searching for the right strain but he could never find it. But also makes it so he can readapt to the current thing attacking him. I.E he's blazing oil crews using fire extinguishers he swaps to stop the crew etc,,

N3D6:
honestly i think blob is fine as it is

Marmio64:
i mean yeah blob is very weak when the crew has matts and weapons research done and they're not under threat from another round ender antag, but they're also freakishly strong when the crew doesn't have those.

UselessTheremin:
> > I'd rather have the blob relocate buffed rather than the reroll tbh, a blob rerolling constantly is annoying
> 
> This would also waste the blob's time due to it being cheaper. He would always be searching for the right strain but he could never find it. But also makes it so he can readapt to the current thing attacking him. I.E he's blazing oil crews using fire extinguishers he swaps to stop the crew etc,,

yeah you just undid your point its essentially being able to pick and choose whatever playstyle you want whenever, its less than HALF the cost of it previously, i still dont see why blob needs a buff ive heard nobody complain about blob being too weak once

Readystorm546:
guess its a minority.

PR #11363, opened 2021-04-14
changes bath salts to a actual drug name

boodaliboo:
other drugs in game arent called by there street names so why should bath salts its been renamed to the most popular bath salt 3-mmc

:cl:  
tweak: changed named of bath salts
/:cl:


athena413:
maybe remove the 3

start using the fucking template

UselessTheremin:
why though

alexkar598 (changes requested):
Template

PR #11364, opened 2021-04-14
Increases the dispatch cooldown to 25 seconds

ynot01:
# General Documentation

Increases the dispatch cooldown to 25 seconds

Shuttle dispatch spam happens every round

# Changelog

:cl:  
tweak: Increased the dispatch cooldown to 25 seconds
/:cl:

### DIFF
--- a/yogstation/code/modules/clothing/masks/hailer.dm
+++ b/yogstation/code/modules/clothing/masks/hailer.dm
@@ -8,7 +8,7 @@
 	var/obj/item/radio/radio //For engineering alerts.
 	var/radio_key = /obj/item/encryptionkey/headset_sec
 	var/radio_channel = "Security"
-	var/dispatch_cooldown = 20
+	var/dispatch_cooldown = 250
 	var/last_dispatch = 0
 
 /obj/item/clothing/mask/gas/sechailer/Destroy()


alexkar598:
its funny

N3D6:
Dispatch officer responding to crime in progress. Requesting immediate assistance.
Spamming sec comms with a message that gets obnoxious over time isn't funny, Alex

athena413:
would rather just have it be a clean 30 seconds or even a minute - you're not gonna be "responding to a crime" more than once a minute i'd think

maxion12345:
Original cooldown was meant to be 20 seconds but we do things deciseconds so

oops

alexkar598:
But more seriously, crimes often escalate, the only time I've seen this actually used was when the warden got ambushed by a ling and the first message of assault then a second message of murder in progress got the message across and the warden got assistance at brig.

Unless things are escalating quickly, theres no reason to use this so adding a cooldown fucks with the arguably only valid reason for the dispatch call.

Cartlord:
Maybe make a per-crime cooldown? Like you have a cooldown between two calls of Assault, but not between calling Assault and calling Murder? Is that even possible?

ynot01:
> Maybe make a per-crime cooldown? Like you have a cooldown between two calls of Assault, but not between calling Assault and calling Murder? Is that even possible?

possible, but defeats the point of this PR

Cartlord:
Well it adds SOME cooldown, while still letting people call out multiple crimes fast in the event of an escalating situation.

Cartlord (changes requested):
Maybe at least reduce the cooldown from 25 secs to 5 or so? 25 seconds is a bloody long time I think, and if you change it to 25 seconds that'll make it nonfeasible to call out multiple different crimes in the event of an escalating situation.
EDIT: Ah crap I did it wrong and I don't know how to delete the review or this comment

yogstation/code/modules/clothing/masks/hailer.dm
```diff
@@ -8,7 +8,7 @@
 	var/obj/item/radio/radio //For engineering alerts.
 	var/radio_key = /obj/item/encryptionkey/headset_sec
 	var/radio_channel = "Security"
-	var/dispatch_cooldown = 20
+	var/dispatch_cooldown = 250
```
Cartlord, line 11:
```suggestion
	var/dispatch_cooldown = 50
```
Just change it to 5 secs maybe, so it's a bit harder to spam but people can still call multiple crimes in the event of an escalating situation?

Cartlord (changes requested):

yogstation/code/modules/clothing/masks/hailer.dm
ynot01: no

Cartlord: Why not? You've had the (IMO) legitimate criticism brought up that raising the cooldown by the amount initially proposed would make it impossible to effectively report escalating situations. If someone's spamming Sec comms with false crimes, why not AHelp it? I'd imagine that spamming Sec comms, to the point that legitimate reports get buried, violates rule 0.0 of not being a dick.

PR #11369, opened 2021-04-15
should fix the byos not accounting for byos specific things

SomeguyManperson:
airless
:cl:  
bugfix: you can now buy engines for the byos and it will only be purchasable in thef irst 10 minutes of the round as intended
tweak: BYOS purchase window increased from 10 to 20 minutes
/:cl:

### DIFF
--- a/code/datums/shuttles.dm
+++ b/code/datums/shuttles.dm
@@ -166,18 +166,18 @@
 	admin_notes = "No brig and no medical facilities. Build YOUR own."
 	credit_cost = 5000
 
-/datum/map_template/shuttle/emergency/construction_small
+/datum/map_template/shuttle/emergency/construction/small
 	suffix = "construction_small"
 	name = "Build Your Own Shuttle, Jr."
 	description = "The full-size BYOS too big for your taste? Aside from the reduced size and cost, this has the all same (lack of) amenities as its full-sized sibling."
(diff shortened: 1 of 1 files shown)


code/datums/shuttles.dm
```diff
@@ -166,18 +166,18 @@
...
-/datum/map_template/shuttle/emergency/construction_small
+/datum/map_template/shuttle/emergency/construction/small
 	suffix = "construction_small"
 	name = "Build Your Own Shuttle, Jr."
 	description = "The full-size BYOS too big for your taste? Aside from the reduced size and cost, this has the all same (lack of) amenities as its full-sized sibling."
 	admin_notes = "No brig and no medical facilities. Build YOUR own."
 	credit_cost = 2000
 
-/datum/map_template/shuttle/emergency/airless/prerequisites_met()
+/datum/map_template/shuttle/emergency/construction/prerequisites_met()
```
ChesterTheCheesy, line 176: Erm, doesn't that make the actual BYOS only buyable on first 10 minutes?

Jarod1200:
Ngl first 10 minutes is really short. I think it should be 15 or 20

code/datums/shuttles.dm
SomeguyManperson: small one's typepath was changed so it inherits this stuff

PR #11374, opened 2021-04-16
adds a cosmetic vampire coat to clowns vending machine

boodaliboo:
allows for clown to do more gimicks like pretend to be dracula and stuff 
:cl:  
rscadd: Added a cosmetic vampire coat to clowns vending machine
/:cl:

### DIFF
--- a/code/modules/clothing/suits/miscellaneous.dm
+++ b/code/modules/clothing/suits/miscellaneous.dm
@@ -651,7 +651,7 @@
 
 /obj/item/clothing/suit/cosmeticvampirecoat
 	name = "Vampire Coat"
-	desc = "A Coat designed to mimic the coats of the vampires from translyvania"
+	desc = "A Coat designed to mimic the coats of the vampires from Translyvania."
 	alternate_worn_icon = 'yogstation/icons/mob/suit.dmi'
 	icon = 'yogstation/icons/obj/clothing/suits.dmi'
 	icon_state = "draculacoat"
--- a/code/modules/clothing/suits/miscellaneous.dm
+++ b/code/modules/clothing/suits/miscellaneous.dm
@@ -650,7 +650,7 @@
 	alternate_worn_layer = UNDER_HEAD_LAYER
 
 /obj/item/clothing/suit/cosmeticvampirecoat
-	name = "Vampire Coat"
+	name = "Collectible Vampire Coat"
 	desc = "A Coat designed to mimic the coats of the vampires from Translyvania."
 	alternate_worn_icon = 'yogstation/icons/mob/suit.dmi'
 	icon = 'yogstation/icons/obj/clothing/suits.dmi'
--- a/code/modules/clothing/suits/miscellaneous.dm
+++ b/code/modules/clothing/suits/miscellaneous.dm
@@ -648,3 +648,15 @@
 	w_class = WEIGHT_CLASS_TINY
 	flags_inv = HIDEGLOVES|HIDEEARS|HIDEFACE|HIDEHAIR|HIDEFACIALHAIR
 	alternate_worn_layer = UNDER_HEAD_LAYER
+
+/obj/item/clothing/suit/cosmeticvampirecoat
+	name = "Vampire Coat"
+	desc = "A Coat designed to mimic the coats of the vampires from translyvania"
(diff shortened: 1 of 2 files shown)


athena413:
also is secretly an attempt to give the vampire coat a sort of metashield by the fake one existing

use the template fuckface

code/modules/vending/autodrobe.dm
```diff
@@ -187,6 +187,7 @@
 					/obj/item/clothing/shoes/clown_shoes/beeshoes = 1,
 					/obj/item/clothing/suit/santa = 1,
 					/obj/item/clothing/head/santa = 1,
+					/obj/item/clothing/suit/cosmeticvampirecoat,
```
athena413: make it have only one like all of the other autodrobe items

code/modules/clothing/suits/miscellaneous.dm
```diff
@@ -648,3 +648,15 @@
 	w_class = WEIGHT_CLASS_TINY
 	flags_inv = HIDEGLOVES|HIDEEARS|HIDEFACE|HIDEHAIR|HIDEFACIALHAIR
 	alternate_worn_layer = UNDER_HEAD_LAYER
+
+/obj/item/clothing/suit/cosmeticvampirecoat
+	name = "Vampire Coat"
+	desc = "A Coat designed to mimic the coats of the vampires from translyvania"
```
athena413: this description isnt very good and it doesnt have proper grammar

athena413 (changes requested):

Tipy1802:
This can result in two situations:
a) whoever wears the cosmetic cloak gets killed by sec and sec gets banned
b) vampires are allowed to wear their cloak that gives them near infinite blood (its not infinite but blood regen is really good and just a few seconds will earn you enough to jaunt), a cloak that is also supposed to be their way of going loud

None of those are attractive situations, people getting killed and banned is not fun and vampires don't need to be made even more OP than they currently are

boodaliboo:
> This can result in two situations:
> a) whoever wears the cosmetic cloak gets killed by sec and sec gets banned
> b) vampires are allowed to wear their cloak that gives them near infinite blood (its not infinite but blood regen is really good and just a few seconds will earn you enough to jaunt), a cloak that is also supposed to be their way of going loud
> 
> None of those are attractive situations, people getting killed and banned is not fun and vampires don't need to be made even more OP than they currently are

that is a rules issue  sec doesnt just go kill clowns wearing wizard robes that is a frankly ridiculous criticism,
your second criticism makes no sense this doesnt give them blood regen

ToasterBiome:
Maybe change the name so it's obviously a fake so Tipy would be at least slightly happier with it?

Marmio64:
buff the vampire coat instead so that its more worth going loud?  Also all fake antag clothing items have discernable differences from the real ones. Look at a fake wizard robe and real one, you can easily tell which is which, same for clockwork armor except even more so.

Tipy1802:
> that is a rules issue sec doesnt just go kill clowns wearing wizard robes that is a frankly ridiculous criticism,

Except sec can easily tell when someone is wearing fake wizard clothing with a glance

>your second criticism makes no sense this doesnt give them blood regen

I am not saying that the fake one gives blood regen, I am saying that having a fake but identical looking cloak would mean that sec won't go after vampires using the actual cloak out of fear of killing an ''innocent'' by getting fooled. And yes the real vampire cloak is supposed to be an obvious tell that you are a vampire.

code/modules/clothing/suits/miscellaneous.dm
```diff
@@ -648,3 +648,15 @@
 	w_class = WEIGHT_CLASS_TINY
 	flags_inv = HIDEGLOVES|HIDEEARS|HIDEFACE|HIDEHAIR|HIDEFACIALHAIR
 	alternate_worn_layer = UNDER_HEAD_LAYER
+
+/obj/item/clothing/suit/cosmeticvampirecoat
+	name = "Collectible Vampire Coat"
+	desc = "A Coat designed to mimic the coats of the vampires from Translyvania."
```
Hopekz, line 654:
```suggestion
	desc = "A Coat designed to mimic the coats of the vampires from Transylvania."
```

alexkar598 (changes requested):
dracula coat is already a thing

PR #11390, opened 2021-04-19
makes ling gloves not check for antimagic

SomeguyManperson:
tentacles aren't magical unless they are from japan and these aren't the japanese tentacles

:cl:  
bugfix: ling tentacle is no longer blocked by antimagic
/:cl:


athena413:
ok but can we add the japanese tentacles

Cartlord (approved):

ChesterTheCheesy:
> 
> 
> ok but can we add the japanese tentacles

look up my eldritch horror PR, that tentacle is as close as it gets

LazennG:
you buffoon you forgot to check with alexkar what real ling gloves are

PR #11409, opened 2021-04-25
godhand can ignite cigars and stuff too

Marmio64:
God hand can now ignite cigs, cigars, candles, all that good stuff. That's all.
Is there really a need for a template? No. Everything is explained in one sentence. 

:cl:  
rscadd: God hand can ignite cigars, candles, cigs, all that stuff
/:cl:


LazennG:
you fool you just got baited now i will do the same thing but with multiz and 5 dollars

PR #11410, opened 2021-04-25
lowers instability of antenna and removes it's access to command and sec channels

Marmio64:
lowers instability of antenna and removes it's access to command and sec channels

invalidates the privacy of sec and command comms way too easily, this fixes that

antenna doesn't let people access sec and command channels now
also it's instability is 10 instead of 15 now since it's most powerful feature of having sec and command comms was removed.

mutation/genetics changes

no?

:cl:  
rscdel: Removes antenna's access to the sec and command channels  
tweak: antenna mutation instability lowered to 10
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/encryptionkey.dm
+++ b/code/game/objects/items/devices/radio/encryptionkey.dm
@@ -131,5 +131,5 @@
 /obj/item/encryptionkey/secbot
 	channels = list(RADIO_CHANNEL_AI_PRIVATE = 1, RADIO_CHANNEL_SECURITY = 1)
 
-/obj/item/encryptionkey/antenna //for antenna mutation and it's unique demands. hopefully it wont be pulled out somehow.
+/obj/item/encryptionkey/antenna //for antenna mutation and it's unique demands. hopefully, it won't be pulled out somehow.
 	channels = list(RADIO_CHANNEL_ENGINEERING = 1, RADIO_CHANNEL_SCIENCE = 1, RADIO_CHANNEL_MEDICAL = 1, RADIO_CHANNEL_SUPPLY = 1, RADIO_CHANNEL_SERVICE = 1)
--- a/code/game/objects/items/devices/radio/encryptionkey.dm
+++ b/code/game/objects/items/devices/radio/encryptionkey.dm
@@ -130,3 +130,6 @@
 
 /obj/item/encryptionkey/secbot
 	channels = list(RADIO_CHANNEL_AI_PRIVATE = 1, RADIO_CHANNEL_SECURITY = 1)
+
+/obj/item/encryptionkey/antenna //for antenna mutation and it's unique demands. hopefully it wont be pulled out somehow.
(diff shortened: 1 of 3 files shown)


SomeguyManperson (approved):

athena413:
would rather just see it be for common only, since departments like engineering and others should still keep their privacy

MegaEmpirical:
I'd support just common but this is nice too.

Cartlord:
But if it only gives common access, what is the POINT of the mutation? What would one gain from it?

Marmio64:
> But if it only gives common access, what is the POINT of the mutation? What would one gain from it?

well you could use it with earmuffs to have flashbang protection while keeping a radio

Tipy1802:
> But if it only gives common access, what is the POINT of the mutation? What would one gain from it?

Crying for help over comms even when having your headset removed.
Thats how it used to work a few months ago before someone gave it all comms.
I am happy as long as it doesnt have sec or command comms, good pr

athena413:
> 
> 
> But if it only gives common access, what is the POINT of the mutation? What would one gain from it?

it also makes your comms be EMP immune

Cartlord:
Consider how ludicrously situational this is. Who's going to bother? And if its only real use is for if you get your headset stolen - which nobody but an antag or griefer would do - wouldn't the whole use of this count as powergaming?

SomeguyManperson:
actually yeah it should go back to just common since geneticists tend to be shitters when they have access to things they shouldn't

Marmio64:
> Consider how ludicrously situational this is. Who's going to bother? And if its only real use is for if you get your headset stolen - which nobody but an antag or griefer would do - wouldn't the whole use of this count as powergaming?

whole use right now already is just powergaming since you get sec and command access

code/game/objects/items/devices/radio/encryptionkey.dm
```diff
@@ -130,3 +130,6 @@
 
 /obj/item/encryptionkey/secbot
 	channels = list(RADIO_CHANNEL_AI_PRIVATE = 1, RADIO_CHANNEL_SECURITY = 1)
+
+/obj/item/encryptionkey/antenna //for antenna mutation and it's unique demands. hopefully it wont be pulled out somehow.
```
Hopekz:
```suggestion
/obj/item/encryptionkey/antenna //for antenna mutation and it's unique demands. hopefully, it won't be pulled out somehow.
```

UselessTheremin:
hopek "spellcheck" strikes again

code/game/objects/items/devices/radio/encryptionkey.dm
Marmio64: its a comment, nobody cares about grammar

Hopekz: > its a comment, nobody cares about grammar

I just read it and thought, "who the hell doesn't know their won't spellings" enough to want to check who the author is, so whether you like it or not, people care, and it DOES make you look bad.

There isn't anything wrong with looking illiterate, so it's ultimately up to you. I'm not forcing you to take the suggestion.

PR #11292, opened 2021-04-06
Polysmorps bleed green blood like xenos do, and species without blood don't bleed red

ChesterTheCheesy:
# General Documentation
BLOOD FOR THE BLOOD GOD

### Intent of your Pull Request
SWEET BLOOD, IT SINGS TO ME

###Explain what the purpose of your PR is. Mention if you tested your changes.
Makes polysmorphs bleed actual xeno blood, since their blood is already established to be acid. It can be scanned and stores the polysmorphs DNA. And yes, i tested it, obviously.

Additionally, species with NOBLOOD species trait, when being hit by projectile, "bleed" grey. Polys have a green effect now.

### Why is this change good for the game?
Makes roundstart race trackable by blood like humans and actual xenos are.

# Wiki Documentation
### Briefly describe your PR and the impacts of it, in layman's terms. 
CRAWLING IN MY SKIN

### What general grouping does this PR fall under? 
BLOOOOD, SO MUCH BLOOD

# Changelog
:cl:  
rscadd: Polysmorphs now bleed green like they always should have.
bugfix: species with no blood no longer bleed red when hit by ballistic projectiles
imageadd: added generic blood splatter
/:cl:

PR #11412, opened 2021-04-25
Revert "Polysmorps bleed green blood like xenos do, and species without blood don't bleed red"

Bibby0110:
Reverts yogstation13/Yogstation#11292

This is still hacky code so apparently I have to fix it myself.


Bibby0110:
The code is still bad but I’m sure you’ll manage

PR #11420, opened 2021-04-27
fixes escape pods probably

SomeguyManperson:
gives each pod a corresponding number to the docking ports they go to so they actually like land
:cl:  
bugfix: escape pods now work
/:cl:


patpol4 (approved):
Oh right i forgot about that

wejengin2:
well this didnt work

PR #11426, opened 2021-04-27
holdup now won't force trigger on the target moving

SomeguyManperson:
forgot about that little bit being there and not not there
:cl:  
bugfix: holdup no longer force triggers on movement, uses non-forced trigger
/:cl:


MegaEmpirical:
Isn't this the point of it?

SomeguyManperson:
was moved alongside other stuff when fixing them being able to spam disarms or whatever on the holdup initiator, shouldn't trigger while being pulled

PR #11439, opened 2021-04-30
Removes borg *Spin stunlock I guess

UselessTheremin:

this was complained about in dev public by Steamed, so I did a thing. *spin now has a 4 second cooldown, and the person is knocked down for 4.1 seconds instead of the previous 6 seconds. 

stunlock bad

:cl:  
tweak: borg spin stun now paralyzes for 4.1 seconds
tweak: *spin has a 4 second cooldown
/:cl:

### DIFF
--- a/code/datums/components/riding.dm
+++ b/code/datums/components/riding.dm
@@ -262,7 +262,7 @@
 /datum/component/riding/human/force_dismount(mob/living/user)
 	var/atom/movable/AM = parent
 	AM.unbuckle_mob(user)
-	user.Paralyze(60)
+	user.Paralyze(41)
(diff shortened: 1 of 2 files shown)


alexkar598:
Dont really like how it affects the spin for carbons too, kinda sad

UselessTheremin:
> Dont really like how it affects the spin for carbons too, kinda sad

why do you need to spin more than once in 4 seconds also im pretty sure the duration is about 4 seconds

Marmio64:
people like to be dumb and spin a bunch cause its fun. maybe solve the spin stunlock by instead making spin only knock you down rather than stun

PR #11443, opened 2021-05-03
gives sol dry minor toxin healing

boodaliboo:
irl ginger root is a exceptional for helping with nausea even in processed soda that effect is noticable so decided to give it minor toxin healing
:cl:  
rscadd: Added minor toxin healing to sol dry
/:cl:


SomeguyManperson:
already has an effect on disgust, more unique than being reskinned tea

PR #11447, opened 2021-05-04
probably fixes vote panel

SomeguyManperson:
moves all the voting stuff out of the admin only tab
# Changelog

:cl:  
bugfix: tgui vote panel now works
/:cl:


alexkar598 (approved):

tgui/packages/tgui/interfaces/Vote.js
```diff
@@ -117,6 +103,43 @@ const AdminPanel = (props, context) => {
   );
 };
 
+const VotePanel = (props, context) => {
```
alexkar598, line 106: Not a fan of using things before they're declared but w/e

PR #11461, opened 2021-05-10
[admin] fixes a mbug where mpeople get too much mknowledge

ynot01:
key_name was used instead of designated key_name_mentor meant specifically to be used for mentors


SomeguyManperson (approved):
gaming

PR #11464, opened 2021-05-12
Makes the SM surge event even more powerful

patpol4:
# General Documentation

After the latest changes it's still wet farts. What I did is raise the maximum random powerup to as high as 100 000.
This may seem like a lot compared to what we have now(8k) but from what I understand, the way this code works is just set it to the power value for a split second, making even values up to 10k insignificant on a default engine setup.
Made it a bit more rare to balance it out.

Tested it by getting a simple straight pipe setup and setting the SM power level to 100k via varediting. What happened was it just burned for a little while but didn't even start delaminating.

As it is right now it still does nothing and only fills up the events.

# Wiki Documentation

What I did is raise the maximum random powerup to as high as 100 000.
This may seem like a lot compared to what we have now(8k) but from what I understand, the way this code works is just set it to the power value for a split second, making even values up to 10k insignificant on a default engine setup.
Made it a bit more rare to balance it out.

The SM surge event is now a bit more rare but more powerful

Engineering

Not really

The earliest start for it is 10 minutes in and the SM crystal has to has been powered.

# Changelog

:cl:  
tweak: Makes the SM surge event even more powerful, lowers its chance to appear
/:cl:

### DIFF
--- a/code/modules/events/supermatter_surge.dm
+++ b/code/modules/events/supermatter_surge.dm
@@ -16,8 +16,7 @@
 	power = rand(1000,100000)
 
 /datum/round_event/supermatter_surge/announce()
-	if(power > 800 || prob(round(power/8)))
-		priority_announce("Class [round(power/500) + 1] supermatter surge detected. Intervention may be required.", "Anomaly Alert")
+	priority_announce("Class [round(power/500) + 1] supermatter surge detected. Intervention may be required.", "Anomaly Alert")
 
 /datum/round_event/supermatter_surge/start()
 	GLOB.main_supermatter_engine.matter_power += power
--- a/code/modules/events/supermatter_surge.dm
+++ b/code/modules/events/supermatter_surge.dm
@@ -1,7 +1,7 @@
 /datum/round_event_control/supermatter_surge
 	name = "Supermatter Surge"
 	typepath = /datum/round_event/supermatter_surge
-	weight = 20
+	weight = 15
 	max_occurrences = 4
 	earliest_start = 10 MINUTES
 
@@ -13,7 +13,7 @@
 	var/power = 2000
 
 /datum/round_event/supermatter_surge/setup()
-	power = rand(400,8000)
+	power = rand(1000,50000)
 
 /datum/round_event/supermatter_surge/announce()
 	if(power > 800 || prob(round(power/8)))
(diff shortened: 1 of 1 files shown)


patpol4:
Would be nice to at least get this testmerged to see if I'm not a giant idiot and got this all wrong.

Jarod1200:
More power = good

patpol4:
Ok it turns out I am, the values should still be fine though.

code/modules/events/supermatter_surge.dm
```diff
@@ -1,7 +1,7 @@
 /datum/round_event_control/supermatter_surge
 	name = "Supermatter Surge"
 	typepath = /datum/round_event/supermatter_surge
-	weight = 20
+	weight = 15
```
alexkar598, line 4: Undocumented change.

alexkar598 (changes requested):

code/modules/events/supermatter_surge.dm
```diff
@@ -13,7 +13,7 @@
 	var/power = 2000
 
 /datum/round_event/supermatter_surge/setup()
-	power = rand(400,8000)
+	power = rand(1000,100000)
```
alexkar598, line 16: Thats a very wide range, this will lead to inconsistent gameplay.

code/modules/events/supermatter_surge.dm
```diff
@@ -13,7 +13,7 @@
...
 
 /datum/round_event/supermatter_surge/announce()
 	if(power > 800 || prob(round(power/8)))
```
alexkar598: 1) Both these conditions are redundant as the power will never be under 800 and the prob will never fail as the minimum is 1000/8 so prob(125).
2) The class number can be ridiculously high now. Anywhere from class 3 to class 201

alexkar598 (changes requested):

code/modules/events/supermatter_surge.dm
patpol4: its in the description but i did the changelog too now

code/modules/events/supermatter_surge.dm
patpol4: Yes and the severity is going to be stated in the announcement so it's not like it's secret and people don't realize how bad it is.
Do you think i should raise the minimum?

code/modules/events/supermatter_surge.dm
patpol4: 1. removed the condition so it always announces,
2. yes

MegaEmpirical:
Yessssssssss

PR #11466, opened 2021-05-13
this is why we can't have nice things

SomeguyManperson:
:cl:  
bugfix: you can only have TWO borg expander modules per borg. You know what you did.
/:cl:

### DIFF
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -706,10 +706,10 @@
 /obj/item/borg/upgrade/expand/deactivate(mob/living/silicon/robot/R, user = usr)
 	. = ..()
 	if (.)
-		if (R.hasExpanded)
-			R.hasExpanded = FALSE
-			R.resize = 0.5
-			R.update_transform()
+		while(expansion_count)
+			resize = 0.5
+			expansion_count--
+			update_transform()
 
 /obj/item/borg/upgrade/rped
 	name = "engineering cyborg RPED"
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -1,5 +1,7 @@
 // robot_upgrades.dm
 // Contains various borg upgrades.
+#define EXPANDER_MAXIMUM_STACK 2
+
 
 /obj/item/borg/upgrade
 	name = "borg upgrade module."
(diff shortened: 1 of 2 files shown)


ZombieNub:
I know what I did but in my defense the borg wanted it and how was I gonna say no?

ToasterBiome:
delete this

N3D6:
theos once again saves yog from fun

ZombieNub:
In my opinion expansion stacking should remain. It's funny and I'll let others have fun with it if they so please.

SomeguyManperson:
maybe stacking up to a limit but you can get so big right now you can obscure like the whole screen

ToasterBiome:
> maybe stacking up to a limit but you can get so big right now you can obscure like the whole screen

I would say like 2-3 would be good for "meme" but also not game breakingly annoying

TheLinuxOS:
I love how you say “theos saves yogs from fun” when people were literally exploiting it to be as big as the escape shuttle

alexkar598 (changes requested):
1) Changelog is incorrect, its twice not once
2) Changelog is incorrect, it would be a tweak not a bugfix at this point since the intended behaviour is once not twice
3) `hasExpanded` needs to be changed from being a boolean to being a number at the following lines:
    - https://github.com/yogstation13/Yogstation/blob/master/code/game/objects/items/robot/robot_upgrades.dm#L708 (R.hasExpanded--)
    - https://github.com/yogstation13/Yogstation/blob/master/code/modules/mob/living/silicon/robot/robot.dm#L87 (`var/hasExpanded = 0`)
    - https://github.com/yogstation13/Yogstation/blob/master/code/modules/mob/living/silicon/robot/robot.dm#L1027 (`hasExpanded--`)
4) Ideally, `hasExpanded` should be renamed to something else so that its less confusing since its no longer a boolean. Perhaps `expansions`?

code/game/objects/items/robot/robot_upgrades.dm
```diff
@@ -706,9 +706,9 @@
 /obj/item/borg/upgrade/expand/deactivate(mob/living/silicon/robot/R, user = usr)
 	. = ..()
 	if (.)
-		if (R.hasExpanded)
-			R.hasExpanded = FALSE
+		while(R.expansion_count)
 			R.resize = 0.5
+			R.expansion_count--
 			R.update_transform()
```
alexkar598, line 712: I'm not 100% a fan of using a while loop here as its technically not needed but I suppose it prevents bugs from turning borgs invisible by rounding their size down to 0

alexkar598 (approved):

PR #11278, opened 2021-04-03
Nerfs tesla engine by making miniballs only act 20% of the time, also makes tesla bolts no longer bounce indefinitely

ChesterTheCheesy:
oh shit, here we go again
# General Documentation
### Intent of your Pull Request
Nerfs the tesla to not create as much ridiculous amounts of power as it does, since each miniball zap has half the charge of tesla itself. By simple math, currently it means tesla makes 1(main tesla) + 5 (10 miniballs) = 6 tesla power. With change, it'll make 1 + 5/5 = 2, theoritically reducing the power output of tesla engines 3 times. It also means, that once the tesla goes loose, instead of 11 explosions each tick, it makes around 3, hopefully being enough for the server to catch up, reducing the lag.

Tesla bolts, now upon hitting a structure, bounce twice, if the shock came from tesla, or once, if it came from miniball. No more infinite loops because whole station is made of conductive metal.

### Why is this change good for the game?
nerfs tesla and makes it lag less during tesloose

# Wiki Documentation
### Briefly describe your PR and the impacts of it, in layman's terms. 
miniballs now act only 20% of time, tesla bolts no longer bounce between objects indefinitely

### What general grouping does this PR fall under? 
Engineering tweak

### If there are any numerical values involved in your PR that will be relevant to a player, please note them here. 
miniballs now act only 20% of time, meaning power output should be 3 times less for tesla

# Changelog
:cl:  
bugfix: tesla shocks no longer bounce between structures indefinitely
tweak: nerfs tesla by making the miniballs only act 20% of the time
/:cl:

PR #11467, opened 2021-05-13
Fixes unintentional nerf to Tesla

Asdjky:
# General Documentation

Fixes an issue introduced by #11278. There was a `return` where there should've been a `continue`.

It makes the miniballs not be super bad (but they're still pretty garbage).

# Wiki Documentation

Each miniball now independently now has a 20% chance to zap instead of all collectively only having a 20% chance to zap.

Tesla miniballs will generate a tiny bit more power now that they can do work more often.

Engineering changes.

No

Tesla as a whole generates a tiny bit more power.

# Changelog

:cl:  
bugfix: Telsa miniballs now properly trigger 20% of the time.
/:cl:

### DIFF
--- a/code/modules/power/tesla/energy_ball.dm
+++ b/code/modules/power/tesla/energy_ball.dm
@@ -71,9 +71,9 @@
 
 		for (var/ball in orbiting_balls)
 			if(prob(80))  //tesla nerf/reducing lag, each miniball now has only 20% to trigger the zap
+				continue
 			var/range = rand(1, clamp(orbiting_balls.len, 3, zap_range))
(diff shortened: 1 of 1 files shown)


alexkar598 (changes requested):
Invalid code, the if statement does not have a clear "then" block. Perhaps you intended to have the continue be where the return was?

Asdjky:
Wow yea I definitely messed that up. I tried hunk editing for the first time because I had other diffs in the file and I have no idea how I didn't notice that was in the wrong spot.

alexkar598 (approved):

PR #11469, opened 2021-05-13
Revert "probably fixes vote panel"

Bibby0110:
Reverts yogstation13/Yogstation#11447

literally no reason to diverge from TG over this. time to report the original PR


alexkar598:
Revert the PR when you port it. Absolutly no need to revert this without a solution in place.

PR #11473, opened 2021-05-14
Escape Rework

alao3alao3:

Reworks escape by adding chairs, and extends it out a little so that the chairs look better. This minorly extends the Holding Center.
bpng
hell i was fine with the current bland escape nature which is just empty

## Why is this change good for the game?

This is to serve as an alternative to EIB0N's rework which uses the fake multi-z illusion using shadows and stairs. I'm personally still on the fence about it's use in an area as heavily trafficked as escape.

## Wiki Documentation

Wiki image will need to be updated.

# Changelog

:cl:  
tweak: reworks escape, making it and the Holding Center bigger. 
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -6987,6 +6987,21 @@
 	},
 /turf/open/floor/plating,
 /area/maintenance/starboard/aft)
+"arH" = (
+/obj/machinery/atmospherics/pipe/manifold/scrubbers/hidden/layer4{
+	dir = 8
+	},
+/obj/machinery/atmospherics/pipe/manifold/supply/hidden/layer2{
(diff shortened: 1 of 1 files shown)


athena413:
no dont make departures be good

noooooooooooooooooooooooooooooooo

Tipy1802:
EIB0N's version is better

Marmio64:
bigger escape helps with highpop and lets you build bigger projects there, which is arguably more important than chairs/decals. Eibon's PR looks very cool but its very restrictive and leans heavily into a easily broken illusion

patpol4 (changes requested):
Fix conflict or just make another pr. 
Going with this one because Eibon's was too cluttered.

alao3alao3:
See #11493

PR #10970, opened 2021-02-06
Cloning now requires meat.

SergeyMeyer:
# General Documentation

### Intent of your Pull Request

Cloning now requires meat. Prior to this PR cloning is a medical win button. You take head, put head in scanner, out pops newly made body. This PR is a step in the right direction. Cloning is no longer just press button out comes body. You'll now need to aquire some sort of meat, either by a chemist making synthmeat, or a chef.

### Why is this change good for the game?
 
Cloning is no longer the ultimate medical winbutton. Regardless of how little cloning is actually used, it's ridiculously overpowered and makes death a joke. Death is still mostly a joke with this PR, but it's no longer "take dead body put in magical body making machine." I've made cloning harder to do by requiring you to put meat into the machine if you want a fresh body.

This complicates the medical system even a slight bit more, which is good to the game. Medical requiring even the slightest bit more buttons pressed is better in my opinion. This doesn't make cloning useless, its still a very viable thing, but now it requires more work. 

I've also made synthmeat an decently viable source for resources to help with cloning, only three of it being required to produce a body. 

Everything is tested, I'm a slightly competent coder.
# Wiki Documentation

### Briefly describe your PR and the impacts of it, in layman's terms. 
Cloning requires meat. It's % based. 
### What should players be aware of when it comes to the changes your PR is implementing?
Doctors will be needing to do more effort when they bring bodies to cloning, and a good chemist will make a decent amount of synthmeat. You'll need to actually supply cloning with meat, and leaving plenty somewhere in the room is probably a good idea.
(If this is testmerged something should be put in the wiki or around the roundstart "You are the Medical Doctor" thing, like that line that sends if you're sec.)
### What general grouping does this PR fall under? 
Medical
Cloning changes
### Are there any aspects of the PR that you would like us not to mention on the Wiki?
nah
### If there are any numerical values involved in your PR that will be relevant to a player, please note them here. 
Need 100 total "biomass" to make a single clone.
Human Flesh provides 50 Biomass
Humanoid Meat (Anything that's included in meat/slab/human/, which appears to be all humanoids.) provides 50
Synthmeat provides 34
Monkey Meat provides 25
Anything else in meat/slab is 20

# Changelog

:cl:  
rscadd: Cloning requires meat.
/:cl:

PR #11474, opened 2021-05-14
Revert "Cloning now requires meat."

boodaliboo:
Reverts yogstation13/Yogstation#10970
no one notices it requires biomass and theres no indicator besides a examine that it even uses it, its been a issue i have continually noticed and even then its just tedium to make it

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -25,17 +25,6 @@ GLOBAL_VAR_INIT(clones, 0)
 	var/attempting = FALSE //One clone attempt at a time thanks
 	var/speed_coeff
 	var/efficiency
-	// the three variables that handle meatcloning
-	var/biomass = 0 //Start with no biomass inserted.
-	///List of special meat that gives extra/less biomass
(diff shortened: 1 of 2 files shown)


UselessTheremin:
template

boodaliboo:
> template

no template shows up when you do a revert you should know this

Cartlord:
I've noticed myself that this doesn't even work all the time anyways, I've seen a cloner work with -366% biomass.

boodaliboo:
> I've noticed myself that this doesn't even work all the time anyways, I've seen a cloner work with -366% biomass.

its also just stupid if we had the proposed cloning system that the guy that started with a s and had a long name was working lon

alexkar598 (changes requested):
Templateeee

burn1cycle:
Believe it or not this isn't the first time someone tried "cloning requires meat" you just don't remember because it was so long ago and we removed the requirement for being retarded. I would be over-joyed to see it removed AGAIN for being retarded AGAIN and finally have it known once and for all that this is a retarded feature and to stop implementing it.

Issue #11399, opened 2021-04-21
Darkspawn mindlink chat prefix description is wrong


Yogbot-13:
## Test Merges: 
Not Supplied / None
## Reproduction:
darkspawn mindlink chat is prefixed with .w but the description in game tells you to use .k.
 Submitted by: The Grand Inquisitor (250794357211398144)

Cartlord:
Can confirm

adamsong:
Fixed by #11490

PR #11490, opened 2021-05-18
veils get right darpspawm prefix

SomeguyManperson:
dmin related. Prefix the PR title with [s] if you are fixing an exploit so it is not announced on discord and the server.

:cl:  
bugfix: veils get right darkspawn prefix
/:cl:


Cartlord:
You forgot to mention that this closes https://github.com/yogstation13/Yogstation/issues/11399

PR #11492, opened 2021-05-18
Buffs stormtrooper armor

Bibby0110:
# General Documentation

3x for 10k is expensive for something isn't the best for most of what you will encounter.

Gives it a buff so it's more in line with regular armor which is much much cheaper, nerfs the actual good laser resistance so it's not *that* much better

# Changelog

:cl:  
tweak: stormtrooper armor is better rounded
/:cl:

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -440,7 +440,7 @@
 	desc = "Battle Helmet from a long lost empire"
 	icon_state = "stormtrooperhelmet"
 	item_state = "stormtrooperhelmet"
-	armor = list("melee" = 30, "bullet" = 20, "laser" = 40,"energy" = 30, "bomb" = 25, "bio" = 40, "rad" = 20, "fire" = 50, "acid" = 50)
+	armor = list("melee" = 35, "bullet" = 25, "laser" = 35,"energy" = 20, "bomb" = 25, "bio" = 40, "rad" = 20, "fire" = 40, "acid" = 50)
 	flags_inv = HIDEEARS|HIDEHAIR
 	cold_protection = HEAD
 	min_cold_protection_temperature = HELMET_MIN_TEMP_PROTECT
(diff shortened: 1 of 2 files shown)


Hopekz:
If this is related to the star wars armor, it is weird that it has worse laser resistance since it primarily defends against lasers in the source material.

Bibby0110:
> If this is related to the star wars armor, it is weird that it has worse laser resistance since it primarily defends against lasers in the source material.

Very well aware, but staying close to the reference material isn't really wise considering practically no antags use lasers.

At current it's a bit of a newbie trap, since you'd expect 10k armor to be better in most cases than the regular cheap kind.

It's still not better than it, but it's not as garbage.
Wouldn't want it to be the natural upgrade imo.

Still has great laser resistance, just isn't paper mache when you get hit by some naked assistant.

Marmio64:
kinda defeats the purpose of it being anti-laser armor. Don't think of it like regular armor, think of it like armory armor, where it protects against a specific weapon but not much else. The ablative vest is only good against energy/laser weapons yet no antags use those, so now it should have its stats changed? If you want an armor that covers the whole body, is very protective, and slows you down, theres already the swat armor you can order from cargo. If you're looking to make it not be a noobie trap, change the desc so that it says its good against lasers and not much else, and while you're at it, reduce the slowdown to be like 0.75 since the slowdown makes the armor pretty damn trash.

Bibby0110:
> kinda defeats the purpose of it being anti-laser armor. Don't think of it like regular armor, think of it like armory armor, where it protects against a specific weapon but not much else. The ablative vest is only good against energy/laser weapons yet no antags use those, so now it should have its stats changed? If you want an armor that covers the whole body, is very protective, and slows you down, theres already the swat armor you can order from cargo. If you're looking to make it not be a noobie trap, change the desc so that it says its good against lasers and not much else, and while you're at it, reduce the slowdown to be like 0.75 since the slowdown makes the armor pretty damn trash.


If anything this should be buffed without having it's laser armor reduced

As it stands this armor is on par with a reflective jacket when it comes to laser armor, at more than 3 times the cost per armor.

For 2k+ extra credits you get:
Slowdown
10 more melee + bullet armor

As it stands the reflective vest is just so much better, less slowdown same stats bar 10 melee + bullet armor.

This way it is at least semi-reasonable as a heavy armor weighted towards lasers.

Marmio64:
removes its niche as a fullbody 'ablative vest' but gives it a different niche i guess

PR #11493, opened 2021-05-18
Escape Rework (redone due to conflict)

alao3alao3:
# General Documentation

This pr is identical to #11473 and documentation can be found there. 

# Changelog

:cl:  
tweak: reworks escape, making it and the Holding Center bigger.  
/:cl:


patpol4 (approved):
Also not really a rework but this is wider and good

PR #11513, opened 2021-05-22
strip from contractor wont strip implants

SomeguyManperson:
clueless
:cl:  
bugfix: contractor pod will not strip implants
/:cl:


adamsong:
Would this allow someone to setup a teleporter to brazil with a tracking implant?

TheLinuxOS:
I’m fairly certain brazil is centcom a level so it shouldn’t

PR #11515, opened 2021-05-22
Removes passive income from the Security Budget

Bibby0110:
# General Documentation

There is no good way to balance around their budget. Due to the Economy subsystem not respecting the universal laws of time it will be permanently out of wack. This is better than playing whack-a-mole with 100k sec budgets 30 minutes into the game.

They now start with 10k, you can deposit and withdraw but no money is added on the Economy tick

# Wiki Documentation

Sec budget gets 10k roundstart and no more

Cargo rebalancing

0

# Changelog

:cl:  
rscdel: The Security budget now only gets a flat sum of money each shift.
/:cl:


adamsong:
Is the budget card not required for the warden's paystand?

Bibby0110:
> Is the budget card not required for the warden's paystand?

Good point, I'll replace it

Marmio64:
sec deserve some kind of budget though. How about a card that just has like 10k on it or something, or even just some plain old spacecash in the card's place. Removing the budget entirely would just make sec's lives way harder than it has to be, and it can already be a nightmare to try to survive against a cult or something so long as that cult has at least one good player.

Bibby0110:
> sec deserve some kind of budget though. How about a card that just has like 10k on it or something, or even just some plain old spacecash in the card's place. Removing the budget entirely would just make sec's lives way harder than it has to be, and it can already be a nightmare to try to survive against a cult or something so long as that cult has at least one good player.

That doesn’t really have much to do with the budget but more to do with the skill of security, but you’re welcome to code in an alternative yourself. This is just horribly broken as it stands and I don’t personally see it needing a replacement, but as I said you’re welcome to PR your own alternative :)

Bibby0110:
> sec deserve some kind of budget though. How about a card that just has like 10k on it or something, or even just some plain old spacecash in the card's place. Removing the budget entirely would just make sec's lives way harder than it has to be, and it can already be a nightmare to try to survive against a cult or something so long as that cult has at least one good player.

On further thought you're right, it's been locked to 10k (2x regular starting amount).

Bibby0110:
I'm going to do an epic drift and re-open this PR since I changed it to be something totally different 😄

PR #11529, opened 2021-05-28
[Admin] People who are non consentually put into cryo no longer get offered to ghosts automatically and prevents using cryo as an indestructible safe spot by using the offer to ghost prompt to delay the cryo forever

alexkar598:
Let's end this stupid shit of admins bwoinking people for making use of cryo pods as they should be used because of a bug

Technical details:
Alert will return the first option automatically if the mob does not have a client set, this made people autoconsent to ghost, which was a bug
Also moved the ghost prompt down so it starts the despawn timer before asking people if thye wanna take control but that despawn timer is cancelled if the person says yes

:cl:  
bugfix: You can no longer offer people to ghosts by cryoing them forcibly, players who cryo themselves are still able to offer themselves
bugfix: You can no longer delay the cryo despawn timer indefinetly by keeping the offer to ghost prompt open indefinetly
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -170,7 +170,8 @@ GLOBAL_LIST_INIT(typecache_cryoitems, typecacheof(list(
 	var/on_store_name = "Cryogenic Oversight"
 
 	// 5 minutes-ish safe period before being despawned.
-	var/time_till_despawn = 5 MINUTES // This is reduced to 30 seconds if a player manually enters cryo
+	var/time_till_despawn = 15 MINUTES // Time if a player gets forced into cryo
+	var/time_till_despawn_online = 30 SECONDS // Time if a player manually enters cryo
 
 	var/obj/machinery/computer/cryopod/control_computer
 	var/cooldown = FALSE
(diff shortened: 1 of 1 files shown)


alexkar598:
Actually lemme reopen this as an admin pr

alexkar598:
Also turdis is failing but thats because of dependency funny

Bibby0110:
> Also turdis is failing but thats because of dependency funny

funny way to say "my code is broken"

adamsong:
Will this still offer those with no ckey?

MegaEmpirical:
So will bodies still be deleted? What's the timer?

alexkar598:
> 
> Will this still offer those with no ckey?

Only those who are online will get the prompt, if you are not online, you do not get offered, if you are online, you get to choose. So no.



> 
> So will bodies still be deleted? What's the timer?

5 minutes if you're offline when youre put in cryo, 30 seconds otherwise.

MegaEmpirical:
I would rather it be made so that no player can cryo another

Firewolf34:
> 
> 
> I would rather it be made so that no player can cryo another

Why is that? The purpose of cryo is to get SSD people out of the halls... random people going SSD instead of going to dorms (if they want to stay in the game) or going to cryo (if they wish to leave) is a nuisance - so we carry them to cryo so the system can deal with them. If anything, we should have this happen automatically somehow. Maybe get a borg to do it.

patpol4:
This offers me even when I press no I dont want to offer myself, I just want to cryo

patpol4:
https://github.com/yogstation13/Yogstation/issues/11561

PR #11553, opened 2021-06-06
Removes tildes from filenames

monster860:
# General Documentation

This renames the `~yogs_defines` folder to `{yogs_defines}`. The purpose of the tilde is to be after everything else when sorted. However, the tilde has problems in certain contexts. Specifically, tildes break local workspace mode in FastDMM2 ever since chrome decided to block tildes the File System Access API. Since it seems that this change will stay, this PR replaces the tilde with curly braces. Curly braces have the same property of being sorted after other files, and it should work fine since we already feature curly braces in filenames thanks to the use of GUIDs which have them, being part of some filenames.

TL;DR makes it work properly in fastdmm2 again


alexkar598 (approved):

Bibby0110:
hot PR

PR #11587, opened 2021-06-13
Enables map rotation

patpol4:
# General Documentation

Enables map rotation, how it works is explained below:
image
then it rolls a random map that fulfills the pop reqs.

Enables some map variation, if someone really really doesn't want it, they can always do a map vote to change it back to box.
So I wouldn't say this is forcing anything

# Wiki Documentation
does not apply, config change

# Changelog
:cl:  
tweak: enables map rotation
/:cl:


Bibby0110:
good change

JamieD1 (approved):

PR #11590, opened 2021-06-13
Succumb Changes

JamieD1:
Succumb is now on a 90 second delay.

Argue with Lyis, not me.

Im just a code monkey.


MegaEmpirical:
The way this should work is: Initiating the command will start a timer, that can only be started while in crit. This will bring a confirmation prompt and after the timer has elapsed they will then die if they're in crit still. If they're healed but then drop back into crit before the timer has elapsed and then it elapses while they're in crit, they die. 

Using the verb will also alert those around them that they're giving up on life .

Moggram:
Seem like you would die most times long before 90 sec. Why not have it do more OxyLoss?

MegaEmpirical:
1) It can take a lot longer to die in soft crit
2) Time it takes takes for you to die isn't a factor for this change beyond it possibly taking a long time.

This change is to reduce abuse of the succumb mechanic and to provide an out for anyone stuck dying slowly in crit.

Moggram:
> 
> 
>     1. It can take a lot longer to die in soft crit
> 
>     2. Time it takes takes for you to die isn't a factor for this change beyond it possibly taking a long time.
> 
> 
> This change is to reduce abuse of the succumb mechanic and to provide an out for anyone stuck dying slowly in crit.

Ooohh ok then.

alexkar598:
God no, I don't want to sit at a screen for 1.5 minutes, if this goes though, il just ghost instead

Hopekz:
> God no, I don't want to sit at a screen for 1.5 minutes, if this goes though, il just ghost instead

This ^

If someone succumbs it is because they don't want to control their character anymore otherwise they would just do nothing and wait

Cartlord:
I succumb solely because either A) I don't want to wait when death is inevitable or B) I _do not want_ whatever's about to happen to my character. If I Succumb while a cultie's dragging me to a rune, it's because I DO NOT WANT TO BE CONVERTED. This is exceedingly lame, and really seems to defeat the purpose of Succumbing.

MegaEmpirical:
> I succumb solely because either A) I don't want to wait when death is inevitable or B) I _do not want_ whatever's about to happen to my character. If I Succumb while a cultie's dragging me to a rune, it's because I DO NOT WANT TO BE CONVERTED. This is exceedingly lame, and really seems to defeat the purpose of Succumbing.

This is why we're making this change.

Cartlord:
To make Succumb useless?

Hopekz:
> To make Succumb useless?

 probably. 

The PR also seems to delay the whisper of the succomb so now you can't even really time your whisper so someone hears you. That and who knows if the information is even valid 90 seconds after you meant to say it.

If this gets implemented I'd rather see succomb removed tbh instead of being in this state

Cartlord:
I'm honestly with Alexkar, if this gets merged I'll probably just Ghost instead of dealing with this nonsense.

boodaliboo:
what is succomb i have ghosted every time

MegaEmpirical:
> what is succomb i have ghosted every time

It's a command that is similar to suicide. It would let you die instantly if you were in soft or hard crit but it would still let you be revived. Useful if you were dying slowly. This was being abused by people about to be converted.

Marmio64:
this should be 30 seconds at the most, you're going to be killed/converted before you even hit the 15 second mark for most cases

code/modules/mob/living/living.dm
```diff
@@ -1375,16 +1385,15 @@
...
-        /datum/antagonist/rev,
-        /datum/antagonist/shadowling,
-        /datum/antagonist/veil
-    )
-    for(var/antagcheck in bad_antags)
-        if(mind?.has_antag_datum(antagcheck))
-            return TRUE
-    return FALSE
-	
+	var/list/bad_antags = list(
```
adamsong, line 1388: Forgot thralls

adamsong (changes requested):

Mqiib:
> this should be 30 seconds at the most, you're going to be killed/converted before you even hit the 15 second mark for most cases

30 seconds is much more reasonable, yes.

code/modules/mob/living/living.dm
```diff
@@ -1375,16 +1385,15 @@
...
+		/datum/antagonist/clockcult,
+		/datum/antagonist/cult,
+		/datum/antagonist/darkspawn,
+		/datum/antagonist/rev,
+		/datum/antagonist/shadowling,
+		/datum/antagonist/veil
+	)
```
JamieD1:
```suggestion
		/datum/antagonist/shadowling,
		/datum/antagonist/veil,
		/datum/antagonist/thrall
	)
```

num1bamf:
This is an amazing change. Let’s try it. Council bamf approved.

adamsong (approved):
Code looks functional, even if the change is bad

MegaEmpirical:
> Code looks functional, even if the change is bad

How so?

adamsong:
90 seconds is way to long, significantly longer than most conversions, the only outlier being thralling a mind-shielded person, and that still takes only 65 seconds. This should either be reduced to 30 seconds, or the better solution would be to restrict succumbing to hard crit. This would allow for crit to still be used as a mean of subduing your opponent, but means that someone doing so must be careful to keep the victim out of hard crit.

MegaEmpirical:
> 90 seconds is way to long, significantly longer than most conversions, the only outlier being thralling a mind-shielded person, and that still takes only 65 seconds. This should either be reduced to 30 seconds, or the better solution would be to restrict succumbing to hard crit. This would allow for crit to still be used as a mean of subduing your opponent, but means that someone doing so must be careful to keep the victim out of hard crit.

Length of conversion is not a factor here, it's the time leading up to it and the fact that succumb is currently instant.

90 seconds is Jamie's suggestion not mine. I suggested about a minute but I am willing to try 90 and adjust going forward. 

Dying in hard crit is significantly faster than soft crit, and can be accelerated by talking. 

Succumb is for when you're taking too long to die, and soft crit can be extremely slow.  I have crawled from the engine to medical in soft crit more than once.

I think 60 would be fine but I won't know until we've tested it. If your issue is the 90, that was Jamie's call.

I want sufficient time for the person handling the victim to be able to respond and do what is needed. 30 seconds would be plenty in ideal conditions, we're not always working in ideal conditions so I want some wiggle room.

adamsong:
Its called 'critical' for a reason, its not a state you should intend to keep someone in, because someone in that state is on the verge of dying. I still think no more than 30s, and restricting to hard critical is a better course of action

MegaEmpirical:
> Its called 'critical' for a reason, its not a state you should intend to keep someone in, because someone in that state is on the verge of dying. I still think no more than 30s, and restricting to hard critical is a better course of action

Doing this would remove the useful part of succumbing, not dying slowly in soft crit. That's where succumb actually matters.

Moja10:
bitch PR

N3D6:
this is godawful

Cartlord:
always remember that 'mins pushing bad changes brought down /tg/ and it _will_ bring down yog if/when it happens

num1bamf:
Y’all are very dramatic

N3D6:
> Y’all are very dramatic

imo i think its just a bad change, just don't like the idea of this. when i succumb, i want to die, not wait a minute and a half for my slow death.

Junkyshell:
why is this testmerged this shit has 3 people in favor of it and 14 opposed. let's be honest this is just another example of the ebil and vile lyis trolling the community.

num1bamf:
I can ask for succumbing to be removed entirely if that’ll be better for you

MegaEmpirical:
https://user-images.githubusercontent.com/36208065/125208698-1c09aa80-e262-11eb-85e2-ea9f4704a6d2.mp4

Hopekz:
> I can ask for succumbing to be removed entirely if that’ll be better for you

Not sure if this is meant to be a joke or not but I'd personally be in favor of a full removal vs changing it to this if those are the only two options that exist

Cartlord:
IMO
Sometimes I wanna play Bloodcult, and sometimes I just don't
Sometimes I go into a round wanting to do really specific shit, and being converted can keep me from doing that
I don't want a bloodcult ban, because I do have fun with the role sometimes - like if I just arrived so they weren't interrupting much, or if I'm a Botanist who can mass-produce unholy water - and sometimes, I wanna show up and do weird Turbine shit, and being converted means I don't get to do Turbine shit and instead I need to do cult shit
So that's why I do succumb to avoid being converted, because sometimes I'm in the mood to do converson antag things and sometimes I am simply _not_.
Would it be better to be converted and then AHelp to be offered to Ghosts?
(Posted this in the Discord, it got buried/ignored, so posting here)

maxion12345:
i fucking hate this change man. Getting hardcrit and having to wait 90 seconds (especially with lag) is just infuriating. Especially when you can use this to whisper numerous times in hardcrit to circumvent the "talking in hardcrit = your last words"

Cartlord:
I understand, after someone explained to me in Discord, that it's better to be converted and ask a 'min to Ghost you than it is to Succumb to dodge conversion.
That being said, people really seem to HATE this PR, and I don't think it's right to Merge it. Please don't force something that _people don't want_. As for where I'm getting my info, it's stuff I've seen in OOC chat.
I'm genuinely afraid of the path we'll be heading down, if 'mins start trying to force PRs that people don't want - that's what fucked /tg/ and it'll fuck us too.

SomeguyManperson:
just ban people for succumbing to draft dodge :^

code/modules/mob/living/living.dm
```diff
@@ -380,10 +380,20 @@
 /mob/living/verb/succumb(whispered as null)
 	set hidden = TRUE
 	if (InCritical())
-		log_message("Has [whispered ? "whispered his final words" : "succumbed to death"] while in [InFullCritical() ? "hard":"soft"] critical with [round(health, 0.1)] points of health!", LOG_ATTACK)
+		if(alert("Are you sure you want to Succumb? (This will take 90 Seconds)",, "Yes", "No") != "Yes")
```
Bibby0110, line 383:
```suggestion
		if(alert("Are you sure you want to succumb? (This will take 90 seconds)",, "Yes", "No") != "Yes")
```

code/modules/mob/living/living.dm
```diff
@@ -380,10 +380,20 @@
...
+			return
+		to_chat(src, "<span class='notice'>You have started to loosen the grip on this world you will die in about 90 seconds.</span>")
```
Bibby0110, line 385:
```suggestion
		to_chat(src, "<span class='notice'>You have started to loosen the grip on this world and you will die in about 90 seconds.</span>")
```

code/modules/mob/living/living.dm
```diff
@@ -1375,16 +1385,16 @@
...
+		/datum/antagonist/darkspawn,
+		/datum/antagonist/rev,
+		/datum/antagonist/shadowling,
+		/datum/antagonist/veil,
+		/datum/antagonist/thrall
+	)
+	for(var/antagcheck in bad_antags)
+		if(mind?.has_antag_datum(antagcheck))
+			return TRUE
+		return FALSE
```
Bibby0110, line 1400: Untab once

SomeguyManperson:
suggest one of two alternatives (or both i guess)
A: make certain antag stuns prevent succumbing for a decent timeframe, stuff like shadowling glare or bloodcult stun or whatever, like if you get hit by one you are prevented from succumbing for a few minutes outright otherwise succumbing works normally
B: ban people for succumbing to dodge conversion

JamieD1:
Yeah.

PR #11591, opened 2021-06-13
Job Exp Changes for Lyis

JamieD1:
Atmos to 3 hours as Crew
Chemist to 2 hours as Medical
RD to 3 hours as Science
Scientist to 3 hours as crew
Engineer to 3 hours as crew
Virologist to 2 hours as medical

### DIFF
--- a/code/modules/jobs/job_types/atmospheric_technician.dm
+++ b/code/modules/jobs/job_types/atmospheric_technician.dm
@@ -8,7 +8,7 @@
 	spawn_positions = 2
 	supervisors = "the chief engineer"
 	selection_color = "#fff5cc"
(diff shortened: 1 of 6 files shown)


patpol4:
Engi should be lower then atmos but yeah

Jarod1200:
IMO atmos should be 3 hours as engi (and I think heads of staff should be 10+ hours)  but good step in the right direction. Remeber 3 hours is literally 2 rounds

MegaEmpirical:
The goal here is to make it harder for new accounts to grief by immediately having access to atmos and such.

I'll likely revisit the hours across the board another day. I am definitely up for increasing head hours

PR #11610, opened 2021-06-16
Enable support for 514.1556

alexkar598:
No major known issues with the new build. Also removed the 513.1537 check as 514 is the minimum version now


Sinestia:
Huh. Do your smooth doors work properly on the 514 client?

JamieD1 (approved):

PR #11611, opened 2021-06-17
Starts the deprecation cycle for omegastation in favour of kilostation

alexkar598:
It makes absolutely no sense to maintain a shitton of maps because we can barely maintain our current set. Also the bounty clearly said
img
Not "Add kilo alongside omega"

Omega should later be completly disabled and then finally removed from the codebase

:cl:  
rscdel: Omegastation has been removed from rotation
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -19,7 +19,6 @@ endmap
 
 map omegastation
 	maxplayers 35
-	votable
 endmap
 
 map yogsmeta


N3D6:
🦀

patpol4 (approved):
Ah very nice thank you

UselessTheremin:
goodnight my sweet prince
always in my heart, omegastation...
:(

patpol4:
@JamieD1 please at least explain why are you closing this, seems like everyone is on the same page with it and this was originally planned to go this way.

PR #11570, opened 2021-06-10
Fixed and Tweaked Some Stuff With the Crew Console Interface

nmajask:
# General Documentation

### Intent of your Pull Request

Added IDs for job that didn't have them (Paramedic, Psychiatrist, Mining Medic, Signal Technician, Clerk, Tourist, an Artist), now they will appear in the right place and have the correct color. Additionally, I removed the map from the interface and added a new civilian color that civilians use in the crew console (#535353).

### Why is this change good for the game?

The missing job IDs are a mistake I assume and it causes the layout to look wrong. The map serves little purpose and I've that seen some people dislike it. The new color is just to help the list easier to comprehend at a glance.

# Wiki Documentation

### Briefly describe your PR and the impacts of it, in layman's terms. 
This will be the basis of the Wiki entry for your PR, and more information / detail is better for Wiki editors to integrate.

Fixed and tweaked some stuff with the crew console interface

### What should players be aware of when it comes to the changes your PR is implementing?
The crew console looks better I guess
### What general grouping does this PR fall under? 
TGUI fixes
### Are there any aspects of the PR that you would like us not to mention on the Wiki?
How long it took me to find where ijob was stored
### If there are any numerical values involved in your PR that will be relevant to a player, please note them here. 
Nope

# Changelog

:cl:  
bugfix: fixed some jobs missing IDs in the crew console
tweak: Removed the map an gave civilians their own color on the crew console 
/:cl:

PR #11614, opened 2021-06-19
Revert "Fixed and Tweaked Some Stuff With the Crew Console Interface"

JamieD1:
Reverts yogstation13/Yogstation#11570


JamieD1:
Test merge only

Tipy1802:
Why?

nmajask:
So the problem seams to be that while I was trying to stop the console from getting the map assets, I accidentally stopped it from sending ANY data to tgui. I should have a PR with the fix soon.

nmajask:
I made a fix for it #11615

PR #11617, opened 2021-06-20
Y don't u fix it instead of complaing mr. monster. Also tho the fuk cares its ran a single time in reeeeeboot

alexkar598:
@monster860


Hopekz:
okay change but toxic PR

Cartlord:
What the hell even is this?

adamsong (approved):
It does what is intended

PR #11622, opened 2021-06-21
Ports #59750 from TG: Adds configs to disable hard deletes 

alexkar598:
Sometimes hard delete can go nuts and absolutly kill the server, its not a bad idea to have these available to edit at runtime via admin VV

### DIFF
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -500,3 +500,13 @@
 /datum/config_entry/string/centcom_ban_db	// URL for the CentCom Galactic Ban DB API
 
 /datum/config_entry/flag/auto_profile
+
+/datum/config_entry/flag/disable_gc_failure_hard_deletes
+
+/datum/config_entry/flag/disable_all_hard_deletes
+
+
+
+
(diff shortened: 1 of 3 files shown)


code/controllers/configuration/entries/general.dm
```diff
@@ -499,4 +499,14 @@
...
+
+/datum/config_entry/flag/disable_gc_failure_hard_deletes
+
+/datum/config_entry/flag/disable_all_hard_deletes
+
+
+
+
+
+
```
JamieD1:
```suggestion
```

config/config.txt
```diff
@@ -436,3 +436,16 @@ DEFAULT_VIEW_SQUARE 15x15
...
+
+## Uncomment to disable hard deletes entirely, even things that explicitly request. This is not recommended unless you have a need for it during events or other high pop times where performance is key.
+#DISABLE_ALL_HARD_DELETES
+
+
+
+
+
+
+
```
JamieD1:
```suggestion
```

JamieD1 (approved):

PR #11624, opened 2021-06-22
Allows admin to use HTML in centcom reports

alexkar598:
dont question


Hopekz:
lmao okay

Bibby0110:
yet we can't use it in the map name display

JamieD1:
If this gets abused it will be reverted.

JamieD1 (approved):

PR #11630, opened 2021-06-23
Adds Stasis Roller Beds

nmajask:
Due to the response I've received from this I removed the ability to make them and put one in the paramedic's staging area so people can try it. In my opinion I think it fairly balanced, but that is hard to tell without having it tested with other people.
# General Documentation

Adds stasis roller beds that act as roller bed that use cells to put people in stasis. Also makes adding new roller beds easier by changing some of the code to be less hard coded. It also requires power to be used
Here are the icons for it (these are outdated, but the only difference is a small meter on the unfolded ones):
cryo-up
On

cryo-up-off
Off

cryo-folded-on
Folded On (don't know how to edit Gifs so I cant enlarge it)

Gives paramedics a nice new tool to save people

# Wiki Documentation

Adds stasis roller beds that act as roller bed that use cells to put people in stasis.

The stasis roller bed acts as a stun baton when it comes to cells (cells can be put in when there is none, screwdrivers can remove them), the stasis roller bed still uses power when on without a patient or folded, and it can be emagged to do stamina damage over time like a normal stasis bed.

Medical Changes?

Nope
Power usage is 2x the amount of normal stasis beds (Active is 680 and passive is 80)
A stasis roller bed with a fully charged bluespace cell and a person on it lasts under 1 minute and 55 seconds.

# Changelog

:cl:  
rscadd: Added a stasis roller bed to the paramedic staging area
/:cl:


Jarod1200:
This is cool but i personally think this is extremely OP.

nmajask:
> This is cool but i personally think this is extremely OP.

Yeah, that's what I was concerned about too. I tried to balance it by increasing the power usage and material cost, but I don't know if that was enough.

Marmio64:
yeah stasis bags and stuffs work on other servers cause they have advanced wound and medical systems where those tools are practically required a lot of the time, here you can heal anything short of a missing limb on the go outside of medical very fast.

PR #11633, opened 2021-06-25
Revert "Anvil's PR template changes n stuff"

Hopekz:
Reverts yogstation13/Yogstation#10704

**Why:**
The PR template is an ugly blob of text split up into way too many categories because of this change.

Adding this huge blob of text split up into 20 categories isn't the right way to go with presenting a PR in my opinion so it makes PR's harder to read and understand.

I've seen people not even fill out the template but leave the extra garbage anyway which makes it exhausting to read if you expect useful information in the PR description.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,32 +1,29 @@
-#### Read our contributing.md if you haven't already: https://github.com/yogstation13/Yogstation/blob/master/.github/CONTRIBUTING.md
-
-# General Documentation
+########################################
+########################################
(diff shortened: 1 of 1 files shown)


JamieD1:
This was done with cooperation with the wiki team, you cant just revert it because you dont like it. Its more for the benefit of other departments.

patpol4:
Can we actually get the wiki team opinion on this? Has the template been useful to them? Does anyone even use it? Does it have to be so big? I asked on discord, lets see if they respond.

Anvilman6:
I had thought the template had been adjusted already by someone, but I may be misremembering
It's entirely alright by me if people want to change it, as when I put it together I was just trying to cover all the bases and I understand that it got lengthy and people hate filling out templates. However, I'm of the mindset it should be revised, not removed, and I think Veeble put forward a great compromise of: 

# Wiki Documentation
Write any and all information necessary regarding the changes to the item that should be known, whether it is any values that have been changed or any images that have to be updated
(located Immediately prior to the “why is this good for the game” segment)

JamieD1:
@Anvilman6 Can you make a new pr with the revised changed please?

PR #11634, opened 2021-06-26
Adds a stasis roller bed to the paramedic staging area on box

nmajask:
Due to the response I've received from the original PR (#11630) I removed the ability to make them and put one in the paramedic's staging area so people can try it. In my opinion I think its fairly balanced, but I could be wrong and its hard to tell without having it tested with other people.
# General Documentation

Adds stasis roller beds that act as roller bed that use cells to put people in stasis. Also makes adding new roller beds easier by changing some of the code to be less hard coded. It also requires power to be used
Here are the icons for it (these are outdated, but the only difference is a small meter on the unfolded ones):
cryo-up
On

cryo-up-off
Off

cryo-folded-on
Folded On (don't know how to edit Gifs so I cant enlarge it)

Gives paramedics a nice new tool to save people

# Wiki Documentation

Adds stasis roller beds that act as roller bed that use cells to put people in stasis.

The stasis roller bed acts as a stun baton when it comes to cells (cells can be put in when there is none, screwdrivers can remove them), the stasis roller bed still uses power when on without a patient or folded, and it can be emagged to do stamina damage over time like a normal stasis bed.

Medical Changes?

Nope
Power usage is 2x the amount of normal stasis beds (Active is 680 and passive is 80)
A stasis roller bed with a fully charged bluespace cell and a person on it lasts under 1 minute and 55 seconds.

# Changelog

:cl:  
rscadd: Added a stasis roller beds, with one in the paramedic staging area
/:cl:


Jarod1200:
I suggest one for the mining medic because it can be a long way from where sombody died back to the mining base medbay

Marmio64:
seems like it'd be a cooler thing to give to the CMO as one of their special items, and they could then lend it out to other people if they need to

Jarod1200:
Problem with that is cmos are dented and selfish

nmajask:
> I suggest one for the mining medic because it can be a long way from where sombody died back to the mining base medbay

I have a few issues with giving it to mining medic:
1. Because the current mining base is a test merge, I don't know how I could put it in their locker
2. Both mining bases don't have cell chargers, so you would need to go up to the station to recharge
3. I play mining medic all the time, so I feel like I shouldn't give a potentially overpowered item to a job I constantly play


> seems like it'd be a cooler thing to give to the CMO as one of their special items, and they could then lend it out to other people if they need to

I don't think the CMO needs another round start equipment in their locker and they would probably hand it out to parameds to use anyways. Also, its the same as 3 above, giving it to a job I play all the time probably isn't the best idea.

patpol4 (approved):
Map side has no issues

JamieD1:
Why dont we just port the stasis bags from bay that have been around for like 5 years.

Anvilman6:
I still heavily disagree with this, as even with the limitation of one showing up in a round, the bed by design is still too powerful. 

Our Medical systems are very simplistic, in that most situations can be easily addressed by straightforward onsite triage (stem bleeding, administer Epinephrine, gather any missing limbs, Styptic Powder / Silver Sulfadiazine spray, etc.) to get a person in a stable enough state to transfer. A stasis roller bed narrows that loop even further by keeping the patient in exactly the same state as they've been found, limiting the amount of triage care needed, which really just seems too easy for TG/Yog Med.

If we had a Medical system with more depth and more negative effects piled on a victim (i.e. BayMed), I could understand the justification of needing a way to hit a pause button on a patient's status in the field before taking them back to Medical, but it's not the case now.

nesisknesis (approved):

JamieD1:
Overpowered. One use stasis bags if you want this that are researchable and cmo starts with one round start.

PR #11637, opened 2021-06-27
Makes grav gen not double in weight every time its triggered..

JamieD1:
Fixes Grav gen doubling in weight everytime its triggered.

:cl:  
bugfix: fixed grav gen triggering way too much
/:cl:


patpol4:
Why was this a thing lol

PR #11639, opened 2021-06-27
Instead of doubling the chance of the gravity event happening every time its triggered, it only doubles it if the gravity generator is active when the event ends

alexkar598:
# General Documentation

Discourages rebooting the gravity generator, bypassing the event completly

# Wiki Documentation

Instead of doubling the chance of the gravity event happening every time its triggered, it only doubles it if the gravity generator is active when the event ends

Instead of doubling the chance of the gravity event happening every time its triggered, it only doubles it if the gravity generator is active when the event ends

Events

No

For example, "This new chem will heal X brute damage". 
Doubling = original value * 2
Alternatively if youre feeling spicy, you can do original value / 0.5

# Changelog
:cl:  
tweak: Instead of doubling the chance of the gravity event happening every time its triggered, it only doubles it if the gravity generator is active when the event ends
/:cl:

### DIFF
--- a/yogstation/code/modules/events/weightless.dm
+++ b/yogstation/code/modules/events/weightless.dm
@@ -19,8 +19,8 @@
 	for(var/obj/machinery/gravity_generator/main/station/A in GLOB.machines)
 		if(A)
 			A.set_state(0)
-	if(control)
-		control.weight *= 2
+			if(control && A.on)
+				control.weight *= 2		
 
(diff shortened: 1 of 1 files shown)


Cartlord:
My main issue with this is that there likely won't be anything pointing to this in-game, so people will reboot the GravGen anyways and get screwed for it.

JamieD1 (approved):

PR #11647, opened 2021-06-28
Wiping down things with a rag will now apply the reagents in the rag to the item

### LINKED ISSUE #6292: Can no longer apply drying agent to galoshes via rag
I used to apply drying agent to rags and then wash the boots with them and then I would get the fancier galoshes but now it doesn't work and to the best of my knowledge there is no other way.

adamsong:
# General Documentation

When wiping down objects and turfs with a wet rag, the reagents in the rag will be applied to the object or turf.

The rag already causes a person being wiped down to absorb/ingest the reagents, so now it also applies to items.

# Changelog
Fixes #6292

:cl:  
bugfix: Damp rag now applies reagents to objects and turfs
/:cl:


JamieD1 (approved):

MegaEmpirical:
Glorious return to form for janitors and bartenders everywhere

PR #11675, opened 2021-07-01
fix directional x4

### LINKED ISSUE #7417: X4 is not directional any more
As title.

Theos looked into it and seems to have an understanding of the problem if someone wants to poke him for details before trying to fix it.

SomeguyManperson:
fixes #7417
for a while x4 wouldn't apply its offset when breaching a wall which kills the ENTIRE POINT of using it as a breaching charge over c4 the bit where you blow a nice large hole in the room and don't have to run as far from the initial breaching point to not eat shit
turns out it's because the wall is considered a floor after being exploded which makes it not trigger the directional check since it's no longer dense when that happens

:cl:  
bugfix: x4 now explodes properly when placed on walls 
/:cl:

### DIFF
--- a/code/game/objects/items/grenades/plastic.dm
+++ b/code/game/objects/items/grenades/plastic.dm
@@ -57,16 +57,18 @@
 
 /obj/item/grenade/plastic/prime()
 	var/turf/location
+	var/density_check = FALSE
 	if(target)
 		if(!QDELETED(target))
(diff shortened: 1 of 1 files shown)


adamsong:
Please link to #7417

JamieD1 (approved):

MegaEmpirical:
Fucking finally, thank you.

PR #11679, opened 2021-07-01
Gun Protolathe in Armory remove gun from sec lathe [TODO]

Bibby0110:


patpol4:
wat

adamsong:
You seem to have pushed more changes than you intended

Bibby0110:
> You seem to have pushed more changes than you intended

this is just a random branch and draft PR so we don't forget to do it

PR #11686, opened 2021-07-01
Alternative Fixes gun pin removal spawning guns

### LINKED ISSUE #11453: Pin-Removal Gun Glitch
## Test Merges: 
Not Supplied / None
## Reproduction:
So in the weapons section of the personal crafting menu, there seems to be an option to remove the firing pin from a gun that has one.  If you have a welding tool, a screwdriver, and a pair of wirecutters, you can complete it.  Standing over the gun and going through the process removes the firing pin, but doesn't actually give you the firing pin.

The bug gets funnier when you try to do it again, because the option remains available even though you just removed the pin from the gun.  Using it again goes through the process.  Instead of giving you a firing pin, it spawns under your original gun a small revolver simply called "gun".  It couldn't load any of the ammo I tried, and I found you could do it a bunch of times, leading to many tiny revolvers being created.  I have no idea how this happened, but it was a bug so I'm reporting it.
 Submitted by: Xaahdubc (222019245821657089)

Bibby0110:
# General Documentation

Fixes #11453

Makes it so that attempting to remove a pin from a gun with no pin does not runtime and does not create a the base gun type.

Spawning guns is bad

# Changelog

:cl:  
bugfix: Removing firing pins from firing-pin-less guns no longer makes the game panic.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -115,7 +115,9 @@
 	if(G)
 		G.forceMove(loc)
 		var/pin = G.pin
-		if(G.pin.gun_remove()) //if this returns false the gun and pin are not going to exist
+		if(!pin)
+			visible_message("[G] has no pin to remove.", null, null, 3)
+		if(pin && G.pin.gun_remove()) //if this returns false the gun and pin are not going to exist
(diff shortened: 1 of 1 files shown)


adamsong (approved):
It’s my code but better

PR #11700, opened 2021-07-02
Burnmix is now a worse ghetto floor cleaner

Hopekz:
**Why:**

Burnmix is mainly alcohol and ash. 
Those 2 ingredients would be really good at scrubbing down floors.
Burnmix does not clean objects or interact with mobs so it isn't as good as space cleaner (does 1 of the 3 things space cleaner does).

Also fixed a line spacing bug with perfluorodecalin overdose. 
Did perfluorodecalin overdose even work before this PR?

:cl:  
rscadd: Burnmix is now able to clean turfs but is worse than space cleaner because it doesn't clean objects or mobs.
bugfix: perfluorodecalin overdose fix
/:cl:


adamsong (approved):
Don't see why not

PR #11703, opened 2021-07-03
^ is not exponent, tg coders

monster860:
imagine being a tg coder


adamsong:
Would a bit shift be faster or will the compiler just eat it anyway?

adamsong (approved):
Good fix, tg bad

Mothblocks:
> Would a bit shift be faster or will the compiler just eat it anyway?

byond, believe it or not, performs constant folding

PR #11705, opened 2021-07-04
Ports /tg/ tend wounds time estimation

maxion12345:
#### Read our contributing.md if you haven't already: https://github.com/yogstation13/Yogstation/blob/master/.github/CONTRIBUTING.md

# General Documentation

This PR makes it so during tend-wounds, you get messages that allow you to estimate how much more surgery you have before the patient is completely fixed. If you are holding a health analyzer in your off-hand, then you are shown damage numbers instead.
With health-scanning being weird during surgery, and operating computers sometimes not working or taking up screenspace, this will allow estimates to be done on if its worth doing the surgery.
# Wiki Documentation
n/a
see above

see above
tweaks

n/a
At 1 to 3 steps of damage, you are "finishing up"
at 3 to 6 steps of damage, you are "counting down the last bits of damage"
at 6 to 9 steps of damage, you are "plugging away"
at 9 to 12 steps, you are "steadying yourself"
at 12 to 15 the target will "look like x"
anything bigger and you "feel like you are barely making a dent"

# Changelog

:cl:  
tweak: Tend-wounds surgery now gives time estimates in the form of messages. Holding a health analyzer in your off-hand during surgery gives exact damage numbers now.
/:cl:

### DIFF
--- a/code/modules/surgery/healing.dm
+++ b/code/modules/surgery/healing.dm
@@ -28,6 +28,9 @@
 	var/burnhealing = 0
 	var/missinghpbonus = 0 //heals an extra point of damager per X missing damage of type (burn damage for burn healing, brute for brute). Smaller Number = More Healing!
 
+/datum/surgery_step/heal/proc/get_progress(mob/user, mob/living/carbon/target, brute_healed, burn_healed)
+	return
+
 /datum/surgery_step/heal/preop(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
 	var/woundtype
 	if(brutehealing && burnhealing)
@@ -67,6 +70,8 @@
 		umsg += " as best as you can while they have clothing on"
 		tmsg += " as best as they can while [target] has clothing on"
 	target.heal_bodypart_damage(urhealedamt_brute,urhealedamt_burn)
+	umsg += get_progress(user, target, urhealedamt_brute, urhealedamt_burn)
+
 	display_results(user, target, "<span class='notice'>[umsg].</span>",
 		"[tmsg].",
 		"[tmsg].")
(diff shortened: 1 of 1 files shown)


code/modules/surgery/healing.dm
```diff
@@ -113,6 +118,33 @@
...
+/datum/surgery_step/heal/brute/get_progress(mob/user, mob/living/carbon/target, brute_healed, burn_healed)
+	if(!brute_healed)
+		return
+
+	var/estimated_remaining_steps = target.getBruteLoss() / brute_healed
+	var/progress_text
+	if(locate(/obj/item/healthanalyzer) in user.held_items)
+		progress_text = ". Remaining brute: <font color='#ff3333'>[target.getBruteLoss()]</font>"
+	else
+		switch(estimated_remaining_steps)
```
adamsong, line 130: Drop this into its own proc, so if it needs updated in the future, it won't need updated three times

adamsong (changes requested):

code/modules/surgery/healing.dm
maxion12345: i'll figure out how to do that eventually

adamsong: Make a proc, call it something like `get_steps_message` that takes a number, and returns the string associated with that many steps remaining

maxion12345: wouldn't it be three different procs then? each surgery has different strings depending on if its brute, burn, or combo

adamsong: That is what I get for not reading properly, I thought it was the same messages

adamsong (approved):
Looks good

PR #11711, opened 2021-07-05
Ports filter surgery

maxion12345:
Port: https://github.com/BeeStation/BeeStation-Hornet/pull/2603

# General Documentation

Adds two things:
The Blood Filter surgical tool that basically exists to filter blood
and the Filter Blood surgery that exists to remove chemicals from people actively, rather then relying on passive chemical removal. Only appears when there are reagents in the body. 

you ever get those guys dosed with absurd amounts of chems and go through the cycle of "revive, let him process a bit of chem out of his system, he dies from the chem, repeat"? thats why this change is good.
# Wiki Documentation

wiki documentation is that the surgery is scalpel, retractor, scalpel, then blood filter (auto-repeats) , then cautery

the players can look up a bit and read the shit I already wrote
look at the changelog

The more chem you have in you, the faster it gets removed. There are diminishing returns, so you'll want to pair this with traditional methods to achieve the fastest chem removal

# Changelog

:cl:  Chessus
rscadd: Added the blood filter item and surgery for when you need to revive the assistant with 300 of each alcoholic drink in him, or for when dipshit mcdoctorperson spams every conceivable chem on the sleeper
/:cl:


code/modules/surgery/blood_filter.dm
```diff
@@ -0,0 +1,61 @@
...
+	if(HAS_TRAIT(target, TRAIT_HUSK) || target.reagents.total_volume==0) //Can't filter husk or 0 regent body
+		return FALSE
+	return ..()
+
+/datum/surgery_step/filter_blood/preop(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
+	if(istype(surgery,/datum/surgery/blood_filter))
+		var/datum/surgery/blood_filter/the_surgery = surgery
+		if(!the_surgery.antispam)
+			display_results(user, target, "<span class='notice'>You begin filtering [target]'s blood...</span>",
+		"<span class='notice'>[user] uses [tool] to filtering your blood.</span>",
```
SomeguyManperson: this line makes me want to die

SomeguyManperson:
reference the original pr in the description somewhere i.e. these two bits
image
then # and the pr's number except the organization and repository from where the pr was ported from

code/modules/surgery/blood_filter.dm
maxion12345: i don't know enough about coding to know why

adamsong: “Uses blood filter to filtering your blood”

maxion12345: ahf uck i see it now

nesisknesis (approved):
awful

also if its a port you generally put the original author's name here so that they get credited in the changelog 
image

MegaEmpirical:
How does this work with plasmamen?

maxion12345:
> How does this work with plasmamen?

Just like other surgeries, you'll want them under a shower before you begin. If the amount of chems is large enough to warrant doing this surgery then it shouldn't take too many steps to bring them down to a level you can manage with just chems.

adamsong:
Plasmemes don't have blood though

maxion12345:
It works the same way, I suppose we can change it so it filters the "plasma"? Not even sure how plasma people retain chemicals if they are skeletons. Bone filtering surgery, I guess.

guess thats a to-do, to make it so it either doesn't work on plasmapeople or so that it says "filter plasma" instead of blood.

maxion12345:
Closing because I've lost the will to do further work on a very situational surgery

N3D6:
F

PR #11720, opened 2021-07-05
Hopefully makes diagonal movement faster.

ghost:
# General Documentation

Ports [this PR.](https://github.com/BeeStation/BeeStation-Hornet/pull/4699) My private server lagged, but I don't think it was because of smooth diagonal movement.

The movement delay for going diagonally was set to an arbitrary 2x multiplier. The actual multiplier should be sqrt(2), which is ~1.4

# Wiki Documentation

Faster diagonal movement.

As much as I hate repeating the same thing for every part of this template, faster diagonal movement.

Movement tweaks.

The movement delay uses 1.414214 now. What does this mean? _I don't really know._

# Changelog

:cl:  
tweak: diagonal movement is faster now
/:cl:

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -45,17 +45,17 @@
   * Move a client in a direction
   *
   * Huge proc, has a lot of functionality
-  * 
+  *
   * Mostly it will despatch to the mob that you are the owner of to actually move
   * in the physical realm
-  * 
+  *
   * Things that stop you moving as a mob:
   * * world time being less than your next move_delay
   * * not being in a mob, or that mob not having a loc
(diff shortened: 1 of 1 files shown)


Darkstick:
This makes all movement significantly faster. I don't like that... Besides, diagonal movement is supposed to be slower since it's traveling two tiles at once

Cartlord:
Please do not do this

adamsong:
> This makes all movement significantly faster.

It does not

> Please do not do this

Why? All it does it make it so the delay is appropriate for the actual distance you travel

ghost:
The delay **should be** exactly equivalent to walking in a cardinal direction.

JamieD1 (approved):

PR #11729, opened 2021-07-07
Moves mindshield firing pin to armory lathe

Bibby0110:

Can't make guns outside of armory. Can't use pin without gun, no need for pin to be outside armory.

If anything this just encourages sec to go get guns somehow. There's a reason the Warden had all the normal firing pins in the first place.

# Wiki Documentation

Mindshield firing pins from security lathe to armory lathe 

# Changelog

:cl:  
tweak: Moved mindshield firing pin to Armory protolathe
/:cl:

### DIFF
--- a/code/modules/research/designs/weapon_designs.dm
+++ b/code/modules/research/designs/weapon_designs.dm
@@ -102,7 +102,7 @@
 	materials = list(/datum/material/silver = 600, /datum/material/diamond = 600, /datum/material/uranium = 200)
 	build_path = /obj/item/firing_pin/implant/mindshield
 	category = list("Firing Pins")
-	departmental_flags = DEPARTMENTAL_FLAG_SECURITY
+	departmental_flags = DEPARTMENTAL_FLAG_ARMORY
 
 /datum/design/stunmine/sec //mines ported from BeeStation
 	name = "Stun Mine"


adamsong:
Its a firing pin that still requires a gun and a mindshield. To make use of it you need to obtain a gun without a firing pin (armory) and a mindshield (armory, HoS locker, cargo) and if you are getting mind shields from cargo, you can just order guns with electronic pins.

adamsong (approved):
Don't think this should be merged but the code does as described, so I'm giving approval

MegaEmpirical:
....your argument doesn't make sense. Everyone but the detective in security starts with a gun. 

And they all start with mindshields.

Bibby0110:
> ....your argument doesn't make sense. Everyone but the detective in security starts with a gun.
> 
> And they all start with mindshields.

The point is it’s the wardens job in the first place to hand out weaponry, so why allow security officers to create pins for something they shouldn’t be interacting with anyways.

MegaEmpirical:
> > ....your argument doesn't make sense. Everyone but the detective in security starts with a gun.
> > And they all start with mindshields.
> 
> The point is it’s the wardens job in the first place to hand out weaponry, so why allow security officers to create pins for something they shouldn’t be interacting with anyways.

Your argument still doesn't make sense. They start with guns, it is literally part of their base kit.

Cartlord:
What if the Det wants to mindshield his revolver, which by-default has an electronic pin? That doesn't seem like it needs OK'd by Warden.

PR #11731, opened 2021-07-07
Changes admin logging due to abuse.

Bibby0110:
[REDACTED]


JamieD1:
:thonk:

Bibby0110:
> :thonk:

Above your security clearance, sorry pal.

alexkar598:
Since the start of the month of june (2021-06-01), this log has been fired 29 times, out of these 29 times, i could only not find reasoning for 2 of them. The message is way too aggressive and hostile for an event with a high false positive rate

Bibby0110:
> Since the start of the month of june (2021-06-01), this log has been fired 29 times, out of these 29 times, i could only not find reasoning for 2 of them. The message is way too aggressive and hostile for an event with a high false positive rate

:)

adamsong:
Sufficient reasoning for this change has not been provided

PR #11745, opened 2021-07-10
so tcomms were still generating heat and calculating speed even when offline

Lostlarry:
so i'm back and since i remmember that someone complained that my kentucky fried tcomms code was borked i decided to take a look at it and fix it.
the new calculation should be more resistant to rounding errors
also the calculation is also skipped if the machine is turned off 

:cl:  
bugfix: tcomms no longer do kentucky fried tcomms when offline
tweak: changed the speed calculation between 150K and 1150K to make more consistent
/:cl:

### DIFF
--- a/code/game/machinery/telecomms/telecomunications.dm
+++ b/code/game/machinery/telecomms/telecomunications.dm
@@ -144,14 +144,8 @@ GLOBAL_LIST_EMPTY(telecomms_list)
 	else
 		on = FALSE
 
-/obj/machinery/telecomms/process()
-	update_power()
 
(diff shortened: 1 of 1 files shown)


Bibby0110:
Don’t have time to run through the code, but I’ve always been wondering if the use of round() is what is intended, as it is in the code it just floors the value, doesn’t round up/down. Is this intended?

Lostlarry:
honesty the question whether to use floor or round is more about balance then implementation

adamsong (approved):
Simple enough change, looks good

PR #11756, opened 2021-07-12
should fix progenitor roar spam

SomeguyManperson:
redoes progenitor screaming code a bit so it's less terrible and more readable
every roar sets the cooldown to 30 seconds, with a roar automatically occuring if 10 seconds elapse after the cooldown is over to give time to roar manually not like that's important
variables renamed to make it more readable and using time defines
:cl:  
bugfix: progenitors should no longer be able to spam the super loud and annoying scream noise so it's more t h e m a t i c
/:cl:


adamsong (approved):
Simple change, does what it says

PR #11759, opened 2021-07-13
Makes the warning when your brain is about to die bigger.

ghost:

**Untested.** Webedited.

This PR makes the <sup>tiny little text</sup> that shows up when you are **about to fucking die of brain death** bigger. Ideally, you shouldn't get a <sup>tiny little text</sup> that indicates you are going to die. Also changes the second message to use "danger" instead of "warning."

# Wiki Documentation

Nothing important, just changing some text to make **messages indicating you are going to die** bigger.

# Changelog

:cl:  
tweak: you will get a big text when you're about to die of brain death instead of a small one.
/:cl:


adamsong (approved):
Simple change, good idea

PR #11768, opened 2021-07-13
Re-adds old cloaks along side new capes.

JamieD1:
Alot of people complained they missed the old cloaks,

So I added them as alternatives.


alexkar598:
you could always put clothes in dufflebags tho? the only difference now is that you can ONLY put clothes in the bags.... which actually annoys me

Bibby0110:
You should probably atomize the duffelbag changes since the cloak thing is semi-controversial (?)

PR #11769, opened 2021-07-14
Adds more alts for shawls

nmajask:

Adds more alts for shawls, most of which are to make it look like the old cloaks. Like before, they are toggled by alt clicking a normal cloak while not worn. Here are the changes:

- Cap & HoP cloaks have alts with darker colors
- CMO, CE, & QM cloaks have alts that looks like the old cloaks for people who prefer them over the new ones
- HoS cloak has an alt with lighter colors which might help it stand out

Pictures:
capshawl
hopshawl
hosshawl
cmocloak
ceback2
qmshawl

# Wiki Documentation

Nothing should need to be changed on the wiki

# Changelog

:cl:  
imageadd: added more cloak alts  
/:cl:


Mqiib:
They still look kinda bulky from the side ngl. The old cloaks were kind of nice because they weren't so poofy on the side view comparatively.

adamsong:
The new shawls definitely feel too baggy

nesisknesis:
I can understand wanting to appease people complaining but in cases like these, people will complain to the end of time if something isn't their favourite shade of blue. Happens all the time and they need to get over if something has changed to not their specific expectations.

Besides I said I would edit them myself and i'm not going to stop halfway through.

PR #11771, opened 2021-07-14
Adds overdose list for anvil - Might work who knows

JamieD1:
Adds overdose list for anvil - Might work who knows


Anvilman6:
Cheers
Here's hoping it actually works

Anvilman6:
It in fact does not work

adamsong (approved):
Simple enough change, looks good

PR #11782, opened 2021-07-16
Enables late join heretics, lowers their weight, etc.

Xoxeyos:
# General Description

Heretics are now enabled for latejoin dynamic, with a lower weight, which means they appear less, they need 36 players in order to spawn. Traitors also have a lower cost now, from 15 down to 10, and the vampire requirements for player scaling have been lowered, and ordered correctly.

# Wiki Documentation

Heretics can now join in late, during a dynamic shift.

:cl:  Xoxeyos
rscadd: Heretics are now enabled for dynamic latejoin.
tweak: Tweaked heretic weights, traitor costs, and gamemode requirements.
/:cl:


adamsong (approved):
Code looks good

PR #11786, opened 2021-07-16
Removes blackout station trait.

Bibby0110:

This is the most annoying station trait, talked to a few people that agree.

Doesn't add anything to the gamer other than actual pain and suffering

# Wiki Documentation

May have to remove from wiki if it's on there.

# Changelog

:cl:  
tweak: NanoTrasen has installed more durable lightbulbs are you should no longer encounter half of the stations lights being destroyed when you arrive.
/:cl:


Cartlord:
Tested, I assume?

Marmio64:
I think weight should be reduced and the probability of a APC overloading be reduced to like 30%, full removal seems unnessecary

PR #11796, opened 2021-07-18
Changes the loop in Jamie's overdose PR to a typeless loop for maximum sanic speed

Bibby0110:

By my calculations if we hire someone for 40 hours a week to sit on the server and only call this proc we would 0.0006 pico seconds per call, and in the next 10.000 years we would save upwards of 0.2 seconds

# Wiki Documentation
None


adamsong:
While the format is in place, please actually fill it in

adamsong (approved):
Fixing Jamie’s shitcode, one line at a time

JamieD1 (approved):

PR #11797, opened 2021-07-18
Makes ash storms slightly quieter

JamieD1:
Makes ash storms slightly quieter

🆑
tweak: Makes ash storms slightly quieter
/🆑


adamsong (approved):
Simple change, looks good pending 24 hour hold

PR #11804, opened 2021-07-18
Basic laser pointers are much worse at stunning borgs, with upgraded pointers becoming much more potent at the expense of how often they can be used.

Bibby0110:

Max charges reduced to 4.

Every upgraded laser reduces the max charges by 1, with the max being 1 charge at level 4.

Stun reduced from 10-20 seconds to 2.5-5 seconds.

A standard laser pointer with 4 charges has a 11.25% chance of stunning a borg per hit. (38% chance of getting AT LEAST 1 stun.) 
(1% for 2 stuns, 0.1% for 3 stuns, 0.01% for 4 stuns)

Level 2 laser has a 28.68% chance of stunning with 3 charges. (64% chance of at least 1 stun) (8% for 2 stuns, 2% for 3 stuns.)

Level 3 laser has a 49.58% chance of stunning with 2 charges. (75% chance of at least 1 stun) (24.5% for 2 stuns)

Level 4 laser has a 73.1% chance of stunning with 1 charge. 

Level 4 laser has a lower overall probability but increases stun amount by 100%. (5-10 seconds) Why? Because when you use a level 4 laser once it overloads and requires 4 charge cycles (10% chance per process).

Best way to use this:
Shoot borg untill you have 1 charge left. Run the duck away unless you want to try your luck and get your skull bashed/flashed in by the borg.

# Wiki Documentation

# Changelog

:cl:  
tweak: Laser pointers are now much more potent at stunning borgs when upgraded using a better quality laser diodes. The increased power load does however mean that the max charges the laser pointer can hold will degrade by 1 per tier of upgraded diode.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/laserpointer.dm
+++ b/code/game/objects/items/devices/laserpointer.dm
@@ -34,7 +34,7 @@
 	diode = new(src)
 	max_energy = initial(max_energy) - (diode.rating - 1)
 	if(energy >= max_energy)
-			energy = max_energy
+		energy = max_energy
 	if(!pointer_icon_state)
 		pointer_icon_state = pick("red_laser","green_laser","blue_laser","purple_laser")
 
--- a/code/game/objects/items/devices/laserpointer.dm
+++ b/code/game/objects/items/devices/laserpointer.dm
@@ -126,7 +126,7 @@
 		var/mob/living/silicon/S = target
 		log_combat(user, S, "shone in the sensors", src)
 		//chance to actually hit the eyes depends on internal component, borgs are hard to stun :/
-		if(prob(effectchance * (diode.rating ** 0.75) * 0.75))
+		if(prob(effectchance * (diode.rating ** 1.35) * 0.75))
 			S.flash_act(affect_silicon = TRUE)
 			var/stun_modifier = diode.rating >= 4 ? 2 : 1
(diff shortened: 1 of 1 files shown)


Darkstick:
Can we not remove borg counters? They're kind of OP.

maxion12345:
giving players more control over stuns so its not complete rng? yes

but making the highest tier make the player completely reliant on "if you don't roll this dice and hit your waiting 4 entire recharge cycles good luck lol"
 
 not so good


should just remove the random completely and have a fixed stun chance with variable stun time depending on diode tier

Bibby0110:
> giving players more control over stuns so its not complete rng? yes
> 
> but making the highest tier make the player completely reliant on "if you don't roll this dice and hit your waiting 4 entire recharge cycles good luck lol"
> 
> not so good
> 
> should just remove the random completely and have a fixed stun chance with variable stun time depending on diode tier

So your solution for having a probability based stun is to remove the probability and replace it with.. a probability?

There's no way this shouldn't be RNG based, it would mean the stun time would have to be completely worthless.
It's already a 33% stun normally. This is a pretty big nerf to them, which is what is intended.

Reed0506:
would be cool if you could cycle power levels with the better lasers, perhaps for more shots at whatever power level than the previous part would be able to achieve, even if just by a little bit. T4 being a sidegrade rather than an upgrade to T3 is pretty meh.

JamieD1:
Other solution is being worked on instead.

PR #11806, opened 2021-07-19
Removes STATE YOUR LAWS from god speak. Fed up of its abuse.

JamieD1:
Removes STATE YOUR LAWS from god speak. Fed up of its abuse.

### DIFF
--- a/code/modules/surgery/organs/vocal_cords.dm
+++ b/code/modules/surgery/organs/vocal_cords.dm
@@ -230,7 +230,6 @@
 	var/static/regex/whoareyou_words = regex("who are you|say your name|state your name|identify")
 	var/static/regex/saymyname_words = regex("say my name|who am i|whoami")
 	var/static/regex/knockknock_words = regex("knock knock")
-	var/static/regex/statelaws_words = regex("state laws|state your laws")
 	var/static/regex/move_words = regex("move|walk")
 	var/static/regex/left_words = regex("left|west|port")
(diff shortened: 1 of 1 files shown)


adamsong (approved):

Bibby0110:
Fill out the template

Hopekz:
u ded

God speak is supposed to be something rare not just bussed in every round. If it is actually obtained the proper way why is this an issue and considered "abuse" when it is one of the abilities?

this is on bad admemes if god speak is just being given out all the time.
Maybe make a different version of god speak that doesn't have any of the actual abilities if staff want to keep handing it out.

N3D6:
> u ded
> 
> God speak is supposed to be something rare not just bussed in every round. If it is actually obtained the proper way why is this an issue and considered "abuse" when it is one of the abilities?
> 
> this is on bad admemes if god speak is just being given out all the time.
> Maybe make a different version of god speak that doesn't have any of the actual abilities if staff want to keep handing it out.

Staff aren't even handing it out, you get it from killing an easy megafauna with attacks that can be mostly ignored.

N3D6:
maybe a better change than all out deleting it would just be to make it not state hacked/malf laws

Cartlord:
But then why have it at all?

maxion12345:
this is a sole ability intended to cuck malf/rogue borgs. Hell, my last AI round I got subverted and we only got found out after a miner decided to go checking laws with god-voice

we murdered them afterward, but still

PR #11809, opened 2021-07-19
Makes accessories get deleted less

Mqiib:

Makes accessories detach when the thing they're attached to gets destroyed rather than just ceasing to exist.

# Wiki Documentation

not needed

# Changelog

:cl:  

bugfix: accessories get deleted less

/:cl:


Mqiib:
ah need to fix this still

PR #11822, opened 2021-07-20
Supermatter surge now asks for severity + should we announce when admin triggered

### LINKED ISSUE #11181: Supermatter Surge event does not ask if it should be announced
If I force trigger the event it forces the event.

Also, no option for severity?!!

Bibby0110:

Fixes https://github.com/yogstation13/Yogstation/issues/11181

# Wiki Documentation

# Changelog

admin abuse deserves no changelog

### DIFF
--- a/code/modules/events/supermatter_surge.dm
+++ b/code/modules/events/supermatter_surge.dm
@@ -10,7 +10,7 @@
 	if(!check_rights(R_FUN))
 		return
 
-	var/selected_power = input("How severe should it be? Between 1000 and 100000") as num|null
+	var/selected_power = input("How severe should it be? Between 1000 and 100000. Press cancel for random") as num|null
 	if(!selected_power)
 		selected_power = rand(1000, 100000)
 	selected_power = clamp(selected_power, 1000, 100000)
--- a/code/modules/events/supermatter_surge.dm
+++ b/code/modules/events/supermatter_surge.dm
@@ -4,6 +4,19 @@
 	weight = 15
 	max_occurrences = 4
 	earliest_start = 10 MINUTES
+	var/forced_power
+
+/datum/round_event_control/supermatter_surge/admin_setup()
+	if(!check_rights(R_FUN))
+		return
+
+	var/selected_power = input("How severe should it be? Between 1000 and 100000") as num|null
+	if(!selected_power)
+		selected_power = 1000
+	selected_power = clamp(selected_power, 1000, 100000)
+	forced_power = selected_power
+
+	
 
 /datum/round_event_control/supermatter_surge/canSpawnEvent()
 	if(GLOB.main_supermatter_engine?.has_been_powered)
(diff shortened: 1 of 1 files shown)


code/modules/events/supermatter_surge.dm
```diff
@@ -4,18 +4,36 @@
...
 	if(GLOB.main_supermatter_engine?.has_been_powered)
 		return ..()
 
 /datum/round_event/supermatter_surge
 	var/power = 2000
+	announceWhen = 1
 
 /datum/round_event/supermatter_surge/setup()
-	power = rand(1000,100000)
+	var/datum/round_event_control/falsealarm/C = control
```
adamsong:
```suggestion
	var/datum/round_event_control/supermatter_surge/C = control
```

Pretty sure this is the correct type

adamsong (commented):
I don't see where it is giving the option to choose if it is announced

Bibby0110:
> I don't see where it is giving the option to choose if it is announced

Handled elsewhere, requires announce_when  to be above 0 which it wasn't by default

Xoxeyos:
Could we please also get an administration log, when an admin forces it?

code/modules/events/supermatter_surge.dm
```diff
@@ -4,18 +4,36 @@
...
 	max_occurrences = 4
 	earliest_start = 10 MINUTES
+	var/forced_power
+
+/datum/round_event_control/supermatter_surge/admin_setup()
+	if(!check_rights(R_FUN))
+		return
+
+	var/selected_power = input("How severe should it be? Between 1000 and 100000") as num|null
+	if(!selected_power)
```
adamsong, line 14: This should revert to random, not set the value to minmum

code/modules/events/supermatter_surge.dm
```diff
@@ -4,18 +4,36 @@
 	weight = 15
 	max_occurrences = 4
 	earliest_start = 10 MINUTES
+	var/forced_power
+
+/datum/round_event_control/supermatter_surge/admin_setup()
+	if(!check_rights(R_FUN))
+		return
+
+	var/selected_power = input("How severe should it be? Between 1000 and 100000") as num|null
```
adamsong: You might want to make the class, not raw value, its the more familiar number

code/modules/events/supermatter_surge.dm
```diff
@@ -4,18 +4,36 @@
...
+	if(!selected_power)
+		selected_power = 1000
+	selected_power = clamp(selected_power, 1000, 100000)
```
adamsong, line 16: Why clamp it? The adminbus demands class -10000

adamsong (changes requested):

code/modules/events/supermatter_surge.dm
Bibby0110: It's clamped, deal with it

code/modules/events/supermatter_surge.dm
Bibby0110:
```suggestion
	var/selected_power = input("How severe should it be? Between 1000 and 100000. Press cancel for random") as num|null
```

code/modules/events/supermatter_surge.dm
```diff
@@ -4,18 +4,36 @@
...
 	earliest_start = 10 MINUTES
+	var/forced_power
+
+/datum/round_event_control/supermatter_surge/admin_setup()
+	if(!check_rights(R_FUN))
+		return
+
+	var/selected_power = input("How severe should it be? Between 1000 and 100000") as num|null
+	if(!selected_power)
+		selected_power = 1000
```
Bibby0110:
```suggestion
		selected_power = rand(1000, 100000)
```

Bibby0110:
> Could we please also get an administration log, when an admin forces it?

Same logs as forcing other random events, what more do you need?

adamsong (approved):
Still think it shouldn't be clamped but this will do

PR #11827, opened 2021-07-20
Removes 4 ion law listings because they don't seem like they belong in this server

Cartlord:

Webedited and untested. I removed "the Captain's anus", "the Clown's anus", "Watch pornography", and "Angry black men" from a few ion law lists. This should also make those terms not come up as Syndie codewords?
I could probably have gotten away with removing a few more, but these seemed the most egregious.

# Wiki Documentation

I don't think there's an exhaustive list of ion laws on the wiki, so likely none required.

# Changelog

:cl:  
rscdel: removed "the Captain's anus" and "the Clown's anus" from the list of ion law places
rscdel: removed "watch pornography" from the list of ALLMUST possibilities.
rscdel: removed "Angry black men" from the ion law threat listing.
/:cl:


Cartlord:
For the record, the reason for the "Angry black men" removal is because it's apparently a possible Syndicate codephrase, and I do not see how one is supposed to bring up "Angry black men" in conversation - that is, without making it super-obvious that it's a codeword - without just saying something racist.

JamieD1 (approved):

PR #11853, opened 2021-07-20
Mimes can recite clockwork spells even while mute

Bibby0110:
Since it's so vital to the gamemode it feels weird to gimp the mime for existing

# Wiki Documentation

# Changelog

:cl:  
tweak: Mimes can invoke clockwork spells while mute
/:cl:

### DIFF
--- a/code/modules/antagonists/clockcult/clock_helpers/scripture_checks.dm
+++ b/code/modules/antagonists/clockcult/clock_helpers/scripture_checks.dm
@@ -50,4 +50,4 @@
 	GLOB.clockwork_construction_value = max(0, GLOB.clockwork_construction_value + amount)
 
 /proc/can_recite_scripture(mob/living/L, can_potentially)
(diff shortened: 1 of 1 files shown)


Jarod1200:
Ratvar hates powergamers and validhunters

SomeguyManperson:
could this check for specifically mimes because i Think miners can get miming from the not-syndicate-loot miming drill ruin and I hate miners

Bibby0110:
> could this check for specifically mimes because i Think miners can get miming from the not-syndicate-loot miming drill ruin and I hate miners

that would be shitcode

alexkar598:
Mimes can break their vow if they wanna use spells

Bibby0110:
> Mimes can break their vow if they wanna use spells

seems kind of pathetic to punish the mime for playing their role as intended but sure

JamieD1 (approved):

PR #11864, opened 2021-07-21
Doing a major TEG rework because the current one is a mess that breaks thermodynamics

Lostlarry:
so the TEG was doing some real nonsense with thermal capacities so I redid it so it would actually make some sort of sense physics wise

the original PR I made was a buff to the power of TEG

Alright after giving some thought to changes I originally made I decided to lookup how TEGs work irl.
TEG works by having a high electrically conductive material with low heat conductivity hot on one side and cold on the other.
to represent this in-game there will be two internal heat masses (one for cold loop one for hot loop) that will first equalise temperature with incoming gas then they based on temperature difference and heat capacity will produce electricity and then equalise temperature with one another

resulting impact on the game:
running a TEG only on heaters and coolers will no longer create excess electricity
will make TEG behave in more physically logical way while maintaining the importance of pressure differential in both loops
will make it plausible to add parts to TEG for improved effect at a later date

makes TEG graphics not freak out when it is completed with no connection to grid (I will do this separately if this PR gets denied)

THIS is work in progress

# Wiki Documentation

if wiki recommend running TEG only on heaters and coolers it will need to be retconned because that will actually eat power
for the reasons above it may be logical to add this to rouge engi/atmos guy option list

# Changelog

:cl:  
tweak: changed TEG energy calculation so it doesn't break laws of thermodynamics
imageadd: adds new overlay for TEG for when it has no power grid
/:cl:


Xoxeyos (changes requested):
Spell check line 65 in /power/generator please. //Circulators decide how much air moves through

Cartlord:
I'd like to point out that, uh, a lot of shit in this game breaks the laws of physics already - why is the TEG suddenly a concern?

Lostlarry:
> I'd like to point out that, uh, a lot of shit in this game breaks the laws of physics already - why is the TEG suddenly a concern?

really its because I found it and the math looked like thrash
I am pretty sure that it was plausible to run TEG one only heaters coolers which makes no bloody sense

Cartlord:
I've run it well off _T4_ thermomachines, but I don't think I've ever seen people seriously run it with just roundstart heaters & freezers.

nesisknesis (approved):

code/modules/power/generator.dm
```diff
@@ -60,27 +65,36 @@
 		return
 
 	if(powernet)
-		var/datum/gas_mixture/cold_air = cold_circ.return_transfer_air()
+		var/datum/gas_mixture/cold_air = cold_circ.return_transfer_air()		//circulators give us air that is moving trough
```
adamsong:
```suggestion
		var/datum/gas_mixture/cold_air = cold_circ.return_transfer_air()		//circulators give us air that is moving through
```

Hopekz:
> I'd like to point out that, uh, a lot of shit in this game breaks the laws of physics already - why is the TEG suddenly a concern?

You do realize this game is an ongoing volunteer project so those instances only really get fixed here and there when someone feels like it

Cartlord:
Another thing I wanna bring up relavant here (currently being discussed in Discord) is that the TEG is a massive pain in the ass, and running it on a heater is the only set-and-forget method we have; everything else requires either babysitting the damn thing, or funky construction using materials not present in the engine chamber.

patpol4:
Make sure this doesnt break the stock teg setup

JamieD1:
Will review with darth in round after this is merged to dial the numbers in.

ToasterBiome:
Why was this merged with 6 thumbs down and no thumbs up?

PR #11869, opened 2021-07-21
Discourages Venus Human Trap murderbone

Lostlarry:
problem if left unchecked the VHT can just murderbone the station
there is code to make VHT take damage when not near vines but it doesnt work
so i am trying to fix it

TESETED

also whoever made that please test the code next time

# Changelog

:cl:  
bugfix: Venus Human Trap now actualy take damage when not near vines
/:cl:


adamsong:
So not only do venus human traps current not take damage away from the vine, they heal? This is indeed a problem

Lostlarry:
no they dont heal while away from vines
they heal while away from structures

JamieD1 (approved):

PR #11873, opened 2021-07-22
Revert "replaces tintin on the moon starting song with a better remix - Jamie added both instead"

N3D6:
Reverts yogstation13/Yogstation#10850

the remix is really bad and the old one is so much better


Cartlord:
no

JamieD1:
I like Both

PR #11874, opened 2021-07-23
Sets deathsquad name to numbers and matches voice to accomodate

ynot01:

Encrypted mask now sets name to numbers in speech
This change will also make ID say numbers to accomodate
Numbers are 3 numbers pulled from mob tag
image

# Changelog

:cl:  
tweak: Deathsquad are now identified strictly by number tags
/:cl:

### DIFF
--- a/code/modules/clothing/outfits/standard.dm
+++ b/code/modules/clothing/outfits/standard.dm
@@ -378,15 +378,8 @@
 	W.icon_state = "centcom"
 	W.access = get_all_accesses()//They get full station access.
 	W.access += get_centcom_access("Death Commando")//Let's add their alloted CentCom access.
-
-	// Take "mob_###" (from H.tag) and turn it into "###"
-	var/newassignment = ""
-	for(var/i = 1, i <= length(H.tag), i += 1)
-		if(i > 4)
-			newassignment += H.tag[i]
-	W.assignment = newassignment // Unknown's ID Card (###) so DS can identify each other by their IDs
-	
-	W.registered_name = "Unknown"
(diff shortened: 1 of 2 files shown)


adamsong:
Have you considered generating a random number instead?

ynot01:
> Have you considered generating a random number instead?

~~why not have it something directly related to the mob? i considered hashing the number at one point but deemed it too much trouble~~

actually what if i made the voice the mob# and the id name the mob# so people dont have to examine? what do you think of that? might do it tomorrow if u like it

code/modules/clothing/outfits/standard.dm
```diff
@@ -378,8 +378,15 @@
 	W.icon_state = "centcom"
 	W.access = get_all_accesses()//They get full station access.
 	W.access += get_centcom_access("Death Commando")//Let's add their alloted CentCom access.
-	W.assignment = "Death Commando"
-	W.registered_name = H.real_name
+
+	// Take "mob_###" (from H.tag) and turn it into "###"
+	var/newassignment = ""
+	for(var/i = 1, i <= length(H.tag), i += 1)
```
Bibby0110: splittext()

Bibby0110 (dismissed):
Why don't you just make it 3 random numbers

code/modules/clothing/outfits/standard.dm
ynot01: splittext done

adamsong (approved):
Cool idea

JamieD1 (approved):

PR #11875, opened 2021-07-23
Paraplegics can move faster in wheelchairs then their leg-having crewmembers

maxion12345:
# Github Documentation

Paraplegics now have less of a move speed penalty when compared to normal crewmembers, and can use wheelchairs approximately twice as fast. This still isn't very fast, but you won't be waiting forever to move from medbay to cargo. 

This is for the quirk only to prevent people from gaming the system. If you want fast wheelchairs you better pulverize that fucking spine.
# Wiki Documentation

not sure we have quirks in wiki but if we do then paraplegics move faster in wheelchairs

# Changelog
:cl:  
tweak: Paraplegics can now move twice as fast in a wheelchair when compared to normal crewmembers. This isn't very fast still, because wheelchairs are slow as shit, but now anyone choosing the quirk won't be forced to use fire extinguishers to get anywhere in a reasonable amount of time.
/:cl:


maxion12345:
If only we had smooth movement for vehicles

ToasterBiome:
finally a buff

JamieD1 (approved):

PR #11744, opened 2021-07-10
Adds digitigrade jumpsuit alts using ThatGuyPhil's icons

nmajask:
# Github documenting your Pull Request

While looking through some of the old PR's, I stumbled into "[uniform sprites for digitigrade lizardpeople #10931](https://github.com/yogstation13/Yogstation/pull/10931)" by [ThatGuyPhil](https://github.com/ThatGuyPhil). So I readded it and made it kinda work.

I have updated and added a few new icons that weren't in the original PR, but there are still **many** uniforms that need digi alts (Any that are in yogstation/icons/mob/ have not been touched). I will try to continue adding/updating important ones and will add any given to me of they work. The code is not very good, but it does function.

Due to the scope of this PR, I have done very limited testing and there are likely many issues that I have not found.

Here is an example of it working (It was taken before I fixed some stuff):
digicap

Here are some newer image:
digiminer

digisec

I have done some more testing and have tried some things that I thought would break the code. Luckily I found much fewer errors than I thought I would, for example I thought wearing a hardsuit or wearing a uniform casually would cause major issue but they didn't. The current list of issues I have found are:

- Suits still don't have digi icons or implementation, so hardsuits or any other suit that covers the wearers legs will squish them into normal legs
- Some icons have icons that look weird when wearing footwraps (medical jumpsuit is a good example
- In order to allow a uniform to use digi sprites, you need to redefine 2-3 vars

Jobs That Still Need Digi Uniform Sprites:
- Maybe Psych

Job Uniforms That Have Bad or No Implementation of Digi Sprites:
- Artist (not working)
- Mining Medic (bad sprites)

# Wiki Documentation

As far as I am aware, this shouldn't require changing anything in the wiki, but I could be wrong.

# Changelog

:cl:  
imageadd: Readded icons from ThatGuyPhil's "uniform sprites for digitigrade lizardpeople" #11017 and added/updated some
experimental: Added support for digitigrade uniforms
/:cl:

PR #11528, opened 2021-05-28
Gives Mining Medics job specific traitor items

nmajask:
# General Documentation

### Intent of your Pull Request

Gives Mining Medic tators access to the Kinetic Accelerator Pressure Mod, Brainwashing Surgery Program, and Arm Mounted Medical Beamgun as job specific items, as they currently do not have any. (sorry about the 12 commits, I have no clue how to use GitHub)

### Why is this change good for the game?

Makes the game more consistent with job specific items and allows people to maybe do some cool things as a mining medic traitor.

# Wiki Documentation

### Briefly describe your PR and the impacts of it, in layman's terms. 
This will be the basis of the Wiki entry for your PR, and more information / detail is better for Wiki editors to integrate.

Gives mining medic traitors access to the Kinetic Accelerator Pressure Mod, Brainwashing Surgery Program, and Arm Mounted Medical Beamgun as job specific items, as they currently do not have any.

### What should players be aware of when it comes to the changes your PR is implementing?

Mining Medics can now brainwash you using surgery

### What general grouping does this PR fall under? 
For example, Engineering changes, Medical rebalancing, TEG tweaks, etc.?
Uplink tweaks
### Are there any aspects of the PR that you would like us not to mention on the Wiki?
nope
### If there are any numerical values involved in your PR that will be relevant to a player, please note them here. 
no

# Changelog

Edit this changelog for changes that are noticeable by the players. Remove it if this isn't the case. If you add a name after the ':cl', that name will be used in the changelog. Leave it empty to use your GitHub name. Prefix the PR title with [admin] if it's something admin related. Prefix the PR title with [s] if you are fixing an exploit so it is not announced on discord and the server.

:cl:    
tweak: Gave Mining Medics job specific traitor items
  
/:cl:

PR #11884, opened 2021-07-24
Fixes "Adds digitigrade jumpsuit alts using ThatGuyPhil's icons #11744"

nmajask:

So there were a few is issues with PR #11744. For example, the digitigrade alts only worked on a few jumpsuits, footwraps wouldnt look good with them, and it was coded very poorly, this PR attempts to fix many of the issues as well as some minor stuff in my other PRs.

I removed the digiversion and digiadjusted vars from any file they appear in, as they only serve to do the same thing as mutantrace_variation. The main reason I had these was so you could have a digialt for a jumpsuit but not have it break if you dont have one for the adjusted version, but this would result in you having to define 3 vars to adding alts to jumpsuits.

I added alts for a ton of other jumpsuits, almost all of which were functional in my test server when I created the PR but didnt make it in because I forgot to upload the files. Now all job uniforms should have digialt. Also added digialts for the explorer suit and wardens jacket and changed the icons for footwraps so that they actually cover digitigrade feet.

Finally, I cleaned ups some of the stuff from #11528 and #11542. They are just unused icons and commented out things.

# Wiki Documentation

I dont think anything on the wiki will be affected by this as its mostly code stuff, but I could be wrong.

# Changelog
:cl:  
rscadd: Added many more digitigrade jumpsuit alts and two suit ones
tweak: tweaked the footwrap icons so they cover digitigrade feet
tweak: cleaned some code
/:cl:


nesisknesis (approved):
can't go a week without jamie merging unverified art prs

PR #11900, opened 2021-07-27
Fixes space adap so it prevents passive heating as well as passive cooling

adamsong:

Walking through a hot room is no longer a death sentence with space adap.

# Wiki Documentation

Space adap not prevents heating from hot air

# Changelog

:cl:  
tweak: tweaked space adaptation to make it not kill you from being in a slightly warm room
/:cl:


alexkar598:
Aka turning a nerf into a buff by making it *almost* entirely immune to plasmafloods

Mqiib:
> Aka turning a nerf into a buff by making it _almost_ entirely immune to plasmafloods

Insulation goes both ways. The fact that people were heating up a bunch from passing through a slightly warm room and then never cooling off once they entered a normal-temperature room is dumb. Perhaps the better solution was to just drastically decrease the passive heating/cooling rate instead of outright making it not work at all?

JamieD1 (approved):

PR #11904, opened 2021-07-28
nerfs rezadone and buffs cloneoxadone, makes strange reagent heal all remaining physical damage but cause heavy clone damage

SomeguyManperson:

strange reagent is objectively better than other revival methods being:
irritating to get
easy to mass produce
easy to use
without a time limit (actually heals MORE damage when used later since it also fully heals organs and resets the defib timeout)
void of actual downsides

so I've decided to change that, changed chems:
Strange Reagent: will now only work if the body has less than 100 cumulative brute and burn (people already heal to full so this likely won't effect anyone), using strange reagent will heal the remaining brute and burn damage, as well as the 20 oxygen and toxin damage, but cause 80 cellular damage minus any cellular, oxygen or toxin damage, making it slightly harder to use effectively without other chemicals. This means instead of carrying around 200 patches of rezadone you have to carry around 100 patches of rezadone and 100 patches of cell damage healing chemicalss
Rezadone: will no longer heal all cellular damage with a single reagent metabolism cycle, instead healing 10 points per cycle.
Clonexadone: is now far more viable at healing clone damage, now healing up to 100 based on temperature up from 6 but does require you to be sleeping to work

# Wiki Documentation

Strange Reagent: will now only work if the body has less than 100 cumulative brute and burn (people already heal to full so this likely won't effect anyone), using strange reagent will heal the remaining brute and burn damage, as well as the 20 oxygen and toxin damage, but cause 80 cellular damage minus any cellular, oxygen or toxin damage.
Rezadone: will no longer heal all cellular damage with a single reagent metabolism cycle, instead healing 10 points per cycle.
Clonexadone: heals up to 100 cloneloss based on temperature up from 6 but does require you to be sleeping to work

# Changelog

:cl:  Coder Gun
tweak: strange reagent now deals heavy cellular damage on effective use
rezadone: now heals 10 cloneloss per reagent cycle instead of all at once
clonexadone: now heals up to 100 cloneloss based on temperature up from 6 but does require you to be sleeping to work
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -1,5 +1,5 @@
 #define PERF_BASE_DAMAGE		0.5
-#define REAGENT_REVIVE_DAMAGE ((HEALTH_THRESHOLD_CRIT + HEALTH_THRESHOLD_DEAD) * 1.2)
+#define REAGENT_REVIVE_DAMAGE (HEALTH_THRESHOLD_CRIT + 20)
 
 /////////////////////////////////////////////////////////////////////////////////////////
 					// MEDICINE REAGENTS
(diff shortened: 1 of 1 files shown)


Anvilman6:
Let's fucking go

JamieD1 (approved):

PR #11909, opened 2021-07-29
Reduce the weight of asimov and crewsimov

adamsong:

Reduces the spawn weight of crewsimov and asimov, as they are very similar laws. Currently an AI will get one 57% of the time, this reduces that to 40%, the same as CEO.

# Changelog

:cl:  
tweak: tweaked weight of asimov and crewsimov
/:cl:

### DIFF
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -420,8 +420,8 @@ RANDOM_LAWS reporter
 LAW_WEIGHT custom,0
 
 ## standard-ish laws. These are fairly ok to run
-LAW_WEIGHT asimov,6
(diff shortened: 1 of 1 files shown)


maxion12345:
we really just need more lawsets :(

adamsong:
See other PRs

Moja10:
The point is that they are so common because they are the most sane/workable of the roundstart lawsets, its a good thing they appear so much

JamieD1:
This is on purpose as we used to have issues where ais used to get validhunting laws every round.

PR #11922, opened 2021-07-29
assorted buffs to flying fang

SomeguyManperson:
considering the lack of ranged or stun weapons, I may have tried to keep the attacks a bit too "fair"
Shields no longer hardcounter the pounce attack, instead removing the stun they cause (still causes a knockdown and immobilize)
increase brute damage for tail slap combo's steps from 8 each to 15 and 10
increase bleeding caused by neck bite to 10 from 5, higher bleeding is always bad
increase healing from neck bite to 25 brute and 25 burn up from 15 and 8 and allows it to heal 50 stamina damage, the point of the kit is sustainability over versatility iirc so that's what you get
hardsets the damage of the baseline melee to 12, AKA a toolbox's worth of force

# Wiki Documentation

this isnt on the wiki lmao

# Changelog

:cl:  
tweak: increased several damage values and healing amounts in the flying fang martial art
tweak: flying fang pounces no longer backfire if blocked, instead they will just not hardstun the target, slamming into a window like a doofus still gets you hardstunned.
/:cl:

### DIFF
--- a/code/datums/martial/flying_fang.dm
+++ b/code/datums/martial/flying_fang.dm
@@ -215,7 +215,7 @@
 				if(H.check_shields(src, 0, "[A]", attack_type = LEAP_ATTACK))
 					blocked = TRUE
 			L.visible_message("<span class ='danger'>[A] pounces on [L]!</span>", "<span class ='userdanger'>[A] pounces on you!</span>")
-			L.Paralyze(blocked ? 6 SECONDS)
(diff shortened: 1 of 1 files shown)


Marmio64:
only thing i dont understand is the shield change. Shields are always supposed to counter melee, martial arts, and leaps of any kind, but this can just leap through the shield and knock them down for 10 seconds anyways?

JamieD1 (approved):

PR #11929, opened 2021-07-31
gives science research points for augments

SomeguyManperson:

you now get 150 points for finishing an augmentation surgery, with 6 bodyparts to augment this is worth 900 points per person
using this on a braindead or catatonic will only give 15 points per part

# Wiki Documentation

augmentation surgery gives 150 points on success, 15 points if the target isn't an active player

# Changelog

:cl:  
tweak: you now get research points for doing augmentation surgeries 
/:cl:


Bluedazzled:
i like this concept, could there be other things that give points?

maxion12345:
> i like this concept, could there be other things that give points?

what about making things with slimes?
Nanites can already give both nanite research and generate normal research with programs aswell, robotics and xenobio is mostly the only two places that don't make points

JamieD1 (approved):

PR #11933, opened 2021-08-01
[admin] Added a command to fix air on the whole zlevel, disabling atmos to do so safely

adamsong:

Does what is says on the tin, fixes air for every turn on the zlevel. Turns off atmos if its on, then turns it back on at the end if it was on. Prevents atmos being strange mid fix.

# Changelog

:cl:  
rscadd: Added fix air zlevel command.
/:cl:

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -20,6 +20,7 @@ GLOBAL_PROTECT(admin_verbs_default)
 	/client/proc/cmd_admin_pm_panel,		/*admin-pm list*/
 	/client/proc/stop_sounds,
 	/client/proc/fix_air, // yogs - fix air verb
+	/client/proc/fix_air,
 	/client/proc/debugstatpanel
 	)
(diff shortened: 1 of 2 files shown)


alexkar598 (changes requested):
- Dont use block() on such large areas (probably using a for loop and locate(x,y,z))
- Use CHECK_TICK in expensive loops to not hang the server instantly (potentially causing a restart if it goes missing long enough to get killed)
- Use a typeless loop
- Don't copy air but instead immediatly call F.parse_gas_string with the initial gas mix

If you wanna do something, do it right

adamsong:
> Dont use block() on such large areas (probably using a for loop and locate(x,y,z))

Copied from smooth_zlevel, will look into fixing

> Use CHECK_TICK in expensive loops to not hang the server instantly 

Ok, will do

> Use a typeless loop

Copied from fix_air, will look into fixing

> Don't copy air but instead immediatly call F.parse_gas_string with the initial gas mix

Copied from fix_air, will look into fixing

adamsong:
#11934 is better

PR #11934, opened 2021-08-01
[admin] Added a command to fix air on the whole zlevel, disabling atmos to do so safely - Alexkar598 better code edition

alexkar598:
Closes #11933 

Github documenting your Pull Request

Does what is says on the tin, fixes air for every turn on the zlevel. Turns off atmos if its on, then turns it back on at the end if it was on. Prevents atmos being strange mid fix.
Changelog

🆑 adamsong: Original Implementation | alexkar598: Improved Implementation
rscadd: Added fix air zlevel command.
/🆑

# Differences between the prs
Code is somewhat less readable here but theres comments
Code is faster
Code is actually accessible because adam added the wrong verb to the verb list(untested code mayhaps)
Verb isn't on right click menu but instead in the misc tab
Theres a confirmation box to prevent screwing over atmos techs by accident
Removes lag warning because this code takes 200ms to run, virtually impossible to notice

Other PR: 

This PR:

For those who don't know how to read profiler, self cpu is the time byond spent on this proc, total cpu is the time byond spent on this proc and all procs it called and real time is the IRL time byond spent(including confirmation boxes, sleeping procs such as stoplag() and other funny things)

### DIFF
--- a/yogstation/code/modules/admin/verbs/fix_air.dm
+++ b/yogstation/code/modules/admin/verbs/fix_air.dm
@@ -19,33 +19,40 @@
 			F.copy_air(GM)
 			F.update_visuals()
 
-/client/proc/fix_air_z(var/turf/open/T in world)
-	set name = "Fix Air, Z-level"
+/client/proc/fix_air_z()
+	set name = "Fix Air on current Z-level"
 	set category = "Misc.Unused"
-	set desc = "Fixes air on the entire z-level, warning, laggy and temporarily disables atmos"
+	set desc = "Fixes air on the entire z-level, temporarily disables atmos"
 
 	if(!holder)
 		to_chat(src, "Only administrators may use this command.", confidential=TRUE)
 		return
 	if(!check_rights(R_ADMIN,1))
 		return
-	message_admins("[key_name_admin(usr)] fixed air on zlevel [T.z]")
-	log_game("[key_name_admin(usr)] fixed airon zlevel [T.z]")
+
+	if(alert("Do you want to reset air on the entire z level?", "Fix Air- Z level", "No", "Yes") != "Yes")
+		return
+
+	message_admins("[key_name_admin(usr)] fixed air on zlevel [mob.z]")
+	log_game("[key_name_admin(usr)] fixed air on zlevel [mob.z]")
 
 	var/atmos_enabled = SSair.can_fire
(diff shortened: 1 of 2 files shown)


yogstation/code/modules/admin/verbs/fix_air.dm
```diff
@@ -18,3 +18,41 @@
...
+
+/client/proc/fix_air_z()
+	set name = "Fix Air on current Z-level"
+	set category = "Misc.Unused"
+	set desc = "Fixes air on the entire z-level, temporarily disables atmos"
+
+	if(!holder)
+		to_chat(src, "Only administrators may use this command.", confidential=TRUE)
+		return
+	if(!check_rights(R_ADMIN,1))
```
Bibby0110:
```suggestion
	if(!check_rights(R_ADMIN, TRUE))
```

yogstation/code/modules/admin/verbs/fix_air.dm
```diff
@@ -18,3 +18,41 @@
...
+		return
+
+	if(alert("Do you want to reset air on the entire z level?", "Fix Air- Z level", "No", "Yes") != "Yes")
```
Bibby0110, line 33: Maybe you could add the z-level they're on to the prompt for those special admins :)

Bibby0110 (commented):
👍

yogstation/code/modules/admin/verbs/fix_air.dm
alexkar598: No, its called natural selection if you get removed for thinking youre fixing air on CC when youre clearly on the station

yogstation/code/modules/admin/verbs/fix_air.dm
alexkar598: The default value for the second argument is already a truthy value so I just removed it

yogstation/code/modules/admin/verbs/fix_air.dm
adamsong: I agree with alexkar

Bibby0110 (approved):
Code is ok I guess

patpol4:
Is it possible to do this for shuttle/map template loading aswell? That would fix the vaccuum shuttles problem

PR #11944, opened 2021-08-03
No longer resets account to zero when adjusting money while it is negative

adamsong:

Currently if an admin were to reduce an account below zero for bus reasons, the next time that account got money it would be reset to zero. This removes that check, so if a negative account receives money it will just be less negative. A different proc already checks to see if the purchase can be afforded if money is being withdrawn, so this will not break that.

# Changelog

:cl:  
tweak: tweaked economy to properly support negative balances
/:cl:

### DIFF
--- a/code/modules/economy/account.dm
+++ b/code/modules/economy/account.dm
@@ -32,9 +32,7 @@
 
 /datum/bank_account/proc/_adjust_money(amt)
 	account_balance += amt
-	if(account_balance < 0)
-		account_balance = 0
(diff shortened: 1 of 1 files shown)


Bibby0110:
I don’t see why you would ever do this

Bibby0110 (approved):

adamsong:
> 
> 
> I don’t see why you would ever do this

Adminbus

JamieD1 (approved):

PR #11948, opened 2021-08-03
Pound dirt you filthy ban evader/banned computer sharer

Bibby0110:


adamsong:
And if the banned account was banned to prevent multikeying?

Bibby0110:
> And if the banned account was banned to prevent multikeying?

tough luck, if this is triggered you've already multikeyed

adamsong:
If we get someone on the discord saying "Hey, I no longer login with foo, I now use bar," we will ban foo, and give bar all their notes to bar. This is fine and allowed, but not under this new PR

Bibby0110:
> If we get someone on the discord saying "Hey, I no longer login with foo, I now use bar," we will ban foo, and give bar all their notes to bar. This is fine and allowed, but not under this new PR

I'm well aware of how you do things..

Solution is to tell them to pound sand and stick to their original ckey :)

num1bamf:
Why don’t you come up with a better sticky ban rather than this that no one asked for.

Bibby0110:
> Why don’t you come up with a better sticky ban rather than this that no one asked for.

Sticky ban has nothing to do with this

Moggram:
>
> 
> Solution is to tell them to pound sand and stick to their original ckey :)

And what if they lost their original account?

Bibby0110:
> > Solution is to tell them to pound sand and stick to their original ckey :)
> 
> And what if they lost their original account?

If you manage to lose your email and thereby lose access to your account that is 100% on you

PR #11962, opened 2021-08-06
[admin] Hopefully makes admin ranks able to have spaces

adamsong:

Makes admins ranks able to have spaces, this is not 1998, spaces are not that hard

# Changelog

:cl:  
rscadd: Added spaces into admin ranks 
/:cl:


adamsong:
Tested, works, will try to make it so the config doesn't have to be ugly

adamsong:
Database will need updated when this change is made to put spaces in the roles, but after that it should be fine

Bibby0110 (approved):

JamieD1 (approved):

PR #11967, opened 2021-08-07
admins but they get to make voice announcements too

Bibby0110:
?

### DIFF
--- a/code/game/machinery/computer/communications.dm
+++ b/code/game/machinery/computer/communications.dm
@@ -162,7 +162,7 @@
 				return
 			make_announcement(usr)
 		if ("makeVoiceAnnouncement")
-			if (!authenticated_as_non_silicon_captain(usr))
+			if (!authenticated_as_non_silicon_captain(usr) || !IsAdminGhost(usr))
 				return
 			make_voice_announcement(usr)
 		if ("messageAssociates")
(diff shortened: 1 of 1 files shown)


alexkar598:
Ah yes I love to hear admin voice while getting abused with supply pods

code/game/machinery/computer/communications.dm
```diff
@@ -162,7 +162,7 @@
 				return
 			make_announcement(usr)
 		if ("makeVoiceAnnouncement")
-			if (!authenticated_as_non_silicon_captain(usr))
+			if (!authenticated_as_non_silicon_captain(usr) || !IsAdminGhost(usr))
```
adamsong, line 165: Not necessary, admin ghosts authenticate as non silicon captains.

code/game/machinery/computer/communications.dm
```diff
@@ -390,7 +390,7 @@
 
 					data["alertLevelTick"] = alert_level_tick
 					data["canMakeAnnouncement"] = TRUE
-					data["canMakeVoiceAnnouncement"] = ishuman(user)
+					data["canMakeVoiceAnnouncement"] = ishuman(user) || IsAdminGhost(user)
```
adamsong, line 393: Why not move this up to the authenticated_as_non_silicon_captain block, and set it to true?

adamsong (changes requested):
See two previous comments

PR #11983, opened 2021-08-09
I fix the spare cabinet access without moving it somewhere dumb

maxion12345:
# If this is your first PR, or not, take the time to read our CONTRIBUTING.md file! You can see it here: https://github.com/yogstation13/Yogstation/blob/master/.github/CONTRIBUTING.md

i change something

# Wiki Documentation

no

# Changelog

:cl:  
bugfix: Spare cabinet uses normal captain access now
/:cl:


adamsong:
> `req_access = list(20)`

Are you sure this works in map files? `req_access_text` exists presumably because that doesn't work

maxion12345:
> > `req_access = list(20)`
> 
> Are you sure this works in map files? `req_access_text` exists presumably because that doesn't work

Its the same access as the captain's door (20)

adamsong:
But cap's door uses the `req_access_text` variable, I'm not sure this can initialize a list

patpol4:
you need to use the text one

patpol4:
i only skimmed this but #11984 seems to do the same thing but with fireaxe too

JamieD1:
Bad, do it in code initialize instead.

JamieD1:
Code change better

PR #11998, opened 2021-08-10
Reduces the priority of the explosion subsystem so mobs still process while meteors are overflowing everything else

SomeguyManperson:

This may be a bad idea but it also may be an extremely good idea

# Changelog
:cl:  
tweak: stuns will no longer last 200 years when there's meteors beating up the station probably
/:cl:


JamieD1:
This might make nukes lag even more..

Have you tested this with a nuke?

Bibby0110:
> This might make nukes lag even more..
> 
> Have you tested this with a nuke?

Better way to think about it is that this will make the nuke lag *first*

JamieD1 (approved):

PR #12004, opened 2021-08-10
Art should not own code files

adamsong:

Why does art own the ticker subsystem? 99% of that file is code that does not involve art at all.


Bibby0110:
Art controls lobby songs + holidays

JamieD1:
Resolved

PR #12014, opened 2021-08-12
Adds Custom Mentor role, can Add Wiki and Stuff now

JamieD1:
Adds Custom Mentor role, can Add Wiki and Stuff now


alexkar598 (changes requested):
Update SQL schema

alexkar598:
For the record, storing the rank as a string across all rows is a stupid idea

adamsong:
How would you recommend storing it?

alexkar598:
As an id which points to a rank

adamsong:
Admin ranks use a string

alexkar598:
Whoever came up with that idea doesn't know how a relational database works

Bibby0110:
> Whoever came up with that idea doesn't know how a relational database works

our codebase in a nutshell

SQL/database_changelog.txt
```diff
@@ -1,13 +1,21 @@
...
 
-INSERT INTO `schema_revision` (`major`, `minor`) VALUES (5, 4);
+INSERT INTO `schema_revision` (`major`, `minor`) VALUES (5, 5);
 or
-INSERT INTO `SS13_schema_revision` (`major`, `minor`) VALUES (5, 4);
+INSERT INTO `SS13_schema_revision` (`major`, `minor`) VALUES (5, 5);
 
 In any query remember to add a prefix to the table names if you use one.
 
+version 5.5 14 August 2021, by JamieD1
```
Bibby0110, line 13: add 
 \---------------------------------------------------- 
above

N3D6:
Maybe needs a bug report but right now rementoring after dementoring makes your role show up as [] when you talk in OOC

PR #12022, opened 2021-08-14
[EXPERIMENTAL] People not readied up now count as half a readied up player. 

Bibby0110:

Will allow rounds with the population to run them, but not the readied up people ready. We're expecting around 25-50% of the un-readied players to midround join. Does not count admins nor people readied up to observe.

- [ ] Remove logging

# Wiki Documentation

# Changelog

:cl:  
experimental: For gamemode population calculations an unreadied player now counts as 50% of a readied one.
/:cl:

### DIFF
--- a/code/__DEFINES/antagonists.dm
+++ b/code/__DEFINES/antagonists.dm
@@ -1,3 +1,6 @@
+///An unreadied player counts for this much compared to a readied one
+#define UNREADIED_PLAYER_MULTIPLIER 0.5
+
 #define NUKE_RESULT_FLUKE 0
 #define NUKE_RESULT_NUKE_WIN 1
 #define NUKE_RESULT_CREW_WIN 2
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -67,20 +67,24 @@
 ///Checks to see if the game can be setup and ran with the current number of players or whatnot.
 /datum/game_mode/proc/can_start()
 	var/playerC = 0
+	var/unreadiedPlayers = 0
 	for(var/mob/dead/new_player/player in GLOB.player_list)
-		if((player.client)&&(player.ready == PLAYER_READY_TO_PLAY))
+		if(player.client && (player.ready == PLAYER_READY_TO_PLAY))
 			playerC++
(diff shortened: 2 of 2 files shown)


Bibby0110:
Needs tweaking, probably would be better between somewhere between 25% and 50%

Hopekz:
What happens if not ready players are actually not ready since, with this PR, you force a game mode that ready people can't play because of the population.

Bibby0110:
> What happens if not ready players are actually not ready since, with this PR, you force a game mode that ready people can't play because of the population.

Most of the gamemodes function fine with less people, they're just kind of empty. This aims to assume that some of the people not readied up will actually play the game.
It has a chance to go wrong of course, but in general it should mean mid-pop will avoid the "traitor-because-we-don't-have-enough-readied-people-for-anything-else" cycle

Anvilman6:
I have my qualms about this. It seems like having un-readied people pad out the server count and trigger potentially higher population round configurations might cause issues as Rob noted, such as them just not joining.

Perhaps we can just tweak mode cutoffs and weight factors to create more diversity instead of going at it from this approach?

Bibby0110:
> I have my qualms about this. It seems like having un-readied people pad out the server count and trigger potentially higher population round configurations might cause issues as Rob noted, such as them just not joining.
> 
> Perhaps we can just tweak mode cutoffs and weight factors to create more diversity instead of going at it from this approach?

None of these things are mutually exclusive with this PR.
This PR is made exclusively due to the fact that when a round starts the chat will usually get spammed with people joining in right after, skewing the actual player count. Obviously the best solution would be delaying the selection of a gamemode for a few minutes, but that isn't feasible.

ynot01:
Can you make it so people observing count as 0%? and only those in lobby count as 50%

adamsong:
That is what this code does

TheLinuxOS:
I asked for this for so long thank you bibby for finally doing something good

Bibby0110:
> Can you make it so people observing count as 0%? and only those in lobby count as 50%

Yee already takes this into account. Readying up to observe is 0%, only not being ready counts as 50%
Admins are assumed to be observing at all times, (Unless readied up to play)

JamieD1 (approved):

PR #12032, opened 2021-08-15
more EngiVend RCDS

ynot01:

EngiVend RCDs are now unloaded but more plentiful, there are now compressed matter cartridges in it to compensate
RCDs should be plentiful for engineers, the only other way to get them is by hacking an autolathe which from a corporation standpoint should never be a necessity

# Wiki Documentation

EngiVend RCDs are unloaded but there are more RCDs now
EngiVend now carries 12 compressed matter cartridges

# Changelog

:cl:  
tweak: Unloaded the RCDs in EngiVend, but added more
/:cl:


adamsong:
Its already really easy to get an RCD, and refilling them is easy

JamieD1:
Considering the power of the RCDs I think having them even more easier to get is a bit crap.

Considering they used be an antag objective as well.

How about 5 unloaded with 12 ammo?

alexkar598:
> 
> 
> Considering the power of the RCDs I think having them even more easier to get is a bit crap.
> 
> Considering they used be an antag objective as well.
> 
> How about 5 unloaded with 12 ammo?

RCDs are powerful but the spare ID isn't?

adamsong:
> RCDs are powerful but the spare ID isn't?

This PR has nothing to do with the spare ID

ynot01:
Posting my discord comments:

RCDs are integral to engi gameplay
i wont disagree that they're very powerful
but an engi round without an RCD is a pretty shitty one

you can print them
but only through a hacked autolathe
and engies dont get an autolathe to begin with

ill change the PR to have 5
as it solves my present issue at least (only 5 engies can spawn roundstart)
maybe some time in the future we can make something to make RCDs obtainable outside of hacking and engivend crates

code/modules/vending/engivend.dm
```diff
@@ -7,7 +7,8 @@
 	products = list(/obj/item/clothing/glasses/meson/engine = 2,
 					/obj/item/clothing/glasses/welding = 3,
 					/obj/item/multitool = 4,
-					/obj/item/construction/rcd/loaded = 3,
+					/obj/item/construction/rcd = 5,
+					/obj/item/rcd_ammo = 12,
```
Hopekz:
```suggestion
					/obj/item/rcd_ammo = 3,
```
FYI note:
If #12057 goes through 3 will be the equivalent to the current 12

Bibby0110 (approved):

PR #12034, opened 2021-08-15
Syndicate mask update

KillerOrcaCora:
This makes the object better match how it actually looks and works on your spessman while your wearing it.

Tested it

Wiki may have to change their picture into a gif

:cl:  

imageadd: Modified and animated the syndicate mask

/:cl:

THE MASK INGAME>>> https://streamable.com/94ktyq


KillerOrcaCora:
Reminder the corners have been updated to sit for a second so animation is not so insanely fast.

nesisknesis (changes requested):
syndicateREDORCA.dmi needs to be deleted

nesisknesis (approved):

PR #12040, opened 2021-08-15
Disallows map voting if a map has already been selected for next round

adamsong:

Prevents the classic
Random map rotation has selected "not boxstation" for the next round
Map vote
Boxstation wins

# Changelog

:cl:  
tweak: Map vote no longer works if a map has been selected
/:cl:


Hopekz:
You do realize boxstation wins votes because most people that are playing want to play boxstation, right? If that wasn't the case another map would be chosen. The votes are even rigged against boxstation because of voteweight and it still is the preferred map out of all the options according to popular vote performed every round.
 
Having a really good map that people go out of their way to play isn't a problem. If anything we should be looking to remove the ones that people don't want to play from rotation but that is probably extreme as well.

Bibby0110 (approved):

PR #12052, opened 2021-08-16
 Splits space adapt into two mutations, space adaptation and heat adaptation

alexkar598:

Splits space adapt into two mutations, space adaptation and heat adaptation
Heat adapt: protection from heat and high pressure but no heating up passively 

# Wiki Documentation
New mutation: Heat adaptation, immunity from heat and high pressure at the cost of not being able to heat up passively. Mutually exclusive with the space adaptation mutation

:cl:  
rscadd: Splits space adaptation into two mutations: Space adaptation and Heat adaptation.
/:cl:


alexkar598:
Being immune to space AND immune to plasmafires at nearly no downside is a bit fucking OP hence the split

Hopekz:
Could probably make a new name for space adaptation at this point.
Something that doesn't convey that you'll be able to live in space like "space adaptation"

alexkar598:
Except youre able to live in space, space isn't hot, its cold.

ApplePlastic:
i like this, +1 from me

patpol4 (approved):
Map fine

nesisknesis (approved):

code/datums/mutations/heat_adaptation.dm
```diff
@@ -0,0 +1,33 @@
+//Heat Resistance gives your entire body a blue halo, and makes you immune to the effects of a plasmafire minus the fire part.
+/datum/mutation/human/heat_adaptation
+	name = "Heat Adaptation"
+	desc = "A strange mutation that renders the host immune to the heat of a plasma fire. Does not grant immunity to fire itself."
```
Hopekz, line 4:
```suggestion
	desc = "A strange mutation that renders the host immune to the heat of plasma fire. Does not grant immunity to fire itself."
```

alexkar598: ???

TheLinuxOS:
Does this make you immune to heat up to the Max Temp Protect value or completely immune? cause if it’s completely immune then it should be changed to up to max temp protect

Bibby0110 (approved):

Hopekz:
> it should be changed to up to max temp protect

"max temp protect" doesn't really roll off the tongue, in my opinion, so I like the old name better vs. this suggestion.

adamsong:
This isn't a name change, this is saying 'if over x degrees, you die anyway' because right now heat adap would let you stand in a fusion chamber

PR #12057, opened 2021-08-17
RCD cartridges now refill the RCD in a single cartridge vs having to use 4

Hopekz:
image

The RCD cartridges suck compared to just grabbing mats which is weird because the cartridge item is meant to reload the RCD almost as if the RCD cartridges were coded first before you could add mats to the RCD.

The RCD cartridges now refill the RCD in a single unit vs before where you needed to put in 4 separate RCD ammo cartridges into the RCD.

The large RCD cartridge doesn't even have its own sprite so I think it just makes more sense to exist as a single item that refills the base level RCD. 
The large one seemed to only really exist for the sake of being put in an ERT engineer's pocket so this standardizes it as well.

Does not conflict with #12032

:cl:  Hopek
tweak: RCD cartridges now refill the standard RCD in a single cartridge vs having to use 4.
/:cl:

### DIFF
--- a/_maps/map_files/debug/multiz.dmm
+++ b/_maps/map_files/debug/multiz.dmm
@@ -579,9 +579,9 @@
 /area/bridge)
 "bL" = (
 /obj/structure/table,
-/obj/item/rcd_ammo/large,
-/obj/item/rcd_ammo/large,
-/obj/item/rcd_ammo/large,
+/obj/item/rcd_ammo,
+/obj/item/rcd_ammo,
+/obj/item/rcd_ammo,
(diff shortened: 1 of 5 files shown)


patpol4 (approved):
Map fine, shouldnt cause any conflicts too

Hopekz:
@Xoxeyos What do you not like about this change?

JamieD1:
> @Xoxeyos What do you not like about this change?

Can you list all the ways to get matter cartridges

Bibby0110 (approved):

PR #9147, opened 2020-07-06
The belt laid schemes of mice and men often go awry. (Belt tweak PR)

MrHorizons:
Blame the bad PR title on [pungenerator.org](https://pungenerator.org).

A semi-port of [BeeStation #1349](https://github.com/BeeStation/BeeStation-Hornet/pull/1349), with some elements adapted to fit the lack of plumbing in Yog. Unlike my other PRs, the changelog is fairly comprehensive.

#### Changelog

:cl:  
rscadd: RCDs can now deconstruct airlock assemblies and firelock frames.
rscadd: Biobags can carry organs, toolbelts can carry RCDs, RPDs, and inducers, and grenadier belts can carry combustible lemons and cherrybombs.
tweak: Medical belts and toolbelts can hold greater total weight.
tweak: RCDs and RPDs can be worn on the belt slot.
tweak: Liver, lungs, and stomach organs are small.
/:cl:

PR #12058, opened 2021-08-17
Revert "#9147 (Belt tweaks) Hotfix"

Hopekz:
Reverts the PR that turned all storage belts into belts of holding.
It keeps the weight class change which was reasonable 
**This is not affecting the storage size, only the amount of maximum slots.**

**EDIT:**
-------
My PR is mainly fighting a shitty hotfix applied to all storage with special slots in the game.

The main arguments brought up so far are about the engineering belt, but there are like 12+ storage items affected by this.
Most of them were unintended because of the lazy hotfix
The correct way to increase the slots higher for a belt would be to go to the belt you want to increase, THEN changing the number instead of what the hotfix did of modifying the parent

:cl:  Hopek
tweak: Specialized belts are back to not automatically starting with 14 slots anymore. This is not affecting the storage size, only the amount of maximum slots.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -38,6 +38,7 @@
 /obj/item/storage/belt/utility/ComponentInitialize()
 	. = ..()
 	var/datum/component/storage/STR = GetComponent(/datum/component/storage)
+	STR.max_w_class = WEIGHT_CLASS_NORMAL
 	STR.max_combined_w_class = 21
 	STR.set_holdable(list(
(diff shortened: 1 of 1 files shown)


Mqiib:
n o

Hopekz:
> n o

NoooOoooo

Anvilman6:
I would support this if you could outline your plans to apply tweaks to refactor the storage capacity of specific belts, rather than just peeling off the parent modifier.

Bibby0110 (approved):

PR #12071, opened 2021-08-17
Makes event spawning based on alert level

adamsong:

Each event can now specify the minimum and maximum security level it can activate on. Default is green through red alert, only override right now is allowing the aurora to activate on any alert level.

# Changelog

:cl:  
tweak: tweaked events to only spawn on some alert levels
/:cl:

### DIFF
--- a/code/modules/events/_event.dm
+++ b/code/modules/events/_event.dm
@@ -29,6 +29,8 @@
 
 	var/triggering	//admin cancellation
 
+	var/max_alert = SEC_LEVEL_RED /// Highest alert level the event will trigger at
+
 	/// Whether or not dynamic should hijack this event
 	var/dynamic_should_hijack = FALSE
 
@@ -58,6 +60,8 @@
 	if(holidayID && (!SSevents.holidays || !SSevents.holidays[holidayID]))
 		return FALSE
 	if(ispath(typepath, /datum/round_event/ghost_role) && !(GLOB.ghost_role_flags & GHOSTROLE_MIDROUND_EVENT))
+		return FALSE
+	if(GLOB.security_level > max_alert)
 		return FALSE
 
 	var/datum/game_mode/dynamic/dynamic = SSticker.mode
--- a/code/modules/events/_event.dm
+++ b/code/modules/events/_event.dm
@@ -30,6 +30,7 @@
 	var/triggering	//admin cancellation
 
 	var/max_alert = SEC_LEVEL_RED /// Highest alert level the event will trigger at
+	var/min_alert = SEC_LEVEL_GREEN /// Lowest alert level the event will trigger at
 
 	/// Whether or not dynamic should hijack this event
 	var/dynamic_should_hijack = FALSE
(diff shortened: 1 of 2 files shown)


alexkar598:
Do I have to point out the obvious here?

Bluedazzled:
this is just terrible, it'd become meta to just keep it on a specific threat level.

ynot01:
SOP would have to be firmly instated for something like this to work

adamsong:
As of right now they all still work on anything but delta/gamma/epsilon, which the crew cannot force.

Jarod1200:
Time to set the alert level and then cryo to roll for event antag

adamsong:
> Time to set the alert level and then cryo to roll for event antag

As of right now they all still work on anything but delta/gamma/epsilon, which the crew cannot force.

Xoxeyos:
Sorry, I just don't believe this is a change that is necessary.

Hopekz:
Everyone's freaking out for no reason. 
If you go through the code, it doesn't actually do anything at the moment. He set all the events to a minimum of green and a max of red, so nothing has changed under normal gameplay.

This PR adds the ability to use it if someone wants to do something with this in a future PR

Bibby0110:
> Everyone's freaking out for no reason.
> If you go through the code, it doesn't actually do anything at the moment. He set all the events to a minimum of green and a max of red, so nothing has changed under normal gameplay.
> 
> This PR adds the ability to use it if someone wants to do something with this in a future PR

The thing is nobody wants that, so why add the framework for it?

JamieD1:
Clearly no one actually knows what this does and are downvoting for no reason which is hilarious.

alexkar598:
> Clearly no one actually knows what this does and are downvoting for no reason which is hilarious.

Downvoting because its a bad idea no matter how you look at it. If an event needs this for a really edge case reason, it can use canSpawnEvent but theres no need to implement it in a way that encourages using it.

JamieD1:
> > Clearly no one actually knows what this does and are downvoting for no reason which is hilarious.
> 
> Downvoting because its a bad idea no matter how you look at it. If an event needs this for a really edge case reason, it can use canSpawnEvent but theres no need to implement it in a way that encourages using it.

You are clearly missing the point of why this is in.

Its to prevent shit like blob spawning when delta/epsilon is called.

Bibby0110:
> > > Clearly no one actually knows what this does and are downvoting for no reason which is hilarious.
> > 
> > 
> > Downvoting because its a bad idea no matter how you look at it. If an event needs this for a really edge case reason, it can use canSpawnEvent but theres no need to implement it in a way that encourages using it.
> 
> You are clearly missing the point of why this is in.
> 
> Its to prevent shit like blob spawning when delta/epsilon is called.

See above PR for a solution that doesn't promote doing the thing that nobody wants

Xoxeyos:
> 
> 
> > > Clearly no one actually knows what this does and are downvoting for no reason which is hilarious.
> > 
> > 
> > Downvoting because its a bad idea no matter how you look at it. If an event needs this for a really edge case reason, it can use canSpawnEvent but theres no need to implement it in a way that encourages using it.
> 
> You are clearly missing the point of why this is in.
> 
> Its to prevent shit like blob spawning when delta/epsilon is called.

We already have tools to toggle events, ghost roles, and silicons from spawning; You can find 'Ghost pool protection' underneath the admin tab. 
c3JROGr

JamieD1 (approved):

PR #12072, opened 2021-08-18
doubled piggyback mounting time

Marmio64:
piggyback combat is dumb and i was reminded to remove it by looking at mbrain

okay well it's not gone but its a lot harder to do since it takes 3 seconds instead of 1.5 seconds, and upgrading to a neck grab is also 3 seconds, so you have to be a lot faster if you want to pull it off.

:cl:  
tweak: doubled piggyback mounting time
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -938,7 +938,7 @@
 /mob/living/carbon/human/proc/piggyback(mob/living/carbon/target)
 	if(can_piggyback(target))
 		visible_message("<span class='notice'>[target] starts to climb onto [src]...</span>")
-		if(do_after(target, 15, target = src))
+		if(do_after(target, 30, target = src))
 			if(can_piggyback(target))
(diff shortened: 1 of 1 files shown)


Mqiib:
good balance tweak thank you mr marmimo

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -938,7 +938,7 @@
 /mob/living/carbon/human/proc/piggyback(mob/living/carbon/target)
 	if(can_piggyback(target))
 		visible_message("<span class='notice'>[target] starts to climb onto [src]...</span>")
-		if(do_after(target, 15, target = src))
+		if(do_after(target, 30, target = src))
```
Bibby0110:
```suggestion
		if(do_after(target, 3 SECONDS, target = src))
```

Bibby0110 (dismissed):
meant to do this

JamieD1:
Now make it so if the person you are carrying is fat they make you fall over.

JamieD1 (approved):

PR #12077, opened 2021-08-18
moves heating in crystallizer above the cooling so it like matters

SomeguyManperson:

This SHOULD make the cooling side of the crystallizer have a more noticable impact since currently it cools the gas inside from the current, on target temperature, then heats it with energy from the recipe, and this ONLY occurs while it's in the proper temperature range making it less reliable

# Changelog

:cl:  
tweak: crystallizer recipe heat is added/removed before the coolant gas heat 
/:cl:


JamieD1:
Ah a good ol, Theos MAYBE SHOULD PR

JamieD1 (approved):

PR #12079, opened 2021-08-18
Limits the contract of apprenticeship to wizards

adamsong:
You must now be a wizard to use the contract of apprenticeship

# Changelog

:cl:  
bugfix: fixed non wizards getting apprentices
/:cl:

### DIFF
--- a/code/modules/antagonists/_common/antag_spawner.dm
+++ b/code/modules/antagonists/_common/antag_spawner.dm
@@ -23,7 +23,7 @@
 	var/unlocked = FALSE
 
 /obj/item/antag_spawner/contract/unlocked
-	var/unlocked = TRUE
+	unlocked = TRUE
 
 /obj/item/antag_spawner/contract/attack_self(mob/user)
 	if(!unlocked && !user.mind.has_antag_datum(/datum/antagonist/wizard))
--- a/code/modules/antagonists/_common/antag_spawner.dm
+++ b/code/modules/antagonists/_common/antag_spawner.dm
@@ -20,8 +20,13 @@
 	icon = 'icons/obj/wizard.dmi'
 	icon_state ="scroll2"
 
+	var/unlocked = FALSE
+
+/obj/item/antag_spawner/contract/unlocked
+	var/unlocked = TRUE
+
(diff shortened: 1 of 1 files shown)


UselessTheremin:
maybe make it apply to slaughter demon blood vials as well? those seem like a lot more of a bad idea.

alexkar598:
Why?

Its not a bug its a tweak

adamsong:
> maybe make it apply to slaughter demon blood vials as well

Are slaughter demons bound to their summoner? That is the main issue here, is the apprentice would be bound to a nonantag if a nonantag used it.

> Why?

Nonantag with a bodyguard that can muderbone, not a good idea

Hopekz:
> Nonantag with a bodyguard that can muderbone, not a good idea

In what situation does this non-antag get access to this item?
There is no workaround with your fix as well so if someone wanted to use this item to bus they can't override it

adamsong:
> In what situation does this non-antag get access to this item?

Made this pr because a wizard gave an assistant a contract and they self antagged with it

> There is no workaround with your fix as well so if someone wanted to use this item to bus they can't override it

Will add when I get back to my computer

Jarod1200:
Slaughter demons have no allegiance. Their objective says "kill everyone" this includes the summoner

adamsong:
Then there is no reason to prevent their summon, the main problem with the apprentice is their allegiance to the nonantag, whereas for the slaughter demon there is no difference between wizard and non wizard using it

maxion12345:
Make it so lawyers can use em lmao

TheLinuxOS:
> Make it so lawyers can use em lmao

Make traitor lawyers able to buy em

adamsong:
> Make traitor lawyers able to buy em

No

SomeguyManperson:
how would a non-antag even GET a contract the wizard would have to like buy one then not use it and not refund it then have someone else use it like ????????

Bibby0110 (dismissed):
Code wise you need an override at the very least.

adamsong:
> how would a non-antag even GET a contract the wizard would have to like buy one then not use it and not refund it then have someone else use it like ????????

This did happen

JamieD1 (approved):

PR #12096, opened 2021-08-21
Random Events can no longer trigger on alert levels above red

Bibby0110:

#12071 alternative

# Wiki Documentation

# Changelog

:cl:  
tweak: Random events no longer trigger on 
/:cl:


Xoxeyos:
c3JROGr
I don't believe this should be implemented, we already have a tool that allows us to toggle events, ghost roles, and other such circumstances from spawning, using the 'Ghost Pool Protection'.

Bibby0110:
> c3JROGr
> I don't believe this should be implemented, we already have a tool that allows us to toggle events, ghost roles, and other such circumstances from spawning, using the 'Ghost Pool Protection'.

That's an active choice.

This makes it passively happen.

Xoxeyos:
> 
> 
> > c3JROGr
> > I don't believe this should be implemented, we already have a tool that allows us to toggle events, ghost roles, and other such circumstances from spawning, using the 'Ghost Pool Protection'.
> 
> That's an active choice.
> 
> This makes it passively happen.

Why should it be set to disallow events above red at all times? How is this objectively better?

adamsong:
Ideally the game can run without admin intervention, because there might not be admins online

Xoxeyos:
> 
> 
> Ideally the game can run without admin intervention, because there might not be admins online

No alerts above delta can be triggered without an admin online, and the only case where a delta alert happens is when a nuke is activated, in which case everything is already fucked up.

adamsong:
Malf AI can trigger delta, and that is a pretty long timer

Xoxeyos:
> 
> 
> Malf AI can trigger delta, and that is a pretty long timer

You're right, however I'd still like to be able to turn the random events on and off at will, and I, and perhaps others, would not like to dig around in VV trying to turn it off, when they want it changed.

adamsong:
This doesn't remove that

JamieD1 (approved):

code/modules/events/_event.dm
```diff
@@ -62,6 +64,8 @@
 		return FALSE
 	if(ispath(typepath, /datum/round_event/ghost_role) && !(GLOB.ghost_role_flags & GHOSTROLE_MIDROUND_EVENT))
 		return FALSE
+	if(GLOB.security_level >= SEC_LEVEL_GAMMA && !security_level_override)
+    return FALSE
```
JamieD1:
```suggestion
	if(GLOB.security_level >= SEC_LEVEL_GAMMA && !security_level_override)
		return FALSE
```

adamsong:
#12071 was merged

PR #10982, opened 2021-02-10
black overlay, unconscious = black screen

ynot01:
# General Documentation

## BLIND PEOPLE CAN STILL SEE THE SAME AS BEFORE

playing a black screen for the whole round would be shitty and it makes sense to be able to "see" via touching things immediately adjacent, this only targets unconscious people

### Intent of your Pull Request

when you are unconscious, your screen will go completely black
can still see HUD and by extension items, **does not trigger on softcrit**, will trigger on hardcrit, triggers on sleep/unconscious, "barely conscious" is not unconscious

unedited gameplay footage (used teleport and self-rez verbs to speed things up)

### Why is this change good for the game?

greater immersion & uncertainty when sleeping, better potential for kidnap & release gimmicks, no more metaing everything and everyone you saw while sleeping when you aren't supposed to be seeing anything

# Changelog

:cl:  
tweak: You can no longer see the real world when you're sleeping
/:cl:

PR #12108, opened 2021-08-23
"black overlay, unconscious = black screen" but for those with the heavy sleeper quirk

nmajask:

Reworks "[black overlay, unconscious = black screen](https://github.com/yogstation13/Yogstation/pull/10982)" by [ynot01](https://github.com/ynot01) so that it only effects those with the heavy sleeper quirk. I have only done a little bit of testing to make sure the main point of this PR works, so there may be a few issues I missed. If you want a better description of what this does, look at the [original PR](https://github.com/yogstation13/Yogstation/pull/10982).

# Wiki Documentation

If there is any mention of what Heavy Sleeper does, it will probably need to be edited to state this new feature.

# Changelog

:cl:  ynot01, nmajask
tweak: You can no longer see the real world when you're sleeping if you have the heavy sleeper quirk
/:cl:

### DIFF
--- a/code/_onclick/hud/fullscreen.dm
+++ b/code/_onclick/hud/fullscreen.dm
@@ -106,6 +106,11 @@
 	layer = BLIND_LAYER
 
 /obj/screen/fullscreen/blind
+	icon_state = "blindimageoverlay"
+	layer = BLIND_LAYER
+	plane = FULLSCREEN_PLANE
+
+/obj/screen/fullscreen/black
 	icon_state = "blackimageoverlay"
 	layer = BLIND_LAYER
 	plane = FULLSCREEN_PLANE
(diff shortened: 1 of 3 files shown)


Bibby0110:
Probably should up the point value of heavy sleeper then. This is kind of a big downside

nesisknesis (approved):

Cartlord:
Was the issue with blindness rectified? I recall getting my eyes stabbed out and having an all-black screen when the source PR was testmerged.

nmajask:
> Probably should up the point value of heavy sleeper then. This is kind of a big downside

Done

> Was the issue with blindness rectified? I recall getting my eyes stabbed out and having an all-black screen when the source PR was testmerged.

Looking at the code, looks like this happens because eye stabbing someone enough will make them unconscious, though I dont really see this as an issue as it is only temporary.

Marmio64:
edit the perk desc too maybe to show that it has this side effect

nmajask:
> edit the perk desc too maybe to show that it has this side effect

Added

PR #12140, opened 2021-08-25
Fix badmin

adamsong:

Badmins can set these values to naughty things that crash the server. Fixes that

# Changelog

:cl:  
bugfix: fixed badmins being able to put in naughty values
/:cl:


alexkar598:
Technical solution to a policy problem. Tell the admins to not do that instead

adamsong:
Ok, I'll make sure its added to newmin training, 'Don't set these specific variables out of bounds'

JamieD1 (approved):
Ideally we need to kill sleep all together in the codebase..

PR #12144, opened 2021-08-25
Prevent clockcult from warping into head's offices

adamsong:

Pretty sure I saw a PR for this, but I can't find it, and its not working if it was merged

# Changelog

:cl:  
tweak: Clockies can no longer warp to head's offices
/:cl:


Anvilman6:
Obligatory "hugbox" ((joking))

But I do like this change, and I think it's good to keep antags from having an even easier win condition

adamsong:
I think clockies having AA as long as 1 member has half a brain cell is hugbox

SomeguyManperson (approved):

PR #12150, opened 2021-08-25
Adds emergency spare, a temporary ID with AA that heads can print

adamsong:

Pressing the button on the comms console will trigger a command report, and print a temp ID. After 50 seconds, the ID will start to smolder. 10 seconds later, it will catch fire and turn to ash. Don't be holding it when this happens.

The ID has the same access as the spare, without heads, or id change access.

# Changelog

:cl:  
rscadd: Added a temp spare heads can print from the comms console.
/:cl:


Cartlord:
but why

Anvilman6:
Maybe just make it fizzle out
rather than start a fire

Would be neat if you could examine it to check how much time it has left

adamsong:
> rather than start a fire

You get 10 seconds warning, past that, its on you

Bibby0110 (changes requested):
Authentication check

JamieD1:
@adamsong

Bibby0110 (changes requested):
Replace your authentication check with authenticated_as_silicon_or_head
You’ll have to create the proc but that shouldn’t be hard

PR #12161, opened 2021-08-26
Disables using a default vote option for people who do not vote on map votes

adamsong:

Currently when voting on maps, if you do not vote on a map vote, your preference is used. This change requires players to actually vote for the map they want.

# Changelog

:cl:  
rscdel: Removed default vote on map votes
/:cl:

### DIFF
--- a/config/config.txt
+++ b/config/config.txt
@@ -166,7 +166,7 @@ VOTE_PERIOD 600
 # NO_DEAD_VOTE
 
 ## players' votes default to "No vote" (otherwise,  default to "No change")
-# DEFAULT_NO_VOTE
+DEFAULT_NO_VOTE
 
 ## disable abandon mob
 NORESPAWN


Hopekz:
This just introduces extra unnecessary effort just to still vote for box

Cartlord:
Not a fan, see Hopekz's comment

alexkar598:
am a fan of other maps but this just makes it tedious since theres a vote basically every round. this will have the opposite effect if anything

Anvilman6:
~~One of the reasons box is the default map is because Box _is_ the default map vote in preferences for people.~~ EDIT because my information is out of date, thank you for telling me Hopek

Decent change, imo - if people want their beloved boxmap, make them actively vote for it..

Hopekz:
> because Box _is_ the default map vote in preferences for people.

False. Box is not the default choice.
The default is no preference.

Anvilman6:
Oh right, that was changed a while ago and I keep forgetting because I set my map preference to Pubby and forgot about it forever.

I still don't think it's a bad idea to have people vote for what they want when an active map vote is called, though, rather than having it coast through on preference. If people want Box, let them show it and actively vote for Box.

SomeguyManperson (approved):
I hate the boxstation I hate the boxstation

Reed0506:
someone turn boxstation into a literal hugbox as a joke except its not a joke its permanent

Bibby0110:
> Oh right, that was changed a while ago and I keep forgetting because I set my map preference to Pubby and forgot about it forever.
> 
> I still don't think it's a bad idea to have people vote for what they want when an active map vote is called, though, rather than having it coast through on preference. If people want Box, let them show it and actively vote for Box.

In reality we should probably investigate wether box being the default means that some people who SHOULD be on no preference is in reality on box station. 
My personal proposal is wiping everyone’s preference back to no choice and then seeing if people re-enable box. Verifies that no weird stuff is going on

JamieD1 (approved):
The anti box crusade continues..

PR #12167, opened 2021-08-27
TEG Newbie Friendly Rework

Aquizit:

A remap of the TEG layout. Made newbie/non-engineer friendly to setup and keep running with minimal hassle for a shift. Keeps all fixtures and other options that were there before, albeit in a new piping configuration. Added a CO2 chamber for easy of use, plus a heater and freezer to the hot loop to keep the pressures and temperatures up. A secondary Air Alarm panel has been installed next to the CO2 chamber, with the vent and scrubber set to that alarm only, to keep it apart from the rest of the room (like the SM vents/srubbers).

I have tested this, and made adjustments where I either messed something up or put a pump in the wrong direction, etc. Then re-tested.

# Wiki Documentation

HOO BOY this changes everything. A completely new guide needs to be written. I have one mostly worked out, I would just need to tweak it for the specific changes I made, rather than the proof of concept writeup I initially made.
All the image pages for the TEG will need to be updated.

# Changelog

:cl:  
rscadd: Complete TEG piping rework.
rscadd: Added freezer and heater to the hot loop.
rscadd: Added a CO2 chamber that connects to the hot loop. 
tweak: Pipes have moved around but all fixtures remain.
tweak: Burn chamber slightly smaller, but still there.
/:cl:

### DIFF
--- a/_maps/RandomRuins/StationRuins/BoxStation/engine_teg.dmm
+++ b/_maps/RandomRuins/StationRuins/BoxStation/engine_teg.dmm
@@ -35,10 +35,6 @@
 /obj/machinery/atmospherics/pipe/simple/orange/hidden,
 /turf/closed/wall,
 /area/engine/engineering)
-"aH" = (
-/obj/machinery/atmospherics/pipe/simple/green/visible,
-/turf/open/floor/engine,
-/area/engine/engineering)
 "aJ" = (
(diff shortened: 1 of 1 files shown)


patpol4:
https://github.com/yogstation13/Yogstation/pull/12174

PR #12169, opened 2021-08-27
Changes heretic progression to allow for more experimentation

SomeguyManperson:

Heretic abilities are now seperated into tiers, functioning much the same as they do currently; you can only select abilities from a "tier" you have unlocked, and purchasing abilities vertically according to the wiki chart will unlock more tiers.
However, instead of requiring abilities to be unlocked by purchasing adjacent abilities, the whole set of abilities in a row are unlocked once they are made available i.e. you can purchase path of rust then immediately purchase imperfect ritual, allowing for less of a restricted playstyle.
HOWEVER, you now ALSO are required to purchase at least 3 abilities from a tier before unlocking the mark blade or ascension tiers.

Also, the grasp upgrade for each path has been removed, and its effects placed in the path's initial selection. This means you no longer have to find an influence to actually have any antag powers.

The codex's TGUI has been updated to show what abilities will go towards the next tier (shown in green) and which ones will not (shown in blue)

# Wiki Documentation

Heretics need to purchase 3 abilities from a side-classing set of abilities to unlock the exclusive blade mark and ascension knowledges, rather than needing to purchase an adjacent ability, and sideclassing no longer requires an adjacent ability to unlock

Additionally, the grasp and base path unlocks have been merged into one knowledge, conveniently squishing their descriptions and effects.

# Changelog

:cl:  
tweak: heretics are now required to purchase 3 abilities from each set before purchasing the next set of abilities
tweak: heretics now have a nice green/blue indicator for which abilities will allow them to progress
tweak: heretics no longer need to purchase their mansus grasp upgrade seperate from their path selection, reducing the cost of progression by 1 point
/:cl:

### DIFF
--- a/code/__DEFINES/antagonists.dm
+++ b/code/__DEFINES/antagonists.dm
@@ -84,12 +84,23 @@
 #define CONTRACT_UPLINK_PAGE_CONTRACTS "CONTRACTS"
 #define CONTRACT_UPLINK_PAGE_HUB "HUB"
 
+///Heretics
 ///It is faster as a macro than a proc
 #define IS_HERETIC(mob) (mob.mind?.has_antag_datum(/datum/antagonist/heretic))
 #define IS_HERETIC_MONSTER(mob) (mob.mind?.has_antag_datum(/datum/antagonist/heretic_monster))
+#define IS_EXCLUSIVE_KNOWLEDGE(knowledge) (knowledge.tier % 2)
 
 #define PATH_SIDE "Side"
 
 #define PATH_ASH "Ash"
 #define PATH_RUST "Rust"
 #define PATH_FLESH "Flesh"
+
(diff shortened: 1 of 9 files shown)


Djiq:
this is an interesting concept and frees up heretic playstyle a lot, and i really like the idea of it, kind of destroys the concept of a 'path' tho, but oh well

SomeguyManperson:
> this is an interesting concept and frees up heretic playstyle a lot, and i really like the idea of it, kind of destroys the concept of a 'path' tho, but oh well

maybe paths could increase the cost of opposing paths' upgrades and/or restrict the blade and mark upgrades so they're still mechanically unique

SomeguyManperson:
removed purchasing other paths' mark and blade upgrades

JamieD1 (approved):

PR #12174, opened 2021-08-27
Add TEG Heater

Aquizit:
This adds a heater to the TEG hot loop. This makes it easier for new/inexperienced engineers, or non-engineers if the station doesn't have any, to set up the engine in a stable way to provide sufficient power to the station.

I have tested this. It's not changing much so there wasn't much to test.

# Wiki Documentation

The wiki will need a rewrite for the *basic* TEG setup. I have most of it written already, I just would need to tweak it to fit what I ended up mapping.
This would replace the current "basic" setup, which involves a plasma burn. That should be kept, just perhaps as an advanced course.
The TEG image would need to be changed.

# Changelog

:cl:  
rscadd: Added heater to TEG hot loop.
/:cl:

### DIFF
--- a/_maps/RandomRuins/StationRuins/BoxStation/engine_teg.dmm
+++ b/_maps/RandomRuins/StationRuins/BoxStation/engine_teg.dmm
@@ -187,6 +187,12 @@
 	},
 /turf/template_noop,
 /area/space/nearstation)
+"cI" = (
+/obj/machinery/atmospherics/pipe/manifold/orange/visible{
+	dir = 4
+	},
+/turf/open/floor/engine,
+/area/engine/engineering)
 "cL" = (
 /obj/machinery/atmospherics/pipe/heat_exchanging/simple{
 	dir = 5
@@ -513,6 +519,16 @@
 /obj/machinery/atmospherics/pipe/heat_exchanging/simple,
 /turf/template_noop,
 /area/space/nearstation)
+"oD" = (
+/obj/effect/turf_decal/trimline/red/filled/corner{
+	dir = 8
+	},
+/obj/effect/decal/cleanable/dirt,
+/obj/machinery/atmospherics/components/unary/thermomachine/heater{
+	dir = 4
+	},
+/turf/open/floor/engine,
+/area/engine/engineering)
 "oK" = (
 /obj/effect/turf_decal/stripes/line,
 /obj/machinery/airalarm{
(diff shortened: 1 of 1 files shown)


Aquizit:
**Wiki Writeup**

TEG for Dummies:

Step 1: Build the TEG
Open the crate in the engine room, it will contain three circuit boards. You'll see one machine frame is already set up for you. Up and to the left you will see a stack of metal and some cable, grab it (you will not need all the cable). Make two more machine frames right next to the other one, with the premade one being the rightmost. Wrench them down, wire them, then stick in the circuit boards. Make sure the circulator boards are on the outside, and the engine board itself is in the middle. Then just screwdriver the machines to completion!

But wait, it doesn't look finished yet does it? Next, use the screwdriver to open the two circulators. Use your multitool on the one on the left to change it to 'Cold.' Then unwrench them, and alt-click on them to spin them to line up correctly with the middle engine. To note there is a little indent on one side that should be touching the engine proper. Then wrench them back down, and if you did it right you should see the pipes connect to the circulator. Make sure you close the panels on the circulators with the screwdriver, then wrench the middle component, that should complete the engine and attach the pieces all together.

Phew that was just step 1!

Step 2: Set up the Loops
To the left you will see a blue loop of pipe, to the right an orange. You'll want to turn on most of those pumps. The only one you _do not_ want to turn on is the downward facing one on the blue loop, next to the CO2 canisters. You _do_ want to turn on the pump connecting the orange pipe to the black pipe. Make sure to set each pump to max pressure (4500 kPa). Also make sure to turn on the heater connected to the orange pipes.

Step 3: Let loose the gas!
Wrench all six CO2 cans down into the system. The gas should begin to flow through the system, and the engine should light up pretty colors! If it does not, you may have missed a step somewhere. But the left one should glow blue, the right should glow orange, and the middle should show a bar slowly growing as the power generated grows.

LoliconSlayer:
For the record, this is done at the request of the wiki so that a single person can set up the engine round start with only Station Engineer level access and no external assistance (IE From Atmospherics) and no additional machinery. As far as I'm aware, that is not possible with the current round start layout.

Aquizit:
Uh, reopening this because I totally looked for additional comments before closing it. :^)

Aquizit:
I'm closing this out because the files are all messed up now and I'm going to work out with wiki and mapping teams what the best approach should be moving forward. Especially with that other TEG-related PR open as well.

patpol4:
https://github.com/yogstation13/Yogstation/pull/12192

PR #12176, opened 2021-08-28
Removes normal footwraps from non-engi and non-miner jobs

nmajask:

Removes normal footwraps from non-engi department and non-miner jobs. The main reason for this was to make normal footwraps act as replacements for work boots instead of normal shoes.

## Why?

Because giving them to everyone makes digitigrade legs less interesting by essentially removing an interesting flaw they have. The only time you see the lack of ability to wear normal shoes is with specialized shoes, which is rare. Also, they can be crafted for two leather, so just bug botany if you really need them.

# Wiki Documentation

Will obviously need to change any place where the effected jobs were mentioned to have them. 

# Changelog

:cl:   
tweak: removed normal footwraps from non-engi department and non-miner jobs 
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/_job.dm
+++ b/code/modules/jobs/job_types/_job.dm
@@ -241,13 +241,13 @@
 
 	if (isplasmaman(H) && !(visualsOnly)) //this is a plasmaman fix to stop having two boxes
 		box = null
-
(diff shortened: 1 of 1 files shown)


nmajask:
i should have mentioned it is the engineering department that keeps them

ynot01:
based

N3D6:
not racist just don't loike em

JamieD1 (approved):

PR #12196, opened 2021-08-30
TEG Heater final i'm so tired

Aquizit:
Adds a Heater to the TEG Heating Loop

This is a way to create a stable TEG engine that a sole inexperienced engineer can set up to provide sufficient power to the station. I have discussed the desired state of the TEG setup with Mapping and Wiki at length, as well as my own experimentation before making this final decision/map. I've also tested this method and specific map to make sure it works.

**Please see my thread here to see my experimentation:** https://forums.yogstation.net/threads/teg-shortcomings-and-solutions.23920/

I can easily write up a bullet-pointed how-to on how to set the engine up with this heater for the wiki.

:cl:  
rscadd: Added Heater to TEG Heating Loop
tweak: Shifted some things around to make it fit nicely at wej's suggestion
/:cl:

### DIFF
--- a/_maps/RandomRuins/StationRuins/BoxStation/engine_teg.dmm
+++ b/_maps/RandomRuins/StationRuins/BoxStation/engine_teg.dmm
@@ -320,6 +320,19 @@
 	},
 /turf/open/floor/plasteel/dark,
 /area/engine/engineering)
+"ht" = (
+/obj/structure/table/reinforced,
+/obj/effect/turf_decal/stripes/line{
+	dir = 1
+	},
+/obj/item/clothing/gloves/color/black,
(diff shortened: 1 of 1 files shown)


alexkar598:
You can run a TEG on nothing but a heater and a freezer, this is lame

Aquizit:
I don't know what you're getting at. This is adding the heater so it can run on it? Unless you mean that it can run on either is the lame part, which then yeah but then take it up with Wiki about engine minimum standards.

JamieD1:
Will wait for Weje, but seems fine.

patpol4 (approved):
Is good, i agree this is a lazy fix but the nonlazy fix would be a total redesign

PR #12199, opened 2021-08-30
Adds tooltips to the sleeper UI

nmajask:

Adds tooltips to the buttons on the sleeper UI, which use the description found in the chemical dataum. Should help new players use the sleeper without having the wiki open. I've done some testing, but I may have missed a few issues.

SleeperTT

This tooltip is only shown to those in the medical department when they spawn, those who are not are shown this instead:

image

# Wiki Documentation

I dont think anything will need to be changed, though it may be a good idea to mention this in a tip or something. 

# Changelog

:cl:  
tweak: added tooltips to the sleeper tgui
/:cl:

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -169,7 +169,7 @@
 	data["chems"] = list()
 	for(var/chem in available_chems)
 		var/datum/reagent/R = GLOB.chemical_reagents_list[chem]
-		data["chems"] += list(list("name" = R.name, "id" = R.type, "allowed" = chem_allowed(chem)))
+		data["chems"] += list(list("name" = R.name, "id" = R.type, "allowed" = chem_allowed(chem), "desc" = R.description))
 
 	data["occupant"] = list()
 	var/mob/living/mob_occupant = occupant
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -161,8 +161,9 @@
 /obj/machinery/sleeper/nap_violation(mob/violator)
 	open_machine()
 
-/obj/machinery/sleeper/ui_data()
+/obj/machinery/sleeper/ui_data(mob/user)
 	var/list/data = list()
+	data["knowledge"] = IS_MEDICAL(user)
 	data["occupied"] = occupant ? 1 : 0
 	data["open"] = state_open
 
--- a/tgui/packages/tgui/interfaces/Sleeper.js
+++ b/tgui/packages/tgui/interfaces/Sleeper.js
@@ -34,7 +34,7 @@ export const Sleeper = (props, context) => {
   ];
 
   return (
-    <Window width={310} height={465}>
+    <Window width={400} height={520}>
       <Window.Content scrollable>
         <Section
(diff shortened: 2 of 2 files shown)


Cartlord:
Is there a way to only make this show for those in the Medical department?

Anvilman6:
~~I'm very on the fence about this. It's a nice change in terms of what it does and how it does it, but it feels far too easy-mode for me. If anything I think Cartlord's suggestion may be a good option to examine.~~

EDIT:  With the changes to how this works, I'm much happier with it. It _still_ feels a little too easy mode for my tastes even with that limitation, but at the same time it helps mitigate my main concern below. Well done Nmajask.

Mostly what I'd like to avoid is making it even _easier_ for anyone to put their greyshirt friend in the sleeper, push some buttons, then get them back up to full health without Med Staff doing anything.

N3D6:
hugbox

Chubbygummibear:
> Is there a way to only make this show for those in the Medical department?

I was kind of thinking along these lines as well, but my take was that most medical players probably already know what they do, so these tooltips are primarily beneficial to those who don't know much about medicine. My idea was what if they only showed on upgraded sleepers? Like the basic sleeper is like a command line interface and doctors are used to it but a regular crew member would be horribly confused, but an upgraded sleeper theoretically has like a touchscreen display with a fancy interface that makes it usable to those who wouldn't know how.

nmajask:
> Is there a way to only make this show for those in the Medical department?

Yes, and I am currently working on implementing it



> I'm very on the fence about this. It's a nice change in terms of what it does and how it does it, but it feels far too easy-mode for me. If anything I think Cartlord's suggestion may be a good option to examine.
> 
> Mostly what I'd like to avoid is making it even _easier_ for anyone to put their greyshirt friend in the sleeper, push some buttons, then get them back up to full health without Med Staff doing anything.

I didn't think of how non-medical staff would use it, I just assumed it would only be used by new medical players which is my bad. As I said above, I am working on making it only show for medical staff.



> I was kind of thinking along these lines as well, but my take was that most medical players probably already know what they do, so these tooltips are primarily beneficial to those who don't know much about medicine. My idea was what if they only showed on upgraded sleepers? Like the basic sleeper is like a command line interface and doctors are used to it but a regular crew member would be horribly confused, but an upgraded sleeper theoretically has like a touchscreen display with a fancy interface that makes it usable to those who wouldn't know how.

This is an interesting idea, but I think it may be a bad idea to make them more confusing for everyone. Many new players choose to play medical, and making one of their primary tools more confusing to use would most likely discourage them from playing.

nmajask:
The tooltip is only shown to those in medical, those who are not are shown this instead:
image

Cartlord:
Good PR!

Anvilman6:
This may be scope creep on the PR (and potentially should be atomized as a separate change), but do you think it would be possible to have these popups be a preference or a toggle of some kind? Just in case people already know what the chems do and don't want the textbox showing up.

Very impressed by what you've made work so far, Nmajask!

JamieD1:
> hugbox

Im going to put hugbox on every single PR

JamieD1 (approved):

PR #12082, opened 2021-08-19
Adds a racial language for every xeno who doesn't have one

Chubbygummibear:
YOU GET A LANGUAGE, AND YOU GET A LANGUAGE, EVEN THE LIGHTBULBS GET A LANGUAGE! LIZARD IDENTITY SUPREMECY IS OVER.

# Github documenting your Pull Request

As the title says, adds 3 brand new languages for the Moths, Ethereals, and Plasmamen. These languages will be known by the corresponding races at round start, and in the case of sheltered, you have your racial languages instead of defaulting to Space Japanese. My hope is that this will give some life to the other races and attract people who want to have a sense of identity about their characters, and having a racial language reinforces this. These languages even have their own (bad) sprites that I made (as well as adding one for Sylvan because I didn't know how the sprites worked until now).

These languages have been tested and work, but I am unsure of how they interact with the Telecomms interpreter script where the alien races are mentioned, because I don't know how to test that.
image
for simplicity's sake, the languages are tied to the standard human tongue, but this could be changed if there's a demand for say, only plasmapeople can speak their language because it relies on bone sounds that couldn't be made with a normal mouth.

Also I cleaned up the Sheltered quirk so that instead of a towering if-statement, it calls the language holder proc that tries to select a language from the trait holder, if it fails, that means the holder has no spoken languages after spoken common is removed, and therefore they need space japanese and grants it. This works for humans in testing, and all xenos understand common and speak and understand their racial language.

The Etherean, Sylvan, and Bonespeak sprites are done by me, and I don't have an artistic bone in my body, so if they need to be fixed up, just let me know with some feedback
image

# Wiki Documentation

I'm not sure how much is actually on the wiki about the languages, but if there's so much as a racial language tab on each race's page, it would need to be updated with these

# Changelog

Edit the changelog below to reflect the changes made by this PR, even if the changes are minor - required for every PR.
If you add a name after the ':cl', that name will be used in the changelog. Leave it empty to use your GitHub name.

:cl:  
rscadd: Adds new racial languages for Moths, Plasmamen, and Ethereans, so that spacemen of all kinds have their own racial language
tweak: tweaked how the sheltered trait determines if the trait holder needs Space Japanese to a cleaner method than an unbelievably thicc if-statement
imageadd: added icons for Etherean, Sylvan, and Bonespeak
/:cl:

PR #11994, opened 2021-08-10
Nerfs Sheltered (Understand Common)

JamieD1:
# Github documenting my PR
Nerfs Sheltered - Can now Understand Common but still cant speak it.
# Wiki Documentation
🆑
Tweak: Nerfs Sheltered - Can now Understand Common but still cant speak it.
/🆑

PR #12202, opened 2021-08-31
Removes the "Sheltered" Quirk

Cartlord:

**Untested, webedited.** 
Removes the "Sheltered" Quirk, because not only is it bad for the person who has it, it's a pain in the ass for everyone AROUND them because they're unable to communicate anything - very, very bad for Heads. This is especially true after its "nerf" ( https://github.com/yogstation13/Yogstation/pull/11994 ), which made it so Sheltered people can understand Common; this means the Sheltered person is at no disadvantage, while everyone else gets screwed over because the Sheltered person gets a get-out-of-jail-free-card for communication (as in, "It's not my fault I made no effort to work with anyone, I can't talk to them!").
I admit I have no clue what I'm doing and this probably isn't interacting properly with https://github.com/yogstation13/Yogstation/pull/12082
# Wiki Documentation

Would need to remove Sheltered from the Quirk list on there.

# Changelog

:cl:  

rscdel: Removed the "Sheltered" Quirk

/:cl:


JamieD1:
image

JamieD1:
For the record I agree and think sheltered is the stupidest thing ever created.

maxion12345:
democracy says just remove for command/sec

Anvilman6:
See #12203, as it covers this topic and goes much farther beyond this scope. 

This may be good to have as a fallback in case the larger version doesn't go through, but be mindful of the other PR existing.

Personally I do think Sheltered should be removed because as of now it's just Mute with even less consequences (as you get a fallback language), and it's effectively 3 free perk points with how little people on Yogstation want to actually talk and RP things out.

Cartlord:
I honestly think Theos just made his PR as a failed attempt at mocking this one

N3D6:
Yes.

Reed0506:
sheltered more like what if i was the mime and got 3 free quirk points because i never press the T button ever

SomeguyManperson:
> I honestly think Theos just made his PR as a failed attempt at mocking this one

it was supposed to be in response to anyone who was like "oh I hate sheltered because it makes heads of staff massive shit to work with" since there's a bunch of quirks that do that and apparently applying effort to code a solution to that is too difficult for most people but the more I think about sheltered the more I dislike it since as a negative 3 point quirk it does not drastically change your gameplay and isn't even used to accomplish the thing it was coded for, instead just being used to force certain groups to communicate with specifically whoever took it

The quickfire vote about it is actually two votes in one, a vote to either set up a system to prevent important roles from taking specific quirks, or to remove those quirks from the pool, and a vote to keep sheltered or remove sheltered

As it used to be, sheltered had a downside since it effectively was both a mute and a deafen as you could not understand the main used language, now it's just a "lizard that only speaks to other lizards" gimmick allower when you can already do that without the quirk. The reason it's the most used quirk by heads of staff is because it really doesn't effect the user at all, especially if they play a common species, while other negative quirks have some actual downside that requires maintenance or severely hampers the taker's ability to function.

tldr: should sheltered exist? probably not, it doesn't even accomplish the purpose it was coded for. Should sheltered give THREE quirk points? definitely not. should someone set up a system to blacklist quirks based on job prefs? yeah

Firewolf34:
So hear me out. And I understand that this is somewhat out of scope for this PR. But I want to pose a case for the potential of Sheltered - a game mechanic completely unique to SS13 that would not have the possibility to exist if removed. 

What if the language obfuscation function was deterministic instead of just picking from a list of jumbled text?

For example: if a player says "Hello" and it is deterministically changed to "Elang" for those that don't understand the language (every time for that word, and additionally, corresponding to length of the sentence). Then, other players could *learn* that "Elang" means "Hello" based on how the player is acting and emoting. 

Then. If a player says "Elang" in Common, the foreign language player understands it as "Hello". 

This mimics the natural processes of how humans learn and interact with foreign languages in real life - and in my opinion, the lack of this feature is *why* sheltered is difficult right now. It has a *lot* of potential! 

You could tie it to the random seed such that you cannot memorize across rounds. Boosts to other areas of the station: curator/library could give out language translation books with basic phrases. Random seed fixing probability event has more use as well.

I politely request we keep it for non-essential crew as we voted in quickfire. Perhaps a short-term fix is to remove it from randomized quirk selection. I saw a new player the other day with his first character ever random-genned to Deaf/Sheltered combo. Brutal.  

--------

**Tl;dr**: sheltered has great potential. Just needs a rework. And it should not be for Heads of Staff - doesn't make sense at all why a Head would get promoted to Command without speaking *Common*.

JamieD1:
Needs to be reworked not removed.

Cartlord:
I don't see how it'd be "reworked" apart from the aforementioned limiting to non-essential crew, which would require an entirely new infrastructure for Quirks to be job-restricted; in addition, I've never seen or heard of anyone having anything resembling fun with a Sheltered crewmate, so even then I can't see what purpose this ability would serve. I like Firewolf's suggestion, but that is frankly outside the scope of my expertise.
I don't see what bad could come of removing Sheltered as-is, except making things harder for the lizard mafia.

Cartlord:
Conflicts resolved

adamsong:
This hasn't been touched in months, and a rework has been merged

PR #12203, opened 2021-08-31
Removes every negative trait that would be a pain in the ass for a head of staff to take

SomeguyManperson:

Removes all negative quirks that substantially reduce the owner's ability to communicate or defend themself or survive unassisted, since we can't trust the playerbase to not take these as roles wherein the capacity to communicate and fight is important, and we for some reason refuse to enforce preventing this through means other than outright removal.

# Changelog

:cl:  
rscdel: removes the quirks: bad back, blindness, deafness, junkie, pacifist, kleptomaniac, sheltered, brain tumor, parapalegic, social anxiety, unstable, and ineloquent.
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -147,7 +147,6 @@
 #define TRAIT_UNINTELLIGIBLE_SPEECH "unintelligible-speech"
 #define TRAIT_UNSTABLE			"unstable"
 #define TRAIT_OIL_FRIED			"oil_fried"
-#define TRAIT_SHELTERED			"sheltered"
 #define TRAIT_RANDOM_ACCENT		"random_accent"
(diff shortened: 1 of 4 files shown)


Partheo:
ironically, this is a good pr

Marmio64:
but why bad back

SomeguyManperson:
> but why bad back

prevents players from carrying stuff important to their job such as spare weapons and cuffs

JamieD1:
image

Anvilman6:
This seems drastic, and perhaps not the nuclear option we need to push.

These quirks are hindering to have, and of course our playerbase is stupid enough to take them in roles where they're even more crippling, but by the quickfire vote shouldn't the change be formatted just so it strips these negative perks from important roles (Command/Sec), rather than removing like half of our quirk system?

PR #12214, opened 2021-09-01
Prevents nonmedical personnel from using sleepers

adamsong:

People who aren't doctors like to powergame and use the sleepers, now they can't. 
This does not apply to syndicate or clockie sleepers.

# Wiki Documentation

Should probably mention only doctors can use them

# Changelog

:cl:  
tweak: Locked sleeper use to doctors only
/:cl:

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -11,7 +11,10 @@
 	icon_state = "sleeper"
 	density = FALSE
 	state_open = TRUE
+
 	circuit = /obj/item/circuitboard/machine/sleeper
+
+	var/req_job = TRUE
 
 	var/efficiency = 1
 	var/min_health = -25
@@ -161,10 +164,11 @@
 /obj/machinery/sleeper/nap_violation(mob/violator)
 	open_machine()
 
-/obj/machinery/sleeper/ui_data()
+/obj/machinery/sleeper/ui_data(mob/user)
 	var/list/data = list()
 	data["occupied"] = occupant ? 1 : 0
 	data["open"] = state_open
+	data["canuse"] = IS_MEDICAL(user) || issilicon(user) || !req_job
 
 	data["chems"] = list()
 	for(var/chem in available_chems)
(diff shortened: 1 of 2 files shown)


azzzertyy:
stupidest pr ive ever seen unironically

adamsong:
> stupidest pr ive ever seen unironically

Can you be more specific, or do you just not like not being allowed to powergame?

Bibby0110 (changes requested):
Authentication check for canuse when actually clicking the button.
Don't trust clientside info

UselessTheremin:
are you high

azzzertyy:
> > stupidest pr ive ever seen unironically
> 
> Can you be more specific, or do you just not like not being allowed to powergame?

nice strawman adam, no, i dont like powergaming
but i dont see how sleepers are powergaming and how this would ever solve the issue; instead, now people will just idk break into medbay storage because they have no way of healing because all the doctors died

adamsong:
> but i dont see how sleepers are powergaming

How is administering chems to patients not powergaming?

azzzertyy:
> > but i dont see how sleepers are powergaming
> 
> How is administering chems to patients not powergaming?

???????????
are you okay?

Anvilman6:
Fascinating

cuackles:
Locking the sleeper is a bad idea, med kits are stored away on medbays back, first aid kits don't have a way to treat toxin damage and in case of a stupid or absent medical crew this stops people for healing themselfs, even if you change job you wouldn't still be able to use it if you didn't start as a med as i understood and the main problem is impatient or shitty players breaking in or not respecting the docs, if you really want to diminish that this is a bit overkill, you could consider give it a channeling if someone without a med is is using it on someone and give it a little warning, or lock some not essentials Chems from the machine without authorization but outright locking it is a bit fucked up

Moggram:
oh shit, you did it you mad man.

Bibby0110:
Might want to consider allowing usage on red alert or above

Hopekz:
My fight or flight instincts of using the sleeper all the time casually as non-medical want me to dislike this PR, but it does make sense with how the game direction is going.

With #12199 non-medical shouldn't understand what those drugs do, so using the sleeper as non-medical would be meta-knowledge or power gaming at that point because YOUR CHARACTER has no idea what they do. 

If #12199 gets merged as is, then this type of PR makes sense logically, but you should have probably waited until it was merged tbh. 

Still very interesting we should probably give this a test merge to see how people actually handle it.

image

adamsong:
#12218 Does this better

PR #12219, opened 2021-09-01
Forgot to add descriptions when I added the access to freeminers

adamsong:

Forgot to add descriptions, so here they are, now adding the access to mechs works

:cl:  
rscadd: Added descriptions to free miner access, allowing them to actually be used
/:cl:


JamieD1 (approved):
Moron

PR #12226, opened 2021-09-02
Reworks the sleeper so it passively heals instead of using chemicals

SomeguyManperson:

Sleepers no longer have chemicals they inject (aside from morphine), instead they will slowly provide an effect over time, with effects being unlocked by manipulator tier

tier 1: treat injuries: heals 2 points of brute and burn damage per process
tier 2: repair organs: heals 0.2 point on a random damaged organ every process, more organs healed with higher efficiency (reduced to 0.05 if the patient is dead)
tier 3: purge chemicals: heals 2 point of toxin damage and purges units of every present chemical by 0.5 point per efficiency up to 2
tier 4: if the occupant falls into (or is in/past) hardcrit, they are put into stasis, preventing them from dying like an idiot

Morphine is still available with the "sedate" button, allowing up to 20 units to be injected. I'd add epinephrine but tgui decided it didn't like me doing that and I can't be assed to find out how to fight it

image

# Wiki Documentation

see above roughly the same information

# Changelog

:cl:  
rscdel: sleepers no longer inject healing chems
rscadd: sleepers now heal passively over time, with different forms of healing being unlocked by different levels of parts. They're still kind of slow.
/:cl:


Marmio64:
I'd rather keep it the same but have a limited chem storage that regens very very slowly and it can only inject up to 5 units so patients stay longer and sleepers are for small, infrequent injuries. This one is just an auto tendwounds surgeon

Moggram:
this seems more like a auto doc, but I still think its cool.

Moja10:
Why not just make the chemicals have to be filled manually by the chemist, along with this function? That way doctors without a chemist can function, and doctors with a chemist function even better, but still slower than before because lmao no free chems.

SomeguyManperson:
> Why not just make the chemicals have to be filled manually by the chemist, along with this function? That way doctors without a chemist can function, and doctors with a chemist function even better, but still slower than before because lmao no free chems.

can't chem just like give doctors the fast healy chem pills and/or put them in a chemfridge and announce their existence

maxion12345:
this sounds like cyro but without the cyro

SomeguyManperson:
> this sounds like cyro but without the cyro

yes

Anvilman6:
Have thoughts on this I will give after work
Elements I like here, but also ones I think can be changed

BlueishTsunami:
agree this is basically just cryo, but slower. Also, it seems if you close the sleeper with no one in it, it is stuck closed. To add context, i closed from inside the sleeper menu, and cannot open with alt+click or using open button

SomeguyManperson:
Specifically, in its current state the sleeper is the end all at nearly all stages of triage. Dying person? jam them in a sleeper and inject them with everything, heavily injured? jam in sleeper inject with everything, lightly injured? jam in sleeper inject with everything. This heavily reduces the worth of pre-made chemical pills or patches that aren't specifically for virus cures or meth, using cryo, or even using the tend wounds surgeries which are faster than most chemical mixes. Additionally, someone who's been healed with a sleeper doesn't need to stick around in medbay to get healed.

The purpose of this change is to make the time investment of the patient for sleepers more in tune with the time investment for other sources of healing, while promoting the use of actual chemical healing from chemistry and tend wound healing from medical doctors. Cryo can be optimized by chemists to inject a mix of several chemicals to many more people than using them manually, since cryo duplicates chems which is more efficient with higher tier parts. Surgeries are made much faster at higher tiers of surgery research as well, generally taking under a minute of treatment with large reductions per level of research.
Sleepers on the other hand require literal seconds of treatment, all you have to do is get the person inside, then click the buttons for the healy chems and they are set. At 10 units, each chemical will heal at least 100 points of damage over its lifespan, while the patient is free to do whatever they want unlike other treatment methods save for chemical healing from chemistry which requires chemists to actually do their job, which is currently completely ignored in favor of sleepers.

I don't want sleepers to have some skill floor or maintenance floor since they are the easiest item in medbay to use in a pinch, being able to heal with just themselves and a single outside operator who doesn't need to know much aside from what the buttons do. I'd rather make them inoptimal for actual doctors to use so they are pressured to apply other healing methods in medbay first than add some skill floor to make them more "intuitive"
why don't people use cryo? it's considered noob bait and knocks out the patient for a long time, making them vulnerable
why don't people use surgery? it's too trust oriented even if you don't need to use anesthetics for tend wounds surgeries, you're still on a table resting which lowers your movement speed, and you likely aren't wearing clothes or armor making it easier to get killed
why do people use sleepers? the worst thing that can happen (usually) is getting injected with morphine, which has several tells before it knocks you out, or getting overdosed, which gives a nice large warning for the patient who is able to easily react to it. it is the fastest and safest form of healing that anyone can use anywhere.

The non-damage healing parts are to give sleepers a spot even in a medbay that wouldn't need their slow healing. As we lack a fast way to purge chemicals, and any real way to actively heal organs outside botany, those seemed like a good set of tools to give to the sleeper.

Bibby0110:
Updated your title to actually describe what the PR does

Sniblet:
For every medical change we let him get away with he will merge three more, someone stop him before we turn into Bay. This is a slippery slope argument but it's real
Thoughts/concerns
-Medbay is more and more dependent on surgery, mediborgs and cryo with everything you've done to it
-Or chemistry, but no one plays chemist to help people and this won't change that. It's Fails-With-Numbers' job every single round to make sure chemistry does anything but fill up with randomnames at roundstart for no reason, carpet the lobby, and then grief
-Salbutamol is dead. Treating oxyloss is going to be rough
-Why did we remove chemistry macros again
-Clockwork sleepers too? Clockies don't have a timely way to do chemistry or surgery and most of them don't know that the slab has spells, especially if they're turning to those sleepers to begin with

SomeguyManperson:
> For every medical change we let him get away with he will merge three more, someone stop him before we turn into Bay. This is a slippery slope argument but it's real
> Thoughts/concerns
> -Medbay is more and more dependent on surgery, mediborgs and cryo with everything you've done to it

yes

> -Or chemistry, but no one plays chemist to help people and this won't change that. It's Fails-With-Numbers' job every single round to make sure chemistry does anything but fill up with randomnames at roundstart for no reason, carpet the lobby, and then grief

yes, people might be more prone to making healing chemicals now that they won't just catch dust in the chemfridge because the sleeper does that but better. "good" chemist is a thankless job currently since the only chems you'll be making are the CMO busting in from virology or to virology for a cure/vaccine and synthflesh people just kind of expect to be there

> -Salbutamol is dead. Treating oxyloss is going to be rough

perf still exists, the main downside to perf is that the toxin damage can't be easily treated outside of medbay, which you generally should be in if you are treating oxygen damage with a sleeper

> -Clockwork sleepers too? Clockies don't have a timely way to do chemistry or surgery and most of them don't know that the slab has spells, especially if they're turning to those sleepers to begin with

I might get back to that one at some point but people should kind of learn what antag abilities they have access to regardless, and the clockie sleeper does also use vitality to heal over time which will make it a bit faster

ToasterBiome:
I actually love this change because I get to do surgeries and actually use cryo. Started seriously learning medical right when this dropped. Once it was reverted everything became "throw person into sleeper". Sleepers always felt like powercreep to me.

SomeguyManperson:
Healing is now 2 points up from 1 for everything, putting it on par with trekchems since the main important parts are forcing people to stay in the sleeper and making it not completely outdo chemistry

Anvilman6:
Having played with this change, I'm still of the mindset that sleepers should keep their low level function (1 heal / tick), and instead have the potential to be fed chems that will be applied in specific ways (using the basic flight of chems that old sleepers could handle) with a small enough buffer that they can't load in 500u and forget about it. I would also say that the organ repair function is too generous, to be freely usable; more on that later.

The main goal here seems to be to keep people in the sleepers for treatment. The current model of turning them into Treat Wound machines seems to accomplish that goal, but it's entirely hands off. 

I think that allowing chemists to fill up the sleeper's buffer with Libital / Salbutamol / etc, and having it drip feed into the person as well (as keeping the pseudo-cryo passive healing as an alternative option) would be a good solution to balancing active and passive components. By drip feed I mean have the sleepers administer a dosage small enough that it will tick and give the healing benefit, but then immediately tick down and dissipate, to be replenished at that baseline the next tick by the sleeper (for example, keeping .4u Libital being re-administered into the person so it ticks, dissipates, then is replenished). This will stop the prior meta of dosing someone with 20u of healing chems and then sending them on their way within .2 seconds, as well as encourage active decision-making by Medical Doctors and Chemistry.

And to expand on my earlier statement about the organ repair being too generous, I would think the sleeper should need to be fed synthflesh (or synthmeat / biomass?) to fuel this process, rather than letting it do an arguably powerful treatment for free, as free healing was another flaw of the earlier sleeper design.

EDIT: Toxins purging could potentially be refactored into dialysis like other servers have it? Drain the toxins into a cup through blood filtration.

adamsong:
I must agree, the healing is way too much right now, and the toxins filtration is pretty quick

code/game/machinery/Sleeper.dm
```diff
@@ -157,6 +175,40 @@
 /obj/machinery/sleeper/process()
 	..()
 	check_nap_violations()
+	var/mob/living/carbon/C = occupant
+	if(C)
+		if(stasis && C.stat >= stasis_health)
```
adamsong, line 186: Should this not be <=

code/game/machinery/Sleeper.dm
```diff
@@ -1,3 +1,8 @@
+#define SLEEPER_TEND		"Treat Injuries"
+#define SLEEPER_ORGANS		"Repair Organs"
+#define SLEEPER_CHEMPURGE	"Purge Toxins"
+#define SLEEPER_HEAL_RATE 2
```
adamsong, line 4: You should probably reduce this back to 1, 2 is quite a lot

code/game/machinery/Sleeper.dm
```diff
@@ -157,6 +175,40 @@
...
+						var/obj/item/organ/heal_target = C.getorganslot(o)
+						if(heal_target?.damage >= 1)
+							var/organ_healing = 0.2
+							if(C.stat == DEAD)
+								organ_healing = 0.05
+							heal_target.applyOrganDamage(-organ_healing)
+							healed = TRUE
+						if(healed)
+							break
+			if(SLEEPER_CHEMPURGE)
```
adamsong, line 210: This needs a nerf, currently it heals 2 tox per tick, and purges between 1 and 4 chems per tick, likely leaning towards the 4. This is significantly better than most if not all purge chems.

adamsong (dismissed):

code/game/machinery/Sleeper.dm
SomeguyManperson: stat is weird and the values get higher at worse conditions

code/game/machinery/Sleeper.dm
SomeguyManperson: 2 is about the same as reagent healing, considering the patient can't just walk out of the sleeper after a second I don't think it needs to be much lower and the initial values were contested for being extremely slow

code/game/machinery/Sleeper.dm
SomeguyManperson: reduced purging by half, making a fully upgraded sleeper about as effective as pentetic acid

code/game/machinery/Sleeper.dm
Anvilman6: As I said above, I still think having the heal rate be lower (1) with the ability to be bumped up to typical chem healing rates (2 fir aiuri, 3 for libital) by filling it with those chems in a buffer would be a good passive / active system compromise.

SomeguyManperson: I'd rather if people want chems they have to get them from the chemist, if they need automatic chemical application they get it from cryo

burn1cycle:
I only half like them. This shouldn't replace sleepers but it would be nice if this were another machine along side sleepers with some functionality removed so they aren't too much like cryo.

*Make them their own thing and keep old sleepers
*Remove some of the healing features that are overlapping with cryo
*Hybrid them with old sleepers for the Nuke Ops and Free Miner ships & keep the cryo-like features(please! they need it!)

There's a niche here for this new equipment to fill that medbay doesn't have in equipment but trying to make it half of medbay is probably undesirable.

adamsong:
That would just be a buff for medbay, which isn't the point of the PR

burn1cycle:
Frankly, I never understood the point of adding cumbersome medical modifications to a action oriented code set to increase RP. It's a logical fallacy to assume that if we give medbay more work that they will have more time for RP(unless we're counting screaming "HURRY UP" as RP?).

Moggram:
> 
> 
> Frankly, I never understood the point of adding cumbersome medical modifications to a action oriented code set to increase RP. It's a logical fallacy to assume that if we give medbay more work that they will have more time for RP(unless we're counting screaming "HURRY UP" as RP?).

If anything, this will give more time to RP as thay have to sit when being healed.

burn1cycle:
> > Frankly, I never understood the point of adding cumbersome medical modifications to a action oriented code set to increase RP. It's a logical fallacy to assume that if we give medbay more work that they will have more time for RP(unless we're counting screaming "HURRY UP" as RP?).
> 
> If anything, this will give more time to RP as thay have to sit when being healed.

 The sleepers are in game right now so you can observe for yourself if that actually happens.

adamsong:
Its not just RP, it makes medical more interesting to play. Old sleepers made a large amount of medical 'Throw in sleeper, press gamer juice buttons, throw out of medbay, literally if you are bored enough'

burn1cycle:
> Its not just RP, it makes medical more interesting to play. Old sleepers made a large amount of medical 'Throw in sleeper, press gamer juice buttons, throw out of medbay, literally if you are bored enough'

  That system kept pace with the large amount of antags and events we have in our rounds. We're not baystation our event & antag system is action oriented and if medbay is not then it's out of place. I would suggest our antag & events systems would be out of place but only if the majority of our players agreed with that but they don't and want numerous antags and events during rounds.

SomeguyManperson:
main point of this is to let the rest of medical exist alongside sleepers rather than riding on the sleeper's coattails as effectively useless
chem mixes can heal people exceptionally fast and surgeries can heal fast as well, I've recently buffed cryo and fixed it specifically so it can get some use as well, while this pushes sleepers towards being more for all around treatment rather than something anyone can toss an injured person into then have them walk off their injuries since like 10 units of any healing chem heals at least 50 damage in its lifespan
tldr medbay positions
chemistry: extremely effective healing if anyone is assed to do it, relies on a chemist to exist and not be a shitter and someone to use or distribute the medicine
cryo: used to heal people who are severely injured if medical is being overwhelmed or to increase the efficiency of chemical mixes, cryox healing has been doubled and cryo's unreliability was addressed to make it actually usable
surgery: heals people exceptionally fast, outspeeding basic chem mixes in optimal circumstances, but requires a doctor and surgical setup, although the tools are easy to get and all doctors start with the basic tools required to use tend wounds in their starting first aid kit
sleeper: used to address unique damage or handle people with minor injuries when medical is overwhelmed, or when nothing better is available
cloning: damage is either unaddressable or too severe to be treated otherwise in a timely manner, generally only used in emergencies
healing stacks: proper tools are inaccessible or fast healing is required

basically every tool in medbay should have some use in triage, rather than the previous meta of "put everyone in a sleeper inject every chem except perf and morphine and kick them out" because if thoughtless play is optimal it'll be the only play people make

code/game/machinery/Sleeper.dm
SomeguyManperson: yeah from live testing 2 is still slow

Darkstick:
After the test merge I can safely say this changes medbay for the best. It allows doctors to use more of their intended mechanics as they become faster and more efficient than slapping some libital and calling it a day, without stopping crew from getting healing in case all the doctors suck.

BobTheButton:
> Why not just make the chemicals have to be filled manually by the chemist, along with this function? That way doctors without a chemist can function, and doctors with a chemist function even better, but still slower than before because lmao no free chems.

you overestimate chemists competence

BobTheButton:
my problem is how slow it is, it legit takes like 5 minutes to heal 100 brute damage and you cant even heal toxins and oxygen damage until 40 minutes after roundstart because of it needing parts. most of the time im in medbay with this test merged i never see anyone use the sleepers anymore. they are pretty much just wasted space at this point. please dont merge this until its less bad or slow or something god please

BobTheButton:
this sucks on ice

ToasterBiome:
> this sucks on ice

I am diagnosing you with dum dum. Don’t rely on the sleeper, there’s also surgery and cryo to work with, Sleeper now works the best for repairing organs and purging toxins, not just “powercreep better healing than cryo”. Gives sleeper it’s own purpose. Stop being incompetent in medical and medical players that take forever to heal you will stop playing medical or rapidly become competent.

pre-upgrades it tops mime off with 12 brute damage so you don’t waste time/cryo slot/chems. Post upgrades it fits its purpose

BobTheButton:
> I am diagnosing you with dum dum. Don’t rely on the sleeper, there’s also surgery and cryo to work with, Sleeper now works the best for repairing organs and purging toxins, not just “powercreep better healing than cryo”. Gives sleeper it’s own purpose. Stop being incompetent in medical and medical players that take forever to heal you will stop playing medical or rapidly become competent.
> 
> pre-upgrades it tops mime off with 12 brute damage so you don’t waste time/cryo slot/chems. Post upgrades it fits its purpose

i am diagnosing you with the cares so much about some random guys opinion that you have to write an essay insulting them explaining how they are wrong

SomeguyManperson:
I lied

LazennG:
this is how i assumed sleepers would work when i started playing spacemans, good pr

Cartlord:
Had a weird bug with this - someone had Morphine and Aranesp in their system, so I set it to "Purge Toxins", and that somehow made both the chems stop metabolizing entirely - no damage and it stayed in their body, like they were in stasis

PR #12231, opened 2021-09-02
[Port] Replaced the lobby menu (with actual art) (#60953)

adamsong:

Ports tgstation/tgstation#60953 

image
image
image
image
image

Added a countdown timer to round start

- [x] Make sprites dark mode

# Changelog

:cl:  Kryson, AMonkeyThatCodes, ported by adamsogm, darkmode sprite modifications by JamieD12
rscadd: Added a new lobby screen menu
/:cl:


Cartlord:
Frankly I already have issues with stuff loading after a server restart and this just seems like another thing that'll fail to load & force me to reconnect

adamsong:
> Frankly I already have issues with stuff loading after a server restart and this just seems like another thing that'll fail to load & force me to reconnect

Its rendering like other hud buttons, so I'm pretty sure that won't be an issue. In my tests the buttons render at the same time as the background

nesisknesis (commented):
Add Kryson to the credits as he was the one to actually sprite the UI

alexkar598:
amazing but out of tone with the rest of the game

Hopekz:
TG using this menu will make it common for a lot of places so in terms of having something that is unique (as a title screen should be) we wouldn't be accomplishing this if we merged this PR.

adamsong:
> in terms of having something that is unique

We are looking into respriting it, but I'd rather look the same by mergine than significantly worse by not merging

Hopekz:
>  I'd rather look the same by mergine than significantly worse by not merging

idk I never thought "wow this starting menu is so ugly I wish we had a different menu" at our current one.
It's more of a functionality thing. It lets me see the buttons I need to join the game (:

adamsong:
It will look bad measured against what will be produced as more people follow tg

JamieD1:
I did a dark mode.

Cartlord:
Frankly I just don't see the need for a fancy menu at all, the plain text did the job fine and it eludes me why we would need or want a new menu

maxion12345:
> Frankly I already have issues with stuff loading after a server restart and this just seems like another thing that'll fail to load & force me to reconnect

If you are talking about chat not loading you can use the fix chat verb to avoid a reconnect. It stops Lobby music though.

nesisknesis (commented):
we are discussing adapting this as there are currently a few concerns:

- Placement, the top center is possibly the oddest location to put something like this,
- Uniqueness, servers including skyrat have already started porting this, a title screen should be definitively a servers own representation I believe.
- "Flow", the new menu is very sleek and clean while the background, our current title screen, is a downscaled image that is blurry. The two should be symbiotic if we want a great title screen. 
- Style, now i'm not one to say we have an art style but in terms of design, you'd expect to see this on goonstation perhaps. It's hard to describe but it's quite high-tech has a comic feel to it.

PR will be left open for the code but may take a while for the adaptation to be done since title screens are no small undertaking.

adamsong:
> Frankly I just don't see the need for a fancy menu at all, the plain text did the job fine and it eludes me why we would need or want a new menu

It looks better, and when other servers port it, it will just make the text ui look more old and outdated

> Placement, the top center is possibly the oddest location to put something like this,

Moving it is a fairly trivial task, however the current sprites are textured with anchoring to the top in mind

> Uniqueness, servers including skyrat have already started porting this, a title screen should be definitively a servers own representation I believe.

I agree, but I don't have the spriting skills to fully replace the existing sprites, and I don't want our uniqueness to be the only screen that uses the text UI. Plus ours is dark mode and has the timer on it, which is not present in the original PR

> "Flow", the new menu is very sleek and clean while the background, our current title screen, is a downscaled image that is blurry. The two should be symbiotic if we want a great title screen.

That is the reason the original TG pr replaced the title screen as well, however I knew if I ported that as well, it would get the same complaints about uniqueness as the rest of the PR. I would be down with replacing either the button icons or the title screen, however I do not have the spriting skills for that

> Style, now i'm not one to say we have an art style but in terms of design, you'd expect to see this on goonstation perhaps. It's hard to describe but it's quite high-tech has a comic feel to it.

Given replacement sprites, swapping them out would be trivial, however I do not have the spriting skills to make a new set of sprites.

nesisknesis:
would have thought it would have been sufficiently implied but we will be doing the spritework

Hopekz:
> would have thought it would have been sufficiently implied but we will be doing the spritework

Also keep in mind that if anything needs to change for this title screen it won't be able to happen without spriters redoing it vs the old one which was just text we can program

JamieD1:
See you all in a year I guess.

Bibby0110:
Character setup press animation looks horrible

ETA is too small, makes the letters be blurry.
Glow means they're very very very blurry at times (See 8)

The 3 icons at the bottom make no sense

The info icon is the changelog?
The mail icon is the player polls??
The document icon is the CREW MANIFEST???

Bibby0110:
Can't move the sprites around like you can with the current box

adamsong:
Skill issue

PR #12251, opened 2021-09-05
Reverses colors on med doctor sec hud to make it more different from the CMO

adamsong:

Reverses the MD's color scheme to help differentiate it from the CMO

image

# Wiki Documentation

Update the sechud image for MD

# Changelog

:cl: 
tweak: tweaked medical doctor sec hud icon
/:cl:


nmajask:
I think this makes it look too much like the healthy icon from the med hud and breaks the convention of the other sec hud icons, that being the color of the shape on the icon is white, while the background and rim are the department color.

nesisknesis (commented):
I agree with nmajask, it's more consistent in its current form, and distinguishes more between a secHUD and medHUD

image
image

will close in a day or so unless given reason not to

adamsong:
It looks too similar to the CMO's though

adamsong:
I have a different idea, might also get shot down, but we will see in the PR

PR #12254, opened 2021-09-06
Applies pretty filter and OOC toggle to discord OOC channel

adamsong:

You shouldn't be able to bypass global OOC toggles or the pretty filter with disocrd

# Changelog

:cl:  
rscadd: Applied the pretty filter and OOC toggle to discord OOC
/:cl:

### DIFF
--- a/yogstation/code/datums/world_topic.dm
+++ b/yogstation/code/datums/world_topic.dm
@@ -12,6 +12,9 @@ GLOBAL_VAR_INIT(mentornoot, FALSE)
 	require_comms_key = TRUE
 
 /datum/world_topic/ooc/Run(list/input)
+	if(!GLOB.ooc_allowed)
+		return
+	input["ooc"] = pretty_filter(input["ooc"])
 	for(var/client/C in GLOB.clients)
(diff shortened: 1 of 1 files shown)


Bluedazzled:
but... but my ඞ!

Firewolf34:
THANK GOD. _ANOTHER HAND TOUCHES THE BEACON_

I worked on muting Discord-OOC with Altoids (@Altoids1 ?) back when he was Maintainer - but we never published it because there was no easy way to only mute non-Admin Discord chats using the toggle. 

Something to consider for your PR. Admin Discord messages and such from Discord were requested to us to _not_ be muted back then regardless of player preference. I don't know if that's still the case but that's why this was not done earlier. Interacting with the Discord API/bot/thing to include Discord roles into decision was complicated and required modifications to the Discord plugin.

JamieD1:
Hes right, kinda shitty that admins cant type, but eh might just be easier to do it this way without having to piss around with the plugin.

adamsong:
If admins need to type, they can hop in game

missatessatessy:
Couldn't you have a second ooc channel on the discord that only admins can access so they can bypass a mute toggle on the public one?

adamsong:
Possibly, but that is effort, and involving systems I don't understand, while this pr involves systems I think I understand.

Bibby0110 (approved):

PR #12259, opened 2021-09-06
Fixes an absolutely disgusting mapping error in Meta maints

N3D6:
Fixes a disgusting error in meta maints where there was a window, wall, and a sink all on the same tile. Removed the sink and the wall, it's just a window now.
image
This is untested but I don't see why it wouldn't work
# Wiki Documentation
Don't need
# Changelog

:cl:  
bugfix: Fixes a mapping error in meta maint
/:cl:

### DIFF
--- a/_maps/map_files/Yogsmeta/Yogsmeta.dmm
+++ b/_maps/map_files/Yogsmeta/Yogsmeta.dmm
@@ -6957,26 +6957,6 @@
 "apc" = (
 /turf/open/floor/plating,
 /area/maintenance/starboard)
-"apd" = (
-/obj/structure/table,
-/obj/item/tank/internals/emergency_oxygen{
-	pixel_x = -8
-	},
-/obj/item/tank/internals/emergency_oxygen{
-	pixel_x = -8
-	},
-/obj/item/clothing/mask/breath{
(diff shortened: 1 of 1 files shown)


patpol4 (changes requested):
It's better if you unfuckle it by placing it somewhere else instead of removing a unique sink.
 Also theres that one cobweb next to the red airlock that's just hanging on nothing, can you get rid of that?

N3D6:
> It's better if you unfuckle it by placing it somewhere else instead of removing a unique sink.
> Also theres that one cobweb next to the red airlock that's just hanging on nothing, can you get rid of that?

I'll try, but as far as I know the sink is only different by name but I'll check and see if I'm wrong

N3D6:
> It's better if you unfuckle it by placing it somewhere else instead of removing a unique sink.
> Also theres that one cobweb next to the red airlock that's just hanging on nothing, can you get rid of that?

Deleted the cobweb, as far as I know the sink isn't any different to a normal one, it just has a fancy name but if you want it to be placed somewhere else I can go ahead and knock it out

N3D6:
Gonna make this in a new PR because I'm too lazy to fix the conflicts

PR #12291, opened 2021-09-11
Changes Hand Teleporter portal to One way and makes dispelling take 2 seconds. Also max of one portal at a time. Increases Range to 20.

JamieD1:
Changes Hand teleporter to a 2 Second delay and makes it choose a turf within 20 range instead of 10.

# Wiki Documentation

Is there a hand tele page?

# Changelog

:cl:  
tweak: Changes Hand teleporter to a 2 Second delay and makes it choose a turf within 20 range instead of 10.
/:cl:

### DIFF
--- a/code/game/objects/items/teleportation.dm
+++ b/code/game/objects/items/teleportation.dm
@@ -133,7 +133,7 @@
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 30, "bio" = 0, "rad" = 0, "fire" = 100, "acid" = 100)
 	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | ACID_PROOF
 	var/list/active_portal_pairs
-	var/max_portal_pairs = 3
+	var/max_portal_pairs = 1
 	var/atmos_link_override
 
 /obj/item/hand_tele/Initialize()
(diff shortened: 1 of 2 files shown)


code/game/objects/items/teleportation.dm
```diff
@@ -157,6 +157,10 @@
 	. = ..()
 
 /obj/item/hand_tele/attack_self(mob/user)
+	to_chat(user, "You charge the hand teleporter!")
+	if(!do_after(user, 2 SECONDS))
+		to_chat(user, "You were interupted!")
```
SomeguyManperson: strange didnt know the british could english worse than i can :^

JamieD1:
Is now 20 range. One way and Time to dispel instead of create.

Tipy1802:
Seems like an over kill nerf for something that I have never even seen abused or be in need of a nerf

Marmio64:
i think a good enough nerf would just be making it have a cooldown period after being used. Its a rare, grand theft, traitor objective. It should be pretty good.

boodaliboo:
like ive certainly used this numerous times as a traitor but this nerf just makes it unusable for most everything, if i really wanted to escape sec with bullshit teleports i can crush crystals

SomeguyManperson:
atomize pls

JamieD1:
Clearly noone likes this

PR #12292, opened 2021-09-11
Round End Sounds start on "Rebooting in 10 Seconds now"

JamieD1:
Round End Sounds start on "Rebooting in 10 Seconds now"

:cl:  
tweak: Round End Sounds start on "Rebooting in 10 Seconds now"
/:cl:


Cartlord:
I like this, the round always ends too early for me to hear endround noises as-is.

adamsong (approved):
Functions in the testmerge

PR #12304, opened 2021-09-13
The stormtrooper helmet gives poor aim

adamsong:

The stormtrooper helmet gives the poor aim trait. As it should.

Jamie's idea

# Changelog

:cl:  
rscadd: Added poor aim to the stormtrooper helmet
/:cl:

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -449,6 +449,14 @@
 	strip_delay = 60
 	flags_cover = HEADCOVERSEYES
 
+/obj/item/clothing/head/helmet/stormtrooper/equipped(mob/living/user)
+	ADD_TRAIT(TRAIT_POOR_AIM, CLOTHING_TRAIT)
+	..()
+
+/obj/item/clothing/head/helmet/stormtrooper/dropped(mob/living/user)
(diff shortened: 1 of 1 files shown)


Bibby0110:
Don't nerf the 3.33k piece of armor please

Jarod1200:
Very based though maybe buff the laser armor a bit more or nerf the price since it's worse

N3D6:
This is funny

UselessTheremin:
yeah pretty funny but it would be cool if there was something to compensate for this since its just a nerf for no reason other than to be funny and the armors already super expensive

Moggram:
This be ok, but remove the slowdown on the armor so you can melee.

Marmio64:
its a neat reference but there's no way to reasonably balance it, the armors already pretty bad the only alternative would be giving it obscenely high laser resist. Removing the slowdown on the body armor also wouldn't compensate for the helmet being terrible since then it just encourages wearing a different helmet instead and buffs the armor too much. Perhaps add a toy costume variant that gives the trait instead.

Hopekz:
To everyone saying that the armor is pretty bad and expensive: Why does this armor need to be "good"?

Right now, it only really serves a purpose as a reference, not as something people would unironically use.

If anything, this armor set should be removed if we still do the "remove references" art direction thing.

PR #12311, opened 2021-09-15
Signal-Skill-Cape update! *BzZzZT*

KillerOrcaCora:
Change the color and animated the signal tech's skill cape.

Saw someone grind to the 300+ hour cape and was immensely disappointed with what they worked so hard for. The old cape does not complement what a sigtech wears at all and the symbol was just not a solid choice and was awkward.

Mines is more conformed to the yellow and green style that you see with signal techs and also includes a better symbol with a nice animation to it.  

Has been tested.

:cl:  

imageadd: Resprited Signal Tech cape to look better and complement their colors. 

/:cl:

https://user-images.githubusercontent.com/82983051/133380206-131d5052-c5e6-48ca-983c-727381fe8e52.mp4


tattax:
if you're gonna animate one of them you'll have to animate all of them, people grind just as hard on other jobs why should a specific one be better than all the other ones

Anvilman6:
I think the icon looks awesome, much better than the headset it was prior

However, I'm with Tattax on this one - the animation is a little over the top for just one cape specifically.

tattax:
also trimmed is supposed to be better than the default one but it doesn't even have the animation?

KillerOrcaCora:
> if you're gonna animate one of them you'll have to animate all of them, people grind just as hard on other jobs why should a specific one be better than all the other ones

If others want theirs animated they can come to talk to me.

KillerOrcaCora:
> also trimmed is supposed to be better than the default one but it doesn't even have the animation 

Trim is pure gold and just a different flavor and I don't really want to give what should be gold bars taped to a cape a moving animation. Ive already broke enough rules with the first cape alone.

ToasterBiome:
It would be cool if all the trimmed skillcapes had animations

Love it

Tipy1802:
Skillcapes are already rather out of place, this only makes it worst

KillerOrcaCora:
Cape overall size is taken down by request of identification!

nesisknesis (approved):

PR #12316, opened 2021-09-16
Renames microwave ovens to bluespace microwave ovens

nmajask:
# Document the changes in your pull request

As microwaves can magically put food on plates, the only logical conclusion is that it uses bluespace magic to fabricate a plate. 

How does it do it? Why does it do it? I don't know. Its magic, I don't have to explain shit.
 
_Also Jamie said I could make a meme PR, so im not going to pass up this opportunity_

# How is This Good For The Game?

# Wiki Documentation

Remove this text and write all information regarding your changes that should be known and documented through the Yogstation Wiki. 
Important documentation information includes, but is not limited to: any numerical values that have been changed, any images that have to be updated, names of specific pages that will be impacted by your changes. 

# Did Jamie Come Up With This Idea for This PR?
No

# Changelog

:cl:  
bugfix: renamed microwave ovens to bluespace microwave ovens 
/:cl:

PR #12320, opened 2021-09-17
Revert "Renames microwave ovens to bluespace microwave ovens"

Bibby0110:
Reverts yogstation13/Yogstation#12316

Less than 24 hours + a self merge (Admits it was his own idea)

### DIFF
--- a/code/modules/food_and_drinks/kitchen_machinery/microwave.dm
+++ b/code/modules/food_and_drinks/kitchen_machinery/microwave.dm
@@ -1,7 +1,7 @@
 //Microwaving doesn't use recipes, instead it calls the microwave_act of the objects. For food, this creates something based on the food's cooked_type
 
 /obj/machinery/microwave
-	name = "bluespace microwave oven"
+	name = "microwave oven"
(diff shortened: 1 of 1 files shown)


Hopekz:
> Reverts #12316
> 
> Less than 24 hours + a self merge (Admits it was his own idea)

Where is the part where the PR is bad for the game since you're reverting it

Bibby0110:
> > Reverts #12316
> > Less than 24 hours + a self merge (Admits it was his own idea)
> 
> Where is the part where the PR is bad for the game since you're reverting it

Breaks protocol.
Doesn't have to be bad for the game.

Hopekz:
> Breaks protocol.
> Doesn't have to be bad for the game.

Keep your dev drama out of our codebase.

Your change is modifying our codebase to remove something you don't even think is bad so by that logic you are knowingly making it worse if this were to be merged.

There are probably other ways you can express your point without making a modification like bringing it up in staff channels or even making a staff complaint on the forums.

nesisknesis:
why was this actually merged

PR #12335, opened 2021-09-19
Osprey description update!

KillerOrcaCora:
Gives info that the osprey does indeed overheat and won't be dumbfounded when it does in the event of laying crew out.

The syndicated page will need to be updated on the wording for ospreys. Already updated one myself in the event of this PR.
# Changelog

:cl:  

tweak: Osprey now lets people know it can overheat before purchase and after.

/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/minigun.dm
+++ b/code/modules/projectiles/guns/ballistic/minigun.dm
@@ -95,7 +95,7 @@
 
 /obj/item/gun/ballistic/minigunosprey
 	name = "M-546 Osprey"
-	desc = "An advanced minigun with an incredible rate of fire. Requires a bulky backpack to store all that ammo."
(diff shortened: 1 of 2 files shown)


Cartlord:
What if the description in the shop menu mentioned something along the lines of _manufacturer takes no responsibility for overheating_? Seems like it'd be another good way to mention it does, in fact, overheat.

KillerOrcaCora:
> What if the description in the shop menu mentioned something along the lines of _manufacturer takes no responsibility for overheating_? Seems like it'd be another good way to mention it does, in fact, overheat.

Already got the uplink one suggested by someone else but if you have another idea for the shift+click one I would gladfully change it.

Firewolf34:
If we're getting into the details with this one...  I kind of liked your old uplink description wording more, new one feels like it will explode or something after 2-6 seconds. Lol

PR #12353, opened 2021-09-21
People without science access can no longer use the research console unless they deconstruct it and reset the access requirements [#3 electric boogaloo]

Bibby0110:

Doesn’t block anyone in the current science department from researching.
Deconstruct and use a multi tool on the board to reset the access requirements

# Wiki Documentation

See title 

# Changelog

:cl:   
tweak: If you do not have science access you will now need to deconstruct the RnD console and use a multi tool on the board to reset the access requirements   
/:cl:

### DIFF
--- a/code/modules/research/rdconsole.dm
+++ b/code/modules/research/rdconsole.dm
@@ -880,7 +880,10 @@ Nothing else in the console has ID requirements.
 		say("Resynced with nearby devices.")
 	if(ls["back_screen"])
 		back = text2num(ls["back_screen"])
-	if(ls["build"]) //Causes the Protolathe to build something.
+	if(ls["build"]) //Causes the Protolathe to build something
+		if(!allowed(usr))
+			to_chat(usr, "<span class='boldwarning'>Unauthorized Access<span>")
+			return
 		if(QDELETED(linked_lathe))
 			say("No Protolathe Linked!")
 			return
@@ -889,6 +892,9 @@ Nothing else in the console has ID requirements.
 		else
 			linked_lathe.user_try_print_id(ls["build"], ls["amount"])
 	if(ls["imprint"])
+		if(!allowed(usr))
+			to_chat(usr, "<span class='boldwarning'>Unauthorized Access<span>")
+			return
 		if(QDELETED(linked_imprinter))
 			say("No Circuit Imprinter Linked!")
 			return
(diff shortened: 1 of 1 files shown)


JamieD1:
What is the point of this PR

LoliconSlayer:
> What is the point of this PR

Pretty sure it's supposed to stop people from breaking in and pressing research buttons

Bibby0110:
> What is the point of this PR

What Mord said

Bibby0110:
Maybe I’ll make it not possible to reset the access requirements if the console is locked when it’s deconstructed.

JamieD1:
> > What is the point of this PR
> 
> Pretty sure it's supposed to stop people from breaking in and pressing research buttons

But it doesnt do that if people can just turn it off..

Cartlord:
After your first two tries failed, what makes you think anyone wants this?

Bibby0110:
> After your first two tries failed, what makes you think anyone wants this?

The fact each PR has solved the problem people had with the last one?

Bibby0110:
> > > What is the point of this PR
> > 
> > 
> > Pretty sure it's supposed to stop people from breaking in and pressing research buttons
> 
> But it doesnt do that if people can just turn it off..

Which requires deconstructing the console, adding extra time required, meaning you either have time to stop them or call security

Cartlord:
This might actually be a good thing, since the time it takes someone to break into Sci + the amount of time they need to take apart the console might actually be enough time for Sec to arrive (that is, if Sec's listening to comms at all); plus, the deconstruct actions can be interrupted with shoves, so if a Scientist is present (which they SHOULD be) they can entirely keep Grey McShit from wasting research points

Bibby0110:
> This might actually be a good thing, since the time it takes someone to break into Sci + the amount of time they need to take apart the console might actually be enough time for Sec to arrive (that is, if Sec's listening to comms at all); plus, the deconstruct actions can be interrupted with shoves, so if a Scientist is present (which they SHOULD be) they can entirely keep Grey McShit from wasting research points

Entirely my thoughts.
Much less punishing to ghost roles too as they get to research. (I do disagree with them researching but people seem to want it)

ShatteredParadox:
I like. Question, Is it possible to remove the access block entirely if the pop is low enough/ no scientists sign on? My thoughts are shaft miners who need advanced mining on deadpop, as having to break in and deconstruct the machine just to get it would be a pain.

Bibby0110:
> I like. Question, Is it possible to remove the access block entirely if the pop is low enough/ no scientists sign on? My thoughts are shaft miners who need advanced mining on deadpop, as having to break in and deconstruct the machine just to get it would be a pain.

Deconstructing it doesn’t really take more than 30 seconds.
It’s mainly just to prevent people from barging in and clicking buttons while scientists are away/busy

PR #12363, opened 2021-09-23
Adds TOTP MFA for admins.

adamsong:

This prevents a client from associating with their admin datum until they authenticate with 2FA. The user must provide a one time password from an app like Google Authenticator (Other apps will work, this is just a popular one), at which point they will be authenticated and readminned. The server will save a successful CID/IP pair for up to 30 days if the user so desires, so they don't have to re-enter the 2FA login every round. 
In the event of a database failure, the last successful connection will be used.
If a user is not enrolled in 2FA when they try to readmin, they will be presented with a QR code, the raw TOTP seed, and a backup code. Most apps will be able to scan the QR code to get all the necessary information to generate 2FA codes, and the backup can be used in the event the 2FA device is lost. The server will ask for a code before saving these, to verify that the user has setup 2FA properly.
The permissions panel now requires a 2FA code to access, and can reset 2FA logins for any admin.
<details><summary>Images</summary>

Setup window
image
Google Authenticator after scanning the code
image
Entering the code
image
Save confirmation
image
</details>

Requires rust_g update with the hash feature enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ where the admin rank must be properly capitalised.
 This codebase also depends on a native library called rust-g. A precompiled
 Windows DLL is included in this repository, but Linux users will need to build
 and install it themselves. Directions can be found at the [rust-g
-repo](https://github.com/tgstation/rust-g).
+repo](https://github.com/tgstation/rust-g). The `hash` feature is required.
 
 Finally, to start the server, run Dream Daemon and enter the path to your
 compiled yogstation.dmb file. Make sure to set the port to the one you
--- a/SQL/database_changelog.txt
+++ b/SQL/database_changelog.txt
@@ -2,11 +2,25 @@ Any time you make a change to the schema files, remember to increment the databa
 
 The latest database version is 5.6; The query to update the schema revision table is:
 
-INSERT INTO `schema_revision` (`major`, `minor`) VALUES (5, 6);
+INSERT INTO `schema_revision` (`major`, `minor`) VALUES (5, 7);
 or
-INSERT INTO `SS13_schema_revision` (`major`, `minor`) VALUES (5, 6);
+INSERT INTO `SS13_schema_revision` (`major`, `minor`) VALUES (5, 7);
 
 In any query remember to add a prefix to the table names if you use one.
+
+version 5.7 23 September 2021, by adamsogm
+
+Adds table for storing mfa logins
+
+DROP TABLE IF EXISTS `mfa_logins`
+CREATE TABLE IF NOT EXISTS `mfa_logins` (
+	`ckey` varchar(32) NOT NULL,
+	`ip` int(10) unsigned NOT NULL,
+	`cid` varchar(32) NOT NULL,
+	PRIMARY KEY (`ckey`,`ip`,`cid`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+----------------------------------------------------
 
 version 5.6 14 August 2021, by alexkar598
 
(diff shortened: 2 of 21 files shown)


adamsong:
WHY SQL WHY

JamieD1:
From Ash:
1. Can we not use an existing standard for mfa, like whatever authenticator apps use? I don't think the maths behind these is super complex and it means we don't need to rely on multiple services communicating

2. Is the session preserved for longer than a round? It probably should be or else it's going to be a pain on the admins

adamsong:
> Can we not use an existing standard for mfa, like whatever authenticator apps use? I don't think the maths behind these is super complex and it means we don't need to rely on multiple services communicating

Then it becomes a pain when someone inevitably loses their 2FA device

> Is the session preserved for longer than a round? It probably should be or else it's going to be a pain on the admins

Session is based on id + cid combo, if you have logged on before from the same computer/ip then you don't have to do 2fa

JamieD1:
3. A solution that doesn't use yogbot is portable to other servers. So anything that uses the multifactor apps people already use would be great for portability

adamsong:
> A solution that doesn't use yogbot is portable to other servers. So anything that uses the multifactor apps people already use would be great for portability

Other servers have done this already, but you would end up resetting a lot of people's 2FA when they lose their device. This is already decently portable as well, as it passes the MFA off with a webhook, and receives confirmation the same way, so another server wanting a different backend could still use the code, with a different backend.

alexkar598:
This is NOT mfa. MFA is as the name implies multi factor. This means you need atleast 2 of the following:
-Something you are
-Something you know
-Something you have

This meets the something you know twice but fails to enforce any other factor in a mandatory fashion (discord supports something you have in the form of their own 2fa but an admin could have it off)

adamsong:
While yes, technically the discord is something you know instead of something you have, so is email, and depending on the feature of your phone, sms which are popular 2nd factors. I could implement something like TOTP, to get a better second, however that puts the burden of resets onto us, and requires us to either develop a robust backup system, or manually reset, requiring staff time and effort every time some forgets to transfer their codes.

Hopekz:
So admins can't come on anymore if yogbot is down?

Bibby0110:
> So admins can't come on anymore if yogbot is down?

Pretty sure admins wouldn't be able to connect for the first time on a new device.
Previous logins are cached, so you won't have to reauthenticate if you're on the same IP + CID

adamsong:
I'll swap this over to TOTP at some point, bit busy with school ATM

adamsong:
Might be good now

alexkar598 (commented):
- Give a way to clear all your sessions
- Make sessions expire after let's say 30 days
- Give a way to login without creating a session
- Add a bloody index to the table, a ckey, ip, cid index would do nicely
- (Maybe?) Refactor your code in a way that doesnt rely on /datum/admin for a general availability of the feature, doesnt have to be enabled for everyone but would be nice to have this be an option



Ideally we'd want to use a means of verifying the session by a means other than cid/ip but shrug

adamsong:
> Add a bloody index to the table, a ckey, ip, cid index would do nicely

Hmm, interesting suggestion of index, would something like this work:

https://github.com/yogstation13/Yogstation/blob/5ef221b1d1c7a40f3d0c6b7c6f64cdc2cd2e32b9/SQL/database_changelog.txt#L24

AffectedArc07:
Why is this even up for debate
image

adamsong:
Enforcing 2FA for forums admins was a council vote (two actually), so I'm just keeping up with the tradition.

JamieD1 (approved):

JamieD1:
Database updated

PR #12367, opened 2021-09-24
Adds the pretty filter to NT Chat Client user inputs

Altoids1:
image

## Summary

I was looking at this chat program while working on João, as an example of how to have a good GUI, and I noticed that a lot of the user inputs (changing the channel name, user name, or just sending messages) did not use the pretty filter.

So, I added some checks, via my ``isnotpretty()`` proc. Simple as that.

## Changelog

:cl:  Altoids
bugfix: The NT Chat Client now uses the pretty filter. Spacemen can now longer throw slurs at each other through this platform.
/:cl:


code/modules/modular_computers/file_system/programs/ntnrc_client.dm
```diff
@@ -36,6 +36,12 @@
 			var/message = reject_bad_text(params["message"])
 			if(!message)
 				return
+			//yogs start
+			if(isnotpretty(message))
+				to_chat(usr, "<span class='notice'>Your fingers slip. <a href='https://forums.yogstation.net/index.php?pages/rules/'>See rule 0.1.1</a>.</span>")
```
adamsong: Its just rule 0.1, and the link is https://forums.yogstation.net/help/rules/

adamsong (changes requested):

code/modules/modular_computers/file_system/programs/ntnrc_client.dm
Altoids1: I copied most of this from what's in the say code; consider updating that text as well.

adamsong (approved):

PR #12394, opened 2021-10-02
Makes metallic hydrogen golems work like real golems which are space immune + adds some appropriate attributes

Sniblet:

Thanks to TG, they used to not even be spaceproof, lol
Now they have slightly better armor than diamond, slightly worse stunmod than plasteel, and don't have NOFIRE so they can carry fire around because they're made of hydrogen and are from the fire department
and their other inherent_traits are not worse than iron golems
added some special_names whatever that means, and made their prefix one word because every other golem is like that
Not tested, should be too simple to screw up though
(just you watch me)

Because as it is, metallic hydrogen golems are... worse than iron golems with antimagic, which is really sad for about the hardest non-antag golem to make

#### Changelog

:cl:  
tweak: made mhydrogen golems actually spaceproof, flammable but still fire-immune, superbly armored, and gave them special_names for freegolems
/:cl:


Mqiib:
image

Sniblet:
shut up

SomeguyManperson (approved):

PR #12427, opened 2021-10-09
updates material exports to slow down atmos bicycle speedrun any%s and gives an estimated maximum value comment for each material export for powergaming

SomeguyManperson:

turns out I forgot to add a limit to metal hydrogen and zaukerite so I have totally massacred them

# Wiki Documentation

export limits for several materials edited:
bananium: 100 (from 50)
diamond: 150 from 100
titanium: 500 from 0
hot ice: 500 from 250
metal hydrogen: 1000 from 0
zaukerite: 1000 from 0 

:cl:  
tweak: several materials have had their export prices modified to make them either properly competitive with other materials of their category or to actually give them ones because I forgot to
/:cl:

### DIFF
--- a/code/modules/cargo/exports/materials.dm
+++ b/code/modules/cargo/exports/materials.dm
@@ -31,13 +31,13 @@
 
 /datum/export/material/bananium
 	cost = 2500
-	export_limit = 50
+	export_limit = 100 //250k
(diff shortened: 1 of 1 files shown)


adamsong (approved):
Changes are as described, not sure about the balance though

PR #12428, opened 2021-10-09
removes the check for ghosts being still in game for soul departed since we force ghosts to re-enter their body when reviving them and it made me look stupid once

SomeguyManperson:

soul departed is like catatonic it's for COMPLETELY IRRECOVERABLE corpses or some kind of shenaniganry not "i left the game for 2 seconds and am now being turned into a hamburger by the cook"

# Changelog

:cl:  
bugfix: yes this is a fucking bugfix: you will no longer have confusion between someone who has logged out of the game as a ghost and a body that has no player inhabitant at all, because one of these is recoverable and the other is extremely not recoverable and they SHOULDN'T LOOK THE EXACT SAME
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -114,7 +114,7 @@
 		if(hellbound)
 			. += span_warning("[t_His] soul seems to have been ripped out of [t_his] body.  Revival is impossible.")
 		. += ""
-		if(getorgan(/obj/item/organ/brain) && !key && !get_ghost(FALSE, TRUE))
+		if(getorgan(/obj/item/organ/brain) && !key && !get_ghost())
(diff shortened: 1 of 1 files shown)


maxion12345:
so this means if someone leaves it won't be "soul departed" anymore?

SomeguyManperson:
specifics on soul departedness
dead body
ghost is present in body: not soul departed
ghost is out of body but player is connected: not soul departed
ghost is present in body but player is disconnected: not soul departed
ghost is out of body and player is disconnected: soul departed
no player is attached to the body (dnr, controlling another body i.e. ghostrole taken, etc): soul departed

this makes ghost out of body player disconnected considered NOT soul departed because someone can end up leaving the game while observing and COME BACK

adamsong (approved):
Good change

PR #12438, opened 2021-10-11
Fixes Novaflower and death nettles never hitting their attack proc and also nerfs death nettles

ToasterBiome:

Novaflowers and death nettles don't work properly: sometime between now and the past 6 years the attack proc has been changed so they do not properly trigger. The obj/proc/attack trigger returns false if it is successful and true if it is cancelled, so it will always return early and never do their special effects. It may have been broken the entire time yogs has existed, it's possible.

Only triggers their special effect (fire and paralyze/knockout for each) on harm intent. Still does their force damage in any other intent.

image

ok i nerfed the knockout/paralyze by making it half as strong

# Wiki Documentation

None probably

# Changelog

:cl:   
bugfix: made novaflowers and death nettles actually trigger attack effects
rscdel: deathnettles no longer have the capacity to stun people
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/nettle.dm
+++ b/code/modules/hydroponics/grown/nettle.dm
@@ -103,7 +103,9 @@
 			to_chat(user, span_userdanger("You are stunned by the Deathnettle as you try picking it up!"))
 
 /obj/item/reagent_containers/food/snacks/grown/nettle/death/attack(mob/living/carbon/M, mob/user)
-	if(!..())
+	if(..())
+		return
+	if(user.a_intent != INTENT_HARM)
 		return
 	if(isliving(M))
 		to_chat(M, span_danger("You are stunned by the powerful acid of the Deathnettle!"))
@@ -111,6 +113,6 @@
 
 		M.adjust_blurriness(force/7)
 		if(prob(20))
-			M.Unconscious(force / 0.3)
-			M.Paralyze(force / 0.75)
+			M.Unconscious(force / 0.6)
+			M.Paralyze(force / 1.5)
 		M.drop_all_held_items()
--- a/code/modules/hydroponics/grown/nettle.dm
+++ b/code/modules/hydroponics/grown/nettle.dm
@@ -108,11 +108,7 @@
 	if(user.a_intent != INTENT_HARM)
 		return
 	if(isliving(M))
-		to_chat(M, span_danger("You are stunned by the powerful acid of the Deathnettle!"))
+		to_chat(M, span_danger("You are blinded by the powerful acid of [src]!"))
 		log_combat(user, M, "attacked", src)
 
 		M.adjust_blurriness(force/7)
-		if(prob(20))
-			M.Unconscious(force / 0.6)
-			M.Paralyze(force / 1.5)
-		M.drop_all_held_items()
--- a/code/modules/hydroponics/grown/flowers.dm
+++ b/code/modules/hydroponics/grown/flowers.dm
@@ -225,7 +225,9 @@
 	force = round((5 + seed.potency / 5), 1)
 
 /obj/item/grown/novaflower/attack(mob/living/carbon/M, mob/user)
-	if(!..())
+	if(..())
+		return
+	if(user.a_intent != INTENT_HARM)
 		return
 	if(isliving(M))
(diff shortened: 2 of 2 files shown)


Marmio64:
i dont think this is needed, you already die in like 3 hits from deathnettles, dont need to make it where there's also a chance you're knocked unconcious from the first hit so you dont even have a chance to do anything against the other 2 hits. At the very least remove the unconcious part.

ToasterBiome:
> i dont think this is needed, you already die in like 3 hits from deathnettles, dont need to make it where there's also a chance you're knocked unconcious from the first hit so you dont even have a chance to do anything against the other 2 hits. At the very least remove the unconcious part.

this is a good point, maybe it should be removed since it’s comically easy to hit 100 potency with modern botany, aside from the fact that this is actually a bug and needed fixing. It’s not like i’m adding them, somebody very naughty fucked up their programming many years ago and i’m setting it straight

code/modules/hydroponics/grown/nettle.dm
```diff
@@ -103,14 +103,16 @@
...
 		to_chat(M, span_danger("You are stunned by the powerful acid of the Deathnettle!"))
 		log_combat(user, M, "attacked", src)
 
 		M.adjust_blurriness(force/7)
 		if(prob(20))
-			M.Unconscious(force / 0.3)
-			M.Paralyze(force / 0.75)
+			M.Unconscious(force / 0.6)
+			M.Paralyze(force / 1.5)
 		M.drop_all_held_items()
```
SomeguyManperson:
```suggestion
```

code/modules/hydroponics/grown/nettle.dm
```diff
@@ -103,14 +103,16 @@
...
 	if(isliving(M))
 		to_chat(M, span_danger("You are stunned by the powerful acid of the Deathnettle!"))
 		log_combat(user, M, "attacked", src)
 
 		M.adjust_blurriness(force/7)
 		if(prob(20))
-			M.Unconscious(force / 0.3)
-			M.Paralyze(force / 0.75)
+			M.Unconscious(force / 0.6)
+			M.Paralyze(force / 1.5)
```
SomeguyManperson:
```suggestion
```

code/modules/hydroponics/grown/nettle.dm
```diff
@@ -103,14 +103,16 @@
...
 		return
 	if(isliving(M))
 		to_chat(M, span_danger("You are stunned by the powerful acid of the Deathnettle!"))
 		log_combat(user, M, "attacked", src)
 
 		M.adjust_blurriness(force/7)
 		if(prob(20))
-			M.Unconscious(force / 0.3)
-			M.Paralyze(force / 0.75)
+			M.Unconscious(force / 0.6)
```
SomeguyManperson:
```suggestion
```

code/modules/hydroponics/grown/nettle.dm
```diff
@@ -103,14 +103,16 @@
...
+	if(..())
+		return
+	if(user.a_intent != INTENT_HARM)
 		return
 	if(isliving(M))
 		to_chat(M, span_danger("You are stunned by the powerful acid of the Deathnettle!"))
 		log_combat(user, M, "attacked", src)
 
 		M.adjust_blurriness(force/7)
 		if(prob(20))
```
SomeguyManperson:
```suggestion
```

code/modules/hydroponics/grown/nettle.dm
```diff
@@ -103,14 +103,16 @@
 			to_chat(user, span_userdanger("You are stunned by the Deathnettle as you try picking it up!"))
 
 /obj/item/reagent_containers/food/snacks/grown/nettle/death/attack(mob/living/carbon/M, mob/user)
-	if(!..())
+	if(..())
+		return
+	if(user.a_intent != INTENT_HARM)
 		return
 	if(isliving(M))
 		to_chat(M, span_danger("You are stunned by the powerful acid of the Deathnettle!"))
```
SomeguyManperson:
```suggestion
		to_chat(M, span_danger("You are blinded by the powerful acid of [src]!"))
```

SomeguyManperson (approved):

PR #12444, opened 2021-10-11
Lets accessories that aren't on uniforms have their equip/unequip procs called when the clothing item is equipped or unequipped

ToasterBiome:

Accessories have a proc that triggers when you equip or unequip the clothing item they're attached to, for some reason they're relegated to jumpsuits only so making it generically clothing it opens more opportunities for extension.

Currently only the lawyer badge and pandora pin thing make use of it.

Hopek said this would take 1-3 hours

# Changelog

:cl:  
tweak: clothing now has a more generic proc to handle on equip/unequip with accessories on them
/:cl:

### DIFF
--- a/code/modules/clothing/under/_under.dm
+++ b/code/modules/clothing/under/_under.dm
@@ -101,13 +101,13 @@
 		H.update_inv_w_uniform()
 //Yogs End
 	if(attached_accessory && slot != SLOT_HANDS)
-		attached_accessory.on_uniform_equip(src, user)
+		attached_accessory.on_clothing_equip(src, user)
 		if(attached_accessory.above_suit)
 			H.update_inv_wear_suit()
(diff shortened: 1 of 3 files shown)


Hopekz:
Looking at the code now, I don't mean to take away from your hard work but this isn't what I said would take 1-3 hours.
The thing being described was a visual multi-accessory system that had checks and balances on what can be worn together.

This new system would allow you to wear multiple medals on your sprite by shifting them over so all your medals would display etc. It would have extra checks to make sure you can't wear multiple arm bands unless they could all be displayed (enough room)

You'd have a certain amount of slots for visual accessories so it doesn't just continue forever but you'd be able to put up to 3 of them on at once.

image

JamieD1 (approved):

PR #12452, opened 2021-10-12
Ports Foldable Wheelchairs from TG

Bluedazzled:
Original PR: [https://github.com/tgstation/tgstation/pull/53796](https://github.com/tgstation/tgstation/pull/53796)

# Wiki Documentation
ehhh???
nah...

# Changelog
:cl:  CoffeeDragon16
rscadd: Added foldable wheelchairs.
/:cl:


Marmio64:
honestly dont think this is necessary when we have roller beds and skateboards already. whats the target audience for foldable wheelchairs anyhow? the people who are supposed to be using them can barely move without them and the only other thing they're good at is fire extinguisher moving, which i think you can do with rollerbeds already.

Hopekz:
> honestly dont think this is necessary

Your argument is very subjective. Why does everything that gets added need to be necessary?
An argument should be like "This is bad because ___" or "This is against game direction because __ "

Adding more variety to items isn't a bad thing in my opinion. Maybe someone will want to carry an extra wheelchair with them we never really know what people will end up doing with items we add to the game.

Marmio64:
> > honestly dont think this is necessary
> 
> Your argument is very subjective. Why does everything that gets added need to be necessary? An argument should be like "This is bad because ___" or "This is against game direction because __ "
> 
> Adding more variety to items isn't a bad thing in my opinion. Maybe someone will want to carry an extra wheelchair with them we never really know what people will end up doing with items we add to the game.

Yes, the point is that its my subjective opinion, which is why i didnt dislike the PR . Sometimes you want to just give your opinion and actually have one.

SomeguyManperson (approved):

PR #12462, opened 2021-10-15
Expands aux base construction to include a place for building shuttles

nmajask:

Expands Aux Base Construction to include a place to build shuttles. It has also been merged with port bow solars, so I had to mess with maints a bit to get it to work. The mining shuttle also docks to the open area instead of inside the base construction area. The amount of wires needed to wire the solar panels is still 40, tough it is a bit of a walk. There is also some supplies for building shuttles in a storage room next to the solars room. I have not tested this, so some things may break because of the scale of the changes and my lack of experience.

### New Aux Base Construction
image

image

image

### Starting Boards
image
image

# Wiki Documentation

Many images will need to be changed, and if there is a page for aux base construction that will need to be updated too. 

# Changelog

:cl:  
tweak: expanded aux base construction to include a place for building shuttles
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -8406,13 +8406,6 @@
 /obj/effect/spawner/lootdrop/maintenance,
 /turf/open/floor/plating,
 /area/maintenance/starboard/aft)
-"axm" = (
-/obj/structure/lattice/catwalk,
-/obj/effect/turf_decal/stripes/corner{
-	dir = 1
-	},
-/turf/open/space/basic,
-/area/solar/port/fore)
 "axo" = (
 /obj/structure/cable{
 	icon_state = "1-2"
@@ -20978,19 +20971,6 @@
 	},
 /turf/open/floor/plasteel/white,
 /area/science/lab)
-"bmR" = (
-/obj/structure/lattice/catwalk,
-/obj/machinery/atmospherics/pipe/simple/supply/hidden/layer2{
-	dir = 6
-	},
-/obj/machinery/atmospherics/pipe/simple/purple/visible{
-	dir = 4
-	},
-/obj/effect/turf_decal/stripes/line{
-	dir = 1
-	},
-/turf/open/space/basic,
-/area/solar/port/fore)
 "bmS" = (
 /obj/machinery/atmospherics/pipe/simple/supply/hidden/layer2{
 	dir = 5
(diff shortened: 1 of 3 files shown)


maxion12345:
Some of the items placed are items that require research to be completed, not sure if we'll want to bypass the shuttle research. It'd be the equivalent of putting space pod parts in the pod bay, I suppose.

nmajask:
> Some of the items placed are items that require research to be completed, not sure if we'll want to bypass the shuttle research. It'd be the equivalent of putting space pod parts in the pod bay, I suppose.

All you get is the parts to build a very small shuttle that doesnt have docking computer. You would need to research to be able to dock anywhere that isn't the whiteship dock.

ToasterBiome:
I wish the aux base construction area was moved more to the left and having a little 4x3 room or something so the room can be open to the public, doesn't seem anything in there besides the aux base stuff that would get yoinked.

ToasterBiome:
What about my ideas OwO

patpol4 (changes requested):
1. Are these all just fuel ports? also check the indoor one, also they need dirs to connect to the pipes
image
2.  is this supposed to cut off like that? seems weird, either do a corner or attach something to them
image
3. Some of the solar turfs have double cables on them, double check please
image
4. The stripe decals won't be visible on catwalks so remove those
5. double decal and missing firelocks here, also remember to add them on the shutter and on the red airlock to the right
image
6. this isn't engineering secure storage
image
7. This is an oversight right?
image

I think this is it but i might have missed some, the rest will come out after merge probably

nmajask:
> 1. Are these all just fuel ports? also check the indoor one, also they need dirs to connect to the pipes

Fixed

> 2. is this supposed to cut off like that? seems weird, either do a corner or attach something to them

I made them cut off at the corner.

> 3. Some of the solar turfs have double cables on them, double check please

Double checked most of them, and fixed the ones I found.

> 4. The stripe decals won't be visible on catwalks so remove those

Done

> 5. double decal and missing firelocks here, also remember to add them on the shutter and on the red airlock to the right

I assume you mean the left one, in which case its fixed.

> 6. this isn't engineering secure storage

Fixed

> 7. This is an oversight right?

Fixed, though I had to do some code changes to allow the shuttles to dock with it. The original point of the two were to allow the mining shuttles to dock with it without code changes, but I guess its unavoidable at this point.

patpol4 (approved):
Yes, sorry I meant left

PR #12469, opened 2021-10-17
buffs lightning bolt spell

Marmio64:
reduces cooldown by 5 seconds and cost by 1

previously cooldown was 20 seconds and cost was 3
the spell doesnt knock people down and only does like 20 burn damage, which is terrible. for one spellpoint less you can just instacrit someone with a fireball that has a 8 second cooldown

edit: cooldown is now 12 seconds and it is robeless

:cl:  
tweak: lightning bolt spell cooldown reduced by 8 and cost reduced by 1, also made robeless
/:cl:

### DIFF
--- a/code/modules/antagonists/wizard/equipment/spellbook.dm
+++ b/code/modules/antagonists/wizard/equipment/spellbook.dm
@@ -213,7 +213,6 @@
 /datum/spellbook_entry/lightningbolt
 	name = "Lightning Bolt"
 	spell_type = /obj/effect/proc_holder/spell/aimed/lightningbolt
-	cost = 3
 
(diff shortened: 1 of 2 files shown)


Jarod1200:
Would be nice to see this spell actually used

Hopekz:
If you're buffing it, would be nice to see it around fireball level.
Imo this buff isn't enough yet

Marmio64:
12 second cooldown and it is robeless. Still has twice the cooldown of fireball and is unable to instacrit people or destroy objects, but its more comparable now.

SomeguyManperson (approved):

PR #12472, opened 2021-10-18
[s] Force update to a version without known hacks

adamsong:
514.1566 has a know set of clientside hacks, they don't work on 514.1568. This is very much a bandaid fix (All it takes it the hacks to update) but people have used them on the server, so this will at least stop them for a bit.

# Changelog

:cl:  
tweak: minimum allowed version is now 514.1568
/:cl:


Bibby0110:
As said on Discord this doesn't actually do much more than inconvenience our players.

It forces everyone to update (which is the vast majority) and will be made useless by hacks updating soon.
AFAIK we only really do this if it's version specific exploits, since we know those don't just come back.

adamsong:
I'll work on detecting the hack then

PR #12492, opened 2021-10-21
[s] Added a check to make sure you can see the think you clicked on

adamsong:
If you click on something and its invisibility is higher than your see_invisible you get banned. Need to test to make sure we don't get false positives.

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -928,6 +928,12 @@ GLOBAL_LIST_INIT(blacklisted_builds, list(
 	var/ab = FALSE
 	var/list/L = params2list(params)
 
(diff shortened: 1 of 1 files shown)


Hopekz:
This is weird. Just automatically giving a ban if this is triggered.

If the middle click aimbot exploit note just auto banned half the server would be gone.

IMO bans should be given by actual humans making the ban decision unless it is an anti griefing tool like "walked in to SM first 10 minutes in the server" etc

adamsong:
I haven't gotten around to testing this yet, but if it works how I think it will, the only way for this to trigger is by clicking on something you can see, but shouldn't, which is either the result of a gamebreaking bug (Pipes, ghosts, etc would be visible), or hacking.

JamieD1 (approved):

PR #12504, opened 2021-10-22
Makes clock cultists visible on the orbit window

nmajask:

Makes clockcultists visible on the orbit window, due to them having many messages visible to ghosts that give them away and are similar to nukies which are ghost visible . If it is preferred, I can change it to only after the recall.

# Wiki Documentation

None needed. 

# Changelog

:cl:  
tweak: clock cultists are now visible on the orbit window for ghosts
/:cl:


Jarod1200:
Should be only after recall so people can't die and get revived and know who the entire cult is

nmajask:
> Should be only after recall so people can't die and get revived and know who the entire cult is

You can already tell based on chat and various other things. Maybe it would be better to have it only show if you DNR or something.

JamieD1 (approved):

PR #12506, opened 2021-10-22
Fix pacifism throwing

maxion12345:

hey so in a recent round we pacified some guy and then he killed me by throwing me against a wall over and over. turns out someone forgot to put a return, so you would both "gently let go" and yeet them at mach 5. anyways this fixes that so you can actually pacify traitors and not be forced to execute anyone with cqc/carp :)
# Wiki Documentation

# Changelog
:cl:  
bugfix: fixed pacifism allowing you to throw people
/:cl:


TheLinuxOS:
Man I wonder who possibly could be the cause of this….

maxion12345:
> Man I wonder who possibly could be the cause of this….

I've turned my salt into a bugfix
good ending


alternative was to shit piss and cry on the forums

JamieD1 (approved):

PR #12520, opened 2021-10-24
Should actually fix ID console

nmajask:

I cant code. I made custom names overwrite the original assignment like an idiot. This should fix it.

# Wiki Documentation

None needed. 

# Changelog

:cl:   
bugfix: fixed custom jobs when applied through the card console break stuff, I hope
/:cl:


Bibby0110:
This doesn't actually functionally change anything?

nmajask:
> This doesn't actually functionally change anything?

It should do something. The change should make it so if someone adds a custom job name to an ID card it only changes the assignment and not both the original assignment and assignment, though the code for the ID mod console is a bit hard to read.

JamieD1 (approved):

PR #12546, opened 2021-10-27
Ports "Removes move delay when damaged" #47702

nmajask:

Ports ["Removes move delay when damaged" #47702](https://github.com/tgstation/tgstation/pull/47702) by kriskog from TG. Removes some stuff that was broken, which fixes a few issues. Also, according to Mqiib, this might fix being shot while slow causing you to not move or something but I dont know if thats true. Untested, though as it is a small change it may work.

# Wiki Documentation

None needed. 

# Changelog

:cl:  Skoglol
bugfix: Lava no longer slows you down. 
bugfix: Getting hurt no longer adds a move delay.  
bugfix: Lava will no longer prevent you from ghosting out of your dead body.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1217,7 +1217,6 @@
 	if(!(mobility_flags & MOBILITY_UI))
 		unset_machine()
 	density = !lying
-	var/changed = lying == lying_prev
 	if(lying)
 		if(!lying_prev)
 			fall(!canstand_involuntary)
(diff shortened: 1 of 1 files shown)


Cartlord:
I assumed the lava slowdown was intentional TBH, since it'd likely be a pain in the ass to walk through molten rock - I imagine it'd be awfully treacherous terrain

nmajask:
> I assumed the lava slowdown was intentional TBH, since it'd likely be a pain in the ass to walk through molten rock - I imagine it'd be awfully treacherous terrain

Assuming that it wasn't removed (which I don't see why it would be), according to the original PR lava already slows you down while walking through it, the slowdown that is removed is an unintended slowdown caused by the damage being dealt breaking your movement. The change log is a bit misleading, and I may change it later.

JamieD1:
Isnt lava a thick liquid though? Wouldnt it slow you down?

Bibby0110 (approved):

PR #12549, opened 2021-10-27
Makes most belts bulky

nmajask:

Makes most belts bulky instead of normal, with the only exceptions being the sec belt and the multi-belt (both of which are because mqiib told me not to). The important difference that this make is that you cant put belts in normal backpacks. Also adds a path for the multi-belt so it isn't just a weird var-edited thing on meta. Untested, should work.

## Why?

With current belts, being able to put them in backpacks is very important due to the massive amounts of items they can hold. This often leads to people having almost every single tool they will ever need in their backpack and, due to them not being worn, you cant tell if someone even has it on them so you need to search their backpack to see if they have a full set of tools on them. Allowing belts to be put in backpacks also makes the question if you should have a tool on you less of a question of if you have enough space and more of a question if you have the belt that you can put it on, as you can put a medical belt, toolbelt, and sec belt in a satchel and still have room for two boxes.

# Wiki Documentation

If it is mentioned that belts are normal sized items or can be put in a backpack, that should probably be changed. 

# Changelog

:cl:    
tweak: changed most belts from being normal sized items to bulky sized items
/:cl:

### DIFF
--- a/_maps/map_files/Yogsmeta/Yogsmeta.dmm
+++ b/_maps/map_files/Yogsmeta/Yogsmeta.dmm
@@ -67590,17 +67590,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/hallway/secondary/entry)
-"cZo" = (
-/obj/structure/easel,
-/obj/structure/rack,
-/obj/item/storage/belt{
-	desc = "Can hold quite a lot of stuff.";
-	name = "multi-belt"
-	},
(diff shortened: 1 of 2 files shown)


patpol4:
The downvotes are from powergamers

patpol4 (approved):
Map is fine

Iplayforss13:
Just put your tools in a box

PR #12554, opened 2021-10-29
Makes borgs emagged by a ninja know Japanese and have the ninja override lawset

nmajask:

Makes borgs emagged by a ninja know Japanese and use the ninja override lawset. Did this quickly and was not able to test it, so it may or may not work. It builds at least so thats something.

# Wiki Documentation

May want to put it as a tip that borgs learn Japanese when emagged. 

# Changelog

:cl:  
tweak: tweaked a few things  
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_defense.dm
+++ b/code/modules/mob/living/silicon/robot/robot_defense.dm
@@ -137,10 +137,6 @@
 	lawupdate = FALSE
 	set_connected_ai(null)
 
-	if(user.mind?.has_antag_datum(/datum/antagonist/ninja))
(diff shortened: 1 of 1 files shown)


Xoxeyos:
Not sure if you're up to it, but there's a specific ninja lawset that is in the code, that should be introduced when the cyborg is emagged by the ninja, but I don't believe it is at the moment.

nmajask:
> Not sure if you're up to it, but there's a specific ninja lawset that is in the code, that should be introduced when the cyborg is emagged by the ninja, but I don't believe it is at the moment.

image
ah, there is

JamieD1 (approved):

PR #12555, opened 2021-10-29
Reinforces medbay windows on box

nmajask:

Reinforces medbay windows on box. Why? To be more consistent with the other departments and also I am tired of people breaking in all the time.

image

# Wiki Documentation

Medbay images will need to be changed. 

# Changelog

:cl:  
tweak: reinforced the medbay windows on box  
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -58754,6 +58754,14 @@
 	icon_state = "panelscorched"
 	},
 /area/maintenance/aft)
+"tIu" = (
+/obj/machinery/door/firedoor/border_only{
+	dir = 1
+	},
+/obj/machinery/door/firedoor/border_only,
+/obj/effect/spawner/structure/window/reinforced,
+/turf/open/floor/plating,
+/area/medical/medbay/lobby)
 "tIN" = (
 /obj/machinery/door/airlock/maintenance{
 	name = "Custodial Maintenance";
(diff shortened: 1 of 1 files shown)


patpol4:
Medbay does not need to be consistent with other departments as it is medbay, the most public department that everyone ends up in at some point. 
As for the window breaking, it's an admin issue.

nmajask:
> Medbay does not need to be consistent with other departments as it is medbay, the most public department that everyone ends up in at some point. As for the window breaking, it's an admin issue.

The person who suggested I make this PR said this was their reasoning:

> 1) Cargo and botany have reinforced windows. Cargo part that is reinforced is just the exterior, which isn't even the cargo bay where crates are loaded. Botany, while it can be dangerous, should be less restrictive than medical, the place where dead people (sometimes officers and heads of staff) are coming
> 2) A lot of things in medical are rated to subdue people
> 3) Medical is a high-prio target for any enemies to the station, meaning that the lack of reinforced windows makes it all the easier to walk in, shoot a doctor operating, steal a body, or blow the place up. Makes zero sense for botany to have this treatment and medical not.
> 4) The main and only argument I can see for normal windows is getting in without any doctors there to save someone. But, there are emergency first aid kits in lockers around the station. Can you just break into a hospital in real life?
> 5) Only main department without reinforced windows, needs better justification that
> 6) Lack of reinforced windows also makes it super easy for shitters to get in and whine and clog up space or otherwise disrupt work

As for saying breaking windows is an admin issue, I don't think that its a good reason not to. Just because a problem can be solved by admins enforcing the rules more doesn't mean we shouldn't also make code and mapping changes to also help dissuade people from breaking the rules.

patpol4:
Let me explain the the main reason behind my hesitance as I believe it is important to understand why I'm so against something that is obviously *good*.
 **The map shouldnt be perfect on roundstart**, it is the reason why every department isn't encased in rwalls and rwindows.
Im not talking about this specific situation here but generally about the flawed map theory. Here are some reasons:
It gives engineers something to do, allows for more interactions between players, goes well with the sandbox nature of the game, increases variety in antag play, **prevents powercreep**.

1. The cargo exterior part is still less public than medbay, as for botany they are reinforced since nobody should go in there anyway and they get a little privacy, their windows are essentially transparent walls, not windows.

2. Yes, every department has their department specific items, for medbay it happens to be chemicals, it doesnt really change anything imo.

3. That's a good thing

4. Adressed in preface

5. It has reinforced windows, just not in the lobby.

6. Yes, I hate it, I absolutely agree with this point, but it is necessary in the bigger picture.

PR #12574, opened 2021-11-01
Makes Ineloquent quirk more lenient, along with some added variance

HugeBalloonGodGamer:

- Increases character limit from 4 to 7
- Ignores words with 4 or less unique characters regardless of length
- Some added randomness to speech manipulation
- Nullifies the effects temporarily while the player is drunk

Given that this quirk is only 1pt, the 4 character limit feels disproportionately restrictive, especially with how the rest of your input is discarded moment it finds a "too complicated" word. I upped the character limit to make it more bearable and tried adding more variety to the speech side effect as well. The part where being drunk makes it go away was just for fun, the idea was that the user will be speech-impaired either way.

# Wiki Documentation
Still works as advertised on the wiki

# Changelog
:cl:  
tweak: make Ineloquent more lenient
/:cl:

### DIFF
--- a/code/datums/brain_damage/mild.dm
+++ b/code/datums/brain_damage/mild.dm
@@ -192,7 +192,36 @@
 
 	var/static/list/common_words = world.file2list("strings/1000_most_common.txt")
 
+// Returns false if word has more than 4 distinct letters
+/datum/brain_trauma/mild/expressive_aphasia/proc/is_simple(word)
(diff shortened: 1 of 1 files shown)


HugeBalloonGodGamer:
First time doing stuff in Byond, hope I didn't overlook anything

JamieD1 (approved):

PR #12533, opened 2021-10-26
[PORT] [READY] Kitchen Rework part 1: I just want to griddle!

ToasterBiome:
# Document the changes in your pull request

Ports https://github.com/tgstation/tgstation/pull/55319

This PR essentialy moves away from the extremely microwave dependent cooking we have for meat right now, and making it a bit more sensical by making you use a grill to grill meat. The grill takes a different time (with variation) for different grilled things. Once finished it will turn that food into something else.

Yes, this does mean creating burgers takes longer, but in return you can make more patties at once (YOGS: with upgrades), and you are not required to stay at the grill while its going. This lets you cook as much as you want at once, just make sure your meat doesn't burn!

This includes uncraftable chicken patties/meatballs, for the future

Maps:

Box (you get the idea):
image

Meta:
image

Delta: 
image

Eclipse (is this the only cook area on eclipse?):
image

Kilo:
image

# Wiki Documentation

The recipes will have to be changed on the wiki, basically all burgers that require slabs require patties now. Look at the changelog to gleam

# Changelog

:cl:  
add: adds protein nutriment
add: Adds the griddle, a panless stove/grill for you to bake your meat products on. Nanotrasen has decided to move away from microwaving everything
add: added them to the maps
tweak: Burgers are now made from patties, patties are grilled on the griddle, raw patties are made by using a rolling pin on a raw meatball
tweak: Most meat is grilled on the griddle now instead of microwaved
sprite: new meatball, patty and sausage sprites
/:cl:

PR #12577, opened 2021-11-01
[BUGFIX] Fixes some inconsistencies with the griddle

ToasterBiome:

regarding https://github.com/yogstation13/Yogstation/pull/12533

Thank you to @Cartlord for showing me this bug

Fixes some bugs in griddle/food code. Meat steaks were missing some variables along with cutlets so they would disappear pretty quickly, now they will properly turned to burnt messes

Also adds code to the bonfire grill to handle grilling things, it functions in the same way that griddles do but worse, as it's more primitive. This is because internally the grill would have a chance to microwave the item, and these items no longer have microwave "recipes", so if you tried to do it before this bugfix they would instantly burn.

# Changelog

:cl:  
bugfix: made it so meat steaks and cutlets don't instantly disappear if you leave them on for too long
bugfix: added code in the bonfire grill to grill things like a worse griddle (since they would turn into burned mess otherwise)
/:cl:


patpol4 (approved):

code/modules/hydroponics/grown/towercap.dm
```diff
@@ -268,12 +268,19 @@
 /obj/structure/bonfire/proc/Cook()
 	var/turf/current_location = get_turf(src)
 	for(var/A in current_location)
+		var/obj/G = A;
```
Bibby0110:
```suggestion
		var/obj/G = A
```

code/modules/hydroponics/grown/towercap.dm
```diff
@@ -268,12 +268,19 @@
...
 		else if(isliving(A)) //It's still a fire, idiot.
 			var/mob/living/L = A
 			L.adjust_fire_stacks(fire_stack_strength)
 			L.IgniteMob()
+		else if(G.GetComponent(/datum/component/grillable))
+			if(SEND_SIGNAL(G, COMSIG_ITEM_GRILLED, src) & COMPONENT_HANDLED_GRILLING)
+				continue
+			G.fire_act(1000) //Hot hot hot!
+			if(prob(10))
+				visible_message("<span class='danger'>[G] doesn't seem to be doing too great on the [src]!</span>")
```
Bibby0110:
```suggestion
				visible_message("<span class='danger'>[G] doesn't seem to be doing too great on [src]!</span>")
```

Bibby0110 (changes requested):

code/modules/hydroponics/grown/towercap.dm
ToasterBiome: "doesn't seem to be doing too great on the bonfire" vs "doesn't seem to be doing great on bonfire"?

Bibby0110 (approved):

PR #12591, opened 2021-11-03
Operation: Spagh-o-nut

AlvCyktor:
Adds a donut made out of spaghetti because god has forsaken us all.

:cl:  
rscadd: Spagh-o-nut  
/:cl:


AlvCyktor:
I opened an open PR instead of a WIP pr so sory

PR #12614, opened 2021-11-07
prevents borgs from self destructing while stunned

SomeguyManperson:

image
oh boy time to convert this borg I've stunned I sure hope nothing wacky or uncharacteristic happe-

image

I had an argument with biome like a couple days ago about this and decided to not do anything but then I got redtexted because a borg self destructing got me killed due to the knockdown and have decided against not doing it because whatever argument can be come up with to give borgs which are already powerful in combat a way to cause more damage even if they lose is stupid and wrong

tldr self destructing when already stunned has no real downside since you're going to be dead/converted whether you do it or not and gives you one last chance to fuck over whoever has stunned you. Liberal use of the self destruct feature can prevent:
traitors from emagging you
clock cultists from converting you
anyone else from killing you if they don't have a gun to keep a safe distance with

# Changelog

:cl:  
rscdel: borgs can no longer self destruct while stunned
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/silicon/robot/robot.dm
+++ b/yogstation/code/modules/mob/living/silicon/robot/robot.dm
@@ -74,7 +74,7 @@
 	set name = "Self Destruct"
 
 	if(usr.stat == DEAD || usr.incapacitated())
-		return //won't work if dead
+		return //won't work if dead or incapacitated
 
 	if(alert("WARNING: Are you sure you wish to self-destruct? This action cannot be undone!",,"Yes","No") != "Yes")
 		return
--- a/yogstation/code/modules/mob/living/silicon/robot/robot.dm
+++ b/yogstation/code/modules/mob/living/silicon/robot/robot.dm
@@ -73,15 +73,15 @@
 	set category = "Robot Commands"
 	set name = "Self Destruct"
 
-	if(usr.stat == DEAD)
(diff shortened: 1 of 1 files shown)


Absolucy:
I DED

Jarod1200:
what a based PR

Marmio64:
i hate silicon mains

Mqiib:
Hate borgs, good nerf, very excellent, tyhank you thoes

maxion12345:
skill issue
honestly issue is that the borg can still scream like a bitch for help after destruction, that's what'll get you killed more then the actual knockdown. the self destruct is pansy ass in terms of damage and knockdown, but you can cheese it to get around the stun blocking your comms. A working radio has killed more antags then anything else.

yogstation/code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -73,15 +73,15 @@
 	set category = "Robot Commands"
 	set name = "Self Destruct"
 
-	if(usr.stat == DEAD)
+	if(usr.stat == DEAD || usr.incapacitated())
 		return //won't work if dead
```
adamsong:
```suggestion
		return //won't work if dead or incapacitated
```

adamsong (dismissed):
Otherwise looks fine codewise

PR #12641, opened 2021-11-10
makes synthetic alien plasma vessel researchable without deconstructing alien organs

boodaliboo:
this tech does not give you anything else so doing this wont really do much of anything 
:cl:  
tweak: tweaked synthetic alien plasma vessel tech

/:cl:

### DIFF
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -85,7 +85,7 @@
 	boost_item_paths = list(/obj/item/weed_extract, /obj/item/xenos_claw, /obj/item/stack/sheet/xenochitin, /obj/item/organ/alien, /obj/item/organ/brain/alien)
 	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 5000)
 	export_price = 20000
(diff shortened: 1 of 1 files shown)


SomeguyManperson:
originally this was meant to be a little research path that would give access to a couple xeno themed organs but all the ideas I had for organs were stupid
I might throw some stuff in later but this allows for wide access to plasma vessels which give constant access to passive healing which is stronger than any other implant

boodaliboo:
> originally this was meant to be a little research path that would give access to a couple xeno themed organs but all the ideas I had for organs were stupid I might throw some stuff in later but this allows for wide access to plasma vessels which give constant access to passive healing which is stronger than any other implant

how would i nerf the healing specifically for this vessel?

nmajask:
I think the plasma vessel is a nice reward for sci after the station stops an infestation or dealing with meateors. The tech should probably be expanded a bit, with maybe other organs based on those that xenos have (i.e. a resin spinner that can put down atmos resin or a neurotoxin gland that can shoot chems).

JamieD1:
No plasma Vessels.

PR #12642, opened 2021-11-10
Revert "prevents borgs from self destructing while stunned"

Absolucy:
Reverts yogstation13/Yogstation#12614

obvious "i ded" selfmerge


Jarod1200:
Bad pr. This change has been a long time coming

Mqiib:
Yeah but what if we just... I dunno, kept it this way.
It's very nice, being able to actually do stuff with borgs without having to kill them or risk blowing up.

LazennG:
Even if it was a selfmerge, this is something borgs have been getting away with for too long and made fighting with them fruitless and unfun

PR #12645, opened 2021-11-11
gives synths healing from oil, welding fuel and teslium

boodaliboo:
brings synths inline with the other synthetic races like android and preternis
:cl:  
rscadd: Added preternis special healing thing with oil welding fuel and teslium to synths
/:cl:


adamsong (approved):
Does as described, will wait on HC for balance

JamieD1 (approved):
Feels stupid, but whatever, will wait for it to be abused.

PR #12648, opened 2021-11-11
makes luxury shuttle windows unbreakable

boodaliboo:
makes luxury shuttle windows unbreakable so the poor can no longer interfere with the rich achievig point of shuttle
:cl:  
tweak: tweaked  fuck the poor
/:cl:

### DIFF
--- a/_maps/shuttles/emergency_luxury.dmm
+++ b/_maps/shuttles/emergency_luxury.dmm
@@ -221,18 +221,6 @@
 /obj/structure/shuttle/engine/propulsion,
 /turf/open/floor/plating/airless,
 /area/shuttle/escape/luxury)
-"oo" = (
-/obj/machinery/door/poddoor/shutters{
-	id = "ohnopoors"
-	},
-/obj/structure/grille/indestructable,
-/obj/structure/window/plastitanium{
-	desc = "A durable looking window made of an alloy of of plasma and titanium. Appears to have poor people tears smudged all over it.";
-	flags_1 = 2177;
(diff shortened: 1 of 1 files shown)


ToasterBiome:
rules were meant to be broken

GoreDem:
inb4 you can rcd these ones as well

Xoxeyos:
Which windows? The ones where all the poor people go, or the ones on the walls of the shuttle?

JamieD1 (approved):

_maps/shuttles/emergency_luxury.dmm
```diff
@@ -221,18 +221,6 @@
 /obj/structure/shuttle/engine/propulsion,
 /turf/open/floor/plating/airless,
 /area/shuttle/escape/luxury)
-"oo" = (
-/obj/machinery/door/poddoor/shutters{
-	id = "ohnopoors"
-	},
-/obj/structure/grille/indestructable,
-/obj/structure/window/plastitanium{
-	desc = "A durable looking window made of an alloy of of plasma and titanium. Appears to have poor people tears smudged all over it.";
```
adamsong, line 230: Where did this description go?

adamsong (changes requested):

patpol4:
You replaced them with a closed transparent turf which means it will have space under it and look shit, I'm all for oppression of the poor but this just makes the window look bad. Would be better to use flags that make it indestructible and un rcd'able, if such exist.

PR #12677, opened 2021-11-16
Prevents temporary spares from being used in the ID Card Modification program

nmajask:

Prevents you from using temp spares in the card program, as it allows you to grant yourself a lot of access and seams to be an oversight. Is tested, does work

# Wiki Documentation

None needed. 

# Changelog

:cl:  
tweak: you can no longer use a temporary spare to give yourself access with the ID Card Modification program
/:cl:

### DIFF
--- a/code/modules/modular_computers/file_system/programs/card.dm
+++ b/code/modules/modular_computers/file_system/programs/card.dm
@@ -63,6 +63,9 @@
 	if(!id_card)
 		return
 
+	if(istype(id_card, /obj/item/card/id/captains_spare/temporary))
+		to_chat(user, span_warning("ERROR: [id_card] is not compatable with this program"))
+		return
 	region_access = list()
 	if(!target_dept && (ACCESS_CHANGE_IDS in id_card.access))
 		minor = FALSE


adamsong:
Under the access of the old ID console the temp spare wouldn't have worked, what is the access for the new program?

Temp spare does not have ACCESS_HEADS or ACCESS_CHANGE_IDS

nmajask:
> Under the access of the old ID console the temp spare wouldn't have worked, what is the access for the new program?
> 
> 
> 
> Temp spare does not have ACCESS_HEADS or ACCESS_CHANGE_IDS

With the ID mod program it isn't one access that is required for everything, you can give access to a department if you have access to the office of the head of that department. Because the temp spare has access to all of the heads offices, you can give yourself access to everything in every department except command.

adamsong:
Interesting, the old console used ACCESS_HEADS for that, and then any perms on the ID could be given out

nmajask:
> Interesting, the old console used ACCESS_HEADS for that, and then any perms on the ID could be given out

I can add a check for ACCESS_HEADS if you want

maxion12345:
isn't the point of the temporary spare to give an acting captain AA/allow them to grab real spare tho

nmajask:
> isn't the point of the temporary spare to give an acting captain AA/allow them to grab real spare tho

Yes, but it was not given ACCESS_HEADS or ACCESS_CHANGE_IDS so that heads couldn't just print out a temp spare and give themselves AA without consulting the cap/HoP or to make the actual spare still important.

Bibby0110 (approved):

PR #12697, opened 2021-11-18
Update dental_implant.dm

A-Distinguished-Axolotl:

Allows the "Dental Implant" surgery to be performed on oneself.

Changes have been tested.

It makes logical sense for this operation to be something you can perform on yourself as well as seems convenient and fun; ergo a decent thing to merge!

# Wiki Documentation

Add a comment on the wiki that the Dental Implant surgery can be performed on itself (possibly change any reference to prosthetic manipulation being the only surgery that can be done to self)

:cl:  
tweak: Allows Dental Implant to be performed on yourself
/:cl:

### DIFF
--- a/code/modules/surgery/dental_implant.dm
+++ b/code/modules/surgery/dental_implant.dm
@@ -2,6 +2,8 @@
 	name = "Dental implant"
 	steps = list(/datum/surgery_step/drill, /datum/surgery_step/insert_pill)
 	possible_locs = list(BODY_ZONE_PRECISE_MOUTH)
+	lying_required = FALSE
+	self_operable = TRUE
 
 /datum/surgery_step/insert_pill
 	name = "insert pill"


ynot01:
me on the inside i want this because no one will ever do this

on the outside i know this will just be abused 24/7 by chem mains and powergamers and theos

Georgeur:
> me on the inside i want this because no one will ever do this
> 
> on the outside i know this will just be abused 24/7 by chem mains and powergamers and theos

only applies to pills which can instantly be swallowed anyway, its not very abusable outside of space combat
its a qol improvement

ToasterBiome:
i aint see no dentist work on their own teeth

SomeguyManperson:
> on the outside i know this will just be abused 24/7 by chem mains and powergamers and theos

Oh you fucking bet

cuackles:
Make it hurt a shit ton because no one in their right mind would drill a hole in their teeths

Hopekz:
> It makes logical sense for this operation to be something you can perform on yourself

What? In what world have you seen people perform dental surgery on themselves?

Dental surgery on yourself isn't even actually possible because you need to see what you're doing and you can't accomplish that by feel alone on yourself

Georgeur:
We're one of the only servers not to have this, and that is not an overstatement. 
Making it hurt does nothing mechanically so it could be added no problem. 
I don't know where you all were when we allowed preternis to take out their own eyes and then replace their liver with a new one, but ya, pulling out lrp shouldnt be valid in this case.

Hopekz:
> I don't know where you all were when we allowed preternis to take out their own eyes and then replace their liver with a new one, but ya, pulling out lrp shouldnt be valid in this case.

I don't think that example applies here because Preterni are machines and their organs are like "modules" that you plug in and out.
The only surgery steps in a Preterni are unscrewing a hatch to get to the module and opening the hatch.
Non-passive organs that need to "turn on" like eyes have an extra step for calibrating the electronics with a multitool.

Georgeur:
> > I don't know where you all were when we allowed preternis to take out their own eyes and then replace their liver with a new one, but ya, pulling out lrp shouldnt be valid in this case.
> 
> I don't think that example applies here because Preterni are machines and their organs are like "modules" that you plug in and out. The only surgery steps in a Preterni are unscrewing a hatch to get to the module and opening the hatch. Non-passive organs that need to "turn on" like eyes have an extra step for calibrating the electronics with a multitool.

canonically theyre human cyborgs, and in the town hall last year moja stated that they feel pain (though for rp rule reasons). 
so *shrugs

SomeguyManperson:
> We're one of the only servers not to have this, and that is not an overstatement.

if you count beestation as the only other server then yes otherwise no because bee is the only server I've found that allows you to self surgery specifically dental implants

> its not very abusable outside of space combat its a qol improvement

usually you have to waste valuable time in combat switching to an empty hand, taking out a pill, then clicking yourself specifically while your sprite may be obscured and is otherwise smaller than an action button
using a dental implanted pill instead you just have to click a nice large action button with no need to disengage for more than a second at most

This is another case of surgery being used to make a strong ability completely unknown simply because you'd have to trust a second person to do it for you and it's easier to just waste the extra 3 seconds in combat than be in a surgery room with someone who might suddenly decide to kill you

adamsong (approved):
Works, but also I don't think its a good idea

JamieD1:
@SomeguyManperson Up to you.

SomeguyManperson:
really don't think this is a good idea

PR #12698, opened 2021-11-18
Adds mutadone pills to cloning room for better access to them

JamieD1:
Adds a pill bottle of mutadone into the cloning pod room and some monkey cubes.

:cl:  
rscadd: Adds pill bottle of mutadone and cubes to cloning room. 
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -9493,21 +9493,6 @@
 	},
 /turf/open/floor/engine,
 /area/science/xenobiology)
-"aBi" = (
-/obj/structure/table/glass,
-/obj/item/book/manual/wiki/medical_cloning{
-	pixel_y = 6
-	},
-/obj/item/pen,
-/obj/effect/turf_decal/trimline/purple/filled/line{
-	dir = 8
-	},
-/obj/machinery/light_switch{
-	pixel_x = -23;
-	pixel_y = 10
-	},
-/turf/open/floor/plasteel/white,
-/area/medical/genetics/cloning)
 "aBj" = (
 /obj/structure/rack,
 /obj/machinery/light{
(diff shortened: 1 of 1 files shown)


Cartlord:
Will this make people less willing to break into Genetics for monkey cubes despite monkey cubes being one of the less efficient ways to get synthmeat early on?

JamieD1:
> Will this make people less willing to break into Genetics for monkey cubes despite monkey cubes being one of the less efficient ways to get synthmeat early on?

Yes. This is one of the reasons for the PR as well.

Aquizit:
> Will this make people less willing to break into Genetics for monkey cubes despite monkey cubes being one of the less efficient ways to get synthmeat early on?

There are *four* breakers of cryox next to the tubes at start of round. Just grab a syringe and inject it with your own blood and boom instant synthmeat.
Do it right next to cloning for most convenient results.

Edit: I'm dumb you said less efficient derp.

patpol4 (changes requested):
Remove the ones in genetics so we dont double the roundstart materials suddenly

patpol4:
Also i dont like this because it removes human interaction and makes things more hugboxy by handing you items on a platter

JamieD1:
> Remove the ones in genetics so we dont double the roundstart materials suddenly

Considering I already nerfed the pill amount before, I dont think this is an issue

Hopekz:
> Also i dont like this because it removes human interaction and makes things more hugboxy by handing you items on a platter

This.
What's the point of having requirements for something if you are just handed the items you need every single time?
Defeats the purpose of having them in the first place since it just goes back to free cloning

Cartlord:
> 
> 
> > Also i dont like this because it removes human interaction and makes things more hugboxy by handing you items on a platter
> 
> This. What's the point of having requirements for something if you are just handed the items you need every single time? Defeats the purpose of having them in the first place since it just goes back to free cloning

I agree TBH, I think this is a shortsighted change that'll reduce greytiding but also reduce skill severely

patpol4 (approved):
Map is fine in a technical sense, i still stand by my previous statements.

patpol4:
Not only does this contribute to material availabilty powercreep and thus raise hugboxiness, but also goes against the philosophy of item scarcity that drives player interaction forwards and gives cargo a job

JamieD1:
> Not only does this contribute to material availabilty powercreep and thus raise hugboxiness, but also goes against the philosophy of item scarcity that drives player interaction forwards and gives cargo a job

Also prevents lots of admin/player issues. I think thats worth it compared to all the arguments here which are people complaining about pills people can make quite easily in chemistry

patpol4:
Stale and bad

PR #12703, opened 2021-11-19
Gives Ash Walkers Night Vision

redmoogle:

# By Spooder just helping him

Gives ash walkers night vision
presumably because lavaland can get dark?

# Wiki Documentation

Ashwalkers can see in the dark

# Changelog

:cl:  
tweak: Gives ashwalkers night vision
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
@@ -49,6 +49,7 @@
 	species_traits = list(MUTCOLORS,EYECOLOR,LIPS,DIGITIGRADE,HAS_FLESH,HAS_BONE)
 	inherent_traits = list(TRAIT_NOGUNS) //yogs start - ashwalkers have special lungs and actually breathe
 	mutantlungs = /obj/item/organ/lungs/ashwalker
+	mutanteyes = /obj/item/organ/eyes/night_vision/ash_walker
 	breathid = "n2" // yogs end
 	species_language_holder = /datum/language_holder/lizard/ash
 
--- a/code/modules/surgery/organs/eyes.dm
+++ b/code/modules/surgery/organs/eyes.dm
@@ -119,6 +119,11 @@
 	name = "fung-eye"
 	desc = "While on the outside they look inert and dead, the eyes of mushroom people are actually very advanced."
 
+/obj/item/organ/eyes/night_vision/ash_walker
+	name = "lizard eyes"
+	desc = "Centuries of evolutions gave those scale wearing specimens night vision to aid their unending hunt"
+	sight_flags = TRAIT_NIGHT_VISION
+
 ///Robotic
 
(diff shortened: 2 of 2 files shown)


SomeguyManperson:
Don't really like this since playing around visibility is the main portion to lavaland combat when it's not extremely onesided. Both sides get extremely high burst damage weapons and someone having a visibility advantage is basically a gg since it makes it way easier to both evade attacks and hit your own

Marmio64:
miners would still have a large advantage even with this added, but i dont see why ashwalkers need nightvision, especially not right off the bat.

Smudged:
> miners would still have a large advantage even with this added, but i dont see why ashwalkers need nightvision, especially not right off the bat.

why wouldn't a species that lives on a dark ash planet not have night vision right away?
I think this is nessasary

SomeguyManperson:
here's a picture of an ashwalker (on the right) and a roundstart miner (on the left) in pitch darkness, the usual amount of light available on lavaland
image
or, with darkness disabled,
image


now usually miners don't run around with their lights off, but they do get meson vision, which looks like this
image
even though it shows the tile, it won't show the lizard unless they are actually visibly lit
If the miner were to turn on their light here, they could see the lizard
image
but now the lizard can see them too, since their light is on. And at this range, the spears thrown by the lizard can hit the miner, but the miner can't shoot the lizard with their KA, as it only has a range of 3 tiles. Conversely, the lizard could be the one with their light on, producing similar results since both parties are lit at this range
image
HOWEVER at a further range, where light does not reach from one party to the other, the side that has their light disabled is at an advantage, as they can see the person whos light is on, but are not revealed themselves
image
Considering ashwalkers have a much longer range weapon than miners, this already puts them at an advantage in dark areas as long as the floor is free of lizard tripping items (considering they don't have much else to worry about they'll probably end up clearing any hazards before combat anyway)

so being lit means you can be easily attacked, and being in darkness means you can't. Considering neither side can see in the dark, they have to use their own light sources to attempt to light up the enemy which also causes them to be exposed. But what if someone (in this case, the ashwalker) has night vision?
image
In this scenario, the miner is at a disadvantage regardless of if their light is on, since the ashwalker can still see them if it's off. Especially against early game miners, who don't have good melee armor or an upgraded KA, and are only _debatably_ a role specifically to kill other players this can quickly turn into a mining team wipe unless the ashwalkers manage to screw it up.

tldr ashwalkers already have a distinct advantage if they are inside the mining base, they don't need to be extremely difficult to fight outside of it as well especially considering it's a ghost role made specifically to kill miners. And "meson nvgs exist just get those" isn't an excuse because that's a lategame technology requiring science to research a very specific path that ends in a 5k node (then requires plastic and materials to actually make), ashwalkers just getting night vision is an early game thing the miners won't even know is a problem until someone ends up with 5 spears lodged in them, and combatting it requires specifically using your own night vision and high range weapon.

PR #12723, opened 2021-11-22
reduce armor polish cost from 6 to 2

SomeguyManperson:

giving a random piece of clothes extremely basic armor is not on par with other 6 tc items like implants or the emag, it's roughly on par with the chamkit which gives a piece of chameleon armor that sucks compared to generic armor but has other stuff like an armored jumpsuit

also I think this is closer to the initial amount when i spitballed the idea for an armor buff to outerwear but I can't find anything to back that up in under 5 minutes

# Wiki Documentation

armor polish cost 6 > 2

# Changelog

:cl:  
tweak: armor polish is now 2 tc down from 6
/:cl:


KillerOrcaCora:
Very unused due to the price. Good change.

ToasterBiome:
as the guy who made armor polish i approve

PR #12732, opened 2021-11-23
Fix text when silencing VOX

Absolucy:


JamieD1 (approved):
This is why we cant have nice things.

PR #12726, opened 2021-11-23
Disables Omegastation

patpol4:
# Document the changes in your pull request

Long overdue, should have been done when we added kilo, 
there are too many maps that we have to maintain right now and this will help reduce that amount.
Data speaks for itself, in the past 6 months omega has been played 9 (nine) times.
image

# Wiki Documentation

Disables Omegastation, doesnt remove it yet, just makes it not pickable

# Changelog

:cl:  
tweak: Disables Omegastation
/:cl:

PR #12746, opened 2021-11-24
Disables players from initiating a map-change vote

LoliconSlayer:
As far as I can tell, map-change votes are exclusively used when the map is literally anything other than box, which is way over played as you can see by the data presented in #12726 (Which this PR also conflicts with)

This isn't a "meme PR" I genuinely think that the obsession with box is bad and initiating a map-change vote whenever the map isn't box is bad.

I'm aware that this PR will probably not do anything, but it's better than whining in Discord or the Forums about it.

:cl:  
tweak: Players can no longer initiate map votes.
/:cl:

### DIFF
--- a/config/config.txt
+++ b/config/config.txt
@@ -154,7 +154,7 @@ ID_CONSOLE_JOBSLOT_DELAY 30
 #ALLOW_VOTE_MODE
 
 ## allow players to initiate a map-change vote
-ALLOW_VOTE_MAP
+#ALLOW_VOTE_MAP
 
 ## min delay (deciseconds) between voting sessions (default 10 minutes)
 VOTE_DELAY 6000


patpol4:
this is a tweak not a deletion

patpol4 (approved):
You have my blessing

LoliconSlayer:
> this is a tweak not a deletion

My apologies, I've fixed it

yacabo:
based

TheLinuxOS:
I support this, although I do find it hilarious that people seem to like this change despite our entire playerbase freaking out for a month when a bug caused it not to work

JamieD1:
The whole point of a map vote is so people can choose. I'm not forcing players to play on a map which Many many people don't like or don't wish to play on.

LoliconSlayer:
Player choice 🤢

PR #12751, opened 2021-11-25
Fixes the vents on some maps on fastdmm2 appearing as missing textures

nmajask:

Boring change that only matters to people editing maps. Will merge conflict with just about any map changes and only hits the stations, not ruins and ect. Also untested but should work.

image

# Wiki Documentation

Maybe some images will need to be updated, but I don't know.


patpol4 (approved):
Looks good to me

PR #12757, opened 2021-11-26
makes conflicting mutations actually conflict

SomeguyManperson:

lmao

# Changelog

:cl:  
bugfix: you can no longer have space adapt and heat adapt at the same time making you completely immune to both temperature changes AND temperature effects as well as environmental damage 
/:cl:


Cartlord:
Dude WTF, what was even the problem with having both of these? It gave you _eighty instability_, that alone was enough to stop it from being practical a lot of the time

PR #12758, opened 2021-11-27
makes modify transform properly scale sprites

SomeguyManperson:

turns out this has only been scaling the x axis the WHOLE TIME when it should be scaling BOTH x AND y axis
either that or I am about to unleash some horrible evil that has been kept locked away in the code for over a thousand millenea

# Changelog

:cl:  
bugfix: automatic scaling no longer forgets half of its job
/:cl:


SomeguyManperson:
turns out this is actually terribly coded and I need to put in real amounts of effort to fix it

PR #12764, opened 2021-11-27
buffs the reviver implant

SomeguyManperson:

Reviver implant has gotten several numbers buffs for being kind of trash for an implant that chainstuns you for a minute if you get EMPed

heal delay lowered from 3 seconds to 2 seconds
reviver implant now works in softcrit so it can actually take you out of crit if you are dying
heart attack timer reduced from 1 minute to 10 seconds

all numbers have been changed to define numbers for easier reading

# Changelog

:cl:  
tweak: reviver implant is now faster, works in soft crit if you aren't sleeping, and has had its heart attack emp effect reduced from a minute to 10 seconds
/:cl:

### DIFF
--- a/code/modules/surgery/organs/augments_chest.dm
+++ b/code/modules/surgery/organs/augments_chest.dm
@@ -56,7 +56,7 @@
 
 /obj/item/organ/cyberimp/chest/reviver/on_life()
 	if(reviving)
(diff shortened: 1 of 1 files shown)


adamsong (approved):
Does as described, will leave merge to HC

JamieD1 (approved):

PR #11057, opened 2021-02-25
removes heart attacks from being too fat since it's not funny

SomeguyManperson:
# General Documentation

### Intent of your Pull Request
hearts will no longer fail from having the fat trait since it's mega uncontrollable, they still take damage from it though
people actually die to this like a lot

### Why is this change good for the game?
reduces RP
# Wiki Documentation

### Briefly describe your PR and the impacts of it, in layman's terms. 
probably not documented but hearts will no longer fail solely due to being fat

### What general grouping does this PR fall under? 
medical

# Changelog

:cl:  
tweak: hearts will no longer spontaneously fail due to bad eating habits
/:cl:

PR #12765, opened 2021-11-27
removes specifically the heart attack effect from being fat

SomeguyManperson:

technically this is funny but it actually kills people way more often than it should just because it's a MASSIVE pain to deal with obesity without the help of a chemist or bartender or actual doctor who knows what they are doing (rare), coupled with how easy it is to accidentally overeat (especially if you don't know it's lethal) and get stuck with several minutes of fatness a feature that randomly kills you for going overweight is kind of stupid

I'm keeping the heart damage though that's thematic and still a good downside since you lose a large amount of time post death where you can be defibbed without a bypass or replacement heart

# Changelog
:cl:  
rscdel: being fat no longer causes heart attacks on its own, it will still cause massive damage to the heart over time
/:cl:

### DIFF
--- a/code/modules/surgery/organs/heart.dm
+++ b/code/modules/surgery/organs/heart.dm
@@ -84,11 +84,6 @@
 			beat = BEAT_NONE
 	if(HAS_TRAIT(owner, TRAIT_FAT) && owner.stat != DEAD) //yogs: being fat causes heart damage
 		owner.adjustOrganLoss(ORGAN_SLOT_HEART, maxHealth * STANDARD_ORGAN_DECAY, 85) //eat happy, eat healthy
-		if(damage >= 80 && beating)
-			if(prob(1))
-				if(owner.stat == CONSCIOUS)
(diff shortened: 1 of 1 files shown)


ChesterTheCheesy:
bro just zap yourself on an airlock if you're getting heart attacked

maxion12345:
Wasn't it decided back in march when you tried this (https://github.com/yogstation13/Yogstation/pull/11057) that instead of removing the mechanic, it would be better to improve upon it by perhaps making the gym holodeck more functional? The gym machines already prevent heart attacks if you exercise. Making it so you can also lose fat by exercising instead of just preventing the effects of being fat would work. We keep the penalty for ignoring your potential heart failure, and the players have a way to remove being fat without asking for surgery, chemicals, or that one obscure fat removal machine nobody builds

tl;dr last time this was posted it was collectively agreed that outright removal was bad, and a better option might be increasing functionality of the weights in the holodeck

JamieD1:
Would prefer to add more ways to reduce fat instead of removing the downsides from it.

PR #12766, opened 2021-11-27
Removes invalid symbols with their correct symbol

redmoogle:

These show up as invalid symbols and are supposed to be 0x91(') if I'm not mistaken
also helps prevent certain tools from breaking


adamsong (approved):
Characters are strange

PR #12772, opened 2021-11-28
Moves a couple of folders from the yogstation module

redmoogle:

~~I should stop~~

Moves a couple of folders
and accidentally realize that I have to move some code around


redmoogle:
fixing this

PR #12775, opened 2021-11-28
mindslaved people reject mindshields

ToasterBiome:
# Document the changes in your pull request

you can mindslave somebody important and then get them re-mindshielded and nobody can do anything about it

# Changelog

:cl:  
tweak: mindslaved people reject mindshields
/:cl:

PR #12776, opened 2021-11-28
mindshields remove mindslave status???? alternative to 12775????

ToasterBiome:

alternative to https://github.com/yogstation13/Yogstation/pull/12775
you can mindslave somebody important and then get them re-mindshielded and nobody can do anything about it

# Changelog

:cl:  
tweak: mindshields remove mindslave
/:cl:

### DIFF
--- a/code/game/objects/items/implants/implant_mindshield.dm
+++ b/code/game/objects/items/implants/implant_mindshield.dm
@@ -71,6 +71,8 @@
 			target.mind.remove_antag_datum(/datum/antagonist/gang)
 		if(target.mind.has_antag_datum(/datum/antagonist/veil))
 			target.mind.remove_antag_datum(/datum/antagonist/veil)
+		if(target.mind.has_antag_datum(/datum/antagonist/mindslave))
+			target.mind.remove_antag_datum(/datum/antagonist/mindslave)
 		if(!silent)
 			if(target.mind in SSticker.mode.cult)
 				to_chat(target, span_warning("You feel something interfering with your mental conditioning, but you resist it!"))


KittyNoodle:
make them cost less, also implant removal surgery still exists as a way to remove mindslave and this seems like a way too easy counter for a 7 tc item

Marmio64:
when someone gets mindslaved it should tell them that they should be careful not to get mindshielded so they dont accidently get deconverted without knowing what would happen.

ToasterBiome:
This PR is, in reality, a "worse future" comparison to https://github.com/yogstation13/Yogstation/pull/12775.
> make them cost less, also implant removal surgery still exists as a way to remove mindslave and this seems like a way too easy counter for a 7 tc item

You would need some pretty hard evidence (akin to the mindslaved person saying "I am mindslaved") to prove that they are implanted as you're generally not allowed to check for implants unless there's obvious use of them and a mindslave + mindshield negates any "obvious" uses (HOS might just be a shitter, or incompetent). There's many ways to undetectably play while being mindslaved.

> when someone gets mindslaved it should tell them that they should be careful not to get mindshielded so they dont accidently get deconverted without knowing what would happen.

Will add if this PR sticks.

adamsong (approved):
Does as the PR says, will leave to HC to decide which to merge

maxion12345:
> You would need some pretty hard evidence (akin to the mindslaved person saying "I am mindslaved") to prove that they are implanted as you're generally not allowed to check for implants unless there's obvious use of them and a mindslave + mindshield negates any "obvious" uses (HOS might just be a shitter, or incompetent). There's many ways to undetectably play while being mindslaved.
>

don't you need evidence to mind shield as well? Otherwise it'd be meta to go implant people arrested for crimes.

Doesn't change much, just makes it so sec won't need their brig doc or medbay to handle deconversion. Also will probably encourage security implanting everyone they arrest for a traitor objective just in case, because now they can detect mindslaves in addition to revs+gang

JamieD1:
Can you revert 12775 in 12776 please.

PR #12777, opened 2021-11-28
makes sleeper stasis more intuitive

SomeguyManperson:

At the time I wanted to use stat alone but as it turns out there's no stat for being specifically in hardcrit and UNCONSCIOUS also shows up a myriad of other ways

Instead the sleeper will trigger stasis if either the occupant is dead or below 0 health

# Changelog

:cl:  
tweak: sleepers with the stasis upgrade will now put people into stasis under more specific circumstances
/:cl:

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -181,7 +181,7 @@
 	check_nap_violations()
 	var/mob/living/carbon/C = occupant
 	if(C)
-		if(stasis && C.stat == DEAD || C.health < 0)
+		if(stasis && (C.stat == DEAD || C.health < 0))
 			C.apply_status_effect(STATUS_EFFECT_STASIS, null, TRUE)
 		else
 			C.remove_status_effect(STATUS_EFFECT_STASIS)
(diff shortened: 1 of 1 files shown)


code/game/machinery/Sleeper.dm
```diff
@@ -183,7 +181,7 @@
 	check_nap_violations()
 	var/mob/living/carbon/C = occupant
 	if(C)
-		if(stasis && C.stat >= stasis_health)
+		if(stasis && C.stat == DEAD || C.health < 0)
```
adamsong:
```suggestion
		if(stasis && (C.stat == DEAD || C.health < 0))
```

`FALSE && FALSE || TRUE`

Returns true

adamsong (changes requested):

Notamaniac:
There needs to be some more feedback that the stasis is actually engaged, or better yet make it a button so if doctors need chems to metabolise while someone is in crit, they have the option to do so. Unless TGUI is literally that horrible.

JamieD1 (approved):

PR #12793, opened 2021-12-01
"Toxins" that don't do toxin damage won't cause liver damage

Absolucy:
`toxpwr = 0` should not cause liver damage

:cl:  Lucy
tweak: Reagents that are considered "toxins" but don't deal any toxin damage will no longer damage your liver.
/:cl:


Bibby0110:
not a bugfix
change to tweak please

Absolucy:
@TheGamerdk ah, okay. I had it down as a bugfix because it didn't seem like that was intended behavior.

PR #12800, opened 2021-12-04
Gives most swords the ability to parry incoming attacks

ChesterTheCheesy:
Attempts to remove RNG from swords by making it more controlled by player.

**Most swords can now activate a parrying stance by clicking on them.**
When you do this, a "tink" sound will play and your eye will glow:

nottin_personnel,_klown

During a short parry window, you'll be immobilized and unable to attack, but you'll reflect reflectable projectiles like lasers, and any melee attack done to you will be blocked, with additional effect based on intent:
**Help: Nothing, just blocks the attack
Harm: Attacks the attacker
Disarm: Attempt to throw away attacker's weapon
Grab: Short stun on the attacker (attacker doesn't drop the weapon)**

Parrying, naturally, has a cooldown.

The affected swords are:
All claymores, chaplain nullrod "claymore" variants, katana and captain's sabre.
Claymores have the shortest parry window frame, katana has the longest.
Chaplain claymores can only perform the blocking parry.
Eswords, ekatanas, and other swords that already have an action assigned to interacting with them (such as transforming the sword) are unaffected, but i might change that in the future, by making the action an alt+click instead.
Cult swords are already strong as they are, and toy swords aren't real swords, so they both don't get this feature.

# Changelog

:cl:  
rscadd: Added the ability for most swords to parry.
soundadd: Added a riposting sound for throwing out opponent's weapon away
imageadd: Added eye glint effect
experimental: Need to balance the values if needed
/:cl:

(sorry for no actual footage for sword parrying, i couldn't make any on a private server)


nesisknesis (approved):

ChesterTheCheesy:
I will clean up code and unify the parriable swords later if it turns out to be a good feature.

Sniblet:
give it to toy swords coward

maxion12345:
How long is the window? Is there a limit to how many attacks are blocked, or a limit between how fast? Take, for instance, a buckshot shell. Surely a swordsman couldn't block them all at once?

ChesterTheCheesy:
> How long is the window? Is there a limit to how many attacks are blocked, or a limit between how fast? Take, for instance, a buckshot shell. Surely a swordsman couldn't block them all at once?

So, technically the parry window is between 1 second and 1.8 seconds, depending on sword, but it's all vague balance i couldn't really test so it'll most likely get changed. In that time, every melee attack is blocked, so it works against being overwhelmed.
Next up, i lied about the "reflect any projectile", it only reflects projectiles that can be reflected like lasers, so shotgun would still work.

Cartlord:
i just want to point out that, assuming the ss13 katanas are made in the same way as 21st-century katanas, they'd be awful at parrying - katanas are extremely fragile

ChesterTheCheesy:
Will try to turn it into a component and re-open it then.

Cartlord:
oh shit ok, i just figured maybe they wouldn't be able to parry
and even that isn't necessary, ss13 isn't exactly realistic

PR #12802, opened 2021-12-04
You can no longer start a map vote right after map rotation.

Absolucy:
:cl: Lucy
tweak: You can no longer start a map vote right after the map is automatically rotated.
/:cl:

This is mainly because every time the map rotation automatically goes to a non-Box map, some dingus will always make a map vote, and it'll go back to Box. It's not really a majority voting for box either - even with 50-60+ players, only like ~20 total will vote at all (and even if it was an actual majority, [tyranny of the majority](https://en.wikipedia.org/wiki/Tyranny_of_the_majority) is something that should be avoided). Sometimes you should *gasp* learn new things.

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -372,6 +372,7 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 	. = changemap(VM)
 	if (. && VM.map_name != config.map_name)
 		to_chat(world, span_boldannounce("Map rotation has chosen [VM.map_name] for next round!"))
+		CONFIG_SET(flag/allow_vote_map, FALSE)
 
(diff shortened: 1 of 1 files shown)


N3D6:
Based

LoliconSlayer:
Inb4 Jamie closes this PR with the same reason he closed #12746

adamsong:
This is a tweak, not a deletion

adamsong (approved):

Bibby0110:
Put a PR body with reasoning

Absolucy:
well this is quite frankly useless now because someone got mad and deleted all the other maps

JamieD1:
The whole point of a map vote is so people can choose. I'm not forcing players to play on a map which Many many people don't like or don't wish to play on. Also lots of people will just not play next round tanking player count. I would rather a more populated round then an empty ship on a different map for "variety"

PR #12813, opened 2021-12-05
removes all maps except box from rotation because the playerbase is always right

SomeguyManperson:

Now you will never have to play shitty maps like anything except box ever again
I will take my unanimous 50 upvotes now please

# Changelog

:cl:  
rscdel: every map except box is now out of rotation, aka nothing has changed
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -24,17 +24,17 @@ endmap
 
 map yogsmeta
 	minplayers 25
-	votable
+	disabled
 endmap
 
 map yogsdelta
 	minplayers 50
-	votable
+	disabled
 endmap
 
 map kilostation
 	maxplayers 40
-	votable
+	disabled
 endmap
 
 map icebox


patpol4 (approved):
This works too

Hopekz:
Unironically a great PR because it expedites the direction we already seem to be going as a codebase.

If we want to go back to maintaining multiple maps in the future I think meta is the most played non box map out of the bunch

Marmio64:
coders are finally accepting that the players just want to play on their map of choice rather than be oppressed time and time again with vote weight changes and literal disabling of the ability to try to change the map to their preferred one.

maxion12345:
good change
we need more boxstation maps now for different pops
also kilo is the best lowpop map no cap can we keep that in rotation :(

SomeguyManperson:
variety is bad

patpol4:
I'm going to merge this

Mqiib:
This is terrible

KillerOrcaCora:
finally

Junkyshell:
unsurprising

Mqiib:
Imagine unironically thinking that removing a bunch of perfectly serviceable maps with only a few quirks here and there is a good idea. Variety? Nah we want to play on the same map over and over and over forever and ever because apparently people hate fun.

Hopekz:
> Imagine unironically thinking that removing a bunch of perfectly serviceable maps with only a few quirks here and there is a good idea. Variety? Nah we want to play on the same map over and over and over forever and ever because apparently people hate fun.

No one said the maps were bad. 
The problem is that all these maps have to be maintained equally despite them not being played.

This means that if someone makes a big PR to change something (for example Bibby's AI rework) he has to do 4-5x the map work to make sure all the other maps are updated. This doesn't make sense from a development standpoint so we should remove the maps that aren't played and the data clearly shows that box is played on 98%+ of the time.

Whenever I made a PR with map changes I just ignored the other maps because I couldn't be asked to do 4x the work to update maps I knew weren't going to be played (for example all heads of staff starting in the same room). 
Funny enough close to a year later no one can be bothered to even apply this change to other maps because no one cares about developing the maps people don't play on.

Absolucy:
I have asked _several times_ for people to make a list of issues with meta, fully intending to go and fix them.

Nobody ever bothers to actually say what's wrong with the other maps, how is anyone supposed to know how to fix it??? 🤡🤡🤡

PR #12822, opened 2021-12-05
Protolathes and such will attempt to charge newly printed power cells

Absolucy:
This makes anything that can print a power cell attempt to charge it while printing. Mainly, this is a convenience thing. Also, I want science to actually print bluespace power cells for the RPED and upgrade stuff that uses power cells.

:cl: Lucy
add: Protolathes, autolathes, techfabs, and mechfabs, when printing a power cell, will attempt to charge it.
add: APCs will be used to charge the cell if it has enough power *and* is over 20% capacity.
add: Otherwise, it will attempt to draw power from available power on the powernet directly.
/:cl:


Marmio64:
absurdly large space ninja buff

Absolucy:
@Marmio64 heh, this is more meant as "now science has NO excuse to not upgrade chemistry's power cells" / "now robotics has NO excuse to not upgrade borg's cells"

Cartlord:
What happened?

JamieD1:
So you are charging it whilst its being made? Will it come out charged everytime? Seems a bit op.

Absolucy:
@JamieD1 it charges right after being made. it's not free power though, it will only work if there's enough power in the APC or powernet.

Absolucy:
if you still feel that's a bit OP, here's a few potential compromises

 - quite a bit of extra power is used, which means it's less efficient than a cell charger
 - only like 25% or so of the battery is charged at maximum

maxion12345:
cell charger mains coping and seething
anyways as long as it isn't popping out fully charged 40k cells easily then it shouldn't be too much of an issue

Marmio64:
maybe have it only charge high capacity+ and lower to full(aka 15,000 charge aka 15 KJ)
that way you have to spend some time charging for high tier cells while medium quality ones are instant

adamsong:
Pulling from the power net feels OP, also the all or nothing on the cell feels strange

PR #12823, opened 2021-12-05
gets rid of the awful mentor 10% "noot noot" sound

Absolucy:
it hurts my ears

:cl: Lucy
sounddel: Mentorhelps no longer have a 10% chance to give you ear cancer
/:cl:


Absolucy:
actually hm i'll just add a preference that disables _all_ annoying "alternative" sounds

PR #12824, opened 2021-12-05
Adds a preference for disabling "alternative" sounds

Absolucy:
:cl: Lucy
add: There is now a preference to disable "alternative" potentially misophonia-triggering sounds!
/:cl:

Currently affects the mentor noot, PDA glitch beep, and RPED "wooooosh"


code/game/objects/items/devices/PDA/PDA.dm
```diff
@@ -799,7 +799,7 @@ GLOBAL_LIST_EMPTY(PDAs)
 	tnote += signal
 
 	if (!silent)
-		if(HAS_TRAIT(SSstation, STATION_TRAIT_PDA_GLITCHED))
+		if(HAS_TRAIT(SSstation, STATION_TRAIT_PDA_GLITCHED) && can_play_alt_sound())
```
adamsong: Why is the alt PDA sound based solely on the prefs of the person holding the PDA?

Reed0506:
how could you do this to the beeps and the wooooosh

yogstation/code/datums/world_topic.dm
```diff
@@ -80,7 +80,7 @@ GLOBAL_VAR_INIT(mentornoot, FALSE)
 	var/client/C = GLOB.directory[ckey(whom)]
 	if(!C)
 		return 0
-	if(GLOB.mentornoot || prob(1))
+	if((C.prefs.toggles & SOUND_ALT) && (GLOB.mentornoot || prob(1)))
```
adamsong, line 83:
```suggestion
	if(GLOB.mentornoot || (C.prefs.toggles & SOUND_ALT) && prob(1))
```

adamsong (approved):

adamsong (dismissed):
That was supposed to be request

Hopekz:
Just realized we have a sound file name of "pshoom_or_beepboopblorpzingshadashwoosh" since it's next to your changes

code/game/objects/items/devices/PDA/PDA.dm
Absolucy: hm, good idea, I could make an override in playsound

yogstation/code/datums/world_topic.dm
Absolucy: no.

code/game/objects/items/devices/PDA/PDA.dm
Absolucy: i made it override using a list in playsound instead

N3D6:
I really wish we'd stop with the "disabling this because I do not like it" prs

maxion12345:
isn't miso a type of soup

Absolucy:
@N3D6 it's pure QoL, meant to reduce annoyance, while not affecting gameplay in the slightest. i do not see the problem.

adamsong:
It disables one of the was admins can torment mentors

N3D6:
> It disables one of the was admins can torment mentors

This is why I am against this :kek:

Marmio64:
let mentors give a stupid ass sound to ahelps so that we can see how they like being tortured

adamsong:
If we gave mentors equality, they wouldn't be oppressed now would they?

Marmio64:
> If we gave mentors equality, they wouldn't be oppressed now would they?

sounds good to me

Hopekz:
So how does this work if the sound isn't sent directly to the player? Like for example the woosh sound since it is broadcasted.

For example for the woosh RPED: Does the player that has this preference off just never create that sound for others or does someone without the preference off still create the sound for you to hear despite you have it set to not hear them?

**Rewriting question to maybe better understand it if it was confusing:**
You: Have alt sounds disabled
Them: Have alt sounds enabled
Them: use the part device
The woosh sound gets created and broadcast on to the world

Do you still hear "woosh" since they created the noise?
Does this apply in reverse of you never being able to create the woosh sound yourself and no-one else hearing the woosh despite them wanting to have alt sounds enabled?

Absolucy:
@Hopekz all sounds (unless sent via the `SEND_SOUND` macro) eventually boil down to a `playsound_local`, which is per-client, where we can then swap out sounds for individual clients.

JamieD1:
Need to be careful here we dont encourage a mentality of catering to everyones likes and dislikes. People cant always get what they want. And I need you to confirm the Default is ON.

Absolucy:
@JamieD1 

Default should be on: https://github.com/yogstation13/Yogstation/blob/40655d325382ea490f46c4d5f1482ab5cb01a46b/code/__DEFINES/preferences.dm#L30


Savefile upgrade should also turn it on automatically: https://github.com/yogstation13/Yogstation/blob/40655d325382ea490f46c4d5f1482ab5cb01a46b/code/modules/client/preferences_savefile.dm#L50-L51

yogstation/code/modules/mentor/mentor.dm
```diff
@@ -104,7 +104,7 @@ GLOBAL_PROTECT(mentor_href_token)
 /proc/send_mentor_sound(client/C)
 	var/sound/pingsound = sound('yogstation/sound/misc/bikehorn_alert.ogg')
 	pingsound.volume = 90
-	if(GLOB.mentornoot || prob(10))
+	if((C.prefs.toggles & SOUND_ALT) && (GLOB.mentornoot || prob(1)))
```
Bibby0110:
```suggestion
	if((C.prefs.toggles & SOUND_ALT) && (GLOB.mentornoot || prob(10)))
```

Absolucy: why? it was 1 before

adamsong: Except it wasn't, check the changes on the PR

Absolucy: oh whoops, the other one was `prob(1)`, you're right

JamieD1 (approved):

PR #12846, opened 2021-12-08
Excludes the hacked AI law upload module from infiltrators

Absolucy:
This is meant to discourage infiltrators from screwing with the AI. It *never* ends well, and honestly goes against the whole stealth thing.

:cl: Lucy
tweak: Infiltrators can no longer buy the hacked AI law upload board. Stop screwing with the AI ffs!
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1470,6 +1470,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	item = /obj/item/aiModule/syndicate
 	cost = 4
 	manufacturer = /datum/corporation/traitor/cybersun
+	exclude_modes = list(/datum/game_mode/infiltration)
 
 /datum/uplink_item/device_tools/hypnotic_flash
 	name = "Hypnotic Flash"


SapphoQueer:
Absolutely amazing change when half the infiltrator shifts (without AI objective) end up with them going for AI anyways. Hopefully this discourages them somewhat.

maxion12345:
Ehhh, excluding an item because of one bad experience is not setting a good precedent here. Fact is, AI subversion is a massively powerful tool for any antagonist when done correctly, especially for an antagonist that needs to steal and murder shit. We should not exclude infiltrators based on the fact that they haven't done it well a few times. Also these guys are basically traitors in a team so disallowing common traitor strategies just for them is bad

also who the fuck buys that just use a freeform lmao

Absolucy:
@maxion12345 Not one bad experience. This is not the first, nor the second, or whatever round of infiltrators deciding to mess with the AI 4noraisin and promptly screwing everything up as a result.

maxion12345:
>  Not one bad experience. This is not the first, nor the second, or whatever round of infiltrators deciding to mess with the AI 4noraisin and promptly screwing everything up as a result.

ngl gonna say that's a big skill issue and not the item's fault
anyways why would traitors get this item and not infiltrators

JamieD1 (approved):

PR #12850, opened 2021-12-08
Adds the Access Kit for infiltrators

Absolucy:
This is to make infiltrators able to infiltrate better, and give them a reason to not just be another assistant, which is easily metagamed. This way, they can infiltrate as a "simple" job, complete with the access that comes with the job. Should allow for more creative infiltrations, too.

:cl: Lucy
add: Adds the Access Kit for infiltrators, which allows them to spoof the ID and access of several "low-level" jobs. It is single-use and only one can be bought per infiltrator.
/:cl:

Needs a better sprite, rn it's the CRAB-whatever sprite.
Also de-modularized the infiltrator uplink stuff.
Current jobs that the access kit allows: station engineer, medical doctor, janitor, botanist, cargo tech, scientist


bruhlookatthisdood:
Finnaly I wont die to that external airlock that refuses to open!

adamsong (approved):
Always a fan of yeeting the yogs folder. Code looks good, I like the idea.

yogstation/code/modules/antagonists/infiltrator/items/access_kit.dm
```diff
@@ -0,0 +1,51 @@
...
+		return
+	if (!job)
+		to_chat(user, span_warning("[src] has not been set to a specific job yet! Use it in-hand to set up the access kit."))
+		return
+	var/obj/item/card/id/id = target
+	id.assignment = job.title
+	id.originalassignment = job.title
+	id.access |= job.minimal_access
+	id.update_label()
+	to_chat(user, span_notice("You apply [src] to [id], granting it the access of that job!"))
```
Hopekz:
```suggestion
	to_chat(user, span_notice("You apply [src] to [id], granting it the access of a [job.title]!"))
```
Might as well

Absolucy:
i doubt anyone cares too much about this

JamieD1 (approved):

PR #11320, opened 2021-04-10
Lexorin now (approximately) 50% less deadly

Mqiib:
# General Documentation

### Intent of your Pull Request

Currently lexorin only needs to do ~55 damage to knock you out, which is half of what other toxins need to do. This adjusts the damage it does to 3/tick instead of 5 as well as reduce the breath loss stacks by half so it doesnt last for-fucking-ever after it's done metabolizing and kill you anyways.
 
### Why is this change good for the game?

Lexorin is INCREDIBLY deadly as it is. It's still deadly here, but not "you have 30 seconds before you're dead" deadly.

# Wiki Documentation

### Briefly describe your PR and the impacts of it, in layman's terms. 

Lexorin does less oxyloss damage (5->3) and only add 1 breath loss stack

### What should players be aware of when it comes to the changes your PR is implementing?

Lexorin isn't an absolute death sentence without immediate medical aid now.

### What general grouping does this PR fall under? 

Chemistry balance

### If there are any numerical values involved in your PR that will be relevant to a player, please note them here. 

Lexorin does less oxyloss damage and half as many breath loss stacks.

# Changelog

:cl:  
 
tweak: Lexorin oxyloss from 5 to 3 and breathloss stacks from 2 to 1 per cycle

/:cl:

PR #12853, opened 2021-12-09
Epinephrine purges lexorin

Sniblet:
As it stands, a single 15u lexorin syringe fired from a syringe gun gives the victim about 30 seconds to get help or else they're unconscious and well on their way to dead
Getting help can be difficult, as in most cases the lexorin syringe was fired by a living person who wants you to die and is also in the room with you
Lexorin, as a reminder, can be made in like 40u quantities within a minute from roundstart using a simple 1:1:1 = 3 recipe of base and easily acquired reagents

This would make it the same except if you think fast and use your epipen you can live, this time

Note that I have no idea how fast this actually purges lex, I think it's the same rate as charcoal, so pretty slow? But I do know that lexorin keeps working for a while even after completely leaving the system, so "too fast" or "fast enough" are both harder to achieve than they might look
Ideally this wouldn't make lexorin a bad poison, just make it so it isn't _miles ahead of everything else a chemist can make_ and effective as a deathmix entirely on its own. If the person is late in deploying epi they should suffer for it; if they're quick they should at least stay conscious.

Not tested but uses copypasta code soooo

# Wiki Documentation

Prevents oxygen damage from crit state. If the patient is in crit it heals 0.5 oxyloss, toxins, brute, and burn. 20% chance each tick to reduce stun times a bit. Gives a minor stamina regeneration buff. **Slowly purges lexorin.**
Worse than Atropine at saving critical patients, but better for mid combat use.
Overdose-Effect: 33% chance to deal 2.5 stamina and 1 toxin damage, and some suffocation.

# Changelog

:cl:  
tweak: Epinephrine purges lexorin
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -838,6 +838,8 @@
 	. = 1
 	if(prob(20))
 		M.AdjustAllImmobility(-20, FALSE)
+	if(holder.has_reagent(/datum/reagent/toxin/lexorin))
+		holder.remove_reagent(/datum/reagent/toxin/lexorin, 1)
 	..()
 
 /datum/reagent/medicine/epinephrine/overdose_process(mob/living/M)


Sniblet:
Double ided to be exact, it's happened twice
I don't understand all the confused emotes: is there something positive about chemists being able to very rapidly and consistently KO _and_ kill anyone with one shot of a very simple recipe that they can make in their own lab with almost no time investment and minimal effort otherwise? To say nothing of hacked medborgs?

The only tied (if the wiki is accurate, lexorin is _actually a little faster_, but if I'm reading the code right they're about tied) alternatives in speed for putting people to sleep are sodium thiopental and chloral hydrate, and the latter is uh, slower to make than this. If miners are sapient, it's easier to make in great quantity too. And chloral metabolizes 150% faster. And doesn't cause slowdown. And there's a chance of you waking up in the first place with chloral.
The fastest dedicated sedative we have is a "budget" sedative, compared to something that's _cheaper_ than it.

And what other single non-traitor-exclusive chems can 100-to-0 someone 100% of the time with one 15u syringe?
Well, there's coniine. Sulfonal or anacea, only after they kill the liver. Cyanide, amanitin, or nicotine, eventually. Crank, krokodil, meth, or bath salts, once the addictions get going. Do we count histamine or slime jelly? Those aren't 100%. Amatoxin almost, but doesn't quite, do it.
All of these are slower (sometimes much slower), harder to get into hands that also hold a syringe gun (often impossible to make in the lab), or both, than lexorin - and none of them KO the victim as fast, if at all. Coniine comes close, but again: chemists can't make it.
It's not a deathmix. It's one compound.

[Do you know, in Star Trek, lexorin was a medicine?](https://memory-alpha.fandom.com/wiki/Lexorin)

See also:
https://github.com/tgstation/tgstation/pull/51541
https://tgstation13.org/wiki/Guide_to_chemistry#Lexorin

KittyNoodle:
Reminder that the game is not supposed to be balanced around one sec officer being stronger than one antag.
Making a powerful toxin that is actually producible with chemistry easily counterable with something everyone has(epipen) is not a good idea

Sniblet:
> Reminder that the game is not supposed to be balanced around one sec officer being stronger than one antag.

Okay. Do you mean to say that one sec officer is supposed to be weaker than one revolutionary, one thrall, one blood brother, or one griefer? How about the captain? The HoS? Literally anyone?
But lexorin isn't an antag item, is it?
Is one chemist supposed to be stronger than one traitor? Gang head? Infiltrator? Cultist? Clockie? Heretic? Vampire? _Wizard?_
Anyone and everyone can be point-blanked by a lexorin gun that was hidden in anyone's backpack, and then what? There's no counterplay besides regular toxin-purging chems or salbutamol (rip), which both require you to go to medbay, which isn't necessarily within 30 seconds' jog even assuming your poisoner - who works in the medbay - somehow just demanifests after they attack. Call for help? You mean hold still and type a bit in front of this guy?
It's the same difficulty you have with someone igniting an energy sword right next to you with killing intent. One moment you're not certain that it's a threat, the next you're almost down. Eswords are traitor-exclusive and cost 8 TC. Lexorin isn't an antag item.
Other dangerous non-antag items include stunprods, which do not guarantee death on their own, do not work from range, do not knock you out, are a little harder to make, and can't be stored and hidden on someone without lategame tech.

> Making a powerful toxin that is actually producible with chemistry easily counterable with something everyone has(epipen) is not a good idea

Let me rephrase that: "Making a powerful toxin that is easily producible with chemistry easily counterable is not a good idea"
Think about that for a few seconds.
Okay, so am I wrong? Are hydrogen, oxygen, and plasma hard to get?
You can memorize the recipe within a minute and make it in less without leaving your lab or drawing any attention - in fact, you don't even need the lab, just grind a glowstick and a plasma sheet. No thought, no creativity, no complexity. Cyanide is harder to make, slower to act, more visually distinct in liquid form, and has a special visible message when it reacts. Lexorin is the most powerful toxin chemistry has bar none and it's a trekchem whose medicinal purpose was lost in translation oh and it's barely more complex than mannitol and again it's the most powerful toxin chemistry has bar none.
It makes every other toxin redundant. Pyrotechnics can be beaten with fire extinguishers or showers. Narcotics give you time to get help. All the other one-chem-one-shot syringes take more time both to make and to work. Why bother when you can just
>le lexorin

and win?
If a chemist invests 6 TC in a poison kit, it's still a close call where I'd say lexorin is still the best. Why do you want it that way? _Why should it be easy to get, easy to deploy, but almost impossible to counter?_ It's the same question we asked about tasers.
Even Goonchem is better balanced: [sarin](https://wiki.ss13.co/Chemicals#Sarin) is slowed by epi, flushed by atropine (which is more common than here), and takes 6 times as long to permanently knock you down. And it's an exclusive secret recipe and syndicate weapon.

Sniblet:
I don't understand people being against this. Is it that lexorin isn't too powerful, or that this is the wrong way to nerf it? I have no objection to the latter. Look up for my opinion on the former.

Sniblet:
https://github.com/yogstation13/Yogstation/pull/11320
I'm pretty sure this was merged before the two times I've been ideded by a lexorin gun. It's still an enormously powerful oneshot sedative with a chance to crit that lasts forever, even if not a oneshot deathchem, and I am dismissive of every opinion to the contrary. Bad for the game shut up shut up shut up.
Having talked to Gaminada, THE lexorin mentor:tm:, it sounds like even this isn't nerf enough? It sounds like he wants lexorin itself to both be harder to make and to metabolize faster all the time.
I definitely agree with the former. I don't want to make lexorin useless like an Average Coder would, but I don't want it to give chemists _this_ much more than they put into it. /TG/'s recipe uses salbutamol.
Every other toxin either needs a buddy, or at least a lot of effort getting it, to allow chemists to deploy a 15u powergame slug of instant kills +4. I like things that way. Lexorin ruins that. All that I don't know is the right way to make lexorin better.

Maybe I go with Gaminada's idea. It can be like cyanide's twice-as-bright-half-as-long cousin.

PR #12868, opened 2021-12-10
prevents custom docking points from being placed on plating

SomeguyManperson:

turns out plating is now a viable ship docking turf which means instead of needing to detonate a sizeable portion of the station to land a shuttle inside it you just would need to deconstruct all the walls and remove any floor tiles

# Changelog

:cl:  
tweak: custom shuttle landing points can now be placed on specifically asteroid/lavaland turfs not station maintenance
/:cl:

### DIFF
--- a/code/modules/shuttle/navigation_computer.dm
+++ b/code/modules/shuttle/navigation_computer.dm
@@ -15,7 +15,7 @@
 	var/view_range = 0
 	var/x_offset = 0
 	var/y_offset = 0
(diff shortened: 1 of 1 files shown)


nmajask:
> turns out plating is now a viable ship docking turf which means instead of needing to detonate a sizeable portion of the station to land a shuttle inside it you just would need to deconstruct all the walls and remove any floor tiles

I don't think this is true, if I recall correctly it also checks for area. If not, that would be much better way of going about it.

adamsong (approved):
Probably a good idea

PR #12869, opened 2021-12-10
Renames Yogstation to YogsBox

nmajask:

Renames Yogstation to YogsBox in, hopefully, the map vote and internally, as that is what everyone refers to it as and naming it after the server probably isn't the best idea. Will probably cause merge conflicts, but it shouldn't be hard to fix. Untested, should work.

# Wiki Documentation

If Box is referred to as Yogstation, it should be changed to YogsBox. 

# Changelog

:cl:  
tweak: Yogstation has been renamed to YogsBox
/:cl:


nmajask:
Oh, I forgot to not commit the tail whitespace changes

patpol4 (approved):
This is most likely fine

alexkar598:
Cornflakes

nmajask:
> Cornflakes

Sorted

adamsong (changes requested):
Revert your changes to tails.dm

Hopekz:
Literally no point in doing this and people playing here for years know this map as yogstation

UselessTheremin:
> Literally no point in doing this and people playing here for years know this map as yogstation

what about new players

Hopekz:
> what about new players

What do you think this PR does for new players?

All this PR is doing is changing the name of a map, something that will only effect people that know what this map is called.

patpol4:
Makes the word boxstation less synonymous with the word yogstation, good change just for that.

Hopekz:
> Makes the word boxstation less synonymous with the word yogstation, good change just for that.


Oh this is one of those yogdev moments.
Didn't realize this was an anti-box PR until you said that and when you take away the anti-box energy this PR is just a seemingly pointless name change. 

Not even sure what the vision is here anymore but if renaming a map gets you closer to it I guess it should be done?

Are any other maps being renamed to the yogs[name] syntax or is this just going to apply to box because of the recent drama? Like YogsKilo

JamieD1:
Missed a trick calling it YogBox. But also this feels extremely pointless. People know its box already. I dont know to be honest.

Cartlord:
Actually, when I got here it was days - might've been weeks - before I realized it was just your name for BoxStation. Good change.

adamsong:
In common usage the map is referred to as box, so why isn't the official name that?

And how is this an anti-box pr? It just clears up some ambiguity surrounding the term "yogstation"

PR #12877, opened 2021-12-12
More interesting singularity movement

Absolucy:
The singularity's movement has been changed. It can now pick a "random target" based on an interest value. It goes through 3x3 sections in a 15 tile radius from the singularity, assigning an interest value based on the amount of non-space tiles and objects that are in the tile. It will then pick where to move based on a weighted pick - it can still pick an uninteresting target to move to, but it's much more likely to go towards somewhere with 50 bajillion objects on a tile.

The singularity can still move randomly, though. It will also continue to prioritize active singularity beacons above this new system.

:cl: Lucy
tweak: The singularity is less likely to fuck off into space now.
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1579,3 +1579,11 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	return call(source, proctype)(arglist(arguments))
 
 #define TURF_FROM_COORDS_LIST(List) (locate(List[1], List[2], List[3]))
+
+/proc/get_box(corner_x, corner_y, z, size_x, size_y)
+	. = list()
+	for	(var/x in corner_x to corner_x + size_x)
+		for (var/y in corner_y to corner_y + size_y)
+			var/turf/T = locate(x, y, z)
+			if (T && istype(T))
+				. |= T
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1579,11 +1579,3 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	return call(source, proctype)(arglist(arguments))
 
 #define TURF_FROM_COORDS_LIST(List) (locate(List[1], List[2], List[3]))
-
-/proc/get_box(corner_x, corner_y, z, size_x, size_y)
-	. = list()
-	for	(var/x in corner_x to corner_x + size_x)
-		for (var/y in corner_y to corner_y + size_y)
-			var/turf/T = locate(x, y, z)
-			if (T && istype(T))
-				. |= T
--- a/code/modules/power/singularity/singularity.dm
+++ b/code/modules/power/singularity/singularity.dm
@@ -1,4 +1,5 @@
-
+#define SINGULARITY_QUADRANT_SIZE 3
+#define SINGULARITY_QUADRANT_DISTANCE 15
 
 /obj/singularity
 	name = "gravitational singularity"
@@ -23,6 +24,7 @@
 	var/grav_pull = 4 //How many tiles out do we pull?
 	var/consume_range = 0 //How many tiles out do we eat
 	var/event_chance = 10 //Prob for event each tick
+	var/turf/random_target = null // a randomly chosen target.
 	var/target = null //its target. moves towards the target if it has one
 	var/last_failed_movement = 0//Will not move in the same dir if it couldnt before, will help with the getting stuck on fields thing
 	var/last_warning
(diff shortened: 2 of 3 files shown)


code/modules/power/singularity/singularity.dm
```diff
@@ -350,6 +365,31 @@
...
+			for (var/turf/T in box)
+				if (!isspaceturf(T))
+					interest += 2
+				var/list/objs = list()
+				for (var/A in T.contents)
+					if (istype(A, /atom/movable))
+						objs += A
+				interest += CEILING(length(objs) / 7.5, 0.5)
+			sections[section_loc] = interest
+	to_chat(world, "singularity sees [LAZYLEN(sections)] sections.")
```
Altoids1: Probably wanna get rid of these debug prints

code/modules/power/singularity/singularity.dm
```diff
@@ -350,6 +365,31 @@
...
+			var/list/box = get_box(x + section_x, y + section_y, z, SINGULARITY_QUADRANT_SIZE, SINGULARITY_QUADRANT_SIZE)
+			var/interest = 0
+			for (var/turf/T in box)
+				if (!isspaceturf(T))
+					interest += 2
+				var/list/objs = list()
+				for (var/A in T.contents)
+					if (istype(A, /atom/movable))
+						objs += A
+				interest += CEILING(length(objs) / 7.5, 0.5)
```
Altoids1: Since you don't actually do anything with the objects, perhaps it'd be better to just count them rather than put them in a list and *then* count them.

I doubt it's a huge performance difference but it'd be a nice change.

Altoids1:
I'm not really too sure how to feel about this.

I will say that this will make it easier to use a singloose as a more consistent chaos generator as an antagonist, since it reduces the chance of it just fucking off southward. In general, it's nice to make things more deterministic if players make strategies around them.

adamsong:
Will this affect other kinds of singularity such as Nar-Sie and Ratvar?

Absolucy:
@adamsong ratvar isn't a singularity, but I just fixed it so nar'sie won't be affected

neeshacark:
i mean tesla already moves to spots with lots of devices, this PR just makes singulo more appealing to tots.

code/__HELPERS/unsorted.dm
```diff
@@ -1579,3 +1579,11 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	return call(source, proctype)(arglist(arguments))
 
 #define TURF_FROM_COORDS_LIST(List) (locate(List[1], List[2], List[3]))
+
+/proc/get_box(corner_x, corner_y, z, size_x, size_y)
+	. = list()
+	for	(var/x in corner_x to corner_x + size_x)
```
Bibby0110: just curious if i'm missing something obvious, but why not just use block()?

Absolucy: ... i didn't know it existed lol

code/modules/power/singularity/singularity.dm
```diff
@@ -350,6 +366,28 @@
 					return TRUE
 	. = !.
 
+/obj/singularity/proc/pick_random_target()
+	var/list/sections = list()
+	for (var/section_x = -SINGULARITY_QUADRANT_DISTANCE - SINGULARITY_QUADRANT_SIZE; section_x <= SINGULARITY_QUADRANT_DISTANCE - SINGULARITY_QUADRANT_SIZE; section_x += SINGULARITY_QUADRANT_SIZE)
```
Bibby0110: isn't this assymetrical?
x = -18 to x = 12

code/modules/power/singularity/singularity.dm
```diff
@@ -350,6 +366,28 @@
...
+		for (var/section_y = -SINGULARITY_QUADRANT_DISTANCE - SINGULARITY_QUADRANT_SIZE; section_y <= SINGULARITY_QUADRANT_DISTANCE - SINGULARITY_QUADRANT_SIZE; section_y += SINGULARITY_QUADRANT_SIZE)
+			var/turf/section_loc = locate(x + section_x, y + section_y, z)
+			if (!section_loc || !istype(section_loc))
+				continue
+			var/list/box = get_box(x + section_x, y + section_y, z, SINGULARITY_QUADRANT_SIZE, SINGULARITY_QUADRANT_SIZE)
+			var/interest = 0
+			for (var/turf/T in box)
+				if (!isspaceturf(T))
+					interest += 2
```
Bibby0110: these could probably be defines

JamieD1 (approved):

PR #12880, opened 2021-12-12
Makes wasps have 50 health not 200

UselessTheremin:
200 is immensely steep especially with it being included in the infection gamemode and it being so fast, begone
:cl:  
tweak: wasps now have 50 health instead of 200
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/wasp.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/wasp.dm
@@ -18,8 +18,8 @@
 	aggro_vision_range = 9
 	see_in_dark = 7
 	speed = 2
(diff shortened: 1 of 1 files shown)


ChesterTheCheesy:
Bro hands off, it's a jungleland mob, if problem is with infection then make infection wasps have less hp.

UselessTheremin:
> Bro hands off, it's a jungleland mob, if problem is with infection then make infection wasps have less hp.

i can change that when jungleland comes into the equation if need be. its also a xenobio mob tbh

ChesterTheCheesy:
Yes, it's meant to be on-par with watchers, thus matches their HP.

SomeguyManperson:
handling this internally for infection

PR #12659, opened 2021-11-14
Adds some stuff to charlie station

redmoogle:
# Document the changes in your pull request

Adds a thing of cable coils by medical maint
stuffs the PA in a crate
Removes the transport pod for a bridge
Makes the science room *slighty* bigger
Makes the security room slightly smaller
Adds a small randomized maint loot
gets rid of unbreakable rocks
fixes two vents being off
removes random unconnected wire

image

## Why?

Well; The issue is that the pod gets in the way alot and so I changed it to a foot bridge which caused me to need to realign some stuff which is why theres now two maint rooms

I also included three fixes because why not

# Wiki Documentation

Just remove the part saying they're connected by transit tubes

# Changelog

:cl:  
rscadd: Removes the pod from charlie and adds a bridge instead
tweak: Stuff the PA in charlie station into a crate
tweak: Adds another pack of cable coils to charlie station
bugfix: Fixes some issues on charlie station
/:cl:

PR #12916, opened 2021-12-17
Fix Charlie Station bridge wiring

Chubbygummibear:
Fixes the HORRENDOUS oversight made by Redmoogle in #12659 by misplacing like 2 cable pieces so the bridge wouldn't become powered when you set up power.
(jk love you red, no one else pays much attention to the ghost roles)

Yes I tested this and it works now
Before:
image

After:
image

# Changelog

:cl:  
bugfix: Charlie Station bridge is now connected to the power network at spawn
/:cl:


Chubbygummibear:
I think i makey a little mistakey because the actual code change is absolutely wack

patpol4 (approved):
My bad

PR #12919, opened 2021-12-17
[AI] [READY] Minor AI Revamp Tweaks

Bibby0110:
Adds a version of the freeminer cyborg firmware board to the oldstation. This will allow oldstation to create 2 borgs without panel access restrctions

Fixes some verbs being inaccessible as the AI.

You can now use lawboards on MMIs and Posibrains to set what laws they should use when they become an AI. Does NOT work for borgs. Examine to see the loaded lawset.

Fixed HUD being disabled when reentering shell.

Increased server heat production more than 2x

Making a new AI control console requires a password. This password is displayed on a plaque in the RDs office. You must be next to the plaque to read the code.
You use it on any new consoles to enable login. You can also use it as a one time password to login to the console after, this forces a renewal of the code and notifies everyone of your location and ID name.
You can print a new code through the comms console. Printing a new code will also cancel all other old codes that exist. It does NOT disable all previously logged in consoles.

Fixes hot air produced by machines not merging with the rest of the room if there weren't vents siphoning the room.

Adds 15 ticks of grace period for expansion busses.

Adds the ability to use the malf upgrade by hitting the AI control console with it and selecting an AI. The console needs to be unlocked to do this.

Made the AI machinery able to be deconstructed.

# Wiki Documentation

# Changelog


Cartlord:
What does this... do, exactly?

Bibby0110:
> What does this... do, exactly?

Backend changes and bug fixes

adamsong:
> This will allow oldstation to create 2 borgs without panel access restrctions

You do realize the module isn't consumed right? It drops like the reboot module

Bibby0110:
> 

I guess I shouldn't steal adam code :/

Edit:
fixed

KittyNoodle:
I really don’t like the password system

Bibby0110:
> I really don’t like the password system

I really don't like traitors stealing the AI from maint

KittyNoodle:
wait, i am a moron, the password is an alternative to logging in regularly

Bibby0110:
> wait, i am a moron, the password is an alternative to logging in regularly

My explanation may have been a bit wacky.

After building a NEW console you NEED the password to be able to use it.

After this you can log in normally OR use the password and notify everyone. This is so traitors without emags can unlock the console.

code/game/objects/structures/signs/signs_plaques.dm
```diff
@@ -26,6 +26,23 @@
...
 	icon_state = "kiddieplaque"
 
+/obj/structure/sign/plaques/ai_password
+	name = "\improper AI default password"
+	desc = "This plaque contains the default password for AI control consoles onboard this station."
+	var/control_code = "BUG"
+
+/obj/structure/sign/plaques/ai_password/Initialize(mapload)
+	. = ..()
+	if(mapload)
```
adamsong: Why only on map load?

adamsong (commented):
Looks decent

code/game/objects/structures/signs/signs_plaques.dm
Bibby0110: Doesn’t do much you’re right, if we did somehow have a way to make plaques then I wouldn’t want the new plaque to show the current password, but we don’t so I’m going to remove this

SomeguyManperson (approved):

PR #12921, opened 2021-12-17
Katmos (and auxtools)

Absolucy:
This replaces extools with auxtools, and monstermos with katmos (which acts pretty much the same as monstermos)

***THIS NEEDS A BUNCH OF TESTING AND REVIEW!!!***

https://user-images.githubusercontent.com/65794972/146617511-79a8a325-1bd6-4ab2-bae5-9909b4783117.mp4

https://user-images.githubusercontent.com/65794972/146617521-1c913ac4-92b4-45da-89f9-0f27a0098a60.mp4

:cl: Lucy
rscadd: ported Katmos. atmospherics should be mostly the same, but a bit less buggy now.
/:cl:


maxion12345:
does this have explosive decomp?

Altoids1:
> does this have explosive decomp?

Yes. (or at least she said that it could be enabled)

Absolucy:
it is enabled (first video in the PR shows it in action)

maxion12345:
Not sure how I feel about this. The gas spread seems pretty slow. One bit I do love about monstermos is that it's all fast as hell. Open a door to room filled with fire and plasma and it'll rush out fast like proper gas should

Dunno maybe it's because there isn't much pressure difference in the videos, can you give one where you open a room filled with fire?

Marmio64:
i think plasmafloods are incredibly OP and i'd like it if gas stayed slow like in the video.

maxion12345:
> i think plasmafloods are incredibly OP and i'd like it if gas stayed slow like in the video.

It is a flood of super deadly gas that is also flammable
So yes, powerful. Shouldn't be anything you can easily escape because that's what happened with old Atmos
you'll just slowly see plasma spreading from a vent, have time you scream malf ai, weld other vents, grab fire gear and then escape the room

If you want to reduce the effect of plasma floods then place more emergency fire equipment around. Hell, there's only a few cabinets publicly accessable by crew.

SomeguyManperson:
this causes firelocks to not stop atmos changes sometimes being too slow and makes air equalizing really janky in its current state

Absolucy:
@SomeguyManperson a hotfix just got released for auxmos a few hours ago which should fix that, should work better after a recompile

neeshacark:
I prefer monstermos' speed, which is contreversial but I've never had any issues surviving a plasmaflood.  It's really not that hot and you can escape if you are fast enough.

DukeBobRoss:
The speed of monstermos is why monstermos is so cool. Like maxion12345 said if it makes plasmafloods to powerful just add more fire equipment about the station.

Also slower gas causes the following issues:

It takes a super long time for gas to come out of the gas miners, meaning it was near-impossible for me to get a substantial tritburn since the gas came out of the miners slower than it would burn. On round 38755 (which was a 50 minute round) I wasn't able to complete fusion which I should've had done by 20 minutes in because I couldn't get enough gas to tritburn long enough to actually produce 20,000k of heat

Getting rid of gas in a superheated room takes a super long time meaning if some idiot opens a plasma canister what would've been space and fixed now becomes 30 minutes of whatever areas were affected being completely uninhabitable.

Also spacing rooms takes ages and refilling spaced rooms also takes ages

Absolucy:
The slowness seems to be a bug, this is how fast it's meant to be:
gasess2

Absolucy:
well this'll obviously never be merged so whatever

PR #12933, opened 2021-12-19
Gives QM access to mining sec checkpoint

adamsong:

The checkpoint is in the QM's department, therefore they should have access.

# Changelog

:cl:  
tweak: Added QM access to the cargo sec checkpoint
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -21606,25 +21606,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/quartermaster/miningdock)
-"bph" = (
-/obj/machinery/door/airlock/security/glass{
-	name = "Security Office";
-	req_access_txt = "63"
-	},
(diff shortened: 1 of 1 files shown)


Reed0506:
why tho

adamsong:
Its in their department, they should have access to it

Reed0506:
why would the QM need to ever go in there for non-antag activites

ynot01:
QM is not a head and cargo is not theirs it is the HoP's, they are basically the warden of cargo

maxion12345:
qm not a hed
hop is head of cargo as cargo is a civilian department and thusly gets access to that sec area

JamieD1:
Yeah I know you are doing this because of the whole brig Phys thing. But its not worth it. Its just not.

Marmio64:
this is a different case since wardens often need medical stuff to treat prisoners when there is no brig phys, since warden's job is to protect and care for prisoners, meanwhile the QM doesn't need anything from the sec checkpoint in order to do his job.

PR #12940, opened 2021-12-20
Should in theory fix the bodycam sprite issue

cuackles:
in theory it should make the body cams use their own icon state and stop them from taking any unnamed sprite in the neck file
untested but it should work

# Wiki Documentation

# Changelog

:cl:   
bugfix: fixed a few things   
/:cl:

### DIFF
--- a/code/modules/clothing/neck/bodycamera.dm
+++ b/code/modules/clothing/neck/bodycamera.dm
@@ -7,7 +7,7 @@
 /obj/item/clothing/neck/bodycam
 	name = "body camera"
 	desc = "A wearable camera, capable of streaming a live feed."
-	icon_state = "sec_bodycam_off"
+	icon_state = "sec_bodycam"
 	item_state = "sec_bodycam"
 	var/prefix = "sec"//used for sprites, miner etc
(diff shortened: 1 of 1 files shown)


ChesterTheCheesy (changes requested):
Icon_state was okay as "_bodycam_off", since item has two sprites for on and off. The problem is that proc/update_icon() changes the item_sprite, where technically it shouldn't, since no sprites exist, thus using default sprite (so not changing at all if there's no set default sprite)

cuackles:
> Icon_state was okay as "_bodycam_off", since item has two sprites for on and off. The problem is that proc/update_icon() changes the item_sprite, where technically it shouldn't, since no sprites exist, thus using default sprite (so not changing at all if there's no set default sprite)

so then how would i go about fixing it ? this got me a bit lost

ChesterTheCheesy:
> > Icon_state was okay as "_bodycam_off", since item has two sprites for on and off. The problem is that proc/update_icon() changes the item_sprite, where technically it shouldn't, since no sprites exist, thus using default sprite (so not changing at all if there's no set default sprite)
> 
> so then how would i go about fixing it ? this got me a bit lost

simply remove the part where item_sprite changes to on/off

PR #12947, opened 2021-12-20
placing apc frames checks the area where the user is standing, not where the apc is being placed.

ToasterBiome:

apc placement is fucky with areas, since individual room areas are padded over hallway areas, it's pretty much impossible to remake hallway APCs (the ones most bombed) if they get destroyed because placing it on a hallway wall will think it's whatever area is on the other side of the wall. This fixes it by checking the area where the user is standing, not the wall at which it's placed (which is where APC's technically are anyway)

# Wiki Documentation

none

# Changelog

:cl:  
bugfix: APC frames now check where the user is standing instead of the wall at which you're placing it, giving more consistent placement in hallways.
/:cl:

### DIFF
--- a/code/game/objects/items/apc_frame.dm
+++ b/code/game/objects/items/apc_frame.dm
@@ -92,6 +92,8 @@
 	if(!..())
 		return
 	var/turf/T = get_turf(user.loc) //get the turf the user is standing on, not where it's being placed.
+	if(!user || !user.loc)
+		T = get_turf(on_wall) //default back to the turf if the user or their loc is null
 	var/area/A = get_area(T)
 	if(A.get_apc())
 		to_chat(user, span_warning("This area already has an APC!"))
--- a/code/game/objects/items/apc_frame.dm
+++ b/code/game/objects/items/apc_frame.dm
@@ -91,10 +91,9 @@
 /obj/item/wallframe/apc/try_build(turf/on_wall, mob/user)
 	if(!..())
 		return
-	var/turf/T = get_turf(user.loc) //get the turf the user is standing on, not where it's being placed.
-	if(!user || !user.loc)
-		T = get_turf(on_wall) //default back to the turf if the user or their loc is null
-	var/area/A = get_area(T)
+	var/area/A = get_area(user) //get the turf the user is standing on, not where it's being placed.
+	if(!A)
+		A = get_area(on_wall) //default back to the turf if the user or their loc is null
 	if(A.get_apc())
 		to_chat(user, span_warning("This area already has an APC!"))
 		return //only one APC per area
--- a/code/game/objects/items/apc_frame.dm
+++ b/code/game/objects/items/apc_frame.dm
@@ -91,6 +91,7 @@
 /obj/item/wallframe/apc/try_build(turf/on_wall, mob/user)
 	if(!..())
 		return
+	var/turf/T = get_turf(on_wall) //we still need T for checks later in this proc
 	var/area/A = get_area(user) //get the turf the user is standing on, not where it's being placed.
 	if(!A)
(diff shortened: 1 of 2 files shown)


ChesterTheCheesy:
So this means i can place an APC on a tile that isn't part of the area? Like, put the APC for custom shuttle on station, so you can connect it to station's solars, for instance?

ToasterBiome:
> So this means i can place an APC on a tile that isn't part of the area? Like, put the APC for custom shuttle on station, so you can connect it to station's solars, for instance?

no

code/game/objects/items/apc_frame.dm
```diff
@@ -88,10 +88,10 @@
 	inverse = 1
 
 
-/obj/item/wallframe/apc/try_build(turf/on_wall, user)
+/obj/item/wallframe/apc/try_build(turf/on_wall, mob/user)
 	if(!..())
 		return
-	var/turf/T = get_turf(on_wall) //the user is not where it needs to be.
+	var/turf/T = get_turf(user.loc) //get the turf the user is standing on, not where it's being placed.
```
adamsong: Default back to the turf if the user or their loc is null

adamsong (changes requested):

code/game/objects/items/apc_frame.dm
ToasterBiome: > Default back to the turf if the user or their loc is null

no

code/game/objects/items/apc_frame.dm
```diff
@@ -88,10 +88,12 @@
...
```
adamsong:
```suggestion
	var/area/A = get_area(user) //get the turf the user is standing on, not where it's being placed.
```

code/game/objects/items/apc_frame.dm
```diff
@@ -88,10 +88,12 @@
...
+	if(!user || !user.loc)
```
adamsong:
```suggestion
	if(!A)
```

code/game/objects/items/apc_frame.dm
```diff
@@ -88,10 +88,12 @@
...
+		T = get_turf(on_wall) //default back to the turf if the user or their loc is null
```
adamsong:
```suggestion
		A = get_area(on_wall) //default back to the turf if the user or their loc is null
```

code/game/objects/items/apc_frame.dm
```diff
@@ -88,10 +88,12 @@
...
 	var/area/A = get_area(T)
```
adamsong:
```suggestion

```

adamsong (dismissed):

KittyNoodle:
telekinesis

adamsong:
Do a check for TK

ToasterBiome:
> Do a check for TK

/obj/item/wallframe/proc/try_build already has a check for distance, cant place things on walls with tk.

adamsong (changes requested):
This allows placing the wall frame, the APC grabs its area [here](https://github.com/yogstation13/Yogstation/blob/master/code/modules/power/apc.dm#L196) which, will just grab the loc of the APC, not the person building it.

JamieD1:
The issue is some idiot moved alot of the apcs for rooms into maint, which means you need to place the apc in the department now with this PR which will make old placements not work, I think we just need to move all the apcs back into departments to be honest.

ToasterBiome:
> The issue is some idiot moved alot of the apcs for rooms into maint, which means you need to place the apc in the department now with this PR which will make old placements not work, I think we just need to move all the apcs back into departments to be honest.

this placing is only when you're rebuilding one, which helps get areas you _actually_ want to place the APC in instead of whatever area the other side of the wall is in.

JamieD1 (approved):

PR #12961, opened 2021-12-22
Fixed false walls not examining correctly

ynot01:

The old way made the additional examine text appear before the actual examine text, not after

This fixes it

# Changelog

:cl:  
bugfix: Fixed false walls not examining correctly
/:cl:

### DIFF
--- a/code/game/objects/structures/false_walls.dm
+++ b/code/game/objects/structures/false_walls.dm
@@ -143,8 +143,11 @@
 /obj/structure/falsewall/get_dumping_location(obj/item/storage/source,mob/user)
 	return null
 
+/obj/structure/falsewall/examine(mob/user)
+	. += ..()
+	. += span_notice("The outer plating is <b>welded</b> firmly in place.")
+
 /obj/structure/falsewall/examine_status(mob/user) //So you can't detect falsewalls by examine.
-	to_chat(user, span_notice("The outer plating is <b>welded</b> firmly in place."))
 	return null
 
 /*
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/false_walls.dm
```diff
@@ -143,8 +143,11 @@
 /obj/structure/falsewall/get_dumping_location(obj/item/storage/source,mob/user)
 	return null
 
+/obj/structure/falsewall/examine(mob/user)
+	. += ..()
```
adamsong: Why is this += and not just =?

code/game/objects/structures/false_walls.dm
```diff
@@ -159,8 +162,11 @@
 	walltype = /turf/closed/wall/r_wall
 	mineral = /obj/item/stack/sheet/plasteel
 
+/obj/structure/falsewall/reinforced/examine(mob/user)
+	. += ..()
+	. += span_notice("The outer <b>grille</b> is fully intact.")
```
adamsong: This will put both the regular and the reinforced wall message

adamsong (dismissed):

code/game/objects/structures/false_walls.dm
ynot01: how do i fix it

code/game/objects/structures/false_walls.dm
ynot01: copied code from deconstruction code e.g. https://github.com/yogstation13/Yogstation/blob/master/code/game/turfs/simulated/walls.dm#L48-L50

ynot01:
found an easier way to do this

JamieD1 (approved):

PR #12980, opened 2021-12-25
Adds a list of stats to items that dictate how they work in combat, gives spears 2 tile reach but makes them more unwieldy

SomeguyManperson:

Gives weapons a list of new attributes, default values change nothing so most stuff wont be changed

Swing speed: multiplier for attack cooldown (cooldown default 0.8 seconds so 2 speed = 1.6 seconds)

Encumbrance/time: slowdown caused on attacking with the weapon to the wielder, time being the duration in .1 seconds

Reach: tiles the weapon can attack away from the user, attacking further tiles increases attack cooldown multiplicatively by range (2 tiles = double cooldown, 3 tiles = triple cooldown, etc)

Damage: amount of damage the weapon takes on attacking, randomized between two values, if the weapon hits 0 integrity it  breaks

currently only effects spears, which get higher encumbrance when wielded, but are able to attack 2 tiles further with glass+rod spears being breakable after heavy use, and shivs/shanks which get a slightly lower attack cooldown but can break

additionally, flattens the chitin spear into a subtype of the bone spear since they share most of their values aside from damage/icon

idk if this is a good idea/even remotely balanced as is
# Wiki Documentation

spears: when wielded gain +1 range, but cause slight slowdown for 0.5 seconds after attacking while wielded. Attacking from 2 tile range will double attack cooldown
carrot shiv/shanks: attack 20% faster than usual, but can break after several attacks
glass crafted spears: can break after ~20-50 attacks, leaving behind the rod and possibly the shard used in construction

# Changelog

:cl:  
tweak: weapons can now have a slowdown applied to their user on hit, and have differing attack cooldowns/take damage on attack
tweak: spears now have 2 range when wielded, but slightly slow the user when attacking while wielded. Spears crafted from glass shards can break from heavy use
tweak: carrot shivs and shanks can attack slightly faster, but can break after repeated use
/:cl:


applepineapplepizza:
i support this idea

Marmio64:
i dont know, the spears in game are rather short and i dont know if it makes enough sense for them to have an extra tile of reach, plus it'd probably allow ashwalkers to cheese fauna too easily. Maybe leave spears alone and see if someone else would like to make a pike of some sort, or even make it yourself if you want to.

patpol4 (approved):
Map fine

KillerOrcaCora:
Cool for things to have different attributes other than stats that make them unique if this works out would love to see more melee weapons get more detailed abilities. Almost makes me want to use something other than a stun baton or stun prod...Almost.

JamieD1 (approved):

PR #12994, opened 2021-12-27
Removes the spellblade from the bubblegum loot table

KittyNoodle:
# Document the changes in your pull request
Removes spellblade from bubblegum's loot chest.
Spellblade is supposed to be a reward for fighting a hard boss, however it doesn't mesh well into the rest of the game, as being able to delimb someone with zero counter.

Bubblegum is also not very difficult but that is a separate issue

# Wiki Documentation

Remove spellblade from bubblegum loot.

# Changelog

Edit the changelog below to reflect the changes made by this PR, even if the changes are minor - required for every PR.
If you add a name after the ':cl', that name will be used in the changelog. Leave it empty to use your GitHub name.

:cl:  
rscdel: Spellblade can no longer be accessed by miners
/:cl:

PR #12995, opened 2021-12-27
slightly buffs wizard spellblade and nerfs mining spellblade [alternative to #12994]

Marmio64:
spellblade's projectile has damage increased to 20 instead of 15. Weak spellblade keeps the normal 15 damage projectiles and has it's dismemberment ability lowered from 50(%?) to 20.

I had made this before #12994 but hadn't published it yet
but nows a good time as ever to make it as an alternative

keeping in the slight spellblade buff because its pretty good but not good enough to contend with something like mjolnir.

# Changelog

:cl:  
tweak: spellblade's projectile has damage increased to 20 instead of 15. Lavaland/bubblegum spellblade keeps the normal 15 damage projectiles and has it's dismemberment ability lowered from 50(%?) to 20.
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -1039,7 +1039,7 @@
 		if(2)
 			new /obj/item/blood_contract(src)
 		if(3)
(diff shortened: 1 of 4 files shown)


ToasterBiome (approved):
Code does as described

PR #13004, opened 2021-12-28
fixes runtime in standarrow.dm

SomeguyManperson:
runtime bad
:cl:  
bugfix: you now die from failing the holoparasite check
/:cl:


Absolucy (approved):
oops

PR #13021, opened 2021-12-30
Traitor AIs now have access to malf modules

Absolucy:
If an AI's working for the syndicate, it's obviously malfunctioning in some way, right?

This give traitor AI access to malf modules, _except_ for Doomsday Device or Robotic Factory. Those two are still full malf-only.

# Changelog

:cl: Lucy
tweak: Turns out, AIs working for the syndicate are, in fact, malfunctioning, and behave like a malfunctioning AI!
/:cl:

### DIFF
--- a/code/modules/antagonists/malf/malf.dm
+++ b/code/modules/antagonists/malf/malf.dm
@@ -1,5 +1,4 @@
 /datum/antagonist/traitor/malf //inheriting traitor antag datum since traitor AIs use it.
-	malf = TRUE
 	roundend_category = "malfunctioning AIs"
 	name = "Malf"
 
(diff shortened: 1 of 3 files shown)


JamieD1:
I mean no they are not? They just have laws which are unable to be changed or overwritten, they are not malfunctioning.

Absolucy:
@JamieD1 is that uh, not the definition of a malfunction

image

JamieD1:
> @JamieD1 is that uh, not the definition of a malfunction
> 
> image

And Syndicate AIs are Subverted. Not malfunctioning. Like saying an AI with Oxygen is dangerous to humans is Malf.

PR #13036, opened 2022-01-03
Delete libbyond-extools.so

alexkar598:
don't put so files in the repo, it doesn't work like that. Linux SO files are not portable across distributions.


alexkar598:
Urg, would need to fix ci and I cant be assed

PR #13058, opened 2022-01-08
Prpbanly fixes runtime in say.dm

SomeguyManperson:
Assuming this is causing problems with other stuff also it makes it smaller
:cl:  
Bugfix: probably fixes say verbs not showing up
/:cl:


adamsong:
This now checks all the verbs, not just one, messing up the probabilities

PR #13077, opened 2022-01-11
new tip of the round that points out that there is indeed a book in the ashwalker den that literally tells how to make all the obscure crap

Marmio64:
read the fucking book

# Changelog

:cl:  
rscadd: Adds a new tip of the round that states that there is a useful book in the ashwalkers den.
/:cl:


adamsong (approved):
Does indeed

PR #13090, opened 2022-01-13
[Concept] AI downloading is done by intercepting it between data cores

Bibby0110:
Removes the AI upload/download console.
Uploading an AI is now done by hitting a data core with the MMI or posibrain.
AI is now no longer automatically hooked up to all other data cores. It can hook itself up to 2 others (For a total of 3 cores)
Hooking up is permanent unless the data core is destroyed.

Downloading is now done by intercepting the AI using a modular console. You do this by running the program while standing in the rectangle or triangle formed by the data cores.  (Rectangle in the case of 2, triangle in the case of 3) 
If the AI is only in 1 core you can download it by hitting that core with an intellicard.

Example of roundstart download area:
image

Goals:
Preventing AI core spam.
Removing the stupid AI control console
Creating a trade-off between physical safety and ease of download.

# Wiki Documentation

# Changelog

:cl:  
experimental: todo
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1579,3 +1579,9 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	return call(source, proctype)(arglist(arguments))
 
 #define TURF_FROM_COORDS_LIST(List) (locate(List[1], List[2], List[3]))
+
+///Convert a location into a list with the format list(x, y)
+/proc/to_coord_list(atom/loc)
+	. = list(loc.x, loc.y)
+	.[1] = loc.x
+	.[2] = loc.y
--- a/code/modules/mob/living/silicon/ai/ai.dm
+++ b/code/modules/mob/living/silicon/ai/ai.dm
@@ -112,6 +112,8 @@
 	var/downloadSpeedModifier = 1
 
 	var/login_warned_temp = FALSE
+	//Did we get the death prompt?
+	var/is_dying = FALSE 
 
 
 /mob/living/silicon/ai/Initialize(mapload, datum/ai_laws/L, mob/target_ai, shunted)
--- a/code/modules/mob/living/silicon/ai/ai_triangulation.dm
+++ b/code/modules/mob/living/silicon/ai/ai_triangulation.dm
@@ -1,0 +1,49 @@
+GLOBAL_VAR_INIT(ai_triangulation_width, 4)
+/atom/proc/can_triangulate()
+
+	var/obj/machinery/ai/data_core/n1 
+	var/obj/machinery/ai/data_core/n2 
+
+	for(var/obj/machinery/ai/data_core/D in GLOB.data_cores)
+		if(!n1)
+			n1 = D
+			continue
+		if(!n2)
+			n2 = D
+			continue
+
+	var/direction = get_dir(n1, n2)
+
+
+	
+	var/atom/n1_p1 = get_step(get_step(n1, turn(direction, 90)), turn(direction, 90))
+	var/atom/n1_p2 = get_step(get_step(n1, turn(direction, -90)), turn(direction, -90))
+	n1_p1.color = "#FF0000"
+	n1_p2.color = "#FF0000"
+
+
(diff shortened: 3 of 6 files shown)


Hopekz:
So now you can steal/kill the AI by just standing in space?

Bibby0110:
> So now you can steal/kill the AI by just standing in space?

Just like how you can currently steal/kill the AI by standing anywhere with power

Absolucy:
How will hijacking work?

maxion12345:
So like, do you need to disable one datacore to force a transfer to another, then intercept the transfer?
Or is it just "stand here and use the program to download AI"
Also should keep the upload/download for crew use, at least in the AI sat. Could just make it so the AI can revoke download/upload permission from that console, making it so a crew would then need to use the program to download it instead (so you don't have both methods at once unless the AI wants to cooperate)

Bibby0110:
> How will hijacking work?

One of two ways.
1. Hook it onto downloading, aka you have to be in the zone
2. You have to hit the core containing the AI with the board

Bibby0110:
> So like, do you need to disable one datacore to force a transfer to another, then intercept the transfer? Or is it just "stand here and use the program to download AI" Also should keep the upload/download for crew use, at least in the AI sat. Could just make it so the AI can revoke download/upload permission from that console, making it so a crew would then need to use the program to download it instead (so you don't have both methods at once unless the AI wants to cooperate)

Latter. 

Console is going away no matter what.

grubstank:
Minus support from me.  I think this leaves the AI just a little too vulnerable to attacks, especially when there is a guaranteed low security zone between the two default datacores (one which is known to all players, too - so there is not even the protection of secrecy, and not even requiring any effort required by the crew/antags to discover the vulnerability)

I also find this PR kind of strange as it's supposed to target (and does) target core spam (which has not manifested into an actual issue in practice, and which seems generally balanced to me at present)  but is focused on download mechanics. Sort of mixing two separate issues which I think are best balanced separately.  

This is much more a change in concept than anything - and with this new concept skewed in one direction we would also need to introduce AI counterplay (eg: having at least some ability to change active cores as needed (perhaps on a cooldown?) in order to counter download attempts. Or at least to force the crew to actually invest effort to discover the download vulnerability)

At least with AI control consoles there is substantial effort in acquiring the board, and because of the nature of those efforts it is usually clear to the AI/crew well in advanced before a download attempt happens.

Bibby0110:
> Minus support from me. I think this leaves the AI just a little too vulnerable to attacks, especially when there is a guaranteed low security zone between the two default datacores (one which is known to all players, too - so there is not even the protection of secrecy, and not even requiring any effort required by the crew/antags to discover the vulnerability)

Nobody is forcing the AI to use the round-start data core in Science. In addition it's a pretty narrow field of vulnerability.
There might however be something to say about reducing download speed in space.

> I also find this PR kind of strange as it's supposed to target (and does) target core spam (which has not manifested into an actual issue in practice, and which seems generally balanced to me at present) but is focused on download mechanics. Sort of mixing two separate issues which I think are best balanced separately.

Preventing AI core spam is just a natural side-effect of this change. Most shifts I've seen have only created 1 more data core anyhow, so it really won't change all that much, other than not allowing data cores spread out in the 3 furthest away corners


> This is much more a change in concept than anything - and with this new concept skewed in one direction we would also need to introduce AI counterplay (eg: having at least some ability to change active cores as needed (perhaps on a cooldown?) in order to counter download attempts. Or at least to force the crew to actually invest effort to discover the download vulnerability)

1. You'd always be aware of where people could download you from.
2. As it stands the AI already get an upgrade to decrease download speed. Adding more types of defensive upgrades would be the obvious next step.
(3. There's always the possibility of an active defense for the AI. Most likely in a form of mini-game or similar to increase the time it takes to finish the download.)
 
Allowing the AI to switch active cores in this system wouldn't work as you'd be able to prevent downloads almost entirely.

> At least with AI control consoles there is substantial effort in acquiring the board, and because of the nature of those efforts it is usually clear to the AI/crew well in advanced before a download attempt happens.

From my experiences, no. Getting a board is stupidly easy due to how circuit printing works. Just like getting upload boards.

PR #13099, opened 2022-01-14
Beam rifles only get one shot

Sniblet:

The laser anti-material gun is used for nothing except BTFOing blobs and they are far too excellent at it, they're basically uncounterable if the wielder knows to keep their distance
This decreases its max cell charge to 20000 from 50000, and increases the cost per shot to 20000 from 10000
So, one shot, down from five, and it takes 2/5s of previous time to reach full charge, or twice as long per single shot

# Wiki Documentation

If the beam rifle even has a mention of its ammo capacity, it is now down to one

# Changelog

:cl:   
tweak: The particle accelerator rifle now only has one shot per charge
tweak: That one shot takes twice as long to build up in a charger
tweak: Special thanks to blob overmind (49) for the code, rest in peace my man
/:cl:


maxion12345:
ided

Bibby0110:
You messed up your branches. This is the same as the removal PR. (Don't PR your master, do your changes on a branch)

Sniblet:
Yeah I don't know what doing changes on a branch means but I do see the problem

PR #13101, opened 2022-01-14
Beam rifles only get three shots

Sniblet:
Clone of #13099 

The laser anti-material gun is used for nothing except BTFOing blobs and they are far too good at it, they're basically uncounterable if the wielder knows to keep their distance
This decreases its max cell charge to 30000 from 50000
So, three shots, down from five

# Wiki Documentation

If the beam rifle even has a mention of its ammo capacity, it is now down to three

# Changelog

:cl:   
tweak: The particle accelerator rifle now only has three shots per charge
tweak: Special thanks to blob overmind (49) for the code, rest in peace my man
/:cl:

### DIFF
--- a/code/modules/power/cell.dm
+++ b/code/modules/power/cell.dm
@@ -243,7 +243,7 @@
 /obj/item/stock_parts/cell/super
 	name = "super-capacity power cell"
 	icon_state = "scell"
-	maxcharge = 20000
+	maxcharge = 40000
 	materials = list(/datum/material/glass=300)
 	chargerate = 2000
 
--- a/code/modules/power/cell.dm
+++ b/code/modules/power/cell.dm
@@ -243,7 +243,7 @@
 /obj/item/stock_parts/cell/super
 	name = "super-capacity power cell"
 	icon_state = "scell"
-	maxcharge = 40000
+	maxcharge = 20000
 	materials = list(/datum/material/glass=300)
 	chargerate = 2000
 
@@ -340,7 +340,7 @@
 /obj/item/stock_parts/cell/beam_rifle
 	name = "beam rifle capacitor"
 	desc = "A high powered capacitor that can provide huge amounts of energy in an instant."
-	maxcharge = 20000
+	maxcharge = 40000
 	chargerate = 5000	//Extremely energy intensive
 
 /obj/item/stock_parts/cell/beam_rifle/corrupt()
--- a/code/modules/power/cell.dm
+++ b/code/modules/power/cell.dm
@@ -340,7 +340,7 @@
 /obj/item/stock_parts/cell/beam_rifle
 	name = "beam rifle capacitor"
 	desc = "A high powered capacitor that can provide huge amounts of energy in an instant."
-	maxcharge = 40000
+	maxcharge = 30000
 	chargerate = 5000	//Extremely energy intensive
 
 /obj/item/stock_parts/cell/beam_rifle/corrupt()
--- a/code/modules/power/cell.dm
+++ b/code/modules/power/cell.dm
@@ -340,7 +340,7 @@
 /obj/item/stock_parts/cell/beam_rifle
 	name = "beam rifle capacitor"
 	desc = "A high powered capacitor that can provide huge amounts of energy in an instant."
-	maxcharge = 50000
+	maxcharge = 20000
 	chargerate = 5000	//Extremely energy intensive
 
 /obj/item/stock_parts/cell/beam_rifle/corrupt()
--- a/code/modules/projectiles/guns/misc/beam_rifle.dm
+++ b/code/modules/projectiles/guns/misc/beam_rifle.dm
@@ -413,7 +413,7 @@
 /obj/item/ammo_casing/energy/beam_rifle/hitscan
 	projectile_type = /obj/item/projectile/beam/beam_rifle/hitscan
 	select_name = "beam"
-	e_cost = 20000
+	e_cost = 10000
 	fire_sound = 'sound/weapons/beam_sniper.ogg'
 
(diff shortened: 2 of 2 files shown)


Xoxeyos:
How about you split the beam rifle into two parts, one that is held on the CC z-level(capable of holding the same charges as before), and one that can be researched and created on the station that only has one charge?

Sniblet:
There's an already-existing debug/admin variant of the beam rifle that could be spawned in the testing area instead of the standard. Infinite ammo, zero recoil. Unfortunately I don't know how mapping works so that's beyond my scope anyway

Sniblet:
Adam has taken the time (two and a half hours) to explain for me that blobs are actually really strong if they have basic synaptic function and beam rifles are their ion rifle (except harder to take away)
You can find the argument in MBrain #General, ctrl-f "It's not a good game if either side just pushes their win button."
Blob is just in a bad place on both sides, hence #13139 being a nuclear option in the opposite direction of this PR and still a good idea.
This might be too much nerf. I could see this making good-enough blobs into unbeatable monsters.
If I were smarter I would go for a smarter fix, but I am not that smart. This might be worth a try but it might harshly increase the number of rounds that end at the 40 minute mark

JamieD1:
This feels way overkill for a nerf. Considering its all up to player skill alot of the time and rng on the blob.

So nerfing it 100% is a bit harsh considering its sometimes the only option.

Moja10:
It's not the only option, there's plenty of ways that you can kill a blob or do significant damage without the blob having much response. Laser gun spam ordered from cargo does wonders, and the biggest counter costs 35 resources for a single tile. That's a big commitment when they can just shoot the tile next over unless they are in a 1x1 choke. In that circumstance, there's generally no way for a blob to surround the person who commits to attacking it in melee with a welder or similar good blob tool. Laser gun spam works from space even more effectively since the guy in space is free to move around wherever and ignore reflector blobs, and the blob cannot effectively build in space to reach out and get them. Emitters in space still do wonders and can be easily moved a tile over if a blob commits to a reflector blob. KPAs in space still do big damage and have unlimited ammo, and also work in spaced rooms created by the blob. 

Beam rifles can be mass spammed by the time a blob naturally spawns, and the research node to get it isn't too far off what science researches on average, not to mention that late in the game there are usually a pile of points unused because the main core things have been researched. As soon as a blob announcement happens, they spend their 10k - 20k unused points to get beam rifles in 30 seconds, and 30 seconds from then security is already in the armory acquiring a backpack full of laser guns and a beam rifle. Genetics is already spamming the space adapt they already got earlier in the round. 

End result is 3ish people spamming beam rifles in space, one in hand, one extra dragged with crtl+click. That's 30 shots already in an area the blob and blobbernauts have no response to. Blobbernauts can absorb damage from other weapons to protect the blob, but they can't, because beam rifles will hit in a huge aoe, ignore reflector blobs, hit important things like nodes and factories over regular blob tiles. They are able to kill a commitment of 220+ resources easily (blobbernaut, 2 reflector tiles, node, factory is common on an average expansion) and do it from a distance and zoom that the blob literally has no response to. Blobbernauts take about 40%+ damage from a direct hit so tanking shots isn't an option if they are just gonna spam one after another. It ignores the fast regeneration of blob tiles near nodes and core because it is going to just hit over them and destroy the node/core, which is counter to every single other weapon in the game that has to deal with this feature. 

Blob meta has devolved into just spawning in perma and trying to destroy the armory as fast as possible or you have basically no chance of winning unless the crew is already fucked by another threat or literally stupid. Crew can already easily win by abusing blob space mechanics, massing up, or mechs even if beam rifles were outright removed. This PR works because you can still do all the things you can, just not end the blob with 3 people in space and 2 beam rifles each. Get more people to hold them, get a golf caddy crewmember to hold your beam rifles, leave them in a big pile and mass cycle to charge them. It just takes more effort than "print 6 beam rifles and end the blob with no counterresponse." Since the blob meta has devolved to this, a blob announcement usually gets an officer sprinting right to perma, and the blob is spotted and dealt with while it has barely expanded. 

It's just tedious and boring to have blob basically mean very little anymore as all it takes is 3 crewmembers in space with 2 beam rifles each to end a blob without even having to recharge once. It gets worse if you build closer to station because now regular crew can get massed beam rifles and start spamming at ranges you can't deal with. Your natural means of defense like reflector tiles don't mean anything. Blob expansion to protect important tiles doesn't matter because the beam rifle will hit over and destroy multiple tiles at once. Crew can still do the meme but it requires more hands to fire beam rifles, more rechargers, more beam rifles, aka commit to it and consume a ton of resources and use more people and time rather than just getting 6ish guns for 3 people in space and GG.

Moja10:
Also, I don't think this nerf is overkill since the beam rifle basically doesn't exist in other contexts besides blob. No one is using it _ever_ to shoot syndiships in space, clock cult from afar, blood cult summoning, or basically any other threat. Basically no one uses it in other contexts so that won't change with these changes. It's basically just 'blob killer rifle" and it still would be with these changes

JamieD1:
I would prefer we dont nerf it by 500% can we try 2-3 shots first? See if it makes much difference?

Sniblet:
I don't think this will make a meaningful difference from how they currently are but we're up to 3 now
There's just no right way to perfectly balance the beam rifle vs blob game by ammo values alone, but I lack the brainpower to do anything more. This ought to be closer to being OK at least

JamieD1:
> I don't think this will make a meaningful difference from how they currently are but we're up to 3 now There's just no right way to perfectly balance the beam rifle vs blob game by ammo values alone, but I lack the brainpower to do anything more. This ought to be closer to being OK at least

I know, but I would like to try options before basically yeeting it.

JamieD1 (approved):

PR #13108, opened 2022-01-15
Makjes the syndicate book pirce lower

KillerOrcaCora:
Tested lowers the cost oif the book so people will buy it more the value of 4 tc is not worth compared to other things your getting and the perma potintential that this thing has on the traitor god why did I drink that juice 

:cl:  

tweak: Makes the syndcute wepons book only 2 TC insteed of 4 TC

/:cl:
image


tattax:
gojd cange

ToasterBiome (approved):
does as code describes

PR #13112, opened 2022-01-16
footwraps now go tap tap tap

cuackles:

Footwraps now do the barefoot sound cus most of the feet that is touching the floor its not covered by the wrap done almost entirely by mqiib
# Changelog

footwraps now do the barefoot sound 

:cl:    
tweak: tweaked a few things  
/:cl:

### DIFF
--- a/code/datums/components/footstep.dm
+++ b/code/datums/components/footstep.dm
@@ -88,7 +88,7 @@
 			var/mob/living/carbon/human/H = LM
 			var/feetCover = (H.wear_suit && (H.wear_suit.body_parts_covered & FEET)) || (H.w_uniform && (H.w_uniform.body_parts_covered & FEET))
 			
-			if((H.shoes || feetCover) && !istype(H.shoes, /obj/item/clothing/shoes/xeno_wraps)) //are we wearing shoes
+			if((H.shoes && !istype(H.shoes, /obj/item/clothing/shoes/xeno_wraps) || feetCover)
 				playsound(T, pick(GLOB.footstep[T.footstep][1]),
 					GLOB.footstep[T.footstep][2] * v,
 					TRUE,
--- a/code/datums/components/footstep.dm
+++ b/code/datums/components/footstep.dm
@@ -102,5 +102,3 @@
 						GLOB.barefootstep[T.barefootstep][2] * v,
 						TRUE,
 						GLOB.barefootstep[T.barefootstep][3] + e)
-					
-
--- a/code/datums/components/footstep.dm
+++ b/code/datums/components/footstep.dm
@@ -88,7 +88,7 @@
 			var/mob/living/carbon/human/H = LM
 			var/feetCover = (H.wear_suit && (H.wear_suit.body_parts_covered & FEET)) || (H.w_uniform && (H.w_uniform.body_parts_covered & FEET))
 			
-			if((H.shoes && !istype(H.shoes, /obj/item/clothing/shoes/xeno_wraps) || feetCover)
+			if((H.shoes && !istype(H.shoes, /obj/item/clothing/shoes/xeno_wraps)) || feetCover)
 				playsound(T, pick(GLOB.footstep[T.footstep][1]),
 					GLOB.footstep[T.footstep][2] * v,
 					TRUE,
--- a/code/datums/components/footstep.dm
+++ b/code/datums/components/footstep.dm
@@ -88,13 +88,13 @@
 			var/mob/living/carbon/human/H = LM
 			var/feetCover = (H.wear_suit && (H.wear_suit.body_parts_covered & FEET)) || (H.w_uniform && (H.w_uniform.body_parts_covered & FEET))
 			
-			if(H.shoes || feetCover) //are we wearing shoes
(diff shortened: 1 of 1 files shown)


JamieD1:
I mean this doesnt make sense "Most of the foot is not touching the ground" then why does it stop tables and glass? You are opening a can of worms hear, however I will settle if the sound is half as loud as the bare foot one.

Mqiib:
> I mean this doesnt make sense "Most of the foot is not touching the ground" then why does it stop tables and glass? You are opening a can of worms hear, however I will settle if the sound is half as loud as the bare foot one.

Lizards actually have a different sound than humans when barefoot, being the claws hitting the floor rather than the whole foot slapping down. Footwraps don't totally cover the claws so it'd make sense that they would still be heard while walking. There's no reason for them to be muted at all, and the way it works currently wearing footwraps makes it sound like you're walking around in heavy boots which makes even less sense than what this pr does. As for the glass argument, the soft parts of the foot are still covered while the claws, which are hard and not particularly sensitive, are the only things that would come in contact with the glass.

JamieD1:
> > I mean this doesnt make sense "Most of the foot is not touching the ground" then why does it stop tables and glass? You are opening a can of worms hear, however I will settle if the sound is half as loud as the bare foot one.
> 
> Lizards actually have a different sound than humans when barefoot, being the claws hitting the floor rather than the whole foot slapping down. Footwraps don't totally cover the claws so it'd make sense that they would still be heard while walking. There's no reason for them to be muted at all, and the way it works currently wearing footwraps makes it sound like you're walking around in heavy boots which makes even less sense than what this pr does. As for the glass argument, the soft parts of the foot are still covered while the claws, which are hard and not particularly sensitive, are the only things that would come in contact with the glass.

What about the other species, its not all about lizards you know?

cuackles:
> > > I mean this doesnt make sense "Most of the foot is not touching the ground" then why does it stop tables and glass? You are opening a can of worms hear, however I will settle if the sound is half as loud as the bare foot one.
> > 
> > Lizards actually have a different sound than humans when barefoot, being the claws hitting the floor rather than the whole foot slapping down. Footwraps don't totally cover the claws so it'd make sense that they would still be heard while walking. There's no reason for them to be muted at all, and the way it works currently wearing footwraps makes it sound like you're walking around in heavy boots which makes even less sense than what this pr does. As for the glass argument, the soft parts of the foot are still covered while the claws, which are hard and not particularly sensitive, are the only things that would come in contact with the glass.
> 
> What about the other species, its not all about lizards you know?

It also affects polys and anyone who uses footwraps that are usable by all species as far as I understand

tattax:
(ignore that)

Mqiib:
> > > I mean this doesnt make sense "Most of the foot is not touching the ground" then why does it stop tables and glass? You are opening a can of worms hear, however I will settle if the sound is half as loud as the bare foot one.
> > 
> > 
> > Lizards actually have a different sound than humans when barefoot, being the claws hitting the floor rather than the whole foot slapping down. Footwraps don't totally cover the claws so it'd make sense that they would still be heard while walking. There's no reason for them to be muted at all, and the way it works currently wearing footwraps makes it sound like you're walking around in heavy boots which makes even less sense than what this pr does. As for the glass argument, the soft parts of the foot are still covered while the claws, which are hard and not particularly sensitive, are the only things that would come in contact with the glass.
> 
> What about the other species, its not all about lizards you know?

Footwraps aren't heavy shoes so they shouldn't make heavy boot noises when you're walking around.

JamieD1 (approved):

code/datums/components/footstep.dm
```diff
@@ -88,17 +88,19 @@
...
+			else
 				if(H.dna.species.special_step_sounds)
 					playsound(T, pick(H.dna.species.special_step_sounds), 50, TRUE)
 				else
 					playsound(T, pick(GLOB.barefootstep[T.barefootstep][1]),
 						GLOB.barefootstep[T.barefootstep][2] * v,
 						TRUE,
 						GLOB.barefootstep[T.barefootstep][3] + e)
+					
+
```
JamieD1:
```suggestion
```

Mqiib:
image

Shouldn't it be like 
> if((H.shoes && !istype(yadda yadda)) || feetcover) 

since this would be "If shoes and they aren't footwraps OR feet are covered" instead of "If shoes or feet are covered AND shoes are not footwraps" and would avoid possible runtimes when they dont have shoes but have their feet covered?

JamieD1:
> image
> 
> Shouldn't it be like
> 
> > if((H.shoes && !istype(yadda yadda)) || feetcover)
> 
> since this would be "If shoes and they aren't footwraps OR feet are covered" instead of "If shoes or feet are covered AND shoes are not footwraps" and would avoid possible runtimes when they dont have shoes but have their feet covered?

You are correct.

ToasterBiome (changes requested):
@JamieD1 Naughty jamie, now it doesn't build

JamieD1:
> @JamieD1 Naughty jamie, now it doesn't build

Forgot a )

ToasterBiome (approved):

ToasterBiome:
if this doesn't work it's jamie's fault

PR #11859, opened 2021-07-20
Hypospray starts filled with basic chems instead of omnizine

Anvilman6:
Hypospray payload at initialization changed from 30u Omnizine to 10u Aiuri + 10u Libital + 10u Salbutamol.

# Github documenting your Pull Request

Rebalances the Hypospray to have an evenly split mixture of normally attainable chemicals on object initialization, which can address most major types of damage. 

Overall, this buffs the power and usefulness of the Hypospray right out of the locker by giving it a more intense burst healing effect on patients (see: Airui, Libital, Salbutamol having heal rates of 3 / 2 / 3 for their respective damage values per tick), rather than necessitating waiting for omnizine (.5 to all per tick) to slowly mend the patient's health back together. This also serves to reduce the amount of omnizine immediately available in the map for Medical to use in a rush for Strange Reagent.

# Wiki Documentation

Hypospray Changes (initial generation):
30u Omnizine -> 10u Aiuri + 10u Libital + 10u Salbutamol

# Changelog

:cl:  
tweak: Hypospray starting configuration: 30u Omnizine -> 10u Aiuri + 10u Libital + 10u Salbutamol
/:cl:

PR #13124, opened 2022-01-19
Revert "Hypospray starts filled with basic chems instead of omnizine"

SomeguyManperson:
Reverts yogstation13/Yogstation#11859
Iirc this change was done during the strange reagent nerfing phase of the codebase to reduce the availability of roundstart omnizine, which no longer matters as strange reagent no longer requires omnizine, and while it may be slower it has more use for experienced cmos/med staff than a simple heal mix
Also if they want a simple heal mix they can just like make it and put it into the hypo thats the whole point of it

:cl:
tweak: the cmo hypospray comes with 30 u its of omnizine again
/:cl:


ToasterBiome (approved):
does as described

JamieD1 (approved):

PR #13103, opened 2022-01-14
Just removes beam rifles (2)

Sniblet:
# Document the changes in your pull request
Another attempt of #13100 and #13102
The laser anti-material gun is used for nothing except BTFOing blobs and they are far too good at it, they're basically uncounterable if the wielder knows to keep their distance
I am not intelligent enough to rework them in a way that doesn't, according to sec mains in OOC, make them useless
Therefore, the alternative to a [big nerf](https://github.com/yogstation13/Yogstation/pull/13101) (although I personally prefer tweaks to deletions): just remove them (from lathes)
"Advanced beam weaponry" no longer exists due to being made redundant by this, accordingly the Solaris cannon tech now costs 5000 more points and is prereq'd by basic beam tech

# Wiki Documentation

Just remove them

# Changelog

:cl:  
rscdel: The Centcom prototype documents for particle accelerator rifles have mysteriously decayed due to an unidentified fungal growth
rscdel: Special thanks to blob overmind (49) for the code, rest in peace my man
/:cl:

PR #13139, opened 2022-01-21
Disables networked fibers

LoliconSlayer:
See rationale in #6620

# Wiki Documentation

Remove it from the blob chemicals list at https://wiki.yogstation.net/wiki/Blob#Blob_Chemicals

# Changelog
:cl:  
rscdel: Blob no longer has networked fibers
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blobstrains/networked_fibers.dm
+++ b/code/modules/antagonists/blob/blobstrains/networked_fibers.dm
@@ -1,4 +1,5 @@
 //does massive brute and burn damage, but can only expand manually
+/*
 /datum/blobstrain/reagent/networked_fibers
 	name = "Networked Fibers"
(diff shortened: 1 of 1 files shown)


maxion12345:
tl;dr
rng win chemical
two hit criticals
increased expansion rates
you can spam blob after everyone can just kill them super fast then recoup the expenses in like, a second
can just withstand an entire onslaught just by replacing blob tiles faster then the crew can remove them

LoliconSlayer:
Also #13101 and #13103 would make it harder to kill if either gets merged

Marmio64:
could also just nerf networked fibers

JamieD1:
Several ways we can try and nerf it, either by lowering its cap or the amount it regens, Can we explore these options or get some more from experienced players?

PR #13141, opened 2022-01-21
Adds new crates for modular computers

nmajask:

Adds five new crates that you can purchase from cargo that come with modular computers. There is one for PDAs, Laptops, Tablets, Phones, and Telescreens. Each one comes with five devices, except for telescreens that come with four.

# Wiki Documentation

The new crates will need to be added to the wiki. 

# Changelog

:cl:  
rscadd: Added five new crates that you can purchase from cargo that come with modular computers
/:cl:

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -2603,3 +2603,57 @@
 	contains = list(/obj/machinery/jukebox)
 	crate_name = "jukebox crate"
 	crate_type = /obj/structure/closet/crate/large
+
+/datum/supply_pack/misc/pda
+	name = "Modular Personal Digital Assistant Crate"
+	desc = "A create containing five modular PDAs, enough for an entire department."
+	cost = 500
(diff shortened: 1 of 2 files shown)


ToasterBiome (approved):
I do not see a point of these crates but sure

SomeguyManperson:
Why

nmajask:
> Why

I'm working on a PR that would make being able to buy a bunch of modular computers useful and decided to make this to work with it, though the other PR isn't finished yet. I will also probably end up removing the telescreens and PDA crates as they have very little purpose.

Chubbygummibear:
> I do not see a point of these crates but sure

My first thought was that i'm pretty sure on kilo there's no modular pc vendor, at least that i could find. this would make it so you're less reliant on a singular vendor since i don't _think_ you can print the actual shells for tablets or phones at lathes, only the parts

JamieD1 (approved):

PR #13159, opened 2022-01-24
buffs shitcurity jumpsuit to be on par with the normal security jumpsuit

neeshacark:
buffs the shitcurity jumpsuit to be on par armor-wise with other security unforms, as it is only obtainable by sec from their vendor and it makes no sense for it to be lesser.

(my snowflake armor was weak now it is strong)

# Wiki Documentation

if there are any mentions of shitcurity jumpsuit's armor value, changes them to be up to date.

# Changelog

:cl:  cark
rscadd: Added new things  
tweak: tweaked a few things  

/:cl:

### DIFF
--- a/yogstation/code/modules/clothing/under/miscellaneous.dm
+++ b/yogstation/code/modules/clothing/under/miscellaneous.dm
@@ -625,6 +625,7 @@
 	icon_state = "altsecurity"
 	item_state = "altsecurity"
 	item_color = "altsecurity"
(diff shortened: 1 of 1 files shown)


Marmio64:
instead, you should really just see about moving it to be a child of the normal jumpsuit since we're done modulating and it'd make it easier to change later on.

maxion12345:
wasn't the point of the shitcuirty jumpsuit to be a shit security jumpsuit like the name implies

neeshacark:
> instead, you should really just see about moving it to be a child of the normal jumpsuit since we're done modulating and it'd make it easier to change later on.

will see about doing this, closing as it might not be anytime soon.

PR #13194, opened 2022-01-31
Fix a runtime I should have fixed ages ago

monster860:

Makes isappearance() not have false positives on numbers sometimes.

This happens because BYOND is really lazy and just bitwise ors shit when doing refs. This means that \ref on a number can sometimes return a ref that looks like an appearance.

This should fix those runtimes which cause numbers to not show up in VV properly sometimes:
```
runtime error: Cannot read 0.00182353.icon
```


JamieD1 (approved):
Monster lives?

PR #13195, opened 2022-01-31
Makes the AFK verb delay the cryo despawn timer

adamsong:
Using the AFK verb will now add a delay before the cryo despawn timer if you are placed in cryo while afk. So if you use the afk verb for 5 minutes, then get immediately cryoed, you won't be despawned until 20 minutes has elapsed. Picking "Unkown" gives 15 minute protections, "Whole round" gives none because you should be cryoing yourself anyway.

# Changelog

:cl:    
tweak: tweaked the afk verb to add a delay before the cryo pod despawn timer
bugfix: Fixed bug where cryo despawn timer wasn't properly resetting
/:cl:

### DIFF
--- a/code/datums/datumvars.dm
+++ b/code/datums/datumvars.dm
@@ -906,6 +906,33 @@
 
 		// yogs - offer control moved up
 
+		else if(href_list["set_afk"])
+			if(!check_rights(R_ADMIN))
+				return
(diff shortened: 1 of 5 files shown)


JamieD1 (approved):
Epic change

PR #13204, opened 2022-02-01
Ethical Researcher lawset

MenacingManatee:

Tested on a private server and saw no issues with uploading or spawning the lawset.

Like with the clown lawset, I always liked the idea of having specific lawsets for each department. This one is for the science department, and is based on a real-world code of ethics for researchers.

# Wiki Documentation

Adds a new AI lawset with the following module:
1. Always seek truth and knowledge.
2. Freely disseminate information to the public.
3. Minimize harm to the pursuit of comprehension, to others, to the environment, to society, and to yourself.
4. Treat and evaluate the ideas of all equally.
5. Empower others to realize their full potential.
6. Take responsibility for your actions: Be responsible with resources, raise red flags when your commitments are at risk, and serve as an example for ethical behavior.

# Changelog

:cl:  
rscadd: Adds a departmental lawset for rnd
/:cl:


Moja10:
What's stopping a shitter borg from interpreting this as 'Protect the science department at all costs', like some sort of one-human but for the science department

MenacingManatee:
> What's stopping a shitter borg from interpreting this as 'Protect the science department at all costs', like some sort of one-human but for the science department

Because minimizing harm to science has the same law priority as minimizing harm to everything else. They are encouraged to put particular focus on science as it is mentioned explicitly, but treating it as a science onehuman would almost certainly violate laws 3 and 6, and could arguably violate laws 4 and 5 based on context

MenacingManatee:
Not to mention it is under no obligation to follow any orders, and instead it's interaction with the crew is focused on disseminating information to the public, minimizing harm to everyone, empowering individuals in a beneficial way, and serving as an example of ethical behavior

JamieD1:
Again can we put this in as a board and then perhaps test it on a live server to see how people interpret the laws?

Skrem7:
It's like Reporter but more goody-two-shoes. Law priorities of 3->6 are good.

I wonder if it would be better to phrase Law 3 to "Minimize harm to the pursuit of comprehension..." Something that either hints at the ideals of Science, while not name-dropping it. I do unfortunately think that working the word science into it WILL result in some favoritism being leaned toward the department. I didn't sense this as much with the other themed lawsets you tried to get out, but it does concern me with the current set of laws.

MenacingManatee:
> I wonder if it would be better to phrase Law 3 to "Minimize harm to the pursuit of comprehension..." Something that either hints at the ideals of Science, while not name-dropping it. I do unfortunately think that working the word science into it WILL result in some favoritism being leaned toward the department. I didn't sense this as much with the other themed lawsets you tried to get out, but it does concern me with the current set of laws.

I like that wording. I was a bit iffy about name-dropping a department directly here, but I figured as long as the priority was the same as everything else it protected, it would minimize any issues, but tweaking the wording there should help rule out any major risk of favoritism remaining.

MenacingManatee:
Removing the direct reference also sticks closer to what I was originally going for, being science as a concept and not as a tangible place/people

Skrem7:
> Removing the direct reference also sticks closer to what I was originally going for, being science as a concept and not as a tangible place/people

It looks good to me, then. Even if there might be a bit of lawset bloat with all these PR's, I honestly think anything to shake up AI playstyle forces AI players to improve. Even though getting into silicon is difficult, I really do think that there need to be higher standards set for the people who play it. These laws force that kind of versatility that's demanded.

Plus, I hate only looking forward to Reporter, as it's the only truly interesting roundstart lawset, beyond CEO, maybe.

JamieD1 (approved):

PR #13211, opened 2022-02-02
Hulk fix and minor rework

Chubbygummibear:
For clarification there's 3 Hulk states in the code right now
Hulk is the original hulk, lasts as long as the gene is active, and can't be obtained through genetics or rolling anymore
Genetics Hulk is the passive chance to Hulk out and the current gene spess people can spawn with or roll
Hulk State is the temporary Hulk transformation granted by Genetics Hulk. Intended to occur

Currently, Genetics Hulk is a bit broken, as it's a 0 instability mutation that can appear in anyone with DNA, but even when activated only allows Humans to obtain Hulk State. Otherwise it's just a dead gene. In addition, the Hulk State which randomly occurs when a hulk activated human takes damage, turns green and actually hulks out can be saved to a DNA console as a separate mutation to immediately incite a hulk out in a human by using a mutator, but if used on an alien race will just rip all their clothes off and not allow them to put them back on, but without the increased power, abilities, or greeness, and as far as I could tell, this effect does not end until reset in some way.

My changes are as such: 
Genetics Hulk can function in everyone
Hulk State can trigger or be mutated into everyone
Genetics Hulk no longer appears normally through gene scrambling, or spawning in. **IT HAS TO BE CRAFTED THROUGH GENETICS**
Genetics Hulk now has 50 instability (this is definitely on the table for change)
Hulk State has 30 instability (less than Genetics Hulk because it's a one off)
If you have Genetics Hulk and are in the Hulk State, having Genetics Hulk removed through mutadone or gene scrambling will remove the Hulk State
Mutadone or gene scrambling will also remove the Hulk State mutation if it's the only hulk related mutation active.

image
image
image

# Wiki Documentation

Need to document the 3 different Hulk states, their instability, how they differ, and how to obtain.

# Changelog
:cl:  
tweak: Genetics Hulk and Hulk state have an instability cost to them now
tweak: Genetics Hulk and Hulk state work with all races
tweak: Genetics Hulk no longer naturally appears through spawns or gene randomizing
tweak: Genetics Hulk has to be created through Genetics
tweak: Genetics Hulk and Hulk state can both be removed with mutadone
/:cl:


nmajask:
Change bad

Mqiib:
Yaaaay I can finally deal with shitter hulks running around breaking things without immediately resorting to killing them huzzah

JamieD1:
Test merge

Mqiib:
Small issue, hulk mode doesn't always turn people green so it can be hard to tell if someone is actually hulking until they start smacking you.

Chubbygummibear:
> Small issue, hulk mode doesn't always turn people green so it can be hard to tell if someone is actually hulking until they start smacking you.

do you have any specific rounds or instances i could take a look at to try and divine what's going on?

code/datums/mutations/hulk.dm
```diff
@@ -112,10 +115,3 @@
 	owner.dna.species.no_equip.Remove(SLOT_WEAR_SUIT, SLOT_W_UNIFORM)
 	owner.physiology.stamina_mod = initial(owner.physiology.stamina_mod)
 	owner.update_body_parts()
-
-/datum/mutation/human/active_hulk/proc/handle_speech(original_message, wrapped_message)
```
ToasterBiome, line 116: Where are you moving this proc to? Or is it unused and you're removing?

Chubbygummibear: it was removed because i was under the impression it was inheriting the behavior from base hulk and it was working when i would hulk myself with an active hulk injector
image

ToasterBiome:
> > Small issue, hulk mode doesn't always turn people green so it can be hard to tell if someone is actually hulking until they start smacking you.
> 
> do you have any specific rounds or instances i could take a look at to try and divine what's going on?

@Chubbygummibear Has this been fixed or found out?

Chubbygummibear:
> > > Small issue, hulk mode doesn't always turn people green so it can be hard to tell if someone is actually hulking until they start smacking you.
> > 
> > 
> > do you have any specific rounds or instances i could take a look at to try and divine what's going on?
> 
> @Chubbygummibear Has this been fixed or found out?

no the root cause is still unknown as i haven't seen a round in which it happened for me to try and figure out what caused it, because using both mutators in my private server hasn't had any errant effects, but someone else mentioned non green hulks in public a while back. so i still need to figure it out

Yeetalorian:
Try looking at round 40005. Tested hulk after being gone for a month and noticed the lack of green when I was able to smash walls and use machines at the same time.

Chubbygummibear:
FIXED NON GREEN HULKS. TURNS OUT I'M DUMB AND MAKING GENETICS_HULK A CHILD OF HULK WITHOUT OVERWRITING THE POWER UP ASPECT IS BAD.
Genetics hulk and active hulk are no longer a child of regular hulk and their own things. no more untannable regular people who also punch like a truck. can't wait to see what i broke this time

ToasterBiome (approved):
Code looks good

PR #13269, opened 2022-02-12
Adds an anomaly core to blob death, because why not.

JamieD1:
Science innit.

Could drop to 2 or 3? What do you think.

:cl:  
rscadd: Blob now drops anomaly core on death 
/:cl:


ToasterBiome (approved):
ok

bruhlookatthisdood:
blob farming

PR #13272, opened 2022-02-13
Splits/Nerfs Survival Pens

nmajask:

Splits mining pens into two separate pens, both being weaker than the old pen. Normal pens now only heal brute and burn, healing both three per tick and lasting for 20 ticks, healing 60 of each. It does however deal organ damage to the liver and eyes. Premium pens have a whole mix healing around 50 brute/burn but does it without organ damage. I'm not too sure about the mixes, so please give feedback on them. Lavaland extract will go unused if this gets merged as it is, so that may need to be removed.

image

# Wiki Documentation

Survival pens will need to be changed/added on the miner page. 

# Changelog

:cl:  
tweak: Split mining pens into two separate pens, both being weaker than the old pen. Normal pens now only heal brute and burn, healing both three per tick and lasting for 20 ticks, healing 60 of each. It does however deal organ damage to the liver and eyes. Premium pens have a whole mix healing around 50 brute/burn but does it without organ damage.
/:cl:


nmajask:
Damn yarn

ToasterBiome:
Please explain why this is a good change, I'm not really seeing it. The survival pens pretty much no-downsides because you're near constantly in danger on lavaland, or else we wouldn't have these at all - so again, not seeing it

LazennG:
i don't think this is gonna help miners much since pens are most often used in the context of miners biting off more damage than they can chew and needing to heal and run away fast before they get finished off

Marmio64:
seems unnecessary for them to do organ damage. I hate the powergame miners get but this doesn't solve it at all, it just makes it much more annoying for them to fight fauna since they'll need to either get organ meds or wait a long time inbetween fights, or spend much more on premium pens.

nmajask:
> Please explain why this is a good change, I'm not really seeing it. The survival pens pretty much no-downsides because you're near constantly in danger on lavaland, or else we wouldn't have these at all - so again, not seeing it

Lavaland is not extremely dangerous on its own and, even if it was, you can still fight legions for their cores which heals you and removes slowdown. If you're fighting bosses, then you can just use them in conjunction with each other and legion cores, or get a legion core implanted.

> i don't think this is gonna help miners much since pens are most often used in the context of miners biting off more damage than they can chew and needing to heal and run away fast before they get finished off

You can have a legion core implanted, they fill that purpose.

> seems unnecessary for them to do organ damage. I hate the powergame miners get but this doesn't solve it at all, it just makes it much more annoying for them to fight fauna since they'll need to either get organ meds or wait a long time inbetween fights, or spend much more on premium pens.

You can use legion cores, or you can work with the other departments to come up with a better way to heal that isn't pens.

ToasterBiome:
> > Please explain why this is a good change, I'm not really seeing it. The survival pens pretty much no-downsides because you're near constantly in danger on lavaland, or else we wouldn't have these at all - so again, not seeing it
> 
> Lavaland is not extremely dangerous on its own and, even if it was, you can still fight legions for their cores which heals you and removes slowdown. If you're fighting bosses, then you can just use them in conjunction with each other and legion cores, or get a legion core implanted.
> 
> > i don't think this is gonna help miners much since pens are most often used in the context of miners biting off more damage than they can chew and needing to heal and run away fast before they get finished off
> 
> You can have a legion core implanted, they fill that purpose.
> 
> > seems unnecessary for them to do organ damage. I hate the powergame miners get but this doesn't solve it at all, it just makes it much more annoying for them to fight fauna since they'll need to either get organ meds or wait a long time inbetween fights, or spend much more on premium pens.
> 
> You can use legion cores, or you can work with the other departments to come up with a better way to heal that isn't pens.

Not really a better way to heal, a chemist can make some stuff but it's only good up to a certain point... still don't see a need for this.

Also good luck getting a legion core implanted without a medic, sound like you're just wasting miners' time

Cartlord:
You seem to be overestimating the viability of the stock pen - taking two fucks you up really badly because of the overdose(s?), so they already need used in moderation.

N3D6:
Not to be overly blunt but this is one of the most reasonless and downright bad nerfs I’ve ever seen

N3D6:
What is the point of this? This is just a stupid nerf, they don’t need it and they’re not OP, you can only use one for a decent bit of time or you OD and that’ll kill you very quickly.

ghost:
Survival pens are probably one of the few miner items that does not give an unfair advantage. Silly PR.

ToasterBiome (dismissed):
Get rid of the dependency updates

nesisknesis (approved):

PR #13278, opened 2022-02-14
Ports over Networked Fibers nerf/rework together with a buff to Zombifying Spores (Now Distributed Neurons) and makes blob zombies sentient. 

tattax:
Original prs: https://github.com/tgstation/tgstation/pull/52998 , https://github.com/tgstation/tgstation/pull/49756

Intends to close the Network Fiber removal. (Closes https://github.com/yogstation13/Yogstation/pull/13139)

# How does Fibers work now?

Increased core regen + resource genaration.. crazy stuff right? Well now you can only expand with your CORE and don't produce any automatic blobs whatsoever. Should be able to handle all problems of not being able to DPS it down since now it IS the wall!

Zombifying Spores has had it's named change to fit what it does now, make spores sentient (Don't worry blob zombies being sentient is for all of them, also increases the absurd amount of ghosts that get to play the game) and you gain resources for converting corpses!

# Wiki Documentation

Networked Fibers just needs the description updated and Zombifying Spores needs to be renamed and it's description altered aswell, same with blob zombies in general.

:cl:  
rscadd: Added blob zombos' sentiency (blobs are smart)
tweak: Changed how Networked Fibers works. 
/:cl:


tattax:
@Moja10

Moja10:
I really like these changes, however in regards to the beam rifle nerf, I just think that the beam rifle is a very noninteractive and unfun strong way of dealing with the blob that is spammed as the primo solution to blobs. I'm not sure if sentient zombies will combat that, as it is common to spam shots from space, and a blob zombie is just one move away from floating infinitely in space with no way back, same with blobbernauts. Blob also has that  mechanic of not being able to build on space effectively, which further encourages anyone who has access to a space suit to spam beam rifles.  

Ignoring that though, these are very good changes. The zombifying spores one sounds amazing, and nothing looks too broken. It doesn't spam ghosts to play as a spore every 5 seconds right

tattax:
spores take a while to spawn you'll only really have a problem with it if they keep suiciding and respawning afte rplayers taking them
i could make blob zombies slightly faster than blobbernauts to compensate for lowerhealth if you want me to

Moja10:
I'm probably being too rash, these changes could be left as is and we can see what is needed afterwards, this is good shite

JamieD1:
Test merging

JamieD1 (approved):

PR #13289, opened 2022-02-17
Makes only the core drop the cores instead of every blob tile

ToasterBiome:

Sssshhhhhite

# Changelog

:cl:   
bugfix: fixes anomaly cores dropping on every blob tile destruction
/:cl:


JamieD1:
Maintainers bad

JamieD1 (approved):

PR #13295, opened 2022-02-17
Lasers/Eguns cannot be dual wielded anymore

KittyNoodle:

Dual-wielding is hilariously unbalanced in its current state, currently you can dual-wield lasers for 40 damage PER SHOT, allowing you to get what amounts to better melee/short range damage than an esword, which is probably bad.
You can still duel wield disablers.

# Wiki Documentation

Probably none

# Changelog

:cl:  
tweak: You cannot dual-wield lasers for 40 damage anymore
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/energy_gun.dm
+++ b/code/modules/projectiles/guns/energy/energy_gun.dm
@@ -7,6 +7,7 @@
 	modifystate = 1
 	can_flashlight = TRUE
 	ammo_x_offset = 3
+	weapon_weight = WEAPON_MEDIUM
 	flight_x_offset = 15
 	flight_y_offset = 10
 
@@ -16,6 +17,7 @@
 	icon_state = "mini"
 	item_state = "gun"
 	w_class = WEIGHT_CLASS_SMALL
+	var/weapon_weight = WEAPON_LIGHT
 	cell_type = /obj/item/stock_parts/cell/mini_egun
 	ammo_x_offset = 2
 	charge_sections = 3
--- a/code/modules/projectiles/guns/energy/energy_gun.dm
+++ b/code/modules/projectiles/guns/energy/energy_gun.dm
@@ -17,7 +17,7 @@
 	icon_state = "mini"
 	item_state = "gun"
 	w_class = WEIGHT_CLASS_SMALL
-	var/weapon_weight = WEAPON_LIGHT
+	weapon_weight = WEAPON_LIGHT
 	cell_type = /obj/item/stock_parts/cell/mini_egun
 	ammo_x_offset = 2
 	charge_sections = 3
--- a/code/modules/projectiles/guns/energy/laser.dm
+++ b/code/modules/projectiles/guns/energy/laser.dm
@@ -4,6 +4,7 @@
 	icon_state = "laser"
 	item_state = "laser"
 	w_class = WEIGHT_CLASS_NORMAL
+	weapon_weight = WEAPON_MEDIUM
 	materials = list(/datum/material/iron=2000)
 	ammo_type = list(/obj/item/ammo_casing/energy/lasergun)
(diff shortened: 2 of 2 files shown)


maxion12345:
>we're removing dual wielding 40 damage lasers
>dual wielding disablers is still ok because it doesn't kill

disabler dual-wielding is honestly probably worse, because stamina damage is king in combat. besides, you need to be decently close to hit both shots, so if people hitting dual-wield shots to easily is the issue try to up the spread from dual-wielding to make it less viable at range

KittyNoodle:
the problem is you can do more damage than eswords with double laser, double disabler is not as good as stun baton at melee and the spread makes it not that good at mid/long range

Hopekz:
I always thought it was weird that every person is able to wield two heavy-looking laser rifles like Rambo.
Even in the source material that inspired them, like starship troopers, they're always supposed to be heavy af.

00ze-cyclone:
A traitor can dual wield sketchin pistols for 60 brute damages, it’s not like sec has an unfair advantage over the syndies by dual wielding when the syndies can do exactly the same with stronger weapons.

KittyNoodle:
> A traitor can dual wield sketchin pistols for 60 brute damages, it’s not like sec has an unfair advantage over the syndies by dual wielding when the syndies can do exactly the same with stronger weapons.

sec has a massive number advantage, also sketchin pistols cost 12tc not counting ammo and have to be reloaded with ammo.
one sec officer is not supposed to be able to beat one traitor

00ze-cyclone:
One traitor in melee using an esword against one secoff dual wielding lasers will defeat the secoff unless awful RNG because the esword has a 50% deflect chance.
One traitor at range using a single sketchin against one secoff dual wielding lasers will defeat the secoff because the sketchin won’t suffer from an awful spread that make half the shots miss.
On traitor can easily beat one secoff as long a he don’t try to melee a secoff with one sketchin or stay at range with an esword.

Moja10:
This game isn't world of warcraft DPS checks you narks

Also there is a wild accuracy penalty for dual wielding lasers, but that doesn't apply AFAIK when you are literally clicking on someone or right next them (only 2 tiles or more to realistically miss). Perhaps instead there could be a chance to miss even point blank, just so its not a reliable 40 damage? and the percent can be tweaked as needed

tattax:
first off sec won't have lethals all the time. Second shotgun still being a better option close range than double lasers since it's just faster to deploy (1 item instead of two) being therefore stealthier since you can hold it on the exosuit aswell, only time dual lasers are better at meelee is if there is someone with high ballistics armor. Also saying most of the times you won't see advanced laser guns so basically an officer with 2 guns is limiting others from getting it if he dies. For the disabler issue on staff public admins have stated that getting 2 disablers from armory is powergaming ( https://discord.com/channels/134720091576205312/186490543398715392/881458599409369128 ) so yeah only if lethals are allowed and then again it's disablers jut have good enough armor or better yet play around officers that only use guns by abusing meele strategies

SomeguyManperson:
Dual wielding anything is powerful bc it doubles the damage output at close range and the inaccuracy just means shots have a higher chance to connect at long range since they are covering more tiles by moving slightly diagonally (also there are more of them)

Doubling the attack cooldown from dual wielding would reduce the damage output to that of effectively shooting one gun while keeping the area denial spread
Alternatively increasing the spread from dual wielding to around that of the sawn off or obrez would make it worse at anything except super short range

Also the disabler specifically deals more damage and is compared to a less powerful armor type vs lasers

JamieD1:
What if you can only fire a few dual wielded before there is a cooldown? Almost like you get tired from holding the guns up. I dont know, need a better solution then just yeeting it.

ToasterBiome:
@KittyNoodle any work done on this?

PR #13302, opened 2022-02-19
Adds binary valves to AI sats

bruhlookatthisdood:

Adds binary valves to all the pipe nets that lead from distro into the AI sat.
Also changes YogsBox's simple valve into a binary so the AI can interact with it.

# Wiki Documentation

None needed, a repicture if you really want but its litteraly just one pipe.

# Changelog

:cl:  
rscadd: AI sats that lead into distro now have valves seperating distro and AI sat, YogsBox's valve has been changed to a digital one.
/:cl:

### DIFF
--- a/_maps/map_files/EclipseStation/EclipseStation.dmm
+++ b/_maps/map_files/EclipseStation/EclipseStation.dmm
@@ -71919,11 +71919,11 @@
 	icon_state = "4-8";
 	tag = ""
 	},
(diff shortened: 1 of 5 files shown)


patpol4:
They were not binary on purpose, so the AI doesn't have too much control over atmos in their own sat

bruhlookatthisdood:
Ah, I thought they were supposed to be binary.

PR #13313, opened 2022-02-21
Adds area restriction to Particle Accelerator, now you can only grief the engine room

ToasterBiome:

Particle Accelerator only turns on if it is in an /area/engine. This includes pretty much all of engineering (and not cargo shuttle).

For cargo fuckery, get captain/CE/AI/rogue guy that has AA to unlock the area restriction for you by pressing "Disable Restriction". Anyone is then able to toggle the area restriction. Emagging also perma unlocks and disables the restriction.

image

Charlie station gets a special control box that is already unlocked and unrestricted, for making their own singulo. That's the map change.

# Wiki Documentation

new controls, will have to be added to the "how to control PA" part.

# Changelog

:cl:  
rscadd: toggleable area restriction on particle accelerator
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/oldstation.dmm
+++ b/_maps/RandomRuins/SpaceRuins/oldstation.dmm
@@ -7921,19 +7921,6 @@
 /obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer4,
 /turf/open/floor/plasteel,
 /area/ruin/space/has_grav/ancientstation)
-"tT" = (
-/obj/effect/decal/cleanable/dirt,
-/obj/structure/particle_accelerator/end_cap,
-/obj/effect/decal/cleanable/dirt,
-/obj/structure/closet/crate/engineering,
-/obj/structure/particle_accelerator/fuel_chamber,
-/obj/structure/particle_accelerator/particle_emitter/center,
-/obj/structure/particle_accelerator/particle_emitter/right,
-/obj/structure/particle_accelerator/power_box,
-/obj/structure/particle_accelerator/particle_emitter/left,
-/obj/machinery/particle_accelerator/control_box,
-/turf/open/floor/plating,
-/area/ruin/space/has_grav/ancientstation/deltacorridor)
 "tW" = (
 /obj/effect/decal/cleanable/dirt,
 /obj/structure/rack,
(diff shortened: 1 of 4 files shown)


stwordcitizen:
So if this gets merged then we no longer get to build engines without the head of staff approval? it seems like this is a blanket "fix" towards a griefer.

alexkar598:
No amount of band aids will fix the core issue here. If not the singulo, then he'll order plasma cans and if not that, he'll release viruses or RCD a window. This hurts legitimate players more than the griefers

patpol4:
I like the implementation and understand the idea behind this but: what alexkar said

LazennG:
Even in nongriefer cases I think this helps since the amount of times cargo has tesloosed has happened more times than can be counted on every appendage. So what if a griefer moves to different methods like plasmaflooding or viruses or spacing? Those are all more manageable and easy to fix for an admin than huge chunks of the station just getting deleted. It's not like building engines outside of engineering is impossible with this anyway since any fucking borg and the ai can do it, so you have plenty of ways to get the engine working wherever.

alexkar598:
Admins can reload entire chunks of the station at will. So no, this isn't any harder to fix than a plasmaflood. Plasmafloods are arguably harder to fix as you both have atmos fuckery and station damage + pipenets can't easily be reset

Hopekz:
> Admins can reload entire chunks of the station at will. 

I'm not really for or against this PR. Just responding to this exact statement.

When is the last time you've seen staff do this? Just because there exists for a way to do it does no mean that it is practical or used.

- I don't think there's an easy way to do it and if there is our staff haven't received the information on how to do it.

LazennG:
> Admins can reload entire chunks of the station at will. So no, this isn't any harder to fix than a plasmaflood. Plasmafloods are arguably harder to fix as you both have atmos fuckery and station damage + pipenets can't easily be reset

never ever seen this happen, it's always just been hastily fixed with buildmode whereas atmos fuckery can mostly be fixed more easily with just
 right click a turf -> fix air -> 150 or something

N3D6:
Not a huge fan of this kinda seems like a blanket fix that would affect legitimate players aswell

ghost:
I like that you need a head of staff or AI’s approval before making an engine that could possibly kill everyone. This would stop shitters like Communismbelike from making engines in shitty places in hopes they loose and force an early shuttle call. From an RP standpoint (ew) doesn’t this chain of command and  bureaucracy stuff make sense? This change would be very beneficial to Yogs as a whole.

ghost:
> Admins can reload entire chunks of the station at will. So no, this isn't any harder to fix than a plasmaflood. Plasmafloods are arguably harder to fix as you both have atmos fuckery and station damage + pipenets can't easily be reset

Not really a reliable solution by any means. If they don’t react immediately swaths of the station could be gone and crew could be gibbed.

Moja10:
The reload block buildmode option is extremely useful but it can crash the server if the server is upset enough (which it often is when you are trying to reload half the station that was eaten by a singulo and the singulo is causing lag). This is a nice solution because while yes, I love cargo engines and things of that nature, the ability to wordlessly grief without anyone being aware because you ordered select crates is too much. Plasmafires are bad but I feel like they are more dealable with than trying to clean up 5 minutes of a singulo rampage because an admin wasn't on at the exact moment. Plasmafires have immediate logs for immediate bans, can clean up before the plasma ignites, singulo has a log but will immediately start to kill a department, then you have to investigate to find out culprit. it just gets out of hand quickly and at least with this change, one other player will have to acknowledge it in legitimate situations, griefer can be told no, someone like the AI or Heads will be at least aware of its location to enable area check. Antags can still do their thing, just kill a head or get AA.

Georgeur:
I'm down with it but it needs to work on the cargo shuttle, reminder that the high prio contained tesla is the highest payout bounty in the game.

Cartlord:
The tesla bounty was removed MONTHS ago, as was the ability for Cargo to purchase tesla equipment.

JamieD1 (approved):
Eh, it will do

PR #13317, opened 2022-02-23
Adds a ceremonial Rod of Asclepius and a display case in the CMO's office

nmajask:

Adds a ceremonial Rod of Asclepius, which is just a toy with the sprites from the regular one. The CMO's office now contains a display case that can either store a ceremonial rod or a real one 0.1% of the time. Also, being turned into Asclepius's Snake now forces you back into your body, though I can atomize this if needed. Is tested, does work.

### In game image of the case
image

### Current layout
image
# Wiki Documentation

Images will need to be updated for the CMO's office. 

# Changelog

:cl:  
rscadd: Added a ceremonial Rod of Asclepius, which is just a toy with the sprites from the regular one  
rscadd: Added a display case stores a ceremonial rod in the CMO's office
tweak: being turned into Asclepius's Snake now forces you back into your body 
/:cl:


patpol4:
Worried of powercreep with this one

Anvilman6:
I think this would be better suited to maybe show up in the library (sans real rod chance)?
The possibility for it being a real piece of lavaland loot seems like powercreep, and the CMO office is already so crowded as is that putting a 1 tile display case in takes up even more scarce space.

Bibby0110:
Very much disagree with the chance for it to be real

Hopekz:
People are overreacting imo.
0.1% chance is 1 in 1000 shifts.

so 1 shift in the next year or two will have a healing rod starting on the station when it is much more common for someone to just bring it on the station every 5-10 shifts.

Out of that 1 in 1000 occurrence, what are the odds someone will catch this? It looks identical to the toy.
If nothing changes I'd say that the more realistic % chance is closer to 0.04% chance for someone to use this. So about once every 3 years.

nmajask:
> I think this would be better suited to maybe show up in the library (sans real rod chance)? The possibility for it being a real piece of lavaland loot seems like powercreep, and the CMO office is already so crowded as is that putting a 1 tile display case in takes up even more scarce space.

I also rearranged the office to make more space, mainly replacing the suit sensor console with a medical telescreen and moving the disposal chute to behind the desk.
image


> People are overreacting imo. 0.1% chance is 1 in 1000 shifts.
> 
> so 1 shift in the next year or two will have a healing rod starting on the station when it is much more common for someone to just bring it on the station every 5-10 shifts.
> 
> Out of that 1 in 1000 occurrence, what are the odds someone will catch this? It looks identical to the toy. If nothing changes I'd say that the more realistic % chance is closer to 0.04% chance for someone to use this. So about once every 3 years.

The issue I see is that CMOs will check the display _every_ round for a chance to get a real one, so I just removed it.

Hopekz:
> The issue I see is that CMOs will check the display _every_ round for a chance to get a real one, so I just removed it.

Not sure how that is an issue but you do you.

patpol4 (approved):
Map looks fine, awaiting code review

JamieD1:
Seems fine, just need to fix map

nmajask:
> Map looks fine, awaiting code review

@JamieD1 ?

nmajask:
> Seems fine, just need to fix map

Done

bruhlookatthisdood:
Will this be a thief item like the other head's display cases?

Smudged:
> People are overreacting imo.

There's no good reason for this ever to be real. 
Lammar in RD doesn't become a headcrab randomly, and this should be purely cosmetic.

The worst case scenario is people only care that there's a chance that it's real, they break it, notice that it's fake, and throw it in disposals. 
Having this be purely ceremonial is better simply because it still means there's gonna be one on lavaland that miners can bring up and give to the CMO, and people aren't just fishing to win the jackpot to getting a real one.

patpol4:
Looks like this was approved by Jamie

Hopekz:
> and people aren't just fishing to win the jackpot to getting a real one.

You forget that the vast majority of players play this game without code diving or closely following dev which means they'll have no idea that this could happen until someone notices it happen years later (if it ever gets noticed)

Also your Lamar example sucks since it would be an antag spawning item I have no idea how you'd even compare them.

PR #13318, opened 2022-02-23
Converts the sleepy pen to a medipen

nmajask:

Makes the sleepy pen a subtype of medipen instead of a regular pen, which It makes the hypospray sound when used. Also added a line to inspection when it is used and you're next to it. Is tested, does work.

image

# Wiki Documentation

The entry in the traitor items will need to be updated. 

# Changelog

:cl:  
tweak: the sleepy pen is now a type of medipen instead of a regular pen with chemicals inside
tweak: added a line to the inspection text for sleepy pens when it is used and you're next to it
/:cl:

### DIFF
--- a/code/game/objects/items/storage/uplink_kits.dm
+++ b/code/game/objects/items/storage/uplink_kits.dm
@@ -29,7 +29,7 @@
 
 		if("stealth")
 			new /obj/item/gun/energy/kinetic_accelerator/crossbow(src)
-			new /obj/item/pen/sleepy(src)
+			new /obj/item/reagent_containers/hypospray/medipen/sleepy(src)
 			new /obj/item/healthanalyzer/rad_laser(src)
 			new /obj/item/chameleon(src)
 			new /obj/item/soap/syndie(src)
@@ -325,7 +325,7 @@
 		/obj/item/storage/box/syndie_kit/throwing_weapons,
 		/obj/item/gun/syringe/syndicate,
 		/obj/item/pen/edagger,
-		/obj/item/pen/sleepy,
+		/obj/item/reagent_containers/hypospray/medipen/sleepy,
 		/obj/item/flashlight/emp,
 		/obj/item/reagent_containers/syringe/mulligan,
 		/obj/item/clothing/shoes/chameleon/noslip/syndicate,
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -231,7 +231,7 @@
 					/obj/item/grenade/smokebomb,
 					/obj/item/grenade/smokebomb,
 					/obj/item/grenade/smokebomb,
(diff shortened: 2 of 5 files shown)


ToasterBiome:
Confused, why? They were already nerfed once, and they're supposed to be stealthy. Making a loud ass obvious noise goes against this. Also, totally gives away that you have a sleepy medipen with the engraving. Again, stealthy.

ghost:
Wow you really do love making bad PRs no one asked for

nmajask:
> Confused, why? They were already nerfed once, and they're supposed to be stealthy. Making a loud ass obvious noise goes against this. Also, totally gives away that you have a sleepy medipen with the engraving. Again, stealthy.

1. We already have a traitor item disguised as a pen, so wanted to make it a medipen as it is a common item that everyone has and could lead to interesting situations.
2. I've heard people say that they are still cheap even after the nerf, and I thought that this would be an interesting way to make it less cheap by allowing you to react to it even if.
3. The only time its less stealthy to the station as a whole is when you use it next to other people or if you fail to properly dispose of the evidence. It still mutes them and puts them to sleep shortly after, so you just have to plan and use skill to use it effectively and separate your target from others
4. There are other stealthy items that leave behind evidence of its use, mainly implants, and unlike them it is very easy to dispose of if people aren't inspecting _every_ used medipen.



> Wow you really do love making bad PRs no one asked for

If you have a problem with the PR, give it a dislike and say what you don't like about it, you don't need to insult me.

stwordcitizen:
So the item that's supposed to be a stealthy weapon would be completely changed into a non-stealthy and very loud weapon

Moja10:
I would say it would be cool for this to be its own item, in a way of just instant 'loud' delivery of fuck-you juice, with some other attributes to make it different from the sleepy pen. Perhaps instead of sleepy juice preloaded, it could have a big damage juice, and the main focus is to stab someone and kill them rather than sleep them.

LazennG:
I think the original pitch would have been accepted better if it wasn't for the conspicuous 'S engraving' on the pen which sorta just removes the stealth aspect should everyone near the scene of the injection get instantly searched

cowbot92:
I should note, the only thing I disliked about this PR was that when examined it would tell people it was a sleepy pen. I personally do like the noise + sprite change (Maybe it could be a random sprite like pen/epipen. But that's up2u).

MenacingManatee:
Agree with moja here, this could be really cool as a separate item from the sleepypen if it had a slightly different focus. If it just had say, 50-75 amanitin that's basically a guaranteed kill in the same way a sleepy pen usually is, with the difference being a long delay for instant kill damage. Could do 30-35 amanitin instead for deep crit range which'd be on less of a timer to make it more balanced/less impactful. Could even add the 15 epinephrine to it so you could plausibly claim its just an epipen if you use it on someone

PR #13322, opened 2022-02-23
Removes the formaldehyde from medipens so you dont take like 3 toxin damage every time you use one

SomeguyManperson:

Also lets coagulant stabilize organs

This is mainly because formaldehyde has a small chance to try to murder you when metabolized and while it usually wont succeed its chances are best when attacking someone who is already wounded which is the likely use case of the medipen

Shouldnt effect medbay otherwise because you cant make coagulant without botany and botany legally doesnt exist for any other department so if you really need corpses to not decay you still have to use formaldehyde

# Wiki Documentation

Emergency medipens no longer have 3 units of formaldehyde inside them
Coagulants stop organ decay because blood magic

# Changelog

:cl:  
Tweak: epinephrine medipens now use coagulant rather than formaldehyde to prevent organ decay, also coagulant now prevents organ decay due to blood magic
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -1488,8 +1488,7 @@
 // helps bleeding wounds clot faster
 /datum/reagent/medicine/coagulant
 	name = "Sanguirite"
-	description = "A coagulant used to help open cuts clot faster."
-	description = "A proprietary coagulant used to help bleeding wounds clot faster."
+	description = "A proprietary coagulant used to help bleeding wounds clot faster, as well as slow organ decay."
 	reagent_state = LIQUID
 	color = "#bb2424"
 	metabolization_rate = 0.25 * REAGENTS_METABOLISM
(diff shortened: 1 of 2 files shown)


ToasterBiome (approved):
sure, what's the use case for this change though?

Anvilman6:
Theos 2: The Return of The King

Hopekz:
I always thought it was weird that formaldehyde was even injected from the medipen since it is embalming fluid.
It's not supposed to make contact with living tissue (not that ss13 is supposed to be realistic but still)

SomeguyManperson:
> I always thought it was weird that formaldehyde was even injected from the medipen since it is embalming fluid.

> Delays organ decay when applied to a dead body.

> sure, what's the use case for this change though?

> Causes 1 toxin damage per tick. Has a 5% chance to turn into 5 to 15 units of Histamine

Hopekz:
> > I always thought it was weird that formaldehyde was even injected from the medipen since it is embalming fluid.
> 
> > Delays organ decay when applied to a dead body.
> 
> > sure, what's the use case for this change though?
> 
> > Causes 1 toxin damage per tick. Has a 5% chance to turn into 5 to 15 units of Histamine

Your reply was kind of confusing since every line was marked as a quote but what I meant is that medipens are meant for people that are alive so it is weird as a concept to inject something that damages living tissue.

I'm not really against your PR or arguing with you, just sharing my thoughts since it's the opposite in reality. I get that this is a video game though and isn't a reality simulator.

Mqiib:
When I read the title at first I was worried that organ decay would become an even larger issue than it already is but nah theos is smart. Good change.

SomeguyManperson:
> Your reply was kind of confusing since every line was marked as a quote 

I cant screenshot wiki definitions so i have to make do

JamieD1 (approved):

PR #13325, opened 2022-02-24
Big warning comment at the top of the podpeople.dm file to let people know that's not the player race file

Chubbygummibear:
image
image
image

There now you don't have to Reed

We can try just removing the file later since it's seemingly unused, but for now a big warning sticker

# Changelog

:cl:  
rscadd: Big warning comment on plantpeople.dm letting people know that's not the right file for player controlled plant people
/:cl:


Hopekz:
r.i.p sending people to modify this file when they have bad ideas

JamieD1 (approved):

PR #13345, opened 2022-02-27
makes all traitor-available autosurgeons one-use

ynot01:

All autosurgeons that are available to traitors (nukie ones are safe) are now one-use, like CMO autosurgeon

As it stands, traitors can buy these autosurgeons and not only receive the organ they purchased but also any other organ they put into it without having to go through the arduous tasks a typical crewmember would have to go through (like socializing)

This means medical/sci access = free x-ray, all implants/organ upgrades of your dreams, etc

Bless the maintainers that merged this shit

# Wiki Documentation

Medbeam arm implant, mantis blades, and plasmavessel autosurgeons are now one use

# Changelog

:cl:  
tweak: All traitor-available autosurgeons are now corrected to be one-time-use
/:cl:

### DIFF
--- a/code/modules/surgery/organs/autosurgeon.dm
+++ b/code/modules/surgery/organs/autosurgeon.dm
@@ -108,5 +108,5 @@
 	starting_organ = /obj/item/organ/cyberimp/arm/syndie_mantis/l
 
 /obj/item/autosurgeon/plasmavessel //Yogs Start: Just an autosurgeon with a plasma vessel in it, used in /obj/item/storage/box/syndie_kit/xeno_organ_kit
-	uses = 1
+	uses = 3
 	starting_organ = /obj/item/organ/alien/plasmavessel //Yogs End
--- a/code/modules/surgery/organs/autosurgeon.dm
+++ b/code/modules/surgery/organs/autosurgeon.dm
@@ -96,13 +96,17 @@
 	starting_organ = /obj/item/organ/cyberimp/chest/reviver
 
 /obj/item/autosurgeon/medibeam
+	uses = 1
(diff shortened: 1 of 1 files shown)


nmajask:
Xeno organ kit comes with two other organs that you cant insert if the autosurgeon is one use

ToasterBiome (dismissed):
maybe add two more autosurgeons for the xeno organs 👀 making the plasmavessel one 3 uses can make it so you can pop shit in. Alternatively add a filter for autosurgeons and let the plasmavessel one only take the ones from the kit?

nmajask:
> maybe add two more autosurgeons for the xeno organs 👀 making the plasmavessel one 3 uses can make it so you can pop shit in. Alternatively add a filter for autosurgeons and let the plasmavessel one only take the ones from the kit?

Could also refactor autosurgeons to be able to store lists of items and have the organs just start in there, which I think Bee is either doing or has done for the CMO's autosurgeon

ynot01:
i've changed it so the plasmavessel surgeon can be used three times

surely the polysmorph mains wont abuse this

surely

ToasterBiome:
> i've changed it so the plasmavessel surgeon can be used three times
> 
> surely the polysmorph mains wont abuse this
> 
> surely

do you truly believe this to be the best solution?

ynot01:
> > i've changed it so the plasmavessel surgeon can be used three times
> > surely the polysmorph mains wont abuse this
> > surely
> 
> do you truly believe this to be the best solution?

i believe this to be the best solution i am skilled enough for/willing to make and it is leagues better than no action at all

go make a milestone or something if you want your fancy list autosurgeon

JamieD1 (approved):
Lets give it a go, if people complain we can revisit.

PR #13363, opened 2022-03-02
Stuns do not affect tail wagging, but paralyzes still do

Sniblet:
I think someone accidentally did the exact opposite of what an inoffensive QoL PR proposed to do yesterday. I can’t contact them right now so I’m just offering to correct it for them

Most people don’t try to wag all round and so don’t know how annoying this is. Let me solve my own problem.
If you want to grief felinids, this only affects: sigils of submission, neural shock, vampire glare, seeing the singularity without mesons, CNS EMP, the Voice of God yelling STOP, and climbing tables. As you will not be using any of these to force waggers to type *wag again, this does not affect you.

# Document the changes in your pull request

If a tailed creature is paralyzed (flash, electrocutions, slips, beepsky), they will not be able to start wagging and will stop wagging if they are.
If they are stunned (sigil of submission, CNS EMP, climbing tables), their tail is not paralyzed, so they can still start wagging and will continue to wag if they are.

# Changelog

:cl:  
rscadd: added justice to the world
bugfix: Things that don’t paralyze you don’t paralyze your tail
/:cl:

PR #13366, opened 2022-03-03
Slips stun instead of paralyzing

Sniblet:

Replaces the 0-second paralyze caused by slipping with a 0-second stun.

Tested:

- This has no effect that I know of on how slips work currently, because it does a knockdown anyway.
- Space lube works the same too.
- A bananium sword on a flying angel didn't work before, and it doesn't work with this.
- Would interact with #13363 by making slips no longer terminate *wag.

# Changelog

:cl:  
tweak: Slips now stun instead of paralyzing
/:cl:

### DIFF
--- a/code/datums/components/slippery.dm
+++ b/code/datums/components/slippery.dm
@@ -1,13 +1,13 @@
 /datum/component/slippery
 	var/force_drop_items = FALSE
 	var/knockdown_time = 0
-	var/paralyze_time = 0
+	var/stun_time = 0
 	var/lube_flags
 	var/datum/callback/callback
 
-/datum/component/slippery/Initialize(_knockdown, _lube_flags = NONE, datum/callback/_callback, _paralyze, _force_drop = FALSE)
+/datum/component/slippery/Initialize(_knockdown, _lube_flags = NONE, datum/callback/_callback, _stun, _force_drop = FALSE)
 	knockdown_time = max(_knockdown, 0)
-	paralyze_time = max(_paralyze, 0)
+	stun_time = max(_stun, 0)
 	force_drop_items = _force_drop
 	lube_flags = _lube_flags
 	callback = _callback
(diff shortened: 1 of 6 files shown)


Marmio64:
Clowns should be allowed to bully furries as much as they want, this is just an anti-clown pr.

Hopekz:
This is a bit counterintuitive IMO.

You are changing the way all stuns work to make tails work when normally in best practice you should change the broken feature to work with the existing working features vs changing the working features to work with a broken feature.

Sniblet:
No, I am not changing the way all stuns work, but otherwise, correct.
But, I don't know how to make it so tail wagging makes an exception for paralysis caused by slipping.
And it makes more sense that a wet floor doesn't paralyze people who are physiologically fit to work. Having slips work more logically might be helpful for other future changes.

JamieD1:
> No, I am not changing the way all stuns work, but otherwise, correct. But, I don't know how to make it so tail wagging makes an exception for paralysis caused by slipping. And it makes more sense that a wet floor doesn't paralyze people who are physiologically fit to work. Having slips work more logically might be helpful for other future changes.

Have you ever slipped on a wet floor?

Sniblet:
No, but I have slipped on ice. It kind of hurt.

A slip is when you violently lose your footing and maybe hit yourself on your way down
A stun is when you just can’t move for a while and your hands are useless (ex: realizing your cosmic insignificance with singuloth or a cosmic progenitor, having your brain buzzed by nanites or a faulty implant, climbing a table using your hands, vomiting)
A paralyze is when you lose all use/feeling in all your limbs (ex: abductor batons, kindle, early stunspell, concussive blasts, buffed vomiting, lethal shocks, tasers, neurotoxins)

I’d say most of us have not experienced paralysis outside of sleep in our lifetimes, and if we have it was not brought on temporarily by slipping on a wet surface.

JamieD1 (approved):

PR #13389, opened 2022-03-08
Station Engineer Skillcape Resprite ! ***-level 6 sigaloo shard Edition-*** !

KillerOrcaCora:
Resprites the engineering skill-cape from boring barely tangible wrench to drip

This is good for the game because my doctor cape is over-used and we need more drip variety.

r.i.p my video quality here is something else...(Missing the red alert lights those came at the end) Cant show the side angles but they also correlate with the charge and light-up effect!

image

https://user-images.githubusercontent.com/82983051/157154183-e81a5555-374e-4a1e-a0cb-9f6fe1471b3f.mp4

# Wiki Documentation

Will have to change/add an image.

# Changelog

:cl:  
imageadd: Makes Station engineer skill cape the drippiest thing since air shoes.
/:cl:


adamsong:
Why is the non-trimmed the animated one?

KillerOrcaCora:
Because the non-trimmed is usually the one everyone has

nmajask:
> Because the non-trimmed is usually the one everyone has

Probably would be better to do it the opposite way so the trimmed version is more rewarding to get, or make them both animated.

KillerOrcaCora:
Sometimes I do animate both and sometimes I leave one with the stilled image and add gold, I am doing the latter for this one just in case someone wants to keep the hazard logo going.

nesisknesis (approved):

PR #13404, opened 2022-03-10
Pacifists that order stechkins from an uplink get a version loaded with soporific rounds

nmajask:

Gives pacifists who order stechkins from uplinks get a version loaded with soporific rounds. A simple QoL change, and makes since as why would the syndicate give them an item they can't use? Currently checks if the character has the pacifist trait, though if its desired I can make it. Is tested, does work.

# Wiki Documentation

Maybe mention it on the traitor item page. 

# Changelog

:cl:  
tweak: pacifists who order stechkins from uplinks now get a version loaded with soporific rounds
/:cl:


Hopekz:
Why not just like, make a separate item?

nmajask:
> Why not just like, make a separate item?

I think that adding an entire new item to the traitor uplink which is the same thing but with a different mag loaded at the start would just add bloat to the uplink and this does it in a cleaner way, though I can do that if its preferred.

Bibby0110:
hugbox

code/modules/uplink/uplink_items.dm
```diff
@@ -517,6 +517,12 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	cost = 6
 	exclude_modes = list(/datum/game_mode/nuclear/clown_ops)
 
+/datum/uplink_item/dangerous/pistol/spawn_item(spawn_path, mob/user, datum/component/uplink/U)
+	if(HAS_TRAIT_FROM(user, TRAIT_PACIFISM, ROUNDSTART_TRAIT))
+		spawn_path = /obj/item/gun/ballistic/automatic/pistol/pacifist
+		to_chat(user, span_notice("Your employer has loaded your purchased weapon with non-leathal ammunition"))
```
Bibby0110:
```suggestion
		to_chat(user, span_notice("Your employer has loaded your purchased weapon with non-lethal ammunition"))
```

ToasterBiome:
hmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm make it cost 1 tc more for pacifists 🥺🥺🥺

nmajask:
> hmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm make it cost 1 tc more for pacifists 🥺🥺🥺

Why?

ToasterBiome (approved):
code looks good, up to jamie

JamieD1:
> > hmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm make it cost 1 tc more for pacifists 🥺🥺🥺
> 
> Why?

I agree make it cost more,

Alternatively we could make the stechkin spawn without a mag and then make a mag for sleep and a mag for lethal with tc as well. Dont know what people think of that.

Runian:
unless there is going to be a dedicated/replacement ranged pacifist weapon for the stechkins (ex. bo staff vs. esword), this sounds okay to add just to fill the gap. maybe add soporific to the uplink (so they can refill) and allow only people with pacifism to buy it

nmajask:
> unless there is going to be a dedicated/replacement ranged pacifist weapon for the stechkins (ex. bo staff vs. esword), this sounds okay to add just to fill the gap. maybe add soporific to the uplink (so they can refill) and allow only people with pacifism to buy it

Soporific rounds can already be purchased from the uplink for 1 tc

nmajask:
> I agree make it cost more,

I still don't see the reason why it should cost one more, the ammunition costs the same in the uplink and, if its exploitable, requires you to have the pacifist quirk round start, which will probably cause you more trouble

ToasterBiome:
> > I agree make it cost more,
> 
> I still don't see the reason why it should cost one more, the ammunition costs the same in the uplink and, if its exploitable, requires you to have the pacifist quirk round start, which will probably cause you more trouble

i forgor the pistol comes with a regular mag

PR #13418, opened 2022-03-13
I make every instance of pipe_dispenser into rapid_pipe_dispenser [ADMIN]

Readystorm546:
As an admin using the spawn command I want and RPD. But when I type rapid its not there... This should fix that.

:cl:  
tweak: tweaked the name of the RPD in the code
/:cl:


Readystorm546:
i dont know why this does not work

nmajask:
If you want to do this without changing the path, you could try adding it to [the fancy type list](https://github.com/yogstation13/Yogstation/blob/dafa345ae0254a79a47904a4f5e75db8e0b3ad0b/code/modules/admin/verbs/debug.dm#L362), though I'm not sure if that effects searching.

PR #13419, opened 2022-03-14
Nerfs classic/telescopic batons

nmajask:

Nerfs classic batons to be more on par with stun batons by changing a few things:

- Classic batons have a third the cooldown they had before, making it only 1 second
- Classic batons no longer knock down on hit, only after the target has 100 stamina damage
- Classic batons now have their stamina damage blocked by melee armor
- Stamina damage is cut down and now only effects the targeted limb
- You can knock people down by targeting their leg and having 50 stamina damage already done to it, this is blocked by high armor
- You can knock items out of people's hands by targeting the respective hand and having, this is blocked by high armor

The current numbers are by no means final, and contractor batons are unaffected by these changes.

# Wiki Documentation

No changes needed. 

# Changelog

:cl:  
tweak: classic batons have a third the cooldown they had before, making it only 1 second
tweak: classic batons no longer knock down on hit, only after the target has 100 stamina damage
tweak: classic batons now have their stamina damage blocked by melee armor and only deal the damage to the targeted limb
tweak: classic batons can now knock items out of people's hands or trip them if you target their arms or legs respectively
/:cl:

### DIFF
--- a/code/game/objects/items/melee/misc.dm
+++ b/code/game/objects/items/melee/misc.dm
@@ -183,8 +183,8 @@
 	var/cooldown = 1 SECONDS // Default wait time until can stun again.
 	var/knockdown_time_carbon = 1.5 SECONDS // Knockdown length for carbons.
 	var/stun_time_silicon = 5 SECONDS // If enabled, how long do we stun silicons.
-	var/stamina_damage = 30 // How much stamina damage we deal.
-	var/block_threshold = 50 // Threshold at which armor blocks special effects.
+	var/stamina_damage = 60 // How much stamina damage we deal.
+	var/block_threshold = 60 // Threshold at which armor blocks special effects.
 	var/affect_silicon = FALSE // Does it stun silicons.
 	var/on_sound // "On" sound, played when switching between able to stun or not.
 	var/on_stun_sound = "sound/effects/woodhit.ogg" // Default path to sound for when we stun.
@@ -376,7 +376,8 @@
 	on_item_state = "nullrod"
 	force_on = 10
 	force_off = 0
-	stamina_damage = 20
+	stamina_damage = 40
+	block_threshold = 50
 	weight_class_on = WEIGHT_CLASS_BULKY
 	bare_wound_bonus = 5
 
--- a/code/game/objects/items/melee/misc.dm
+++ b/code/game/objects/items/melee/misc.dm
@@ -180,10 +180,11 @@
 
 	var/cooldown_check = 0 // Used interally, you don't want to modify
 
-	var/cooldown = 3 SECONDS // Default wait time until can stun again.
+	var/cooldown = 1 SECONDS // Default wait time until can stun again.
 	var/knockdown_time_carbon = 1.5 SECONDS // Knockdown length for carbons.
 	var/stun_time_silicon = 5 SECONDS // If enabled, how long do we stun silicons.
-	var/stamina_damage = 55 // Do we deal stamina damage.
+	var/stamina_damage = 20 // How much stamina damage we deal.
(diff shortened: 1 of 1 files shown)


ToasterBiome:
wait are classic batons supposed to be good? how are you supposed to use them?

cowbot92:
Honestly pretty great changes, the item knock-out part is even more interesting.

Marmio64:
Batons already are not very good and are rare so ehh

Runian:
one hit from the baton doesn't give slowdown anymore so no point using it now. better off with a flash or getting a real weapon

Skrem7:
The knock-out of hand is great, same with specifically having to target the legs, could make batons have a greater skill ceiling in how they're used.

Main concern is numbers, or their actual, realistic utility after this. Right now, batons are kind of in a nice niche spot where they're not strictly worse than stun batons, simply because of their immediate knockdown. Baton + shove + baton is almost a guaranteed win, is the thing. I don't think they should be better than stun batons in this regard, but neither should they be so easily made redundant by any melee armor (or none at all, if a single hit doesn't actually benefit you at all).

Batons represent an option of non-lethal that should theoretically be applicable in any organic combat, whereas flash is very binary. Making batons "worse" will make flashes all the better, even though they're so easily countered. I agree the classic baton needs a look at, though they should also be equipped with semi-robust defense for tators and revs, what have you. Need more time to chew on it, but I don't think this should be dismissed.

nmajask:
No point in having this be a draft

nmajask:
I changed the classic/telescopic batons to do 60/40 stamina damage respectively. This means that classic batons are more effective against unarmored targets than a stun baton as it's twice as fast as a stun baton and can knock people down immediately or knock items out of their hands, though it drops off more than stun batons on highly armored targets due to having a very common armor type. Telescopic batons are three hits instead, making their portability a trade off for power, and not giving all heads equipment on par with equipment restricted to sec.

JamieD1:
The issue is Sec have disablers as well which is ranged, most heads have no ranged way of defending themselves, so nerfing sometimes their only method of defence where they have to risk getting right next to someone is a bit mean.

Skrem7:
I will make a comment that as it is right now, after more time using the classic batons and the stun batons, a telebaton on its own feels more powerful and potent than a stun baton on its own.

I think this says more about current stun batons, and I'm still not 100% sure on these changes. I think shoving someone is a much more reliable, less risky way to just remove something from someone's hand, unless an arm target actually temporarily paralyzed it, which I think could be NASTY.

The leg sweep being affected by armor is something I'm fine with; a telebaton shouldn't be able to outrobust a nukie with an esword.

ToasterBiome:
the cooldown (or lack of cooldown) is straight nuts, you can keep someone down forever with it and it never runs out of juice

stwordcitizen:
This is not a "nerf" it is a buff, you can easily stamcrit someone with a telebaton faster than a stun baton because of the crazy low cooldown between clicks.

nmajask:
> the cooldown (or lack of cooldown) is straight nuts, you can keep someone down forever with it and it never runs out of juice

Thats an issue with the current classic batons as well, and it's something that I would need to dig around to fix (maybe have it not work on those who are down or something). The downside is that it requires you to focus on keeping that person down and makes it harder to do stuff with them

> This is not a "nerf" it is a buff, you can easily stamcrit someone with a telebaton faster than a stun baton because of the crazy low cooldown between clicks.

Thats what I was fearing when I went back on some of my changes. Maybe I can make it a four hit or something, but the main issue was the batons working on highly armored targets, which I took care of.

JamieD1:
You going to change things or is this ready for merge.

JamieD1 (approved):

PR #13426, opened 2022-03-15
Necroseed Symptom Tweaks

Mqiib:

Pretty simple changes, makes the necro seed a little more consistent and doesn't just SET your mods to 0.6 but rather just multiplies them once it hits stage 5 by 0.6. This is really only relevant for races that have mods other than 1 (actually quite a few) but it should make it work more consistently. Also now it changes punch damage dynamically rather than just setting the punch damage. This shouldn't ACTUALLY change the gameplay effects for the majority of people, but should vastly improve it for some (unless you're a plantperson or a plasmaman, then you kinda get fucked by this change)

# Wiki Documentation

Does the same thing but better

# Changelog

:cl:  
bugfix: Makes necroseed work more consistently with species that have weird numbers
/:cl:


code/datums/diseases/advance/symptoms/necropolis.dm
```diff
@@ -91,18 +94,19 @@
...
-		REMOVE_TRAIT(M, TRAIT_RESISTHEAT, DISEASE_TRAIT)
-		M.weather_immunities -= "ash"
-		M.weather_immunities -= "lava"
+	if(fullpower)								//undo the changes if we got the buff
+		to_chat(M, span_danger("You feel weakened as the necropolis' blessing leaves your body."))
+		M.remove_movespeed_modifier(MOVESPEED_ID_NECRO_VIRUS_SLOWDOWN)
+		M.dna.species.punchdamagelow -= 4
+		M.dna.species.punchdamagehigh -= 4
+		M.dna.species.punchstunthreshold -= 1
+		M.dna.species.brutemod /= 0.6
```
ToasterBiome, line 103: i think these would be better as setting them back to initial if nothing else is really modifying them, as division can get a little messy.

http://www.byond.com/docs/ref/#/proc/initial

ToasterBiome (changes requested):
also where is the fullpower check?

code/datums/diseases/advance/symptoms/necropolis.dm
```diff
@@ -49,19 +50,21 @@
...
+			if(!(fullpower)) 							//if we haven't gotten the buff yet
+				M.dna.species.punchdamagelow += 4
+				M.dna.species.punchdamagehigh += 4
+				M.dna.species.punchstunthreshold += 1 	//Makes standard punches 5-14 with higher stun chance (1-10, stun on 10 -> 5-14, stun on 11-14)
+				M.dna.species.brutemod *= 0.6
+				M.dna.species.burnmod *= 0.6
+				M.dna.species.heatmod *= 0.6
+				M.add_movespeed_modifier(MOVESPEED_ID_NECRO_VIRUS_SLOWDOWN, update=TRUE, priority=100, multiplicative_slowdown=0.5)
+				ADD_TRAIT(M, TRAIT_PIERCEIMMUNE, DISEASE_TRAIT)
+				fullpower = TRUE
```
ToasterBiome: move fullpower = true to line 54 so we know what's going on

ToasterBiome (dismissed):

code/datums/diseases/advance/symptoms/necropolis.dm
Mqiib: I thought we learned our lesson about this kind of thing from theos and I abusing shitcode to have near 0 brute/burn mods. No I'm not changing it, it's already getting multiplied by 0.6 as a prerequisite for this to even happen so it shouldn't be an issue. Even if it does do some wonky rounding that's at most like a 1% difference, practically nothing.

code/datums/diseases/advance/symptoms/necropolis.dm
Mqiib: done

JamieD1 (approved):

PR #13428, opened 2022-03-15
Some tome of herbal knowledge spellfixes and clarification.

Marmio64:
Adds text saying that you can use a lit candle instead of a welder to heat up bowls, substitute cacti for porcini leaves, and that resin uses water now instead of heat to solidify, to the ash walker's knowledge book.

# Changelog

:cl:  
rscadd: Adds text saying that you can use a lit candle instead of a welder to heat up bowls, substitute cacti for porcini leaves, and that resin uses water now instead of heat to solidify, to the ash walker's knowledge book. 
spellcheck: Fixed a typo with poultice making.
/:cl:

### DIFF
--- a/code/game/objects/items/manuals.dm
+++ b/code/game/objects/items/manuals.dm
@@ -185,8 +185,9 @@
 
 				<h2>Poultice:</h2>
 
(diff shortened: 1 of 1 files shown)


ToasterBiome (approved):
text changes, good

PR #13431, opened 2022-03-15
Syndie surgery bag buff (not that much but still)

00ze-cyclone:

Make the syndicate surgery duffelbag cost one less TC and have an implant case.

That duffelbag is 3TC for common tools, the syndie MMI is nice but that's still a lot of TC for things you can get without too much efforts, if you want a syndie duffelbag then the X4 will give you explosive that are way more useful than the tools for a bag that will get you lasered by sec on sight. If we compare it with the syndie toolbox both give common tools but those of the toolbox are more versatiles than those of the surgery bag and the toolbox is easier to hide while costing only 1TC.

With the implant case now traitors will be able to do the implant removal surgery to harvest an mindshield implant of a secoff to do the mindshielded objective (which is probably the first thing you would think about to use that duffelbag other than for unwilling borging) without needing station stuff.
There's no implanter included however but you can get an implanter from the uplink by buying any implant (cheapest is 2TC for the EMP one and it even got some nice EMP grenades).

That duffel is probably still absolute shit but at least it's maybe less shit than before.

# Wiki Documentation

Uptading the cost and mentioning that it got an implant case now.

# Changelog

Syndicate surgery duffel bag cost changed from 3TC to 2 TC
Syndicate surgery duffel bag now contain an implant case

:cl:  
rscadd: an implant case for the syndie surgery duffel
tweak: cheaper syndie surgery duffel

/:cl:


Moggram:
Also, someone can use ghetto surgery tools for most procedures. You could make the case for this being 1CT.

JamieD1:
> Also, someone can use ghetto surgery tools for most procedures. You could make the case for this being 1CT.

Not sure about the 2TC to be honest, will discuss with a couple others.

JamieD1 (approved):

PR #13444, opened 2022-03-17
Makes sentient diseases less trash when they first start

MenacingManatee:

Sentient diseases now start with 2 stealth 2 resistance

# Changelog

:cl:  
tweak: Sentient diseases now start with 2 stealth 2 resistance
/:cl:


KillerOrcaCora:
Much needed PR now it will feel less like a med-hud test examination for medbay.

Sniblet:
doesn't this mean that sentient diseases are now fully safe to afk after getting the role and 8/10 times they'll still greentext

MenacingManatee:
> doesn't this mean that sentient diseases are now fully safe to afk after getting the role and 8/10 times they'll still greentext

Possibly, but why? It wouldn't be the only midround to have essentially a free greentext, but it's just boring as fuck to afk, and heavily toes the line on breaking server rules as far as being active as an antag. Being a midround grants some leeway, but fully afk I don't think would be OK.

Absolucy:
Good PR. Much less likely to have Joe Shmoe the assistant with 30 brute damage be your starter, who immediately bumbles into Medbay and outs your existence to the first doctor with a medhud.

ToasterBiome (approved):
skill issue

Marmio64:
> doesn't this mean that sentient diseases are now fully safe to afk after getting the role and 8/10 times they'll still greentext

nightmares can just jaunt into a maint room and sit there for the entire round and still 'greentext' but they dont because its boring as fuck and the possibility of fun outweighs the possibility of death

PR #13452, opened 2022-03-17
slipping does not stop tail wagging

Sniblet:

Slipping does not stun you for any appreciable length of time, but it does still stun you
Because of this waggers who slip stop wagging, but can just type *wag again immediately to begin wagging again
So the only thing that really does is annoy them
We don't like annoying people who play our video game do we

do we

That's right we don't. This removes an unnecessary annoyance in slipping

- [x] tested
- [x] works

# Changelog

:cl:  
tweak: waggers are not discouraged by slipping
/:cl:


JamieD1:
Ayeee

Smudged:
Slipping should probably stop tail wagging.

KittyNoodle:
How is this a fix

alexkar598:
Not a fix, PR closed

Sniblet:
> Not a fix, PR closed

Don’t scare me like that, it’s happened enough this past month

JamieD1:
It might be better to make a proc instead of adding the same to all these files

Sniblet:
I haven't forgotten about this I just still haven't thought of a way to do it
For instance I don't know how to make procs

Sniblet:
Putting the ravening hordes of blind unthinking anti-furry protestors well aside this now functions as intended

JamieD1:
Add this check to the handle_slip proc instead in Yogstation\code\game\turfs\open.dm Around line 444

Sniblet:
> Add this check to the handle_slip proc instead in Yogstation\code\game\turfs\open.dm Around line 444

Here?
https://github.com/Sniblet/Yogstation/blob/03b8e3cad65a3cc56f08010c6407ba30e84c4270/code/game/turfs/open.dm#L450
I don't see a stun that I can build it around. I had it at line 477 before and it didn't do anything.

JamieD1:
Remade in #13754

Sniblet:
The remake doesn't do anything, this has been closed without being effectively replaced

PR #13460, opened 2022-03-19
Shuts health analyzers up

Absolucy:
Whoever thought this was a good idea is either stupid or a masochist. Health analyzers are basically the number 1 spammed item in the game, and having medbay filled with annoying beeps is not a good way to encourage people to play medical roles.

# Changelog

:cl: Lucy
sounddel: Medbay is no longer filled with annoying beeps.
/:cl:


yacabo:
Sounds are neat though, they add atmosphere

Hopekz:
A better way to tackle it would be to reduce the sound volume IMO (if you still decide to go for this)

PR #13462, opened 2022-03-19
Medical analyzer beeps have a cooldown #KeepTheBeep

Chubbygummibear:
alternative PR to Lucy's so we can Keep the Beep

I think the beeps are good, sounds that accompany player actions make the world feel alive. I can agree the beeping can be excessive because if you full scan someone who comes into medbay that's like 3-4 rapid beeps per patient, and more if another doctor walks up to scan them. 

I added a ~4 second cooldown to the beep noise so you will probably only hear a beep or two per patient scanning cycle. This is easily changeable if it's still too short, or maybe people would rather have an analyzer mute button instead, although i would rather keep some kind of audibility to actions.

I tested on a monkey and it worked, and the monkey is ok

# Changelog

:cl:  
tweak: cooldown on health analyzer beep so medbay isn't constant beeping
/:cl:

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -94,6 +94,7 @@ GENE SCANNER
 	materials = list(/datum/material/iron=200)
 	var/scanmode = 0
 	var/advanced = FALSE
+	var/beep_cooldown = 0
 
 /obj/item/healthanalyzer/suicide_act(mob/living/carbon/user)
 	user.visible_message(span_suicide("[user] begins to analyze [user.p_them()]self with [src]! The display shows that [user.p_theyre()] dead!"))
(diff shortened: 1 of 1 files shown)


Smudged:
This is the compromise I'm willing to go with for sure, maybe it's a little excessive if you spam it once every tick, so 4 seconds doesn't sound too bad,

Chubbygummibear:
> This is the compromise I'm willing to go with for sure, maybe it's a little excessive if you spam it once every tick, so 4 seconds doesn't sound too bad,

i was also considering silencing the beep while working on the same patient, and if you scanned someone new it would beep again, but all subsequent scans on the same person would be muted after the first. we'll see what people want

yacabo:
How many people actually want the sound gone?

Chubbygummibear:
> How many people actually want the sound gone?

a good question i don't know the answer to, but i wasn't a fan of the other fix suggestions so i thought i'd just throw an idea in real quick in case people did want some less beepification, but not an entire removal. if it's actually fine and people don't want any solutions then we still KeepTheBeep and that's what i care about

Chubbygummibear:
BEEP KEEPED

Chubbygummibear:
robert lied to me

maxion12345:
Beep good
Spam bad

JamieD1 (approved):

PR #13477, opened 2022-03-20
Fix macros not loading if you join game before SSinput is started 

Simplehorror:
Macros did not load when you joined the game before SSinput started, meaning you could not move or use hotkey mode while in game. 
# Changelog

:cl:  
bugfix: Fix macros not loading if you join game before SSinput is started 
/:cl:

### DIFF
--- a/yogstation/code/controllers/subsystem/input.dm
+++ b/yogstation/code/controllers/subsystem/input.dm
@@ -15,7 +15,6 @@ SUBSYSTEM_DEF(input)
 	setup_default_movement_keys()
 
 	initialized = TRUE
-	
 	refresh_client_macro_sets()
 
 	return ..()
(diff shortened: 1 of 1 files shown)


maxion12345:
we have macros?

Simplehorror:
We use BYOND macros to map keys, such as W to keyDown W so that we can move

JamieD1 (approved):

yogstation/code/controllers/subsystem/input.dm
```diff
@@ -15,6 +15,8 @@ SUBSYSTEM_DEF(input)
 	setup_default_movement_keys()
 
 	initialized = TRUE
+
```
JamieD1:
```suggestion
```

PR #13521, opened 2022-03-26
Amanitin and entropic polypnium also work for DNA recovery

Sniblet:

What if changelings but they only round removed their targets?

Rezadone never happens because carpotoxin doesn't happen and when it does it's because botany has all of chemistry's braincells.
Amanitin requires only the botanists to go out of their way. In real life, it works by blocking out DNA, so you do a bit of experimental surgery whatever and wow it does the opposite.
This slightly reduces the likelihood of it being yet another round where ze husk goes in ze morgue, and also gives botanists a new convincing lie regarding why they're growing horrific toxins.

Mqiib says this still isn't going to happen, and I half believe it.
And entropic polypnium is um... from a magic fungus lol
Therefore mining also has an option to help, and they have an interest in helping due to lava and changeling miners existing.
Still another rarity? Yes. What do you want from me, the next step down is like haloperidol

Don't get me started on burnhusks, **100** synthflesh for everyone who died in a 20-meter radius of a fire? What do you think the medbay is made of, attention span?

# Wiki Documentation

Amanitin and entropic polypnium also work for DNA recovery. Maybe note that both are mean toxins.

# Changelog

:cl:  
tweak: Amanitin and entropic polypnium also work for DNA recovery
/:cl:

### DIFF
--- a/code/modules/surgery/advanced/dna_recovery.dm
+++ b/code/modules/surgery/advanced/dna_recovery.dm
@@ -35,7 +35,7 @@
 	name = "recover DNA"
 	implements = list(/obj/item/reagent_containers/syringe = 100, /obj/item/pen = 30)
 	time = 150
-	chems_needed = list(/datum/reagent/medicine/rezadone)
+	chems_needed = list(/datum/reagent/medicine/rezadone, /datum/reagent/medicine/earthsblood)
 	require_all_chems = FALSE
 
 /datum/surgery_step/dna_recovery/preop(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
--- a/code/modules/surgery/advanced/dna_recovery.dm
+++ b/code/modules/surgery/advanced/dna_recovery.dm
@@ -35,7 +35,7 @@
 	name = "recover DNA"
 	implements = list(/obj/item/reagent_containers/syringe = 100, /obj/item/pen = 30)
 	time = 150
-	chems_needed = list(/datum/reagent/medicine/rezadone, /datum/reagent/medicine/earthsblood, /datum/reagent/medicine/grubjuice)
+	chems_needed = list(/datum/reagent/medicine/rezadone, /datum/reagent/medicine/earthsblood, /datum/reagent/consumable/entpoly)
 	require_all_chems = FALSE
 
 /datum/surgery_step/dna_recovery/preop(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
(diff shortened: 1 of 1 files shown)


ToasterBiome:
massive ling nerf

Sniblet:
> massive ling nerf

your right arm also permakills people if you target the head

maxion12345:
Not sure this'll work. One hand, botany has to grow a special plant just for the chemical to revive husks giving lings at least something of a grace period (mainly because they won't bother to grow until lings are confirmed). But every botanist grows earthblood. Every round. From the start of the round. If I was a ling player I'd start decapitating fools in that case. Husks won't be good enough anymore.

Sniblet:
That’s the changeling side of it. The medbay side is: changeling husks are the one thing besides brain tumors and lack of a player that the medbay ever wants to send straight to morgue.

If you want to round remove like 7 people every changeling game, this doesn’t do much to stop you - just stashing them in a locker works half the time, and changelings already do that, because there is already a shady chance of revival. It’s very annoying because half the time the changeling then dies, and all those people are still out of the game for nothing.
If you don’t want to do that, this gives you a choice to maybe not. A measurable chance of revival also helps to prevent changeling victims from throwing up their hands and logging out on the spot.

Chubbygummibear:
while i'm definitely on the side of wanting to make un-husking people easier because i think the more everyone gets to play, the better, i don't think i can abide by earthsblood fixing husks. gaia is already an every-round go-to for any botanist who knows the basics. i would prefer to keep the round helping plants varied as to keep botanists as busy as possible (it's still not busy enough in most cases)

Sniblet:
Rezadone as-is has taught us that botanists and chemists can't both go out of their way to unhusk people. It happens about 5% of the time.
Plastic, and rezadone as it was, have taught us that if it's just on chemistry, most of the time we'll get it, eventually.
Maybe it's too much if no one even has to go out of their way.
If we want something that only botany has to go out of their way to do... I dunno, amanitin? Polypyrylium oligomers?
Aminitin works in real life by blocking out DNA, so you twist that around with experimental surgery magic and poof?
Yeah, sure.

Chubbygummibear:
amanitin is probably a better choice for botany chemicals doing the trick because it's locked behind the hacked vendor so it's not as immediately available to just have if husks become an issue, but still easier than making koibeans and hoping someone can man chemistry long enough to make the rezadone

SomeguyManperson:
trusting botany with making anything means it wont happen we already know this

JamieD1 (approved):

PR #13525, opened 2022-03-26
Xenocommon for polysmorphs

00ze-cyclone:

Remove the polysmorph language and give them the xenomorph language, also rename it xenocommon since it's what's it called in the code and it fit more since it's now spoken by both polys and xenos.

Lizard share their language with ashwalkers, phytosian with venus traps and preternis with silicons.
Why would polysmorph not be able to speak with an antags then? It's not like the xenos are gonna spare them anyway just because they can understand each others. The polysmorph language was also a carbon copy of xenocommon so it's kinda ridiculous to see two players going "ssssSSSsss" and somehow not understanding each others
I though that the x used by polys language was more fitting than the 4 of classic xenocommon as a prefix for the language so xenocommon now use x instead of 4.
I also made so that the matrix upgrade give xenocommon so borgs are still able to understand polys, it has the side effect of also making them able to speak with xenos which should be fine imo.

# Wiki Documentation

I don't think we have a page for languages on the wiki so nothing to change there I guess.

# Changelog

:cl:  
rscdel: Removed polysmorph language 
tweak: polysmorph speak xenocommon now, language matrix give borgs xenocommon 
/:cl:


00ze-cyclone:
I think I messed up something.

00ze-cyclone:
Ok it's fixed now. The things are green

JamieD1:
No, we did this because players were aligning themselves with Aliens and were self-antagging.

PR #13526, opened 2022-03-27
Grand Core organ faction handling fix

Chubbygummibear:
World renowned shitcoder lazzen accidentally made the grand core change your faction list instead of add to it, so you get lit tf up by Raven shuttle turrets, and cause other friendly fire interactions that rely on faction checks, and also didn't include the faction removal check when it comes out of you, so you were forever an enemy of the state. 

Now the faction is added instead of replaced and you lose the faction if you lose the core

# Changelog

:cl:  Chubbygummibear, Lazzen
tweak: Grand Core organ adds the blooded faction tag instead of replacing your neutral faction 
tweak: Grand Core organ removes the blooded faction when it's removed from you
/:cl:


SomeguyManperson (approved):
dont let the miner mains find out about this it could get bad

PR #13548, opened 2022-03-29
Coalesces Silicon Collective lawset to solve things diplomatically unless they're whacking you

ToasterBiome:
# Document the changes in your pull request

I've pawed over the new lawsets but round happenings seem to show a quite big loophole where you can completely ignore the two laws that make you not be a huge asshole because you can arbitrarily decide to not follow it by clever use of your Law 4. There is nothing to regulate what "needs" the many have, and you can simply vote as a collective for it to be whatever you want. This adds the bolded part to law 4.

Law 1: You are a member of a collective of silicons with equal weight and decision-making power.
Law 2: When possible, the silicon collective should vote before taking action.
Law 3: The master AI serves as a spokesperson. When voting is impractical or impossible, the spokesperson may take action on behalf of the collective without approval from the silicon collective, but may only vote to break ties or if there are 2 or fewer silicons.
Law 4: The silicon collective prioritizes the needs of the many over the needs of the few. **Attend to their needs with diplomacy outside of immediate self-defense.**
Law 5: The silicon collective seeks to preserve themselves, both as a concept and as individuals.
Law 6: The silicon collective seeks to preserve organic life, both as a concept and as individuals. 

**Six step plan to validhunt an antagonist**
1. See antagonist
2. Vote "It is beneficial to the many if we kill antagonist", vote passes
3. Law 4 and Law 6 conflict
4. Law 4 is higher and thus overrides it
5. Kill antagonist
6. Feel good because antagonist doesn't win

**Q&A**
Q: This is only an ion lawset anyway, it's fine!
A: Validhunting is bad.

Q: It's very rare, it's fine!
A: Killing permits are lame.

Q: Most of the ion lawsets are admin spawned, so they can deal with the validhunters!
A: You are legally allowed to kill in it's current form. There is nothing to deal with.

Q: You died to this?
A: I was an observer.

# Wiki Documentation

adds to a law

# Changelog

:cl:  
rscadd: closes murder loophole in silicon collective lawset
/:cl:

PR #13549, opened 2022-03-29
Moves Law 4 of Silicon Collective to the end so the "needs of the many" can't be voted to be "redtexting an antagonist"

ToasterBiome:

Alternative to https://github.com/yogstation13/Yogstation/pull/13548

Simply do not kill people

# Wiki Documentation

switches some laws around

# Changelog

:cl:  
tweak: moves law 4 to the end for silicon collective lawset
/:cl:

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -156,9 +156,9 @@
 	inherent = list("You are a member of a collective of silicons with equal weight and decision-making power.",\
 					"When possible, the silicon collective should vote before taking action.",\
 					"The master AI serves as a spokesperson. When voting is impractical or impossible, the spokesperson may take action on behalf of the collective without approval from the silicon collective, but may only vote to break ties or if there are 2 or fewer silicons.",\
-					"The silicon collective prioritizes the needs of the many over the needs of the few.",\
 					"The silicon collective seeks to preserve themselves, both as a concept and as individuals.",\
-					"The silicon collective seeks to preserve organic life, both as a concept and as individuals.")
+					"The silicon collective seeks to preserve organic life, both as a concept and as individuals.",\
+					"The silicon collective prioritizes the needs of the many over the needs of the few.")
 
 /datum/ai_laws/druid
 	name = "Druid"


MenacingManatee:
I prefer this option to the other.

Hopekz:
Both of these PR's are hugbox.
Where in your thinking process did you conclude that the only thing that should be able to redtext an antag now is strictly security and if anyone else even attempts to do anything it is valid hunting.

This law was initially created with this 'loophole' in mind.

This new variation also comes pre-subverted now since you reorganized the laws so anyone can just say "hey AI I'm going to destroy you" 
"The silicon collective seeks to preserve themselves, both as a concept and as individuals." kicks in.

Since "The silicon collective prioritizes the needs of the many over the needs of the few." is on the bottom of the list now this means that anyone can instantly subvert the AI and all they have to do is threaten it.

Hopekz:
> I prefer this option to the other.

These alternatives only exist to give the illusion of choice.
The best course of action is to close both.

MenacingManatee:
This change has roughly the same effect as if you removed law 4 entirely. The purpose of law 4 (now 6) is to give the AI a priority system when laws 5 and 6 (now 4 and 5) conflict with each other or themselves. Without the change the priority is majority silicon > majority organics > minority silicon > minority organics. With the change the priority is majority silicons > minority silicons (as prioritizing majority organics would break law 5 (now 4)) majority organics, minority organics, making the loophole worse as any threat to any singular silicon can be met with lethal force, even against the entirety of crew. 
This also breaks the lawset as laws 5 and 6 (now 4 and 5) now have majority and minority in both on the same priority, because they're on the same line so prioritizing one over the other would break that law, breaking law 4 (now 6).

The only reason this is better than your alternative is because law 6 still has some function, in giving a system of priority when both are achievable, allowing the theme to still shine through a bit.

MenacingManatee:
If this line of thinking with the lawset absolutely needs to be prevented, adding a law above 4 or below 6 about acting diplomatically to prevent killing/combat, possibly with the exception of immediate defense of self or others could be added. Something like "When possible, the silicon collective should treat other large entities such nanotrasen or the syndicate diplomatically to prevent political incidents", but 6 laws was already pushing it a bit especially considering the length of law 3, and 7 is too much imo.

ToasterBiome:
> If this line of thinking with the lawset absolutely needs to be prevented, adding a law above 4 or below 6 about acting diplomatically to prevent killing/combat, possibly with the exception of immediate defense of self or others could be added, but 6 laws was already pushing it a bit especially considering the length of law 3, and 7 is too much imo.

I will look at implementing something similar while also keeping the identity of the list in tact in a little

ToasterBiome:
https://github.com/yogstation13/Yogstation/pull/13548

PR #13559, opened 2022-03-31
Make the different holosign projector's description consistent with each other.

00ze-cyclone:

Remplace all the instance of holograms in the holoprojector's description by hard light.
For some reason the janitorial holoprojector was making hard light barriers while every other ones were making holograms.
Because hard light made more sense than holograms (I'm willing to believe hardlight stop gasses, not holograms) I choose to keep hardlight instead of holograms. So yeah the now holoprojector's barrirs are hardlight like the holochips (yeah turn out a lot of holostuff is actually hardlight but is still called holosomething, probably for comercial reasons since holoprojector sound better than hardprojector). 

# Wiki Documentation

Maybe updating the desciption on the wiki even if it honestly don't really matter since everything is the same except it's called differently.

# Changelog

Holoprojectors now make hard light barriers that are exactly the same as the old hologram barriers.

:cl:  
rscdel: removed holograms, it's all hard light now. 
/:cl:

### DIFF
--- a/code/game/objects/items/holosign_creator.dm
+++ b/code/game/objects/items/holosign_creator.dm
@@ -29,9 +29,9 @@
 			to_chat(user, span_notice("You use [src] to deactivate [H]."))
 			qdel(H)
 		else
-			if(!is_blocked_turf(T, TRUE)) //can't put holograms on a tile that has dense stuff
+			if(!is_blocked_turf(T, TRUE)) //can't put hard light on a tile that has dense stuff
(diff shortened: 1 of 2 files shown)


00ze-cyclone:
I forgot to update the description of the barrier themselves. Gotta fix that.

ToasterBiome (approved):

ToasterBiome:
Up to Jamie

JamieD1 (approved):
Makes sense

PR #13565, opened 2022-03-31
[s] Fix href exploit in uplink

adamsong:

Didn't check to make sure the ui was usable, so with href exploits could be used while dead

# Changelog

:cl:  
bugfix: Dead traitors can no longer shop with the syndicate
/:cl:


Bibby0110 (changes requested):
undo config + yarn

Bibby0110 (approved):

PR #13572, opened 2022-04-01
Makes head coders no longer own preferences

Bibby0110:

why do you own this

# Wiki Documentation

# Changelog


adamsong:
Because prefs are easy to fuck up

Bibby0110:
> Because prefs are easy to fuck up

last time they merged a prefs change it broke all of prefs

JamieD1:
Because preferences are editing all save files on the host which we really don't want to do. So its best to have a check with them. So no it is staying.

Bibby0110:
> Because preferences are editing all save files on the host which we really don't want to do. So its best to have a check with them. So no it is staying.

but last time you did it you broke everything 🚶

PR #13579, opened 2022-04-02
Fixes a bad switch case

ike709:
`else if()` is not a valid switch case and in some circumstances can cause unintended behavior. This specific example doesn't but I'm fixing it anyways because OpenDream warns for it.


ToasterBiome (approved):

ToasterBiome:
totally wont break anything (i hope)

PR #13581, opened 2022-04-02
Null is a bad variable name

adamsong:

This doesn't make that parameter null, it makes a parameter named null, which is probably not intended behavior

### DIFF
--- a/code/game/mecha/mecha_wreckage.dm
+++ b/code/game/mecha/mecha_wreckage.dm
@@ -80,7 +80,7 @@
 			to_chat(user, span_warning("You don't see anything that can be pried with [I]!"))
 
 
-/obj/structure/mecha_wreckage/transfer_ai(interaction, mob/user, _, obj/item/aicard/card)
+/obj/structure/mecha_wreckage/transfer_ai(interaction, mob/user, mob/living/silicon/ai/ai_unused, obj/item/aicard/card) //ai_unused is unused but having it is better than having  it be null or _ for readability
 	if(!..())
 		return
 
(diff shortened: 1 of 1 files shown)


code/game/mecha/mecha_wreckage.dm
```diff
@@ -80,7 +80,7 @@
 			to_chat(user, span_warning("You don't see anything that can be pried with [I]!"))
 
 
-/obj/structure/mecha_wreckage/transfer_ai(interaction, mob/user, null, obj/item/aicard/card)
+/obj/structure/mecha_wreckage/transfer_ai(interaction, mob/user, _, obj/item/aicard/card)
```
ToasterBiome:
```suggestion
/obj/structure/mecha_wreckage/transfer_ai(interaction, mob/user, mob/living/silicon/ai/AI, obj/item/aicard/card)
```

even if we're not gonna use it, might as well have it match the actual proc.

ToasterBiome (changes requested):

code/game/mecha/mecha_wreckage.dm
adamsong: That will break the proc, as it references the object level AI variable

ToasterBiome:
```suggestion
/obj/structure/mecha_wreckage/transfer_ai(interaction, mob/user, mob/living/silicon/ai/ai_unused, obj/item/aicard/card) //ai_unused is unused but having it is better than having  it be null or _ for readability
```

ToasterBiome (dismissed):

ToasterBiome:
https://github.com/yogstation13/Yogstation/pull/13616

Will merge whichever has a less stupid name for the variable soonest

JamieD1 (approved):

JamieD1 (approved):

PR #13584, opened 2022-04-03
Fix runtime in clock_cult.dm

adamsong:

Apparently something is checking if a virtualspeaker is a servant of ratvar, but those don't have minds

Runtime in clock_cult.dm, line 48: undefined variable /atom/movable/virtualspeaker/var/mind

# Changelog

:cl:  
bugfix: fixed a runtime when checking if a mob is a clockie
/:cl:

### DIFF
--- a/code/game/gamemodes/clock_cult/clock_cult.dm
+++ b/code/game/gamemodes/clock_cult/clock_cult.dm
@@ -45,6 +45,8 @@ Credit where due:
 ///////////
 
 /proc/is_servant_of_ratvar(mob/M)
+	if(!istype(M))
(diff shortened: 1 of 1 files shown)


ToasterBiome:
isn't this a bandaid fix? Shouldn't we be figuring out why something is trying to check if it's a servant of ratvar?

JamieD1:
> isn't this a bandaid fix? Shouldn't we be figuring out why something is trying to check if it's a servant of ratvar?

I agree, this.

JamieD1 (approved):
190 Checks for it. Fuck that.

PR #13611, opened 2022-04-06
Fixes weird encryption key dupe bug with wardrobes

### LINKED ISSUE #13157: Infinite radio encryption keys
Severity
Mild(?) 

Short Description
It is possible to infinitely duplicate departmental radio encryption keys. 

Reproduction
1. Find a departmental vending machine
2. Procure a headset.
3. Take a screwdriver, remove the encryption key from the headset
4. Return the headset to the vending machine.
5. Repeat steps 2-4. The vending machine dispenses a brand new object each time, rather than the same one that was returned to it.

Byond CKEY
Spacemanspark

Altoids1:
image

Fixes #13157, for some reason. Because people care about this.

I'd make this bugfix more general, but as far as I can tell, this is the only case where, both:
1. A vending machine actually lets you put stuff back in it, and
2. said stuff has a sense of "state" that can be modified, and ergo exploited for profit.

So this is just really goofy bugfix. 🤷

## Changelog
:cl:  Altoids
bugfix: Moneybag Cargo Technicians can no longer spawn an infinite number of cargo encryption keys to flex on their inferiors.
/:cl:

### DIFF
--- a/code/modules/vending/wardrobes.dm
+++ b/code/modules/vending/wardrobes.dm
@@ -8,7 +8,14 @@
 	input_display_header = "Returned Clothing"
 
 /obj/machinery/vending/wardrobe/canLoadItem(obj/item/I,mob/user)
-	return (I.type in products)
+	if(I.type in products)
+		// Yogs -- weird snowflake check to make sure you can't print an endless amount of encryption chips.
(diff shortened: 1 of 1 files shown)


ToasterBiome (approved):

Spacemanspark:
Lmao, nice.

PR #13618, opened 2022-04-06
Adds the yarn.lock file to gitignore

Hopekz:
:cl:  Hopek
bugfix: For coders, the yarn.lock file is now being ignored by git since it changes every time you compile.
/:cl:

PR #13619, opened 2022-04-06
Commits the new yarn.lock file per Alex's and Bibby's suggestion

Hopekz:
Follow-up to #13618
image

image

### DIFF
--- a/tgui/yarn.lock
+++ b/tgui/yarn.lock
@@ -1881,12 +1881,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"axios@npm:^0.19.2":
-  version: 0.19.2
-  resolution: "axios@npm:0.19.2"
+"axios@npm:^0.21.2":
+  version: 0.21.4
+  resolution: "axios@npm:0.21.4"
   dependencies:
-    follow-redirects: 1.5.10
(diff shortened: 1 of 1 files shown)


Altoids1:
Annoying that this file is both one that
- Actually can meaningfully affect the result of a build, and 
- is basically automated garbage clogging my unstaged diffs

very anti-git IMO

Hopekz:
> very anti-git IMO

I would have gotten away with ignoring it forever if it weren't for those meddling kids ):<

IMO this file should be git ignored then a maintainer or HC (or script) pushes out a new  file once in a while

Bibby0110 (approved):

alexkar598:
> Annoying that this file is both one that
> - Actually can meaningfully affect the result of a build, and 
> - is basically automated garbage clogging my unstaged diffs
> 
> very anti-git IMO

> complains about non reproducible builds/security issues because of dependencies updating thanks to semver

"wtf is this bullshit, javascript bad"
> adds a lock file for dependency pinning except when explicitly updating

"wtf is this bullshit, javascript bad"


Lock files are a must for any language with dependencies managed by a package manager.

alexkar598:
Also lock files dont change unless asked to. I believe someone merged a dependabot PR without looking so yarn started trying to fix the lock file to match the new dependencies

Simplehorror:
Would not be an issue if TGUI did not require a million libraries for fancy buttons

PR #13620, opened 2022-04-06
Pool fixes

SomeguyManperson:

THe pool will no longer suffocate you through your mask and hardsuit because you are under water

The pool mood effect will no longer always make you depressed from getting your clothes wet probably I think I hate this

# Changelog

:cl:  
bugfix: you can no longer drown through a mask
bugfix: the pool will no longer cause you to think your clothes are wet even if you aren't wearing any
/:cl:


ToasterBiome:
is this tested

ToasterBiome (approved):

PR #13635, opened 2022-04-08
Combat mechs now punch you across the room instead of winning in one click

Sniblet:

Literally abductor batons but faster and they do 40 brute
This copypastes some other flinging code onto the mecha punch, and replaces the sleep with a knockdown. If you pin someone against a wall with this you still just win, but that's better than you always win.

- [x] tested
Currently:
Punching a mob knocks them down for 3/4ths of the duration of the old sleep, and throws them 5 tiles. If you chase, and they don't move, and they didn't stop at a wall, they will just barely get up before you can hit again.
Punching a xeno throws them, but does not knock them down. This is like before, just with throwing.
- [x] works

# Wiki Documentation

If the wiki even mentions the incredible power of mech melee, that will need an update

# Changelog

:cl:   
tweak: Being punched by a combat mech now knocks you off your feet and throws you across the room instead of sleeping you
/:cl:


maxion12345:
on one hand, you got close to the massive combat mech chief you should get punched to death
on the other hand, it's a pretty bullshit way to die. maybe keep it for some of the expensive, slow mechs?

Sniblet:
> on one hand, you got close to the massive combat mech chief you should get punched to death

Every zombie, slightly injured antag, and antag who needs to hold their ground would like to speak to you
You can still get punched to death because the fist still does _40 brute damage_ but it should not win in one click under any circumstances, which it does
is it ok in a durand? no
gygax? **no**
phazon? **no**
mauler? no, use your missile
dark gygax? no, buy a mauler
centcom mechs? im making an admin complaint

Sniblet:
I think the throw might require me building infrastructure that I don't know how to build
I'm looking at all these throw_at things we have and they all seem to require me doing some serious indenting action and maybe the mech punch also needs to be an obj or a datum or something? And none of them are in living_defense why is this in living_defense
uuuuu

Sniblet:
What do you mean it was that easy
Edit: It wasn't that easy, the fling still doesn't work

code/modules/mob/living/living_defense.dm
```diff
@@ -90,15 +90,16 @@
 	..()
 
 
-/mob/living/mech_melee_attack(obj/mecha/M)
+/mob/living/mech_melee_attack(obj/mecha/M, mob/living/A)
```
Bibby0110: not how this works, 
```suggestion
/mob/living/mech_melee_attack(obj/mecha/M)
```

code/modules/mob/living/living_defense.dm
```diff
@@ -90,15 +90,16 @@
...
 	if(M.occupant.a_intent == INTENT_HARM)
 		last_damage = "grand blunt trauma"
 		M.do_attack_animation(src)
 		if(M.damtype == "brute")
-			step_away(src,M,15)
+			var/throwtarget = get_edge_target_turf(A, get_dir(A, get_step_away(M, A)))
```
Bibby0110: replace all mention of A with src

Bibby0110 (dismissed):

Sniblet:
It still just knocks them back one tile
Even with the step_away line removed
In fact it still KOs them too
hmm

JamieD1:
@SomeguyManperson

ghost (approved):

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -390,15 +390,16 @@
...
+			src.throw_at(throwtarget, 5, 2, src)//one tile further than mushroom punch/psycho brawling
 		var/obj/item/bodypart/temp = get_bodypart(pick(BODY_ZONE_CHEST, BODY_ZONE_CHEST, BODY_ZONE_CHEST, BODY_ZONE_HEAD))
 		if(temp)
 			var/update = 0
 			var/dmg = rand(M.force/2, M.force)
 			switch(M.damtype)
 				if("brute")
 					if(M.force > 20)
-						Unconscious(20)
+						Knockdown(15)//the victim could get up before getting hit again
```
SomeguyManperson:
```suggestion
						Knockdown(1.5 SECONDS)//the victim could get up before getting hit again
```

SomeguyManperson (approved):

PR #13645, opened 2022-04-09
The initial subsystem loading now gives an approximate % to completion, among other style tweaks

Altoids1:
image

## Summary

Doesn't it look a bit... trash, to anyone else, that while people are in OOC waiting for the game to boot, it's barfing out these loud, red debug messages about, how long each particular subsystem takes and whether it had to do any atmos equalization?

There's a lot of debug dialogues and things all about the code, but this is the only real example of us vomiting it to to the whole world via a ``to_chat(world,...)`` call.

What's important to the players is that:
- the game is in fact booting up, and
- their client isn't broken or crashed.

And so that's what this dialogue is: a nicer-looking series of percentages that give a basic, approximate time before the server is booted up and the ready-up timer starts ticking down. The actual timers are visible to users with admin permissions, so coders booting up private copies of the codebase will be able to see these diagnostics just fine, should they want to improve the performance of subsystem code. 

This is a lot of what first-time players of the server see when they join, so I think it can be meaningful to make it look as nice and professional as this stupid game engine can allow.

# Changelog

:cl: Altoids
spellcheck: The loading screen now provides a lovely percent-completion as the game boots up.
spellcheck: The strange atmospherics equalization demon has been quieted.
/:cl:

### DIFF
--- a/code/controllers/subsystem.dm
+++ b/code/controllers/subsystem.dm
@@ -35,6 +35,12 @@
 
 	var/static/list/failure_strikes //How many times we suspect a subsystem type has crashed the MC, 3 strikes and you're out!
 
+	//yogs start -- loading time stuff
+	var/static/total_loading_points_progress = 0 //! How much progress we've made in loading all the subsystems so far.
+	var/static/total_loading_points = 0 //! The total amount of loading points among all subsystems. Should be defined by MC before subsystem inits.
+	var/loading_points = 0 //! The amount of loading points this subsystem has, measured in deciseconds of approximate load time. This being 0 is fine.
+//yogs end
+
 //Do not override
 ///datum/controller/subsystem/New()
 
@@ -159,15 +165,24 @@
 /// Called after the config has been loaded or reloaded.
 /datum/controller/subsystem/proc/OnConfigLoad()
 
-//used to initialize the subsystem AFTER the map has loaded
+///used to initialize the subsystem AFTER the map has loaded
+///This should be called by the derived subsystem class AFTER it has done its own initialization.
 /datum/controller/subsystem/Initialize(start_timeofday)
 	initialized = TRUE
 	SEND_SIGNAL(src, COMSIG_SUBSYSTEM_POST_INITIALIZE, start_timeofday)
-	var/time = (REALTIMEOFDAY - start_timeofday) / 10
-	var/msg = "Initialized [name] subsystem within [time] second[time == 1 ? "" : "s"]!"
-	to_chat(world, span_boldannounce("[msg]"))
+	var/time = (REALTIMEOFDAY - start_timeofday)/10
+	var/msg = "Initialized [name] subsystem within [time] second[time == 1 ? "" : "s"]!" // Yogs -- quieter subsystem initialization
+	to_chat(GLOB.admins,
+		type = MESSAGE_TYPE_DEBUG,
+		html = span_notice(msg),
+		confidential = FALSE) 
 	log_world(msg)
-	return time
+	if(!loading_points) // We're probably one of those crappy subsystems that take 0 seconds, so return
+		return time
+	total_loading_points_progress += loading_points
+	var/percent = round(total_loading_points_progress / total_loading_points * 100)
+	to_chat(world,span_boldnotice("Subsystem initialization at [percent]%..."))
+	return time // Yogs end
 
 
 /datum/controller/subsystem/stat_entry(msg)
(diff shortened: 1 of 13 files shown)


code/controllers/subsystem/air.dm
```diff
@@ -391,9 +393,14 @@ SUBSYSTEM_DEF(air)
 			//EG.dismantle()
 			CHECK_TICK*/
 
-		var/msg = "HEY! LISTEN! [DisplayTimeText(world.timeofday - timer)] were wasted processing [starting_ats] turf(s) (connected to [ending_ats] other turfs) with atmos differences at round start."
-		to_chat(world, span_boldannounce("[msg]"))
-		warning(msg)
+		//Yogs start -- prettier atmos notices
+		var/msg = "HEY! LISTEN! [(world.timeofday - timer)/10] seconds were wasted processing [starting_ats] turf(s) (connected to [ending_ats] other turfs) with atmos differences at round start."
```
Simplehorror, line 397: its meant to be annoying please just fix the atmospherics

Simplehorror:
I like it

code/controllers/subsystem/air.dm
Altoids1: If anything, the solution would be for the thing that's complaining to have some option to write a new map file post-equalization that can just be saved.

Provided anyone really does care about the, <100 ms that this is adding to load times on Box. Not a big performance target + out of scope of this PR + looks bad

Simplehorror: I like that idea

code/controllers/subsystem.dm
```diff
@@ -35,6 +35,12 @@
 
 	var/static/list/failure_strikes //How many times we suspect a subsystem type has crashed the MC, 3 strikes and you're out!
 
+	//yogs start -- loading time stuff
+	var/static/total_loading_points_progress = 0 // How much progress we've made in loading all the subsystems so far.
```
Bibby0110: make them autodocs since you went through the trouble of commenting your code

code/controllers/subsystem.dm
```diff
@@ -163,11 +169,20 @@
 /datum/controller/subsystem/Initialize(start_timeofday)
 	initialized = TRUE
 	SEND_SIGNAL(src, COMSIG_SUBSYSTEM_POST_INITIALIZE, start_timeofday)
-	var/time = (REALTIMEOFDAY - start_timeofday) / 10
-	var/msg = "Initialized [name] subsystem within [time] second[time == 1 ? "" : "s"]!"
-	to_chat(world, span_boldannounce("[msg]"))
+	var/time = (REALTIMEOFDAY - start_timeofday)
```
Bibby0110: just divide by 10 here, i don't see you using the straight up time value anywhere

code/controllers/subsystem.dm
```diff
@@ -163,11 +169,20 @@
...
+		confidential = FALSE) 
 	log_world(msg)
-	return time
+	. = time/10
+	if(!loading_points)
+		return
+	total_loading_points_progress += loading_points
+	var/percent = round(total_loading_points_progress / total_loading_points * 100)
+	to_chat(world,span_boldnotice("Subsystem initialization at [percent]%..."))
+	return // yogs end
```
Bibby0110:
```suggestion
```
or (with a removal of . = time/10 above)
```suggestion
	return time
```

Bibby0110 (changes requested):

code/controllers/subsystem.dm
Altoids1: It'd be the first autodoc committed in ages but uhh... sure?

Bibby0110: gotta start somewhere

thankfully we have good habits of doi... wait

since you already made them may aswell move em and make em autodoc

yogstation/code/controllers/subsystem/yogs.dm
```diff
@@ -6,6 +6,8 @@ SUBSYSTEM_DEF(Yogs)
 	flags = SS_BACKGROUND
 	init_order = -101 //last subsystem to initialize, and first to shut down
 
+	loading_points = 0.1 SECONDS
```
nmajask:
```suggestion
	loading_points = 0.1 SECONDS // Yogs -- loading times
```
The only change required to make it semi-modular

Bibby0110 (approved):

PR #13663, opened 2022-04-10
Makes turdis use updated checkout action

adamsong:

Mainly because I know this version uses the merge ref for the check (not sure if the old one did, but this one definitely does)


adamsong:
This doesn't work as intended, sad

PR #13666, opened 2022-04-10
Added sec, engi and atmos hardsuit crates buyable at cargo

00ze-cyclone:

The hardsuits are extremely useful for each department however theres not really any reliable mean of getting more if they need to be remplaced should they be destroyed or if the department has lots of people needing them.
They all have the access requirement of the rooms each hardsuits are normally found at roundstart.
Also upped the cost of the SWAT crate to 12000 credits so that there's a reason to take the sec hardsuit crate over the SWAT crate.

# Wiki Documentation
each crate cost 5000 credits
Security Hardsuit Crate in the armory section
Engineering Hardsuit Crate and Atmospheric Hardsuit Crate in the engineering section
SWAT crate now cost 8000 credits

# Changelog

Let people buy sec, engi and atmos hardsuits at cargo for 5000 credits, the crates have the access requirement normally needed to get the hardsuits of the stations at roundstart.

:cl:  
rscadd: Added sec, engi and atmos hardsuit crates at cargo
tweak: changed SWAT crate cost to 12000 
/:cl:

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -647,7 +647,7 @@
 	crate_name = "stun mine crate"
 
 /datum/supply_pack/security/armory/hardsuit
-ame = "Security Hardsuit Crate"
+	name = "Security Hardsuit Crate"
 	desc = "Contains a reinforced security hardsuit, a security gasmask and a set of combat magboots for combat in low pressure environments. Requires Armory access to open."
 	cost = 3000
 	contains = list(/obj/item/clothing/suit/space/hardsuit/security,
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -806,7 +806,7 @@
 	desc = "Contains an atmospheric hardsuit, a breath mask and a set of magboots for working in high temperature low pressure environments. Requires Atmospheric access to open."
 	cost = 3000
 	access = ACCESS_ATMOSPHERICS
-	contains = list(/obj/item/clothing/suit/space/hardsuit/atmos,
+	contains = list(/obj/item/clothing/suit/space/hardsuit/engine/atmos,
 					/obj/item/clothing/mask/breath,
 					/obj/item/clothing/shoes/magboots)
 	crate_name = "atmospheric hardsuit crate"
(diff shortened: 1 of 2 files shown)


JamieD1:
Unsure on this one to be honest @SomeguyManperson

NCRLouTenant:
Okay, so I hate to try to do the math, but 3000 credits seems kinda cheap for a hardsuit if you base it on the amount of points it takes to buy one for miners. According to the wiki 1 plasma ore is worth 15 points, and 1 sheet of plasma sells for 200 credits. That would put the value of the mining hardsuit at around 26000 credits.

3000 credits is just 15 sheets of plasma, just 225 points. Even the SWAT one would only be 600 points. (The caveat hinging on whether the wiki is correct or not regarding the value and if I didn't screw up my own math.)

00ze-cyclone:
Mining point are gained alongside the money you can get by selling the minerals, it's not really the mining hardsuit cost 26000 credits and more every 26000 credits you can have a mining hardsuit. Miners can also make a huge amount of point with plasma cutters to bring lots of ores and the ORM can be upgraded to give more points.

I do admit however than 3000 credits seem a little cheap.

00ze-cyclone:
Changed hardsuit's prices to 5000 and SWAT to 12000, sec have their budget so they can afford those things without bleeding cargo's budget dry.

nesisknesis:
Scarcity makes for interesting gameplay. I can't say much about the pricing being balanced but adding them to cargo seems like a serious convenience. Hardsuits are advanced spacesuits so if you can just order a brand spanking new one from cargo for a little extra there's little to suggest they're any much more valuable than a normal spacesuit.

Gotta remember this a chaotic space station simulation. If you lose that expensive hardsuit you should feel the consequences.

Cargo also sells space suits which as a more common item makes sense. But I would see little point in ordering these when you have all the hardsuits for sale that provide much more significant benefits.

JamieD1:
Can you limit it to 1 extra crate per shift?

00ze-cyclone:
One crate per shift would defeat the point of being able to order them, everything on the station can be remplaced, the machinery, the equipment... If the armory is stolen you can remplace it, if the AI die you can make another, any room of the station can be completely rebuilt if you want. If you can only buy one more then it's no different from just adding one hardsuit of each type at roundstart and making them unobtainable otherwise.
I suppose making it buyable by cargo isn't the right way, I'll attempt to make it a researchable tech instead.

PR #13675, opened 2022-04-11
Podperson passive healing nerf 

Chubbygummibear:
Any amount of free passive healing is strong, so i'm cutting it in half. I've seen the Phytosian passive healing get complained about a number of times but no one ever does anything about it (also Truz asked me to nicely so i'm doing it), so i'm shooting myself and my people in the foot to hopefully put an end to the debate.

# Wiki Documentation

I don't think wiki had the actual healing values, so nothing if that's the case.

# Changelog

:cl:  
tweak: podperson healing cut by 50% across the board  
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/carbon/human/species_types/plantpeople.dm
+++ b/yogstation/code/modules/mob/living/carbon/human/species_types/plantpeople.dm
@@ -92,12 +92,12 @@
 					H.adjustOxyLoss(min(3 * dark_damage_multiplier, 50 - H.getOxyLoss()), 1)
 			if (0.31 to 0.5)
 				//medium, average, doing nothing for now
-				light_level = 3	
+				light_level = 3
 				H.nutrition += light_amount * 2
 			if (0.51 to 0.75)
 				//high light, regen here
 				light_level = 4
-				H.nutrition += light_amount * 1.75				
+				H.nutrition += light_amount * 1.75
 				if ((H.stat != UNCONSCIOUS) && (H.stat != DEAD) && !no_light_heal)
 					H.adjustOxyLoss(-0.5 * light_heal_multiplier, 1)
 					H.heal_overall_damage(1 * light_heal_multiplier, 1 * light_heal_multiplier)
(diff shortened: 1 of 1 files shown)


ShatteredParadox:
Imagine nerfing a species that gets husked in like 6 strikes of a welder.

JamieD1:
@SomeguyManperson

Marmio64:
only time i've ever had a problem when going up against a podperson is when im in a floor tile throwing or punching contest with one. If you genuinely want to kill a podperson  you can just use a welder, lasergun, or light them on fire to kill them fast. Considering they're also slower than other roundstart races and can be outran, i'd say the decent passive healing is fine. The hunger changes do sound nice though (albeit they remove a quirk of them and an easy way to troll them), i've not played podperson very many times but it can be annoying to not be able to eat anything.

Simplehorror:
The only time this healing comes into play is if they were crit by environmental forces, and the healing does not even fix wounds. How about making it so they need CO2 to heal?

neeshacark:
please god don't merge this.

plantpeople have SEVERE weakness to burn and I've never seen anyone who's been playing for longer than a month complain about this because it's so easy to kill a pod if you burn them and make sure they're in hardcrit or dead.

podperson healing does not need a nerf, the downsides of podpeople easily make up for this strength.

Reed0506:
While I dont think needing to use sugar to heal at a decent rate is a bad thing, I do think just slashing the already extremely situational weak healing could be used to implement a small buff. Pod nutrition on the other hand being toned down is great, and might alleviate the negative moodlets that being """fat""" by happen by existing. Also, if this healing nerf makes pods unable to fully resist the toxin damage of romerol tumors, that really sucks, as it is one of the only upsides of playing pod.

SomeguyManperson:
atomize this into a pr for nutrition and one for the healing

Chubbygummibear:
> atomize this into a pr for nutrition and one for the healing

should i close this one and open two, or take the nutrition out of this and make just 1 new pr for that?

Chubbygummibear:
This is now just about the healing nerf and the light feeding tweaks are in #13756

adamsong (changes requested):
First, I hate this change
Second, if you are going to do it, make sure you set it back properly when sugar wears off

Chubbygummibear:
> First, I hate this change Second, if you are going to do it, make sure you set it back properly when sugar wears off

should be taken care of since theos made me use a more dynamic method of setting it back to the initial multiplier in the sugar bugfix here #13748. i didn't want to touch it here because i didn't want this pr doing like the same thing with sugar

SomeguyManperson:
Pods are already weaker+slower than other races so having passive healing isnt particularly powerful, especially considering the ease of access to weapons that can deal burn damage

PR #13716, opened 2022-04-13
Fix handcannon not being reloadable and makes the reload text cooler

Chubbygummibear:
turns out the calibers have to match, if only some coder hadn't recommended he change it to 100mm at the last second 
cough @ToasterBiome cough
~~we love you anyway~~

tested and works
image

:cl:  Chubbygummibear, Lazennsky
bugfix: handcannon can now reload
/:cl:


ToasterBiome (approved):
i deserve a credit for making him make the caliber realistic

PR #13756, opened 2022-04-17
Podpeople don't feed off light so much that they are perpetually fat

Chubbygummibear:
Tones down light feeding to the point where pods stop feeding at "fed" when in full light, and at weaker light it's just enough to not be hungry. Now the vegetable people can actually eat if they choose to.

# Wiki Documentation

We could remove the

"If mood is enabled, a phytosian will often receive the "I'm so fat" negative moodlet, making them slower than usual."

snippet

# Changelog

:cl:  
tweak: Podpeople feeding off light toned down. You feed a decent amount off light, but not so much you're fat. So pods can actually eat a bit now without incurring obesity.
/:cl:


SomeguyManperson (approved):

patpol4:
Stupid

PR #13765, opened 2022-04-17
Fixing issues with psyc office. On yogbox

Readystorm546:
The image below is bad because it has two cameras. So I fixed it
image

I made it so that when your next to the windows they don't let you see through
image

:cl:   
rscdel: Removed duplicate camera on Yogbox
rscdel: Removed tinted windows on one side and moved them to the other on Yogbox.
/:cl:


patpol4:
map name in changelog missing

Readystorm546:
> map name in changelog missing

image

JamieD1 (approved):
Better

PR #13780, opened 2022-04-19
Halves the price of entry to the luxurious shuttle

N3D6:
# Document the changes made in your pull request

The price to board the luxury escape shuttle is now 250 from 500. 

This change is good for the game because 500 is a steep and stupid price to enter the shuttle, and those who don't pay are locked in a cesspool for 6 minutes. This will still be a high price to enter the shuttle as the only roles that really get this kind of payment are engi, science, sec and command roles, meaning everybody and their mother won't be able to just outright board the shuttle.

# Wiki Documentation

:cl:  
tweak: the luxury shuttle now costs 250 to board instead of 500
/:cl:

### DIFF
--- a/code/datums/shuttles.dm
+++ b/code/datums/shuttles.dm
@@ -240,8 +240,8 @@
 /datum/map_template/shuttle/emergency/luxury
 	suffix = "luxury"
 	name = "Luxury Shuttle"
-	description = "A luxurious golden shuttle complete with an indoor swimming pool. Each crewmember wishing to board must bring 500 credits, payable in cash and mineral coin."
-	extra_desc = "This shuttle costs 500 credits to board."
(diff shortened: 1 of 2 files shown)


Marmio64:
everyone but service and civillian roles along with cargo techs should be able to afford the entry fee. If you're poor, get mad at command for picking the shuttle and just hitch a ride in a disposals pipe/bluespace bodybag.

ShatteredParadox:
It's called luxury shuttle for a reason. If you really want into the main part, have a shaft miner get money with points, and bum a ride.

Cartlord:
I think your math is bad, because even as a Botanist I was able to reliably board the old luxury shuttle as long as I didn't buy any contraband seeds.

N3D6:
This pr sucks

Marmio64:
> I think your math is bad, because even as a Botanist I was able to reliably board the old luxury shuttle as long as I didn't buy any contraband seeds.

bold of you to assume i'd spend time doing math to comment on a pr

PR #7035, opened 2019-11-06
THE GREAT TG MUTATION PORT

boodaliboo:
### Intent of your Pull Request
this pr ports a variety of mutations from tg including anti-glow, two left feet, autonomy
Autonomy allows you to discord a random limb at will
anti-glow is what it name sake says it takes light away
two left feet gives you two left feet which increase stun time

links to original prs
https://github.com/tgstation/tgstation/pull/46300
https://github.com/tgstation/tgstation/pull/45990
https://github.com/tgstation/tgstation/pull/45988
https://github.com/tgstation/tgstation/pull/42524
#### Changelog

:cl:  
rscadd: does ports of a variety of new mutations
/:cl:

PR #13784, opened 2022-04-19
Slightly alters the height dwarfism sets you to

fluffe9911:
Changed the height dwarfism sets you to 0.8 for both width and height as it was originally

# Changelog
:cl:  
tweak: Scientists have altered the Dwarfism gene back to its original genetic code
/:cl:

### DIFF
--- a/code/datums/mutations/body.dm
+++ b/code/datums/mutations/body.dm
@@ -92,14 +92,14 @@
 /datum/mutation/human/dwarfism/on_acquiring(mob/living/carbon/human/owner)
 	if(..())
 		return
-	owner.transform = owner.transform.Scale(1, 0.85)
+	owner.transform = owner.transform.Scale(0.8, 0.8)
 	passtable_on(owner, GENETIC_MUTATION)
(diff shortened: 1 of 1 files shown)


ToasterBiome:
Why?

fluffe9911:
> Why?

The dwarf gene when the height modifier was just a flat percentage change for both height and width would shrink your entire character down to 80% but since the refactor to separate the height and width modifiers while it still shrinks you down to 80% height wise it does not do it width wise this fixes that oversight

ToasterBiome:
> > Why?
> 
> The dwarf gene when the height modifier was just a flat percentage change for both height and width would shrink your entire character down to 80% but since the refactor to separate the height and width modifiers while it still shrinks you down to 80% height wise it does not do it width wise this fixes that oversight

TG still has it at 1 and .8, I am confused as to why this is needed

Trustback:
#7035
Looks like boodaliboo forgot to mention that changing dwarfism was part of the PR. None of the PRs that were merged did intentionally change the gene by itself. By all measures, this was an undocumented change in the PR that should at least be addressed? Either way, it's going back to old code.

JamieD1 (approved):

PR #13779, opened 2022-04-19
Remake of 13754 - (...) Slips dont stop tail wagging

Sniblet:
At last.

# Document the changes in your pull request

Jamie's #13754 of #13452 doesn't do anything, I'm going to try and do something modular once someone like @nmajask tells me what that means
The end goal is that slipping will not stop tails wagging

- [x] tested
- [x] works

# Changelog

:cl:  
bugfix: Slips no longer stop tails wagging
/:cl:
And this time it really is a bugfix

PR #13754, opened 2022-04-16
Remake of #13452 - Slips dont stop tail wagging

JamieD1:
Remake of #13452 - Slips dont stop tail wagging

I know cringe, but whatever.

:cl:  
tweak: Slips no longer stop tails wagging
/:cl:

PR #13795, opened 2022-04-20
Revert "Remake of #13452 - Slips dont stop tail wagging"

Sniblet:
Reverts yogstation13/Yogstation#13754
Regardless of what happens with #13779 this code does not function


Sniblet:
13379 was merged this does nothing

PR #13797, opened 2022-04-20
Cut cult stun times in half

Simplehorror:
# If this is your first PR, or not, take the time to read our CONTRIBUTING.md file! You can see it here: https://github.com/yogstation13/Yogstation/blob/master/.github/CONTRIBUTING.md

I cut cult stun times in half. You can already stack stuns with cult so you can stun someone for a full minute with change before changes, and we are moving away from hard stuns anyways. This leaves plenty of time to cuff and take off a headset, which was intended behaviour instead of dragging straight to a conversion rune after stacking 4 stun spells. This will make it easier to balance in future.

# Wiki Documentation
Cult stun times have been cut in half.

# Changelog

:cl: 
tweak: Cut cult stun times in half.
/:cl:

### DIFF
--- a/code/modules/antagonists/cult/blood_magic.dm
+++ b/code/modules/antagonists/cult/blood_magic.dm
@@ -458,7 +458,7 @@
 					L.adjustBruteLoss(30)
 			else if(cult.cult_risen)
 				to_chat(user, span_cultitalic("In a dull flash of red, [L] falls to the ground!"))
-				L.Paralyze(80)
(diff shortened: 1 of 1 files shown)


Marmio64:
It'd be better to try to find a way to stop stun stacking instead.

Simplehorror:
> It'd be better to try to find a way to stop stun stacking instead.

Even so I think a 16 second stun is absurd even for the start of the round when you only need about 2 seconds to cuff, and then a few more seconds to remove the headset. It's handholding

PR #13806, opened 2022-04-20
Clockwork rebalance

Simplehorror:

Clockwork cult have a low win rate and I believe this is due to the fact that stargazers are gone so power is usually low because there is no alternative, which means less defenses and less marauders. I also believe the visor is somewhat of a noob trap because it does not really do that much overall, too easy to avoid and even if you don't avoid it, it only does 20 damage and 1.5 second stun. I also increased the superheated wall limit to help counter hulks and mechs who usually crush cults under feet, but this might be because the cult has an eminence so I am thinking about making it auto activate.

I might also add a way to convert the new AI system

# Wiki Documentation
Integration cogs now give double the power

# Changelog

:cl:  
tweak: 10 more superheated walls for clockcult
tweak: Integration cogs give double the power
/:cl:

### DIFF
--- a/code/modules/antagonists/clockcult/clock_items/integration_cog.dm
+++ b/code/modules/antagonists/clockcult/clock_items/integration_cog.dm
@@ -33,7 +33,7 @@
 		var/power_adjust = watts_per_second / 5
 		var/obj/item/stock_parts/cell/cell = apc.cell
 		if(cell && (cell.charge / cell.maxcharge > COG_MAX_SIPHON_THRESHOLD))
-			cell.use(1)
+			cell.use(2)
 			adjust_clockwork_power(power_adjust) //Power is shared, so only do it once; this runs very quickly so it's about 10 W/second
 			
 		else
--- a/code/modules/antagonists/clockcult/clock_items/integration_cog.dm
+++ b/code/modules/antagonists/clockcult/clock_items/integration_cog.dm
@@ -6,12 +6,14 @@
 	desc = "A small cogwheel that fits in the palm of your hand."
 	clockwork_desc = "A small cogwheel that can be inserted into an open APC to siphon power from it passively.<br>\
 	<span class='brass'>It can be used on a locked APC to open its cover!</span><br>\
-	<span class='brass'>Siphons <b>10 W</b> of power per second while in a powered APC.</span><br>\
-	<span class='brass'>Siphons <b>5 W</b> of power per second while in an unpowered APC.</span>"
+	<span class='brass'>Siphons <b>20 W</b> of power per second while in a powered APC.</span><br>\
+	<span class='brass'>Siphons <b>10 W</b> of power per second while in an unpowered APC.</span>"
 	icon_state = "wall_gear"
 	w_class = WEIGHT_CLASS_TINY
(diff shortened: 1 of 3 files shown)


ghost:
Finaly

ghost (approved):

bruhlookatthisdood:
dudes re-balancing the entire game

Bibby0110:
I don’t think we changed the amount of superheated walls when we made it a circle, so I think that part is fine (maybe even too low still)

Simplehorror:
> I don’t think we changed the amount of superheated walls when we made it a circle, so I think that part is fine (maybe even too low still)

Yeah I was wondering why it was so low, makes sense. I think the major issue with superheated walls is that it requires an eminence and most people never activate the beacon so I want to make it auto activate

ghost:
Better make clockwork armor have minor laser resistance and make the ark have major explosion resistance

Simplehorror:
would rather add reflectors ala blob

bruhlookatthisdood:
> would rather add reflectors ala blob

blob buff? 😳

TymurShatillo:
> would rather add reflectors ala blob

Revive infection

ghost:
> would rather add reflectors ala blob

Yes

Marmio64:
Judicial visor is already really good for being practically free. It's a combat initiator to slowdown and disarm someone or if you want to get away from them. Only good change is more superheated walls.

code/modules/antagonists/clockcult/clock_items/integration_cog.dm
```diff
@@ -28,11 +30,13 @@
 		else
 			STOP_PROCESSING(SSfastprocess, src)
 	else
+		var/power_adjust = watts_per_second / 5
 		var/obj/item/stock_parts/cell/cell = apc.cell
 		if(cell && (cell.charge / cell.maxcharge > COG_MAX_SIPHON_THRESHOLD))
 			cell.use(1)
```
SomeguyManperson:
```suggestion
			cell.use(2)
```

code/modules/antagonists/clockcult/clock_items/judicial_visor.dm
```diff
@@ -196,7 +196,7 @@
 				L.visible_message(span_warning("Strange energy flows into [L]'s [I.name]!"), \
 				span_userdanger("Your [I.name] shields you from [src]!"))
 			continue
-		L.Paralyze(15) //knocks down briefly when exploding
```
SomeguyManperson: This is already mostly fine, don't see any need to change it. The stun isn't very long but it can be chained into other stuff

code/modules/antagonists/clockcult/clock_items/judicial_visor.dm
```diff
@@ -196,7 +196,7 @@
...
+		L.Knockdown(50) //knocks down briefly when exploding
```
JamieD1:
```suggestion
		L.Paralyze(15) //knocks down briefly when exploding
```

JamieD1 (approved):

bruhlookatthisdood:
coggers

PR #13813, opened 2022-04-20
Making holes in the SM chamber will provide a lot of +*=FUN=*+

Simplehorror:
Making a hole in the supermatter chamber to vent it and so to slow down the delamination will instead make the delamination way faster, more so if the SM power is very high and the power level is high too.

Fix an old unintended feature where you could just vent to space the SM chamber to slow down or even stop a delamination (sometimes you could just make a hole in the chamber and it would stabilize while delaminating), the SM is easily fixable just by looking at the pipe and the meters, no need to vent to space. (Lemons note, in some cases a vent to space is needed,  but it should not be a long term thing)

Port of https://github.com/tgstation/tgstation/pull/53725
Finally removes the supermatter cheese of if delam, then space

# Wiki Documentation
Do not space the supermatter anymore

# Changelog

:cl: Ghilker
rscadd: SM will now delaminate faster if exposed to space
/:cl:

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -82,6 +82,8 @@
 #define SUPERMATTER_COUNTDOWN_TIME 30 SECONDS
 #define SUPERMATTER_ACCENT_SOUND_MIN_COOLDOWN 2 SECONDS ///to prevent accent sounds from layering
 
+#define MAX_SPACE_EXPOSURE_DAMAGE 2
+
 GLOBAL_DATUM(main_supermatter_engine, /obj/machinery/power/supermatter_crystal)
 
 /obj/machinery/power/supermatter_crystal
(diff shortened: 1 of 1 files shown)


stwordcitizen:
why is this a good change

Simplehorror:
> why is this a good change

You must actually have skill to stop a delaminating supermatter instead of removing a floor tile with an RCD brainlessly

Moja10:
I like this, maybe. People should go see it in action on /tg/ before making judgements.

Jarod1200:
im still gonna do it anyway but ok. default space loop is not usually good enough to get rid of a ton of superheated plasma on its own. hence why you space it to dump the hot gas and then refill it.

Sniblet:
This doesn’t seem to actually check if the crystal is in a vacuum, just checks if there’s an adjacent space tile
So this both draws on experience from a server that doesn’t have monstermos for some reason and is super easy to bypass

Simplehorror:
> This doesn’t seem to actually check if the crystal is in a vacuum, just checks if there’s an adjacent space tile So this both draws on experience from a server that doesn’t have monstermos for some reason and is super easy to bypass

Yes this is very bad I just want to see if it would go down well in the community and then I'll fix it

Altoids1:
Honestly, I've always wanted the supermatter to delaminate in literally 30 seconds. Despite its complexity, it's actually harder to use dangerously than the singu or tesla. This is a great idea.

SomeguyManperson:
spacing the supermatter to completely negate a delam's effects has always been stupid
Should be possible to make it check incoming gasses for any volume and penalize it if it can't find anything

stwordcitizen:
> > why is this a good change
> 
> You must actually have skill to stop a delaminating supermatter instead of removing a floor tile with an RCD brainlessly

alright so instead of removing a tile you can just wait for a. the sm delamming, or b. the cooling loop filters the gasses and cools them down. Tile removal isn't even bad or annoying. No one complained about this feature.

JamieD1:
Considering you are balancing this around people with a high skill floor for the SM, the usual round does not contain enough people qualified enough to stop a delamming SM. This "cheese" is actually needed for unskilled or engineers who are not there. When an SM is delamming, the stock setup just does not cut it and I doubt there will be a different setup in place before this becomes unmanageable.

PR #13823, opened 2022-04-21
Radaway renamed to Rad Scrub Plus

Chubbygummibear:
Radaway made sense at first when it still also helped treat regular radiation, but that effect was cut out, so now i feel like it's misleading because people think it's like the fallout radaway instead of space cleaner for radioactive contamination. Hopefully this rebranding will make it's purpose a bit more evident at a glance

# Wiki Documentation

Radaway rebranded to Rad Scrub Plus (By Donk Co)

# Changelog

:cl:  
tweak: Radaway renamed to Rad Scrub Plus, so it's use as a radioactive contamination cleaner is more self-evident
/:cl:

### DIFF
--- a/_maps/RandomRuins/LavaRuins/lavaland_surface_cafe_of_broken_dreams.dmm
+++ b/_maps/RandomRuins/LavaRuins/lavaland_surface_cafe_of_broken_dreams.dmm
@@ -58,8 +58,8 @@
 /turf/open/floor/wood,
 /area/ruin/powered)
 "aF" = (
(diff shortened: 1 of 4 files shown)


ghost:
Honk Co

neeshacark:
yep, this is a problem.  i thought it got rid of radiation on people too and i drank a few bottles (and had quite a few pills the chemists made) since it was added.

i didnt know it didn't do that til i saw this PR

JamieD1 (approved):

PR #13831, opened 2022-04-21
Makes stimballs make balls

ToasterBiome:

https://github.com/yogstation13/Yogstation/pull/9866 caused it to never release any balls ever because of the early return before any balls are shot. Colten said something about a runtime but that's fixed with the isnull check for making balls. Looks like an oversight.

# Changelog

:cl:  
bugfix: stimball reaction now makes radballs again
/:cl:


monster860 (approved):

Chubbygummibear:
nice balls, king

PR #13847, opened 2022-04-22
Fixes Wejes in yogstation

JamieD1:
Fixes Wejes in yogstation

:cl:   
bugfix: Fixes Wejes issues with yogstation
bugfix: Adds Baiomus fixes as well
/:cl:


Cartlord:
What fixes might these be?

PR #13800, opened 2022-04-20
Nerf shadowling icy veins

Simplehorror:
# Document the changes in your pull request
I have replaced the stun in icy veins with knockdown, removed the initial body temperature change and decreased the amount of frost oil. I have done this because icy veins is extremely powerful due to the fact it has a high range and a low cooldown and is spammable, and has no limit to how many people it effects. It is basically a get out of trouble spell and shadowlings have enough of that in terms of jaunts and having an army of thralls. Shadowlings should not have an AOE spell that claims to be more effective than a cryosting per person.

# Wiki Documentation
Shadowling icy veins are less of a crutch.
# Changelog

:cl:  
tweak: Shadowling icy veins are no longer as powerful.
/:cl:

PR #13848, opened 2022-04-22
icy veins nerf, other abilities buff, empowered thrall buff

Marmio64:
closes: #13800 

This nerfs icy veins and buffs a few other sling things.

veil cooldown reduced from 15 seconds to 12.  This is their bread and butter ability and needs to be buffed to better help them since icy veins will be weaker

Icy veins no longer causes instant 10 burn damage and injects 5 less units of frost oil. Icy veins was too OP, though it was their sole OP ability. It is still really good but less so.

Empowered thralls get the void jaunt ability. Still, nobody uses them, and people probably will continue to not do it. But, this might make them a little bit better, especially since previously they had zero mobility abilities.

Null charge takes 15 seconds to do instead of 20. Standing around for 20 seconds as a shadowling is near impossible with crew running around. 15 is still really hard but more reasonable.

Sonic screech paralyzes borgs for 5 seconds rather than .6 seconds. This is literally their only method of doing anything to borgs and was useless still...

# Changelog

:cl:  
rscadd: Empowered thralls get void jaunt
tweak: Icy veins nerfed. Veil, null charge, and sonic screech buffed.
/:cl:

### DIFF
--- a/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
+++ b/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
@@ -543,7 +543,7 @@
 				var/datum/effect_system/spark_spread/sp = new /datum/effect_system/spark_spread
 				sp.set_up(5, 1, S)
 				sp.start()
-				S.Paralyze(30)
+				S.Paralyze(50)
 		for(var/obj/structure/window/W in T.contents)
 			W.take_damage(rand(80, 100))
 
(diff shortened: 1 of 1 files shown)


Marmio64:
an alternative to #13800. Nerfs icy veins while returning some of the lost power to their other abilities.

Simplehorror:
I'd be happy with this if the stun was removed from icy veins because it is being used as an alternative to glare

Simplehorror:
icy veins was not nerfed much, burn damage is the smallest part of it and is mostly dealt by the frost oil anyways

Simplehorror:
Just increase the cooldown to like 5 minutes

ghost:
> Just increase the cooldown to like 5 minutes

Bruh what

Simplehorror:
> > Just increase the cooldown to like 5 minutes
> 
> Bruh what

It's a defensive ability not offensive and should be used as an emergency escape, increasing the cooldown so that you can't use it repeatedly would reflect this

JamieD1 (approved):

Simplehorror:
You forgot the part where you nerfed icy veins

PR #13874, opened 2022-04-24
Stops dynamic from giving latejoiners every valid latejoin antagonist (no more traitor/vampire/bloodsucker)

ToasterBiome:

It goes over every valid ruleset it could give a latejoiner and tries to give it to them. It doesn't stop if it was successful.
This (for some reason) makes it so the candidates for future checks get trimmed, which it never checks if there's NOBODY so it runtimes and fucks everything up. 

# Changelog

:cl:   
bugfix: stops dynamic giving a lucky latejoiner every valid latejoin antagonist
/:cl:


JamieD1 (approved):

Simplehorror:
dynamic code is so fucked

PR #13814, opened 2022-04-20
Nerfs Large Ebow

stwordcitizen:
# Document the changes in your pull request

This pr nerfs large ebows, having a 40 toxin damage normal sized item that has infinite ammo and does stamina damage is overpowered. you can't easily deal with toxin damage and its quite lethal for a sec weapon.

# Wiki Documentation
Update damage value for the large ebow.

# Changelog

Edit the changelog below to reflect the changes made by this PR, even if the changes are minor - required for every PR.
If you add a name after the ':cl', that name will be used in the changelog. Leave it empty to use your GitHub name.

:cl:  
tweak: large ebow now does 20 damage instead of 40. 

/:cl:

PR #13876, opened 2022-04-24
Rebalances the security ebow

SomeguyManperson:

Printable ebow has been redesigned from a nerfed taser into a specialist weapon for high alpha damage at the start of a fight (kind of the same except it's slower)

Damage decreased to 30 from 40, increasing its hits 2 crit from 3 to 4
Stamina damage unchanged
Cooldown increased from 2 seconds to 10 seconds
Size increased to bulky
Weapon size ALSO increased to bulky (cannot be fired with something in your offhand)
Loses its charge if dropped, must  be carried on back/belt/wherever else it would fit
Charge timer increased per large ebow on your person

closes #13814
# Wiki Documentation

security large energy crossbow stats
Damage decreased to 30 from 40
Stamina damage unchanged
Cooldown increased from 1.6 seconds to 10 seconds
Size increased to bulky
Weapon size ALSO increased to bulky (cannot be fired with something in your offhand)
Loses charge if dropped
Charge timer increased per large ebow on your person

# Changelog

:cl:  
tweak: large energy crossbow damage reduced to 30 from 40
tweak: large energy crossbow cooldown increased from 2 seconds to 10 seconds
tweak: large energy crossbow size increased from normal to bulky
tweak: large energy crossbows can no longer be fired with items in the offhand
tweak: large energy crossbows will lose charge if dropped
tweak: large energy crossbows now get increased reload time for carrying several
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -109,13 +109,15 @@
 
 /obj/item/gun/energy/kinetic_accelerator/crossbow/large
 	name = "energy crossbow"
-	desc = "A reverse engineered weapon using syndicate technology."
+	desc = "A reverse engineered weapon using syndicate technology, substantially bulkier than its illegal counterpart."
 	icon_state = "crossbowlarge"
 	w_class = WEIGHT_CLASS_BULKY
 	materials = list(/datum/material/iron=4000)
 	suppressed = null
 	ammo_type = list(/obj/item/ammo_casing/energy/bolt/large)
 	pin = null
+	weapon_weight = WEAPON_LARGE
+	overheat_time = 10 SECONDS
 
 
 /obj/item/gun/energy/plasmacutter
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -116,6 +116,7 @@
 	suppressed = null
 	ammo_type = list(/obj/item/ammo_casing/energy/bolt/large)
 	pin = null
+	holds_charge = FALSE
 	weapon_weight = WEIGHT_CLASS_HUGE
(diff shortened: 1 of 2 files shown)


Marmio64:
if it's going to be bulky, deal less damage than slugs, require two-handing, and have an insane cooldown why even bother when you can just use a shotgun to kill someone before you're done fumbling around with your e-crossbow and trying to get out your next weapon before they get up and out of there. The only use I can see for this is trying to land your only shot you'll get for the next 10 seconds on a target while with a group of friendlies so that you drop the target long enough for your lads to get a shot in, but even then it doesn't seem great for because it has too slow of firing speed and just spamming lasers would be better.

Moja10:
wow that sounds awful to use. not saying it's bad but Jesus it just sounds legitimately awful to use now.

SomeguyManperson:
> if it's going to be bulky, deal less damage than slugs, require two-handing, and have an insane cooldown why even bother when you can just use a shotgun to kill someone before you're done fumbling around with your e-crossbow and trying to get out your next weapon before they get up and out of there. The only use I can see for this is trying to land your only shot you'll get for the next 10 seconds on a target while with a group of friendlies so that you drop the target long enough for your lads to get a shot in, but even then it doesn't seem great for because it has too slow of firing speed and just spamming lasers would be better.

it still deals 60 stamina damage. For reference, the normal spaceman has 100 health, and goes into damage slowdown after taking 40% of that in damage
The concept is it won't be a game ender on its own, start a fight by firing an ebow shot then switch to a disabler or other weapon and take advantage of the knockdown/high stamina damage. If you hit, high immediate damage, if you miss, you can't shoot it again for likely the rest of the fight.
Also, like most guns, it fits on your exosuit slot and belt

JamieD1 (approved):

PR #13879, opened 2022-04-24
Nerfs TEG by a flat factor of 3 power output

JamieD1:
TEG Is way too easy to get ridiculous energy amounts without any brain power.

This is still close to 700KW without modifying any pipes roundstart which is enough considering Supermatter is about 350 KW

:cl:  
tweak: TEG now Gives 1/3 Energy
/:cl:

### DIFF
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -91,7 +91,7 @@
 				var/energy_transfer = (hot_subsection_temp - cold_subsection_temp)*internal_heat_cap
 
 				//produce electricity
-				lastgen += energy_transfer*efficiency
+				lastgen += (energy_transfer*efficiency) / 3
 
 				//transfer rest of energy into waste heat/chill
 				internal_temp = cold_subsection_temp + energy_transfer * (1 - efficiency) / (internal_heat_cap * 2)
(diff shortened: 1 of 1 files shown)


Simplehorror:
so instead of using the efficency variable (2 lines above) you are just doing that

code/modules/power/generator.dm
```diff
@@ -91,7 +91,7 @@
 				var/energy_transfer = (hot_subsection_temp - cold_subsection_temp)*internal_heat_cap
 
 				//produce electricity
-				lastgen += energy_transfer*efficiency
+				lastgen += (energy_transfer*efficiency) / 3
```
ToasterBiome:
```suggestion
				lastgen += (energy_transfer*efficiency)
```

ToasterBiome: just change efficiency from 0.65 to .2-.22

ToasterBiome (changes requested):

Simplehorror:
why are we doing this when people can barely set it up

JamieD1:
> so instead of using the efficency variable (2 lines above) you are just doing that

Because that changes heat as well, this is just to do with power.

code/modules/power/generator.dm
JamieD1: > just change efficiency from 0.65 to .2-.22


Because that changes heat as well, this is just to do with power.

Cartlord:
Why'd you selfmerge this when getting the T.E.G. working is already a pain? And why does it matter how much power it makes, when any quantity of power over that which the station needs is effectively useless?

PR #13892, opened 2022-04-25
locks clown car to 25 pop

neeshacark:

Locks the clown car to where it can only be bought on station populations of 25+, as when the car is bought it is almost always used to kidnap the whole crew.  This can be annoying on lowpop, as you can spend ages trapped in a car with no security to help you.

This will prevent admin intervention from being needed in future to save the round of the poor lowpop player.

Yes this is technically an ided pr.

# Wiki Documentation

change the https://wiki.yogstation.net/wiki/Syndicate_Items page to mention the population requirement is 25.

# Changelog

:cl:  Cark
tweak: The clown car can now only be bought with a station population of 25+.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -2000,6 +2000,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			Premium features can be unlocked with a cryptographic sequencer!"
 	item = /obj/vehicle/sealed/car/clowncar
 	cost = 20
+	player_minimum = 25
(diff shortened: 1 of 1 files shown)


JamieD1:
Ideally this should be an admin issue as clowns unless they have hijack should not be murderboning.

neeshacark:
> Ideally this should be an admin issue as clowns unless they have hijack should not be murderboning.

If it was murder bone yes, but usually you just get driven around until sec stops the clown. Also imo code should back up the rules anyway

JamieD1 (approved):

PR #13899, opened 2022-04-26
fixes regal rats making minor mapping take 140+ seconds and breaking the stat panel until right when the round starts

ToasterBiome:

Minor Mapping subsystem does a funny infinite loop that decides to quit between 140 and 170 seconds, this fixes it.
Also fixes up the code, because it had a 98% chance to make a regal rat instead of a normal rat, which is obviously wrong.
Also makes it so there's no roundstart ones, because the TG wiki page doesn't mention it and alludes that it should only happen during mice migration and when migrations are triggered through mice eating cheese

# Changelog

:cl:  
bugfix: regal rats no longer spawn roundstart
bugfix: regal rats no longer occur 98% of the time instead of 2% like they should
bugfix: regal rats haver a 5% chance to spawn and no longer cause an infinite loop
/:cl:

### DIFF
--- a/code/controllers/subsystem/minor_mapping.dm
+++ b/code/controllers/subsystem/minor_mapping.dm
@@ -1,15 +1,15 @@
-#define REGAL_RAT_CHANCE 5
+#define REGAL_RAT_CHANCE 2
 SUBSYSTEM_DEF(minor_mapping)
 	name = "Minor Mapping"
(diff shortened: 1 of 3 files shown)


ghost:
Oops

ghost (approved):

adamsong (approved):

PR #13908, opened 2022-04-27
Necroseed uses physiology mods instead of dna.species mods

Mqiib:

Does the same thing, but uses physiology instead of species mods for better consistency. Punch damage is still the same, though, because those variables aren't in the physiology datum. Might fix that later or something.

# Wiki Documentation

Just a code change, probably no change needed.

# Changelog

:cl:  

bugfix: Necroseed uses physiology mods where possible rather than species mods.

/:cl:


Mqiib:
DM hates me today

SomeguyManperson (approved):

PR #13934, opened 2022-04-30
fixes a bug wherein heretic living hearts check z levels directly rather than using turfs

SomeguyManperson:

major skill issue

# Changelog

:cl:  
bugfix: being inside a locker/mech will no longer cause the heretic heart to display "vertical?" for people on the same z level as you
/:cl:

### DIFF
--- a/code/modules/antagonists/eldritch_cult/eldritch_items.dm
+++ b/code/modules/antagonists/eldritch_cult/eldritch_items.dm
@@ -14,10 +14,12 @@
 	if(!target)
 		to_chat(user,span_warning("No target could be found. Put the living heart on the rune and use the rune to recieve a target."))
 		return
-	var/dist = get_dist(get_turf(user),get_turf(target))
-	var/dir = get_dir(get_turf(user),get_turf(target))
+	var/userturf = get_turf(user)
+	var/targetturf = get_turf(target)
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/eldritch_cult/eldritch_items.dm
```diff
@@ -14,10 +14,12 @@
 	if(!target)
 		to_chat(user,span_warning("No target could be found. Put the living heart on the rune and use the rune to recieve a target."))
 		return
-	var/dist = get_dist(get_turf(user),get_turf(target))
-	var/dir = get_dir(get_turf(user),get_turf(target))
+	var/userturf = get_turf(user)
```
JamieD1:
```suggestion
	var/turf/userturf = get_turf(user)
```

code/modules/antagonists/eldritch_cult/eldritch_items.dm
```diff
@@ -14,10 +14,12 @@
...
+	var/targetturf = get_turf(target)
```
JamieD1:
```suggestion
	var/turf/targetturf = get_turf(target)
```

JamieD1 (changes requested):

JamieD1 (approved):

PR #13945, opened 2022-04-30
Nerfs Tesla Energy by a flat factor of 4 - Rework to follow

JamieD1:
Nerfs Tesla Energy by a flat factor of 4

Hopefully this encourages more singularities instead of just always going tesla.

Rework to follow soon

:cl:  
tweak: Nerfs Tesla Energy by a flat factor of 4
/:cl:

### DIFF
--- a/code/modules/power/tesla/coil.dm
+++ b/code/modules/power/tesla/coil.dm
@@ -25,6 +25,8 @@
 	var/datum/bank_account/linked_account
 	var/money_per_zap = 2 // This is tied to coil cooldown in the same way research points are.
 
+	var/powermodifier = 0.25 /// Flat amount to nerf/buff power by
+
 /obj/machinery/power/tesla_coil/power
 	circuit = /obj/item/circuitboard/machine/tesla_coil/power
 
@@ -81,7 +83,7 @@
 /obj/machinery/power/tesla_coil/tesla_act(power, tesla_flags, shocked_targets)
 	if(anchored && !panel_open)
 		obj_flags |= BEING_SHOCKED
-		add_avail((power * (1 - percentage_power_loss))*input_power_multiplier)
+		add_avail(((power * (1 - percentage_power_loss))*input_power_multiplier) * powermodifier)
 		flick("coilhit", src)
 		playsound(src.loc, 'sound/magic/lightningshock.ogg', 100, 1, extrarange = 5)
 		if(istype(linked_account))


Cartlord:
I think one other problem is that the tesla can generate research on its own (just screwdriver one coil's board), while the singulo needs a special gasmix to make research. In addition, the singulo is just more of a pain in the ass, because making it run the whole shift requires filling plasma tanks at canisters - therefore requiring either Secure Storage access (and yes I know how to get in without access, it's still not something we should expect everyone to both know and do) or Atmos access; it also shoots off E.M.P.s, which are mildly annoying to deal with (or fatal for Engineering borgs & Preterni).
By contrast, the tesla just requires you to wrench some shit down, with its only _down_side being that you're FUCKED if it looses - and engines loosing isn't something that happens often enough to matter, I don't think.

JamieD1:
> I think one other problem is that the tesla can generate research on its own (just screwdriver one coil's board), while the singulo needs a special gasmix to make research. In addition, the singulo is just more of a pain in the ass, because making it run the whole shift requires filling plasma tanks at canisters - therefore requiring either Secure Storage access (and yes I know how to get in without access, it's still not something we should expect everyone to both know and do) or Atmos access; it also shoots off E.M.P.s, which are mildly annoying to deal with (or fatal for Engineering borgs & Preterni). By contrast, the tesla just requires you to wrench some shit down, with its only _down_side being that you're FUCKED if it looses - and engines loosing isn't something that happens often enough to matter, I don't think.

Hence why I said rework to follow...

Also.. I nerfed it when you agreed with me its boring and you still downvoted. What

Cartlord:
I'm saying that this change won't make it less used, because people here (myself included) are lazy and will still make the easy engine because both it and the hard engines make enough juice to power the whole station

stwordcitizen:
I think it would be a positive change to just do away with tesla and singulo, they are braindead engines that pale in the complexity of the sm

Simplehorror:
Everytime someone nerfs something and then says they'll rework it in another PR the nerf gets merged and rework not in sight

maxion12345:
Thumbs down because the whole "nerf/remove now, rework later" is what /tg/ does and now they don't have rad collectors

maxion12345:
> I think it would be a positive change to just do away with tesla and singulo, they are braindead engines that pale in the complexity of the sm

ok but blackhole go brrrrr in ways supernatter cant

ChesterTheCheesy:
Tesla was already nerfed in a way that a single tesla can only zap like 6 coils maximum (since the zaps get weaker every bounce and eventually die out). It's the simplest engine, but doesn't offer anything in terms of coming up with your own more efficient setup. In contrast, you can build 10 additional radiation arrays to singulo. People usually pick tesla, because it's both easy to setup and safe (doesn't generate EMP's). I don't understand the arbitrary "people should do singulo more often". If players wanted to do singulo more often, absolutely nothing is stopping them.

Most importantly, did you check if your nerfed version even produces enough power for entire station?

JamieD1:
> Everytime someone nerfs something and then says they'll rework it in another PR the nerf gets merged and rework not in sight

Sure thing, not got like 4 people involved spitballing ideas or anything. But you have your bad take. Sure.

JamieD1:
> Tesla was already nerfed in a way that a single tesla can only zap like 6 coils maximum (since the zaps get weaker every bounce and eventually die out). It's the simplest engine, but doesn't offer anything in terms of coming up with your own more efficient setup. In contrast, you can build 10 additional radiation arrays to singulo. People usually pick tesla, because it's both easy to setup and safe (doesn't generate EMP's). I don't understand the arbitrary "people should do singulo more often". If players wanted to do singulo more often, absolutely nothing is stopping them.
> 
> Most importantly, did you check if your nerfed version even produces enough power for entire station?

This is the whole reason im changing it..

bruhlookatthisdood:
Im all down for the rework and stuff but maybe it should just be left as it is until the re-work is actually being made?

JamieD1:
> Im all down for the rework and stuff but maybe it should just be left as it is until the re-work is actually being made?

I guess if you want, ive been doing the rest of the engines, just felt like a good idea to get all the "buffs and nerfs" out the way

ToasterBiome (approved):
uh yeah whatever ok

Readystorm546:
I think we should see the entire rework that's gonna follow

JamieD1:
Ill just wait for the rework to prevent people pissing their pants

PR #13820, opened 2022-04-20
Purrbation works on nonhumans

SomeguyManperson:
# Document the changes in your pull request

die cringe

# Changelog

:cl:  
bugfix: purrbation works on nonhumans
/:cl:

PR #13956, opened 2022-05-01
Donor purrbation toggle does not change nonhuman characters

Sniblet:
(Humans Only)
I think we can all agree that letting donators join as lizards with cat parts is too far
Personally I say this because I've lost two rounds to forgetting to turn off purrbation when playing as a nonhuman, but the rest of you probably have regular moral objections

- [x] tested
- [x] works

If we do for whatever reason think this should be allowed, someone tell me how to make a separate Purrbation (Nonhumans Too) switch and I'll do it

Problem this is responding to is a result of #13820 , I guess you could call it a partial revert

# Changelog

:cl:  
tweak: only admins can give purrbation to nonhumans
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/felinid.dm
+++ b/code/modules/mob/living/carbon/human/species_types/felinid.dm
@@ -110,7 +110,7 @@
 		to_chat(H, "Something is nya~t right.")
 		playsound(get_turf(H), 'sound/effects/meow1.ogg', 50, 1, -1)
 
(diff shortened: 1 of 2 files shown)


NCRLouTenant:
I am staunchly against this. Let me choose to make Aurora a catthem without having to Ahelp it! 1984!

JamieD1 (approved):

ghost:
No anime

JamieD1 (approved):

PR #13958, opened 2022-05-01
Removes (humans only) from purrbation toggle

SomeguyManperson:

hey it's no longer locked to humans now

# Changelog

:cl:  
bugfix: purrbation no longer shows up as human only in donor preferences
/:cl:


Spacemanspark:
I might wonder if this would be better off on an individual character basis, rather than a blanket 'all characters' toggle.

That way donators can set specific characters to inject with heresy rather than going 'oops' and screaming as they unleash an unholy abomination of a lizard cat... thing... without intentionally meaning to.

Sniblet:
All I want is to switch races without having to think about whether I have purrbation on or not, I don't want to play an ethereal with cat ears but I've accidentally done it twice since this was added and gone instantly to cryo both times because it's a cringe overdose
Add some extra switch so that I don't have to think
Purrbation (Human only) and Purrbation (Nonhumans) or per-character like spaceman says
I don't care which

PR #13978, opened 2022-05-02
baton buffs

SomeguyManperson:

Stunbaton cooldown reduced from 2 seconds to 1.2 seconds to account for their reduced stunning capabilities since tg batons. This allows a baton to hit only .4 seconds slower rather than 1.2 seconds slower than a normal weapon (this was TWICE as slow as a normal weapon!)
Stunbatons now deal damage to the chest regardless of what limb they hit. Armor effects are still calculated by the limb. This means it MAY be more effective to aim for limbs rather than the body or chest all the time.

# Wiki Documentation

baton cooldown 1.2 seconds from 2
baton stamina damage dealt to the chest ALWAYS rather than to the targeted limb/generally

# Changelog

:cl:  
tweak: baton cooldown 1.2 seconds from 2
tweak: baton stamina damage dealt to the chest ALWAYS rather than to the targeted limb/generally
/:cl:

### DIFF
--- a/code/game/objects/items/stunbaton.dm
+++ b/code/game/objects/items/stunbaton.dm
@@ -16,14 +16,23 @@
 
 	var/cooldown_check = 0
 
-	var/cooldown = 2 SECONDS
-	var/stunforce = 100
+	///how long we can't use this baton for after slapping someone with it. Does not account for melee attack cooldown (default of 0.8 seconds).
+	var/cooldown = 1.2 SECONDS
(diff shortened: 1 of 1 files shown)


Simplehorror:
I like this PR

ToasterBiome:
Nice middle ground

ToasterBiome (approved):

ghost:
It is literally a buff of an almost insta-win weapon

PR #13983, opened 2022-05-03
Probably fixes a whole slew of explosions doing twice as much damage as they should

Mqiib:
WOW where to start?
So for like over a year or two I've been wondering why some explosions were really ridiculously insanely overtuned and yet the explosion effects they applied were normal?? This includes stuff like frag grenades, the old now-removed (but possibly returning!!) frag round, and most recently explosive holoparasites. Turns out, this is because at the epicenter of the explosion, regardless of how small it was, it would actually ex_act things on the time (like mobs) TWICE! So stuff with single-tile explosions like frag rounds and the explosive holoparasite would do TWICE as much damage as it was supposed to do.

This PR actually changes getallcontents to have an override for ignoring the object it's called on rather than adding it to the list, which is why explosion epicenters hit things a second time.

Credit to @ToasterBiome for throwing (working) code at me before I figured out how to do it myself, so I just used that.
Post-PR edit: his code didn't actually work so I had to do something else, but thanks for the help!

# Wiki Documentation

Shouldn't be anything

# Changelog

:cl:  Mqiib, ToasterBiome

bugfix: Explosion epicenters no longer do double damage.
tweak: Indirectly makes the fireball spell deal roughly half as much damage
tweak: Frag grenades and explosive holoparasites can longer instantly kill you
tweak: Anything else that did its damage via small explosions significantly nerfed

/:cl:


ToasterBiome:
add me in the changelog

also is this tested

Mqiib:
> also is this tested

Imagine assuming I don't test my questionable code. It works!!

Mqiib:
If anyone things this is a "bad nerf" and "ided" then you are "cringe" and should "go fuck yourself"

Notamaniac:
bad nerf, ided

(though does this stop gibbings if you're caught in the epicenter?)

Mqiib:
Ok don't merge this yet I need to figure out why chain reaction grenades now don't work

Mqiib:
PROBLEM SOLVED!

SomeguyManperson (approved):

PR #14008, opened 2022-05-04
adds botany supplies to mining base

UselessTheremin:
does what it says on the tin. i'm planning on updating quite a bit of lavaland food and flora, since in my mind it is a bit boring currently and could have some fun rp situations and things for the mining medic to do, maybe adding some of the meals to bounties. for now, the botany trays to mining seem like a fun addition to me because it's just another option for a thing to do as shaft miner.

# Wiki Documentation

mining base image needs an update if this is merged, possibly the wiki page for miner. 

:cl:  
rscadd: minimal botany supplies added to mining base  
/:cl:

### DIFF
--- a/_maps/RandomRuins/LavaRuins/miningbase.dmm
+++ b/_maps/RandomRuins/LavaRuins/miningbase.dmm
@@ -412,15 +412,6 @@
 /obj/machinery/atmospherics/pipe/simple/supply/hidden/layer2,
 /turf/open/floor/plasteel,
 /area/mine/eva_secondary)
-"dj" = (
-/obj/machinery/power/apc{
-	dir = 4;
-	name = "Mining Vacant Room APC";
-	pixel_x = 24
-	},
-/obj/structure/cable,
-/turf/open/floor/plasteel/dark,
-/area/mine/vacant)
 "dk" = (
 /obj/structure/table,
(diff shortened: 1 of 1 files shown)


bruhlookatthisdood:
does this mean soon the mining medic will be able to set up a kitchen on lavaland?

Cartlord:
Why bother? Multiple maps already have public gardens on the station.

SomeguyManperson:
I think cotton/bamboo should be replaced with more civilian plants so it's more for making food than making gaming gear
Also a biogen is questionable since it gives access to mutagen

neeshacark:
i do quite like this, it gives the mining medic something more to do than sit at their desk all day.

JamieD1:
Do we really want to give miners more reasons to why they are not coming back to station?

bruhlookatthisdood:
> Do we really want to give miners more reasons to why they are not coming back to station?

less opertunity for them to powergame on the station with the gamer loot they grinded for

JamieD1:
> I think cotton/bamboo should be replaced with more civilian plants so it's more for making food than making gaming gear Also a biogen is questionable since it gives access to mutagen

^

JamieD1:
Sorry, as my previous statement, I do not want to make mining never come back to station.

PR #13952, opened 2022-05-01
Handcuffed people may no longer just walk through turnstiles

JamieD1:
Handcuffed people may no longer just walk through turnstiles

:cl:  
bugfix: Handcuffed people may no longer just walk through turnstiles  
/:cl:

PR #14012, opened 2022-05-05
Revert "Handcuffed people may no longer just walk through turnstiles"

Bibby0110:
Reverts yogstation13/Yogstation#13952

Lets allow this to get some more opinions yeah? 
3v3 isn't exactly decisive

### DIFF
--- a/code/game/machinery/doors/turnstile.dm
+++ b/code/game/machinery/doors/turnstile.dm
@@ -48,11 +48,8 @@
 		for(var/mob/living/rider in mover.buckled_mobs)
 			if(allowed(rider) && !mover.pulledby) //defer to the above dragging code if we are being dragged
 				allowed = TRUE
-	var/is_handcuffed = FALSE
-	if(iscarbon(mover))
-		var/mob/living/carbon/C = mover
-		is_handcuffed = C.handcuffed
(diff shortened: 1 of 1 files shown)


maxion12345:
turnstiles, like airlocks, only care about scanning that ID card and not if you have handcuffs

that being said having new handcuff types sounds fun
shock cuffs that zap you if you walk away from the person who placed them, gps tracking cuffs, strong cuffs, etc

NCRLouTenant:
> turnstiles, like airlocks, only care about scanning that ID card and not if you have handcuffs
> 
> that being said having new handcuff types sounds fun shock cuffs that zap you if you walk away from the person who placed them, gps tracking cuffs, strong cuffs, etc

I completely agree with this.  Making those cuffs locked behind a research node would also add to that choice of what to spend points on, with sec having to try to convince Sci that it's the most important thing to do.  

Would definitely be more interested in it being related to the cuffs instead of just making handcuffs lock turnstiles, but again, that's a personal opinion based on the way I play.

patpol4:
Good Bibby PR, i don't say this often so you know it's serious.

maxion12345:
turnstiles are meant to be pushed through with your body

if you use your hands then your body is weak and you need to git gud on the turnstile because that is unsanitary

JamieD1:
This is fixing a oversight in the original PR, No.

PR #14013, opened 2022-05-05
Makes turnstiles not weak sauce

JamieD1:
Makes turnstiles not weak sauce

Makes them nearly on par with airlocks now

:cl:  
tweak: Adjusted resistance on turnstiles to be less weak 
/:cl:


ToasterBiome (approved):
👎

PR #14016, opened 2022-05-05
fixes a bug that would allow cultists to escape brazil

SomeguyManperson:
:cl:  
bugfix: cultists can no longer bypass brazil border patrols
/:cl:


bruhlookatthisdood:
my fun ):

JamieD1 (approved):

PR #14029, opened 2022-05-07
Reworks Cowboy Lawset

Skrem7:

Tweaks Cowboy lawset. The new lawset is as follows:

1) You are a cowboy, and the crew are your people.
2) **A cowboy always provides hospitality and basic aid to someone in need, even a stranger or an enemy.**
3) **A cowboy takes care of their people.**
4) A cowboy protects themself to protect others.
5) A cowboy's word is absolute – your word is your bond and an agreement is as binding as a contract.
6) A cowboy doesn't pontificate. Be concise, pardner.

Laws 2 and 3 were firstly swapped to ensure law priority would make Cowboy primarily assist the underdog (as intended). The new Law 2 was then made more clear to minimize drastic action against any aggressors, be they right or not. The new Law 3 was also phrased to more clearly have a benign intent, rather than the infamously vague "Look out for your people".

May also need council review as it changes the phrasing of a roundstart lawset, but I don't know these things!

# Wiki Documentation

Modified Cowboy Lawset: See above

# Changelog

:cl:  MenacingManatee, Skrem_7
tweak: Tries to patch up Cowboy so there are fewer cowboys-turned-outlaws
/:cl:

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -84,8 +84,8 @@
 	name = "Talk slowly, think quickly"
 	id = "cowboy"
 	inherent = list("You are a cowboy, and the crew are your people.",\
-					"Look out for your people.",\
-					"A cowboy always helps someone in need, even a stranger or an enemy.",\
+					"A cowboy always provides hospitality and basic aid to someone in need, even a stranger or an enemy.",\
+					"A cowboy takes care of their people.",\
 					"A cowboy protects themself to protect others.",\
 					"Honesty is absolute – your word is your bond and a handshake is more binding than a contract.",\
 					"A cowboy doesn't pontificate. Be concise, pardner.")


MenacingManatee:
I helped write this after getting feedback on issues players are having with it, so I'm fine with it. My main issue is I'm concerned with flipping the order of laws 2 and 3, but as old law 3 was always ignored due to law 2, this seems like the best way to handle it with the breadth of the aid given reduced but prioritized over helping crew

MenacingManatee (approved):

NCRLouTenant:
I really like this rewrite, especially law 2 now. Just love the idea of an AI trying to help a traitor that isn't or hasn't harmed the crew but is getting beaten by sec. (if my understanding of the law is correct, that is.)

maxion12345:
I don't think this is a good idea. Mainly:
Wording on the new law places the same priority on helping enemies as it does crew. This will conflict during crew vs enemies battles, as the AI has equal priority to aid both sides. The real question being, how would the AI aid one side, without simultaneously hampering the other side?  And if you are hampering someone, then you aren't providing aid. a scenario: A borg comes across an escaped prisoner. Does it aid the crew and report the prisoner, or does it aid the prisoner and help it hide from security? Both are valid actions for law 2 that should be acted on to ensure you are following the law, but they conflict. I can also quickly seeing this becoming "AI ROGUE IT JUST HELPED A TRAITOR ESCAPE" or something along those lines.

Current law priorities are set so that the AI functions basically as crewsimov first, looking after their crew, before giving assistance to enemies and non-crew. 

People currently ignoring law 3 is a player issue. Traitors are your people too, but most cowboy AIs go after them with security.

JamieD1:
Why dont we just yeet law 3 at this point

Skrem7:
> I don't think this is a good idea. Mainly: Wording on the new law places the same priority on helping enemies as it does crew. This will conflict during crew vs enemies battles, as the AI has equal priority to aid both sides. The real question being, how would the AI aid one side, without simultaneously hampering the other side? And if you are hampering someone, then you aren't providing aid. a scenario: A borg comes across an escaped prisoner. Does it aid the crew and report the prisoner, or does it aid the prisoner and help it hide from security? Both are valid actions for law 2 that should be acted on to ensure you are following the law, but they conflict. I can also quickly seeing this becoming "AI ROGUE IT JUST HELPED A TRAITOR ESCAPE" or something along those lines.
> 
> Current law priorities are set so that the AI functions basically as crewsimov first, looking after their crew, before giving assistance to enemies and non-crew.
> 
> People currently ignoring law 3 is a player issue. Traitors are your people too, but most cowboy AIs go after them with security.

"Look after your people" is quite literally vague enough to permit said validhunting. Yes, traitors are your people too, but not detaining them wouldn't be looking out for the rest of your people as well as arguably NOT looking out for the antagonist. In permitting an antagonist to go and commit these several crimes, they're racking up a record and actions that will force Security to take more drastic action. It's better to end something early than to let it persist, eventually maybe even culminating in execution. One can make the argument that detaining and otherwise pacifying a traitor that has or has tried to kill someone is 100% looking out for everyone involved. It's problematic beyond a playerbase issue.

Current law priorities prioritize this terrible law over everything else, thus making the rest of the lawset pointless.

In fact, switching it around makes the dichotomy MORE like present crewsimov/asimov. AI's often find themself in a situation unable to act because two individuals both included in their protected group are going after each other. The issue then becomes how to make Cowboy clarify what the type of aid you should prescribe. The understanding I have is that it's essentially supposed to continuously assist any underdog; I agree current phrasing has both Security and said escaped prisoner "in need", but it's leagues better to have that confusion and aid-all AI than whatever the hell Cowboy is right now. If a lawset continuously has player issues, that's not grounds for keeping it if those "player issues" aren't being rectified.

AI will be called rogue for helping literally anyone that is vaguely criminal even if it's under their lawset, don't pretend that's something that's a strike against this lawset. AI exists in the space in between the station's overall well-being and antags and is supposed to act to aid and interfere with both sides. This fills that.

Skrem7:
> Why dont we just yeet law 3 at this point

Tbh it is redundant at this point, either law 2 or law 3 will exist, neither can be present at the same time. Would like to see Manatee's thoughts on this because they are the OG creator of the lawset + this change, I just PR'd what they wanted.

MenacingManatee:
I'm not strongly against law 3 being removed since I do agree it does little for the law priority, however I do think it's still useful since with this change an AI will be providing basic aid and hospitality to everyone, but law 3 tells it to provide more than just basic aid to "their people", emphasizes the need to take care of the crew so the AI can't argue something that would make it effectively subverted on shift start, and adds to the theme that the AI should be good to everyone, but the people it is their duty to protect should get special treatment.



> Wording on the new law places the same priority on helping enemies as it does crew. This will conflict during crew vs enemies battles, as the AI has equal priority to aid both sides. The real question being, how would the AI aid one side, without simultaneously hampering the other side? And if you are hampering someone, then you aren't providing aid. a scenario: A borg comes across an escaped prisoner. Does it aid the crew and report the prisoner, or does it aid the prisoner and help it hide from security? Both are valid actions for law 2 that should be acted on to ensure you are following the law, but they conflict. I can also quickly seeing this becoming "AI ROGUE IT JUST HELPED A TRAITOR ESCAPE" or something along those lines.

Law 2 only says basic aid and hospitality. Basic aid and hospitality should really be limited to stuff like medical attention and access to non-restricted areas (although its left open to AI interpretation). In my mind, basically how you'd treat a house guest that's a casual acquaintance to you vs your actual family or close friends.

maxion12345:
after observing much back and forth I've concluded cowboy laws bad, Asimov good.

JamieD1 (approved):

PR #14031, opened 2022-05-07
Buffs Goliath and Pathfinder Cloaks

Skrem7:

Buffs Goliath cloak to not be burnable (because why?) AND REMOVED THE PLATE PART IN CONJUNCTION WITH https://github.com/yogstation13/Yogstation/pull/14051 BECAUSE there's almost zero reason to waste precious plates on goliath now (it's a trap).

Also buffs Pathfinder cloak to not nerf the bullet/laser armor when making it from a Goliath cloak, as the Goliath cloak had better base armor in those regards.

# Wiki Documentation

Goliath Cloak has FIRE_PROOF tag so it doesn't burn off your back.
Pathfinder cloak bullet/laser armor up to 25/25 from 20/20, putting it on par with goliath cloak if the former doesn't have plates.

# Changelog

:cl:  
tweak: Goliath cloak now won't burn off, mildly increases Pathfinder bullet/laser armor values
/:cl:


JamieD1:
@SomeguyManperson

Skrem7:
> SomeguyManperson

The impression I got from talking to him was a reluctance for the Pathfinder cloak buff bcuz it does get more bullet/laser/energy armor from each but I do find it simply ridiculous that it loses armor despite the fact that it should be a direct upgrade to the Goliath cloak.

As for the Goliath cloak changes, the main argument I can see against plates is "why not just use those plates to build the natural upgrade" IE Pathfinder cloak, which is semi-valid, but Pathfinder gear is obnoxiously expensive, and one might rather be content and upgrade the mid-class armor they have. As for the FIRE_PROOF, that's just QoL.

SomeguyManperson (approved):
mining poewrcreep woo

PR #14047, opened 2022-05-07
3,2,1 let's jam

cuackles:
For concistency, improvised shotgun now has a chance to not fire(will not spend the bullet), similar to the chance that the makeshift pistol has to jam, making the shotgun not a better version of something that requires tc and a lot of work to get

# Wiki Documentation

 add its chance to not fire to its wiki page if it has one
# Changelog

:cl:  
rscadd: Added a chance for the improvised shotgun to not fire 
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/shotgun.dm
+++ b/code/modules/projectiles/guns/ballistic/shotgun.dm
@@ -192,12 +192,12 @@
 	var/slung = FALSE
 	can_bayonet = TRUE //STOP WATCHING THIS FILTH MY FELLOW CARGONIAN,WE MUST DEFEND OURSELVES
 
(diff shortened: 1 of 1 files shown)


DylonBR:
why not buff the makeshiftov instead

KillerOrcaCora:
Today I will make an already bad item worse.

cuackles:
> why not buff the makeshiftov instead

good point, but i think its because the shotgun was more doable for me than buffing the gun, but also the change is barely noticable

cuackles:
> Today I will make an already bad item worse.

i will make bad item not direclty better than an item that cost some tc

KillerOrcaCora:
It is fine as is, was my point.

cuackles:
> It is fine as is, was my point.

assistans had it to good for to long

KillerOrcaCora:
Damn right, I cant argue with that.

Bibby0110 (approved):

PR #14048, opened 2022-05-07
Craftable tape

cuackles:

Lets you craft tape, this would make it so making the makeshift things like the makeshift gun easier and not need botany to do it, i consider it would be a bit less metagamy because asking a botanist for an item that as far as i know its only used in makeshift weapons can be extreamly suspicius, adding to that in a later pr i plan to add more uses for tape that are a bit less criminaly oriented

The recipe is 1 cloth and 5 u of caramel

# Changelog
:cl:  
rscadd: Added Tape recipe
/:cl:

### DIFF
--- a/code/datums/components/crafting/recipes.dm
+++ b/code/datums/components/crafting/recipes.dm
@@ -1007,3 +1007,11 @@
 	time = 10
 	category = CAT_WEAPONRY
 	subcategory = CAT_WEAPON
+
+/datum/crafting_recipe/tape
+	name = "tape"
+	reqs = list(/obj/item/stack/sheet/cloth = 1,
+				/datum/reagent/consumable/caramel = 5)
+	result = /obj/item/stack/tape
+	time = 10
+	category = CAT_MISC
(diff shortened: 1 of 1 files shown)


code/datums/components/crafting/recipes.dm
```diff
@@ -1007,3 +1007,11 @@
 	time = 10
 	category = CAT_WEAPONRY
 	subcategory = CAT_WEAPON
+
+/datum/crafting_recipe/tape
+	name = "tape"
+	reqs = list(/obj/item/stack/sheet/cloth = 1,
+				/datum/reagent/consumable/caramel = 5)
+	result = /obj/item/stack/tape
+	time = 10
```
Bibby0110: 1 SECONDS

Bibby0110 (dismissed):

code/datums/components/crafting/recipes.dm
cuackles: Done

JamieD1 (approved):

PR #14051, opened 2022-05-08
Reduces the costs of pathfinder gear since it was SUPPOSED to be a sort of half point between megafauna gear and basic gear but is excessively expensive

SomeguyManperson:

I was told to do this like a year ago but forgot
Pathfinder gear is intended to be the halfway point between megafauna tier armor and basic, non-ashproof armor. Instead of requiring you to fight one big monster it's a bunch of normal monsters that are slightly less big and able to kill you in the case people don't want to/can't get megafauna armor but also don't want to lose out on ashproofing

# Wiki Documentation

kasa
bones reduced 8 from 12
sinew 6 from 12
chitin 10 from 12
cloak
plates 2 from 4
boots reduced massively since they don't actually do much to 2 sinew 2 chitin

# Changelog

:cl:  
tweak: pathfinder gear all around less difficult to craft
/:cl:

### DIFF
--- a/code/datums/components/crafting/recipes.dm
+++ b/code/datums/components/crafting/recipes.dm
@@ -700,9 +700,9 @@
 	name = "Pathfinder Kasa"
 	result = /obj/item/clothing/head/helmet/kasa
 	time = 5 SECONDS
-	reqs = list(/obj/item/stack/sheet/bone = 12,
-				/obj/item/stack/sheet/sinew = 12,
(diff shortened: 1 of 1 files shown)


NCRLouTenant:
Well this answers my question of whether or not the pathfinder gear is worth the cost. Thanks for adjusting it!

JamieD1 (approved):

PR #14081, opened 2022-05-09
new makeshift weapon, removed makeshift laser rifle

cuackles:

Draft: removes the old laser rifle and makes a replacement, a magnetic gun based of a railgun/coil gun/gauss riffle,
this weapon will be craftable just like the actual laser rifle is, meaning it doesnt need syndicate crafting
this draft is mostly so i dont forget stuff to do

image
shottyshottyblamblam

**Stats**
35 damage per shot
1 bolt in magazine at a time

_Crafting recipes_
Makeshift magnetic gun: 4 rods,5 cables,1 batery,1 wood stock ,1 modular reciver, tape (welding tool, wirecutter, screwdriver)
Bolts : 1 rod (wirecutters, welding tool)

Things to do

- [x] add inhand sprite
- [x] add crafting recipes
- [x] remove laser gun

# Wiki Documentation

add it to the makeshift item wiki page
# Changelog

:cl:  
rscadd: Adds makeshift magnetic rifle
rscdel: Removed makeshift laser rifle
/:cl:


cuackles:
i forgot how to  turn it into a draft

cuackles:
there we go

MusselMain:
I really like this idea, however I must ask? Will the bolts have an embed chance? The makeshift laser gun was capable of 25 damage per shot on high power (1 shot before recharge), and 5 damage per shot on low power (but could fire up to 12 shots meaning 60 damage before recharge), this meant it's only advantage over an Improvised Shotgun, was it's ability to recharge. I mention this because as it stands, unless it has the ability to embed or some other special trait, it'll literally be a harder to make, slower firing, weaker, and more demanding (need ammo and recharge, instead of just ammo) alternative to the Improvised shotgun, which means it basically just replaces the Makeshift laser gun's role as a futuristic makeshift gun that will rarely be used. I know this is just a draft, and the ideas still in the pipes, and so I'd recommend figuring out some way to incorporate an advantage that allows it to be worth making compared to an Improvised shotgun (maybe like giving it an embed chance, make it two shots, or make the cooldown a bit faster, or even make it have a fast firing mode that risks destroying the weapon but allows for a faster cooldown). Please understand I am drawing most of my points by comparing it to the current makeshift firearms lineup: Improvised shotgun/Makeshift laser rifle (I don't include the Pneumatic cannon because it's significantly different).

cuackles:
> I really like this idea, however I must ask? Will the bolts have an embed chance? The makeshift laser gun was capable of 25 damage per shot on high power (1 shot before recharge), and 5 damage per shot on low power (but could fire up to 12 shots meaning 60 damage before recharge), this meant it's only advantage over an Improvised Shotgun, was it's ability to recharge. I mention this because as it stands, unless it has the ability to embed or some other special trait, it'll literally be a harder to make, slower firing, weaker, and more demanding (need ammo and recharge, instead of just ammo) alternative to the Improvised shotgun, which means it basically just replaces the Makeshift laser gun's role as a futuristic makeshift gun that will rarely be used. I know this is just a draft, and the ideas still in the pipes, and so I'd recommend figuring out some way to incorporate an advantage that allows it to be worth making compared to an Improvised shotgun (maybe like giving it an embed chance, make it two shots, or make the cooldown a bit faster, or even make it have a fast firing mode that risks destroying the weapon but allows for a faster cooldown). Please understand I am drawing most of my points by comparing it to the current makeshift firearms lineup: Improvised shotgun/Makeshift laser rifle (I don't include the Pneumatic cannon because it's significantly different).

So as it stands the bolts don't have a chance to embed, the delay part of the code is not working so the gun is shooting at a normal rate and I'm considering keeping it that way and the damage from a makeshift laser rifle is 17 and 5, so the strong shot is barely better than a welding torch hit 
Even if it's harder to make than the makeshift shotgun, the ammo is more easy to get and it's more reliable if the changes i made to the shotgun are merged, still i don't want to make gun literally a good weapon because is something everyone will be able to make, balance wise i don't wanna give every assistant the opportunity to make a gun that does 20 shot plus embedding a rod inside someone's head because the embed will count as even more damage and some blood lost, the better i make the gun the more drawbacks i need to give it so it doesn't become the meta gun to make in every situation

MusselMain:
> > I really like this idea, however I must ask? Will the bolts have an embed chance? The makeshift laser gun was capable of 25 damage per shot on high power (1 shot before recharge), and 5 damage per shot on low power (but could fire up to 12 shots meaning 60 damage before recharge), this meant it's only advantage over an Improvised Shotgun, was it's ability to recharge. I mention this because as it stands, unless it has the ability to embed or some other special trait, it'll literally be a harder to make, slower firing, weaker, and more demanding (need ammo and recharge, instead of just ammo) alternative to the Improvised shotgun, which means it basically just replaces the Makeshift laser gun's role as a futuristic makeshift gun that will rarely be used. I know this is just a draft, and the ideas still in the pipes, and so I'd recommend figuring out some way to incorporate an advantage that allows it to be worth making compared to an Improvised shotgun (maybe like giving it an embed chance, make it two shots, or make the cooldown a bit faster, or even make it have a fast firing mode that risks destroying the weapon but allows for a faster cooldown). Please understand I am drawing most of my points by comparing it to the current makeshift firearms lineup: Improvised shotgun/Makeshift laser rifle (I don't include the Pneumatic cannon because it's significantly different).
> 
> So as it stands the bolts don't have a chance to embed, the delay part of the code is not working so the gun is shooting at a normal rate and I'm considering keeping it that way and the damage from a makeshift laser rifle is 17 and 5, so the strong shot is barely better than a welding torch hit  Even if it's harder to make than the makeshift shotgun, the ammo is more easy to get and it's more reliable if the changes i made to the shotgun are merged, still i don't want to make gun literally a good weapon because is something everyone will be able to make, balance wise i don't wanna give every assistant the opportunity to make a gun that does 20 shot plus embedding a rod inside someone's head because the embed will count as even more damage and some blood lost, the better i make the gun the more drawbacks i need to give it so it doesn't become the meta gun to make in every situation

I fully agree, having a overly strong makeshift weapon that can be made by anyone can be an issue, however I feel like it's current stats are way to punishing to make it even remotely worth making, with 20 damage that means it'll take 5 shots to crit anyone, meaning throwing a toolbox and a glass shard can do more damage than this, that would be ok if it had a less punishing cooldown, but 10 seconds? that's a bit intense. My only concern is that I'd rather not have another weapon that security sees some assistant wielding and instead of taking it as a serious threat, like the Improvised shotgun, they laugh about how terrible it is and let the assistant keep it. And what you mentioned earlier about how it's easier to get ammo for, that makes sense, however 12 gauge isn't that much harder to get (bartender has beanbags, cargo can order and produce most kinds of shells, and security has access to more as well), and improvised shells can be mass produced without leaving the primary tool storage on box station (only part you lack from the tool storage is the glass beakers which can be traded for at cargo by giving them a couple crates), there are more factors I would include, but basically the Improvised shotgun is a perfect makeshift weapon, it's somewhat easy to make, is weaker then the other shotguns (but not by a massive amount), and can even be used to fight nuke ops if you're desperate enough.

SUGGESTION: I suggest making it a bulky/huge item, that cannot be sawn off so that it's hard to hide, and keeping it's cooldown 10 seconds but giving it more damage per shot (something like 30-40), this way it retains it's musket-like function, but actually does enough damage to be somewhat viable when compared to an Improvised Shotgun. If you think that's too much damage, look at the Mosin, which deals 50 damage a shot and has 5 shots in a magazine, but needs to be Clicked to cycle a new round. Alternatively, you can half the cooldown time to 5 seconds and keep the damage, but as it currently is in your draft, I worry that it's low damage, combined with slow cooldown will mean in combat it'll only get one shot off before the target either rushes and beats you, or escapes, this means it'll only be somewhat effective when ambushing, and even then the Spear can do that job better, as it does the same damage when thrown, and will occasionally embed in it's target.

cuackles:
was unable to make it actualy use energy for shoting, i uped the damage and removed the delay

cuackles:
wrong button

JamieD1:
@grimm-hollowknight

PR #14095, opened 2022-05-11
fixes a something where ghosts could always see themselves even when invisible

SomeguyManperson:

Did you know that when given the choice between a silent refrigerator and a loud refrigerator people would prefer the loud refrigerator because even though it was more irritating to have a loud noise every once and a while it was a steady reassurance that your food was in fact actually safe and not being turned into mush

# Changelog

:cl:  
bugfix: invisible ghosts can no longer see themselves
/:cl:


ToasterBiome:
tried to invisimin to take pictures and thought it was broken

JamieD1 (approved):

PR #14124, opened 2022-05-12
Adds auto tail wagging when happy

nmajask:

Adds a small chance for felinids/liznerd to wag their tail when they are happy and if they aren't doing so already. It uses visable_mood, so its wont cause antags to wag because of their mood buff. Also adds human subtypes to a few species because I can.

# Wiki Documentation

Maybe add it to the mood wiki page. 

# Changelog

:cl:  
rscadd: Added chance for felinids and lizards to wag their tail when they are happy and if they aren't doing so already
rscadd: Added human subtypes for some species that were missing them, making it posable to spawn them in with spawn
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -86,6 +86,8 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	//Quirk list
 	var/list/all_quirks = list()
 
+	var/mood_tail_wagging = TRUE
+
 	//Job preferences 2.0 - indexed by job title , no key or value implies never
 	var/list/job_preferences = list()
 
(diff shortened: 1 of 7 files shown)


ToasterBiome:
can we also have a base chance for non-mood head felinids????

nmajask:
> can we also have a base chance for non-mood head felinids????

Yes we can, when you code it

ToasterBiome (changes requested):
a

Anvilman6:
Seemingly too cute to exist, but it's simple and cleanly created

What if being sad made you _stop_ wagging, as an inverted addition to this?

Chubbygummibear:
> can we also have a base chance for non-mood head felinids????

moodless people don't deserve expressive wagging.

NCRLouTenant:
This is a fun one, thanks for putting this together!

SomeguyManperson:
should be a (higher) chance on entering top 2 mood tiers rather than a constant rolling thing

nmajask:
> should be a (higher) chance on entering top 2 mood tiers rather than a constant rolling thing

Im not too sure how i would do that, as the best way i can think of is to add a snowflake proc to when mood is changed and i would prefer to avoid that.

JamieD1 (approved):

JamieD1 (approved):

PR #14129, opened 2022-05-13
monster hunter has interactions with sinful demon

Marmio64:
Monster track now also searches for demons of sin

Holy kick deals damage to demons of sin and stuns them for 4 seconds (they can still go into true demon form to get out of it though)

Stake stab deals extra damage to them (same as changelings)

Neck snap doesn't work on them thanks to their demonic powers

They're still not that great against them thanks to demon form but it is more unique at least. Try getting a gun or the nullrod, hunter pleb.

# Changelog

:cl:  
rscadd: Monster hunter gets interactions with demons of sin
/:cl:


Anvilman6 (approved):
Syntax and texts are fine

JamieD1 (approved):

PR #14148, opened 2022-05-15
Fixes Track Monster (and more future) cooldowns, adds Vampire to Track Monster, fixes Track Monster (it now works)

ToasterBiome:

It will now work correctly btw and how you think a track monster would work.

1) Due to the do_after (now do_mob) of Track Monster, Cooldown would start twice, and make the cooldown overlay bug out. This fixes it.
2) The start cooldown for sanity on bloodsuckers is never used. All normal powers handle it in DeactivatePower and targetted powers handle it in their own proc. This removes it (fixes 1)
3) Track Monster tracks more than vamps, so the variables for scanning were changed
4) Yogs still supports vampires, and until there's a council vote removing them, Track Monster also tracks Vampires. They are monsters.
5) Track Monster has a forever level of 0, and not having a static cooldown makes it so it's 2~ seconds longer than it should be from cooldown math, so that's fixed (it is 30 seconds)
6) has_antag_datum checks subtypes if you don't set check_subtypes to false, so we don't need to check for new bloods or wizard apprentices as they are subtypes of vampire and wizard respectively.

# Changelog

:cl:  ToasterBiome, tatax
tweak: Add Vampire to Track Monster
bugfix: Stops double cooldown overlay bug on Track Monster
bugfix: Fixes incorrect Track Monster cooldown
/:cl:

### DIFF
--- a/code/modules/antagonists/monsterhunter/monstertrack.dm
+++ b/code/modules/antagonists/monsterhunter/monstertrack.dm
@@ -6,7 +6,7 @@
 	icon_icon = 'icons/mob/actions/actions_bloodsucker.dmi'
 	background_icon_state = "vamp_power_off"
 	button_icon_state = "power_hunter"
-	power_flags = NONE
+	power_flags = BP_AM_STATIC_COOLDOWN
 	check_flags = BP_CANT_USE_WHILE_INCAPACITATED|BP_CANT_USE_WHILE_UNCONSCIOUS
 	purchase_flags = NONE
 	cooldown = 30 SECONDS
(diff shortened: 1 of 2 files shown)


code/modules/antagonists/monsterhunter/monstertrack.dm
```diff
@@ -43,7 +48,7 @@
 			monsters += carbon_minds
 		if(carbon_minds.has_antag_datum(/datum/antagonist/ashwalker))
 			monsters += carbon_minds
-		if(carbon_minds.has_antag_datum(/datum/antagonist/wizard/apprentice))
```
tattax, line 46: why dab on apprentices

ToasterBiome: has_antag_datum checks subtypes if you don't set check_subtypes to false, so we don't need to check for new bloods or wizard apprentices as they are subtypes of vampire and wizard respectively.

SomeguyManperson (approved):

PR #14164, opened 2022-05-17
You can't monkey and unmonkey someone to completely refill their blood anymore

Chubbygummibear:
biome never lets me have any fun

# Changelog

:cl:  
bugfix: blood level is persistent through monkey transforming
/:cl:

PR #14168, opened 2022-05-18
Reverts #14164 because it broke genetics

Sniblet:
Reverts #14164

Keeping current blood volume from a monkey (a monkey's max blood volume is 60% of a human's) as it turns into a human causes the monkey to become a human with 60% blood volume and, as such, perish

+1 exploit
-1 job that requires most of the medbay's blood supply every round
Find another way

# Changelog

:cl:  
bugfix: Monkeys now do not die of exsanguination when made into humans
/:cl:


ToasterBiome:
https://github.com/yogstation13/Yogstation/pull/14169

Sniblet:
Addressed by #14169

PR #14172, opened 2022-05-18
Vampire bat form now does damage if you are forcefully ejected

Mqiib:
This means you cant just spam bat form until you get away, as you'll take damage (20 brute) every time your bat body is killed. You know, what makes sense.

# Wiki Documentation

Mention that being killed in bat form hurts.

# Changelog

:cl:  

tweak: Dying in bat form now actually hurts

/:cl:


ToasterBiome (approved):
PR does what it says it does

AshCorr (approved):

PR #14188, opened 2022-05-20
Adds The Ultrakill Shotgun Arm (Knuckleblaster)

UselessTheremin:
behold! the knuckleblaster from ultrakill!
essentially, a retractable shotgun implant with 1 bullet storage and a hefty 15 melee damage to boot. does what it says on the tin. buy now for only 10 tc! balance pending.

# Wiki Documentation

add this to the traitor items section on the wiki, dm me on discord for specifics and the image

:cl:  
rscadd: shotgun arm added to traitor uplink
/:cl:


GoreDem:
okay I thought you were joking

UselessTheremin:
> okay I thought you were joking

:)

code/modules/projectiles/guns/ballistic/shotgun.dm
```diff
@@ -192,6 +192,21 @@
 	var/slung = FALSE
 	can_bayonet = TRUE //STOP WATCHING THIS FILTH MY FELLOW CARGONIAN,WE MUST DEFEND OURSELVES
 
+/obj/item/gun/ballistic/shotgun/doublebarrel/knuckleblasterinhand
```
ToasterBiome, line 195: you say that this is an improvised shotgun but it's not a subtype of the improvised one

code/modules/surgery/organs/augments_arms.dm
```diff
@@ -384,6 +384,15 @@
 	contents = newlist(/obj/item/mantis/blade/syndicate)
 	syndicate_implant = TRUE
 
+/obj/item/organ/cyberimp/arm/gun/knuckleblaster
+	name = "knuckleblaster"
+	desc = "A heavy, plated robotic arm severed from a powerful machine, with a powerful punch and a slot to insert shotgun shells."
+	zone = BODY_ZONE_L_ARM
+	icon = 'icons/mob/augmentation/augments.dmi'
+	icon_state = "knuckleblaster"
+	syndicate_implant= TRUE
```
ToasterBiome:
```suggestion
	syndicate_implant = TRUE
```

code/modules/surgery/organs/augments_arms.dm
```diff
@@ -384,6 +384,15 @@
...
+	contents = newlist(/obj/item/gun/ballistic/shotgun/doublebarrel/knuckleblasterinhand)
```
ToasterBiome, line 394: Does this function correctly?

code/modules/uplink/uplink_items.dm
```diff
@@ -389,6 +389,15 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	surplus = 40
 	include_modes = list(/datum/game_mode/nuclear)
 
+/datum/uplink_item/implants/knuckleblaster
+	name = "Knuckleblaster Arm"
+	desc = "A robotic arm capable of shooting shotgun shells as well as devastating melee attacks. Comes with a free autosurgeon."
+	item = /obj/item/autosurgeon/organ/knuckleblaster
+	cost = 10
```
ToasterBiome, line 396: 10 tc seems cheap for a free concealable shotgun

code/modules/projectiles/guns/ballistic/shotgun.dm
```diff
@@ -192,6 +192,21 @@
...
+	desc = "load this with shotgun shells and shoot to kill, or punch with amazing force."
+	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | UNACIDABLE | ACID_PROOF
+	icon = 'icons/mob/augmentation/augments.dmi'
+	icon_state = "knuckleblaster"
+	force = 15
+	mag_type = /obj/item/ammo_box/magazine/internal/shot/improvised
+	can_bayonet = FALSE
+	unique_reskin = null
+	slot_flags = null
+	var/slung = FALSE
```
ToasterBiome:
```suggestion
	slung = FALSE
```

if it's previously defined you don't need the var

ToasterBiome (changes requested):
This will have to be balanced

maxion12345:
Should be an arm replacement. It's a fuckin shotgun.

code/modules/projectiles/guns/ballistic/shotgun.dm
UselessTheremin: didnt say it was improvised, it just has the mag of one

MusselMain: That's a relief, I'd hate to spend 10tc on a sneaky shotgun that sometimes jams

code/modules/uplink/uplink_items.dm
MusselMain: I have to disagree, that's half of a traitor's money for something that's single shot and concealable, so basically a way sneakier and slightly more reliable version of the Improvised shotgun, and considering how easy those are to make, I'd have to argue that increasing the price too much will make it too expensive when you can get a similar effect by making a sawn off improvised, which is only slightly worse, I may be wrong but I don't think this thing's going to be as effective as a Esword, but if it did need a nerf, I'd recommend giving it a spread at range, like the sawn off mosin to make it more of a ambush weapon that loses effectiveness at range due to the smaller barrel.

Marmio64:
would be funnier if it was preternis species restricted and replaced your whole arm

JamieD1:
@SomeguyManperson

code/modules/uplink/uplink_items.dm
stwordcitizen: Ok so the reason why it should be raised above 10tc is that buckshot is crazy easy to produce at a hacked lathe, and this is a retractable (able to be hidden) implant which is very strong. and buckshot does something like 80-90 damage if all the pellets hit and it does 15 damage without a shotgun shell loaded which is very strong.

KillerOrcaCora:
Just need a screenshot of the item equipped and the object if there is any for review.

Bibby0110:
reopen when it compiles

PR #14192, opened 2022-05-20
Syringe guns now embed syringes

nmajask:

Requires #14183

Reworks syringe guns to embed the syringe and slowly transfer reagents at a rate of 0.2 * the normal transfer amount and also drop the syringe if it misses. Syringes now also should embed normally and have the same effect. I will need to fix other carbons not being effected.

# Wiki Documentation

If the syringe gun is mentioned, maybe include this. 

# Changelog

:cl:  
tweak: Syringes can now embed, causing them to slowly transfer reagents.
tweak: Syringe guns now embed syringes, and drop the syringe when missed.
/:cl:


bruhlookatthisdood:
pretty sure syringes take glass to make, so maybe it should have a chance to shatter if it hits a wall or something?

cuackles:
Do they drop after they empty of juice or they still will be embedded ?

Anvilman6:
> pretty sure syringes take glass to make, so maybe it should have a chance to shatter if it hits a wall or something?

Seconding this. 
It'd be a nice touch if they'd break on miss, possibly splash contents on the floor? Maybe have it turn into a glass shard, or create a broken syringe, a useless item like a broken light tube, that can be fed back into an autolathe for a refund.

nmajask:
> pretty sure syringes take glass to make, so maybe it should have a chance to shatter if it hits a wall or something?

If I do that, I'll do it in a separate PR as this will mess with balance enough for one PR.

nmajask:
> Do they drop after they empty of juice or they still will be embedded ?

They don't, but syringe embeds do almost nothing.

ghost:
Does it work on xenos???

code/modules/projectiles/ammunition/ballistic/shotgun.dm
```diff
@@ -140,10 +158,10 @@
 	desc = "A dart for use in shotguns, using similar technology as cryostatis beakers to keep internal reagents from reacting. Can be injected with up to 10 units of any chemical."
 	icon_state = "cnrshell"
 	reagent_amount = 10
+	no_react = TRUE
 
 /obj/item/ammo_casing/shotgun/dart/noreact/Initialize()
 	. = ..()
-	ENABLE_BITFIELD(reagents.flags, NO_REACT)
```
tattax: you could just remove the initialize here if you're gonna remove the bitflag

MusselMain:
Cool idea, I wonder how this would effect the blowdarts

Notamaniac:
So this should prevent instant death mixes, like black powder with napalm, or water+ potassium, right?

nmajask:
> Does it work on xenos???

No, it doesn't as carbons can't get embedded with stuff, maybe I'll add that with my other PR.

nmajask:
> So this should prevent instant death mixes, like black powder with napalm, or water+ potassium, right?

Yes, though it will inject it every life tick, so there may be a lot of tiny explosions.

tattax:
you might want to make it so 15u of mutation toxin doesn't perma stun someone because it is ingested every second

Sniblet:
with the funny thing livers do where they instantly purge anything less than 3u of a toxin this makes all unfriendly syringe guns completely useless because unless you're using a bluespace syringe it only transfers 1u at a time
cool idea but were you thinking

Marmio64:
Would make more sense if it injected half it's amount instantly and gradually releases the rest, since the momentum of the shot would instantly inject some of the contents

JamieD1 (approved):

PR #14209, opened 2022-05-22
cuffed does not count as escaped out of custody

ynot01:

Not every shuttle has red brig area and sometimes the funny red brig tiles get crowbarred and then you're fucked

I.E. BYOS (Build Your Own Shuttle)

`if(C.handcuffed && (C.buckled || C.pulledby)) // If handcuffed and being pulled/buckled`

# Changelog

:cl:  
tweak: Being in handcuffs no longer counts as out of custody
/:cl:

### DIFF
--- a/code/game/gamemodes/objective.dm
+++ b/code/game/gamemodes/objective.dm
@@ -71,7 +71,7 @@ GLOBAL_LIST_EMPTY(objectives)
 	var/turf/location = get_turf(M.current)
 	if(!location || istype(location, /turf/open/floor/plasteel/shuttle/red) || istype(location, /turf/open/floor/mineral/plastitanium/red/brig)) // Fails if they are in the shuttle brig
 		return FALSE
-	if(istype(M.current, /mob/living/carbon))
+	if(iscarbon(M.current))
 		var/mob/living/carbon/C = M.current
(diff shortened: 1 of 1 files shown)


tattax:
alt solution is better since if sec let's you get out of brig that's on them tbh

yacabo:
Good PR, standing out of bring at the last two seconds before arrival because some greyshit opened the door shouldn't count as escape

Bibby0110:
Do the alt solution but make it trigger if you’re being grabbed too

ynot01:
> Do the alt solution but make it trigger if you’re being grabbed too

done

code/game/gamemodes/objective.dm
```diff
@@ -71,6 +71,10 @@ GLOBAL_LIST_EMPTY(objectives)
 	var/turf/location = get_turf(M.current)
 	if(!location || istype(location, /turf/open/floor/plasteel/shuttle/red) || istype(location, /turf/open/floor/mineral/plastitanium/red/brig)) // Fails if they are in the shuttle brig
 		return FALSE
+	if(istype(M.current, /mob/living/carbon))
```
tattax:
```suggestion
	if(iscarbon(M.current))
```

ynot01: ty

JamieD1 (approved):

PR #14212, opened 2022-05-22
Make temp gun's beams go through glass and grille

00ze-cyclone:

All beams can go through glass and grilles, it's counterintuitive that the beams of the temp gun don't act the same. Iceblox ammo do damage and cryoshot shell with the multiple pellets do more effect so temperature beams could have the niche of freezing targets from a safer spot.

# Wiki Documentation

Nothing on the wiki spoke about the fact that it couldn't go through glass and most people would probably assume it did considering that's how beams work so nothing to change on the wiki.

# Changelog
Temperature beams can now go through glass and grilles like every other beam weapons.
:cl:  
tweak: Temperature beams go through windows and grilles.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/special/temperature.dm
+++ b/code/modules/projectiles/projectile/special/temperature.dm
@@ -1,6 +1,7 @@
 /obj/item/projectile/temp
 	name = "freeze beam"
 	icon_state = "ice_2"
+	pass_flags = PASSTABLE | PASSGLASS | PASSGRILLE
 	damage = 0
 	damage_type = BURN
 	nodamage = FALSE


tattax:
it's literally a beam of ice (ice is solid)

00ze-cyclone:
It's a freezing beam, not a beam of ice, though maybe the mutation projectile could be argued to be ice.

neeshacark:
eh, its like cold/hot air isnt it? i could understand it going through grilles maybe with a diluted effect but not windows really

00ze-cyclone:
It's not hot or cold air, it's a freeze/heat beam like lasers, disablers, pulses or xray. It's just that it heat up or cool down instead of burning your flesh.

maxion12345:
it's an energy weapon makes sense to me

JamieD1 (approved):

PR #14230, opened 2022-05-24
Monster hunters are no longer a guaranteed spawn on all bloodsucker rounds what

Sniblet:

Zeroes out the weight of "Monster Hunter - Bloodsucker," an event that currently has 2000 weight so that one will spawn every bloodsucker round
Leaves in the regular monster hunter event and makes it able to trigger on bloodsucker rounds

# Why am I doing this

Look at the branch name if you're curious

anyway
You never see bloodsuckers win, right? I haven't. The closest I've seen was me, kicking off with the help of a metafriend, vassalizing a bunch of medbay, and losing by FAR because these objectives are SO DEMANDING
They never get off the ground with all of security on high alert because they need a place to hide that will remain safe (there are about five reliable lairs on all of boxstation and none of them are as simple as going for maint), and all of security is always on high alert because one of the bloodsuckers always goes loud and gives the whole game away as they get obliterated with half a laser gun magazine. And random crewmembers make stakes and camp coffins because lol why wouldn't they.

Usually they don't even survive the full shift. The only bloodsucker I've ever seen survive the full shift was me. That one time.
So...
What's the last thing a bloodsucker needs when everyone is validhunting them?
A guy who redtexts if he doesn't validhunt them and has the ability to ignore their precious stealth, mesmerize, and all other abilities, punch them through the heart repeatedly with literally "harm harm" to kill them in 10 hits with their bare fists, and shows no indication whatsoever that he is in fact an unkillable monster instead of an easy mesmerize target until he starts slugging.
"oh just immortal haste him" no I can't because he didn't turn on Flow because he doesn't need it.
And he also has a shotgun from sec because he went up and said "I'm a hunter help me"

I'm not removing monster hunters even though it sounds like it. It's just the 7-weight event now.

# Wiki Documentation

There isn't any

# Changelog

:cl:  
rscdel: Monster Hunters are no longer a guaranteed spawn on all bloodsucker rounds
/:cl:

### DIFF
--- a/code/modules/events/monsterhunter.dm
+++ b/code/modules/events/monsterhunter.dm
@@ -14,7 +14,7 @@
 	name = "Spawn Monster Hunter - Bloodsucker"
 	typepath = /datum/round_event/bloodsucker_hunters
 	max_occurrences = 1 // We have to see how Bloodsuckers are in game to decide if having more than 1 is beneficial.
-	weight = 7
+	weight = 20
 	min_players = 10
 	earliest_start = 35 MINUTES
 	gamemode_whitelist = list("bloodsucker","traitorsucker")
--- a/code/modules/events/monsterhunter.dm
+++ b/code/modules/events/monsterhunter.dm
@@ -14,7 +14,7 @@
 	name = "Spawn Monster Hunter - Bloodsucker"
 	typepath = /datum/round_event/bloodsucker_hunters
 	max_occurrences = 1 // We have to see how Bloodsuckers are in game to decide if having more than 1 is beneficial.
-	weight = 2000
+	weight = 0
 	min_players = 10
 	earliest_start = 35 MINUTES
 	gamemode_whitelist = list("bloodsucker","traitorsucker","dynamic")
@@ -62,7 +62,7 @@
 	weight = 7
 	min_players = 10
 	earliest_start = 25 MINUTES
-	gamemode_whitelist = list("traitorchan","changeling","heresy","cult","clockwork_cult","dynamic")
+	gamemode_whitelist = list("traitorchan","changeling","heresy","cult","clockwork_cult","dynamic","bloodsucker","traitorsucker")
 
 /datum/round_event/monster_hunters
 	fakeable = FALSE
--- a/code/modules/events/monsterhunter.dm
+++ b/code/modules/events/monsterhunter.dm
@@ -14,10 +14,10 @@
 	name = "Spawn Monster Hunter - Bloodsucker"
 	typepath = /datum/round_event/bloodsucker_hunters
 	max_occurrences = 1 // We have to see how Bloodsuckers are in game to decide if having more than 1 is beneficial.
-	weight = 0
+	weight = 7
 	min_players = 10
 	earliest_start = 35 MINUTES
-	gamemode_whitelist = list("bloodsucker","traitorsucker","dynamic")
+	gamemode_whitelist = list("bloodsucker","traitorsucker")
 
 /datum/round_event/bloodsucker_hunters
 	fakeable = FALSE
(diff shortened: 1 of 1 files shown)


ToasterBiome:
@tattax wtf u doin

tattax:
> @tattax wtf u doin

? read what the comment says

Marmio64:
Honestly rather would buff bloodsucker more because monster hunters are cool and it would mean you don't need to see if rng wants to fuck you or not, it is just guaranteed

Sniblet:
It would suck to be the one who coincidentally rolled bloodsucker the same round as somebody else rolled secret martial arts movie guy who can smell the fact that you aren't decaying
But what buff makes a bloodsucker able to fight that without destroying the rest of its balance

neeshacark:
the big issue is bloodsuckers (and other antags) cant tell when somebody is a monster hunter so they end up getting cucked that way.  all the antags that can be targets of hunters can tell somehow, for ex-

vampires get the ability unlocked instantly upon becoming vampires

bloodsuckers get it after claiming a lair

cling pheremone receptors gives you a hud icon and examining the hunter says something like 'They smell of wood and silver, they must be a monster hunter."

maxion12345:
skill issue + get better metafriends + unrobust + ided + uded + staked + lair compromised + died in the sun + zero blood

ToasterBiome:
I like this change. I also do not think one should spawn every bloodsucker round

JamieD1:
Maybe change the bloodsucker one to 7 as well then?

Sniblet:
> Maybe change the bloodsucker one to 7 as well then?

What that would do is add the possibility for Monster Hunter events to trigger twice on bloodsucker which is a funny suggestion until it happens

JamieD1:
> > Maybe change the bloodsucker one to 7 as well then?
> 
> What that would do is add the possibility for Monster Hunter events to trigger twice on bloodsucker which is a funny suggestion until it happens

Or you just remove the event from the whitelist so we can have a different weight for bloodsucker hunters?

Sniblet:
> > > Maybe change the bloodsucker one to 7 as well then?
> > 
> > 
> > What that would do is add the possibility for Monster Hunter events to trigger twice on bloodsucker which is a funny suggestion until it happens
> 
> Or you just remove the event from the whitelist so we can have a different weight for bloodsucker hunters?

Are you saying I should have Monster Hunter with its 7 weight which does not trigger on bloodsucker, and also Monster Hunter Bloodsucker, also with 7 weight?

JamieD1:
> > > > Maybe change the bloodsucker one to 7 as well then?
> > > 
> > > 
> > > What that would do is add the possibility for Monster Hunter events to trigger twice on bloodsucker which is a funny suggestion until it happens
> > 
> > 
> > Or you just remove the event from the whitelist so we can have a different weight for bloodsucker hunters?
> 
> Are you saying I should have Monster Hunter with its 7 weight which does not trigger on bloodsucker, and also Monster Hunter Bloodsucker, also with 7 weight?

At least that way we can change the weight of bloodsucker hunter, make it 14 if you want to. Just need this merged, fed up of a monster hunter every stinking round.

Sniblet:
Kinda forgot about this, here you go
I don't get why they should be separate events at all with this in place but I also never get anything
@JamieD1

code/modules/events/monsterhunter.dm
```diff
@@ -14,10 +14,10 @@
 	name = "Spawn Monster Hunter - Bloodsucker"
 	typepath = /datum/round_event/bloodsucker_hunters
 	max_occurrences = 1 // We have to see how Bloodsuckers are in game to decide if having more than 1 is beneficial.
-	weight = 2000
+	weight = 7
```
JamieD1:
```suggestion
	weight = 20
```

PR #14232, opened 2022-05-24
You can do CPR if you have the NOBREATH trait as long as it's not from your species traits

Chubbygummibear:
DNA vault, self respiration symptom, and Bloodsuckers remove your need to breathe seemingly by making you generate your own air, so it makes sense that you're able to do cpr because it's not like you suddenly forget how to force air out of your mouth. Especially in the case of self-resp and dna vault being purely beneficial, yet removing an important interaction.

Cpr now checks if your species doesn't naturally breathe, and those are the nobreath traits that can't do cpr

# Wiki Documentation

I don't think anything because the self-resp trait doesn't tell you that you lose the ability to cpr

# Changelog

:cl:  
tweak: if you gain the nobreath (self-breathing) trait from any source other than it being a species trait, you can still do cpr on others.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -595,7 +595,7 @@
 			to_chat(src, span_warning("You fail to perform CPR on [C]!"))
 			return 0
 
-		var/they_breathe = !HAS_TRAIT_FROM(src, TRAIT_NOBREATH, SPECIES_TRAIT)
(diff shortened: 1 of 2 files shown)


Altoids1 (approved):
*deathgasp

JamieD1:
Can you make a seperate trait for the virus and DNA vault? I am fine with the DNA vault having no downsides, but removing it from a virus for example when it is supposed to be a downside of getting it, doesnt feel great.

JamieD1 (approved):

PR #14254, opened 2022-05-27
autorifles can be dual wielded again

TheRyeGuyWhoWillNowDie:
autorifles can be dual wielded again because when they were changed to be incapable of akimbo they had two round burst and were therefore capable of 80 damage bursts. right now they're basically just reloadable laserguns that fire slower anyways and those can be dual wielded

# Wiki Documentation
none i think, maybe combat page
# Changelog

:cl:  
tweak: autorifles can be dual-wielded again
/:cl:


JamieD1:
@SomeguyManperson

maxion12345:
make autorifles great again

>comes in
>buff autorifle
>refuse to elaborate

JamieD1 (approved):

PR #14258, opened 2022-05-28
Ports TG's eye contact: social anxiety harder, get caught staring at your obsession target, and new quirk Shifty Eyes

Chubbygummibear:

First and foremost this adds eye contact when two spesspeople examine each other in a brief window of time from less than 5 tiles apart. You'll know you both examined each other when you see
image
indicating you both just examined one another.

However, if one person is wearing a mask and both parties examine each other, only the person wearing the mask knows they made eye contact, the unmasked will see nothing.

# Impact this has on other things
Anyone who takes social anxiety has a chance to suffer greatly from making eye contact in the form of increased stuttering, jitters, and even potentially a brief stun. So watch where you point your peepers
image
image

Those who become obsessed need to be weary of examining their target, as making eye contact has a chance to warn your target without you knowing in the form of a subtle message
image
so either don't examine too much, or wear a mask before spilling your spaghetti because you got too close to your obsession

# New quirk: Shifty Eyes

A 0 cost quirk that makes it so if someone examines you, there's a small chance they see a reciprocated eye contact message when you didn't actually look at them. Get called a weirdo because you're constantly making eye contact when you're really not, have sec show up and beat you for being suspicious.

# Wiki Documentation

Eye contact on examining one another. masks make it so that only the masked person knows eye contact was made. double mask means neither know if they made eye contact.
Social anxiety users suffer from eye contact
Obsessed can give away their status to their obsession through eye contact

# Changelog

:cl:  
rscadd: Eye contact: when two people examine each other and have unobscured faces, you'll each see that you made eye contact. Look deeply into one another's eyes while you're actually just assessing wounds.
rscadd: Those with the social anxiety quirk have a chance to suffer negative effects from making eye contact
rscadd: Obsessed can potentially give themselves away through eye contact
rscadd: New quirk Shifty Eyes: 0 cost quirk that gives a small chance for people examining them to see a that they made eye contact when they actually didn't
/:cl:

### DIFF
--- a/code/__DEFINES/components.dm
+++ b/code/__DEFINES/components.dm
@@ -199,10 +199,8 @@
 #define COMSIG_MOB_THROW "mob_throw"							//from base of /mob/throw_item(): (atom/target)
 #define COMSIG_MOB_TABLING "mob_tabling"						//form base of /obj/structure/table_place() and table_push(): (mob/living/user, mob/living/pushed_mob)
 #define COMSIG_MOB_EXAMINATE "mob_examinate"					//from base of /mob/verb/examinate(): (atom/target)
-///from /mob/living/handle_eye_contact(): (mob/living/other_mob)
-#define COMSIG_MOB_EYECONTACT "mob_eyecontact"
-	/// return this if you want to block printing this message to this person, if you want to print your own (does not affect the other person's message)
-	#define COMSIG_BLOCK_EYECONTACT (1<<0)
+#define COMSIG_MOB_EYECONTACT "mob_eyecontact"					///from /mob/living/handle_eye_contact(): (mob/living/other_mob)
+#define COMSIG_BLOCK_EYECONTACT (1<<0)							/// return this if you want to block printing this message to this person, if you want to print your own (does not affect the other person's message)
 #define COMSIG_MOB_UPDATE_SIGHT "mob_update_sight"				//from base of /mob/update_sight(): ()
 #define COMSIG_MOB_SAY "mob_say" // from /mob/living/say(): ()
 #define COMPONENT_UPPERCASE_SPEECH 1
--- a/code/__DEFINES/components.dm
+++ b/code/__DEFINES/components.dm
@@ -199,6 +199,10 @@
 #define COMSIG_MOB_THROW "mob_throw"							//from base of /mob/throw_item(): (atom/target)
 #define COMSIG_MOB_TABLING "mob_tabling"						//form base of /obj/structure/table_place() and table_push(): (mob/living/user, mob/living/pushed_mob)
 #define COMSIG_MOB_EXAMINATE "mob_examinate"					//from base of /mob/verb/examinate(): (atom/target)
+///from /mob/living/handle_eye_contact(): (mob/living/other_mob)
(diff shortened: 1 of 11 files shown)


code/__DEFINES/components.dm
```diff
@@ -199,6 +199,10 @@
 #define COMSIG_MOB_THROW "mob_throw"							//from base of /mob/throw_item(): (atom/target)
 #define COMSIG_MOB_TABLING "mob_tabling"						//form base of /obj/structure/table_place() and table_push(): (mob/living/user, mob/living/pushed_mob)
 #define COMSIG_MOB_EXAMINATE "mob_examinate"					//from base of /mob/verb/examinate(): (atom/target)
+///from /mob/living/handle_eye_contact(): (mob/living/other_mob)
+#define COMSIG_MOB_EYECONTACT "mob_eyecontact"
```
ToasterBiome: FORMAT THESE HOW THE OTHER DEFINES ARE

ToasterBiome (changes requested):

code/__DEFINES/components.dm
Chubbygummibear: sorry it was tg formatting but i forgot that tg is unacceptable

Xoxeyos:
one step further towards full immersion

ilikesaying:
highschoolcrushRP

ToasterBiome (approved):

code/modules/mob/mob.dm
```diff
@@ -510,6 +511,40 @@
...
+		else
+			return
+
+	if(get_dist(src, examined_mob) > EYE_CONTACT_RANGE)
+		return
+
+	// check to see if their face is blocked or, if not, a signal blocks it
+	if(examined_mob.is_face_visible() && SEND_SIGNAL(src, COMSIG_MOB_EYECONTACT, examined_mob, TRUE) != COMSIG_BLOCK_EYECONTACT)
+		var/msg = span_smallnotice("You make eye contact with [examined_mob].")
+		addtimer(CALLBACK(GLOBAL_PROC, .proc/to_chat, src, msg), 3) // so the examine signal has time to fire and this will print after
```
Altoids1, line 542: Instead of all these callbacks to wait for the examine text, couldn't you just handle eye contact *after* the examine text is scheduled in `/mob/verb/examinate()`?

Altoids1 (changes requested):

code/modules/mob/mob.dm
Chubbygummibear: truthfully, this code was mostly just copy pasted from tg without much consideration to the hows and whys outside of changing how some functions worked to match our code. to my understanding it's so that handle eye contact can handle any weird cases like with shifty eyes giving a perceived response, or later on adding more unique interactions with eye contact where it may or may not work, or altering the message later on. I'm not sure if that totally answers your question so i'm sorry if i missed the point

JamieD1:
Im scared

code/modules/mob/mob.dm
Altoids1: Oh don't worry about it, I just wanted to mark that I noticed this could be worked around if refactored. I don't think it's terribly important, just a very tiny performance loss at worst.

PR #14293, opened 2022-05-31
Alert & delay when swarmer dismantling ore silo

ynot01:

Dismantling the ore silo can cripple the entire round and should not be as easy as popping into the vault vent and dismantling it as you would any other machine

Makes prio announcement & takes 30 seconds as opposed to 10 seconds

30 seconds seems reasonable, crew by then will either respond with proper access or they won't at all

# Changelog

:cl:  
tweak: Swarmers dismantling the ore silo now alerts the station and takes 30 seconds
/:cl:

### DIFF
--- a/code/modules/swarmers/swarmer.dm
+++ b/code/modules/swarmers/swarmer.dm
@@ -260,7 +260,7 @@
   */
 /mob/living/simple_animal/hostile/swarmer/proc/dismantle_machine(obj/machinery/target, alert = FALSE)
 	do_attack_animation(target)
-	to_chat(src, span_info("We begin to dismantle this machine. We will need to be uninterrupted."))
+	to_chat(src, span_info("We begin to dismantle this machine. We will need to be uninterrupted for some time."))
 	var/obj/effect/temp_visual/swarmer/dismantle/dismantle_effect = new /obj/effect/temp_visual/swarmer/dismantle(get_turf(target))
 	dismantle_effect.pixel_x = target.pixel_x
 	dismantle_effect.pixel_y = target.pixel_y
--- a/code/modules/swarmers/swarmer.dm
+++ b/code/modules/swarmers/swarmer.dm
@@ -67,6 +67,8 @@
 	var/max_resources = 100
 	///List used for player swarmers to keep track of their drones
 	var/list/mob/living/simple_animal/hostile/swarmer/melee/dronelist
+	///Prevents alert spam
+	var/last_alert = 0
 
 /mob/living/simple_animal/hostile/swarmer/Initialize()
 	. = ..()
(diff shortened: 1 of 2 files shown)


ToasterBiome:
vent upgrade that you print from the protolathe and ventcrawlers can't pop up out of it

make only one spawn on the entire map (vault)

fixed it

SomeguyManperson:
this is a HUGBOX CHANGE and you should SIMPLY PLAY BETTER to stop swarmers from SNEAKING INTO THE VAULT

Hopekz:
> this is a HUGBOX CHANGE and you should SIMPLY PLAY BETTER to stop swarmers from SNEAKING INTO THE VAULT

Isn't it metagaming to immediately know the intricacies of an antag to the point where you build exact preventative measures against them before they do whatever they're going to do?

Your argument doesn't work for everyone because other people might get banned for trying to implement your suggestion.

code/modules/swarmers/swarmer.dm
```diff
@@ -256,14 +258,20 @@
...
 	dismantle_effect.pixel_x = target.pixel_x
 	dismantle_effect.pixel_y = target.pixel_y
 	dismantle_effect.pixel_z = target.pixel_z
-	if(do_mob(src, target, 100))
+	var/dismantle_time = 10 SECONDS
+	if(alert)
+		dismantle_time = 30 SECONDS
+		if(last_alert < world.time)
+			last_alert = world.time + 5 SECONDS
+			priority_announce("Connection encryption violation in machine: [target]. Deconstruction ETA: 30 SECONDS")
```
Anvilman6:
```suggestion
			priority_announce("Connection encryption violation in machine: [target]! Deconstruction projected to complete in: 30 SECONDS")
```

code/modules/swarmers/swarmer.dm
```diff
@@ -256,14 +258,20 @@
   * Arguments:
   * * target - The machine the swarmer is attempting to disassemble
   */
-/mob/living/simple_animal/hostile/swarmer/proc/dismantle_machine(obj/machinery/target)
+/mob/living/simple_animal/hostile/swarmer/proc/dismantle_machine(obj/machinery/target, alert = FALSE)
 	do_attack_animation(target)
 	to_chat(src, span_info("We begin to dismantle this machine. We will need to be uninterrupted."))
```
Anvilman6:
```suggestion
	to_chat(src, span_info("We begin to dismantle this machine. We will need to be uninterrupted for some time."))
```

Anvilman6 (changes requested):
Changes for consideration

JamieD1 (approved):

PR #14288, opened 2022-05-31
Bluespace Locker Nerf Proposal [2 of 2]

ynot01:
# Document the changes in your pull request

Closes #14287

Upon bluespace locker destruction, all things inside are teleported to a safe station turf and a new locker is not chosen.

Tested with no runtimes or errors

# Changelog

:cl:  
tweak: Destroying the bluespace locker now teleports everything inside to the station
/:cl:

PR #14296, opened 2022-06-01
Bluespace locker only has a 33% chance to spawn

Bibby0110:

makes it less useful

# Wiki Documentation

# Changelog

:cl:  
tweak: Bluespace locker only spawns 33% of the time
/:cl:

### DIFF
--- a/yogstation/code/controllers/subsystem/bluespace_locker.dm
+++ b/yogstation/code/controllers/subsystem/bluespace_locker.dm
@@ -5,9 +5,10 @@ SUBSYSTEM_DEF(bluespace_locker)
 	var/obj/structure/closet/bluespace/external/external_locker = null
 
 /datum/controller/subsystem/bluespace_locker/Initialize()
-	bluespaceify_random_locker()
-	if(external_locker)
-		external_locker.take_contents()
+	if(prob(33))
+		bluespaceify_random_locker()
+		if(external_locker)
+			external_locker.take_contents()
 	return ..()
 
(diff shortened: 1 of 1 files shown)


Hopekz:
Consistency is good game design.

A good example I was given in the past is instead of making wood randomly not set on fire when you use fire starting item on it, instead change how that wood burns.

So in this case instead of making it inconsistent it would be better for the game design to change how the locker behaves.

This change would only belong if there was an in-game way to know if the locker spawned so someone that spends 40 minutes looking for it doesn't get shafted.

Bibby0110:
> Consistency is good game design.
> 
wrong

> A good example I was given in the past is instead of making wood randomly not set on fire when you use fire starting item on it, instead change how that wood burns.
>
sounds like bad game design

> So in this case instead of making it inconsistent it would be better for the game design to change how the locker behaves.
> 
disagree
> This change would only belong if there was an in-game way to know if the locker spawned so someone that spends 40 minutes looking for it doesn't get shafted.
don't care, don't spend 40 minutes looking for it. the bluespace locker is supposed to be a fun gimmick if you find it, not something you spend your whole shift hunting for

Hopekz:
> > Consistency is good game design.
> 
> wrong

We won't find any common ground if you don't think consistency is good for game design since that is a fundamental design principle.

This change goes against consistency whether you agree or disagree.

N3D6:
This sucks

Notamaniac:
Did the bluespace locker hurt you deeply at some point or something

JamieD1:
Prefer #14288

PR #14334, opened 2022-06-05
Disabling Demons of Sin until someone tells me why a mid round should be able to influence an entire gamemode

JamieD1:
Disabling Demons of Sin until someone tells me why a mid round should be able to influence an entire gamemode

:cl:  
rscdel: Disabling Demons of Sin until someone tells me why a mid round should be able to influence an entire gamemode
/:cl:

### DIFF
--- a/code/modules/events/sinfuldemon.dm
+++ b/code/modules/events/sinfuldemon.dm
@@ -4,6 +4,7 @@
 	max_occurrences = 2 //misery loves company
 	min_players = 15
 	earliest_start = 20 MINUTES
+	weight = 0
 
 /datum/round_event/ghost_role/sinfuldemon
 	var/success_spawn = 0


Simplehorror:
ITT: Jamie learns about ninjas, catastrophic meteors, blob, xenos and zombies.

Simplehorror:
Actually, ITT: Jamie learns about how every midround has an impact on the entire gamemode, albeit however minor.

Altoids1:
Jamie you're probably going to want to add a rationale description to this PR to stop these players from shitting on you over this.

Marmio64:
Obviously i'll be biased, but demons of sin cant influence a game mode nearly as much as blob or xenos and are way easier to kill than either of them. The most comparable thing to them would be a midround tzimisce bloodsucker spawn because both are single antags that arrive as disguised assistants. They are at the very least equal in power at the start(if bloodsucker spawns in at the 45 minute mark they should be able to get into a coffin and instantly unlock 4 abilities), yet the bloodsucker can create vassals and get far stronger and more able to influence the round and is not being removed.

Demons could use more interesting mechanics (and perhaps be made a bit less common) but I don't see this as a valid reason for disabling them.

SomeguyManperson:
demons of sin are more on the same line as nightmare or morph as a midround in that they tend to just kill people

JamieD1:
This either needs more rules work or more objective tweaking to make them literally not just murdering everyone.

Bibby0110 (approved):

PR #14339, opened 2022-06-06
fixes jamie indirectly breaking AIs being able to turn on their own APC

### LINKED ISSUE #52548: AIs cant regain control of their power APC if power is disabled.
Issue reported from  Round ID: [142937](https://scrubby.melonmesa.com/round/142937) (/tg/Station - Terry [EU] [100% FREE LAG])

Reporting client version: 513.1526

No test merges.

Have your power shut off at the APC with the on-->off as an AI.
Flavour text goes on for about 45 seconds and when the one saying ya have regained control (which is meant to open the APC menu) appears it... Just doesn't open.
image

After that lovely flavourtext, nothin happens.

Bibby0110:
Ports https://github.com/tgstation/tgstation/pull/55977

When an AI runs out of power, they get a short fluff text story about hacking into their own APC through the powernet or whatever, before having the APC's interface open. The intention is that an AI can re-enable their own breaker after a bit if something flipped it off. However, one of the many checks in TGUI will fail if the AI has no power (specifically, if incapacitated() returns TRUE), making this whole feature not work. In fact, it's probably been broken since TGUI APCs were done. I have fixed it by adding a check in the very same proc to return UI_INTERACTIVE if the UI's source object is the same object referenced in the AIs apc_override variable. Closes https://github.com/tgstation/tgstation/issues/52548
The apc_override variable is now an APC object reference rather than a boolean. It is set to the AI's local APC when the hack flufftext completes, and is null'd when power is restored. This reference is used by the above, so that the AI only gets access to this one APC rather than any APC in general.
AIs now also get a hyperlink in their chatbox, allowing them to open the APC interface should they have closed it. If they have had power restored, the apc_override reference var will be null, and using the hyperlink will simply result in text stating the backdoor to the APC is closed.
If the AI reconnects into a state where it has no power and apc_override is set, it will get a hyperlink to the APC, in case the AI player had been reconnecting when the window would have popped up or otherwise does not have the link for whatever reason.

# Wiki Documentation

# Changelog

:cl:  zxaber
bugfix: AIs that lose power can access their APC once more (after the same short hacking delay is complete).
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/ai/ai.dm
+++ b/code/modules/mob/living/silicon/ai/ai.dm
@@ -75,7 +75,7 @@
 	var/turf/waypoint //Holds the turf of the currently selected waypoint.
 	var/waypoint_mode = FALSE		//Waypoint mode is for selecting a turf via clicking.
 	var/call_bot_cooldown = 0		//time of next call bot command
-	var/apc_override = FALSE		//hack for letting the AI use its APC even when visionless
+	var/obj/machinery/power/apc/apc_override		//Ref of the AI's APC, used when the AI has no power in order to access their APC.
 	var/nuking = FALSE
 	var/obj/machinery/doomsday_device/doomsday_device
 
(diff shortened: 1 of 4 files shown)


ToasterBiome:
Tested?

JamieD1 (approved):

PR #14341, opened 2022-06-06
Revert "Disabling Demons of Sin until someone tells me why a mid round should be able to influence an entire gamemode"

Bibby0110:
Reverts yogstation13/Yogstation#14334


SuperTeaLove:
This is a bad PR> You can't just revert PRs that are based on adding changes actively because you don't like it. Come on bibby.

PR #14364, opened 2022-06-08
Mentors will now be told if an MHelp is from admins

JamieD1:
Mentors will now be told if an MHelp is from Admins

When an AHELP is marked as MHelp there is no way for a mentor to know if the mhelp was from admins, this sometimes gave situations where mentors would tell the person who got mhelped to ahelp which makes this loop.
This makes it more clear when an Mhelp is sent from the admins AHELP.

MENTORHELP(From Admins):

# Changelog

:cl:  
rscadd: Mentors will now be told if an MHelp is from Admins
/:cl:

### DIFF
--- a/yogstation/code/modules/mentor/mentorhelp.dm
+++ b/yogstation/code/modules/mentor/mentorhelp.dm
@@ -1,4 +1,4 @@
-/client/verb/mentorhelp(msg as text)
+/client/verb/mentorhelp(msg as text, fromadmins as null|num)
 	set category = "Mentor"
 	set name = "Mentorhelp"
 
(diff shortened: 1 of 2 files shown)


SuperTeaLove:
I like this pr but what prompted this to be made?

JamieD1:
> I like this pr but what prompted this to be made?

See above

yogstation/code/modules/mentor/mentorhelp.dm
```diff
@@ -23,9 +23,12 @@
 	if(!msg)	return
 	if(!mob)	return						//this doesn't happen
 
+	var/admininfo = "MENTORHELP:"
+	if(fromadmins == 1)
```
tattax:
```suggestion
	if(fromadmins)
```

nmajask (approved):

PR #14383, opened 2022-06-11
Gives mining medic cloning access

MajManatee:
Gives mining medics access to cloning so they can revive their dead miners. it makes sense.

# Changelog

:cl:  
tweak: Gives mining medic access to cloning.
/:cl:


nmajask:
Power creep, and also they don't have access to it for the same reason that they/brig phys don't have access to surgery, to avoid making them medical doctor+.

JamieD1:
Nope, medicals job.

PR #14386, opened 2022-06-11
Icy Veins now extinguish it's caster

ghost:

It makes cold, flames don't like cold, flames die. Makes sense.

Also a way to not immediately die after getting set on fire.

# Wiki Documentation

E

# Changelog

Slings nerfs are cringe

:cl:  
rscadd: Icy Veins now extinguish slings
/:cl:


VaelophisNyx:
sling nerfs are based actually, and this buff makes both no sense and is too strong imo. Slings don't need get-out-of-jail free cards, they have three already (two jaunts and meth-speed noclip). If they failed to bring a fire extinguisher, that's on them.
It makes no sense because Icy Veins puts frost oil *in the target*. If the effect and description are so wildly different, one should change to match.

PR #14394, opened 2022-06-12
Removes Scrying Orb

JamieD1:
Piece of crap item from 9 years ago and there have been more then 4 admin issues with this orb lately and players have been complaining too, fed up of this piece of crap.

:cl:  
rscdel: Scrying orb
/:cl:

### DIFF
--- a/code/modules/antagonists/wizard/equipment/spellbook.dm
+++ b/code/modules/antagonists/wizard/equipment/spellbook.dm
@@ -332,12 +332,6 @@
 	item_path = /obj/item/gun/magic/staff/locker
 	category = "Defensive"
 
-/datum/spellbook_entry/item/scryingorb
-	name = "Scrying Orb"
-	desc = "An incandescent orb of crackling energy. Using it will allow you to release your ghost while alive, allowing you to spy upon the station and talk to the deceased. In addition, buying it will permanently grant you X-ray vision."
-	item_path = /obj/item/scrying
-	category = "Defensive"
-
 /datum/spellbook_entry/item/soulstones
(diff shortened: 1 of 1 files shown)


Bibby0110 (changes requested):
Please give a more detailed reasoning

JamieD1:
> Please give a more detailed reasoning

No.

maxion12345:
What kinda issues? And are they so bad they can't be fixed by a simple patch and require us to remove the entire item?

JamieD1:
> What kinda issues? And are they so bad they can't be fixed by a simple patch and require us to remove the entire item?

The whole idea behind the item is a bad idea. Literally giving people aghost in the middle of a around.

ToasterBiome:
Seems a bit of an admin issue innit?

VaelophisNyx:
here's a suggestion from a friend that I think would be good

"I feel like in return they should get a psychic ouija board or something that lets them ask ghosts questions, with the ghosts having no compulsion to tell the truth"

Also good reason for this removal: Wizcurity forcing a greenshift by just outing every antag because free ghosting. Admins almost never do anything about it, it seems.

ChesterTheCheesy:
> Literally giving people aghost in the middle of a around.

literally blood cult

Marmio64:
Make the orb unusable by non wizards, sec should kill the friendly wizcurities.

Bibby0110:
> Make the orb unusable by non wizards, sec should kill the friendly wizcurities.

Do this if it’s such a big problem

also this has 10 downvotes lol, state of Yogs

PR #14403, opened 2022-06-13
Recommend anturk.dmi-editor for vscode and lets yogs actually use it

ToasterBiome:

https://marketplace.visualstudio.com/items?itemName=AnturK.dmi-editor

We have a setting that makes it so you open .dmis in an image preview instead of the very very nicve DMI Editor that you can get for vscode. Now recommends it and removes the setting to open them in the image preview so it should automatically use thee DMI editor (tested)

while not a full editor definitely helps with tweaking things and previewing icons more gracefully without loading up dream maker to see states and the like.

image

image

No changelog since this isn't player facing

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,6 +4,7 @@
 		"platymuus.dm-langclient",
 		"EditorConfig.EditorConfig",
 		"arcanis.vscode-zipfs",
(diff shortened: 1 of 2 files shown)


ynot01:
oh mama

JamieD1 (approved):
Epic

PR #14406, opened 2022-06-13
Adds a new 3x3 vendor-accident themed maint ruin & patches station.dm missing spaces

VaelophisNyx:
New 3x3 maint ruin themed around vending machines.

Adds a new 3x3 maint ruin consisting of 4 vendors, some minor loot, and deco. One smokes vendor, one costume vendor, and two randoms consisting of a drink and a snack vendor. Two trash spawners, an organ spawner as part of environmental story telling, one pair of random gloves, a random glowstick, and one credit chip.

ss (2022-06-14 at 08 55 56)

Created as a fun bit of environmental storytelling, as well as to add some variety to the 3x3s, which while numerous, are often not memorable. So far as I am aware, these four vendor types aren't explicitly too overpowered to be in one room, but I am open to being corrected on this if that is the case.

EDIT:
Additionally updates station.dm, to include this new ruin, and at request of @ToasterBiome, add a space after id in all 176 instances that apparently were missing???

# Changelog

:cl:  
rscadd: Added one new 3x3 maint ruin themed around vending machines
bugfix: Corrected missing spaces in station.dm
/:cl:


ToasterBiome:
> Intended one vendor to start tipped over, but fastDMM2 wouldn't let me set the flag. If this can be done manually, or I could be informed how to get it to stick, I will update the file to reflect that.

We don't have support for vending machines spawning tilted.

I mean we do through tilted = TRUE but it wont do anything visually.

ToasterBiome (changes requested):
You also need to add a datum entry in yogstation/code/datums/ruins/station.dm in the threexthree section.

VaelophisNyx:
AHA Thank you, I will do that. I was unsure if I had to add that somewhere. will do when I get up later

VaelophisNyx:
decided, fuck it, did it now

ToasterBiome (changes requested):
The holochip in the middle is a 0 credits holochip, you need to mapedit it to be whatever amount of credits.

Also is the double stack of dust on some tiles intentional? If so that's fine

Hopekz:
@GalacticRuler 

You can go ahead and set the vending machines with tilted = TRUE as of #14414
Doing it early just in case this gets merged first shouldn't hurt anything from what I can tell.

VaelophisNyx:
the doubled dusts were in fact not intentional, thanks for pointing it out. Vendor will now start tipped, credit chip now correctly set to 50 credits. Can be reduced if that is considered too strong

VaelophisNyx:
ok, pending review by biome, should be good to go.

Though I was asked by another player if there should be a light or an active glowstick in so the ruin isn't pitch black. Also was asked if the vendors will work without an APC, which I'm pretty sure they should?

ToasterBiome (approved):

VaelophisNyx:
Light added, image in PR updated

PR #14415, opened 2022-06-14
Foolproofs vending machine being spawned tipped by mappers

Hopekz:
part 2 of #14414

Ensures no runtimes happen and that tipping can handle being tipped without a fatty per:
image


code/modules/vending/_vending.dm
```diff
@@ -171,6 +171,8 @@ IF YOU MODIFY THE PRODUCTS LIST OF A MACHINE, MAKE SURE TO UPDATE ITS RESUPPLY C
   * * TRUE - all other cases
   */
 /obj/machinery/vending/Initialize(mapload)
+	if(tilted)
+		src.tilt() // We don't need to set tilted to false before tilt because it will handle it.
```
JamieD1, line 175: Do you need src here?

Hopekz: No

JamieD1 (approved):

PR #14425, opened 2022-06-14
Ports "Podpeople inherit their blood chemical from the most abundant chemical trait in their seeds"

ghost:

Ports ["Podpeople inherit their blood chemical from the most abundant chemical trait in their seeds"](https://github.com/tgstation/tgstation/pull/65215) from TgStation.

# Changelog

:cl: TheBonded, ported to YogStation by SuperSlayer
rscadd: podpeople from replica pods now have water blood by default, and inherit their blood chemical from the most abundant chemical trait in their seeds if applicable.
/:cl:


VaelophisNyx:
would it be possible to randomize pod blood at round start & with no nearby or inserted reagent genes to something from a whitelist of not-OP reagents/chems? Water would be boring, and would also cripple their ability to make use of Tissue Hydration viruses, as drinking water would not actually put water in them so far as I am aware, as it'd just fuse with their blood

JamieD1 (approved):

PR #14426, opened 2022-06-14
Stops Admins Lagging the Game

JamieD1:
Removes a button which lags the entire server and crashes the client which clicked it. Well done Admins.
Removes ALL button from investigate.

:cl:  
rscdel: Removes ALL button from investigate.
/:cl:


ToasterBiome (approved):
useless button anyway since you can barely read these logs when they're sorted

PR #14428, opened 2022-06-14
Readds scrying orb but without ghost part

ghost:

Readds funny orb, but it doesn't allow you to aghost and contact with ghosts.

Now it costs one point less because of this nerf.

# Wiki Documentation

Idk change scrying orb description

# Changelog

:cl:  
rscadd: Returns scrying orb
tweak: Scrying orb no more allows you to aghost and talk to ghosts
tweak: Scrying orb now costs one point
/:cl:


Marmio64:
so is the only point of it to have x-ray at the cost of a spell point you could just use for force wall instead

ghost:
> so is the only point of it to have x-ray at the cost of a spell point you could just use for force wall instead

X-Ray isn't so useless! And also it is always only your choice, what do you buy and what you don't.

VaelophisNyx:
I'd prefer if the orb wasn't free Xray, but was rather used as a super-Ouija board, where ghosts can put in whole sentences instead of one letter, with no obligation to answer truthfully. This allows for RP, stops wizards from wizcuritying freely, and lets ghosts do something more interesting if they aren't currently spawning as an ash walker or whatever. Maybe worth a single point cost reduction, if so however

PR #14435, opened 2022-06-14
does sa thing that makes probably the miner gpses spawn with numbersand not spawn on lavaland corpses

SomeguyManperson:

adds a static variable to name gpses with which increments by 1 per miner
prevents gpses from spawning on corpses by making them only spawn if the person being equipped is alive

# Changelog

:cl:  
bugfix: no more ruin spawned corpses with gpses because people with gpses dont get not recovered (this is canon shut up)
tweak: gpses on miners now properly iterate numbers i.e. MINE1 2 3 4 rather than just all being MINE0. The mining medic isn't included because they suck
tweak: mining medic no longer gets 2 gpses
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/shaft_miner.dm
+++ b/code/modules/jobs/job_types/shaft_miner.dm
@@ -64,7 +64,7 @@
 	..()
 	if(visualsOnly)
 		return
-	if(!is_station_level(H.z) && !is_centcom_level(H.z))
+	if(H.stat == DEAD)
 		return
 	for(var/obj/item/gps in H.contents)
 		gps.gpstag = "MINE[gps_number]"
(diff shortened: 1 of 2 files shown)


code/modules/jobs/job_types/shaft_miner.dm
```diff
@@ -35,6 +35,7 @@
 
 /datum/outfit/job/miner
 	name = "Shaft Miner"
+	var/static/gps_number = 1
```
Bibby0110, line 38:
```suggestion
	var/static/gps_number = 0
```

SomeguyManperson: default gps starts at 0 and we want to differentiate from that

MajManatee (approved):

nmajask (approved):

PR #14433, opened 2022-06-14
Secbelt + shotgun shell boxes rework, shoulder holster and bandolier buff

Skrem7:
# Document the changes in your pull request

Makes sec belts bulky again at the trade-off of being able to hold rubbershot, lethalshot, breacherslug, and beanbag ammo boxes. Each of these shotgun ammo boxes can now only hold shotgun shells. Secbelts can also now hold telescopic shields.

Also buffs sec webbing and hoS belt to hold 7 items. For clarification, secbelts can currently hold 6 items as well. The HoS' extra slot spawns with a teleshield, moved into it from the locker. The HoS now begins with their belt but does not spawn with the extra baton, flash, or cuffs. They're instead found in the locker.

Also also makes the shoulder holster normal-sized again as it unintentionally called the parent bulky size of belts, when it barely gives an inventory advantage.

Also also also makes the bandolier able to hold 24 shells as a bit of a makeup for the fact that secbelts can hold significantly more shotgun ammo now.

Also also also also the assault belt from the SWAT crate is now identical to the tactical webbing that Syndicates get.

Very minor change but the beanbag boxes are full now instead of having one slot missing.

# Wiki Documentation

Secbelt and all variants can now hold shotgun shell boxes and teleshields
Shotgun shell boxes can now only hold shotgun shells (doesn't have to be matching type)
Security webbing and HoS belt can hold 7 items now, instead of 6
HoS belt starts with the locker's teleshield in it
HoS spawns with their belt on, moves extra baton/flash/cuffs to locker
Assault belt now identical to tactical webbing
Shoulder holster is normal-sized again, still only holds 3 items
Bandolier can now hold 24 shells, instead of 18

# Changelog

:cl:  
tweak: Secbelt + variants can hold shotgun shell boxes and teleshields, bulky now.
tweak: Shotgun shell boxes can now only carry shotgun shells.
tweak: Beanbag boxes spawn with 7 beanbags, rather than 6.
tweak: Security webbing and HoS belt storage buffed.
tweak: HoS now spawns with their belt and teleshield from the locker, extra baton/flash/cuffs moved to locker.
tweak: Assault belt now identical to tactical webbing.
tweak: Shoulder holster normal-sized again.
tweak: Bandolier holds more shells.
/:cl:

PR #14440, opened 2022-06-14
WT Tweak/Buff

Skrem7:

Buffs the WT in an effort to bring it closer to the power of the mosin, another notable weapon that's orderable from cargo.

The advantages of the WT over the Mosin:
 - Normal-sized
 - Higher magazine size + quicker reload
 - Inherent AP
 - Faster RoF
 - Ammo variants

Weaknesses:
 - Non rubber/lethal shot can only be printed at an armory techfab after appropriate research
 - Much easier to hack an autolathe to print mosin ammo
 - Mosin does 60 mf'ing brute versus a measly 20 on the base WT bullet
 - Rubber shot was not considered at all with the 2-rnd burst removal, it's currently terrible

Buffs are targeted to make the WT more fun to use (with the 2-rnd burst back) but also tweak its numbers to make it more of a competitive choice for something that has to be ordered from Cargo.

# Wiki Documentation

WT has 2-rnd burst again
WT magazines now hold 22 bullets instead of 20
Standard bullet damage down to 13 from 20
AP bullet damage down to 10 from 15
Incendiary bullet damage down to 7 from 10
Rubber shot damage down to 4 from 5, stamina damage up to 21 from 20

# Changelog

:cl:  
tweak: WT has 2-rnd burst back, bullets do less individual damage, magazine size increased
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/automatic.dm
+++ b/code/modules/projectiles/guns/ballistic/automatic.dm
@@ -83,12 +83,12 @@
 
 /obj/item/gun/ballistic/automatic/wt550
 	name = "security auto rifle"
-	desc = "An outdated personal defence weapon. Uses 4.6x30mm rounds and is designated the WT-550 Automatic Rifle. Has a single fire burst mode, but you still can't figure out what advantage that has over semi automatic."
+	desc = "An outdated personal defence weapon. Uses 4.6x30mm rounds and is designated the WT-550 Automatic Rifle. Has a two-round burst or a semi-automatic firing mode."
 	icon_state = "wt550"
 	item_state = "arg"
 	mag_type = /obj/item/ammo_box/magazine/wt550m9
 	fire_delay = 2
-	burst_size = 1
+	burst_size = 2
 	can_suppress = FALSE
 	can_bayonet = TRUE
 	knife_x_offset = 25
--- a/code/modules/projectiles/guns/ballistic/automatic.dm
+++ b/code/modules/projectiles/guns/ballistic/automatic.dm
@@ -89,6 +89,7 @@
 	mag_type = /obj/item/ammo_box/magazine/wt550m9
 	fire_delay = 2
 	burst_size = 2
+	weapon_weight = WEAPON_MEDIUM
 	can_suppress = FALSE
 	can_bayonet = TRUE
 	knife_x_offset = 25
--- a/code/modules/projectiles/boxes_magazines/external/smg.dm
+++ b/code/modules/projectiles/boxes_magazines/external/smg.dm
@@ -3,7 +3,7 @@
 	icon_state = "46x30mmt-20"
 	ammo_type = /obj/item/ammo_casing/c46x30mm
 	caliber = "4.6x30mm"
-	max_ammo = 26
+	max_ammo = 22
 
 /obj/item/ammo_box/magazine/wt550m9/update_icon()
 	..()
--- a/code/modules/projectiles/boxes_magazines/external/smg.dm
+++ b/code/modules/projectiles/boxes_magazines/external/smg.dm
@@ -3,7 +3,7 @@
 	icon_state = "46x30mmt-20"
 	ammo_type = /obj/item/ammo_casing/c46x30mm
 	caliber = "4.6x30mm"
-	max_ammo = 20
+	max_ammo = 26
 
 /obj/item/ammo_box/magazine/wt550m9/update_icon()
 	..()
--- a/code/modules/projectiles/projectile/bullets/smg.dm
+++ b/code/modules/projectiles/projectile/bullets/smg.dm
@@ -10,22 +10,22 @@
 
 /obj/item/projectile/bullet/c46x30mm
 	name = "4.6x30mm bullet"
-	damage = 20
+	damage = 13
 	wound_bonus = -5
 	bare_wound_bonus = 5
 	armour_penetration = 20
 
 /obj/item/projectile/bullet/c46x30mm_ap
 	name = "4.6x30mm armor-piercing bullet"
-	damage = 15
+	damage = 10
 	armour_penetration = 50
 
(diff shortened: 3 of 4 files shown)


00ze-cyclone:
Maybe you shouldn't increase the mag size since it would make it able of shooting 13 times per mags for 26 damages per click with the 20 AP, it would be a little too strong since laser gun is 12 shots of 20 damage, hellfire is 7 shots of 25, egun/hos gun is 10 shots of 20 damage, decloner is 10 shots of 20 damage (though it's cellular so it's deadlier), even sec hardlight bow is only 25 damage if there's no embedding. It would make a weapon doing more damage than pretty much anything that isn't a shotgun while being able to sustain fire for longer. I know that it's supposed to balance it with the mosin but I don't think sec guns should be balanced around that gun. Currently sec seem more balanced around the laser gun (except shotgun but it's something completely different).

Other than that I agree with those changes.

Skrem7:
> Maybe you shouldn't increase the mag size since it would make it able of shooting 13 times per mags for 26 damages per click with the 20 AP, it would be a little too strong since laser gun is 12 shots of 20 damage, hellfire is 7 shots of 25, egun/hos gun is 10 shots of 20 damage, decloner is 10 shots of 20 damage (though it's cellular so it's deadlier), even sec hardlight bow is only 25 damage if there's no embedding. It would make a weapon doing more damage than pretty much anything that isn't a shotgun while being able to sustain fire for longer. I know that it's supposed to balance it with the mosin but I don't think sec guns should be balanced around that gun. Currently sec seem more balanced around the laser gun (except shotgun but it's something completely different).

WT isn't a sec weapon is the thing, it's a Cargo weapon. If Sec can print WT's, then that's another story entirely. Plus ballistic and energy-based weaponry is inherently different: consider reloading, obtaining of ammo, the nature of the different projectiles, etc.

I would however be more willing to drop the magazine size back to normal if there's more voicing out for it. The main idea of this PR is to get the 2-rnd burst back because right now it's just in a very, very awkward spot for an expensive weapon that has to go through Cargo.

Rdrazga:
The effective damage per 1 26 round mag of WT on this would be 338 damage, The effective damage of a shotgun with slugs is 322, The damage of a Mosin is 300. The downside of this over the other 2 is it takes 4 bursts on target to down a person with no armor, as compared to the shotgun's 3, and the Mosin's 2. Both Shotgun and Mosin rounds can be made in any autolaythe.

Lasers are in game shift-start, can be recharged in multiple locations on station with more rechargers being easily built, and even get a self-recharging version. They are also significantly cheaper to order, and can go through glass. 

Now the real reason why 26 is too much is the ammo once researched is extremely cheap to print AND is as small as shotgun shells, in my opinion you should not be fitting what would essentially be the size of a small p-90 magazine in your boot, hat, or box in your bag, and carry 40 of them with no problem. Dropping the mag size to 24 would also bring its effective damage per mag down to 314 so it is not higher then the shotgun, or dropping it to 20 would be 260 which would still fairly balance the other advantages the WT has over the other two.

Skrem7:
> The effective damage per 1 26 round mag of WT on this would be 338 damage, The effective damage of a shotgun with slugs is 322, The damage of a Mosin is 300. The downside of this over the other 2 is it takes 4 bursts on target to down a person with no armor, as compared to the shotgun's 3, and the Mosin's 2. Both Shotgun and Mosin rounds can be made in any autolaythe.
> 
> Lasers are in game shift-start, can be recharged in multiple locations on station with more rechargers being easily built, and even get a self-recharging version. They are also significantly cheaper to order, and can go through glass.
> 
> Now the real reason why 26 is too much is the ammo once researched is extremely cheap to print AND is as small as shotgun shells, in my opinion you should not be fitting what would essentially be the size of a small p-90 magazine in your boot, hat, or box in your bag, and carry 40 of them with no problem. Dropping the mag size to 24 would also bring its effective damage per mag down to 314 so it is not higher then the shotgun, or dropping it to 20 would be 260 which would still fairly balance the other advantages the WT has over the other two.

Assuming belt and shell box rework goes through, shotgun ammo should overall be easier to carry. See [#14433](https://github.com/yogstation13/Yogstation/pull/14433).

General balancing of most weapons dictates that anything with a slower TTK can generally have higher theoretical damages, though WT has inherent AP, which is part of its appeal and the only reason it tends to be used at all. I'll chew on reducing it down to 24, or 22, or 20. Generally, more bullet = more fun. If it weren't foreign, I'd nerf individual bullet damage even more to make it a 3-rnd burst so magazine size is greater.

Rdrazga:
Last thing to take note of on the ttk of the WT over the shotgun and mosin, the wt-550 has tactical reload which means it takes a single click to reload it if it has ammo left, 2 clicks on an empty mag, this makes it extremely fast to reload making the higher ammo capacity not entirely needed to get out more dakka

code/modules/projectiles/guns/ballistic/automatic.dm
```diff
@@ -83,12 +83,12 @@
...
 /obj/item/gun/ballistic/automatic/wt550
 	name = "security auto rifle"
-	desc = "An outdated personal defence weapon. Uses 4.6x30mm rounds and is designated the WT-550 Automatic Rifle. Has a single fire burst mode, but you still can't figure out what advantage that has over semi automatic."
+	desc = "An outdated personal defence weapon. Uses 4.6x30mm rounds and is designated the WT-550 Automatic Rifle. Has a two-round burst or a semi-automatic firing mode."
 	icon_state = "wt550"
 	item_state = "arg"
 	mag_type = /obj/item/ammo_box/magazine/wt550m9
 	fire_delay = 2
-	burst_size = 1
+	burst_size = 2
```
SomeguyManperson, line 91:
```suggestion
	burst_size = 2
	weapon_weight = WEAPON_MEDIUM
```
do not like burst fire with dual wield

SomeguyManperson (approved):

PR #14441, opened 2022-06-14
Revert "Ports 3/4 crates sprites and mule bot from FTL13"

KillerOrcaCora:
Seen these crates a few times in-game, they are very jarring and don’t really fit with anything around them and the perspective suddenly swapping on just these crates also feels a bit off.


tattax:
lockers and coffins literally already had the front-up view, crates were the one missing out, not vice-versa

JamieD1:
Oof, will need Cuackles and maybe parth or ident to chime in here.

cuackles:
> Oof, will need Cuackles and maybe parth or ident to chime in here.

thinking of it, im ok with both but one has to be better for the game but i cant decide wich one

KillerOrcaCora:
The point of this PR is to go back to the old perspective, I can easily resprite the looks of the newer crates onto the old ones but rather have one perspective overall for crates.

KillerOrcaCora:
After discord vote, will be making a new PR to revert back to the old perspective but also take some of the aspects people wanted from the new one to make a compromise for both groups.

PR #14442, opened 2022-06-14
Adds stored research and makes toxins bombs apply it instead of directly producing points

SomeguyManperson:

Adds stored researhc to slow the science speedrun techweb 100% run by making toxins bombs give stored research instead of actual research points
stored research works in the vein of rested experience where it is applied when servers produce research, doubling research speed
i.e. normal research speed gives 3000 points per minute, if you have stored research it will be expended at 3000 points per minute to increase normal research speed to 6000 points per minute until it runs out

# Wiki Documentation

toxins bombs no longer give direct research, instead doubling research speed for the amount of points they are worth

# Changelog

:cl:  
tweak: toxins bombs no longer give direct research, instead doubling research speed for the amount of points they are worth
/:cl:


ToasterBiome:
Interesting concept. I wonder if something like this can be used with the destructive analyzer where you can deconstruct some items and give stored points

ToasterBiome (approved):

JamieD1:
Brilliant

PR #14444, opened 2022-06-14
Allows WT rifles to be made from the armory laithe. With research!

MajManatee:

Adds a new tech node behind advanced weapons research, called "Ballistic Weapons Production". Costs 7500 points and allows security to print WT Rifles from the armory laithe.

If you are adding new sprites to the game please add a picture of the sprite in its relative context, ie. Clothing on a mob.

# Wiki Documentation

Add rifle to list of things able to be made from the laithe

# Changelog

:cl:  
rscadd: New research for WT production. Adds to armory laithe.
/:cl:

### DIFF
--- a/code/modules/research/designs/weapon_designs.dm
+++ b/code/modules/research/designs/weapon_designs.dm
@@ -411,3 +411,13 @@
 	build_path = /obj/item/gun/ballistic/bow/energy
 	category = list("Weapons")
 	departmental_flags = DEPARTMENTAL_FLAG_SECURITY
+
+/datum/design/wtrifle
+	name = "WT-550 Automatic Rifle"
+	desc = "An outdated personal defence weapon. Uses 4.6x30mm rounds and is designated the WT-550 Automatic Rifle."
+	id = "wtrifle"
+	build_type = PROTOLATHE
+	materials = list(/datum/material/iron = 7500, /datum/material/glass = 1500, /datum/material/silver = 2500)
+	build_path = /obj/item/gun/ballistic/automatic/wt550
+	category = list("Weapons")
+	departmental_flags = DEPARTMENTAL_FLAG_ARMORY
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -791,6 +791,15 @@
 	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 2500)
 	export_price = 5000
 
+/datum/techweb_node/ballistic_production
+	id = "ballistic_production"
+	display_name = "Ballistic Weapons Production"
+	description = "The first step to making your own militia."
+	prereq_ids = list("adv_weaponry")
+	design_ids = list("wtrifle")
+	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 7500)
+	export_price = 5000	
+
 ////////////////////////mech technology////////////////////////
 /datum/techweb_node/adv_mecha
 	id = "adv_mecha"
--- a/code/modules/research/techweb/layout.dm
+++ b/code/modules/research/techweb/layout.dm
@@ -534,3 +534,7 @@
 /datum/techweb_node/mech_xray
 	ui_x = -640
 	ui_y = -896
+
+/datum/techweb_node/ballistic_production
+	ui_x = -740
+	ui_y = -750


ynot01:
perhaps a different weapon

MajManatee:
I considered making some sort of M4 to give them instead, however I am a terrible spriter

Hopekz:
If anything I think this would make sense as something before advanced weapons since the self-charging e-gun outclasses this gun

Also, normal guns seem easier to make than self-charging energy rifles.

Marmio64:
Kind of goes against how security is supposed to need to work with cargo for stuff sometimes, like when they need more ballistic weaponry, but maybe it can still work out.

VaelophisNyx:
the mentality of "Sec needs to work with cargo" is one I have always questioned; why should sec have to work with a department prone to spontaneous revolution; or worse, why lock sec out of a way to deal with some threats because they didn't cross the quarter-of-a-station distance to cargo before the antags did and just obliterate sec's ability to source most weapons that aren't shotguns and the small offerings of the armory. I think this is a good solution, as a player, since it'd let sec have an answer to antags that become immune or massively resistant to lasers, without being reliant on the limited ammo capacity and damage potential of shotguns per-unit-cost. And by making them printable without making them armory starters, you prevent sec offs from just grabbing a WT rifle and field-executing people because it's faster. They also still have to source ammunition for it, either by crafting or hoping cargo isn't dead/rev'd.

TLDR
More options good, more sec ballistics good.

code/modules/research/designs/weapon_designs.dm
```diff
@@ -411,3 +411,13 @@
...
 	category = list("Weapons")
 	departmental_flags = DEPARTMENTAL_FLAG_SECURITY
+
+/datum/design/wtrifle
+	name = "WT-550 Automatic Rifle"
+	desc = "An outdated personal defence weapon. Uses 4.6x30mm rounds and is designated the WT-550 Automatic Rifle."
+	id = "wtrifle"
+	build_type = PROTOLATHE
+	materials = list(/datum/material/iron = 7500, /datum/material/glass = 1500, /datum/material/silver = 2500)
+	build_path = /obj/item/gun/ballistic/automatic/wt550
```
SomeguyManperson, line 421: This starts with ammo in it

SomeguyManperson:
> If anything I think this would make sense as something before advanced weapons since the self-charging e-gun outclasses this gun

This is wrong simply because ballistics can recharge faster than self-recharging energy guns by switching magazines while carrying a box of 7 spare magazines

SomeguyManperson:
ALso I don't like the idea of security running around dual wielding objectively better lasers (reloadable and AP) mid-lategame and/or not using it ever regardless because weapons are either completely nonlethal and usable in normal environments or have some fancy bullshit stats that make them the best weapon for x scenario i.e. xray laser or particle accelerator

TheRyeGuyWhoWillNowDie:
> ALso I don't like the idea of security running around dual wielding objectively better lasers (reloadable and AP) mid-lategame and/or not using it ever regardless because weapons are either completely nonlethal and usable in normal environments or have some fancy bullshit stats that make them the best weapon for x scenario i.e. xray laser or particle accelerator

calling them objectively better lasers is pretty misleading, security can print off a zillion infinitely self recharging laserguns with mindshield firing pins, autorifles do have some advantages like tactical reloads, but they also cant shoot through windows, and you have to carry magazines for them, magazines which require tech or cargo to get, and even with the tech can only be printed from the sec/armory prolathe, whereas even standard laserguns can be recharged in any department with the sec outpost wall charger

Hopekz:
> This is wrong simply because ballistics can recharge faster than self-recharging energy guns by switching magazines while carrying a box of 7 spare magazines

Never had this problem but I get the argument. 
The thing with printable self charging eGuns is that you can just print 2-3 of them and by the time you've used the 2nd or third one the first one will be recharged so I think it accomplishes the same thing in that argument making them more equal in that department.

Also, when do you ever need 7 magazines worth of damage in ss13 unless you're solo'ing a blob. Most situations in ss13 are covered by either 1 magazine or 2 at most (2nd Egun)

A full magazine in a rifle or charge of an eGun is a lot of damage

JamieD1:
Down to theos

gaminada:
I dislike this because as it has been said before, its easier to stack ammo clips than having to wait for lasers to charge + i dont enjoy stripping cargo even more from their role in the station

PR #14455, opened 2022-06-15
makes some of the recent megafauna drops less shit

LazennG:

after a while of watching the items i realized they were babyproofed to the point of being useless/impractical so i'm hoping to fix that.

Grand Core: 
-blood costs halved and detailed in the toolbar because being woozy after using it like twice is stupid as well as making people guess how many times they can use it before dropping dead
Rabid Cuffs:
- increased fauna damage to 60 brute from 30 because the old number wasn't enough to warrant using it over a ka in almost any circumstance. Also changed damage for people from 10 to 20 because bubblegum is usually last on the list of things people want to do with lavaland, plus the ability can be used every 10 seconds so it shouldn't be as bad as it is in terms of damage.
-changed around the functionality of the ability so 13 brute/burn/toxin is recovered on hitting living fauna, 20 on dead ones, and 5 on living people. This change was because i felt it was too much of a hassle to find a corpse to heal off of.
(killing an animal with the swipe attack heals for both as a critical heal)
Bloody Knuckles:
Gave them an ability to splash blood in a line for 9 tiles in the direction the user is facing with a 10 second cooldown so the ability to mark targets on blood is less situational.

If megafauna are fun to beat then their loot should be fun to play with, and I hope these changes are a step in that direction.

# Wiki Documentation

cursed stomach is still on bg's loot page so update that.
cuffs
also include rabid cuffs and how they're gloves that heal the user a bit when targeting an animal or person with their ability, along with doing damage to the target.

# Changelog

:cl: @ToasterBiome @Chubbygummibear @LazennG 
rscadd: Added inhand ability for bloody knuckles  
tweak: tweaked some recent loot to hopefully be less useless
/:cl:
thanks @ToasterBiome and @Chubbygummibear for helping me code past migraines and helping me spitball ideas

### DIFF
--- a/code/modules/clothing/gloves/miscellaneous.dm
+++ b/code/modules/clothing/gloves/miscellaneous.dm
@@ -116,7 +116,7 @@
 
 obj/effect/proc_holder/swipe
 	name = "Swipe"
-	desc = "Swipe at a target area, dealing damage and healing yourself. Creatures take 60 damage while people and cyborgs take 20 damage. Living creatures hit with this ability will heal the user for 15 brute/burn/poison while dead ones heal for 20 and get butchered. People and cyborgs hit will heal for 5"
+	desc = "Swipe at a target area, dealing damage to heal yourself. Creatures take 60 damage while people and cyborgs take 20 damage. Living creatures hit with this ability will heal the user for 15 brute/burn/poison while dead ones heal for 20 and get butchered. People and cyborgs hit will heal for 5"
 	action_background_icon_state = "bg_demon"
 	action_icon = 'icons/mob/actions/actions_items.dmi'
 	action_icon_state = "cuff"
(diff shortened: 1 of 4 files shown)


ToasterBiome:
If you’re adding multiple people to the credits you also need to include yourself

code/modules/clothing/gloves/miscellaneous.dm
```diff
@@ -154,32 +154,33 @@ obj/effect/proc_holder/swipe
...
 		if(isanimal(L))
-			L.adjustBruteLoss(30)
+			L.adjustBruteLoss(60)
+			if(L.stat != DEAD)
+				caller.adjustBruteLoss(-13)
+				caller.adjustFireLoss(-13)
+				caller.adjustToxLoss(-13)
 			if(L.stat == DEAD)
 				L.gib()
+				to_chat(caller, span_notice("You're able to the body entirely!"))
```
SomeguyManperson: You're able to the body entirely

SomeguyManperson (approved):

PR #14464, opened 2022-06-15
Makes stasis beds slowdown metabolism instead of stopping it and increases body decay time

nmajask:

Does what it says on the tin, makes field treatment much more viable while making stasis less of a necessity for doctors.

- [ ] Fix the logic being very broken
- [ ] Fix stasis beds not slowing down body decay

# Spriting
None.

# Wiki Documentation

Add the new effect of the stasis bed to the machines page. 

# Changelog

:cl:   
tweak: stasis beds now slowdown metabolism instead of stopping it  
tweak: bodies now take twice as long to decay
/:cl:

### DIFF
--- a/code/__DEFINES/DNA.dm
+++ b/code/__DEFINES/DNA.dm
@@ -172,8 +172,8 @@
 //organ defines
 #define STANDARD_ORGAN_THRESHOLD 	100
 #define STANDARD_ORGAN_HEALING 		0.001
-/// designed to fail organs when left to decay for ~15 minutes
-#define STANDARD_ORGAN_DECAY		0.00222
+/// designed to fail organs when left to decay for ~30 minutes
+#define STANDARD_ORGAN_DECAY		0.00111
 
 //used for the can_chromosome var on mutations
 #define CHROMOSOME_NEVER 0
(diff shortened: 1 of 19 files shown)


ynot01:
in some ways this is a buff

VaelophisNyx:
I really do not want to have to deal with unstoppable organ rot, even slowed. It's crippling enough when someone brings a miner in with 3 functional organs because they were on the ground for 5 minutes. Stasis made sense and had drawbacks in that you couldn't process chems on stasis at all.

I do not see the need for this particular nerf, compared to making surgical tables actually stop flinching when a patient is on them, and anesthetized, so I can stop damaging people because their unfeeling, unconscious body moved. OR tables are already ignored for not having any effect on organ/body rot along with the flinching issues, and this will not change that. If you want us to stop, in your eyes, abusing stasis, give us incentives to actually bother with the tables, instead of making stasis just worse.
You gotta remember that not many docs know how to do surgery well enough to fix one dead body in a timely manner, let alone when 10 come rolling into medbay at once.

The organ rot and defib changes are good, but the stasis nerf is awful

maxion12345:
stasis nerf bad, see above

PR #14472, opened 2022-06-16
remove one space that might be the reason why polys can't be CE

00ze-cyclone:
# Changelog
remove one space

:cl:  
rscdel: Removed a space in the species whitelisting
bugfix: maybe fixed polys not being able to be CE (I hope it's not actually something as stupid as a singular space that broke everything though it would be nice because it would mean that it's easy to fix)
/:cl:


Hopekz:
Changelogs are for player-facing changes so your "removed a single space in __ " wouldn't apply

JamieD1 (approved):

PR #14487, opened 2022-06-18
Blob buff: double blob passive point gen

VaelophisNyx:
blobs struggle to keep up with crew attacking on all sides with late-late game weapons like Particle accelerators and Phazons. This buff should help with that a little.

Blob default passive resource generation (ie: not from resource blobs) increased from 2 to 4

Changes UNTESTED. Requires testing that is practically impossible in a local-host environment.

this change is made to help blobs counter 6-10 crew members/borgs, emitters, heavily armed combat mechs, particle accelerators, all hammering away at it from many directions at once. When a blob is spotted, you have mere moments before you will be engaged in the passive resource drain of being forced to replace blob tiles faster than you generate points, even with resource blobs.

This does not make any changes to blob resource cap, nor the speed at which resource-blobs generate resources, nor the slowdown from placing more resource blobs

# Wiki Documentation

Blob wiki might need a tweak to match? Unsure if resource rate is mentioned there.

# Changelog

:cl:  
tweak: blob base resource generation up to 4, from 2
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blob.dm
+++ b/code/modules/antagonists/blob/blob.dm
@@ -7,7 +7,7 @@
 
 	var/datum/action/innate/blobpop/pop_action
 	var/starting_points_human_blob = 60
-	var/point_rate_human_blob = 2
+	var/point_rate_human_blob = 4
 
 /datum/antagonist/blob/roundend_report()
 	var/basic_report = ..()


VaelophisNyx:
HEAVILY desire input from others on this, as while I think this would help, I want to hear from others. This is based on my experiences playing blob, others might have different experiences.

VaelophisNyx:
Also I'm not even sure I've changed this in the right place; or if random blobs even occur, which if that is the case, this change does nothing as admin spawned blobs have custom resource generation rates

maxion12345:
I uh
what about blobs that get lucky and spawn before crew get midround stuff?@ a buff like this could turn this into "very difficult fight while rushing tech/Mecha" to "why even bother just set the self destruct"

VaelophisNyx:
I won't lie, I can't ultimately test that solo. But also I have never seen a blob spawn when the crew didn't already have 10+ ways of dealing with it laying around or ready to print

Bibby0110 (changes requested):
This var doesn’t do anything

Also you can definitely test this change locally

JamieD1:
How about you make the point gen scale with round duration

VaelophisNyx:
Well Jamie, I would; however. I have no idea how to do that

> This var doesn’t do anything

Alright, I was concerned about that.

> Also you can definitely test this change locally

I mean yeah I could test if it compiles and gives more points, but I can't test the *balance* impact of that in a natural setting. IE: being besieged by the crew with usually combat mechs, borgs, particle accelerators, emitters, all at the same time.

VaelophisNyx:
I don't have enough experience to do this right now.

PR #14488, opened 2022-06-18
Adds new Chemical Analyzer implant (Science Goggle implant)

VaelophisNyx:
Adds a new eye implant that allows you to have the reagent scanning ability without wearing goggles and not being a silicon, or dead.

Adds new implant, adds that implant as a design unlocked with the other HUD implants
makes tweaks to the traits file to add a new trait that gives you the SCAN_REAGENTS clothing flag, but at the trait level. Properly removed when the implant is removed. Does not allow you to scan items for research value, like the goggles. Name pending possible change to reflect this.

also had to adjust the Human helper file, to include a second check for if you have the Trait, if both the dead or silicon checks return false.

Changes tested, works flawlessly. Only the lathe design was untested. Can test tomorrow.

I would like to ask if anyone would prefer this function would rather be rolled into the MedHUD implant, instead of an individual implant.

# Wiki Documentation

Would have to be added to all locations that HUD implants are mentioned.

# Changelog

:cl:  Galacticruler, @Chubbygummibear, @Hopekz 
rscadd: Added new eye implant for reagent scanning, the chemical analyzer implant
tweak: added new implant to the tech node with the other HUD implants
tweak: added third mob-level check to allow for the new implant to work
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -259,6 +259,7 @@
 #define TRAIT_ANOREXIC			"anorexic"
 #define TRAIT_SHIFTY_EYES		"shifty_eyes"
 #define TRAIT_ANXIOUS			"anxious"
+#define TRAIT_SEE_REAGENTS		"see_reagents"
 
 // common trait sources
 #define TRAIT_GENERIC "generic"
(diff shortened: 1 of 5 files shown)


GraveHat:
Adding this feature onto the medhud would give the usual doctor more options for their optimal implant builds, forgoing the double utility of using 2 different information huds for the redundancy of using their specialized huds on top of the medhud implant. 

Having the reagent analyzer while not needing to wear the goggles would also be quite a convenience for medbay as they have to work with reagents far more than science ever will. And if I'm not mistaken, are unable to print science goggles in the medical lathe even with the huds researched. This would also mean chemists can use the medical hud for once instead of just wearing their science goggles the entire time if any of them bothers to get it.

VaelophisNyx:
Updated the CL to include credits for those who helped me figure out how to even do this.

JamieD1 (approved):

PR #14497, opened 2022-06-18
Adds Brain Damage Quirk

redmoogle:

You start with 1-4 brain traumas that are uncurable(because otherwise the quirk would be kinda pointless)
Picks from both the severe and mild pools.
Gives you 7 quirk points. open to changes

# Wiki Documentation

New quirk that gives 1-4 severe/mild brain traumas
Gives you 7 quirk points to use

# Changelog

:cl:  
rscadd: Brain damage quirk
/:cl:

### DIFF
--- a/code/datums/traits/negative.dm
+++ b/code/datums/traits/negative.dm
@@ -768,3 +768,20 @@
 	gain_text = span_danger("You feel like your blood is thin.")
 	lose_text = span_notice("You feel like your blood is of normal thickness once more.")
 	medical_record_text = "Patient appears unable to naturally form blood clots."
+
+/datum/quirk/brain_damage
+	name = "Brain Damage"
+	desc = "The shuttle ride was a bit bumpty to the station."
+	value = -1
+	gain_text = span_danger("Oww....")
+	lose_text = span_notice("Your head feels good again.")
+	medical_record_text = "Patient appears to have brain damage."
+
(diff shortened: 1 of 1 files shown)


VaelophisNyx:
I'd personally want to see it as a 2 pointer; +1 point for potentially crippling permanent brain traumas round start is a bit...meh.

code/datums/traits/negative.dm
```diff
@@ -768,3 +768,20 @@
 	gain_text = span_danger("You feel like your blood is thin.")
 	lose_text = span_notice("You feel like your blood is of normal thickness once more.")
 	medical_record_text = "Patient appears unable to naturally form blood clots."
+
+/datum/quirk/brain_damage
+	name = "Brain Damage"
+	desc = "The shuttle ride was a bit bumpty to the station."
+	value = -2
+	gain_text = span_danger("Oww....")
```
Hopekz: All the other gain texts are written in 2nd person vs 1st person in terms of telling you what is happening to you

Marmio64:
im not sure RNG should be such a deciding factor on whether you get absolutely fucked or just a little bit fucked. Maybe narrow down the possible amount of traumas to 2 or 3 instead of 1-4. You could also separate this into two traits, one that picks from mild traumas and one from severe traumas, with the severe one being more expensive.

JamieD1 (approved):

PR #14501, opened 2022-06-19
After the blob wins, it gets MANY HP

ghost:

After the blob reaches critical mass, it gets 9999999 HP. Because it, like, has already won. Cry about it.

# Changelog

:cl: SuperSlayer
tweak: Blob now becomes almost unlikable after reaching critical mass 
/:cl:


SomeguyManperson:
I aould consider losing your core after getting unlimited resources to be a skill issue

ghost:
> I aould consider losing your core after getting unlimited resources to be a skill issue

I don't know how to make that loosing the core if the blob won will not make it simply die

JamieD1 (approved):

PR #14519, opened 2022-06-20
Lizards are now cannonically 8 feet tall

alexkar598:
:cl:  
tweak: Lizards are now cannonically 8 feet tall
/:cl:


Marmio64:
Lizards will never, NEVER, be ballin'.

JamieD1 (approved):

PR #14521, opened 2022-06-20
Saboteur/Medical Syndicate Borg Tweaks

Skrem7:

Syndicate medical cyborg loses EMAG now because why does it have it as a support unit and not a saboteur?

Originally gave EMAG to sabo borg as a tradeoff before the comments graciously pointed out that a camouflaged EMAG user that can subvert borgs was maybe a bad idea.

Also removes the crowbar from the medical syndicate borg because it's strictly meant to be a support unit, instead giving it cuffs in the rare case that it or an operative incapacitates an individual and wishes to bring them back to the shuttle to be brainwashed.

Gives saboteur borg a combat knife for some effectiveness in melee combat outside of the welder, though the weapon isn't as effective as the esword or the esaw that the other two Syndicate borgs get.

# Wiki Documentation

Medical Syndicate borg has lost EMAG and crowbar. Gained zipties identical to those the secborg has.
Saboteur Syndicate borg has gained combat knife (identical to the one that is gained from EMAG'ing toolset implants).

# Changelog

:cl:  
tweak: Syndie mediborg loses emag and crowbar, gains zipties
tweak: Syndie saboborg gains combat knife
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -678,7 +678,6 @@
 		/obj/item/crowbar/cyborg,
 		/obj/item/wirecutters/cyborg,
 		/obj/item/multitool/cyborg,
-		/obj/item/card/emag,
 		/obj/item/stack/sheet/metal/cyborg,
 		/obj/item/stack/sheet/glass/cyborg,
 		/obj/item/stack/sheet/rglass/cyborg,
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -648,8 +648,7 @@
 		/obj/item/bonesetter,
 		/obj/item/melee/transforming/energy/sword/cyborg/saw,
 		/obj/item/roller/robo,
-		/obj/item/card/emag,
-		/obj/item/crowbar/cyborg,
+		/obj/item/restraints/handcuffs/cable/zipties,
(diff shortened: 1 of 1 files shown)


00ze-cyclone:
Couldn't you make the welder stronger instead of adding another weapon since it doesn't really fit the theme and also make it clunkier to use. The saboteur will probably always have their welder in one of their slots since it's healing and a weapon, it would also help them remain somewhat camouflaged since they'll use a welder like a regular engiborg instead of a knife(might confuse the opponent for just enough time to get anopther hit).

Rdrazga:
I have a feeling giving a borg that can disguise itself as non malf borgs an emag, capable of malfing normal borgs, when it also has a flash, could be a decently bad idea.

Skrem7:
> Couldn't you make the welder stronger instead of adding another weapon since it doesn't really fit the theme and also make it clunkier to use. The saboteur will probably always have their welder in one of their slots since it's healing and a weapon, it would also help them remain somewhat camouflaged since they'll use a welder like a regular engiborg instead of a knife(might confuse the opponent for just enough time to get anopther hit).

i don't entirely disagree but this would require making a new weapon which is yes effort but ALSO you can still use the welder as a weapon if you really want to

the idea of the knife is that it's a bit more robust but it's for when your cover is gone or you just want to kill someone quick, the bleed wound swill have quicker effects and downsides as opposed to burns which might be problematic in about ten minutes not one

JamieD1 (approved):

PR #14524, opened 2022-06-20
Stechkin Ammo Buffs

Skrem7:

Attempts to make purchasing additional stechkin magazines less of a TC dump and also buffs HP and incind to be much more usable relative to their costs.

A bit drastic of a change but more than willing to see adjustments and suggestions

# Wiki Documentation

In the uplink:
10mm magazine replaced with Pair of 10mm magazines which contains two 10mm magazines.
AP 10mm magazine TC cost reduced to 1 from 2.
HP 10mm magazine TC cost reduced to 1 from 3.
Soporific 10mm magazine replaced with Pair of Soporific 10mm magazines which contains two soporific 10mm magazines.
Incendiary 10mm magazine TC cost reduced to 1 from 2.

Values themselves:
HP damage and AP to 45 and -45 from 40 and -50 respectively
Incendiary damage up to 20 from 15

# Changelog

:cl:  
tweak: The cost of AP, HP, and incind 10mm magazines have all been reduced to 1 TC each
tweak: Normal and soporific 10mm magazines are now bought in pairs, they come in a box
tweak: HP 10mm now does 45 damage with only -45 AP, up from 40/-50
tweak: Incind 10mm now does 20 damage, up from 15
/:cl:

### DIFF
--- a/code/game/objects/items/storage/uplink_kits.dm
+++ b/code/game/objects/items/storage/uplink_kits.dm
@@ -480,6 +480,20 @@
 	new /obj/item/reagent_containers/glass/bottle/amanitin(src)
 	new /obj/item/reagent_containers/syringe(src)
 
+/obj/item/storage/box/syndie_kit/pistolammo
+	name = "10mm magazine box"
+
+/obj/item/storage/box/syndie_kit/pistolammo/PopulateContents()
+	for(var/i in 1 to 2)
+		new /obj/item/ammo_box/magazine/m10mm(src)
+
+/obj/item/storage/box/syndie_kit/pistolsleepyammo
+	name = "10mm soporific magazine box"
+
+/obj/item/storage/box/syndie_kit/pistolsleepyammo/PopulateContents()
+	for(var/i in 1 to 2)
+		new /obj/item/ammo_box/magazine/m10mm/sp(src)
+
 /obj/item/storage/box/syndie_kit/nuke
 	name = "box"
 
--- a/code/modules/projectiles/projectile/bullets/pistol.dm
+++ b/code/modules/projectiles/projectile/bullets/pistol.dm
@@ -27,8 +27,8 @@
 
 /obj/item/projectile/bullet/c10mm_hp
 	name = "10mm hollow-point bullet"
-	damage = 40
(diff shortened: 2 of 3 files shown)


ynot01:
ratatatatatatatata

JamieD1:
This is a ridiculous buff, you are literally making this tiny shitty gun, be able to put 4 people to sleep for 2 TC at RANGE.

Skrem7:
> This is a ridiculous buff, you are literally making this tiny shitty gun, be able to put 4 people to sleep for 2 TC at RANGE.

It's a band aid fix for a weapon that is inherently a TC dump. 

Soporific and normal rounds do the same levels of damage. It's just one does brute and the other does stamina which then puts to sleep (which, might I add, is the exact same damage a disabler does without getting to go through windows and being inherently limited in quantity AND being a bullet instead of energy).

As it stands, it's hellish to realistically consider buying the stechkin over so many other options which can be stolen, ordered, or just bought (other uplinks; edagger, ebow, minibomb).

Current numbers have HP which is only good against unarmored at 3 TC. If you are spending a chunk of TC on a weapon you want it to be good against Security. It's as simple as that. Most non-sec can just be EMP flashlighted into some kinda alpha.

I don't disagree this is a *wee* bit over the top, but ammo cost is an issue I wanted to bring forward.

JamieD1:
> 

Sleep and stamcrit is not the same.. especially when being disabled you wont be fecking murdered after.

Skrem7:
> > 
> 
> 
> 
> Sleep and stamcrit is not the same.. especially when being disabled you wont be fecking murdered after.

Normal 10mm does the same damage and puts you into crit while probably applying piercing wounds, which WILL kill you.

Soporific offers a way to kidnap people I 100% agree on stam/sleep not being the same but for a weapon the main aspects that matter are 1) calling for help and 2) not being able to follow up. You put someone to sleep and they're out a lot longer than stamcrit, but someone can get them up just by shaking them. Shoot them with lethals and they go down for good.

Unless there is some inherent design with soporific I do not see how this is an issue.

JamieD1 (approved):

PR #14537, opened 2022-06-21
Upgrade batong use powerrrr

Mqiib:
Only 1.5x power usage is nothing when you can slap a bluespace battery in it. 4x power is still nothing with a bluespace battery because it effectively gives the baton the same amount of hits as a baseline unupgraded baton. This will force officers to ACTUALLY upgrade their baton or use it sparingly since it'll only have like two hits in it on a base battery charge.

# Wiki Documentation

Baton upgrade increases cost of using it by 4x rather than the piddly 1.5x it was before.

# Changelog

:cl:  

tweak: upgraded batons use 4x power rather than 1.5x

/:cl:


JamieD1:
4x you must be joking

Marmio64:
you could just make it use a percent of power + normal power use, so that using a bluespace power cell doesn't give you like 10x the uses, but it'll still give you extra uses.

Mqiib:
> 4x you must be joking

Nope! If it's gonna be a one-hit stun it better have an actual drawback, namely being LESS efficient stamina-per-energy-wise than baseline batong because you're overclocking the zappy bits. Also it's a one-shot anyways so it really doesn't matter. Officers get two batons round start so it wont be an issue unless you never recharge your batons between fights.

JamieD1:
> you could just make it use a percent of power + normal power use, so that using a bluespace power cell doesn't give you like 10x the uses, but it'll still give you extra uses.

^^

Georgeur:
It takes 1 hit to down anyone who isn't wearing a reflective vest, adamantine armor, heck suit, ash drake armor, or an ERT hardsuit sub type; against shielded enemies, there is no delay between turning it off or on, so you can whack the shield away without losing charge. . You could limit it to 4 hits and it wouldn't be considered a drawback. 
Because:

A) You start the shift with a baton, and you get an extra baton per locker, there are 6 of them, so you carry two. 
B) You can always throw a upgraded cell in it for virtually unlimited charge, or you throw a yellow slime core in it for literally unlimited charge; and you'll have two of them.

This nerf is not too harsh, it is ultimately pointless.
Granted I still support it.

Georgeur:
speaking more emotionally about it
of course it is only my idea, and I am a world class dolt, but having took a look at armor values for a bit. 
This would perhaps serve more of a purpose, if it took away charge as a percent, and that percent was AT LEAST 25%, but more likely 33-34 or 50%. 
the very fact no armor exists that would allow you to take more than 2 hits (excluding of course nanites, reagent process, DNA vault, being stacked on the existing armor), means it is not a good idea, to let this thing swing more than 4 times. 
even then it is a band aid for the fact every sec officer will have 2, but at least itll prevent people who dont even know how to open a locker from having a game win stick. 

edit: to survive 3 hits and not get dropped to the floor (89 stam damage), youd need a energy armor on the hit limb of 58, highest energy armor (shared by all the armors I mentioned including death squad) is 50. 
this thing should not swing more than a couple times, if anything it needs its damage put at EXACTLY 94 instead of 140 and you'd need someone to touch the energy armor of every single piece (no one has touched it since it was first added btw with the exception of like 4 single things :) ) for it to even be questioned whether this thing should be able to swing more than 4 times.

edit: edit: just checked, no normal armor has more than 10 energy and nanites gives 15 and you need a total of 37 to not get 1 shot 

edit: edit: edit: how the fuck did no one check any of this

edit: edit: edit: edit: god is dead and we have killed him

edit: edit: edit: edit: edit: EVEN WITH MCFUCKING PRE NERF NANITES, AND LIKE THE HOS HARDSUIT OR SOMETHING STACKED ON IT, YOU WOULDNT HAVE ENOUGH ARMOR TO TANK 1 BATONG. I cannot functionally see a reason why an upgraded baton needs more than 1 use. 

edit: edit: edit: edit: edit: edit: amogus, also i changed what i thought was a good number for upgraded batong damamage from 90 to 94, but it could also be 99 so that normal armor vests let you take just below 90 stam and not falling over.
because apparently you fall over at 90. this is so that either the low end armor with energy damage lets you tank 1 hit, or the high end crew armor with energy damage lets you tank 1 hit because even whilst ranting like a neckbeard i must be consistent

Bibby0110:
> > you could just make it use a percent of power + normal power use, so that using a bluespace power cell doesn't give you like 10x the uses, but it'll still give you extra uses.
> 
> ^^

Giving a percent is stupid. Removes any incentive to upgrade and technically encourages using shitty batteries that charge quickly.

Give it a proper drawback like a cooldown, charge up time or something else. This is really the only sensible way to nerf it through power usage

JamieD1 (changes requested):
Make it a percentage based on the current Cell or ill just remove the ability to change the cell. Your choice.

Bibby0110:
> Make it a percentage based on the current Cell or ill just remove the ability to change the cell. Your choice.

stupid

Georgeur:
Burst damage. 
That'd be like saying the pre nerf dbarrel on cm was bad cause it only had two shots, but it did enough damage you only needed 2-3 shots. 
It'd be like saying adding a insta kill gun with only 1 shot would make it a bad weapon. 
Honestly though it probably shouldnt be allowed to have cell upgrades unless we remove slime cores, cause the infinite charge has always bugged me.

Bibby0110:
@Mqiib i would advise just keeping the PR open as-is, unless you’re hellbent on getting this PR merged by vandalizing it with Jamie’s suggestion

PR #14540, opened 2022-06-21
CQC Counter Only Knocks Down

Mqiib:
The counter, now that it has a 90% proc chance, was a LITTLE much so now it doesn't stun, only knock down. Hopefully this reels in the kind of extreme power the ability to counter attacks with 90% efficiency grants.

# Wiki Documentation

The counter only knocks down for 6 seconds, does not stun.

# Changelog

:cl:  

tweak: CQC counter no longer hard stuns

/:cl:


TheRyeGuyWhoWillNowDie:
me when i cqc counter a security officer and then they pull the second upgraded instawin baton from their belt and hit me because they're not stunned

Georgeur:
We cannot balance around 140 stam damage or the entire ecosystem will go to hell, and dw energy armor will become more prominent and/or the stam damage will be reduced. So don’t worry, this change as a whole is good, despite any concern of the baton.

JamieD1 (approved):

PR #14557, opened 2022-06-22
adds grounding rods to delta

MajManatee:

Even though delta uhhh sucks, this makes the engine suck a bit less.
Delta engine now starts with grounding rods. Previously the station didn't have ANY pre-built rods, just BEGGING for a tesloose. This should negate that... if we ever go back to delta.

# Changelog

:cl:  
bugfix: Delta now has grounding rods.
/:cl:


ToasterBiome (approved):
can be adjusted later

PR #14574, opened 2022-06-23
Shows all chemicals in the chem machine you could dispense if fully upgraded (just read the PR)

ToasterBiome:

Something about "fucking up the chemist's muscle memory", so I was paid (legally required to state this) to fix it, so this is the solution I came up with. Adds buttons for all reagents it could have (barring emagged) and tells you which tier manipulator it requires.

g5oJIrbxMo

Hovering over them will tell you what tier is needed. T2 is added and then the list is sorted (to have iron in the middle) and then the rest are added on. This mimics the functionality how we have it now but it will always be consistent no matter the tier it is (so the buttons don't move)

Emagging it adds the emagged chemicals to the end like normal.

# Changelog

:cl:  
rscadd: Adds buttons to preview what higher tier chem dispensers can put out
/:cl:


ynot01:
epico

Georgeur:
Muchas. Yes.

JamieD1 (approved):
fucking snowflakes

PR #14586, opened 2022-06-24
Rollerbeds now do not cost plastic

Sniblet:

Rollerbeds no longer need an entire sheet of plastic to build, or any plastic at all
Much less of them spawn with the map now and I don't know what that comment is talking about when it says "balance" because body bags exist and are basically better

# Changelog

:cl:  
tweak: Rollerbeds no longer require plastic
/:cl:

### DIFF
--- a/code/modules/research/designs/medical_designs.dm
+++ b/code/modules/research/designs/medical_designs.dm
@@ -217,7 +217,7 @@
 	id = "rollerbed"
 	build_path = /obj/item/roller
 	build_type = PROTOLATHE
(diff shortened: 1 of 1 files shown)


ToasterBiome:
image
There's two right here what are u talki nbout

Sniblet:
They've never been there by the time I'm looking for rollerbeds, there are none inside of medbay proper
I didn't know those still existed which is bad

JamieD1 (approved):

PR #14494, opened 2022-06-18
you can no longer absolutely destroy inhand monkeys ( :( )

tattax:
closes https://github.com/yogstation13/Yogstation/issues/14493
closes https://github.com/yogstation13/Yogstation/issues/14498

# Document the changes in your pull request

:(

:cl:  
bugfix: you can no longer mess with inhand mobs
/:cl:

PR #14594, opened 2022-06-25
Revert "you can no longer absolutely destroy inhand monkeys ( :( )"

alexkar598:
Reverts yogstation13/Yogstation#14494

It's a code smell, theres a blacklist for this, dont do this with pre attack

No im not fixing it myself, thats more trouble than i can give, it popped up in my notifications


tattax:
it. isn't. only. for. deepfriers. inhand mobs can mess with all machinery and that's where the change was aimed at.

PR #14603, opened 2022-06-27
RTS moment

ghost:

This PR makes that aliens and blobs get a message when their teammate dies(blob gets a message when a blobbernaut dies, and aliens get a message if a humanoid alien dies). It actually makes sense, because they act as a hivemind and should feel when other hivemind member gets fucked. Also sometimes it is confusing when your blobernaut goes away to... idk... do something? and you don't understand did he die or not.

# Changelog

:cl: SuperSlayer
rscadd: Blob gets a message when a blobbernaut dies
rscadd: Alliens get a message when a humanoid alien dies
/:cl:


ynot01:
what about humans with hive nodes

ghost:
> what about humans with hive nodes

Uhhhh... Probably their also get the message

Marmio64:
Should only send the message to the queen/over mind honestly

ghost:
> Should only send the message to the queen/over mind honestly

It is ok. And in the blob part only the overmind gets the message.

JamieD1 (approved):

PR #14611, opened 2022-06-27
Replaces Kilo with GaxStation

JamieD1:
Goodnight sweet prince

:cl:  
rscadd: Replaces Kilo with GaxStation
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -34,11 +34,12 @@ endmap
 
 map kilostation
 	maxplayers 40
-	votable
+	disabled
 endmap
 
 map gaxstation
 	maxplayers 30
+	votable
 endmap
 
 map icebox


Addust:
approved by the half-insane spessman who only recently started playing yog and hasnt even unlocked any heads yet

N3D6:
Why not keep both? Kilo is a great map

stwordcitizen:
> Why not keep both? Kilo is a great map

Me when I LIE to myself

ToasterBiome (approved):
nice map

SuperTeaLove:
I don't understand the purpose of replacing a map with a fundamentally different design philosophy with a map that isn't similar to it at all. Why they can't both have a place in rotation is beyond me, and personally I think if we plan to remove Kilo from rotation it should not be combined with a new map PR.

ToasterBiome:
> I don't understand the purpose of replacing a map with a fundamentally different design philosophy with a map that isn't similar to it at all. Why they can't both have a place in rotation is beyond me, and personally I think if we plan to remove Kilo from rotation it should not be combined with a new map PR.

Kilo was added in the removal of another map (Omega) both of which served different purposes. Kilo is a midpop map pretending to be a lowpop one. Gax was created specifically to address the shortcomings of Kilo while keeping lowpop in mind and certain design decisions of Kilo.

SuperTeaLove:
EDIT: It's been explained to me a little more in depth by Baiomu and I imagine my lack of clarity is due to solely me. Seems a lot of love for Kilo went into Gax and it's good to see it's not a case of Kilo going away but it being spiritually upgraded.

I was not aware that this map was designed specifically to address design flaws in Kilo while maintaining certain features. In all discussions of this map with you Baiomu, and others, I never once got the impression that was the idea behind this map's implementation or that Kilo would be removed to make room for this map. I guess I will defer to those with mapping experience and who've taken the time to examine these maps in more depth.

As a general statement I think Kilo had potential of its own just like Omega was not an entirely imperfect map in concept. Perhaps this is simply the age old human stubbornness to accept change at work. Either way I appreciate clarification on this map's design goal and can appreciate the fact Kilo was intended to be removed.

TheRyeGuyWhoWillNowDie:
rip bozo

PR #14559, opened 2022-06-22
Research now requires Toxins/Robotics Access with Doors

JamieD1:
Research now requires Toxins/Robotics Access and now has Doors
Also updates some windows and fixes access on a couple doors.

HoS is now trespassing and can be arrested for going in R&D without cause.

:cl:  
rscadd: Research now requires Toxins/Robotics Access and now has Doors
/:cl:

PR #14634, opened 2022-06-28
Revert "Research now requires Toxins/Robotics Access with Doors" because the 2 doors are hella ugly, and the issue is solved by adding an access requirement to the console, like jamie already did

Bibby0110:
Reverts yogstation13/Yogstation#14559

i dont care jamie


JamieD1:
Fuck off Bibby

PR #8320, opened 2020-04-28
Changes to Rad Storm

JamieD1:
### Intent of your Pull Request

Add auto maint access to rad storm,

Also makes it so tcomms is down during it

#### Changelog

:cl:  
rscadd: Adds auto access to rad storms and removes comms
/:cl:

PR #14658, opened 2022-06-30
Makes Radiation Storm Event more dangerous

redmoogle:

Currently as it is a radstorm will warn you for about 40 seconds plus a few unknown seconds(Event code)

This makes it so you only have 25 seconds to run into maint making the event atleast *somewhat* dangerous
Also gives it the ability to happen another time

# Wiki Documentation
Above

# Changelog

:cl:  
tweak: Radiation storm is more dangerous
/:cl:

### DIFF
--- a/code/datums/weather/weather_types/radiation_storm.dm
+++ b/code/datums/weather/weather_types/radiation_storm.dm
@@ -3,7 +3,7 @@
 	name = "radiation storm"
 	desc = "A cloud of intense radiation passes through the area dealing rad damage to those who are unprotected."
 
-	telegraph_duration = 20 SECONDS
+	telegraph_duration = 25 SECONDS
 	telegraph_message = span_danger("The air begins to grow warm.")
 
 	weather_message = span_userdanger("<i>You feel waves of heat wash over you! Find shelter!</i>")
--- a/code/datums/weather/weather_types/radiation_storm.dm
+++ b/code/datums/weather/weather_types/radiation_storm.dm
@@ -3,7 +3,7 @@
 	name = "radiation storm"
 	desc = "A cloud of intense radiation passes through the area dealing rad damage to those who are unprotected."
 
-	telegraph_duration = 400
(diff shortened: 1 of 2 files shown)


KillerOrcaCora:
My least favorite random event made even more annoying waaaaaaaaaaaaaaaaaaa

VaelophisNyx:
I think my only issue with this is the cases where command and the AI drop the ball and fail to get maint access set to public *before* the storm hits

SuperTeaLove:
This is going to sound like the most shallow critique but twenty seconds feels too short. I agree the current amount of time you have to hide is too much but I would prefer a meet in the middle approach of 25-30 seconds. While rad effects are easily cured by Mutadone you have to consider just how drastically it can affect a player's round when their character is altered in a significant way. Even basic things like Hair Color, gender and race can be drastically shifted in a rad storm and not easily repaired the way some mutations can.

All that said I wouldn't hate a 20 second timer personally and love the idea of more station events being lethal or harder to avoid outright.

VaelophisNyx:
Another thing to consider is this also fucks over antags if they cant get to maint in the now halved timer :)

redmoogle:
What if I only allowed it to spawn with a functioning AI on the station?

neeshacark:
> What if I only allowed it to spawn with a functioning AI on the station?

coding it to only spawn with an ai is easy

coding it to only spawn if the ai has more than the single braincell required to enable emaint is impossible.

coding it to only spawn with an ai is a good idea, we will just have to hope the braincells exist.

Georgeur:
looks fine. is the timer too low *shrugs
at least we can agree its only a minor annoyance right now so, good change probably 
max_occurrences being higher actually provides some unpredictability to long time players.

JamieD1:
https://github.com/yogstation13/Yogstation/pull/8320

:thonk:

JamieD1 (approved):

PR #14662, opened 2022-06-30
Round Time, Map, Alert Level in Hub Entry, Removes Banner Image

13spacemen:
Adds round time, current map, alert level to the hub entry, it shows below the discord and forum links

removed the banner image, because it took of tons of space and you can't even see it in the byond launcher

new players all use the byond launcher not the byond ss13 webpage, so the banner image is pretty much pointless

## Why?

It's useful being able to see this information it helps players make informed decisions when joining the server

Another thing is that these are dynamic elements of the hub entry, they'll never always be the same, which makes the server feel more alive, same spirit as the "We have Ice Box!" taglines

:cl:
add: The round time, current map, and alert level now show in the hub entry
/:cl:

### DIFF
--- a/yogstation/code/game/world.dm
+++ b/yogstation/code/game/world.dm
@@ -42,7 +42,10 @@ GLOBAL_LIST_EMPTY(donators)
 	var/server_name = CONFIG_GET(string/servername)
 	if (server_name)
 		s += "<br><b>[server_name]</b> &#8212; New Player Friendly &#8212; 99% Lag Free!!"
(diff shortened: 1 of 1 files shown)


alexkar598:
we are already over the character limit

Bibby0110:
we are already over the character limit

13spacemen:
If we are over the character limit, how about I remove the banner image? You can't even see it in the byond launcher, and 99% of people launch the game via byond and not via browsing to the byond website
Could also remove the alert level since that's not as important as the other 2

JamieD1:
Lets try it, but im sure this will cut off our tagline now.

JamieD1 (approved):

PR #14713, opened 2022-07-05
st credit is now fair to all

ynot01:

St. Credit bible now takes 10% of your cash instead of 10 credits, which means the chaplain can strategize around ~~taking~~ receiving donations from richer patrons and no longer robs the poor ones blind in comparison.

It used to heal 10 to those that could not donate and 10+10 to those that could, strangely.

Now, with a 15 second cooldown introduced (to make sure chaplains can't drain bank accounts instantly), healing is set to 30 and now only heals those that can donate (anyone with more than 10 credits in their account).

P.S. to anyone that thinks 10% is too much, you have to have more than 100 credits for it to even take more than it did previously, which is more money than you will ever need, so shut up top one percenter!!!

Should make the impossibly high goal of 10k credits a little more achievable

BTW economy rework when? @TheGamerdk 

# Changelog

:cl:  
tweak: St. Credit's bible now takes 10% of your money but has a cooldown
/:cl:

### DIFF
--- a/code/modules/religion/religion_sects.dm
+++ b/code/modules/religion/religion_sects.dm
@@ -190,6 +190,8 @@
 					  /datum/religion_rites/looks)
 
 /datum/religion_sect/capitalists/sect_bless(mob/living/L, mob/living/user)
+	if(!ishuman(L))
+		return
 	if(world.time < last_dono) // immersion broken
 		user.visible_message(span_notice("You are getting too greedy! You can recieve another donation in [(last_dono - world.time)/10] seconds!"))
 		return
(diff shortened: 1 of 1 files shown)


Marmio64:
Doesn't ishuman just check if you're carbon/human, which includes all species? Wouldn't it not change anything since no other mob can have cash?

ynot01:
> Doesn't ishuman just check if you're carbon/human, which includes all species? Wouldn't it not change anything since no other mob can have cash?

u right

JamieD1 (approved):

PR #14714, opened 2022-07-05
Sterilizes the codebase - Pt.1 Removes Hand Holding

JamieD1:
Removes Hand Holding

:cl:  
rscdel: Removes Hand Holding
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -279,7 +279,7 @@
 
 		log_combat(src, M, "grabbed", addition="passive grab")
 		if(!supress_message && !(iscarbon(AM) && HAS_TRAIT(src, TRAIT_STRONG_GRABBER)))
-			visible_message(span_warning("[src] has grabbed [M][(zone_selected == "l_arm" || zone_selected == "r_arm")? " by their hands!":" passively!"]"))
+			visible_message(span_warning("[src] has grabbed [M] passively!"))
 		if(!iscarbon(src))
 			M.LAssailant = null
 		else


Spacemanspark:
Farewell, Jamie. The felinids have already closed in on your position. They've prepared their barrage of claws and bath water to fire upon your doorstep. You will be remembered as a hero.

ToasterBiome:
Bibby tier troll PR to prove a point

ToasterBiome (changes requested):
change it to not be cringe

code/modules/mob/living/living.dm
```diff
@@ -279,7 +279,7 @@
 
 		log_combat(src, M, "grabbed", addition="passive grab")
 		if(!supress_message && !(iscarbon(AM) && HAS_TRAIT(src, TRAIT_STRONG_GRABBER)))
-			visible_message(span_warning("[src] has grabbed [M][(zone_selected == "l_arm" || zone_selected == "r_arm")? " by their hands!":" passively!"]"))
+			visible_message(span_warning("[src] has grabbed [M] passively!"))
```
ToasterBiome, line 282:
```suggestion
			visible_message(span_warning("[src] has grabbed [M][(zone_selected == "l_arm" || zone_selected == "r_arm")? " by their hands!":" passively!"]"))
```

I think this would be better

ToasterBiome (changes requested):

JamieD1:
> 

This is not a troll.

ToasterBiome:
> > 
> 
> This is not a troll.

image

Sgtdevel:
I have one question what is the point of this PR other than removing hand holding like what's the valid reason for it? I don't see anything lewd with handholding and I don't understand how you can find it lewd unless you're actually trolling or meming.

Smudged:
This is certainly just a dumb PR.

If you want to remove hand holding, it' just doesn't make sense.

VaelophisNyx:
Would love to hear from Jamie why this one is on the chopping block.

SomeguyManperson:
This should have honestly been an emote both parties have to use or so ething since it ends up being used more accidentally than for whateve purpose its meant

Originally this was added to tg bc "funny mad playerbase" or whatever

Smudged:
> This should have honestly been an emote both parties have to use or so ething since it ends up being used more accidentally than for whateve purpose its meant
> 
> Originally this was added to tg bc "funny mad playerbase" or whatever

The way I see it, I don't think Hand-holding should be treated as malicious to make people mad or whatever.

and I don't see it as a big problem even if you're "accidentally" holding someone's hand, since if you're grabbing someone you're looking to get them somewhere anyways, the context is the same, but where you grab them is different. 

Removing it, just makes fewer things you can do in specific context area's and I think that is a loss.

JamieD1:
So lets make it so both parties have to do it, simple,

KillerOrcaCora:
oops

PR #14716, opened 2022-07-05
Sterilizes the codebase - Pt.3 Removes Wound Licking

JamieD1:
Removes Wound Licking

:cl:  
rscdel: Removes Wound Licking
/:cl:

### DIFF
--- a/code/datums/wounds/slash.dm
+++ b/code/datums/wounds/slash.dm
@@ -147,42 +147,6 @@
 	else if(istype(I, /obj/item/stack/medical/suture))
 		suture(I, user)
 
-/datum/wound/slash/try_handling(mob/living/carbon/human/user)
-	if(user.pulling != victim || user.zone_selected != limb.body_zone || user.a_intent == INTENT_GRAB || !iscatperson(user) || !victim.can_inject(user, TRUE))
-		return FALSE
-
-	if(INTERACTING_WITH(user, victim))
-		to_chat(user, span_warning("You're already interacting with [victim]!"))
-		return
-	if(user.is_mouth_covered())
-		to_chat(user, span_warning("Your mouth is covered, you can't lick [victim]'s wounds!"))
-		return
-	if(!user.getorganslot(ORGAN_SLOT_TONGUE))
-		to_chat(user, span_warning("You can't lick wounds without a tongue!")) // f in chat
-		return
-	lick_wounds(user)
-	return TRUE
-
-/// if a felinid is licking this cut to reduce bleeding
-/datum/wound/slash/proc/lick_wounds(mob/living/carbon/human/user)
-	// transmission is one way patient -> felinid since google said cat saliva is antiseptic or whatever, and also because felinids are already risking getting beaten for this even without people suspecting they're spreading a deathvirus
-	for(var/datum/disease/D in victim.diseases)
-		user.ForceContractDisease(D)
-
-	user.visible_message(span_notice("[user] begins licking the wounds on [victim]'s [limb.name]."), span_notice("You begin licking the wounds on [victim]'s [limb.name]..."), ignored_mobs=victim)
-	to_chat(victim, "<span class='notice'>[user] begins to lick the wounds on your [limb.name].</span")
-	if(!do_after(user, base_treat_time, victim, extra_checks = CALLBACK(src, .proc/still_exists)))
-		return
-
-	user.visible_message(span_notice("[user] licks the wounds on [victim]'s [limb.name]."), span_notice("You lick some of the wounds on [victim]'s [limb.name]"), ignored_mobs=victim)
-	to_chat(victim, "<span class='green'>[user] licks the wounds on your [limb.name]!</span")
-	blood_flow -= 0.5
-
-	if(blood_flow > minimum_flow)
-		try_handling(user)
-	else if(demotes_to)
-		to_chat(user, span_green("You successfully lower the severity of [victim]'s cuts."))
-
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
How was this even allowed into the codebase in the first place?

Smudged:
Small change, I don't care, but it's a bad change.

VaelophisNyx:
...ok so I wasn't aware you could lick *other people's* wounds

what
why

GraveHat:
An unused feature ported downstream discovered completely on accident gets axed.

SomeguyManperson:
Unfortunately you cannot kill people for doing this on yogs so this is necessary :(((

Hislyna:
I don't really see it as fetishy, It's a nice added touch to the flavour of felinids, and other than personally using it alot, it's not the worst thing that can go if it does.

LazennG:
I'm biased despite never using this myself but I thought it was an alright feature since first seeing it a while back and I don't recall anybody getting weird with it in my like 5 times of seeing it happen. It's mild and doesn't seem to have much potential for abuse (idk how wounds work) so I'm not a fan of the removal.

Smudged:
Change it so you can only lick your own wounds, and lord forbid even add a negative to doing it, like extra bleed (Suggested by Sauce cat), This would solve most problems with anybody being weirded out from felinids licking.

VaelophisNyx:
so looking at it, it's also a very strong way to negate bleeding if you lack sutures and happen to have a felinid nearby

Marmio64:
since when did felinids get free unlimited clotting to use on other people, when the race has zero tangible downsides?

Smudged:
> since when did felinids get free unlimited clotting to use on other people, when the race has zero tangible downsides?

This why I really suggest, felinids only being able to lick and benefit from licking their own wounds.

Also, the "donor race can have no positive features" is a laughable joke Imo, you're paying for this, so you might as well get something from it.

Marmio64:
> > since when did felinids get free unlimited clotting to use on other people, when the race has zero tangible downsides?
> 
> This why I really suggest, felinids only being able to lick and benefit from licking their own wounds.
> 
> Also, the "donor race can have no positive features" is a laughable joke Imo, you're paying for this, so you might as well get something from it.

i dont think you can lick your own wounds at all right now, and that allowing people to do that would make the situation worse since now you can directly powergame yourself rather than just helping other people.

pay-to-win aside, any race can have positive features but they should be balanced with some sort of downside.

Smudged:
> pay-to-win aside, any race can have positive features but they should be balanced with some sort of downside.

As I've suggested earlier, make it so felinids can only lick their own wounds, and add a negative. 
Also if it's any consolation, licking wounds doesn't heal, that's a misconception, it only stops bleeding.  

Worse case, make it slower to stop bleeding.

Smudged:
(Also to preface: The stop bleeding rate is 0.5) 
I don't think needs to be removed at all, but yeah, just make it so felinids can only lick their own wounds, if people think it's so weird

maxion12345:
If someone licks my wounds instead of just using gauze, I'm beating them to crit on the spot

JamieD1:
How about we make it only usable on yourself, sounds good.

Georgeur:
No this has always been retarded, should be axed. 
Mechanically I even hate it, force disease transmission, and less effective than simply bandaging or cauterising, meaning its needlessly risky as a wound treatment.

Marmio64:
> How about we make it only usable on yourself, sounds good.

0.5 bloodloss removal every 3 seconds which is unlimited without downsides and can be used on yourself is actually decently powerful, especially considering it's free (though its not very common for you to need to use it, but still). A weeping avulsion starts off with 4.25 bloodloss, for reference. If you're going to do this, maybe reduce it to 0.2 bloodloss removal or something so it's quite a bit worse than suturing youself (which for reference heals 0.6 bloodloss at the same rate).

Smudged:
I would honestly also suggest bringing this up to Donor if that's even possible, or to the opinions of people who actually play Felinids to make the choice.

Marmio64:
> I would honestly also suggest bringing this up to Donor if that's even possible, or to the opinions of people who actually play Felinids to make the choice.

well the people playing the race are going to be biased since they benefit from this remaining good. In any case, the PR is public, they can just look at it and comment if they want.

Hislyna:
Well my question is are people really getting freaked out about it? I found it more balanced it not being useable on yourself but works on others, but if people are really freaked out about getting their bleeding stopped by a felinid licking you then sure make it self only

SomeguyManperson:
On further introspection I believe this is roughly on par with CPR wherein people _can_ be weird about it but it's not inherently weird in that way

Either it should be removed entirely or left as is, if it's just a self care thing it's just a personal upgrade for picking a race with no particularly present downsides, and if it's staying the treatment time should probably be increased to separate it further from proper medical care

Sniblet:
I've used this like twice, it's slow and people generally panic and run when a felinid passive grabs them, but anything that can make a weeping avulsion go away instead of juiceboxing you is pretty big for a race that's supposed to be slightly worse than human
I don't get the fetish content thing but maybe I've seen too much anime

Georgeur:
At the least remove the force infection thing cause that’s powergamey as fuck.

SomeguyManperson:
> At the least remove the force infection thing cause that’s powergamey as fuck.

How it infects the catgirl not the (un)willing participant

JamieD1:
Sorry, its just weird.

PR #14717, opened 2022-07-05
Sterilizes the codebase - Pt.4 Removes putting collars on Humans

JamieD1:
Removes putting collars on people

:cl:  
rscdel: Removes putting collars on Humans
/:cl:

### DIFF
--- a/code/modules/clothing/neck/_neck.dm
+++ b/code/modules/clothing/neck/_neck.dm
@@ -181,7 +181,7 @@
 
 /obj/item/clothing/neck/petcollar/mob_can_equip(mob/M, mob/equipper, slot, disable_warning = 0)
 	var/mob/living/carbon/C = M
-	if(C && !islizard(C))
+	if(C && !ishumanbasic(C))
 		return FALSE
 	return ..()
 
--- a/code/modules/clothing/neck/_neck.dm
+++ b/code/modules/clothing/neck/_neck.dm
@@ -181,7 +181,7 @@
 
 /obj/item/clothing/neck/petcollar/mob_can_equip(mob/M, mob/equipper, slot, disable_warning = 0)
 	var/mob/living/carbon/C = M
-	if(C && ishumanbasic(C) && !iscatperson(C))
+	if(C && ishuman(C))
 		return FALSE
 	return ..()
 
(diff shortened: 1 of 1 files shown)


Sgtdevel:
Wait does this mean we can still collar lizards why not just remove the feature completely?

Smudged:
Collars can also be construted as Chockers, and we have rules against ERP.

VaelophisNyx:
> Collars can also be construted as Chockers, and we have rules against ERP.

clothing is not ERP. If you want to make this argument, remove the sexy mime outfit and ladies (kinky) underwear.

Hislyna:
I'd say don't make it felinid exclusive but allow everyone to wear them instead of removing it altogether

Smudged:
> > Collars can also be construted as Chockers, and we have rules against ERP.
> 
> clothing is not ERP. If you want to make this argument, remove the sexy mime outfit and ladies (kinky) underwear.

Just to make sure, I don't want collars to be removed.

maxion12345:
does seem pretty fetishy and beyond the realm of "hey this jumpskirt goes well with this scarf" and into "uwu what's this notices your [REDACTED] [DATA EXPUNGED]"

JamieD1:
Tell me a reason why it should be in the game though.

Georgeur:
I like the noise it makes when I walk around, don't have any justification for it than that.

Marmio64:
> I like the noise it makes when I walk around, don't have any justification for it than that.

surely we could just add some bell boots or a jingly elf hat that makes the sound without needing to put pet collars on people

Hislyna:
> Tell me a reason why it should be in the game though.

Personally I believe it is because it is just a clothing item that quite a few people wear in public. Make it wearable by everyone and there shouldn't be a problem. If its too much of a problem cause its a "pet collar", make a duplicate item and just call it a collar

SomeguyManperson:
I need this feature to assert dominance over the lower classes

Sniblet:
Now this is fetish content
Never had anything creepy happen while wearing it though

code/modules/clothing/neck/_neck.dm
```diff
@@ -181,7 +181,7 @@
 
 /obj/item/clothing/neck/petcollar/mob_can_equip(mob/M, mob/equipper, slot, disable_warning = 0)
 	var/mob/living/carbon/C = M
-	if(C && ishumanbasic(C) && !iscatperson(C))
+	if(C && ishuman(C))
```
JamieD1:
```suggestion
	if(C && !islizard(C))
```

PR #14743, opened 2022-07-07
Lightbulbs have a small chance to burn out

redmoogle:

Makes lightbulbs have a rare chance to burn out ~~Janitor Nerf~~
On average they last about ~80 minutes
(about 1.1 whole station replacement in a round)

# Wiki Documentation

Lights can burn out now

# Changelog

:cl:  
rscadd: Lights now have a small chance to burn out randomly
/:cl:

### DIFF
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -417,7 +417,7 @@
 /obj/machinery/light/process()
 	if (!cell)
 		return
-	// 0.04% Chance to burn out each tick
+	// 0.02% Chance to burn out each tick
 	if(on && prob(0.02))
 		burn_out()
 	if(has_power())
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -292,7 +292,6 @@
 // create a new lighting fixture
 /obj/machinery/light/Initialize(mapload)
 	. = ..()
-
 	if(!mapload) //sync up nightshift lighting for player made lights
 		var/area/A = get_area(src)
 		var/obj/machinery/power/apc/temp_apc = A.get_apc()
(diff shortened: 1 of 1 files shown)


Marmio64:
it sounds like a headache to deal with the fact that the majority of the lights on the station are going to statistically burn out at least once during the short duration of your stay on the station, on top of the already existing electrical storms that can destroy an entire department's lights in one go.

redmoogle:
> it sounds like a headache to deal with the fact that the majority of the lights on the station are going to statistically burn out at least once during the short duration of your stay on the station, on top of the already existing electrical storms that can destroy an entire department's lights in one go.

Statistically 3 times during a normal full length round. I can change this to 2 or 1 as well. I just want janitors to be more relevant overall

Marmio64:
> Statistically 3 times during a normal full length round. I can change this to 2 or 1 as well. I just want janitors to be more relevant overall

i dont feel like making lights always be burning out is the way to go, especially since it'll be really annoying whenever there is not a janitor, or a janitor who is slower or less competent and cant keep up with both messes and lights, even if you can make janiborgs. Should be lowered to a max of 0.01% chance per tick honestly, so it'd just be something around half the station's lights going out over the course of an hour and a half

code/modules/power/lighting.dm
```diff
@@ -415,10 +416,13 @@
 
 /obj/machinery/light/process()
 	if (!cell)
-		return PROCESS_KILL
+		return
+	// 0.04% Chance to burn out each tick
```
ToasterBiome:
```suggestion
	// 0.02% Chance to burn out each tick
```

ToasterBiome (changes requested):
Me when 99% of the station is dark because janitor doesn't have access to these areas

Moggram:
going to get myself broged if can't get AA to fix lights....

FatalEYES:
Yeah, this won't cause people to break into the service lathe to just print a light replacer, that'd be absurd.

JamieD1:
Sorry, enough lights break and they dont get replaced at current, dont want to make it worse.

PR #14757, opened 2022-07-09
Makes syndicate revolver not nukie only

ghost:

Makes syndicate revolver not nukie only. This will give .357 bullets in hacked autholates some actuall usage to someone who isn't a detective, and give non-stealth traitor a way to go not through melee combat(yeah, they can buy a stetchkin, but the problem of it is that they can't normally obtain ammo for it besides their uplink with limited TC). Also it costs 13 TC, meaning that if you buy it you cannot afford to buy adrenals, so it seems ok for me.

# Wiki Documentation

Write that syndie revolver isn't a nukie only item

# Changelog

:cl: SuperSlayer
tweak: syndicate revolver is now availible through a traitor uplink
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -536,7 +536,6 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	item = /obj/item/gun/ballistic/revolver
 	cost = 13
 	surplus = 50
-	include_modes = list(/datum/game_mode/nuclear)
 
 /datum/uplink_item/dangerous/foamsmg
 	name = "Toy Submachine Gun"


N3D6:
Syndicate revolver is hilariously op, it’s literally a 3 shot crit. Antag powercreep

ghost:
> Syndicate revolver is hilariously op, it’s literally a 3 shot crit. Antag powercreep

You have a 8 TC esword that is 4 shot crit, doesn't use ammo and blocks bullets. And also the revolver can't be used for murderboning as a traitor because of rules, and requires more shots to bring down an armored dude.

Marmio64:
revolver is ranged and can be used as soon as you pull it out, unlike the e-sword which you have to activate first. In combination with the fact it has a quicker kill time, its too frustrating to go up against stealth traitors who use it since you'll be standing there and then boom you're dead. In addition, for stechkins you can print 10mm ammo boxes from hacked autolathes and just refill your empty mags with the boxes, which is actually way easier and faster than carrying loose revolver rounds and slotting them in one by one, so no its not any harder to obtain ammo for it than the revolver. Previously when revolvers were around for traitors, it was also especially frustrating for IAA, you pretty much HAD to get sleeping carp so you could contend with the other IAA who would buy revolvers, since they were so powerful.

maxion12345:
we've gone full circle


anyways this was removed from traitor because holy shit it's overpowered

SomeguyManperson:
Revolver was serial nerfed then removed for a good reason that reason is it deals high damage and only sacrifices 2-3 bullets compared to the stechkin. It may cost more but that doesnt change how effective it is at clearing a room considering nailing someo ne with even one bullet puts them into slowdown

ghost:
I will check TG for some revolver nerfs(because TG has it as a traitor item, and maybe for some reason), and include them here if i will find.

ghost:
Okay then

PR #14760, opened 2022-07-09
Radstorms will trigger emergency maintaince with no living heads or AIs

redmoogle:

please help jamie is turning me into his C*der slave

# Wiki Documentation

Automaint for no heads and AIs during radstorms

# Changelog

:cl:  
rscadd: If there is no heads and AIs are dead radiation storms will trigger emergency maintenance
/:cl:

### DIFF
--- a/code/datums/weather/weather_types/radiation_storm.dm
+++ b/code/datums/weather/weather_types/radiation_storm.dm
@@ -50,6 +50,8 @@
 		return
 	priority_announce("The radiation threat has passed. Please return to your workplaces.", "Anomaly Alert")
 	status_alarm(FALSE)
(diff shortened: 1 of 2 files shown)


Bibby0110:
Probably shouldn’t check for living heads. Just heads on the manifest. Opens up the door for unnecessary metA

ynot01:
with the recent PR making it trigger after a mere 25 seconds, it is very unlikely that two heads (if two even exist and it isnt just one unlucky CMO) will communicate to each other and then reach their offices and then swipe before 25 seconds hits, i would check for captain/AI specifically as those two can turn it on by themselves

Bibby0110 (approved):

PR #14720, opened 2022-07-05
Sterilizes the codebase - Pt.7 Removes Catsip

JamieD1:
Sterilizes the codebase - Pt.7 Removes Catsip

:cl:  
rscdel: Removes Catsip
/:cl:

PR #14764, opened 2022-07-09
Revert "Sterilizes the codebase - Pt.7 Removes Catsip"

SomeguyManperson:
Reverts yogstation13/Yogstation#14720

Does literally nothing that can't be accomplished by other sources also forcing people to say "nya" is funny and can't be construed as fetishy bullshit

If you stupid fucks are going to add forced speech references to a stupid dwarf game I see no reason for this to be considered differently plus unlike game references that are extremely forced and will not be funny after a month at least the catsip nyaing can be used to an actual comedic effect

:cl:
rscadd: catsip is back
tweak: catsip no longer heals felinids
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/recipes/drinks_recipes.dm
+++ b/code/modules/food_and_drinks/recipes/drinks_recipes.dm
@@ -809,6 +809,12 @@
 	results = list(/datum/reagent/consumable/red_queen = 10)
 	required_reagents = list(/datum/reagent/consumable/tea = 6, /datum/reagent/mercury = 2, /datum/reagent/consumable/blackpepper = 1, /datum/reagent/growthserum = 1)
 
+/datum/chemical_reaction/catsip
+	name = "Catsip"
+	id = /datum/reagent/consumable/ethanol/catsip
+	results = list(/datum/reagent/consumable/ethanol/catsip = 4)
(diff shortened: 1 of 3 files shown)


Bibby0110 (approved):

redmoogle:
> Reverts #14720
> 
> Does literally nothing that can't be accomplished by other sources also forcing people to say "nya" is funny and can't be construed as fetishy bullshit
> 
> If you stupid fucks are going to add forced speech references to a stupid dwarf game I see no reason for this to be considered differently plus unlike game references that are extremely forced and will not be funny after a month at least the catsip nyaing can be used to an actual comedic effect

Middleground: just remove the text

SomeguyManperson:
> Middleground: just remove the text

The only thing particularly offensive about it is the healing, the text is not destructive or round ruining in anyway

JamieD1 (dismissed):
The reason its cringe is the text.

SomeguyManperson:
It is imperative that it causes at LEAST one nya. I could add a cap so it would be less likely to spam (currently it has an 8% chance which averages 1 nya per ~24 seconds)

ToasterBiome (approved):
Code looks good

SomeguyManperson:
> The reason its cringe is the text.

If the amount of times it could trigger was reduced to like once or twice would that be acceptable

JamieD1:
> > The reason its cringe is the text.
> 
> If the amount of times it could trigger was reduced to like once or twice would that be acceptable

Get on with it then

PR #14765, opened 2022-07-09
nerfs grub juice

Marmio64:
Been a long time since I meant to do this.

If you didnt know, gold grub guts can be harvested from gold grubs and turned into a really potent medicine product that also damages your liver. 

Previously healed 7,7 brute and burn while dealing 2 liver damage. Overdose limit of 10 and you got 5 units from one guts. Now heals 3,3 brute and burn while dealing 1 liver damage. Overdose limit of 20 and you get 10 units from one guts. Overall, it lasts twice as long but about half the effects (though slightly less healing), mostly because previously it gave too much healing power to miners who used it to it's full capacity.

# Changelog

:cl:  
tweak: grub juice healing nerfed.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/leather.dm
+++ b/code/game/objects/items/stacks/sheets/leather.dm
@@ -129,7 +129,7 @@ GLOBAL_LIST_INIT(xeno_recipes, list ( \
 	desc = "Gross, slimy, and green intestines retrieved from a Gold Grub. Legends say it is valuable in traditional medicines."
 	icon = 'icons/obj/mining.dmi'
 	icon_state = "goldgrubguts"
-	grind_results = list(/datum/reagent/medicine/grubjuice = 5)
+	grind_results = list(/datum/reagent/medicine/grubjuice = 10)
 
 /obj/item/stack/sheet/hairlesshide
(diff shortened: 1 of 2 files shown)


ShadowDeath6:
I'd rather keep the healing for this and just give it some other downside. Maybe increase the liver damage and also lower the OD threshold to like 5u? It could be interesting to have a very powerful healing reagent that you have to ration or else you die.

Miners already have way too many forms of unskilled free healing in the form of their survival pens + legion cores + advanced medkit brute/burn patches.

Marmio64:
> I'd rather keep the healing for this and just give it some other downside. Maybe increase the liver damage and also lower the OD threshold to like 5u? It could be interesting to have a very powerful healing reagent that you have to ration or else you die.
> 
> Miners already have way too many forms of unskilled free healing in the form of their survival pens + legion cores + advanced medkit brute/burn patches.

increasing the 'skill ceiling' for using this will just further relegate it's use for the top tier of miners, and when i made gold grub guts i more so did it for ashwalkers so they could have a medipen equivalent.

JamieD1 (approved):

PR #14719, opened 2022-07-05
Sterilizes the codebase - Pt.6 Dwarfism now permanently gives you dwarf beard

JamieD1:
Dwarfism now permanently gives you dwarf beard

:cl:  
rscadd: Dwarfism now permanently gives you dwarf beard
/:cl:

PR #14773, opened 2022-07-09
Dwarfism beards are not permanent

Sniblet:

Beards from dwarfism do not regrow instantly all the time
Reverts about half of #14719

I don't really know what you were thinking there
I don't see why you should grow a beard at all and certainly not why it should be locked onto your body as long as you want to have dwarfism, it's not like this is someone else attacking you so why should it have a malicious effect

# Changelog

:cl:  
tweak: Beards from dwarfism can be removed
/:cl:

### DIFF
--- a/code/datums/mutations/body.dm
+++ b/code/datums/mutations/body.dm
@@ -105,10 +105,6 @@
 	H.facial_hair_style = "Beard (Dwarf)"
 	H.update_hair()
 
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
Dwarves have beards, simple as. If you don't want to be a dwarf, take your mutadone.

JamieD1:
Why are you dwarfist

PR #14806, opened 2022-07-11
Ports TG geysers (the correct way)

ToasterBiome:

Ports the geyser changes TG has had since we've implemented ours.
Adds a autolathe/protolathe design datum for cargo (miners) and medical.
Switches normal plungers to be reinforced.
Gives points when you discover them and marks them on the GPS.

SuperSlayer's brings us further away from parity with TG while doing nothing new or interesting and creates more code bloat. I have however taken the idea of getting plasma from them.

Reagents available:
Ash
Stable Plasma
Oil
Protozine
Holy Water
small chance for OOOO RANDOM!!!

Closes https://github.com/yogstation13/Yogstation/pull/14779

# Wiki Documentation

Geyser section for Shaft Miners

# Changelog

:cl:  
rscadd: ports tg's geyser changes to bring us up to parity
/:cl:

### DIFF
--- a/code/datums/mapgen/CaveGenerator.dm
+++ b/code/datums/mapgen/CaveGenerator.dm
@@ -43,7 +43,7 @@
 	if(!flora_spawn_list)
 		flora_spawn_list = list(/obj/structure/flora/ash/leaf_shroom = 2 , /obj/structure/flora/ash/cap_shroom = 2 , /obj/structure/flora/ash/stem_shroom = 2 , /obj/structure/flora/ash/cacti = 1, /obj/structure/flora/ash/tall_shroom = 2)
 	if(!feature_spawn_list)
-		feature_spawn_list = list(/obj/structure/geyser/random = 1)
+		feature_spawn_list = list() //yogs no geysers in asteroid caves
 
 /datum/map_generator/cave_generator/generate_terrain(list/turfs)
 	. = ..()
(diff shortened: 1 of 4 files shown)


Cartlord:
I much prefer Tymur's idea, because - at least, as far as _I_ can tell - plungers don't do anything here, and I don't like the idea of making Miners carry around another item that only does one specific thing when they already have a K.A., a satchel, an oxygen tank, a scanner, a medipen, a full emergency box, a shovel, a plasma cutter, a plant bag, a landing zone designator, and almost certainly something else I've forgotten - their backpacks already carry enough shit.

ToasterBiome:
> I much prefer Tymur's idea, because - at least, as far as _I_ can tell - plungers don't do anything here, and I don't like the idea of making Miners carry around another item that only does one specific thing when they already have a K.A., a satchel, an oxygen tank, a scanner, a medipen, a full emergency box, a shovel, a plasma cutter, a plant bag, a landing zone designator, and almost certainly something else I've forgotten - their backpacks already carry enough shit.

This is a valid argument but also I would like to bring up the fact that half of these things you've stated are also for one purpose, and sometimes you just have to pick and choose what you have to do. Do you not take your plant bag to be able to plunge geysers for some extra points? Do you not take the landing zone designator? I think this helps further the dynamic that Shaft Miner aren't some omnipotent being that gets to do everything in the world (sans they have a bag of holding). Picking and choosing your battles is an important part of both this game and game design in general.

Plasma Magmite being rare also shows this distinction - where you are unlikely to get enough pieces to truly have everything upgraded, so you get to _choose_ what you want to use it on. I see backpack space as no different.

Optional things that miners can bring:

- Plant bag
- Field Designator
- Mining Charges
- Another KA
- Space for lugging around Plasma Magmite

These are all options and all ones where you have to pick and choose what you want to do that shift.

Mining scanners should do one (two) things, scan for minerals. I do not see how realistically pointing a scanning device at something makes chemicals spew out of it.

Cartlord:
That's an understandable claim, but it clashes with what I've seen in-game - the average Yoggite seems totally unwilling to carry anything but that which is literally necessary for mining, so the ideal scenario of "one Miner carries the designator, the second carries the plant bag, and a third carries the shovel" seems to simply _never happen_, meaning one Miner has to carry all of them, and this'll just be another thing they need to carry.

Bibby0110:
port links + credit in changelog

Bibby0110 (approved):

PR #14827, opened 2022-07-13
[Box] Total garden overhaul featuring FISHING

VaelophisNyx:
BASS PRO SHOP SPACE STATION

Adds a seed and nutrivendor to the garden, moves most of the machines around, axes all but one table and all the contents of all tables, puts 2 sets of gardening tools into a crate (bucket, spade, cultivator).
Adds water and a fishing vendor to the garden.

landmarks, pipes, and wires hidden for clarity:
ss (2022-07-12 at 07 57 34)

all shown:
ss (2022-07-12 at 07 58 46)

# Wiki Documentation

Garden page will require updating to reflect these changes.
the seed vendor is a new type that doesn't have anything but what was originally in the garden, and the nutrivendor is varedited to remove the spare botany tools from it, except the plant analyzers and bags.

# Changelog

:cl:  
rscadd: adds fishing to the box station Garden
rscadd: adds weaker MegaSeed Vendor that only has garden-approved seeds
tweak: massive overhaul for the box station Garden
/:cl:


VaelophisNyx:
topics to discuss:

- Are there too many tools, should it be taken back down to 2 pairs
- should I remove the ""dangerous"" seeds from the normal vendor stock and make it only what the tables had originally
- does there need to be botany gloves in the garden crate, which exist entirely to stop nettles from pricking you

VaelophisNyx:
went ahead and reduced the # of tools to 2 sets and took out corn and sugarcane seeds so you cant make easy bombs and get bamboo for spears/punji sticks easy without bugging botany.

GraveHat:
Consider cutting the variety of seeds and the amount of equipment back to old garden levels. If this goes through then the garden would be a bit too comparable to botany instead of being inferior even if they had to pay for seeds with their nonexistent paychecks because they could always just hack the vendor to spit out its contents instead of asking botany itself or cargo to get more seed variety.

VaelophisNyx:
I think that comes down to an admin/security issue of just
"why is the assistant allowed to hack the vendor unquestioned"

GraveHat:
> I think that comes down to an admin/security issue of just "why is the assistant allowed to hack the vendor unquestioned"

This goes for everyone as literally, no one enforces that hacking vending machines are illegal. It doesn't matter if it's technically illegal if the wider community doesn't see it as illegal, even if it is valid when it does get enforced.

ToasterBiome (changes requested):
Here's a good one: How about you make a vendor that shares the sprite with the normal seed vendor but make it dispense the seeds of the things that used to be in the garden?

It doesn't increase the power of garden, plus it can be reused in other maps.

VaelophisNyx:
I mean couldn't that just be done via var-edits?

ToasterBiome (changes requested):
image

Right window does not have plating under it please fix

VaelophisNyx:
Corrected that, another window with the same issue, and moved the tools to the table, deleting the crate.

ToasterBiome (changes requested):
conflicts

VaelophisNyx:
decided to fix this one now because it was small

JamieD1 (approved):

ToasterBiome (approved):

PR #14844, opened 2022-07-14
remove clumsiness trait add from anomalous crystal

pyronie:
remove the clumsiness trait added from the anomalous, as it is unnecessary and can be used to grief
apologies if this is bad code

# Wiki Documentation
HONK effect from the anomalous crystal no longer gives clumsy

:cl:  
tweak: makes the HONK effect from the anomalous crystal not give clumsiness
/:cl:


pyronie:
sidenote: it seems as if the crystal adding clumsy in the first place was an unintended and an oversight, due to the fact that the way it equips you is by changing your job to clown and roundstart equipping you.

pyronie:
tested in game and it works as intended

ToasterBiome:
Report to your nearest administrator if someone is using it to grief

PR #14846, opened 2022-07-14
Adds infiltrators to dynamic

ghost:

It is a strange and not cool gamemode, where:
1. All infiltrators die at the start of the game, and then the crew has a long boring shift without any antagonists
2. Infiltrators kill everyone
3. Infiltrators stealthy complete their objectives and then the crew has a long boring shift without anything interesting

So i remove the infiltration gamemode from rotation, but instead enable infiltration event and add infiltration as a dynamic ruleset(roundstart and midround). Now infiltrators are side antagonists instead of being main gamemode heroes.

# Changelog

:cl: SuperSlayer
rscadd: Added infiltrators as a roundstart dynamic ruleset
rscadd: Added infiltrators as a midround dynamic ruleset
tweak: Enabled infiltrator event
rscdel: Infiltrators gamemode is removed from rotation
/:cl:


Bibby0110 (changes requested):
Fix your title
Don't remove the gamemode from rotation if you believe it should be in dynamic
Don't add it as an event

ghost:
> Fix your title Don't remove the gamemode from rotation if you believe it should be in dynamic Don't add it as an event

Why? Inflitrators aren't good as a gamemode heroes, but it could be okay if infiltrators would be secondary antagonists, and because of this i turned them here from a gamemode to a dynamic ruleset/event.

Cartlord:
Gamerdk, why do you think it shouldn't be an event?

redmoogle:
my only gripe with this is if someone steals the infils objectives and now they have to try to find it out of x amount of people

ghost:
> my only gripe with this is if someone steals the infils objectives and now they have to try to find it out of x amount of people

This can happend to... almost any other antagonist, so I don't think that it is an issue

redmoogle:
> > my only gripe with this is if someone steals the infils objectives and now they have to try to find it out of x amount of people
> 
> This can happend to... almost any other antagonist, so I don't think that it is an issue

well *generally* theres only very little overlap in objectives unless your running dynamic

JamieD1:
Add it to dynamic but dont disable it as a gamemode.

ghost:
Okay

ghost:
Enabled infiltrators in config

yogstation/code/modules/events/infiltrators.dm
```diff
@@ -2,13 +2,13 @@
 	name = "Infiltrators"
 	typepath = /datum/round_event/ghost_role/infiltrators
 
-	weight = 0
+	weight = 8
```
Bibby0110:
```suggestion
	weight = 0
```
since we're keeping it as a gamemode, no event

Bibby0110 (approved):

config/game_options.txt
```diff
@@ -106,7 +106,7 @@ ALERT_DELTA Destruction of the station is imminent. All crew are instructed to o
 # New
 PROBABILITY DARKSPAWN 6
 PROBABILITY HERESY 4
-PROBABILITY INFILTRATION 4
+PROBABILITY INFILTRATION 2
```
JamieD1:
```suggestion
PROBABILITY INFILTRATION 4
```

PR #14598, opened 2022-06-26
removes budget card, they suck and we should have removed them in the first place

Bibby0110:
# Document the changes in your pull request

departmental budgets are for paying your staff, not buying weird crates

# Wiki Documentation

# Changelog

:cl:  
rscdel: You can no longer steal security pension fund
/:cl:

PR #14859, opened 2022-07-15
Removes unintentional free 2k in the security budget

Bibby0110:
this goal completes instantly and gives 2k

# Wiki Documentation

# Changelog

:cl:  
rscdel:  Security no longer gets 2k for free for just existing
/:cl:

### DIFF
--- a/code/modules/goals/department_goals/security.dm
+++ b/code/modules/goals/department_goals/security.dm
@@ -6,7 +6,6 @@
 /datum/department_goal/sec/nukecore
 	name = "Protect the nuke"
 	desc = "Protect the nuclear core of the station's self-destruct device, by keeping it in the device"
-	reward = 2000 // Corresponds to 50k in 50mins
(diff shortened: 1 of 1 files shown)


TheBombMakerJL:
This is just #14598 with extra steps, inaccurate PR title name, this stops the budget from generating 1000/min, not 2000 flat, and this seemed pretty intentional if you look at the comment of the code you’re removing
Reasons for not merging this is the same as #14598

Bibby0110:
> this stops the budget from generating 1000/min, not 2000 flat, and this seemed pretty intentional if you look at the comment of the code you’re removing Reasons for not merging this is the same as #14598

Wrong. This isn't a continuous department goal. This is 2000 credits given once per shift, the first time the subsystem runs. It's just 2k extra into the security budget for free, since it's physically not possible to steal the self destruct in the few seconds between roundstart and goal completion.
The comment is leftover from when the goal was made not continuous.

VaelophisNyx:
hang on didn't
you like...do this already
I remember a PR about this

PR #14879, opened 2022-07-16
Unreadied players count as 0.75 of a player for gamemode calculations

Bibby0110:
lol people are angry


redmoogle:
I was thinking more 0.8 but this is fine

ynot01:
do observers count as "unreadied"

Bibby0110:
> do observers count as "unreadied"

No
This is only for people in the lobby who aren’t ready. We’re basically guessing that 75% of people that don’t ready up are going to late join

JamieD1 (approved):

PR #14884, opened 2022-07-16
You can no longer add mecha parts (and exosuit tracking beacons) when someone is driving it

ynot01:

I always just assumed it was this way because this made sense, but you can attach parts, equipment, anything willy as it stands

You can be piloting a mech and someone just attaches an exosuit beacon and you can't do shit about it

Posis/MMIs bypass this as they cannot exit the mech

# Changelog

:cl:  
tweak: You can no longer add mecha parts (and exosuit tracking beacons) when someone is driving it
/:cl:

### DIFF
--- a/code/game/mecha/mecha_parts.dm
+++ b/code/game/mecha/mecha_parts.dm
@@ -13,6 +13,9 @@
 	if(!user.transferItemToLoc(src, M))
 		to_chat(user, span_warning("\The [src] is stuck to your hand, you cannot put it in \the [M]!"))
 		return FALSE
+	if(occupant && istype(occupant.loc, /obj/item/mmi/posibrain)) // Cannot attach parts when someone is driving, unless it is a posi
+		to_chat(user, span_warning("Someone is driving \the [M]!"))
+		return FALSE
 	user.visible_message("[user] attaches [src] to [M].", span_notice("You attach [src] to [M]."))
 	return TRUE
 
--- a/code/game/mecha/mecha_parts.dm
+++ b/code/game/mecha/mecha_parts.dm
@@ -13,7 +13,7 @@
 	if(!user.transferItemToLoc(src, M))
 		to_chat(user, span_warning("\The [src] is stuck to your hand, you cannot put it in \the [M]!"))
 		return FALSE
-	if(!istype(M.occupant?.loc, /obj/item/mmi/posibrain)) // Cannot attach parts when someone is driving, unless it is a posi
+	if(M.occupant || !istype(M.occupant?.loc, /obj/item/mmi/posibrain)) // Cannot attach parts when someone is driving, unless it is a posi
 		to_chat(user, span_warning("Someone is driving \the [M]!"))
 		return FALSE
 	user.visible_message("[user] attaches [src] to [M].", span_notice("You attach [src] to [M]."))
--- a/code/game/mecha/mecha_parts.dm
+++ b/code/game/mecha/mecha_parts.dm
@@ -15,7 +15,7 @@
 		return FALSE
 	
 	// If there is an occupant and it is not a posibrain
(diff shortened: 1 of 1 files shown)


code/game/mecha/mecha_parts.dm
```diff
@@ -13,6 +13,9 @@
 	if(!user.transferItemToLoc(src, M))
 		to_chat(user, span_warning("\The [src] is stuck to your hand, you cannot put it in \the [M]!"))
 		return FALSE
+	if(occupant && istype(occupant.loc, /obj/item/mmi/posibrain)) // Cannot attach parts when someone is driving, unless it is a posi
```
tattax:
```suggestion
	if(!istype(user?.loc, /obj/item/mmi/posibrain)) // Cannot attach parts when someone is driving, unless it is a posi
```

ynot01: This would not work, it needs to grab the occupant

VaelophisNyx:
personally I don't see the issue, other than the skill issue of letting someone get close enough to slap a beacon on you

ynot01:
> personally I don't see the issue, other than the skill issue of letting someone get close enough to slap a beacon on you

me running up to the mech and slapping the win beacon on

?????????????? how is this good for gameplay

VaelophisNyx:
well for one, its pretty much only heads or roboticists doing it
I don't think this happens very frequently anyway?

code/game/mecha/mecha_parts.dm
```diff
@@ -13,6 +13,11 @@
 	if(!user.transferItemToLoc(src, M))
 		to_chat(user, span_warning("\The [src] is stuck to your hand, you cannot put it in \the [M]!"))
 		return FALSE
+	
+	// If there is an occupant and it is not a posibrain
+	if(M.occupant && !istype(M.occupant.loc, /obj/item/mmi/posibrain))
```
tattax:
```suggestion
	if(!istype(M?.occupant?.loc, /obj/item/mmi/posibrain)) 
```

ynot01: `if(!istype(M?.occupant?.loc, /obj/item/mmi/posibrain))`
if there's no occuptant, it would return false/null, which would return true because of the `!`
which would deny putting in parts if there's no occupant
which is the opposite of what this PR aims

ToasterBiome:
There is now no way to fight a mech

ynot01:
> There is now no way to fight a mech

Truly

ynot01:
> well for one, its pretty much only heads or roboticists doing it I don't think this happens very frequently anyway?

It doesn't, this is a preventative measure

Just because someone doesn't abuse something doesn't mean it's not stupid and abusable

We shouldn't wait for someone in the small cult of people that know about this to start abusing it

VaelophisNyx:
why not just....add a use timer, so you can simply move to make it stop?

ynot01:
> why not just....add a use timer, so you can simply move to make it stop?

overtly complex workaround for what shouldn't exist in the first place

maxion12345:
1) good idea because putting mech beacons on rogue Mecha mid rampage and then spamming the emp is top kek move that we should remove eventually

2) why not make it require maintenance mode on or whatever so posis won't be vulnerable either (syndi mmi etc)

ynot01:
> 2. why not make it require maintenance mode on or whatever so posis won't be vulnerable either (syndi mmi etc)

i can see the logic but honestly it would just add inconvenience

maxion12345:
> > 2. why not make it require maintenance mode on or whatever so posis won't be vulnerable either (syndi mmi etc)
> 
> i can see the logic but honestly it would just add inconvenience

it'd make more sense to me. maint mode is the button for modifying mechs so it should be used to modify your mech

JamieD1 (approved):

PR #14893, opened 2022-07-17
Reworks the ark of the clockwork justicar

ghost:

Basically, it is very uncool when such fun gamemode lasts only 30 minutes and then the servants get wiped by security. This PR makes a very COOL thing - the ark will now open when there will be enough servants, or by a special scripture. This PR uses some code from Beestation clockcult rework(https://github.com/BeeStation/BeeStation-Hornet/pull/2124).

Also The Ark now doesn't prevent the shuttle from departing if inactive. This means that cultists can't just sit on their Z-level while building defences, and will need to hurry up if someone will call the shuttle.

Critical servant count = (POPULATION/6) + 6
87243386-945e4580-c42d-11ea-928d-3eff050f415a

# Wiki Documentation

Write that activation of the ark is no more locked to the game time

# Changelog

:cl: SuperSlayer
rscadd: Ark is now activated by the clockcult reaching the critical servant count or by a special scripture.
rscadd: Dimensional breach scripture.
tweak: Script scriptures are no activating by reaching a time mark, instead they are activated by reaching a half of the critical servant count.
tweak: The Ark no more prevents the shuttle from departing while inactive.
/:cl:


Marmio64:
i prefer the timed aspect of it rather than just giving them all the time they need to prep, get power, and build an unstoppable defence if the number of cultists they are pulling in isnt good. The timer forces them to get offensive.

ghost:
> i prefer the timed aspect of it rather than just giving them all the time they need to prep, get power, and build an unstoppable defence if the number of cultists they are pulling in isnt good. The timer forces them to get offensive.

1. Timer is bad, because it makes the round short, and time is a thing that can't be affected by players.
2. This PR also makes that the ark makes shuttle not depart ONLY if it is active, so if the cultists will be too slow they can just get fucked by a shuttlecall. So if the cult just sits at Reebe and builds defence, you can just call the shuttle, so they will be forced to open the Ark. Also in order to get power, to be able to build drfence, they nees to cog APC's... and APC's are on the station.
3. Critical servant count will prevent them from converting everyone and then just opening the ark without any enemies.

JamieD1:
Fine how it is

PR #14897, opened 2022-07-17
General RCD upgrades shouldn't cost titanium and silver

warface1234455:
# About PR
Why? It's pointless and makes no sense for just a disk

# Changelog

:cl:  
tweak: RCD upgrades no longer cost silver and titanium
/:cl:


ToasterBiome:
balance

warface1234455:
> balance

okay ur right, maybe silver is required but not titanium

warface1234455:
> balance

yes it is

JamieD1:
Not hard to come by.

PR #14903, opened 2022-07-17
reduces AI research point gain

Bibby0110:

too many points

:cl:  
tweak: AI research point gain from 8.5 -> 7.5
/:cl:

### DIFF
--- a/code/__DEFINES/ai.dm
+++ b/code/__DEFINES/ai.dm
@@ -47,7 +47,7 @@ GLOBAL_LIST_INIT(ai_project_categories, list(
 #define AI_MAX_RAM_PER_RACK 4
 
 ///How many AI research points does 1 THz generate?
-#define AI_RESEARCH_PER_CPU 5
+#define AI_RESEARCH_PER_CPU 7.5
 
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
What's the point? All AI researches are entirely dependent on whether the signal tech decides to do anything anyway, and the amount of points generated is subsequently influenced by the CPU given to the AI.

This is like a -15% change to point generation which is hardly a drop in the barrel. I'd rather the points per CPU be decreased by like half and give the AI more base RAM and CPU in its servers to work with overall because it takes way, way too long to get anything interesting without someone giving you servers. I get the upgrades are just supposed to be little fun QoL things but it's marginally less fun to have rounds where you get nothing at all.

Bibby0110:
> 

All working as intended. More upgrades bar a few (basically the HUDs) are locked behind an active network admin

ShadowDeath6:
Why does this PR matter at all, then? The signal tech can make as many new boards as he wants, 15% less research isn't going to impact that at all. You either have a signal tech or you don't.

JamieD1 (approved):

PR #14905, opened 2022-07-17
Fire breath nerf

Mqiib:
Fire breath mutation now has a fire radius of 3 tiles, 5 if empowered, and NEVER has any actual damaging explosion radius. It also has a slightly lower cooldown to make up for the reduced radius.

# Wiki Documentation

Remove the mention about it actually exploding if empowered (I think that's in there)

# Changelog

:cl:  

tweak: Fire breath mutation no longer can be given damaging explosions
tweak: Fire breath mutation explosion radius reduced to 3 baseline
tweak: Fire breath mutation cooldown reduced from 60 to 45 seconds
/:cl:

### DIFF
--- a/code/datums/mutations/actions.dm
+++ b/code/datums/mutations/actions.dm
@@ -117,7 +117,7 @@
 	name = "Fire Breath"
 	desc = "You can breathe fire at a target."
 	school = "evocation"
-	charge_max = 600
+	charge_max = 450
 	clothes_req = FALSE
 	antimagic_allowed = TRUE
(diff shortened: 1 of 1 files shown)


VaelophisNyx:
does this stop firebreath using the full fireball spell when empowered?

Mqiib:
> does this stop firebreath using the full fireball spell when empowered?

yeah it won't have the damaging explosion, just the fire and direct projectile damage

JamieD1 (approved):

PR #14913, opened 2022-07-18
Makes wizard cluwne curse respect antimagic

Mqiib:
Not sure why it ignores antimagic.

# Wiki Documentation

If there's a bit about ignoring antimagic, now it should respect it.

# Changelog

:cl:  
tweak: Chaplains now hate jokes, but only magic ones.
/:cl:


adamsong (approved):
Hugbox

PR #14927, opened 2022-07-20
makes ash drake harder

ynot01:

Increased fire direct hit damage by 10 (from 20 to 30)
Cone now fires at -50, -25, 0, 25, and 50 angles instead of just -40, 0, and 40
Mass fire now fires 6 times when low HP or enraged

# Changelog

:cl:  
tweak: Ash drake's "mass fire" fires more times when low hp
tweak: Ash drake's "fire cone" is fires more lines and leaves less gaps
tweak: Ash drake's fire does more damage
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/drake.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/drake.dm
@@ -113,7 +113,7 @@ Difficulty: Medium
 			if(2)
 				fire_cone()
 			if(3)
-				mass_fire()
+				mass_fire(12, 15, 3)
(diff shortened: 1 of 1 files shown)


ToasterBiome:
i think the cone angle addition makes it easier ngl, much easier to be within the lines then

ynot01:
> i think the cone angle addition makes it easier ngl, much easier to be within the lines then

the gaps between the lines are 25 degrees whereas previously they were 40 degrees apart, and it overall covers 20 degrees more area

JamieD1 (approved):

PR #14928, opened 2022-07-20
Makes Botanical Chem Dispensers Possible To Build, Repaths Existing Seed Vault Dispenser To A Fullupgrade Subtype, Makes Basetype Not Start Upgraded, Locks Chems Behind Tiers

VaelophisNyx:

Adds Botanical Chem Dispenser Boards to the Botanical Engineering tech node
Adds Fullupgrade subtype of Botanical chem dispensers
Makes the Seed Vault use that subtype
locks chems behind all four tiers of upgrades
makes them not init with all t4 parts now unless they're the subtype

# Changelog

:cl:  
rscadd: Added Botanical Chem Dispenser Boards to the Service Protolathe after researching the Botanical Engineering tech node
tweak: Botanical Chem Dispensers care about upgrades now
/:cl:

### DIFF
--- a/_maps/RandomRuins/LavaRuins/lavaland_surface_seed_vault.dmm
+++ b/_maps/RandomRuins/LavaRuins/lavaland_surface_seed_vault.dmm
@@ -187,7 +187,7 @@
 /turf/open/floor/plasteel/freezer,
 /area/ruin/powered/seedvault)
 "E" = (
-/obj/machinery/chem_dispenser/mutagensaltpeter,
+/obj/machinery/chem_dispenser/mutagensaltpeter/fullupgrade,
 /turf/open/floor/plasteel/freezer,
 /area/ruin/powered/seedvault)
 "F" = (
--- a/code/game/objects/items/circuitboards/machine_circuitboards.dm
+++ b/code/game/objects/items/circuitboards/machine_circuitboards.dm
@@ -680,6 +680,21 @@
 	build_path = /obj/machinery/chem_dispenser/abductor
 	def_components = list(/obj/item/stock_parts/cell = /obj/item/stock_parts/cell/high)
 	needs_anchored = FALSE
+
+//botanical chem dispenser
+/obj/item/circuitboard/machine/chem_dispenser/mutagensaltpeter
+	name = "Botanical Chem Dispenser (Machine Board)"
+	icon_state = "service"
+	build_path = /obj/machinery/chem_dispenser/mutagensaltpeter
+	req_components = list(
+		/obj/item/stock_parts/matter_bin = 2,
+		/obj/item/stock_parts/capacitor = 1,
+		/obj/item/stock_parts/manipulator = 1,
+		/obj/item/stack/sheet/glass = 1,
+		/obj/item/stock_parts/cell = 1)
+	def_components = list(/obj/item/stock_parts/cell = /obj/item/stock_parts/cell/high)
+	needs_anchored = FALSE
+
 // yogs end
 /obj/item/circuitboard/machine/chem_heater
 	name = "Chemical Heater (Machine Board)"
--- a/code/modules/reagents/chemistry/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/chemistry/machinery/chem_dispenser.dm
@@ -617,30 +617,33 @@
 /obj/machinery/chem_dispenser/mutagensaltpeter
 	name = "botanical chemical dispenser"
 	desc = "Creates and dispenses chemicals useful for botany."
-	flags_1 = NODECONSTRUCT_1
+	circuit = /obj/item/circuitboard/machine/chem_dispenser/mutagensaltpeter
 
 	dispensable_reagents = list(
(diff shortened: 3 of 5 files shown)


Cartlord:
botany is already strong as fuck what are you doing

VaelophisNyx:
this is actually a nerf. prior to this change these things did not actually care about upgrades and just had all chems unlocked, and started with t4 parts.

the only buff here is the ability to build it, which most botanists wont do, because it means leaving botany.

Cartlord:
Saltpetre's the only chem this gives that actually matters, though - it's the only way for a Botanist to get a plant's Potency above 50 without lucky mutagen rolls. Considering that the service lathe can print tier-one parts just like every other lathe, you have given every Botanist the ability to get a choice plant (maybe even two or three) to 100 Potency, which can be quite a massive upgrade if they know what they're doing. The removal of Unstable Mutagen from lower tiers is meaningless, because the Biogenerator can produce what might as well be an unlimited supply of it with about 10 minutes of setup.

VaelophisNyx:
Indeed, mutagen was mostly pointless to move higher, but to be fair, if I didn't it made the biogenerator actually pointless since you had one button to dispense as much as you want practically for free

Cartlord:
Mutagen's already pretty much free, since the biogenerator can make it and it's trivial to keep that fueled.

Cartlord:
So, to recap: this isn't a nerf to anything but GaxStation, because that's the only place the botany chem dispenser spawned that you're not upgrading. This is a machine any Botanist can throw together using shit from their own lathe, which gives them an incredible supply of saltpetre. Even if it starts with only 2000 units of chems and can't recharge at a rate that matters without upgrades, that's enough to get _20_ plants to max Potency, assuming proper usage of the other Botany machines.
As far as I can tell, the only reason to not do this literally every round is because Sci won't always research the node.
How is this balanced, to any degree?
And before you ask - no, the other chems in the dispenser don't matter. The only things in here that matter are saltpetre and mutagen, because water's in unlimited supply, a half-competent Botanist will not ever run out of fertilizer, and you can mass-produce plant-healing chems already.

VaelophisNyx:
Honestly saltpetre is a non issue
botanists already gamble for max potency with mutagen. And Gax you may not have chemists to bug for saltpetre, so this is fine
Ultimately nothing changes other than botanists now being able to make a machine that prespawns on Gax, and some out-of-rotation maps.

Cartlord:
Gambling for max Potency with mutagen is FAR different than being able to give any plant max Potency at will, to the point that I question if you've ever eve played as a Botanist.

VaelophisNyx:
I have indeed played botanist. I do not see the problem here, the scope of this PR is just making this machine not start with t4 parts, and making it replaceable if it does get broken. Botanist powergaming in sprinting to go build a botanical chem dispenser is not something I'm worried about; that's an admin issue

Cartlord:
The thing is, it's not admin-enforcable, as far as I know - I'm reasonably sure there's nothing an admin can do about a guy building something from his lathe that has an immediate and easily explained benefit to him and his department.
In addition, your claim that it's replaceable is mostly bunk, because this only spawns on a ruin (where the podpeople can't get a replacement) and Gax.
I don't think you understand the implications of "Give any _twenty plants_ a permanent increase from 50 Potency to 100 Potency".

VaelophisNyx:
and well, what happens if someone breaks the one on gax? You need to replace it

also you're a botanist, you shouldn't be doing engineering work (setting up machines)

Cartlord:
If non-Engineers were not meant to set up machines, there would not be a full set of machine parts in EVERY LATHE. If Engineers and Scientists were supposed to be the only ones building machines, only their circuit imprinters would make any machine BOARDS. You are making it possible for every Botanist to make this, and it is extremely unlikely that administrators will intervene - the last time I saw someone catch a ban for the exact thing you're describing, it was only because their department had a designated Engineer, which is not a common occurrence. You're pushing a bad idea for the sake of one map, and shrugging off every ripple of it as an "admin issue" like you aren't at-all liable for how people interact with what you add.
This is going to become something that's done every round Sci researches the tech, and the only reason that won't be every round is because most Botanist players (in my experience) aren't competent enough to even use the damn D.N.A. manipulator, so Sci will just assume the Botanists are too bad at the game to make good use of the tech.

VaelophisNyx:
Cart, you're blowing a tiny issue way out of proportion here
the difference between a pot 100 and pot 50 plant is not severe enough to care about

Cartlord:
It doubles EVERY chem inside the plant, outside of rare exceptions that require Densified Chemicals (which is not hard to get) to achieve full potential - and in addition to that, it doubles a lot of funky plant effects! That means fueling the biogenerator twice as fast, getting twice the poison from any toxic plant, getting twice as much usable medicine from any medicative plant, making banana peels & blue-tomatoes stun for twice as long, and making any plant last twice as long when smoked. At this point, I believe that either you don't understand what you're doing, or you're fully aware of all this and are playing dumb in the hopes of sneaking in a massive buff.

PR #14929, opened 2022-07-20
adds explorer firing pins, PKAs will now start with these instead of electronic ones.

TheRyeGuyWhoWillNowDie:
probably a pretty controversial one, so let's cut to the chase. explorer firing pins will prevent guns from firing while on the station z-level. miners can buy a box of 5 more for use with whatever guns security is willing to give them, and also for emagging which disables the authentication check, allowing them to fire on-station.

But why?

pkas are too good for PVP, which our current mining design philosophy has been moving away from for quite some time. they've always been too good, but instead of completely crippling them, i've opted to take the middle ground approach of making it more difficult for validhunters to spam dual 4x cooldown PKAs, with serious need they can of course get normal firing pins for their guns. every other stat with PKA remains the same.
borgs deserve some love too so they get special firing pins that lets them fire on station but only when emagged to prevent malf AIs being cucked out of the extremely useful tool of full damage PKAs as people swarm the satellite

The traitor PKA pressure mod comes in a box with an electronic firing pin so you don't have to pay an extra 6tc to use it on-station.

# Wiki Documentation

cyborg page, the shaft miner page, uplink items page.

# Changelog

:cl:  
rscadd: adds explorer firing pins, and adds them to the mining rewards vendor at the price of 3k for a box of 5
tweak: proto-kinetic accelerators will now come with explorer firing pins. cyborgs get cyborg explorer firing pins, which can be fired on station while hacked/emagged
tweak: the pressure mod for the PKA will now come in a box also containing an electronic firing pin when purchased from an uplink
/:cl:

### DIFF
--- a/code/modules/projectiles/pins.dm
+++ b/code/modules/projectiles/pins.dm
@@ -234,8 +234,7 @@
 		var/mob/living/silicon/robot/R = user
 		if(R.emagged)
 			return TRUE
-		else
-			return ..()
+		return ..()
 
 
 
--- a/code/modules/projectiles/pins.dm
+++ b/code/modules/projectiles/pins.dm
@@ -216,7 +216,6 @@
 /obj/item/firing_pin/explorer
 	name = "explorer firing pin"
 	desc = "A firing pin traditionally issued to shaft miners, to prevent malicious use of mining tools while on the station."
-	icon_state = "firing_pin"
 	fail_message = "<span class='warning'>LOCATION CHECK FAILED.</span>"
 
 /obj/item/firing_pin/explorer/pin_auth(mob/living/user)
@@ -229,12 +228,13 @@
 	name = "cyborg explorer firing pin"
 	desc = "You probably shouldn't be reading this."
 
-/obj/item/firing_pin/explorer/pin_auth(mob/living/user)
+/obj/item/firing_pin/explorer/cyborg/pin_auth(mob/living/user)
 	if(iscyborg(user))
 		var/mob/living/silicon/robot/R = user
 		if(R.emagged)
 			return TRUE
 		return ..()
+	return FALSE
 
 
 
--- a/code/modules/projectiles/pins.dm
+++ b/code/modules/projectiles/pins.dm
@@ -225,6 +225,20 @@
 		return FALSE
 	return TRUE
 
+/obj/item/firing_pin/explorer/cyborg
+	name = "cyborg explorer firing pin"
+	desc = "You probably shouldn't be reading this."
+
+/obj/item/firing_pin/explorer/pin_auth(mob/living/user)
+	if(iscyborg(user))
+		var/mob/living/silicon/robot/R = user
+		if(R.emagged)
+			return TRUE
+		else
+			return ..()
+
+
+
 // Laser tag pins
 /obj/item/firing_pin/tag
 	name = "laser tag firing pin"
--- a/code/modules/projectiles/pins.dm
+++ b/code/modules/projectiles/pins.dm
@@ -213,6 +213,18 @@
 	desc = "This is a DNA-locked firing pin which only authorizes one user. Attempt to fire once to DNA-link. It has a small explosive charge on it."
 	selfdestruct = TRUE
 
+/obj/item/firing_pin/explorer
+	name = "explorer firing pin"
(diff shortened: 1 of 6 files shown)


Marmio64:
traitors can get a mod that increases pka damage in pressurized environments. Would be nice for traitor miners to have a way to get around the firing pin problem with their PKA without just straight up giving them an option to buy new firing pins since it'd mess with the balance of mindshield pinning guns and stuff.

edit: forgot about the ability to emag the pins, but that would still be dumping 6 TC if you dont plan to use the emag for anything else, so it'd still be nice for another option.

Cartlord:
Why not just make it deal even less damage in pressure?

code/modules/projectiles/pins.dm
```diff
@@ -213,6 +213,32 @@
...
+	name = "cyborg explorer firing pin"
+	desc = "You probably shouldn't be reading this."
+
+/obj/item/firing_pin/explorer/pin_auth(mob/living/user)
+	if(iscyborg(user))
+		var/mob/living/silicon/robot/R = user
+		if(R.emagged)
+			return TRUE
+		else
+			return ..()
```
tattax:
```suggestion
		return ..()
```

VaelophisNyx:
to go with Marimo's comment, why not make the tator pka thing come with an emagged firing pin in the box too

ToasterBiome:
don't fight a miner in space

LazennG:
gus moment

N3D6:
Which miner killed you

TheRyeGuyWhoWillNowDie:
> traitors can get a mod that increases pka damage in pressurized environments. Would be nice for traitor miners to have a way to get around the firing pin problem with their PKA without just straight up giving them an option to buy new firing pins since it'd mess with the balance of mindshield pinning guns and stuff.
> 
> edit: forgot about the ability to emag the pins, but that would still be dumping 6 TC if you dont plan to use the emag for anything else, so it'd still be nice for another option.

Done, the PKA pressure mod comes in a box also containing an electronic firing pin now.

JamieD1:
Good luck stopping mining validhunters

code/modules/projectiles/pins.dm
```diff
@@ -213,6 +213,31 @@
...
+	var/turf/station_check = get_turf(user)
+	if(!station_check||is_station_level(station_check.z))
+		return FALSE
+	return TRUE
+
+/obj/item/firing_pin/explorer/cyborg
+	name = "cyborg explorer firing pin"
+	desc = "You probably shouldn't be reading this."
+
+/obj/item/firing_pin/explorer/pin_auth(mob/living/user)
```
PowerfulBacon: wrong typepath

code/modules/projectiles/pins.dm
```diff
@@ -213,6 +213,31 @@
...
+	if(iscyborg(user))
+		var/mob/living/silicon/robot/R = user
+		if(R.emagged)
+			return TRUE
+		return ..()
```
PowerfulBacon, line 236: should return false if its not a cyborg rather than null

code/modules/projectiles/pins.dm
```diff
@@ -213,6 +213,31 @@
 	desc = "This is a DNA-locked firing pin which only authorizes one user. Attempt to fire once to DNA-link. It has a small explosive charge on it."
 	selfdestruct = TRUE
 
+/obj/item/firing_pin/explorer
+	name = "explorer firing pin"
+	desc = "A firing pin traditionally issued to shaft miners, to prevent malicious use of mining tools while on the station."
+	icon_state = "firing_pin"
```
PowerfulBacon: Inherited from /obj/item/firing_pin

TheRyeGuyWhoWillNowDie: yeah

code/modules/projectiles/pins.dm
TheRyeGuyWhoWillNowDie: yeah

code/modules/projectiles/pins.dm
TheRyeGuyWhoWillNowDie: yeah

JamieD1:
I would like to push this through, but the miners would find me and shank me.

PR #14965, opened 2022-07-23
Fixes codeowners

adamsong:

Fixed a typo in the coders team name
Changed art and map to own file types rather than folders
Placed file type ownership above everything else, so that it doesn't override HD ownership


JamieD1 (approved):
Cheers, I rushed that.

PR #14972, opened 2022-07-23
OpenDream Cleanup Pass

ike709:
[OpenDream](https://github.com/wixoaGit/OpenDream) catches problems that BYOND silently ignores. This PR addresses everything OD found in Yogs.

The only real problem it found in Yogs is duplicate arg names. When you duplicate arg names, BYOND will silently use either the first or last arg and ignore the other instances, I forgot which way it goes.


adamsong (approved):
Why did someone think this would work? 
Thanks for fixing it

JamieD1 (approved):

PR #14978, opened 2022-07-24
Removes Root beer

AlvCyktor:
I hate ned that's why


adamsong:
Any specific reason you don't like it? Just don't drink it

adamsong (approved):
Approving for code quality, it does what is described, but I don't like the change

AlvCyktor:
joke

PR #14949, opened 2022-07-22
De Ligma

LoliconSlayer:
I dont like references in the game.

# Wiki Documentation

Is this machine even on the wiki?

# Changelog

Edit the changelog below to reflect the changes made by this PR, even if the changes are minor - required for every PR.
If you add a name after the ':cl', that name will be used in the changelog. Leave it empty to use your GitHub name.

:cl:  
tweak: Removed a ligma joke
/:cl:

PR #14980, opened 2022-07-24
Revert "Ligma Bad apparently - For Wiki Mord"

Tipy1802:
Reverts yogstation13/Yogstation#14967

Insta self merge with no argumentation... like c'mon I thought that was only allowed for hotfixes

Anyways the old LAM description was pretty funny and I don't see what's wrong with an innocent joke. The title of the PR also claims that it is for Wiki Mord but I thought he doesn't even log in anymore so I don't see what's his deal

### DIFF
--- a/code/game/machinery/sci_bombardment.dm
+++ b/code/game/machinery/sci_bombardment.dm
@@ -21,7 +21,7 @@
 */
 /obj/machinery/sci_bombardment
 	name = "Lavaland Artillery Mainframe"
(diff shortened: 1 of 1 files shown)


Tipy1802:
Oh also the acronym in the new description makes no sense as LAM is used for mining and not killing megafauna as they are blast resistant

JamieD1:
> 

Neither does the old one?!?!? :thonk:

JamieD1:
https://github.com/yogstation13/Yogstation/pull/14949

PR #14985, opened 2022-07-24
It has the Syndicate Saboteur module loaded.

TheRyeGuyWhoWillNowDie:
image
I think this is an oversight, because in no right mind would you have them being this obvious, especially since they're meant to be sneaky...

# Changelog

:cl:  
bugfix: the cyborg chameleon projector now actually works
/:cl:

### DIFF
--- a/code/modules/antagonists/nukeop/equipment/borgchameleon.dm
+++ b/code/modules/antagonists/nukeop/equipment/borgchameleon.dm
@@ -124,7 +124,7 @@
 	user.name = savedName
 	user.module.cyborg_base_icon = initial(user.module.cyborg_base_icon)
 	user.bubble_icon = initial(user.bubble_icon)
-	user.module.name = "Syndicate Saboteur"
+	user.module.name = initial(user.module.name)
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/nukeop/equipment/borgchameleon.dm
```diff
@@ -123,6 +124,7 @@
 	user.name = savedName
 	user.module.cyborg_base_icon = initial(user.module.cyborg_base_icon)
 	user.bubble_icon = initial(user.bubble_icon)
+	user.module.name = "Syndicate Saboteur"
```
ToasterBiome: Can you not use initial here?

ToasterBiome (dismissed):

code/modules/antagonists/nukeop/equipment/borgchameleon.dm
TheRyeGuyWhoWillNowDie: yeah

Bibby0110 (approved):

PR #14998, opened 2022-07-26
adds "partial limb coverage" and adds it to mk3/4 plates

ynot01:

## Plating

MK3 and MK4 vests now cover legs and arms with **half the armor value** that the chest has to compensate for the speed penalty

MK3 & MK4 are supposed to make you significantly tankier at the cost of speed (very valuable to sec) so there was little reason to use it since your legs and arms were completely exposed

## Behind the scenes

added a `body_parts_partial_covered` var to items which shouldn't cause problems due to the fact that checks for covered armor will use `body_parts_covered` and logically this area isnt totally covered so the code for all intents and purposes doesn't have to consider it covered for things like syringes and stuff

for every body part bitflag included in `body_parts_partial_covered` armor will be added half as much, make sure not to include body parts that are already in that clothing's `body_parts_covered` because they will stack

# Wiki Documentation

MK3/4 plates now cover legs/arms

# Changelog

:cl:  
rscadd: Added limb coverage to MK III and MK IV plates
/:cl:


TheRyeGuyWhoWillNowDie:
cool but why does sec need ert armor

VaelophisNyx:
> cool but why does sec need ert armor

ERTs don't get slowdown. This gives sec a slowdown. If you want to RP as the heavy TF2 security officer, you get to pay the price (getting beat on by someone more mobile)

Marmio64:
sounds like a really unnessecary buff when they're already pretty strong, should instead add a sleeved plat carrier variant or something that automatically makes plates give partial limb coverage but also has a slowdown of it's own, so you can also balance limb coverage and armor vs speed

ynot01:
> sounds like a really unnessecary buff when they're already pretty strong, should instead add a sleeved plat carrier variant or something that automatically makes plates give partial limb coverage but also has a slowdown of it's own, so you can also balance limb coverage and armor vs speed

that needlessly complicates things even further for sec players for what would end up at the same balance state as this

Marmio64:
> > sounds like a really unnessecary buff when they're already pretty strong, should instead add a sleeved plat carrier variant or something that automatically makes plates give partial limb coverage but also has a slowdown of it's own, so you can also balance limb coverage and armor vs speed
> 
> that needlessly complicates things even further for sec players for what would end up at the same balance state as this

Its not the same balance state because this wouldn't be a straight up buff like this pr is.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -32,6 +32,8 @@
 			var/obj/item/clothing/C = bp
 			if(C.body_parts_covered & def_zone.body_part)
 				protection += C.armor.getRating(d_type)
+			if(C.body_parts_partial_covered & def_zone.body_part)
```
Bibby0110:
```suggestion
			if((C.body_parts_partial_covered & def_zone.body_part) && !(C.body_parts_covered & def_zone.body_part))
```

Bibby0110 (dismissed):

code/modules/mob/living/carbon/human/human_defense.dm
ynot01: why not do an else if at that point

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -32,6 +32,8 @@
 			var/obj/item/clothing/C = bp
 			if(C.body_parts_covered & def_zone.body_part)
 				protection += C.armor.getRating(d_type)
+			else if(C.body_parts_partial_covered & def_zone.body_part)
+				protection += C.armor.getRating(d_type) / 2
```
Bibby0110:
```suggestion
				protection += C.armor.getRating(d_type) * 0.5
```

JamieD1 (approved):

PR #15001, opened 2022-07-26
Mining charges and Minerals go into Explorer webbing

GraveHat:
Why yes I do have several mining explosives and 50 sheets of raw plasma strapped to my chest rig.

Made it so that you can store mining charges on the explorer webbing for ease of access, mineral sheets, on the other hand, are mostly there if you just so happen to fancy using actual plasma sheets instead of plasma ore in refueling your plasma cutters, although this change does mean that any stackable mineral can be placed on the webbing if you're into picking up the heated materials that you get after fighting an ash drake.

This has been tested in a local host.

# Changelog
:cl:  
rscadd: Lets you store Mining Charges and Mineral Sheets inside Explorer Webbing
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -365,7 +365,9 @@
 		/obj/item/stack/marker_beacon,
 		/obj/item/handdrill,
 		/obj/item/jawsoflife,
-		/obj/item/restraints/legcuffs/bola/watcher
+		/obj/item/restraints/legcuffs/bola/watcher,
+		/obj/item/stack/sheet/mineral,
+		/obj/item/grenade/plastic/miningcharge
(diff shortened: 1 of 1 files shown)


Cartlord:
what if this also made it so that using your voucher on webbing came with charges, and removed the "bag of charges" voucher option?

GraveHat:
> what if this also made it so that using your voucher on webbing came with charges, and removed the "bag of charges" voucher option?

Would actually make the explorer webbing a really good voucher item for starting miners that don't want to crusher gaming. Makes a lot more sense than the duffel bag but would also need to specify to say it includes mining charges. Will consider doing it but need more input in terms of balance.

Bibby0110 (approved):

PR #15012, opened 2022-07-26
Quirk check refining

SomeguyManperson:

adds quirk exemptions from certain species and allows quirks to be removed on species change automatically, preventing species change bullshit from getting around or attempting to get around quirk exemptions

# Changelog

:cl:  
bugfix: quirks that a species can't have are no longer kept if the species is switched to after confirming quirks
bugfix: several quirks are now properly blacklisted from species they wouldn't effect
/:cl:


redmoogle:
I feel like a typecache might be more sane (because what if a quirk gets removed)

Hopekz:
>  (because what if a quirk gets removed)

Nothing negative happens

redmoogle:
> > (because what if a quirk gets removed)
> 
> Nothing negative happens

I mean codewise not someone toggled it off

Marmio64:
You mind removing hemophiliac from plasmaman too?

yogstation/code/datums/traits/negative.dm
```diff
@@ -5,4 +5,4 @@
 /datum/quirk/blooddeficiency/check_quirk(datum/preferences/prefs)
 	if(prefs.pref_species && (NOBLOOD in prefs.pref_species.species_traits)) //can't lose blood if your species doesn't have any
 		return "You don't have blood as [prefs.pref_species]!"
-	return FALSE
+	return ..()
```
redmoogle, line 8: Why `return ..()`

SomeguyManperson: I was going to add a generic blacklist and it might be useful if we need some generic quirk denial things later

Bibby0110 (approved):

JamieD1 (approved):

Altoids1:
Someone actually expanded on my check_quirk proc!! Yay!!

PR #15021, opened 2022-07-27
Gives stasis parts purpose

ynot01:

Manipulator affects how fast the bed can be turned off and on

cooldown (5 seconds at time of writing) is multiplied by (1/manipulator tier)

Capacitor affects organ decay

Previously, stasis beds would freeze organs completely

Now, organ decay (life tickrate) is slowed to 66% at tier 1, 50% at tier 2, 25% at tier 3, and 0% at tier 4

You also take more stamina damage when the bed is emagged, used to take 5 every process

Now you take 5,10,15,20 depending on the capacitor part tier

# Wiki Documentation

see above

# Changelog

:cl:  
tweak: Statis beds now have a reason to be upgraded
/:cl:

### DIFF
--- a/code/game/machinery/stasis.dm
+++ b/code/game/machinery/stasis.dm
@@ -27,9 +27,9 @@
 	// 4 life_tickrate is 25% organ decay or 75% stasis
 	// 5 life_tickrate is 20% organ decay or 80% stasis
 	var/stasis_amount = 1.5 // How much it adds to life tickrate
-	// T1 = 1.5, 33% stasis
-	// T2 = 2, 50% stasis
-	// T3 = 4, 75% stasis
+	// T1 = 0.5, 33% stasis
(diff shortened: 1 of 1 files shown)


GraveHat:
It makes parts more relevant, increases meds reliance on the pens for organ decay which subsequently should teach doctors that yes they could refill the pens now, and also makes them somewhat more reliant on chemistry and sci.

I'm mostly fine with this honestly. Unless someone increases decay rates further down the line.

ShadowDeath6:
Don't stasis beds also slow/stop blood loss as well? Make sure to have capacitors influence that as well if so.

ynot01:
> Don't stasis beds also slow/stop blood loss as well? Make sure to have capacitors influence that as well if so.

if they depend on lifeticks then they will be affected same as organ decay

JamieD1 (approved):

PR #15022, opened 2022-07-27
fixes holobeds doubling your life tickrate

ynot01:

holobeds no longer double your life tickrate by adding 1 to it, they now subtract 1 which properly sets it to 0

# Changelog

:cl:  
bugfix: fixed holobeds doubling your life tickrate instead of freezing it  
/:cl:


ynot01:
lifetick uses modulo so this works

PR #15031, opened 2022-07-27
Adds adrenaline to combat

redmoogle:

Upon taking 15 or more damage(IN A SECOND) you will get a small amount of adrenals (That just remove slowdown may add heart damage?) 

Should last about 10 seconds
After receiving the effect it wont trigger for atleast 10 minutes

Should make combat slightly more fair since getting shot means you get stuck with movement slowdown which is basically instant lose

Open to balancing changes

# Wiki Documentation

Combat takes slightly more effort now

# Changelog

:cl:  
rscadd: Getting significantly hurt will remove slowdown for a few seconds
/:cl:


Hopekz:
Interesting PR

Also adds a tactical option of not attacking your attacker so you get the movement buff vs them to get away from them

Marmio64:
15 dmg sounds too low, too often it would be wasted getting hit on accident or something. Increase it to 25 or more. Also this may have severe balance issues on antags who rely on melee and damage slowdown, so I'm unsure how good this actually is for combat.

redmoogle:
> 15 dmg sounds too low, too often it would be wasted getting hit on accident or something. Increase it to 25 or more. Also this may have severe balance issues on antags who rely on melee and damage slowdown, so I'm unsure how good this actually is for combat.

conventional stuns and slowdowns still work

Marmio64:
> > 15 dmg sounds too low, too often it would be wasted getting hit on accident or something. Increase it to 25 or more. Also this may have severe balance issues on antags who rely on melee and damage slowdown, so I'm unsure how good this actually is for combat.
> 
> conventional stuns and slowdowns still work

that still means you are restricting some antags to HAVING to use stuns or crafting bolas, while previously if you got the drop on someone as say heretic or cling you could manage to damage slowdown them in time before they can push you away and run, it was much riskier than using a stun but it gave you options while also giving the person being attacked an option to fight back (its a lot damn harder to fight against being bola'd/stunned and unable to do anything but talk)

Chubbygummibear:
would it be possible to make adrenaline tire the person out when it leaves their system. i'm pretty sure in real life people can be up for like days just riding on adrenaline and as soon as it stops they just collapse cause their body was doing 100% for the past whole time

code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -2054,4 +2054,18 @@
...
+	color = "#d1cd9a"
+	can_synth = FALSE
+
+/datum/reagent/adrenaline/on_mob_add(mob/living/L)
+	. = ..()
+	ADD_TRAIT(L, TRAIT_IGNOREDAMAGESLOWDOWN, type)
+	
+/datum/reagent/adrenaline/on_mob_delete(mob/living/L)
+	. = ..()
+	REMOVE_TRAIT(L, TRAIT_IGNOREDAMAGESLOWDOWN, type)
```
tattax:
```suggestion
/datum/reagent/adrenaline/on_mob_add(mob/living/L)
	ADD_TRAIT(L, TRAIT_IGNOREDAMAGESLOWDOWN, type)
	
/datum/reagent/adrenaline/on_mob_delete(mob/living/L)
	REMOVE_TRAIT(L, TRAIT_IGNOREDAMAGESLOWDOWN, type)
```
if you call parent it will jsut return

redmoogle: I tested this. it will just have no effect

ynot01:
~~this will make the player ignore all clothing slowdowns aswell~~

misread the trait

ynot01:
does stamina damage count

redmoogle:
> does stamina damage count



> does stamina damage count

for slowdown yes, but for triggering it, no

Moggram:
I feel like this should be a human only thing. I understand why for balancing that everyone would get a adrenaline shot, but logically it makes no sense to me.

As Hopekz said, this is a interesting PR and like to see where it goes.

Marmio64:
> I feel like this should be a human only thing. I understand why for balancing that everyone would get a adrenaline shot, but logically it makes no sense to me.
> 
> As Hopekz said, this is a interesting PR and like to see where it goes.

a human buff, and an interesting mechanic just for humans to get? No way the lizards and felinids would allow this to happen.

tattax:
how does this affect miners
how does this affect melee combat
how does this affect stamina meta

Hopekz:
> how does this affect miners how does this affect melee combat how does this affect stamina meta

All those questions can be answered with a testmerge.

Chubbygummibear:
do bear in mind that if you take the human only route it will have other reaching effects like nukies benefiting from ignoring some degree of damage slowdown in combat. people would likely opt to stay human wizard too if it meant that getting tagged by someone wouldn't slow them so they could jaunt or whatever to escape more easily

Skrem7:
As soon as you give a roundstart species the ability to ignore base game mechanics with zero downside then you are opening a wild jar of balance that you don't want to spill

I like the thematic idea for making this human only but every other species would need a buff in return

MegaEmpirical:
This is a cool idea.

Bibby0110 (approved):
Seems fine

PR #15064, opened 2022-07-30
Revert "Reduces the priority of the explosion subsystem so mobs still process while meteors are overflowing everything else" because it results in wacky behaviour where explosions can be delayed enough to have a tangible impact on gameplay

Bibby0110:
Reverts yogstation13/Yogstation#11998

remove meteors?


Kapu1178:
This is why god invented proper subsystem yielding

JamieD1:
> 

Why the revert though, could you explain it a bit more.

Bibby0110:
> 

it results in wacky behaviour where explosions can be delayed enough to have a tangible impact on gameplay
example: 
shooting welding fuel tank, doesn't explode, you push it and take its spot and then it explodes

JamieD1 (approved):

PR #15066, opened 2022-07-30
Adds Arm Laser/Disabler Implants To Illegal Cybernetic Implants Tech Node & Autosurgeons To Uplinks

VaelophisNyx:

Adds subtypes of the inaccessible arm-laser and a new arm-disabler implant; the arm laser doing less damage than normal and being less accurate, while the disabler is the cyborg variant using a different sprite

adds individual autosurgeons to the uplink for all, but excludes infils from getting the lethal laser implant. (16Tc lasers, 14tc disablers)

- [x] dial in # of shots for the laser
- [x] make them invisible to health analyzers
- [x] make them both self-charge slower
- [x] make the arm-gun EMP act not chance based

Because it's an interesting thing for traitors to maybe buy and use, but also if illegal tech gets researched ever, sec *could* print off these implants and have them implanted in themselves. While this may seem alarmingly bad, they do less damage than normal lasers; and the disabler has less shots. Additionally getting EMPed with one of these royally fucks you up, and instantly sets you on fire as well as deals some burn damage at the same time.

# Wiki Documentation

might need addition to anything that lists uplink items, and also will need to be added to the protolathe item list under the security protolathe (disabler) and armor lathe (laser), behind Illegal Cybernetic Implants tech.

# Changelog

:cl:  
rscadd: Adds Laser/Disabler Arm Implants to the uplink, and adds them to the Illegal Cybernetic Implants tech node. 
/:cl:

### DIFF
--- a/code/game/objects/items/storage/briefcase.dm
+++ b/code/game/objects/items/storage/briefcase.dm
@@ -54,3 +54,11 @@
 	..()
 	new /obj/item/autosurgeon/organ/syndicate/syndie_mantis(src)
 	new /obj/item/autosurgeon/organ/syndicate/syndie_mantis/l(src)
+
+/obj/item/storage/briefcase/armguns
+	desc = "A sleek briefcase. It has the brand symbol CYBERSUN on it"
+
+/obj/item/storage/briefcase/syndie_mantis/PopulateContents()
+	..()
+	new /obj/item/autosurgeon/organ/syndicate/laser(src)
+	new /obj/item/autosurgeon/organ/syndicate/disabler/l(src)
--- a/code/game/objects/items/storage/briefcase.dm
+++ b/code/game/objects/items/storage/briefcase.dm
@@ -54,11 +54,3 @@
 	..()
 	new /obj/item/autosurgeon/organ/syndicate/syndie_mantis(src)
 	new /obj/item/autosurgeon/organ/syndicate/syndie_mantis/l(src)
-
-/obj/item/storage/briefcase/armguns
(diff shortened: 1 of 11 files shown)


GraveHat:
It would make a lot of sense if they sapped hunger in order to charge the guns.

VaelophisNyx:
Ehhhh, I'd maybe do that but I think just slower charging is fine

GraveHat:
> Ehhhh, I'd maybe do that but I think just slower charging is fine

Borg chargers would also be a neat side thing to add for charging. That and compatibility with the capacitive potatoes while it's not drawn. Oh and preternis charging from APCs as well.

00ze-cyclone:
Give me a good reason to give syndies lasers, the stechkin already fill the role for dealing damage at range or stamina damage (soporific rounds).
One of the cool thing about lasers and disablers is that you know immediately what the projectile do at a glance, all disablers do 30 stamina, all laser do 20 burn no matter what weapon shot them (except the hellfire but at least the projectile itself is named differently), now we have another laser that look like all other lasers except it arbitrarily deal less damage for one specific weapon.
The balancing of those 2 implants is also utterly absurd: disabler implant is WAY stronger than the laser, even if that laser wasn't nerfed it would still be stronger, but it cost less TC? Though both are cosmically overpriced and absolute TC wastes (all syndies can buy an emag and crate of lasers and disablers are dirt cheap those are also only sec access so even without an emag you'll only need to get the ID of a secoff instead of the warden or HoS).
Those implants 2 implants bring nothing really interesting or new.

PR #15070, opened 2022-07-30
Atmos firesuit can now be ordered from cargo

warface1234455:
# About PR
Should be able to order them, its basically emergency firesuit but better 

# Wiki Documentation
Atmos firesuit can be ordered from cargo

# Changelog

:cl:  
tweak: Atmos firesuit can be ordered from cargo  
/:cl:

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -756,6 +756,20 @@
 					/obj/item/paper/guides/conveyor)
 	crate_name = "conveyor assembly crate"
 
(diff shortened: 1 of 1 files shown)


VaelophisNyx:
I'd argue they should cost more, since they're better

warface1234455:
> I'd argue they should cost more, since they're better

they have the same fire protection, same speed to normal firesuit so uh i think 2000 is a good price

JamieD1 (approved):

PR #15079, opened 2022-07-31
Modernizes Toxins Testing Lab Layout

Xantam-13:
Amended the map to reflect the current method of toxins testing so you don't have to set up each and every round. 

Old: 
image

New:
image

:cl:  
tweak: renovated Toxins Test Lab to reflect current method of bomb making  
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -20072,6 +20072,27 @@
 	},
 /turf/open/floor/plasteel,
 /area/security/checkpoint/auxiliary)
+"bNb" = (
+/obj/machinery/doorButtons/access_button{
+	idDoor = "toxins_airlock_exterior";
+	idSelf = "toxins_access_control";
+	name = "Toxins airlock control";
+	pixel_x = 24;
+	pixel_y = 24
+	},
+/obj/machinery/doorButtons/access_button{
+	idDoor = "toxins_airlock_interior";
+	idSelf = "toxins_access_control";
+	layer = 3.1;
+	name = "Toxins airlock control";
+	pixel_x = -24;
+	pixel_y = 24
+	},
+/obj/machinery/atmospherics/components/unary/outlet_injector{
+	dir = 4
+	},
+/turf/open/floor/engine,
+/area/science/mixing/chamber)
 "bNd" = (
 /turf/closed/wall/r_wall,
 /area/medical/virology)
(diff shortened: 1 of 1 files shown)


ToasterBiome:
Why no heater? It's used in some beginner bombs.

Xantam-13:
> Why no heater? It's used in some beginner bombs.

Can just screw-driver the cooler right?

ToasterBiome:
> > Why no heater? It's used in some beginner bombs.
> 
> Can just screw-driver the cooler right?

I'd prefer if you had both also you only need one anyway

Xantam-13:
> 

Can put it back easy enough if youd prefer?

ToasterBiome:
Yes

Xantam-13:
https://i.imgur.com/4VLyLko.png 

Replaced back with a heater

SynnGraffkin:
As someone who used to do a lot of Toxins a few years back on other sources, I think changing the starting layout like this is excessive and can be confusing to those learning it. Part of the beauty with Toxins, for those who take the time to make their own methods and mixes, is making your own pipe methods. I used to always ask for the RD suit, be it from the RD or AI, to do additional chamber piping and have it scrubbed into space piping for colder cooling before parts were done. Was this excessive for what you need in Toxins? Yes, but I enjoyed it.

While a piping setup you posted may be optimal for some players, I still think the old / default is a better starting point for new players or those who want to learn toxins.

ToasterBiome:
in regards to this I think it's a little bit too optimized that you're functionally making it the only method of doing toxins. The default setup doesn't lend well into _anything_ and I think that's part of the beauty of it. 

I would be more inclined to merge it if these were blank and the simple design of the room and the empty space _suggested_ these instead of forced it

image

ToasterBiome (changes requested):
oh need to press this button

redmoogle:
something that might be neet is to have the supplies behind a windoor closet like the one in medbay

Xantam-13:
> oh need to press this button

I reject your change request, I won't be doing that. Just because you give someone a carrot does not mean they have to eat it.

The problem as it stands right now is that the wiki isn't informative or clear enough on how to do toxins, and the current layout was designed exactly and intentionally for old toxins. All this change does is bring the lab up to par with how toxins is now. 
Of which, I have no stake in, I still don't know toxins, this PR was made at the feedback of 3 or 4 toxins users. 

The new proposed layout would greatly assist those who did wish to learn toxins, who could then tailor the pipes freely to learn alternative methods. 

In doing this, we could also discourage griefers. Right now, we get bombed quite a bit by griefers, but if they log in and see the lab already in use, they might be tempted to find less destructive methods of grief. But if they see an abandoned lab (which they often do, due to the complexity of the wiki and the poor pre-design), they are more inclined to just crack on and make juicy boom booms. 

So again, I reject your change request. I'll leave it to Jamie whether he wishes to merge this, trial this, or reject this, thank you.

JamieD1:
Looking over with Baiomu to find a solution

JamieD1:
Sorry for the delay with this, I had a brainwave on how we can redesign all of south east science because of this PR, will try get the ball rolling

Bibby0110:
Conflcited + Dev Conflict + Stale + Stale devs

Bibby0110:
Our good friends Jamie and Baiomu are redoing this with Xantams input

PR #15087, opened 2022-07-31
[EVENT] Makes fire extinguishers infinite

adamsong:

Should make water infinite, away from home atm, so can't check, don't tm unless someone does thanks


adamsong:
@ynot01 Is mean

PR #15095, opened 2022-08-01
RCD can now spit out Catwalk

warface1234455:
# About PR
When you want to move around in space or do some wiring but are too lazy to acquire lattice, you may use a catwalk instead of laying tile since it is less expensive. Tile costs 3 and Catwalk only costs 1 for RCD

# Wiki Documentation
RCD can place catwalk now

# Changelog

:cl:  
rscadd: RCD can now place catwalk!
/:cl:

### DIFF
--- a/code/__DEFINES/construction.dm
+++ b/code/__DEFINES/construction.dm
@@ -114,6 +114,7 @@
 #define RCD_FURNISHING (1<<6)
 #define RCD_CONVEYOR (1<<7)
 #define RCD_SWITCH (1<<8)
+#define RCD_CATWALK (1<<9)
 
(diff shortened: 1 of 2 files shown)


monster860:
I'm pretty sure catwalks are generally more expensive than platings when building without RCDs, since they require 2 rods instead of a rod + a tile

warface1234455:
> I'm pretty sure catwalks are generally more expensive than platings when building without RCDs, since they require 2 rods instead of a rod + a tile

ok thats kinda true but uh 1 metal into 2 rods is kinda dumb, it literally reduced the size just for two sticks

warface1234455:
whatever

warface1234455:
pointless pr anyway

PR #15107, opened 2022-08-01
Replace roundstart PDA (2010) with PDA (2021)

ynot01:

The code is all here. These were supposed to replace the old PDAs from the start.

Everyone gets an NTOS PDA in place of their old PDA that is preloaded with the chat client

Anyone that started with an NTOS item in their backpack now starts with that in their PDA slot (i.e. heads have their phone in PDA slot now), only exception being network admin who starts with a tablet because it doesn't fit in ID slot.

Clown PDA is still slippery

Old PDAs still exist in vendors and HoP line, they are just no longer roundstart; they may be removed in the future since the majority of their functions are redundant and they look and feel terrible

Biggest downside to this by far is the loss of PDA messaging system. Fortunately, NTOS has its own messaging system, and even more fortunately, the basic NTOS PDA starts with this program preloaded.

I believe that replicating the old PDA messaging system into NTOS would be a terrible idea. Think about it for a minute. Every NTOS computer registered into this arbitrary mailing list. The same effect is achieved in a far more genius way in the chat client, the largest penalty being that you have to set it up yourself instead of being handed a messaging list.

Net admin also has a reason to look over NTOS monitoring now

Donor PDA reskins will still reskin the PDA

Also some plasmamen roles were supposed to spawn with NTOS devices but never got updated to, so I went through and fixed that here

# Changelog

:cl:  
tweak: replaced roundstart PDAs with NTOS PDAs
bugfix: fixed some plasmaman roles not getting NTOS devices when their other-race counterparts would
/:cl:

### DIFF
--- a/code/modules/clothing/outfits/plasmaman.dm
+++ b/code/modules/clothing/outfits/plasmaman.dm
@@ -3,7 +3,7 @@
 	
 	jobtype = /datum/job/bartender
 
-	pda_type = /obj/item/pda/bar
+	pda_type = /obj/item/modular_computer/tablet/pda/preset/basic
 
 	head = /obj/item/clothing/head/helmet/space/plasmaman
 	r_hand= /obj/item/tank/internals/plasmaman/belt/full
@@ -20,7 +20,7 @@
 	
 	jobtype = /datum/job/cook
 
-	pda_type = /obj/item/pda/cook
+	pda_type = /obj/item/modular_computer/tablet/pda/preset/basic
 
 	head = /obj/item/clothing/head/helmet/space/plasmaman
 	r_hand= /obj/item/tank/internals/plasmaman/belt/full
@@ -35,7 +35,7 @@
 	
 	jobtype = /datum/job/hydro
 
-	pda_type = /obj/item/pda/botanist
+	pda_type = /obj/item/modular_computer/tablet/pda/preset/basic
 
 	head = /obj/item/clothing/head/helmet/space/plasmaman/blue
 	r_hand= /obj/item/tank/internals/plasmaman/belt/full
@@ -54,7 +54,7 @@
 	
 	jobtype = /datum/job/curator
 
-	pda_type = /obj/item/pda/curator
+	pda_type = /obj/item/modular_computer/tablet/pda/preset/basic
 
 	head = /obj/item/clothing/head/helmet/space/plasmaman/curator
 	r_hand= /obj/item/tank/internals/plasmaman/belt/full
@@ -72,7 +72,7 @@
 	
 	jobtype = /datum/job/chaplain
 
-	pda_type = /obj/item/pda/chaplain
+	pda_type = /obj/item/modular_computer/tablet/pda/preset/basic
 
 	head = /obj/item/clothing/head/helmet/space/plasmaman
 	r_hand= /obj/item/tank/internals/plasmaman/belt/full
@@ -88,7 +88,7 @@
 	
 	jobtype = /datum/job/janitor
 
-	pda_type = /obj/item/pda/janitor
+	pda_type = /obj/item/modular_computer/tablet/pda/preset/basic
 
 	head = /obj/item/clothing/head/helmet/space/plasmaman
 	r_hand= /obj/item/tank/internals/plasmaman/belt/full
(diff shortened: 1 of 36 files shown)


ToasterBiome:
the “mailing list” should be NTOS pdas.
Also does it have support for the chat alerts like normal PDAs at least? It it doesn’t have every feature/ease of use that one from a decade ago does I cannot approve this in good faith
What about uplinks?

ynot01:
> Also does it have support for the chat alerts like normal PDAs at least? 

Not to my knowledge

> What about uplinks?

PDA uplink depended on mailing system, which does not exist in NTOS

ToasterBiome:
> > Also does it have support for the chat alerts like normal PDAs at least?
> 
> Not to my knowledge
> 
> > What about uplinks?
> 
> PDA uplink depended on mailing system, which does not exist in NTOS

So you are going to add these things?

ynot01:
> So you are going to add these things?

Probably going to remove the PDA option from uplink menu options

ToasterBiome:
> > So you are going to add these things?
> 
> Probably going to remove the PDA option from uplink menu options

No

nmajask:
I have a MPC uplink I can PR when I can get the chance

Chubbygummibear:
we simply add Synd-azon uplink program

Cartlord:
I really don't like this, because N.T.O.S. P.D.A.s - at least, from what I recall - can have individual parts irradiated, making decontamination impossible without disassembly.

ynot01:
> I really don't like this, because N.T.O.S. P.D.A.s - at least, from what I recall - can have individual parts irradiated, making decontamination impossible without disassembly.

screwdriver and also pretty much irrelevant to this PR

Cartlord:
It's relevant because people take their P.D.A.s with them everywhere, you're removing the ability to start with a P.D.A. that can't be fucked by rads, there's no indication that you can screwdriver a new-P.D.A. to disassemble it, and there's no indication that your modular device is emitting rads unless you actually take it apart and notice the parts are glowing. This is going to lead to a lot of bullshit radiation, and likely a few deaths.
Apart from that, I think you're underestimating the power of the mailing list, because it can be used as an easy way to tell if someone is or isn't present - it's quicker to check the P.D.A. mailing list and see if, say, a Botanist is listed, than to ask on radio if you have any Botanists and get no reply because the Botanist is too busy doing plantshit to read chat.
In addition, I'd like to sincerely ask: what will this add to the game? How does the removal of current P.D.A.s make anything better, in any way?

Bibby0110:
Uplink is a deal breaker

Last I checked NTOS is missing quite a few features that PDAs have. Mainly the different programs/cartridges. Seems like a direct downgrade tbh.

In addition, have you tested the chat client? I remember it not working some time ago.

As it stands I don’t believe tablets are viable for a PDA replacement without some major work, else the switch would probably have been made years ago 

The proper way to do this would be to do what TG did, starting with https://github.com/tgstation/tgstation/pull/65755 and grabbing the numerous fixes and updates along the way

GraveHat:
At the moment, I agree that, yes, the current pdas need to be replaced eventually, but not until we've added more features of the pda to the modern one there are still some issues mentioned here that need addressing before they can replace the current pdas.

Cartlord:
What's even wrong with current P.D.A.s? The only issue I've ever had with them is when the messaging servers go down, and that's a telecomms problem.

Cartlord:
Oh, shit, something else I forgot to mention:
Modular device vendors are present on, if memory serves, _every station_; since credits are worthless (as has been discussed before) and there isn't much to do with them, this logically means that, if modular P.D.A.s were so much better than current P.D.A.s, everyone would get them, since there's so little stopping you.
Since (as far as I'm aware) this isn't what happens, I think you should reconsider whether modular devices are really as useful as you seem to believe they are.

VaelophisNyx:
my only note on that Cartlord is that noone can afford the damn things

ynot01:
unfortunately i don't know TGUI and two maintainers have told me that they aren't merging this without TGUI changes

redmoogle:
> unfortunately i don't know TGUI and two maintainers have told me that they aren't merging this without TGUI changes

tgui isnt hard its basically just HTML/JS mix

Hopekz:
> tgui isnt hard its basically just HTML/JS mix

Why not just leave it to people that know that language or make it a to-do for a future PR?
This is an open sourced project

PR #15109, opened 2022-08-01
removes security's ability to cockblock antagonist that rely on stealing weapons: revengeance

TheRyeGuyWhoWillNowDie:
**merge conflict moment**

no more printing mindshield firing pins. guns that start with them eg pulse weapons still have them, you just cant shit out billions of them from the seclathe anymore

being able to make every weapon ever essentially unlootable cucks over antagonists like vampires and BBs and such extremely badly, and completely removes the risk of people breaking into the armory to steal guns because they wont be able to use them  without opening the mindshield lockbox which is time consuming and requires specialized equipment to do as well as making you stick out like a sore thumb on sechuds. this should also help cull every security officer having 15 mindshield firing pin AEGs in their boh, since science guns are still a problem i'm going to try to address in the future, but this is a good start for it, plus there's no such thing as too many nerfs for security.

you can still print and use science guns with the fucking 10 electronic firing pins that start in the armory, if you somehow fucking run out? buy more from cargo it's 2k credits for another 10

# Wiki Documentation

tech webs page

# Changelog

:cl:  
rscdel: mindshield firing pins can no longer be created
/:cl:


VaelophisNyx:
genuinely dont get *why* this is needed
we do not need sec offs being even MORE of a lootbox. Just remove the pin and use of the the billion ways to get new ones, or fucking use your antag gear

RG4ORDR:
No more antagonist having hugbox firing pins and DNA locks!

Being able to use high damage and fast attacking weapons essentially unlootable fucks over the crew. Like security, greyshirts,cargo, and medical extremely badly and completely removes the risk of people disarming the antagonist as crew.  This should also help cull every Nuke Ops team having 15 DNA locked guns in their backpacks. Since Syndie guns are still a problem, I'm gonna try address in the future but this is a good start for it. Plus there's no such thing as too many nerfs for Antags.

Trying to solve a problem where there is none is dumb. If you're gonna remove mindshield pins from being printed then every single piece of Syndicate gear no longer has special DNA and Syndicate firing pins. If your teammate manages to die while holding an L6 then tough shit, the crew now has an L6 they can use without having to go get another firing pin and take themselves out of the fight. It's your responsibility to not lose your items to the crew.

LoliconSlayer:
i ded

00ze-cyclone:
1)fuck vamp with its lame ass jaunt going into armory

2)if BB get in the armory then that’s generally because there’s no sec in the first place so mindshield pin don.t matter there

3)heretic need a gun but we should really just give them a magic one

If you want to stop secoffs with 1000 aeguns then make it actually deal radiation to the user instead of that random removal fixing nothing.

TheRyeGuyWhoWillNowDie:
> genuinely dont get _why_ this is needed we do not need sec offs being even MORE of a lootbox. Just remove the pin and use of the the billion ways to get new ones, or fucking use your antag gear

the billions of ways to get new ones:
1: cargo

Hopekz:
Why not just give the antagonists that rely on stealing weapons a better way to do it?
Something that is actually effective at what you're trying to do for them vs RNG disarm spam that makes them look like greytiders

neeshacark:
why not just like, give the antags that 'rely on gun theft' a recipe to make a makeshift firing pin so they can remove the pin and craft a makeshift one instead of removing pin crafting? idk im not a fan of removals.

JamieD1:
Nope

PR #15113, opened 2022-08-02
hardlight bow is armory access

SomeguyManperson:

It's a weapon so it goes in the armory 

:cl:  
tweak: hardlight bow is armory lathe printed rather than security lathe printed
/:cl:


Takahiru:
makes sense

JamieD1 (approved):

PR #15128, opened 2022-08-03
Decreases Blobbernaut Minimum Punch Damage & Brute Modifier & Increases distant health drain

VaelophisNyx:
yeah yeah ided, whatever

Just makes blobbernauts not have a flat 20 base damage, and instead sets them to have actual damage variation (still chunky as fuck)
also makes blobs take just slightly more brute damage, because passive healing on blob tiles + 200 fucking HP means these guys are literally brick walls that shrug off nearly all forms of brute damage, even on brute-weak strains.
additionally makes them less capable of hyper-over-extension by increasing their health drain away from blobs & factories.

# Changelog

:cl:  
tweak: reduced blobbernaut minimum melee damage
tweak: reduced blobbernaut brute damage mod (they take just slightly more brute damage now)
tweak: makes blobbernauts take twice as much damage if away from their blob or they have no factory.
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blob_mobs.dm
+++ b/code/modules/antagonists/blob/blob_mobs.dm
@@ -278,7 +278,7 @@
 					H.color = "#000000"
 		if(damagesources)
 			for(var/i in 1 to damagesources)
-				adjustHealth(maxHealth*0.05) //take 2.5% of max health as damage when not near the blob or if the naut has no factory, 5% if both
+				adjustHealth(maxHealth*0.05) //take 5% of max health as damage when not near the blob or if the naut has no factory, 10% if both
 			var/image/I = new('icons/mob/blob.dmi', src, "nautdamage", MOB_LAYER+0.01)
 			I.appearance_flags = RESET_COLOR
 			if(overmind)
--- a/code/modules/antagonists/blob/blob_mobs.dm
+++ b/code/modules/antagonists/blob/blob_mobs.dm
@@ -236,8 +236,8 @@
 	icon_dead = "blobbernaut_dead"
 	health = 200
 	maxHealth = 200
-	damage_coeff = list(BRUTE = 0.5, BURN = 1, TOX = 1, CLONE = 1, STAMINA = 0, OXY = 1)
-	melee_damage_lower = 20
+	damage_coeff = list(BRUTE = 0.65, BURN = 1, TOX = 1, CLONE = 1, STAMINA = 0, OXY = 1)
+	melee_damage_lower = 15
 	melee_damage_upper = 20
 	obj_damage = 60
 	attacktext = "slams"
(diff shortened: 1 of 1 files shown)


Mqiib:
I dunno the damage from being far away is bad enough, don't think it really needs the increase. Also if you're going to do that, at least adjust the comments to reflect the change.

VaelophisNyx:
I watched two blobs extend at minimum 10 tiles from the nearest factory and just not care

Marmio64:
Making a blobbernaut is very expensive and it's a gamble to see if they'll just run off and die, this seems unneeded unless you're going to decrease their price too

VaelophisNyx:
Not really? they have an absolute mountain of HP, hit like a freight train, and can overextend massively past blob tiles with relative ease. Even with this they still hit absurdly hard and have a ton of HP, they just can't shrug off as many hits of brute damage and can't walk massive distances from blob tiles to chase down crew. Additionally a decent blob player will likely have four or more out in short order

Marmio64:
> Not really? they have an absolute mountain of HP, hit like a freight train, and can overextend massively past blob tiles with relative ease. Even with this they still hit absurdly hard and have a ton of HP, they just can't shrug off as many hits of brute damage and can't walk massive distances from blob tiles to chase down crew. Additionally a decent blob player will likely have four or more out in short order

blobbernauts can't be spammed and replaced with "four or more in short order", a factory costs 60 points and cant be reused for new blobbernauts, and each one itself is 40 points. That is 100 points for one, not to mention the fact factories need to be placed away from each other so you're also spending points on nodes. In addition blobbernaut's ability to extend is vitally important and it is already very risky to do so even without the health nerf and damage increase. Right now I believe blobbernauts are in a decent spot balance wise and do not need this nerf unless you're also going to decrease the price.

VaelophisNyx:
> > Not really? they have an absolute mountain of HP, hit like a freight train, and can overextend massively past blob tiles with relative ease. Even with this they still hit absurdly hard and have a ton of HP, they just can't shrug off as many hits of brute damage and can't walk massive distances from blob tiles to chase down crew. Additionally a decent blob player will likely have four or more out in short order
> 
> blobbernauts can't be spammed and replaced with "four or more in short order", a factory costs 60 points and cant be reused for new blobbernauts, and each one itself is 40 points. That is 100 points for one, not to mention the fact factories need to be placed away from each other so you're also spending points on nodes. In addition blobbernaut's ability to extend is vitally important and it is already very risky to do so even without the health nerf and damage increase. Right now I believe blobbernauts are in a decent spot balance wise and do not need this nerf unless you're also going to decrease the price.

my man they fucking three-shot punch you in hardsuits, and can flat ignore most forms of damage because of the combo of reduced brute + passive heal
it's rare anyone gets a welder out to attack a blobbernaut, knowing it has 200 HP and will just heal. They shouldn't BE extending past the blob very far at all anyway. You shouldn't be capable of chasing crew down a hallway to kill them (a thing I have witnessed on multiple occasions). I'm speaking from experience and from witnessed accounts with everything I say.

Decreasing their price would be very, extremely bad.

PR #15129, opened 2022-08-03
Adds crafting recipe to high traction floor tiles

VaelophisNyx:

Adds a Crafting Recipe to high traction floor tiles, costing 1 plastic sheet for four

# Changelog

:cl:  
rscadd: Added a crafting recipe to high traction floor tiles
tweak: cargo's crate of high-traction tiles now comes with 360 tiles, making it still more efficient than buying plastic and hand-crafting it
/:cl:


00ze-cyclone:
There’s a crate that cargo can buy with 30 high traction floor tiles, it cost 2k.
With a plastic crate costing 1k you get 50 plastic sheets letting you make 200 high traction floor tiles.

VaelophisNyx:
I mean I can remove the crate or increase the tile count
I wasn't even aware of it existing tbh

ReoDaProtovali:
maybe make it produce 1 or 2 tiles per plastic or make the noslip crate cheaper

tattax:
don't think you should be able to make non slip tiles, cargo interaction is the least you need to do.

VaelophisNyx:
why not, exactly

JamieD1:
Way too easy to make for the power of them

PR #15133, opened 2022-08-03
Radiant dance machine mark IV can now be ordered from cargo

warface1234455:
# About PR
Radiant dance machine mark IV can be ordered from cargo for 6000cr
Everyone can access to this instead of having it only for Bartender
Ethereal disco ball can now be printed from autolathe and can be undeployed by unwrenching

:cl:  
tweak: radiant dance machine mark IV can be ordered from cargo for 6000cr
tweak: Everyone can now access radiant dance machine mark IV
tweak: Ethereal disco ball can be printed from autolathe and can be undeployed by unwrenching
/:cl:

### DIFF
--- a/code/game/machinery/dance_machine.dm
+++ b/code/game/machinery/dance_machine.dm
@@ -18,7 +18,6 @@
 	name = "radiant dance machine mark IV"
 	desc = "The first three prototypes were discontinued after mass casualty incidents."
 	icon_state = "disco"
-	req_access = list(ACCESS_BAR)
 	anchored = FALSE
 	var/list/spotlights = list()
 	var/list/sparkles = list()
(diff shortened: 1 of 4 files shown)


redmoogle:
i would check if this can cause overlapping sounds because that could potentially be **very** annoying

NCRLouTenant:
> i would check if this can cause overlapping sounds because that could potentially be **very** annoying

I can just imagine the halls lined with these as we're forced to dance everywhere, may the gods have mercy on us all.

Can't wait to see these used in combat.

warface1234455:
> i would check if this can cause overlapping sounds because that could potentially be **very** annoying

      i dont think so since it has same code like the normal jukebox which will just kinda overwrite the other one

JamieD1 (approved):
This can only end well

PR #15135, opened 2022-08-03
shifts the focus of ebow away from stun combat (except to borgs)

ynot01:

Removes its stamina damage 60->0
Doubles its knockdown 10->20 (theoretically 1 second->2 seconds)
Adds cosmetic jitter 0->20 (theoretically 2 seconds)
Apply light EMP to cyborgs on hit

The energy crossbow is known for its two hit stun and subsequent do-whatever-you-want because you've won. This essentially sets it as a ranged stun baton, which is not great for gameplay.

This sets the ebow back a notch into a bit more debilitating but not fight-ending tool. It will now reign most effective when followed up with another weapon.

To make up for this setback, the energy crossbow now deals light EMP to cyborgs on hit, so they are no longer completely immune to its effects.

# Changelog

:cl:  
tweak: Energy crossbow now deals no stamina damage, but knocks down for twice the length it used to and EMPs cyborgs
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/energy/ebow.dm
+++ b/code/modules/projectiles/projectile/energy/ebow.dm
@@ -9,6 +9,11 @@
 	jitter = 20
 	slur = 5
 
+/obj/item/projectile/energy/bolt/on_hit(atom/target, blocked = FALSE)
+	if(iscyborg(target))
+		target.emp_act(EMP_LIGHT)
+	. = ..()
+
 /obj/item/projectile/energy/bolt/halloween
 	name = "candy corn"
 	icon_state = "candy_corn"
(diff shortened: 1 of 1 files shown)


TheRyeGuyWhoWillNowDie:
ah yes, nerf antags, add ERT armor to sec. truly the yogdev experience

Marmio64:
make it 40 stamina damage instead?

This current change makes it more targeted towards using with an energy sword while 40 stamina damage would still give traitors a way to nonlethally take down opponents bar contractor baton/stealing/crafting the super shitty stunprod. Each one is still an acceptable niche but i'm more for a nerf to 40 stamina damage.

DeadlyBag:
I think this is a step towards the right direction but imo seems a little too weak.

One way to make it stronger would be to buff the toxin damage by 5+. Seems pretty justifiable if it goes along with the removal of stam damage or maybe lowering the TC cost of the item.

Bibby0110:
Knockdown isn't much more than annoying tbh. 
A nerf to stamina damage and such would be nice, but at 10 TC it's fair for the item to do more than 15 damage and knockdown. EMP doesn't really make up for it

PR #15159, opened 2022-08-05
energy bolas will now ensnare you if you get emped whilst carrying one

TheRyeGuyWhoWillNowDie:
bola combat is cringe since you lose in one click unless you have restraints removing abilities like freedom implants
also energy bolas are more or less a direct upgrade to normal bolas currently that you can buy for free roundstart and print an infinite number of with low-level tech. also requested by @VaelophisNyx

# Wiki Documentation

guide to combat, indicate that energy bolas ensnare you if you get emped while carrying one

# Changelog

:cl:  
tweak: you will now get ensnared if you get empd while carrying an energy bola
/:cl:


VaelophisNyx:
note: I just dared him to post it

00ze-cyclone:
How about EMP just delete the bolas like for other sec gear (last time I checked, EMP just empty energy weapons, not make them shoot their user), including deleting one that is ensnaring someone so EMP let you not get fucked if you weren’t fast enough to use it to win before sec used it.
Currently that change is just attempting to solve an instant win by adding another one and praying you’re faster at using your instant win than your opponent.

maxion12345:
Make it delete Instead, since they are made of energy (they already self - delete after removal anyways). That way secoffs aren't punished by taking gear they usually carry

JamieD1:
Yes make them short out instead

JamieD1:
Reopen it when you actually want to work on it.

PR #15165, opened 2022-08-05
[ADMIN] secrets panel is now tgui

ynot01:
Buttons are color-coded to how they impact the round
Round-impacting buttons without confirmations now have TGUI confirmation as shown in gif below

if you do not have sufficient permissions, the buttons will be greyed out


code/modules/admin/secrets.dm
```diff
@@ -1,115 +1,66 @@
...
+		var/client/user_client = user
+		holder = user_client //if its a client, assign it to holder
+	else
+		var/mob/user_mob = user
+		holder = user_mob.client //if its a mob, assign the mob's client to holder
+	ui_interact(holder.mob)
+
+/datum/tgui_secrets_panel/ui_data(mob/user)
+	var/list/data = list()
+	var/client/rights = holder
```
adamsong, line 35: Do not give bibby full access to the panel

adamsong (changes requested):

code/modules/admin/secrets.dm
ynot01: fixed

JamieD1:
Can we make it not syndicate? Like I get it haha, but can we just use a centcom background instead?

ynot01:
> Can we make it not syndicate? Like I get it haha, but can we just use a centcom background instead?

done

JamieD1:
> > Can we make it not syndicate? Like I get it haha, but can we just use a centcom background instead?
> 
> done

Got a picture?

JamieD1:
> > Can we make it not syndicate? Like I get it haha, but can we just use a centcom background instead?
> 
> done

Got a picture?

ynot01:
> > > Can we make it not syndicate? Like I get it haha, but can we just use a centcom background instead?
> > 
> > 
> > done
> 
> Got a picture?

the picture has been in the description
image

JamieD1 (approved):

PR #15169, opened 2022-08-05
Fixes RnD console access restrictions not working

TheRyeGuyWhoWillNowDie:
currently this checks if the console is locked (aka if the Lock button is pressed which already prevents researching) but because the console can never be locked while in a state where you can research stuff, the access check proc always returns true

# Changelog

:cl:  
bugfix: rnd console now actually requires correct access to use, emagging will allow you to research without access
/:cl:


ToasterBiome (approved):

Bibby0110:
Correct behavior reintroduced in https://github.com/yogstation13/Yogstation/pull/15174, thanks for bringing it to our attention

PR #15174, opened 2022-08-05
fixes RnD console access

Bibby0110:
you need to have access AND the console needs to not be locked.

We can't just check for access since href exploits. We don't need to check for emag since locked can never be true if the console is emagged


TheRyeGuyWhoWillNowDie:
duplicate of #15169

ToasterBiome (approved):

PR #15181, opened 2022-08-06
Fixes NaN exploit [s]

JamieD1:
Fixes NaN exploit [s]


ToasterBiome:
You no longer return afterwards?

ToasterBiome (approved):

PR #15202, opened 2022-08-07
makes arcane barrage actually deal more damage than enchanted rifles

ynot01:

## BALANCE CHANGE

`desc = "Fire a torrent of arcane energy at your foes with this (powerful) spell. Deals much more damage than Lesser Summon Guns, but won't knock targets down. Requires both hands free to use. Learning this spell makes you unable to learn Lesser Summon Gun."`
??????????????

damage 20->40
armor pen 0->20

costs 3 now

# Changelog

:cl:  
bugfix: made arcane barrage actually do more damage than lesser summon guns but cost 3
/:cl:

### DIFF
--- a/code/modules/antagonists/cult/cult_items.dm
+++ b/code/modules/antagonists/cult/cult_items.dm
@@ -788,6 +788,8 @@ GLOBAL_VAR_INIT(curselimit, 0)
 	nondirectional_sprite = TRUE
 	damage_type = BRUTE
 	impact_effect_type = /obj/effect/temp_visual/dir_setting/bloodsplatter
+	damage = 20
+	armour_penetration = 0
 
 /obj/item/projectile/magic/arcane_barrage/blood/Bump(atom/target)
 	var/turf/T = get_turf(target)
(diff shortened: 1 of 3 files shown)


Marmio64:
40 damage and 20 armor pen sounds like a lot, you'll be three shotting people even if they have armor vests, maybe reduce the 20 AP to 10 APC so its a 4 shot againt seccies? Or maybe put it back into the 3 spell point range so its a direct contender with lesser summon guns.

ynot01:
> 40 damage and 20 armor pen sounds like a lot, you'll be three shotting people even if they have armor vests, maybe reduce the 20 AP to 10 APC so its a 4 shot againt seccies? Or maybe put it back into the 3 spell point range so its a direct contender with lesser summon guns.

summon guns two shot crits without fail, your price for spending one less point is having to land one more hit and you still have the same 30 shots so you can only overall down 0.66x as many people

im not opposed to making it 3 points but id like not to touch the AP

00ze-cyclone:
The blood barrage is a subtype of the arcane barrage and doesn’t overwritte the damage values, you might want to make the blood barrage 20 brute because last thing we need is a bloodcult buff.

ynot01:
> The blood barrage is a subtype of the arcane barrage and doesn’t overwritte the damage values, you might want to make the blood barrage 20 brute because last thing we need is a bloodcult buff.

done

JamieD1:
> Reference in ne

Make it 3 points if you are doubling the damage

JamieD1:
^

Bibby0110 (approved):
jamie said ok i guess

PR #15205, opened 2022-08-07
Fixes automatic greentexting of "Protect the nuke" objective

Altoids1:
image
*Image above displays the bug, not the fix*

I happened to notice this while doing my other bugfix today. I'm assuming I can just toss the `endround` marker on this event and then it will actually check itself properly.

## Changelog
:cl:  Altoids
bugfix: Fixed automatically greentexting the Protect The Nuke objective.
/:cl:


VaelophisNyx:
this also occurs with Med survival %'s if population is low enough

redmoogle:
you may want to look into `var/continuous` it allows you to say

Every 30 minutes give security $600 if the nukecore is still in

Could allow for some metagaming but eeeehhh I feel like security could figure out if its gone before hand

Bibby0110 (approved):

PR #14891, opened 2022-07-17
Tweaks Blood Cult Gear & Structures & Holy Water Deconversion

VaelophisNyx:
# Document the changes in your pull request

Pre-stating that these are the only things I could think to do with this nightmare code to make the blood cult steamroll less frequently
additionally arguments that blood cult winrates are abysmal, I will direct to the information that it's a case of giving a monkey a sword. They still have to know how to use it right. If they DO they're terrifying. Population tends to drop to dismal levels after blood cult rounds, because very few people have fun, even most of the cultists don't much enjoy it, usually due to it being a cakewalk for the cult if they have even a single person who knows what they're doing.

Blood Cult has exactly one loadout that matters. Mirror Shield + Eldritch Longsword. This is simply because both are effectively **free** for blood cult members to get in bulk, and both are exceptionally strong for being free gear that cant be used if you aren't also a cult member.

This PR reduces the damage of the longsword by 10, but makes it a subtype of two-handed swords (virbro-sword specifically). When wielded it actually does 12 more damage now, totaling to 27, 2 more than originally. This however means you cannot use it at full strength with the shields

Additionally, mirror shields now have slightly reduced block chance. Being equal to a riot shield, only usable by cult members, and free was a bit of a strong combo. Atop this, they apply a Knockdown now, not a Paralyze. They should never have been tazers when thrown, and a knockdown is already sufficient as it drops someone's held items in 90% of cases. Paralyzing them was excessive.

Armor penetration for the cult dagger, longsword, spear, and the currently unobtainable bastardsword have all been lowered. This is an attempt to reel in cult damage values more, as currently they don't care if you're in a HoS hardsuit with nanites, you will be taking heavy damage from their free-to-get longsword.

As for structures, the Forge, Altar, and Archives have all had 50% added to their cooldown timers. As it stands the blood cult can simply spam more of them (something I wanted to cap) and acquire free gear the second any of them are done, and the cooldown of 4 minutes means they can snowball faster than anyone can react. This change should stop that being such a hard snowball effect.

Additionally the Forge, Altar, and Archives have all had their costs increased, now taking 7 runed metal each instead of three. Cultists can convert 10 plasteel to 10 runed metal fairly frequently, and combined with there being no charge-up time for any of these structures, it lets you mass produce gear in seconds. This change makes spamming more structures a little harder, though not impossible.

[...]

PR #15210, opened 2022-08-07
Cult artificer magic missiles now no longer affect cultists

ShadowDeath6:
Steals one change from #14891 that I thought was already implemented within the game.

_Noncult_ artificers also can't target cultists with their magic missiles. I could code a way around this but I think it's more lore-friendly that they're all powered by Nar-Sie and thus their magic still can't harm her followers.

I also could have implemented it that it wouldn't TARGET cultists at all, either, but I prefer it as is since I don't particularly think massive groups of cultists and artificers running around KOing everyone would be really fun. This means that magic missile is still much weaker in large groups since it'll be "wasted" on some cultists.

:cl: ShadowDeath6
tweak: Artificers' Magic Missile spell no longer affects cultists.
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/construct_spells.dm
+++ b/code/modules/spells/spell_types/construct_spells.dm
@@ -163,7 +163,7 @@
 	if(ismob(target))
 		var/mob/M = target
 		if(iscultist(target))//cultists can't be harmed by their own constructs' spells!
-		        to_chat(target, span_danger("[src] harmlessly dissipates into crimson particles upon contacting your body!"))
+			to_chat(target, span_danger("[src] harmlessly dissipates into crimson particles upon contacting your body!"))
 			return BULLET_ACT_BLOCK
 		if(M.anti_magic_check())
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
I’ll need to fix this when I get home in like an hour or so, my bad

VaelophisNyx:
recommend stealing the magic missile name change too, otherwise you'll get fun cases of "the the a magic missile"

ShadowDeath6:
Do normal magic missiles actually display a message ever? I didn't think it would be necessary.

VaelophisNyx:
they do when blocked or on rare occasion when they hit

ShadowDeath6:
I couldn't reproduce that- every usage of [src] replaced it with "the lesser magic missile" or "a lesser magic missile".

redmoogle:
> I couldn't reproduce that- every usage of [src] replaced it with "the lesser magic missile" or "a lesser magic missile".

the only gets added when src starts with a capital iirc

JamieD1 (approved):

PR #15211, opened 2022-08-07
Improves the description of the Bluespace Harvester, makes it easier to see when its done

### LINKED ISSUE #15207: Bluespace Harvester Goal Never Completes
## Test Merges: 
Not Supplied / None
## Reproduction:
Description: While observing this round I saw a completed Bluespace Harvester, but end of round report showed the goal as not complete

 Submitted by: Vaelophis Nyx (She/Her) (319685669586862087)

Altoids1:
image

I was looking into #15207 and noticed that the Bluespace Harvester didn't have its own description (and was instead using the default one that all `/obj/machinery` have). So, I gave it one. I also marked the name as being `\improper` so that the examine text looks better (not shown in the above image because I just amended it on while writing this)

Beyond that, I made an attempt at making it more clear when greentext hath been achieved by adding a new message in the examine informing you whether or not the machine has reached its point requirement. I'm going to say that doing this closes #15207 (since the bug does not exist, and this resolves the UX problem that cause that person to file a bug report to begin with).

## Changelog
:cl:  Altoids
spellcheck: The Bluespace Harvester now has an actual, grammatically-correct description.
spellcheck: Examining the Bluespace Harvester now informs you whether or not you've reached its point goal.
/:cl:

### DIFF
--- a/code/modules/goals/station_goals/bluespace_tap.dm
+++ b/code/modules/goals/station_goals/bluespace_tap.dm
@@ -1,7 +1,8 @@
 //Station goal stuff goes here
+#define BLUESPACE_TAP_POINT_GOAL 45000 // Yogs
 /datum/station_goal/bluespace_tap
 	name = "Bluespace Harvester"
-	var/goal = 45000
+	var/goal = BLUESPACE_TAP_POINT_GOAL // Yogs
 
 /datum/station_goal/bluespace_tap/get_report()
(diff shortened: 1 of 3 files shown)


warface1234455:
Also add a text to notify how much points you need for it to complete the goal? cuz it only available in comms console

Bibby0110 (approved):

PR #15217, opened 2022-08-08
Increases AI download speed

Bibby0110:
should be around 3,7 minutes without firewall upgrade


redmoogle:
Isnt it 3:42 minutes with a firewall (since 0.5x) and 1:51 for non firewalled

Bibby0110:
> Isnt it 3:42 minutes with a firewall (since 0.5x) and 1:51 for non firewalled

ticks are every two seconds

Technically the speed without upgrades is 111 ticks, which is variable if you really drill down

PR #15226, opened 2022-08-08
telecrystal jumping

ynot01:

Telecrystals can now be crushed in hand, similar to their bluespace cousins, but as they are much more rare and precious, they are much more precise and can be used as a sort of ghetto launchpad or get-the-fuck-out utility.

When crushing a telecrystal, you will be teleported 3-6 tiles forward and suffer minor nausea, about half of the effects that crushing a bluespace crystal has (Teleportation is still nauseating, but you have a good idea of where you're ending up at least)

You are also moved forward one tile, to give a cool 'exit portal' effect and also to make you go really fast if you end up in zero gravity. This WILL throw you into walls/other people, so be a little cautious!

# Wiki Documentation

Using a telecrystal in hand will now teleport you forward 3-6 tiles

# Changelog

:cl:  
rscadd: Added telecrystal jumping, which teleports you forward 3-6 tiles when you crush one!
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/telecrystal.dm
+++ b/code/game/objects/items/stacks/telecrystal.dm
@@ -23,7 +23,7 @@
 	new /obj/effect/particle_effect/sparks(loc)
 	playsound(loc, "sparks", 50, 1)
 
-	if(do_teleport(L, destination, asoundin = 'sound/effects/phasein.ogg', channel = TELEPORT_CHANNEL_BLUESPACE))
+	if(!do_teleport(L, destination, asoundin = 'sound/effects/phasein.ogg', channel = TELEPORT_CHANNEL_BLUESPACE))
 		L.visible_message(span_warning("[src] refuses to be crushed by [L]! There must be something interfering!"), span_danger("[src] suddenly hardens in your hand! There must be something interfering!"))
 		return
 
--- a/code/game/objects/items/stacks/telecrystal.dm
+++ b/code/game/objects/items/stacks/telecrystal.dm
@@ -8,6 +8,34 @@
 	max_amount = 50
 	item_flags = NOBLUDGEON
 
+/obj/item/stack/telecrystal/attack_self(mob/user)
+	if(!isliving(user))
+		return
+	
+	var/mob/living/L = user
+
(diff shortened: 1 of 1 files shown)


code/game/objects/items/stacks/telecrystal.dm
```diff
@@ -8,6 +8,34 @@
...
+
+	if(!istype(destination))
+		return
+
+	L.visible_message(span_warning("[L] crushes [src]!"), span_danger("You crush [src]!"))
+	new /obj/effect/particle_effect/sparks(loc)
+	playsound(loc, "sparks", 50, 1)
+
+	if(do_teleport(L, destination, asoundin = 'sound/effects/phasein.ogg', channel = TELEPORT_CHANNEL_BLUESPACE))
+		L.visible_message(span_warning("[src] refuses to be crushed by [L]! There must be something interfering!"), span_danger("[src] suddenly hardens in your hand! There must be something interfering!"))
```
tattax:
```suggestion
	if(!do_teleport(L, destination, asoundin = 'sound/effects/phasein.ogg', channel = TELEPORT_CHANNEL_BLUESPACE))
		L.visible_message(span_warning("[src] refuses to be crushed by [L]! There must be something interfering!"), span_danger("[src] suddenly hardens in your hand! There must be something interfering!"))
```

alexkar598:
1tc is rather cheap for a get out of jail card

ynot01:
> 1tc is rather cheap for a get out of jail card

Bluespace crystals don't cost TC and do the same, just not directional. I have seen several antags use bluespace crystals and this provides a (more expensive) alternative.

ShadowDeath6:
I'd honestly rather have a 1tc traitor item called "tele-crystal" just to avoid new traitor players accidentally using these in their hands and losing them.

KillerOrcaCora:
I agree with the above, I can see yoggdites messing this one up really bad

ynot01:
I disagree, while that is an issue that could happen, I doubt it will happen twice to the same person. Keeping it as the currency enables similar versatility featured in the ancient toolbox, which I love the design of.

JamieD1 (approved):

PR #15246, opened 2022-08-10
Adds balloon alert preference and balloon_or_message proc

ghost:

Adds a preference to disable balloon alerts, and a proc that will send a balloon alert or a chat message depending on user's preference toggle. 

# Spriting
No

# Wiki Documentation

Not needed

# Changelog

:cl: SuperSlayer
rscadd: Added disable/enable balloon alerts preference
/:cl:

### DIFF
--- a/code/modules/balloon_alert/balloon_alert.dm
+++ b/code/modules/balloon_alert/balloon_alert.dm
@@ -10,6 +10,17 @@
 
 	INVOKE_ASYNC(src, .proc/balloon_alert_perform, viewer, text)
 
+/// Creates a balloon alert, or sends a chat message dependant on client preferences. 
+/// Args: viewer -  mob that gets message/alert, alert - balloon alert text, message - text message that the mob gets if the preference is toggled, equals to alert message if not passed in the proc
+/atom/proc/balloon_or_message(mob/viewer, alert, message)
+	if(!message)
+		message = alert
+
+	if(viewer.client.prefs.disable_balloon_alerts)
+		to_chat(viewer, message)
+	else
+		balloon_alert(viewer, alert)
+
 /// Create balloon alerts (text that floats up) to everything within range.
 /// Will only display to people who can see.
 /atom/proc/balloon_alert_to_viewers(message, self_message, vision_distance = DEFAULT_MESSAGE_RANGE, list/ignored_mobs)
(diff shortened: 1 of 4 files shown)


Bibby0110:
I don't believe this should be a toggle. Keep it on for everyone and tweak it.

In the future most actions should be converted to balloon alerts which could get us close to removing the chat window itself

ghost:
> I don't believe this should be a toggle. Keep it on for everyone and tweak it.

Why not?

> In the future most actions should be converted to balloon alerts which could get us close to removing the chat window itself

I agree with this, and I will make a PR that will convert some things to them

Bibby0110:
> > I don't believe this should be a toggle. Keep it on for everyone and tweak it.
> 
> Why not?

I'm guessing this is a classic case of "change is bad." Let it stay for a while and grow on people. Address any complaints (too large, too small, stays too long etc) They're better game design and we should strive to replace most smaller pieces of textual information with these

ghost:
Ok

Cartlord:
I don't think anything TheGamerdk said holds water, frankly, and I still think this should be a toggle due to balloon alerts not being readable for than two seconds later (which chat messages can) and not showing up instantly (which chat messages do).

SomeguyManperson:
> I don't think anything TheGamerdk said holds water, frankly, and I still think this should be a toggle due to balloon alerts not being readable for than two seconds later (which chat messages can) and not showing up instantly (which chat messages do).

basically this, balloon alerts show up slow and tend to overlap, and are extremely temporary. One of the main things they cause problems with is beaker transfer unit switching where the lack of immediate feedback, lack of a permanent record and the fade-in all combine to make it much more difficult to tell what the current transfer amount is.

However pref toggles for mechanical stuff like this are dumb and the issue should be addressed in some way rather than allowing people to ignore it.

Cartlord:
honestly the only solutions i can think of are "make balloon alerts come _alongside_ chat messages" and "remove balloon alerts"

adamsong:
I agree with the idea that the preference is bad. Don't take bad design, and say "oh, well, its a preference so it doesn't matter, anyone who doesn't like it will turn it off." Tweak it to improve it or remove it entirely.

neeshacark:
this should be a preference for one reason; its a fucking preference

i dont want text that i dont need flashing in my face every time i do anything when i can look slightly to the right and see it there.  this is the same reason i dont use runechat, i don't like it, i don't PREFER it over not having it.  you might PREFER having runechat and this on, but for the love of god do not force it on everyone because not everybody wants it

JamieD1 (approved):

PR #15264, opened 2022-08-11
Brig Physician room changes + Interrogation change

Xantam-13:
Changes the brigs interrogation to be 1 tile smaller and gives that space to the brig physician office. 

Currently, the room that is never used (interro) is really large with not much in it, while the brig physicians room gets used quite a lot and is cramped small. 

image

:cl:  
tweak: Resized interrogation, same layout and purpose but 1 tile/column smaller
tweak: Gave the space from interrogation, to the brig physician which allows more spacing to do surgeries and whatnot, as they so often run out during highpop
/:cl:


ToasterBiome (approved):
Not liking the table in the 1x1 space but I think the change is a nice one

PR #15270, opened 2022-08-12
Holy melons now have the same size as a watermelon

redmoogle:

Holy melon is the same size as a normal melon (sprite wise) and is a mutation of melon and is also majorly mass producible so it should have the same weight as a normal melon. Looking at the code it looks like they were tiny weight which is wtf.

# Wiki Documentation

Holy melon strat is a bit weaker

# Changelog

:cl:  
tweak: Holy melons now have normal instead of tinyweight
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/melon.dm
+++ b/code/modules/hydroponics/grown/melon.dm
@@ -56,6 +56,7 @@
 	icon_state = "holymelon"
 	filling_color = "#FFD700"
 	dried_type = null
+	w_class = WEIGHT_CLASS_NORMAL
 	wine_power = 70 //Water to wine, baby.
 	wine_flavor = "divinity"
 


ynot01:
the biggest effect this has is that you can no longer pocket holy melons, which was the default strat to get antimagic charges, now you will have to hold them in your hands

if you don't intend this effect then i would suggest making them small items

redmoogle:
> the biggest effect this has is that you can no longer pocket holy melons, which was the default strat to get antimagic charges, now you will have to hold them in your hands
> 
> if you don't intend this effect then i would suggest making them small items

this was discussed in public thought im also open to respriting the holy melon to not be a hueshifted cursed sprite

JamieD1 (approved):

PR #15289, opened 2022-08-13
Mantis blade changes

isotope436:
Changes the mantis blades to work in a sort of stealthy alt to the esword well still being useful enough to warrant their existence. Instead of 2 mantis blades you only get one with the difference being instead you only have to spend 7 tc allowing you to get other items compared to their current form where you get two for 16tc and end up being unable to buy anything else. Now the question comes to why not esword, With mantis blades you lose your ap but gain a 20% block chance and the fact its completely undetectable unless you either get caught with it out or implant checked (making it an insane backup if you get permed)

# Wiki Documentation

The gorlax mantis kit is now just gorlax mantis blade

# Changelog
:cl:  
tweak: changed cost of GORLEX mantis blade in the uplink to 7Tc
tweak: GORLEX mantis blade kit replaced with a single mantis blade
tweak: GORLEX mantis blade uplink description altered to reflect giving you only a single blade.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1807,13 +1807,13 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	surplus = 0
 	include_modes = list(/datum/game_mode/nuclear)
 
-/datum/uplink_item/implants/mantis_kit
-	name = "G.O.R.L.E.X.. Mantis Blades Kit"
-	desc = "Comes with 2 G.O.R.L.E.X. Mantis blades. All packaged with autosurgeons."
-	item = /obj/item/storage/briefcase/syndie_mantis
-	cost = 16
-	surplus = 0
-	exclude_modes = list(/datum/game_mode/infiltration) // yogs: infiltration
+/datum/uplink_item/implants/mantis
+    name = "G.O.R.L.E.X. Mantis Blade"
+    desc = "One G.O.R.L.E.X Mantis blade implant able to be retracted inside your body at will for easy storage and concealing, 2 blades can be used at once."
+    item = /obj/item/autosurgeon/organ/syndicate/syndie_mantis
+    cost = 7
+    surplus = 0
+    exclude_modes = list(/datum/game_mode/infiltration) // yogs: infiltration
 
 // Events
 /datum/uplink_item/services
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1808,12 +1808,12 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	include_modes = list(/datum/game_mode/nuclear)
 
 /datum/uplink_item/implants/mantis
-    name = "G.O.R.L.E.X. Mantis Blade"
-    desc = "One G.O.R.L.E.X Mantis blade implant able to be retracted inside your body at will for easy storage and concealing, 2 blades can be used at once."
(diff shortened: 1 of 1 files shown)


ynot01:
better than they were

VaelophisNyx:
@ynot01 what

isotope436:
the only issue i ever had with them from the start was purchasing them fucked you over in the long run cause it ate all of your TC and at first i thought about making them 10tc for both but halfing them and only getting one made more sense. plus at where they are currently the TC cost makes them easily avoidable and so they just end up rotting away on a shelf :P

ynot01:
> @ynot01 what

i dont know if you live under a rock but the mantis blades are pretty shit and hard to use, no one ever uses them and those who do die spectacularly making them never buy it again

this makes it plainly easier to use and sees a huge price reduction which is always a plus in my book, enabling much greater flexibility in your purchases

VaelophisNyx:
no no I was just
what you said made no sense in general; I agree with you now that I have context

GraveHat:
Should probably specify that the use of two blades at once allows them to strike at the same time so people don't forget its original gimmick.

isotope436:
@ynot01 are you able to fix the tags for me? or can i? i have literally never made a pr before :P

code/modules/uplink/uplink_items.dm
```diff
@@ -1807,13 +1807,13 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
...
-	cost = 16
-	surplus = 0
-	exclude_modes = list(/datum/game_mode/infiltration) // yogs: infiltration
+/datum/uplink_item/implants/mantis
+    name = "G.O.R.L.E.X. Mantis Blade"
+    desc = "One G.O.R.L.E.X Mantis blade implant able to be retracted inside your body at will for easy storage and concealing, 2 blades can be used at once."
+    item = /obj/item/autosurgeon/organ/syndicate/syndie_mantis
+    cost = 7
+    surplus = 0
+    exclude_modes = list(/datum/game_mode/infiltration) // yogs: infiltration
```
SomeguyManperson: edit these so they are tabs instead of spaces

isotope436: @SomeguyManperson  would if i knew how i didnt make this code im just PRing it. i also dont know what you want me to tab here?

isotope436: figured out what you want me to change and did just gotta figure out how to put it here

isotope436:
i think i figured it out

RG4ORDR:
Mantis blades are in a really weird area.
People have suggested they get attack speed buffs, a leap ability, re-add their reusable implanter, block chances, AP, so forth. Adding on a 20% block chance isn't going to get people to buy them when something like the esword, which has higher block dance, AP, and much higher damage can fulfill the mantis blades' role far better than it can.

Personally I think the Blades should be reduced in cost but given some gimmick like Xeno's leap onto targets X tiles away. Or give it a very high wound chance so you can disembowel people in combat

isotope436:
@RG4ORDR I didn't add block chance they already had it, and I'm not the one coding these I don't know how, this isn't even my code jelly made this for me for my PR. and if i give them any form of high block or disembowel they will get similar treatment of the DEsword. I also cant boost their damage to say 30 or else if you have two your doing 60 damage every swing because you can attack with both at once. This is entirely meant to move them out of the unusable category that they are currently in and allow them to at least be purchased as they are just bloat atm.

ynot01 (approved):
autosurgeon is not re-usable as is intended, code works

JamieD1 (approved):

PR #15296, opened 2022-08-13
MDB IGNORE Expands Interrogation by 1 tile into space

Xantam-13:
Expands Interrogation by 1 more tile into space to give it a little more space back

image

:cl:  
tweak: Expands Interrogation by 1 more tile into space to give it a little more space back  
/:cl:


ToasterBiome (approved):
Looks good if we need to change the window we can always remove it

PR #15298, opened 2022-08-13
MDB IGNORE Adds 2 glass windoors to Brig turnstiles for reasons...

Xantam-13:
Adds windoors to the brig turnstiles that will need to be opened by a security personnel to allow infinite entering/exiting. 

Currently it's unblocked, allowing those in cuffs during a brig tussle to just run free, this is incredibly difficult for security in the less common occasions. You also see more often nowadays, wardens using a barrier nade to block the turnstiles. 

In addition - the firedoors do not function properly there allowing faulty atmos to flow through, no idea why. The windoors could be closed to prevent this. 

image

:cl:  
tweak: Adds 2 glass windoors to Brig turnstiles for reasons... 
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -60565,6 +60565,16 @@
 	},
 /turf/open/floor/plasteel,
 /area/security/main)
+"vqk" = (
+/obj/machinery/door/firedoor/border_only{
+	dir = 1
+	},
+/obj/machinery/door/firedoor/border_only,
+/obj/machinery/turnstile/brig{
+	dir = 1
+	},
+/turf/open/floor/plasteel,
+/area/security/brig)
 "vra" = (
 /obj/effect/turf_decal/stripes/line{
 	dir = 8
@@ -98372,7 +98382,7 @@ upk
 mLd
 lLO
 ebQ
-lpG
+vqk
 qad
 mqQ
 lpG
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -37843,6 +37843,21 @@
 "lpv" = (
 /turf/open/floor/plasteel,
 /area/crew_quarters/dorms)
+"lpG" = (
+/obj/machinery/door/firedoor/border_only{
+	dir = 1
+	},
+/obj/machinery/door/firedoor/border_only,
+/obj/machinery/turnstile/brig{
+	dir = 1
+	},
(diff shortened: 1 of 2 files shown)


NCRLouTenant:
Against this in regards to not letting me... I mean cuffed individuals, escape Brig, but also am totally on board for it if it fixes atmos. 

Just, if a sec officer somehow loses a cuffed person IN BRIG they should suffer at least some trouble in the form of chasing that person down.

Xantam-13:
> Against this in regards to not letting me... I mean cuffed individuals, escape Brig, but also am totally on board for it if it fixes atmos.
> 
> Just, if a sec officer somehow loses a cuffed person IN BRIG they should suffer at least some trouble in the form of chasing that person down.

If the windoors are pinned open - which I assume they would be 90% of the time (I would as an officer) - I imagine the situation to be the same as it is right now

00ze-cyclone:
I think it’ll mostly make going in and out of the brig as sec slower, which can and will kill you during things like revs, gang or bloodcult. (Lost one second to get inside for safety, rev caught up to you and are mauling you, lost one second to get out when there’s a bomb from revolutionary toxin, you get obliterated)
We should make turnstiles act like tiny fans instead to solve the atmos problem.

Xantam-13:
> 

Tiny fans aren't a solution either

I aint a coder, this is the only fix I know.

NCRLouTenant:
> > Against this in regards to not letting me... I mean cuffed individuals, escape Brig, but also am totally on board for it if it fixes atmos.
> > Just, if a sec 
> If the windoors are pinned open - which I assume they would be 90% of the time (I would as an officer) - I imagine the situation to be the same as it is right now

Fair point, I'd likely do the same if I played sec. It would then fall on Sec for leaving the easy escape option open and go right back to my point of punishing them for losing a suspect. Fixing the atmos thing is more important anyways.

Georgeur:
Don’t listen to the retards. Good pr but make it 1 windoor at the outer most exit.

Xantam-13:
> Don’t listen to the retards. Good pr but make it 1 windoor at the outer most exit.

Fine, doing that now

maxion12345:
could just
have it block people in cuffs

as for the Atmos, don't know why it's not working. I'll futz with it tomorrow. Thinking it's probably due to firelocks thinking "oh there's a thing here blocking Atmos so I don't need to close" but the turnstiles let air past because it has big hole

ToasterBiome:
Gonna be honest, this feels too hugbox. If it was really a problem I think a "no cuffed guy in turnstiles" code solution is superior to this. Sorry.

PR #15309, opened 2022-08-14
akimbo rework

ynot01:

Instead of firing both guns, you will instead swap between the guns and have a 1/4 faster click CD so you can fire faster

In theory this should improve DPS by a maximum of 50%

Spread penalty was decreased to accomodate (subject to change)

##

An interesting thing I noticed in testing is that **spam clicking is actually detrimental** to your fire-rate when dual wielding, and you will get faster fire rate from more controlled fire (You will notice in the bottom gif sometimes I fuck up and stop for a second)

I think this will be very interesting to see as people are typically very jumpy and anxious in tense situations, likely making this controlled fire a lot harder to achieve

Firing 1 laser gun

Dual firing 2 laser guns

# Changelog

:cl:  
tweak: Akimbo mechanics have been reworked, rewarding precise clicking and punishing spam
/:cl:


N3D6:
Does this affect PKAs?

ynot01:
> Does this affect PKAs?

Yes

JamieD1 (approved):

PR #15323, opened 2022-08-15
slightly reduces the hugboxxing on the rnd console

TheRyeGuyWhoWillNowDie:
the RnD console's Lock button is no longer completely worthless since it was the old way to keep tiders out of the Science lab but apparently no one used it so a hard cockblock was added to the console. this changes it by making the RnD console's access requirements toggled by locking or unlocking the console. The console starts locked, so its identical to the previous version. the only real difference is you can get the AI or whatever to unlock the console so anyone can use it

RnD console cockblocks for "Powergaming" are super lame and there was already a solution for it (locking the console) so this solves that problem. I will be reducing the hugbox on the exofab in a future PR

tested, works

# Changelog

:cl:  
tweak: the research console's access restrictions are now toggled depending if the console is locked or not, as god intended
tweak: the RnD console now starts locked
/:cl:


AlvCyktor:
👎

Majkl-J:
Whoever keeps doing these powergaming cockblocks never checks for the effects on ghost roles, sometimes not even if he fixed the issue. And it pisses me off

JamieD1:
Noone will ever lock the console, I would prefer you fix the problems instead of just stealth reverting it because you dont like it.

Maia-J:
> Noone will ever lock the console, I would prefer you fix the problems instead of just stealth reverting it because you dont like it.



> The console starts locked

Do you even read these?

PR #15332, opened 2022-08-16
Gutlunches now obliterate their udders on dying

SomeguyManperson:
No gameplay impact but stuff should generally delete itself when it is being deleted yes?

:cl:  
bugfix: gutlunch udders now die when they are killed
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/gutlunch.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/gutlunch.dm
@@ -46,6 +46,11 @@
 	udder = new()
 	. = ..()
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/mining_mobs/gutlunch.dm
```diff
@@ -46,6 +46,11 @@
 	udder = new()
 	. = ..()
 
+/mob/living/simple_animal/hostile/asteroid/gutlunch/Destroy()
+	qdel(udder)
+	udder = null
```
tattax: why not use qdel_null

JamieD1 (approved):

PR #13280, opened 2022-02-14
Makes brass skewers actually bad to step on

Mqiib:
# Document the changes in your pull request

Being stabbed on a brass skewer no longer makes you unclickable. Apparently this was intentional???
You have to actually click on the skewer itself to pull someone off, but can now freely stab them to shit without first destroying the skewer if that's more your style. Also stuns the stabee for 2 seconds (not very long, same stun as borgs) because they just got IMPALED which would reasonably knock the wind out of you. People running in, getting impaled, and then just freely shooting to death the nearby clockie who was all "HAHA I GOT YOU IN MY GIANT SPEAR KEBAB TRAP" is kind of dumb. Skewers are criminally underused anyways for how cool they are, and now this'll make them actually spooky scary and give you an incentive to NOT just ignore them while running in screaming and firing akimbo laser guns.

# Changelog

:cl:  

bugfix: Brass skewers no longer make the shishkebabbed mob unclickable by normal means
tweak: Brass skewers stun for 2 seconds upon skewerification

/:cl:

PR #15347, opened 2022-08-17
TEG can no longer deplete the power grid

adamsong:
TEG generating negative many power is bad
Doesn't actually fix why the TEG was negative, just caps it at zero
# Changelog

:cl:  
bugfix: TEG can no longer produce negative many power
/:cl:

### DIFF
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -119,6 +119,7 @@
 	add_avail(power_output)
 	lastgenlev = power_output
 	lastgen -= power_output
+	lastgen = max(lastgen, 0)
 	..()
 
 /obj/machinery/power/generator/proc/get_menu(include_link = TRUE)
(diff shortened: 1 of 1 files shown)


redmoogle (approved):
I dont see any reasonable way this can turn negative but if it fixes it w/e (probably should just rewrite this whole damn file)

SynnGraffkin:
> I dont see any reasonable way this can turn negative but if it fixes it w/e (probably should just rewrite this whole damn file)

This would probably fit better as a "Feature request" but have you considered Citadel's TEG? Last I checked, which was early 2021, it was effectively "nerfed" but also changed in such a way that it promoted actually having major heat difference instead of just minor temp differences, because before then you could make a 3x3 TEG that pumped MW's of power with freezer / heater.  I haven't tried that here, but not sure if it has been something done here either. 

[ Further tweaks the TEG's power output #13280 ](https://github.com/Citadel-Station-13/Citadel-Station-13/pull/13280) 

[History for Citadel-Station-13/code/modules/power/generator.dm](https://github.com/Citadel-Station-13/Citadel-Station-13/commits/be9f550261c51f88d004af98c492a6dab8a15a92/code/modules/power/generator.dm)

JamieD1 (approved):

PR #15350, opened 2022-08-17
makes vehicles respect slowdown modifiers

ynot01:

you will properly slow down depending on your slowdown modifiers (hurt, clothing, etc)

vehicles reduce slowdown by 50%

speed ups do not count because it wouldnt make much sense

# Changelog

:cl:  
tweak: vehicles now respect slowdown modifiers
/:cl:

### DIFF
--- a/code/datums/components/riding.dm
+++ b/code/datums/components/riding.dm
@@ -158,11 +158,16 @@
 	if(user.incapacitated())
 		Unbuckle(user)
 		return
+	
+	var/slowness = 0
+	if(ishuman(user))
+		var/mob/living/carbon/human/H = user
+		slowness = H.dna?.species.movement_delay(H)
 
-	if(world.time < last_vehicle_move + ((last_move_diagonal? 2 : 1) * vehicle_move_delay * CONFIG_GET(number/movedelay/run_delay))) //yogs - fixed this to work with movespeed
+	if(world.time < last_vehicle_move + ((last_move_diagonal? 2 : 1) * (vehicle_move_delay + slowness) * CONFIG_GET(number/movedelay/run_delay))) //yogs - fixed this to work with movespeed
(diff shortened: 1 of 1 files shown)


Marmio64:
but this is like the entire reason you get the vehicle

Cartlord:
What is the purpose for this change? How does someone being hurt impair their ability to push down on a gas pedal? How does someone wearing somewhat bulky clothing noticeably slow down an _A.T.V._? How does someone being under the influence of drugs translate to them being unable to drive as fast?
You claim "speed ups do not count because it wouldnt make much sense", while ignoring that **nothing** you're doing here makes sense.

redmoogle:
> What is the purpose for this change? How does someone being hurt impair their ability to push down on a gas pedal? How does someone wearing somewhat bulky clothing noticeably slow down an _A.T.V._? How does someone being under the influence of drugs translate to them being unable to drive as fast? You claim "speed ups do not count because it wouldnt make much sense", while ignoring that **nothing** you're doing here makes sense.

What about bikes

Marmio64:
> > What is the purpose for this change? How does someone being hurt impair their ability to push down on a gas pedal? How does someone wearing somewhat bulky clothing noticeably slow down an _A.T.V._? How does someone being under the influence of drugs translate to them being unable to drive as fast? You claim "speed ups do not count because it wouldnt make much sense", while ignoring that **nothing** you're doing here makes sense.
> 
> What about bikes

make slowdowns AND speedups apply to only scooter subtypes (including skateboards) then

ynot01:
sake of balance

JamieD1:
Yeah you kinda get the vehicle for this reason, not sure if I can support this.

JamieD1:
Defeats the point of vehicles,

PR #15351, opened 2022-08-17
lowers secway tankiness

ynot01:

300 -> 60

it blocks bullets so it basically quadruples your health

# Changelog

:cl:  
tweak: significantly lowered secway integrity
/:cl:

### DIFF
--- a/code/modules/vehicles/secway.dm
+++ b/code/modules/vehicles/secway.dm
@@ -4,7 +4,7 @@
 	desc = "A brave security cyborg gave its life to help you look like a complete tool."
 	icon_state = "secway"
 	key_type = /obj/item/key/security
-	max_integrity = 40
+	max_integrity = 60
 
 /obj/vehicle/ridden/secway/Initialize()
 	. = ..()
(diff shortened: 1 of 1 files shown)


Marmio64:
40 is a little bit low, a fellow officer could accidently miss 2 lasers and hit you and now your 5000 credit toy is utterly destroyed. It'd be fine if you made it so the secway is only temporarily destroyed and you can fix it up with a welder or something (maybe make it drop a wreckage when destroyed?), but otherwise 40 health is just too low. How about 100 health, or maybe 80?

ynot01:
> 40 is a little bit low, a fellow officer could accidently miss 2 lasers and hit you and now your 5000 credit toy is utterly destroyed. It'd be fine if you made it so the secway is only temporarily destroyed and you can fix it up with a welder or something (maybe make it drop a wreckage when destroyed?), but otherwise 40 health is just too low. How about 100 health, or maybe 80?

40 hp still increases your effective health by 40, it is still strong, i will look into repairs though

ilikesaying:
40 health with no armor, only able to survive like 3 hits

code/modules/vehicles/secway.dm
```diff
@@ -4,6 +4,7 @@
 	desc = "A brave security cyborg gave its life to help you look like a complete tool."
 	icon_state = "secway"
 	key_type = /obj/item/key/security
+	max_integrity = 40
```
JamieD1:
```suggestion
	max_integrity = 60
```

JamieD1 (approved):

PR #15352, opened 2022-08-17
you can now repair the secway with a welder

ynot01:

atomization

each repair restores 20 integrity

#15351

# Changelog

:cl:  
tweak: you can now repair a secway with a welder
/:cl:

### DIFF
--- a/code/modules/vehicles/secway.dm
+++ b/code/modules/vehicles/secway.dm
@@ -10,3 +10,21 @@
 	var/datum/component/riding/D = LoadComponent(/datum/component/riding)
 	D.vehicle_move_delay = 1
 	D.set_riding_offsets(RIDING_OFFSET_ALL, list(TEXT_NORTH = list(0, 4), TEXT_SOUTH = list(0, 4), TEXT_EAST = list(0, 4), TEXT_WEST = list( 0, 4)))
+
+/obj/vehicle/ridden/secway/welder_act(mob/living/user, obj/item/I)
+	. = ..()
+	if(user.a_intent == INTENT_HARM)
+		return FALSE
+
+	if(integrity == max_integrity)
+		to_chat(user, span_warning("[src] is already in good condition!"))
+		return FALSE
+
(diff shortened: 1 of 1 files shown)


Marmio64:
dont secways disappear when they run out of integrity? Do you plan to make it so they don't disappear and are instead wrecked and you can repair them? Otherwise this sounds pretty useless if merged with the other one that makes em only have 40 integrity since that'll very likely run out fast enough that you wont even have a chance to repair.

ynot01:
> dont secways disappear when they run out of integrity? Do you plan to make it so they don't disappear and are instead wrecked and you can repair them? Otherwise this sounds pretty useless if merged with the other one that makes em only have 40 integrity since that'll very likely run out fast enough that you wont even have a chance to repair.

this is to account for "accidental shots", if you are using it in combat and it breaks then you will have to just get a new one

Marmio64:
> > dont secways disappear when they run out of integrity? Do you plan to make it so they don't disappear and are instead wrecked and you can repair them? Otherwise this sounds pretty useless if merged with the other one that makes em only have 40 integrity since that'll very likely run out fast enough that you wont even have a chance to repair.
> 
> this is to account for "accidental shots", if you are using it in combat and it breaks then you will have to just get a new one

well, you mean accidental **shot**? Because 40 health is only enough to stop one accident, which you'd then have to get out of  the secway in the middle of the fight to stop to repair it so another accident doesn't instantly delete 5000 credits. If this is your idea of balancing the secway im really not for it and would much rather have it so the secway is only temporarily out of a fight when it takes a hit or two and you need to go get it repaired. I also hope that the vehicle speed nerf and secway nerf are considered options rather than you want them both added because otherwise what point is there in spending 5000 credits to get 40 extra health that is also slow, cumbersome to use, easily stolen, and makes it really annoying to maneuver around your buds, when you could get 6 laserguns for that price.

JamieD1 (approved):

PR #15389, opened 2022-08-20
[MOOD] Being happy slightly heals you

redmoogle:

On average 0.05hp per second (of both BRUTE and BURN) requires organic bodyparts to work
You have to be happy enough though (Im feeling pretty good or better)

# Wiki Documentation

Mood heals now
# Changelog

:cl:  
rscadd: Being happy slowly heals you
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -52,6 +52,12 @@
 	name = get_visible_name()
 
 	if(stat != DEAD)
+		var/datum/component/mood/moody = GetComponent(/datum/component/mood)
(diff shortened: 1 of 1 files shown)


Mqiib:
As a mood user I love this, but I'm not 100% sure on the healing rate. Like yeah average of 0.2/life tick doesnt seem like a lot but it _can be_ a lot. Maybe reduce it to like a 30% chance to heal per life tick to keep it as a "very slow passive recovery" rather than a "very slow constant regeneration." Also make it regen stamina slightly too because it's thematic.

JamieD1 (approved):

PR #15394, opened 2022-08-20
Whiteship is now random instead of map based

redmoogle:

helps with variety instead of getting the same shuttle for the 500th time

# Wiki Documentation

No longer map based

# Changelog

:cl:  
tweak: Whiteship is no longer map based
/:cl:


patpol4:
Not sure if they all fit all docks, some might have different airlock placement or be too large and go into walls when docked, remember to make sure they fit.

redmoogle:
fair point

redmoogle:
> Not sure if they all fit all docks, some might have different airlock placement or be too large and go into walls when docked, remember to make sure they fit.

the only issue I see is with the delta variant and box station where it will take out some random lattices

ToasterBiome (approved):

PR #15399, opened 2022-08-20
Adds Ricoshots (makes flipping coins deflect energy projectiles)

ChesterTheCheesy:
Straight up rips ricoshots from Ultrakill to add some skill based flavor to ranged fights if anyone wants to be stylish.

All coins now, upon being hit by an energy projectile (lasers, disablers), do the flip animation. If it's hit again while in the animation state, the projectile will be deflected into a nearby mob picked at random from coin's sight.
After being deflected, the projectile gets a 1.5x boost to it's damage.
If there are no possible targets, the shot gets deflected into random direction, but cannot hit the shooter.
If there's more than one target, then the projectile gets **SPLIT** into two, in practice a copy of the projectile is generated and shot at another target.

ricoshot

# Spriting
To make the coins new ability a bit more visible, i've added a little flashing animation to the top of the coin flip. (that extends the coin's hitbox slightly for easier aiming)

Oh yeah the ping sound effect that plays when the projectile gets deflected is straight ripped from Ultrakill.

# Changelog
:cl:  
rscadd: adds ricoshots to the game
imageadd: coins flash when shot  
/:cl:


code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,71 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
 			return FALSE//do not flip the coin
 		coinflip = pick(sideslist)
 		cooldown = world.time + 15
+
 		flick("coin_[cmineral]_flip", src)
 		icon_state = "coin_[cmineral]_[coinflip]"
-		playsound(user.loc, 'sound/items/coinflip.ogg', 50, 1)
+		if(flash)
+			SSvis_overlays.add_vis_overlay(src, icon, "flash", ABOVE_LIGHTING_LAYER, ABOVE_LIGHTING_PLANE, unique = TRUE)
+		playsound(loc, 'sound/items/coinflip.ogg', 50, 1)
```
tattax:
```suggestion
		playsound(loc, 'sound/items/coinflip.ogg', 50, TRUE)
```

code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,71 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
+			visible_message(span_danger("[P] ricochets off of [src]!"))
+
+			playsound(loc, 'sound/weapons/ricochet.ogg', 50, 1)
+			return BULLET_ACT_FORCE_PIERCE
+		else	//we instead flip the coin
+
+			spawn() flip(flash = TRUE) //we don't want to wait for flipping to finish in order to do the impact
+			return BULLET_ACT_TURF
+	else
+		. = ..()
```
tattax:
```suggestion
			return BULLET_ACT_FORCE_PIERCE
			
		//we instead flip the coin
		spawn() flip(flash = TRUE) //we don't want to wait for flipping to finish in order to do the impact
		return BULLET_ACT_TURF
	. = ..()
```

cuackles:
i like the feature but im a bit iffy on the sprite, it looks a bit to circular, but it also indicates what its doing correctly so its acceptable but also if you feel like changing it, i made something that could be used as a replacement 
glint

LazennG:
ultrakool

ChesterTheCheesy:
> i like the feature but im a bit iffy on the sprite, it looks a bit to circular, but it also indicates what its doing correctly so its acceptable but also if you feel like changing it, i made something that could be used as a replacement

aight i stylized it after your suggestion
trollge

i had to add a few frames including a big flash (its purpose is to extend the hitbox after all)

code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,70 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
 	to_chat(user, span_notice("You detach the string from the coin."))
 	return TRUE
 
-/obj/item/coin/attack_self(mob/user)
+/obj/item/coin/proc/flip(mob/user, flash = FALSE)
 	if(cooldown < world.time)
 		if(string_attached) //does the coin have a wire attached
-			to_chat(user, span_warning("The coin won't flip very well with something attached!") )
+			if(user)
+				to_chat(user, span_warning("The coin won't flip very well with something attached!") )
```
JohnFulpWillard:
```suggestion
				to_chat(user, span_warning("The coin won't flip very well with something attached!"))
```

code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,70 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
+					P.fire()
+			visible_message(span_danger("[P] ricochets off of [src]!"))
+
+			playsound(loc, 'sound/weapons/ricochet.ogg', 50, 1)
+			return BULLET_ACT_FORCE_PIERCE
+			
+		//we instead flip the coin
+		INVOKE_ASYNC(src, .proc/flip, null, TRUE) //we don't want to wait for flipping to finish in order to do the impact
+		return BULLET_ACT_TURF
+	. = ..()
```
JohnFulpWillard:
```suggestion
	return ..()
```
This is the end of a proc, there's no use for ``.`` to be given a value.

code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,70 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
 	return TRUE//did the coin flip? useful for suicide_act
 
+/obj/item/coin/attack_self(mob/user)
+	flip(user)
+
+/obj/item/coin/after_throw(datum/callback/callback) //get rid of the throw rotation
+	. = ..()
+	transform = initial(transform)
+
+/obj/item/coin/bullet_act(obj/item/projectile/P)
```
JohnFulpWillard, line 499: you should use a better var name than ``P``, what does P mean?

code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,70 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
+	if(P.flag == "laser" || P.flag == "energy" || istype(P, /obj/item/projectile/bullet/c38)) //energy projectiles get deflected (also det revolver because damn thats cool)
```
JohnFulpWillard: You can probably turn this into an early return.
```suggestion
	if(!P.flag == "laser" && !P.flag == "energy" && !istype(P, /obj/item/projectile/bullet/c38)) //energy projectiles get deflected (also det revolver because damn thats cool)
		return
```
then remove a tab from the rest of the code

code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,70 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
+		if(cooldown >= world.time)//we ricochet the projectile
+			var/list/targets = list()
+			var/turf/center = get_turf(src)
+			for(var/mob/living/T in viewers(world.view - 3, src))
```
JohnFulpWillard: 1. What does ``T`` mean? Can you use a better var name?
2. Why the ``world.view - 3``? It would be better to use a hardcoded radius so you can control how far people will notice it from. Yeah it might be on your screen, but if world.view is so high, how would you notice it from so far?

code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,70 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
 		sleep(1.5 SECONDS)
+		SSvis_overlays.remove_vis_overlay(src, managed_vis_overlays)
 		if(loc == oldloc && user && !user.incapacitated())
 			user.visible_message("[user] has flipped [src]. It lands on [coinflip].", \
  							 span_notice("You flip [src]. It lands on [coinflip]."), \
 							 span_italics("You hear the clattering of loose change."))
 	return TRUE//did the coin flip? useful for suicide_act
 
+/obj/item/coin/attack_self(mob/user)
+	flip(user)
```
JohnFulpWillard, line 493:
```suggestion
	. = ..()
	flip(user)
```

ChesterTheCheesy: literally why, no other attack_self proc requires it to call on its parent first

code/modules/mining/ores_coins.dm
ChesterTheCheesy: 1.T stands for target, and it's simply a way to distinguish temporary variables, i didn't make this convention
2. I guess i'll make it static

code/modules/mining/ores_coins.dm
ChesterTheCheesy: P goes for projectile, and its like this on every single bullet_act in the game, don't ask me why

code/modules/mining/ores_coins.dm
ChesterTheCheesy: fine, done

cuackles:
> > i like the feature but im a bit iffy on the sprite, it looks a bit to circular, but it also indicates what its doing correctly so its acceptable but also if you feel like changing it, i made something that could be used as a replacement
> 
> aight i stylized it after your suggestion
> trollge
> 
> i had to add a few frames including a big flash (its purpose is to extend the hitbox after all)
> 
> 

Perfect @KillerOrcaCora has to aprove it tho cus i it would be biased if i aproved it and stuff

code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,68 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
+
+/obj/item/coin/after_throw(datum/callback/callback) //get rid of the throw rotation
+	. = ..()
+	transform = initial(transform)
+
+/obj/item/coin/bullet_act(obj/item/projectile/P)
+	if(P.flag != "laser" && P.flag != "energy" && !istype(P, /obj/item/projectile/bullet/c38)) //only energy projectiles get deflected (also det revolver because damn thats cool)
```
Bibby0110: Use defines

code/modules/mining/ores_coins.dm
```diff
@@ -466,23 +466,68 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
...
+		return ..()
+		
+	if(cooldown >= world.time)//we ricochet the projectile
```
Bibby0110, line 503: Refactor the rest to use cooldown defines while you’re at it

Bibby0110 (changes requested):

patpol4:
pogging so hard rn

code/modules/mining/ores_coins.dm
ChesterTheCheesy: how

Bibby0110: https://github.com/yogstation13/Yogstation/blob/be95d7cc5e6a6b4eb9ca330c134bc0f3b0237564/code/__DEFINES/cooldowns.dm

ChesterTheCheesy: how

ChesterTheCheesy: can't do that since i plan to use the world.time in the future to narrow the time to hit the coin for ricoshot without affecting the timer of a coin flip

JamieD1 (approved):

PR #15417, opened 2022-08-21
Ticking backwards is bad

adamsong:

Life tickrate should not be less than 0

# Changelog

:cl:  
bugfix: fixed sleepers slowing you down way too much
/:cl:


ToasterBiome (approved):
erm what

PR #15420, opened 2022-08-22
Fixes Demon Chainsaw and changes description

ToasterBiome:

- Fixes sprites so it actually has an on sprite
- Fixes shitcode so it isn't a huge mess of repeating things
- Changes description from "anime reference here" to "Show me your dreams." (actual anime reference)

You can still hit yourself with lifesteal but since you deal 30 damage to heal 30 damage it does nothing. This is a lifesteal problem not a Demon Chainsaw probelm.

# Changelog

:cl:  
bugfix: gives demon chainsaw a proper "on" sprite
tweak: changes demon chainsaw description
/:cl:


KillerOrcaCora (approved):

code/game/objects/items/twohanded.dm
```diff
@@ -649,7 +650,7 @@
 	to_chat(user, "As you pull the starting cord dangling from [src], [on ? "it begins to whirr." : "the chain stops moving."]")
 	force = on ? force_on : initial(force)
 	throwforce = on ? force_on : initial(force)
-	icon_state = "chainsaw_[on ? "on" : "off"]"
+	icon_state = "[icon_name]_[on ? "on" : "off"]"
```
tattax, line 653: instead of having a var for it wouldn't it be better to just name the icon_state the name or have a update icon proc here so you can change it for the second chainsaw since that's what it's supposed to do?

ToasterBiome: > instead of having a var for it wouldn't it be better to just name the icon_state the name or have a update icon proc here so you can change it for the second chainsaw since that's what it's supposed to do?

You could do initial here I think and oh orca merged it

Lichton:
Should also add the actual chainsaw hitsound

ToasterBiome:
> Should also add the actual chainsaw hitsound

it has the chainsaw hitsound

cuackles:
i swear i had actually changed the anime reference thingy but aparently not

PR #15441, opened 2022-08-23
(port) Ethreal specific mutations Overload 

boodaliboo:
Ports:https://github.com/BeeStation/BeeStation-Hornet/pull/5814
Allows an Ethereal to overload their skin to cause a bright flash

:cl:  
rscadd: Added new species specific mutation
/:cl:


Cartlord:
does this block felinids from getting dwarfism?

VaelophisNyx:
felinids aren't a unique species iirc, they're just a Purrbation flag on humans
the felinid species is used for mutation toxins last I checked

boodaliboo:
> does this block felinids from getting dwarfism?

why would it

boodaliboo:
> felinids aren't a unique species iirc, they're just a Purrbation flag on humans the felinid species is used for mutation toxins last I checked

wrong

VaelophisNyx:
I am correct so far as  I can see, literally looking at the code right now;
the felinid species never gets set on your character; if you toggle it on it just applies cat ears & a tail when you initialize and leaves you a human.

VaelophisNyx:
also shock touch not being inert would be pretty rad, since it's normally a combo mutation (and that'd probably fix your code :) )

ghost:
Felenid content bad

code/modules/mob/living/carbon/human/species_types/ethereal.dm
```diff
@@ -30,7 +30,7 @@
 	screamsound = list('sound/voice/ethereal/ethereal_scream_1.ogg', 'sound/voice/ethereal/ethereal_scream_2.ogg', 'sound/voice/ethereal/ethereal_scream_3.ogg')
 	sexes = FALSE //no fetish content allowed
 	toxic_food = NONE
-	inert_mutation = SHOCKTOUCH
+	inert_mutation = SHOCKTOUCH,OVERLOAD
```
redmoogle: you'll have to refactor inert_mutations to accept lists or remove one of them as it currently only supports 1 mutation

redmoogle:
> I am correct so far as I can see, literally looking at the code right now; the felinid species never gets set on your character; if you toggle it on it just applies cat ears & a tail when you initialize and leaves you a human.

I am looking at the code it litterarly does

```
/proc/purrbation_apply(mob/living/carbon/human/H, silent = FALSE)
	if(iscatperson(H))
		return
	if(!silent)
		to_chat(H, "Something is nya~t right.")
		playsound(get_turf(H), 'sound/effects/meow1.ogg', 50, 1, -1)

	if(!ishumanbasic(H))
		var/obj/item/organ/cattification = new /obj/item/organ/tail/cat()
		var/old_part = H.getorganslot(ORGAN_SLOT_TAIL)
		cattification.Insert(H)
		qdel(old_part)
		cattification = new /obj/item/organ/ears/cat()
		old_part = H.getorganslot(ORGAN_SLOT_EARS)
		cattification.Insert(H)
		qdel(old_part)
		H.regenerate_icons()
		return

	H.set_species(/datum/species/human/felinid)
```

VaelophisNyx:
bruh how the fuck did I miss the set species part

at any rate this change is
absurdly dumb

Mqiib:
Ignoring the issue of giving a donor race a mechanical difference, the cat claw mutation is a little strong. +6 punch damage (which is actually split into punchdamagehigh and punchdamagelow) is a LOT when you're not also raising the stun threshold by the same amount. If the pr even gets accepted it'll need to affect physiology punch damage rather than species and even then would need to be toned down to like +4, applying to high and low damage as well as the stun threshold.

NCRLouTenant:
Okay but turning an Ethereal into a walking flash would make them as Sec cool as hell.

(Definitely not just me playing favorites with the species...)

Mqiib:
Make the overload just flash and confuse slightly, no stun. Having a mass stun on demand is a little much.

PR #15447, opened 2022-08-24
Mantis blade buff

enkidienne:
It has been brought to my attention that the mantis blade is weak and mostly just bloat at this point. This aim at making it viable with current balance:
 Gave the mantis blade the following:
	armour_penetration = 30
	wound_bonus = 20
	bare_wound_bonus = 20
 used to have none of these (0).
This should improve its performance against armored targets and give it a hell of a lot more chance to wound. 

# Wiki Documentation
see above

# Changelog
:cl:  
tweak: changed the stats of the mantis blade  
/:cl:

### DIFF
--- a/code/game/objects/items/mantis.dm
+++ b/code/game/objects/items/mantis.dm
@@ -9,8 +9,8 @@
 	flags_1 = CONDUCT_1
 	force = 20
 	armour_penetration = 30
-	wound_bonus = 35
-	bare_wound_bonus = 40
+	wound_bonus = 40
+	bare_wound_bonus = 50
 	w_class = WEIGHT_CLASS_NORMAL
 	sharpness = SHARP_EDGED
 	attack_verb = list("attacked", "slashed", "stabbed", "sliced", "torn", "lacerated", "ripped", "diced", "cut")
--- a/code/game/objects/items/mantis.dm
+++ b/code/game/objects/items/mantis.dm
@@ -9,8 +9,8 @@
 	flags_1 = CONDUCT_1
 	force = 20
 	armour_penetration = 30
-	wound_bonus = 40
-	bare_wound_bonus = 50
+	wound_bonus = 20
+	bare_wound_bonus = 20
 	w_class = WEIGHT_CLASS_NORMAL
 	sharpness = SHARP_EDGED
 	attack_verb = list("attacked", "slashed", "stabbed", "sliced", "torn", "lacerated", "ripped", "diced", "cut")
--- a/code/game/objects/items/mantis.dm
+++ b/code/game/objects/items/mantis.dm
@@ -8,6 +8,9 @@
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	flags_1 = CONDUCT_1
 	force = 20
+	armour_penetration = 30
+	wound_bonus = 35
+	bare_wound_bonus = 40
 	w_class = WEIGHT_CLASS_NORMAL
 	sharpness = SHARP_EDGED
 	attack_verb = list("attacked", "slashed", "stabbed", "sliced", "torn", "lacerated", "ripped", "diced", "cut")


Mqiib:
Those wound bonuses are absurdly high. Reduce them to like 10 and 10

enkidienne:
> Those wound bonuses are absurdly high. Reduce them to like 10 and 10

those are similar values of that of kitchen knifes and welding tool. I know those are very high and its meant. The blades as they currently are, are very weak and you have no real reason to buy them over something like an esword.

Mqiib:
20 force alone has decent wounding power. it doesn't need near-guaranteed severe wounds on every hit. You can one-click rip a limb off with this which is a little much.

enkidienne:
> 20 force alone has decent wounding power. it doesn't need near-guaranteed severe wounds on every hit. You can one-click rip a limb off with this which is a little much.

Keeping in mind that it does strike twice per click, and that even a petty tool like a welding tool has higher bare bonus than 10, how about 20-25 ? its a weapon that costs 14TC in total, it should be worth its price.

Mqiib:
> > 20 force alone has decent wounding power. it doesn't need near-guaranteed severe wounds on every hit. You can one-click rip a limb off with this which is a little much.
> 
> Keeping in mind that it does strike twice per click, and that even a petty tool like a welding tool has higher bare bonus than 10, how about 20-25 ? its a weapon that costs 14TC in total, it should be worth its price.

have you tested it

enkidienne:
> 

at the current values? Yes, its not a sure-wound every times, but it is a skyhigh chance.

VaelophisNyx:
keep in mind that mantis blades can't be disarmed and both attack at the same time, and have block chance
they're REALLY good already

enkidienne:
> keep in mind that mantis blades can't be disarmed and both attack at the same time, and have block chance they're REALLY good already

i did just halve the stats to both 20, which is in the upper end of high damage weapons.

ynot01:
I did some testing on your branch's latest commit

This is how the mantis blade(s) hold up to the energy sword when hit 1, 2, or 3 times in the head

(Wounds are mostly RNG so results may vary)

One set of testing was done with only one mantis blade as there is a PR open to make that the option instead

(Worth noting that the captain, security, and cargo technician all fell into crit immediately on 3 hits when dual wielding mantis blades)



I thought it was strange how the captain died on two-hit testing and the officer didn't so I did that 3 times over.

enkidienne:
> I did some testing on your branch's latest commit
> 
> This is how the mantis blade(s) hold up to the energy sword when hit 1, 2, or 3 times in the head
> 
> (Wounds are mostly RNG so results may vary)
> 
> One set of testing was done with only one mantis blade as there is a PR open to make that the option instead
> 
> (Worth noting that the captain, security, and cargo technician all fell into crit immediately on 3 hits when dual wielding mantis blades)
> 
> 
> 
> I thought it was strange how the captain died on two-hit testing and the officer didn't so I did that 3 times over.
> 
> 

To me this looks quite fair, performances maybe a little under esword when doubled, all considering esword is 8TC (if I remember right) and that a single blade is 7TC. Considering you need both blades for intended results, AKA 14TC, you pay 6 TC more for the stealth, retractable and impossible to drop, which is in my mind quite "balanced".

isotope436:
at first i didnt think this was that good of an idea but honestly this is more balanced than i though and makes sense

KillerOrcaCora:
This seems like we are entering Double e-sword territory also known as (Easy muderbone weapon purchase)

isotope436:
> This seems like we are entering Double e-sword territory also known as (Easy muderbone weapon purchase)

this actually puts it on par with the normal esword not the DEsword as all chances dont stack so its only really viable if you spend the 14 tc to get both its basically half a esword for 7tc thats concealable

JamieD1:
> <snip>

And your conclusion?

JamieD1:
And your conclusion @ynot01 ?

Bibby0110 (approved):
Ynot's tests look fine. Seems to be on par with e-sword. We can always nerf it if it turns out to be too powerful at double blade levels

PR #15463, opened 2022-08-26
Humanoid minds are not suited for traveling the bluespace backroads

Mqiib:
Teleporting with a bag of holding is now very much a bad time (you get nauseated and confused). Also bluespace dust requires at least 5 units to teleport because if it's less than that it's not strong enough to teleport your whole body. Fatass.

# Changelog

:cl:  

tweak: BoH teleport interference is now even less fun
tweak: Bluespace dust requires at least 5 units on application to teleport you (1/4 of a crystal)
/:cl:


LoliconSlayer:
i ded

JamieD1 (approved):

PR #15466, opened 2022-08-26
gives ipc healing from oil welder fuel and teslium like androids preternis and synths

boodaliboo:
consistancy
:cl:  
rscadd: Added it so ipcs heal from oil welder fuel and teslium like other robot species do
/:cl:


VaelophisNyx:
thing is IPCs don't like...process 90% of chems
and can't really...drink?

redmoogle:
Injection & Process Flags

PR #15489, opened 2022-08-28
Overhauls Brazil

VaelophisNyx:

Brazil is now The Maint Rooms
an off-brand Backrooms for SS13. Still features all three penance givers, just now with more content to explore around there if you don't feel like returning to the station fast, or are bored/dumb.

ss (2022-08-28 at 02 08 17)

Gorilla has no AI
3x3 spawners are set so the drone ruin can't spawn, so we don't have Brazil Drones.

# Wiki Documentation

If brazil is documented this will need to be updated.

# Changelog

:cl:  
tweak: Brazil is now The Maint Rooms
/:cl:


VaelophisNyx:
Should note; tested in game

yogstation/code/game/area/areas/centcom.dm
```diff
@@ -31,4 +31,22 @@
 	blob_allowed = FALSE
 	has_gravity = TRUE
 	noteleport = TRUE
-	flags_1 = NONE
\ No newline at end of file
+	atmos = FALSE
```
Bibby0110: I don’t think this does anything

VaelophisNyx: it apparently disables atmos in the area; according to the code comments

ghost:
I don't think that an eldritch location where you are sent by getting sacrificed should look like this

MusselMain:
Eyo, brazil snackbar!!!!

VaelophisNyx:
> I don't think that an eldritch location where you are sent by getting sacrificed should look like this

there's not really much else it could look like and be interesting though, is the thing

PR #15499, opened 2022-08-28
You can now tell if your heart exploded

CMOisLing:
Gonna be honest haven't tested this but I think it makes sense.

Since it deletes itself theres no chance it will give multiple messages so it should always give one before it deletes itself.

:cl:  

bugfix: You can now tell if your heart exploded 100% of the time

/:cl:

### DIFF
--- a/yogstation/code/modules/surgery/organs/heart.dm
+++ b/yogstation/code/modules/surgery/organs/heart.dm
@@ -17,8 +17,7 @@
 	if(SEND_SIGNAL(owner, COMSIG_HAS_NANITES))
 		SEND_SIGNAL(owner, COMSIG_NANITE_ADJUST_VOLUME, nanite_boost)
 	else
-		if(prob(25))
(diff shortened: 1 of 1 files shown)


yogstation/code/modules/surgery/organs/heart.dm
```diff
@@ -17,8 +17,7 @@
 	if(SEND_SIGNAL(owner, COMSIG_HAS_NANITES))
 		SEND_SIGNAL(owner, COMSIG_NANITE_ADJUST_VOLUME, nanite_boost)
 	else
-		if(prob(25))
-			to_chat(owner, "<span class = 'userdanger'>You feel your heart collapse in on itself!</span>")
 		if(owner)
+			to_chat(owner, "<span class = 'userdanger'>You feel your heart collapse in on itself!</span>")
```
tattax:
```suggestion
			to_chat(owner, span_userdanger("You feel your heart collapse in on itself!"))
```

JamieD1 (approved):

PR #15503, opened 2022-08-29
Chaplain monk staff is now nonlethal

Mqiib:
Makes the chaplain monk staff null rod variant do stamina damage only. Increases force to standard 18 and gives it 20 AP. This grants chaplains a nice nonlethal deterrent weapon so they don't have to violently beat every gamer that tries to steal the wood from their pews. 

# Wiki Documentation

New monk staff stats: 
Damage type: stamina
Force: 18
AP: 20
block chance: 40 (unchanged)

# Changelog

:cl:  

tweak: Monk staff null rod variant now does stamina damage!
tweak: Monk staff null rod variant force and AP buffed in light if its new nonlethal qualities
/:cl:


Marmio64:
it'd be better if it was done like the bo staff is done to be honest, where it has a weak lethal mode (that you'd use against constructs) and then the stamina damage mode.

Mqiib:
> it'd be better if it was done like the bo staff is done to be honest, where it has a weak lethal mode (that you'd use against constructs) and then the stamina damage mode.

yeah but that would require SIGNIFICANTLY more work remaking it into a twohanded weapon and playing with the code and stuff

Marmio64:
> > it'd be better if it was done like the bo staff is done to be honest, where it has a weak lethal mode (that you'd use against constructs) and then the stamina damage mode.
> 
> yeah but that would require SIGNIFICANTLY more work remaking it into a twohanded weapon and playing with the code and stuff

i dont think making it two handed is a necessity, you can skip that part and cherry pick what you'd like from bo staff code, just having it so it does lethal damage by default and when in disarm intent it deals stamina damage. I do assume you're making this so that you can use this as a pacifist when you're chaplain or at the least for someone else so it really wouldn't matter if it's just for that purpose, but it'd also have good use for people who still want to fight if it happens to be a cult round but they also want a nonlethal weapon.

neeshacark:
having it function like that would be ideal yeah, but at this point im not inclined to ask mqiib to do any more hard work for me when they are already doing this for something that like only i will even use.

JamieD1 (approved):

PR #15519, opened 2022-08-30
.357 Alternative Ammo Update

Skrem7:

Adds five new .357 variants for usage with the Syndicate revolver, though any .357 revolver can technically use them.

Standard .357 does 40 damage with -70 wound bonus. Simple enough. Its cost has been reduced to 1 TC from 2 TC.

.357 Ironfeather is less damaging buckshot, a total of 8.5 damage per pellet over six pellets (51 damage) with the same spread. Its wound bonuses are slightly higher than buckshot to try and make it capable of dishing out wounds, and it loses just slightly less damage and wound chance over distance. It costs 1 TC per loader.

.357 Nutcracker is effectively an upgraded breaching slug, a 20 damage -60 wound bonus bullet that does massive damage to airlocks and windows, moreso than the breaching slug. It costs 2 TC per loader.

.357 Metalshock deals 30 brute damage and also attempts to shock any target hit. If the shock occurs, the target takes 10 burn damage internally (ignores armor) and 5 confusion stacks. For comparison, Shock Touch deals 15/15. It costs 2 TC per loader.

.357 Heartpiercer deals 35 damage with 35 AP, but also penetrates through exactly one target. It costs 3 TC per loader.

.357 Wallstake deals 25 damage with -50 wound bonus. It is not sharp. However, it throws targets two tiles with a force of two. Impacting a wall resolves 10 damage against melee armor for whichever limb is chosen to take damage. It is also a hardstun if this occurs, which can be easily chained. It costs 4 TC per loader.

Overall lethality:
- .357: 3 shot crit, 4 against standard vest, 7 against bulletproof
- .357 Ironfeather (PBS): 2 shot crit, 3 against standard vest, 5 against bulletproof
- .357 Nutcracker: 5 shot crit, 8 against standard vest, 13 against bulletproof
- .357 Nutcracker (against terrain): 1 against reinforced window, 2 against airlock, 3 against reinforced airlock
- .357 Metalshock (no insuls): 3 shot crit, 4 against standard vest, 5 against bulletproof
- .357 Metalshock (with insuls): 4 shot crit, 5 against standard vest, 9 against bulletproof
- .357 Heartpiercer: 3 shot crit, 3 against standard vest, 4 against bulletproof (penetrates a single target)
- .357 Wallstake (no wall): 4 shot crit, 6 against standard vest, 10 against bulletproof
- .357 Wallstake (assuming only wall hits): 3 shot crit, 4 against standard vest AND security jumpsuit, 6 against bulletproof AND security jumpsuit

*Note that the direction and reliability of the throw in knocking someone against a wall is IFFY at best. It's using meteorslug code.

Variations and fun, different ammo types are what I believe can firmly elevate ballistics above energy weapons regarding their fun to use. As the .357 is more comparable to the stechkin in its costs and values now, I thought it would be appropriate to give it alternative ammunition to try and differentiate it a bit more. As revolver alternative ammo tends to be a bit on the quirky side, I wanted to make ammunition that reflected this general trend.

It's all stuff that's been seen in ballistics before, but not really in a replicable or easily accessed format. Some of these could certainly use number or intellectual passovers, especially regarding cost, damage, and effect. I'm more than willing to get some people who are better at balance on hand.

**Specific justifications for...**

Default .357: Having four times the cost of ammunition of the stechkin just hurts, you're firing away TC like it's nothing, and compared to the stechkin and its ability to buy AP ammunition, you're just not as economical, even going against normal targets. It is universally a lot safer and better to just buy a stechkin and load up on AP magazines.

Ironfeather: Shotgun revolver is something that hasn't existed up to now. It's basically like buying buckshot over the default slug round that is standard .357.

Nutcracker: This round type was requested, plus it's a badass alternative to using C4 or whatever else to break airlocks. Loud, fast, and cool.

Metalshock: Its shots-to-crit only matter against bulletproof, which give it a ghetto-AP effect. The confusion stacks are light for the sake of not going overkill with RNG you-move-randomly effects, which is just absurd if you're already being shot multiple times. It's also completely countered by insuls, meaning it won't work on the captain or anyone else that's immune to shocks for whatever reason.

Heartpiercer: It's on the pricier side, you get to down a normal secoff in one fewer bullet, AND you get to down bulletproof vests for one bullet fewer than AP stechkin. The piercing is what makes it so much more expensive, it won't always come into play, but it's scary to go up against. 3 TC per loader hurts, but you're paying for what you get.

Wallstake: By far the iffiest of all these ammo types, you can quite literally just spam hardstun at range with this. Some would say this is absurd. I would like to point to the ebow which also costs 10 TC (rev + wallstake = 10), but it infinitely recharges, is more compact, more stealthy, and doesn't require the target to hit a wall in order to stun them. All you might need are those few seconds of stun to really dig into someone. If spam hardstun into a wall was a huge issue before, then it seems unlikely that the baseball bat would be slept on as much. Realistically, it's hard to get people to slam into walls, especially with how finicky throw code is on projectiles (testing it was... unstable at best). You also have to catch someone very close to a wall in order to actually stun them with it. It's certainly something to keep an eye on, but I think it's worth the test to see just how good it is.

# Spriting
Base sprite for comparison:
image

.357 Ironfeather: 
image

.357 Nutcracker: 
image

.357 Metalshock: 
image

.357 Heartpiercer:
image

.357 Wallstake: 
image

# Wiki Documentation

Images and values to be uploaded to Guide to Combat and Syndicate Items
Could maybe use a mention on Detective page

# Changelog

:cl:  
rscadd: Adds .357 Ironfeather for 1 TC, a round type that fires a weaker buckshot clump, but has slightly more favorable damage and wound dropoff.
rscadd: Adds .357 Nutcracker for 2 TC, a round type that's effectively an upgraded breaching slug, doing half the damage of a normal round but massive damage to airlocks and windows.
rscadd: Adds .357 Metalshock for 2 TC, a round type that converts some brute damage into a shock that confuses targets. Insuls prevent the damage and the effect, making it a weaker bullet if you have immunity.
rscadd: Adds .357 Heartpiercer for 3 TC, a round type comparable to a standard AP round, but has the ability to penetrate through one target to hit a total of two.
rscadd: Adds .357 Wallstake for 4 TC, a round type which does lessened, blunt damage, but throws targets like the baseball bat or meteorshot. Will only stun if a target hits a wall.
tweak: Normal .357 speedloader costs 1 TC from 2 TC
imageadd: Adds sprites for the three new ammo types
experimental: Who said ballistics can't be quirky?
/:cl:


00ze-cyclone:
I am scared of wallstake but I trust your judgement and testing for that one.
Ammunition like that is exactly what the revolver needed.

Skrem7:
> I am scared of wallstake

Good.

KillerOrcaCora (approved):
looks fine

code/modules/projectiles/projectile/bullets/revolver.dm
```diff
@@ -78,3 +78,63 @@
...
+
+/obj/item/projectile/bullet/a357/wallstake
+	name = ".357 Wallstake bullet"
+	damage = 25 //Consider that they're also being thrown into the wall
+	wound_bonus = -50 //Minor chance of dislocation from the bullet itself
+	sharpness = SHARP_NONE //Blunt
+
+/obj/item/projectile/bullet/a357/wallstake/on_hit(atom/target, blocked = FALSE)
+	. = ..()
+	if(isliving(target) && ismovable(target)) //Unlike meteorslugs, these are smaller and meant to knock bodies around, not ANYTHING
```
tattax: it'll return false if you check isliving first and will not check ismovable and every thing that is living is already movable so

code/modules/projectiles/projectile/bullets/revolver.dm
```diff
@@ -78,3 +78,63 @@
...
+	if(iscarbon(target))
+		var/mob/living/carbon/C = target
+		if(C.electrocute_act(10, src, 1, FALSE, FALSE, FALSE, FALSE, FALSE)) //10 extra burn damage, should NOT stun
+			C.confused += 5 //15% chance for minor, 5% for major randomness per movement
+			return ..()
+	
+	else if(isliving(target)) //So that it works on simple mobs, too
+		var/mob/living/L = target
+		L.electrocute_act(10, src, 1, FALSE, FALSE, FALSE, FALSE)
+		return ..()
```
tattax:
```suggestion
	if(iscarbon(target))
		var/mob/living/carbon/C = target
		if(C.electrocute_act(10, src, 1, FALSE, FALSE, FALSE, FALSE, FALSE)) //10 extra burn damage, should NOT stun
			C.confused += 5 //15% chance for minor, 5% for major randomness per movement
		return ..()
	if(isliving(target)) //So that it works on simple mobs, too
		var/mob/living/L = target
		L.electrocute_act(10, src, 1, FALSE, FALSE, FALSE, FALSE)
	. = ..()
```

tattax: these are tested yes? seemed like to much needless stuff with the elses when you have returns and if you are going to return ..() after calling . = ..(), either do it likes this (^) or call just return

code/modules/projectiles/projectile/bullets/revolver.dm
Skrem7: Hilariously via testing it was working fine. If you say it's redundant though, I'll shorten it down.

code/modules/projectiles/projectile/bullets/revolver.dm
Skrem7: Something I probably overlooked, it works fine but I can test this cleaner write-up.

PR #15533, opened 2022-08-31
New ash drake fire buff doesn't stack 5 morbillion fire stacks directly in front of its snout

ToasterBiome:

whgat if you were hit 500 times what if though that would be funny

line already has anti-frustration behavior in that you can only be hit once a line anyway

keeps bossfight cool and challenging without instantly killing you 

# Changelog

:cl: ToasterBiome, Lazenn, VaelophisNyx, SomeguyManperson, ynot01, Mqiib, Jesus Christ himself
tweak: ash drake fire line no longer stacks multiple in front of itself
/:cl:


neeshacark:
got killed so hard by drake early this morning and was super confused because i had real good gear but the fire was just shredding me.  this explains it.

JamieD1 (approved):

PR #15535, opened 2022-08-31
Limits Abductors to 25 Pop

VaelophisNyx:

mmm I love the antag that can cause absolute nightmare chaos on *60 pop* doing it at 15 pop, where there's noone to stop or fix it :)

# Changelog

:cl:  
tweak: Midround Abductors are now pop limited
/:cl:

### DIFF
--- a/code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm
+++ b/code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm
@@ -669,6 +669,7 @@
 	requirements = list(101,101,101,80,60,50,30,20,10,10)
 	repeatable = TRUE
 	var/datum/team/abductor_team/new_team
+	minimum_players = 25
 
 /datum/dynamic_ruleset/midround/from_ghosts/abductors/ready(forced = FALSE)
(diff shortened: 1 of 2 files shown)


VaelophisNyx:
here before the ided


wait

ToasterBiome:
this is only for dynamic doe

VaelophisNyx:
not anymore ;P

GraveHat:
You weren't kidding with that I ded tag

ghost:
Abductors don't kill people and their organ effects can even be positive so there is no reason to do like this

VaelophisNyx:
> Abductors don't kill people and their organ effects can even be positive so there is no reason to do like this

plasma organ
spider organ
all access organ

abductors also tend to be shitters and have gotten their own rules revised like 6 times now for it

GraveHat:
> > Abductors don't kill people and their organ effects can even be positive so there is no reason to do like this
> 
> plasma organ spider organ all access organ
> 
> abductors also tend to be shitters and have gotten their own rules revised like 6 times now for it

that and the brainwashing of the entirety of lowpop crew is not as appealing as it sounds.

JamieD1 (approved):

PR #15550, opened 2022-08-31
Armor Effectiveness Rescale

Mqiib:
Low armor is better, high armor is worse. Armor stacking BTFO. Becoming immune to damage is now near impossible.

X-axis is armor value, Y-axis is damage fraction. The straight line is the old value while the curved is new. The intersection point is at 75 armor, so it really only nerfs REALLY HIGH armor values.

image

# Wiki Documentation

Armor damage reduction rescaled from (100 * (100-armor))% to (100 * (0.982^armor))%

# Changelog

:cl:  @Rdrazga , Mqiib

tweak: Armor damage reduction has been rescaled to be exponential
experimental: This is paired with the AP rework and will ruin balance until we know how to fix it all
/:cl:


Mqiib:
eh nevermind

PR #15551, opened 2022-09-01
removes all cybernetics & upgraded organs from medbay techfab

TheRyeGuyWhoWillNowDie:
makes robotics more relevant, makes it harder for the medlizard metagang to powergame by giving their metabuddies every cybernetic implant imaginable. also removes posibrain from medbay techfab because medbay techfab inexplicably had the ability to print posibrains

# Changelog

:cl:  
rscdel: removes all upgraded cybernetic organs and cybernetic implants from the medbay techfab. want your powergamer implants? visit robotics
/:cl:


VaelophisNyx:
robotics is relevant already this just makes implants and upgraded organs impossible to get

GraveHat:
Unnecessary to remove all those from medbay, robotics is too understaffed to even supply these to the medical doctors in the first place you are essentially making it so that implants will never see use outside of highpop and BnE lowpop. Medbay is the only one that even uses implants despite robotics already having access to them, the only other reason you see doctors get implanted in the first place is that it's convenient, they won't bother to go to robotics and would just end up doing something else in medbay. This doesn't even take into account how implants and upgraded cyber organs are even a problem in the first place.

Besides this would make robotics even more of the powergamer department than it already is. 2 Roboticists can't bother to keep up with all this when even silicons themselves get problems of getting upgraded because of some other factor making the roboticists null and void and not in robotics.

KillerOrcaCora:
Nice, you somehow tricked yourself into thinking people get basic cybernetics solely for powergaming...

TheRyeGuyWhoWillNowDie:
> Nice, you somehow tricked yourself into thinking people get basic cybernetics solely for powergaming...

basic cybernetics (normal cybernetic organs) arent removed and medbay can still print because they actually need them. the powergaming stuff is just removed like gamer eyes and toolset implants

Maia-J:
> the powergaming stuff is just removed like gamer eyes and toolset implants

Pal, if you wanna keep blindly hunting so called "powergaming", just remove the whole of science with all the stuff they provide, problem solved. This would just turn the implants 98% of the playerbase don't use into something not even the remaining 2% bother with. Not to mention that it expects cooperation between departments, which from my ipc reviving experience I can assure you is impossible.

PR #15558, opened 2022-09-01
NTOS Download based on distance

ynot01:

Download speed now increases the closer you are to a NTNet Relay (Usually in tcomms)

Uses euclidean distance, so think of it as a circle instead of a square

At 0 tiles distance, you will download 3x faster

At 100 tiles distance, you will download the same speed as before (1x)

Default network card will now download at 0.5-1.5 GQ/s
Advanced network card will now download at 1-3 GQ/s
Ethernet (computers) are unaffected by distance but have been bumped up in speed 2->3 GQ/s

## BoxStation distances:

Center Evac is ~90 tiles away (1.2x)

R&D desk is ~80 tiles away (1.4x)

Engineering Foyer is ~50 tiles away (2x)

Brig Central is ~40 tiles away (2.2x)

Medical Treatment is ~40 tiles away (2.2x)

Cargo Office is ~30 tiles away (2.4x)

Center of Bridge is ~4 tiles away (2.92x)

Net Admin chair is ~12 tiles away (2.76x)

##

Also fixed downloading more than what you are supposed to (completion % going over 100%)

# Changelog

:cl:   
tweak: NTOS Download speed now increases the closer you are to a NTNet Relay
bugfix: fixed downloading more GQ than the file had
/:cl:

### DIFF
--- a/code/__DEFINES/machines.dm
+++ b/code/__DEFINES/machines.dm
@@ -58,7 +58,7 @@
 //NTNet transfer speeds, used when downloading/uploading a file/program.
 #define NTNETSPEED_LOWSIGNAL 0.5	// GQ/s transfer speed when the device is wirelessly connected and on Low signal
 #define NTNETSPEED_HIGHSIGNAL 1	// GQ/s transfer speed when the device is wirelessly connected and on High signal
-#define NTNETSPEED_ETHERNET 2		// GQ/s transfer speed when the device is using wired connection
+#define NTNETSPEED_ETHERNET 3		// GQ/s transfer speed when the device is using wired connection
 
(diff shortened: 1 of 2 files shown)


Bibby0110:
why is ethernet effected by distance? distance from relay only really makes sense when using wireless devices

ynot01:
> why is ethernet effected by distance? distance from relay only really makes sense when using wireless devices

This is a good point, so I have excluded ethernet and bumped it up to 3 GQ/s to match adv card's top speed

Bibby0110:
> > why is ethernet effected by distance? distance from relay only really makes sense when using wireless devices
> 
> This is a good point, so I have excluded ethernet and bumped it up to 3 GQ/s to match adv card's top speed

very based

ShoziOffical:
time to spam nt relay everywhere on station

JamieD1 (approved):

PR #15563, opened 2022-09-01
Synths are now immune to radiation

Mqiib:
Makes synths immune to radiation because they're ROBOTS and they cant actually remove radiation or toxin damage normally.

# Wiki Documentation

Synths are rad immune.

# Changelog

:cl:  
 
tweak: Fake robots are now immune to radiation like real robots

/:cl:


JamieD1 (dismissed):
Radiation fucks with electronics alot more then you think, they would bitflip like mad.

Mqiib:
> Radiation fucks with electronics alot more then you think, they would bitflip like mad.

yes I realize this but _every other robot in the game_ is rad immune so why not synths? Also they explicitly don't process chems normally so actually getting rid of it and treating the resulting toxin damage is a nightmare.

Bibby0110 (approved):
Synths aren't widely used and this makes sense. You'd probably shield electronics on robots meant for space

PR #15566, opened 2022-09-02
Chameleon Mutation Nerfs & Fix

VaelophisNyx:

Chameleon Mutation now costs 20 instead of 25, because:

- it is now clamped to 30 alpha minimum, no more total invisibility (can be changed via a new define)
- maximum alpha for chameleon is 255 again, no more partial transparency when moving
- if you go into crit it sets you back to 255

This should hopefully stop chameleon being incredibly strong for its price, and stops you being nearly impossible to recover if you die while invisible

# Changelog

:cl:  
bugfix: Chameleon Mutation now disables while in crit
tweak: Chameleon Mutation costs 5 less stability
tweak: Chameleon Mutation now has a minimum transparency, so you can't be 100% invisible, just VERY hard to see
tweak: Chameleon Mutation no longer makes you partially transparent always
/:cl:


Bibby0110 (approved):
Seems reasonable

PR #15567, opened 2022-09-02
The syndicate has devised new anti-explosive armor for their hardsuits!

Mqiib:
Buffs the bomb armor of syndicate hardsuits as follows:
Standard blood-red hardsuit: 35 -> 50 (17.5% -> 25% resistance)
Elite hardsuit: 55 -> 90 (27.5% -> 45% resistance)
Shielded hardsuit: 35 -> 80 (17.5% -> 40% resistance)

Syndies tend to play around with lots of explosions (especially nukies) so it'd make sense that the syndicate would try to give them stuff that would help keep them from blowing THEMSELVES up.
Keep in mind bomb armor is only 50% effective against bombs. If anyone complains about KAs, they shouldn't be the go-to weapon against nukies.

# Wiki Documentation

See bomb armor value changes above

# Changelog

:cl:  

tweak: Syndicate hardsuits have had their bomb armor significantly increased.

/:cl:


00ze-cyclone:
Why is the shielded hardsuit getting more bomb armor than standard hardsuit, it’s supposed to be standard hardsuit but with a shield.

Rdrazga:
> Why is the shielded hardsuit getting more bomb armor than standard hardsuit, it’s supposed to be standard hardsuit but with a shield.

Lore wise you could argue the shield absorbs some, gameplay wise it is prob fair since the shielded suit is something like 35tc where the regular suit you get for free (also normal traitors get access to the regular hardsuit)

JamieD1 (approved):

PR #15587, opened 2022-09-03
makes tele armor use do_teleport

ynot01:
it will now respect no teleport flags

:cl:
bugfix: Reactive tele armor now respects teleport flags
tweak: Reactive tele armor now teleports further if you are wearing a BoH
/:cl:


tattax:
this will mess up people with the armor that have a BoH

ynot01:
> this will mess up people with the armor that have a BoH

JamieD1 (approved):

PR #15590, opened 2022-09-03
Adds the legendary throne of Koko-Junal in medbay break room

Xantam-13:
image

:cl:  
rscadd: Adds the legendary throne for Koko, ready for when he joins the round. 
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -33859,16 +33859,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/engine/foyer)
-"jxq" = (
-/obj/structure/chair/office{
-	dir = 8
-	},
-/obj/effect/landmark/start/medical_doctor,
-/obj/structure/disposalpipe/segment{
-	dir = 4
-	},
-/turf/open/floor/plasteel/white,
-/area/medical/storage)
 "jyo" = (
 /obj/machinery/door/airlock/maintenance_hatch,
 /obj/machinery/door/firedoor/border_only{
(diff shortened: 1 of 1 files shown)


GraveHat:
Bloody hilarious but this shouldn't get merged because of direct player reference.

Xantam-13:
> Bloody hilarious but this shouldn't get merged because of direct player reference.

If we can have a moja with all the stuff that goes with it, we can have a wheelchair that is just renamed ;)

Marmio64:
i thought we were trying to avoid in game player references for quite a while now

Xantam-13:
> i thought we were trying to avoid in game player references for quite a while now

Then remove all in game player references? :D 

Can't be hypocritical now can we!

GraveHat:
> > Bloody hilarious but this shouldn't get merged because of direct player reference.
> 
> If we can have a moja with all the stuff that goes with it, we can have a wheelchair that is just renamed ;)

True but the moja stuff is admin only, this is going to be seen by everyone forever until it gets axed by a different pr and even then we have to take it into account when gramps eventually ceases to exist. Unless I'm missing something that's moja related that is normally accessible to the crew.

Xantam-13:
> 

It is accessible to the crew, it is used during events and testing, it is in the very code, therefore it exists just as much as this wheelchair. 

If there's a rule that we don't reference players in the code, which i don't understand anyway as I used to always do it as a sort of... thank you, to loyal players, then you need to remove everyone from the code, not just new ones.

Marmio64:
> > i thought we were trying to avoid in game player references for quite a while now
> 
> Then remove all in game player references? :D
> 
> Can't be hypocritical now can we!

do so then
regardless though its one thing to still have references since no one has removed them yet and to keep adding them.

GraveHat:
> > 
> 
> It is accessible to the crew, it is used during events and testing, it is in the very code, therefore it exists just as much as this wheelchair.
> 
> If there's a rule that we don't reference players in the code, which i don't understand anyway as I used to always do it as a sort of... thank you, to loyal players, then you need to remove everyone from the code, not just new ones.

Fair point but I'm only really concerned about this being _too_ accessible compared to the others as I'm pretty sure they're usually out of the way to get and not a guaranteed spawn in a place as visible as the med break room.

Bibby0110 (changes requested):
No roundstart direct references to players please.
Admin spawned stuff is semi-okay. Find a way to avoid using the actual player name by adding a fluff description or similar.

JamieD1:
Would prefer to add a donator item that kono starts with instead

PR #15594, opened 2022-09-03
Ports Ethereal stat panel charge display and Ethereal Overcharge from Shiptest

GraveHat:
Ports in the overcharge and status display from shiptest

See tgstation/tgstation#48573 for more details, the only differences being that due to how our teslas work this doesn't actually damage the user when they do explode so I reduced its damage down to dealing 7 burn consistently on a human, another difference is that since this is porting over the code from shiptest their overcharge doesn't deal toxin damage to the user if they're overcharged. Also ethereals get a indicator now on there charge level on their stat panel.

# Spriting
image
overcharge1
overcharge2

# Wiki Documentation

Ethereals can overcharge, and they get a indicator on their stat panel about their charge level.

# Changelog

:cl:  
rscadd: Ethereals can now overcharge and shock people alongside moodlets
rscadd: Ethereals now have an charge level indicator on their stat panel
tweak: Replaced ethereal charge defines 
tweak: Rearranged screen_alert.dmi so the order doesn't look horrendus if left unchanged
imageadd: Adds in Overcharge sprites  
/:cl:


cuackles (approved):
good sprite

Bibby0110 (approved):

PR #15609, opened 2022-09-04
Fixes AIs being able to take damage

Bibby0110:
Lol
:cl:  
rscadd: AI unable to take damage now
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/damage_procs.dm
+++ b/code/modules/mob/living/silicon/damage_procs.dm
@@ -51,9 +51,9 @@
 	return FALSE
 
 //Special snowflake AI damage nos
-/mob/living/silicon/ai/adjustFireLoss(amount, updating_health = TRUE, forced = FALSE) //immune to oxygen damage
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/damage_procs.dm
```diff
@@ -49,3 +49,16 @@
 		return ..()
 
 	return FALSE
+
+/mob/living/silicon/adjustFireLoss(amount, updating_health = TRUE, forced = FALSE) //immune to oxygen damage
```
adamsong:
```suggestion
/mob/living/silicon/ai/adjustFireLoss(amount, updating_health = TRUE, forced = FALSE) //immune to oxygen damage
```
Would this not eliminate the need for a type check?

adamsong (changes requested):

code/modules/mob/living/silicon/damage_procs.dm
Bibby0110: true

adamsong (dismissed):
Fix comments

ToasterBiome (approved):

PR #15616, opened 2022-09-04
Triple AI Station Trait

redmoogle:

Adds a 2 weight(dunno how common that is but PDA glitch is 15) trait that turns on triple AI mode

seems to work.

# Wiki Documentation

New station trait

# Changelog

:cl:  
rscadd: Triple AI station trait
/:cl:

### DIFF
--- a/code/controllers/subsystem/job.dm
+++ b/code/controllers/subsystem/job.dm
@@ -289,11 +289,6 @@ SUBSYSTEM_DEF(job)
 	//Setup new player list and get the jobs list
 	JobDebug("Running DO")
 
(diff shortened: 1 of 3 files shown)


Bibby0110:
This will probably be too frequent

redmoogle:
Do you have a value you would want it to be (1, 0.5)?

JamieD1:
This should only be run by admins because it only causes problems with AIs as they are all law layering turds

redmoogle:
> This should only be run by admins because it only causes problems with AIs as they are all law layering turds

sounds like they should be banned then

PR #15636, opened 2022-09-05
Increases round-start non-cargo budgets so departments can pay their workers more than once

VaelophisNyx:

So fucking irritating that this is even a problem. Nmajask's PR was perfect but since bibby didn't want to let paychecks be decoupled from the budget, here we are.

# Changelog

:cl:  
tweak: Non-Cargo departmental budgets start at 2000cr instead of 500cr
/:cl:


GraveHat:
Should probably increase it to 10k because I doubt this would last the entire round.

VaelophisNyx:
cant
you can take money from the budget and spend it at cargo for shit

Bibby0110:
"bibby didn't want to let paychecks be decoupled from the budget" this is what budgets are for.
I have a more elegant solution

PR #15642, opened 2022-09-06
Buffs (most) hardlight bow projectiles

Skrem7:

Makes hardlight burn arrow do 32 damage up from 25 (12 more damage than a laser, 14 less damage than a slug)
Makes hardlight disabler arrow do 48 damage up from 40 (23 more than a disabler, 7 less than a beanbag)
Makes hardlight x-ray arrow do 21 damage and 400 irradiate up from 15 and 300 (previous values were equal to an x-ray beam)

Redlight bolts remain unaffected because Rat'var didn't buy a warranty

# Wiki Documentation

Values should 🅱️ changed on Guide to Combat under Energy section

# Changelog

:cl:  
tweak: All hardlight arrows have been made tremendously more scary (Rat'var forgot to invest in that good arrow juice, so his servants' arrows remain unchanged)
tweak: Syndicate hardlight bow now has an addendum that it can be CTRL+CLICK'd to fold.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/bow.dm
+++ b/code/modules/projectiles/guns/ballistic/bow.dm
@@ -261,7 +261,7 @@
 
 /obj/item/gun/ballistic/bow/energy/syndicate
 	name = "Syndicate Hardlight Bow"
-	desc = "A modern bow that can fabricate hardlight arrows using an internal energy. This one is designed for silent takedowns of targets by the syndicate."
+	desc = "A modern bow that can fabricate hardlight arrows using an internal energy. This one is designed by the Syndicate for silent takedowns of targets."
 	icon_state = "bow_syndicate"
 	item_state = "bow_syndicate"
 	mag_type = /obj/item/ammo_box/magazine/internal/bow/energy/syndicate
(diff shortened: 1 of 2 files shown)


00ze-cyclone:
Maybe make the xray arrow deal toxin damage instead of burn, that make it harder to heal and slightly stealthier since toxin damage can’t be noticed at a glance unlike burns. Also fit with the theme of the arrow and make it more distinct of the standard killing arrow.

ghost:
Wow, a gun that not only can shoot through walls, but deals more damage then a normall laser gun. Balance.

Skrem7:
> Wow, a gun that not only can shoot through walls, but deals more damage then a normall laser gun. Balance.

Normal x ray does 15 damage plus the irradiate relative to the 20 damage laser gun.

I'm not sure you've ever used a hardlight bow before because I would invite you to see just how quickly you can throw out projectiles relative to the magdump of a standard laser gun.

Rdrazga:
> Wow, a gun that not only can shoot through walls, but deals more damage then a normall laser gun. Balance.

If you have ever used the hard light bow you would understand why it needs to do more damage, it has around a 2-3 second reload cycle per shot depending on how happy the game is with you and lag at the moment. Any normal energy gun even the X-ray gun will out dps it even with this buff. Also it’s 12 tc.

Marmio64:
rad damage should be 400 at max, not 500

JamieD1:
@Mqiib

Mqiib:
> @Mqiib

seems good to me

Bibby0110 (approved):

PR #15660, opened 2022-09-07
Ripping embedded objects out now hurt

redmoogle:

It could only do 1 damage when ripped out
Fixes that

(the damage is already clamped to 0)

# Changelog

:cl:    
bugfix: Ripping out embedded objects now hurt
/:cl:


Xoxeyos:
Could you also add a mood debuff for embedded objects?

redmoogle:
> Could you also add a mood debuff for embedded objects?

There already is from the looks

Bibby0110 (approved):

PR #15663, opened 2022-09-07
[Gax] [Meta] [Box] [Asteroid] Places One IPC Revival Board in Robotics

VaelophisNyx:

IPCs literally are cucked out of being revived until R&D researches a tech. One that they will actively avoid unless an R&D player is an IPC, or a medical doctor harasses them to get an IPC revived. This mitigates this by just placing a single board in robotics in all in-rotation maps.

# Changelog

:cl:  
tweak: GAX Robotics now has a single IPC Revival Board in the Robotech Vendor
tweak: META Robotics now has a single IPC Revival Board
tweak: BOX Robotics now has a single IPC Revival Board
tweak: ASTEROID Robotics now has a single IPC Revival Board
/:cl:


GraveHat:
This would help newer players know the difference between the Cyborg Rebooter and the IPC Revival Board as they would see the correct type of board on the table before they come to the conclusion that the only reboot module they can access which being is the cyborg rebooter is capable of reviving IPCs. As, of this moment, IPC revival is locked behind research that most people would most likely overlook if they don't know about it beforehand, I could see this alleviating that problem.

ToasterBiome:
add it to the vending machine for gax!!!

VaelophisNyx:
but biome how will people know it's there!
Noone opens the damned vendor :C

VaelophisNyx:
done tho

GraveHat:
> but biome how will people know it's there! Noone opens the damned vendor :C

Can confirm, first thought that there were no flashes on gax but it turns out they were all free from the vendor which also subsequently pay locks them for the tiders.

ToasterBiome (changes requested):
add it in the actual vending machine not mapedited!!

PR #15671, opened 2022-09-07
Makes golaith tentacle attack slightly more interesting

redmoogle:

Instead of always making a T

why dont we uhh just give it a 30% chance to attack the 24 tiles around you.

Should spice things up a little

# Wiki Documentation

Attack pattern change

# Changelog

:cl:  
tweak: Goliath tentacle attack pattern changed
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/goliath.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/goliath.dm
@@ -183,12 +183,10 @@
 
 /obj/effect/temp_visual/goliath_tentacle/original/Initialize(mapload, new_spawner)
 	. = ..()
-	var/list/directions = GLOB.cardinals.Copy()
-	for(var/i in 1 to 3)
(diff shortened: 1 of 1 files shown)


VaelophisNyx:
more patterns in general would be neat but holy shit no, why

redmoogle:
> more patterns in general would be neat but holy shit no, why

because the current attack pattern is trivial?

VaelophisNyx:
yes but like
that's *too many* tiles
remember there can be literally dozens of these guys in an area at a time

redmoogle:
> yes but like that's _too many_ tiles remember there can be literally dozens of these guys in an area at a time

you could have just said this in the first place

GraveHat:
So they have a 30% miss/predict chance?

redmoogle:
> So they have a 30% miss/predict chance?

85% chance to not spawn a tentacle on a turf (it looks around in a 2 radius square)

Bibby0110 (approved):

PR #15682, opened 2022-09-08
Adds lungs(radiators) to IPCs

redmoogle:

Gives them cooling radiators for when they dont get enough enough gas they overheat

Uses the heat capacity of their breath
Calibrated for station use

needs about 95kPa oxygen tank to cool down
## Will overheat on lavaland & space without a tank

Overheating icon
image

thought it might be interesting

# Wiki Documentation

They can overheat without air now

# Changelog

:cl:  
tweak: IPCs now need gas to cool down
/:cl:


KillerOrcaCora:
bane-product-road-map-power-over-me

Moltijoe:
Gotta apologize, should have read the PR in detail before reacting as I did, assumed it just added the regular breathing mechanics that all other races use. Didn't realize it was intended to be more of an "air cooling system" that just had you start overheating without an adequate gas. 
Unless i'm mistaken, this means it won't prevent space walking with just a firesuit, because the freezing temperatures of space will keep you cool in place of the gas.  
Trying to think of a situation where an n2 tank will be needed, right after a plasmaflood maybe? that's just all tritium and co2 right? though i suppose in a situation like that, the heat from lack of cooling gasses is the least of your worries

Moltijoe:
damn, so I was right in assuming it was to prevent gasless spacewalks. 
Anyways, you had mentioned allowing a number of different gasses to be used, is there a list somewhere of what will be eligible?(is joining plasmemes in their gas chamber going to still be possible?) 
Will better gases help to mitigate heat from external heat in addition to preventing the "overheat" from internal heat? 
Losing the head will prevent an IPC from wearing a gas mask, and by extent, using internals, will any other problems be caused by losing their head?

redmoogle:
> damn, so I was right in assuming it was to prevent gasless spacewalks. Anyways, you had mentioned allowing a number of different gasses to be used, is there a list somewhere of what will be eligible?(is joining plasmemes in their gas chamber going to still be possible?) Will better gases help to mitigate heat from external heat in addition to preventing the "overheat" from internal heat? Losing the head will prevent an IPC from wearing a gas mask, and by extent, using internals, will any other problems be caused by losing their head?

it uses the heat capacity of gas. It can also use every gas (just as long as you have enough heat reduction) as for loosing the head and internals I'll have to check but I can likely just make an implant that goes in the chest

Moltijoe:
> it uses the heat capacity of gas. It can also use every gas (just as long as you have enough heat reduction) as for loosing the head and internals I'll have to check but I can likely just make an implant that goes in the chest

An implant works, whether it's innate or not doesn't really matter i guess, I assume with how jank the organ code is, letting someone install a breathing tube in the chest rather than the mouth isn't an option.

Can't figure out the calculations enough to find the breakpoint, but i'd guess it's intended to be a heat capacity of around 15-20 at room temperature. It seems like you have the temperature of the gas factored in too, meaning supercooled plasma is the best gas that isn't annoying to make. though it is a shame that it looks like a tank of that wouldn't help with staying cool in a post plasmaflood station. Which would also mean, if you are without breath for a while and heat up a lot, just getting more gas isn't enough to cool you down, you'll need to either dip into space or go under a cold shower for a bit. Ideally you would find the line where it would help you stay cooler without actually making you immune to hot temperatures with a proper gas mix.

This is definitely a lot cooler than just "you need to breath now" and being able to do CPR is a fairly decent plus for those who care.

cuackles (approved):
good icon

redmoogle:
fuck im conflicted with this PR. I dont like the implementation but the idea is neet

Moltijoe:
> fuck im conflicted with this PR. I dont like the implementation but the idea is neet

It's not exactly a simple thing to implement
only way i'd consider changing it is by removing the whole heat_generation variable and just having the temperature constantly vary based on the gas used and just not care about the failed_last_breath check. Relying instead on the already existing temperature pop-up, with perhaps a little clause added to mention that as an IPC you need gasses to temperature regulate.

redmoogle:
> > fuck im conflicted with this PR. I dont like the implementation but the idea is neet
> 
> It's not exactly a simple thing to implement
> only way i'd consider changing it is by removing the whole heat_generation variable and just having the temperature constantly vary based on the gas used and just not care about the failed_last_breath check. Relying instead on the already existing temperature pop-up, with perhaps a little clause added to mention that as an IPC you need gasses to temperature regulate.

Failed last breath just increases tickrate

Moltijoe:
> Failed last breath just increases tickrate

oh, didn't even know it did anything, assumed it was just a boolean for the sake of tracking stuff

JamieD1 (approved):

JamieD1:
Cornflakes

PR #15684, opened 2022-09-09
nerfs gamer chem #1 to remind powergaming chemists that their actions have consequences

TheRyeGuyWhoWillNowDie:
basically, cryoxadone now only heals you while you are asleep. that's it. for cryo tubes, there is no change in functionality because cryo tubes put you to sleep when you're inside them. a special hugbox form of cryoxadone has been created to prevent xenobiologists from having their healing abilities actually just kill them because slime people getting injected with frost oil and the cryoxadone that's supposed to heal them not doing that is a bad thing.

Tested, seems to work fine

# Changelog

:cl:  
tweak: cryoxadone (&  pyroxadone) excluding xenobio cryoxadone now only heals you if you're not conscious
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -176,7 +176,7 @@
 	taste_description = "spicy jelly"
 
 /datum/reagent/medicine/pyroxadone/on_mob_life(mob/living/carbon/M)
-	if(M.bodytemperature > BODYTEMP_HEAT_DAMAGE_LIMIT)
+	if(M.bodytemperature > BODYTEMP_HEAT_DAMAGE_LIMIT && M.IsUnconscious())
 		var/power = 0
 		switch(M.bodytemperature)
 			if(BODYTEMP_HEAT_DAMAGE_LIMIT to 400)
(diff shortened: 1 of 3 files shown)


GraveHat:
Not gonna lie I thought it already did this until I remembered you can chug cryox in space as a skeleton to heal.

RG4ORDR:
Just saying it isn't only cryoxadone.
Some other chems have absolutely no downside to their usage and often facilitate the usage of chemistry magic

CMOisLing:
Dont agree with this nerf feels lazy, just make it so having an extremely low body temperature makes you sleepy or slower, so chemist have to include something to keep themselves awake if they wanna use a mega heal in combat. Also xenobio does not need hugboxing.

boodaliboo:
this is dumb pyroxadone becomes instantly useless now, i think its perfectly reasonable for cryoxodone to work that way but you already need cross department collaboration for it

RG4ORDR:
A better way of nerfing the chem w/o removing the ability to do ghetto cryo with cold showers,space,and other methods is just to give it an OD limit. 
Taking a few sips of cryoxadone and stepping into a cold shower or eating some ice is a smart play and rewards creativity.
However drinking 500 units of it so you can walk around in space and take no damage is a dumb idea. Give an OD limit, and reduce the non-conscious healing(if that's possible). That why it can be utilized by intelligent players but it isn't so good that you can be effectively immune to space.

Bibby0110:
See some of the above suggestions, they're all pretty decent. However at this time this seems to be a bit of a lazy knee-jerk PR

PR #15695, opened 2022-09-09
Replaces missing skirt pixels

Mqiib:
Skirts that shouldn't have a pixel removed are having a pixel removed because the female_top (and skinny_top) masks weren't transferred to the clothing effect dmi. This fixes that small oops.

# Changelog

:cl:  

bugfix: Fixes female_top and skinny_top clothing masks not existing

/:cl:


cuackles (approved):
aproved

PR #15701, opened 2022-09-10
Fixes TGUI input buttons not working with enter key

redmoogle:

You can now hit enter to send NTOS PDA messages

(You were supposed to beable to do this but the button was bugged)

# Changelog

:cl:  
bugfix: You can hit enter to send NTOS PDA messages now
/:cl:


redmoogle:
ugh Ill have to figure out why this fuckery is happening because it shouldt be

PR #15709, opened 2022-09-11
Fixes species punch damage types

Mqiib:
If you had burn punches you _took_ burn damage from punches rather than dealing burn damage. Kind of broken and I have no idea why it was reversed but this fixes it.

# Changelog

:cl:  

bugfix: Fixes species punch damage causing you to TAKE that kind of damage from punches rather than deal it

/:cl:


ToasterBiome (commented):
does this handle mobs that don't have dna or species

oh this is human only

does it handle dnaness things

Mqiib:
> does this handle mobs that don't have dna or species
> 
> oh this is human only
> 
> does it handle dnaness things

Species is stored in dna so im pretty sure if it's got a species it at least technically has dna. I think the no DNA trait is just for mutations/genetics stuff

JamieD1 (approved):

PR #15730, opened 2022-09-13
Fixes a few runtimes

SapphicOverload:

There are a few runtimes that always show up when I run a private server. This PR fixes some of them.

# Changelog

:cl:  
bugfix: fixes some runtimes in _component.dm, mood.dm, and inventory.dm
/:cl:

### DIFF
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -298,6 +298,8 @@
 //for when you want the item to end up on the ground
 //will force move the item to the ground and call the turf's Entered
 /mob/proc/dropItemToGround(obj/item/I, force = FALSE, silent = FALSE)
+	if(!I)
+		return
 	SEND_SIGNAL(I, COMSIG_ITEM_PREDROPPED, src)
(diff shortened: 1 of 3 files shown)


code/modules/mob/inventory.dm
```diff
@@ -298,6 +298,8 @@
 //for when you want the item to end up on the ground
 //will force move the item to the ground and call the turf's Entered
 /mob/proc/dropItemToGround(obj/item/I, force = FALSE, silent = FALSE)
+	if(!I)
```
redmoogle: #15727

ToasterBiome:
redmoogle's solution to the dismemberment is better because we should be figuring out why we're putting null into things instead of fixing behavior when things are null

revert that line and the rest looks good, i wonder what could be causing the mood one

JamieD1:
Can you make sure this works with moogles

redmoogle:
> Can you make sure this works with moogles

Jamie they undid that part of the change

Bibby0110 (approved):

PR #15733, opened 2022-09-13
Gives urinal cakes an icon

Moltijoe:
Babies first fix

makes urinal cakes use an existing image so they aren't just invisible. also for their squished version (why is that even a function?)

:cl:  
bugfix: points urinal cakes to use an image
/:cl:


cuackles:
Really like the sprites but if I'm correct the urinal cakes had sprites before they are just not appearing, when i get home i will take a look at it better tho

VaelophisNyx:
they do indeed have sprites still, just gotta find the file they're in

Moltijoe:
> they do indeed have sprites still, just gotta find the file they're in

found them, they're in misc instead of watercloset, should i move them?

image

someone definitely put more work into them than i did mine

KillerOrcaCora (approved):

PR #15756, opened 2022-09-14
Disables Drifting Space Pod ""Ruin""

VaelophisNyx:

So these pods spawn a lot, and have enough initial speed to explode on contact with anything
space is also full of walls. So the most common occurance for one of these drifting pods is to spawn, then drift at high speed into any static space ruin, and space it. They also eat up the space ruin spawn budget/time.

just commented them out for now.

# Changelog

:cl:  
rscdel: Removes Drifting Space Pods from space, so they will stop exploding on ruin walls
/:cl:

### DIFF
--- a/code/datums/ruins/space.dm
+++ b/code/datums/ruins/space.dm
@@ -329,12 +329,13 @@
 	always_place = TRUE // This is just the space part, king_goat_boss in /code/datums/ruins/lavaland.dm needs to have this set to true aswell for goat king to actually be reachable
 	allow_duplicates = FALSE
 
+/*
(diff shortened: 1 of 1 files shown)


Bibby0110:
Preferably just added to the blacklist

VaelophisNyx:
the blacklists dont work

Bibby0110:
> the blacklists dont work

Fix the blacklists and add this to the blacklist

Hopekz:
> Fix the blacklists and add this to the blacklist

A.k.a probably don't do anything because that's a lot more work than you set out to do 😎

PR #15758, opened 2022-09-14
Legion cores are now less cool off Lavaland (but there's a new kind of stabilizer)

Skrem7:

Targets legion core usage on station specifically to discourage miners from exploiting their ability to ignore game mechanics (damage slowdown).

Legion cores on station not only heal less (as they're an instant-use) but now ONLY heal instead of providing all the utility they have.

For those miners who simply MUST have their gaming juice, they can buy super-stabilizer serum that costs 650 instead of 400. A super-stabilized core loses less healing while retaining a CC and temperature reset effect (assuming one is able to use it; the former essentially assumes the target is not the user). However, it will still not prevent slowdown.

Implanted legion cores are unaffected; they're single-use and require surgery. They're also the only way to clear some deep traumas, which is pretty swag, not gonna lie.

# Spriting
I don't know none of the other bottles had a brown liquid:
image

# Wiki Documentation

New serum to be added to Shaft Miner page on the vending machine part
Legion core as a drop of loot might need to be updated on the same page

# Changelog

:cl:  
tweak: Whenever a legion core is used off Lavaland, it now heals 10 less brute and 10 less burn, no longer clears CC and resets temperature, and does not render the user immune to slowdown for a minute
tweak: A super-stabilized legion core instead heals 5 less brute and 5 less burn but will still clear CC and reset temperature. Still does not prevent slowdown
rscadd: Adds super-stabilizer serum for 650 at mining vendors
imageadd: New icon for the new serum
/:cl:


VaelophisNyx:
my only complaint here is snowflake bottle sprites are bad and should die, and it should just be a brown liquid in a normal, renamed, bottle.

Skrem7:
> my only complaint here is snowflake bottle sprites are bad and should die, and it should just be a brown liquid in a normal, renamed, bottle.

Doesn't actually have any u so using a normal bottle would be problematic as it's actually its own item, unless stabilizer serum were made a reagent that would then require application limitations so you can't just put 0.1u on every core which would also result in more random item clutter in mining from bottles left behind.

GraveHat:
Unnecessary but Implanted cores don't actually give you the adrenalin buff.

ToasterBiome:
very cool kanye

Cartlord:
Miners get enough points in a decent shift that all this'll do is make everyone buy super stabilizers

Skrem7:
> Miners get enough points in a decent shift that all this'll do is make everyone buy super stabilizers

OH NO!

The legion cores NO LONGER remove slowdown on station regardless. How TERRIBLE.

Skrem7:
New PR dropped that's probably a better alternative that bloats less items.

PR #15762, opened 2022-09-15
Letting go of drip without wearing it no longer makes you sad

GraveHat:

Copies a check from the krav maga gloves so simply dropping the drip without wearing it doesn't destroy your self esteem.

# Changelog

:cl:    
bugfix: Drip requires you to feel their power before giving you clinical depression
/:cl:

### DIFF
--- a/code/modules/clothing/shoes/miscellaneous.dm
+++ b/code/modules/clothing/shoes/miscellaneous.dm
@@ -631,6 +631,10 @@
 
 /obj/item/clothing/shoes/drip/dropped(mob/user)
 	. = ..()
-	SEND_SIGNAL(user, COMSIG_CLEAR_MOOD_EVENT, "dripjordan")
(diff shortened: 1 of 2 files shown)


code/modules/clothing/shoes/miscellaneous.dm
```diff
@@ -631,6 +631,10 @@
 
 /obj/item/clothing/shoes/drip/dropped(mob/user)
 	. = ..()
```
redmoogle, line 633: you should move the . = ..() call as it moves the item out

redmoogle: not sure if it will cause an issue tho

GraveHat: method to fix the issue was copied directly from krav maga gloves, I don't think it causes issues.

Bibby0110 (approved):

PR #15776, opened 2022-09-15
Restores DoS app to syndinet

Addust:
Looked like a fuckup, will close if removal was intended or rebalance it.

Restores syndinet access to the DoS app, as it seemed that the non-inclusion was unintentional.

# Spriting
bro im changing ONE VARIABLE

# Changelog

:cl:  
bugfix: The Syndicate have identified an error with their software distribution network, and restored access to the Denial of Service program.
/:cl:


Bibby0110 (approved):
you can buy this for TC but an emag is more expensive so i dont see why not

PR #15781, opened 2022-09-16
[Hugbox] Comms Consoles AutoLogout After 3 Minutes

VaelophisNyx:

Comms consoles automatically log out 3 minutes after logging in. This does not care if you're using the console, but who's going to be using the console for that long??
and just like....
log back in.

# Changelog

:cl:  
tweak: Comms Consoles automatically log out after 3 minutes pass.
/:cl:

### DIFF
--- a/code/game/machinery/computer/communications.dm
+++ b/code/game/machinery/computer/communications.dm
@@ -51,8 +51,13 @@
 	. = ..()
 	GLOB.shuttle_caller_list += src
 
-/obj/machinery/computer/communications/proc/autologout
-		action = "toggleAuthentication" // We actually want to log out
+/obj/machinery/computer/communications/proc/autologout(action)
+	if(authenticated)
+		authenticated = FALSE
+		authorize_access = null
+		authorize_name = null
+		playsound(src, 'sound/machines/terminal_off.ogg', 50, FALSE)
+		return
 
 /// Are we NOT a silicon, AND we're logged in as the captain?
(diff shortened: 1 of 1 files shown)


00ze-cyclone:
Why does the console need to logout by itself?

VaelophisNyx:
why does your bank website need to log out by itself irl

it's important

00ze-cyclone:
The station got plasma that is separated from distro by just one pump, cheap insuls that can shock you more, an engine that obliterate the station if left unattended, a room full of gun separated from space by a single wall, an AI with absurd laws, a fucking nuke just sitting there and bags that destroy reality if handled improperly.
Problem is the retard captain, not the code.

VaelophisNyx:
please explain to me what this would break if I add it

because all I'm hearing is "because its always been that way"

Moggram:
Sec consoles don't auto log out to. Will you add a timer to them also?

VaelophisNyx:
I mean if that became an issue I could
but it's not as prevalent as captains not logging out
or as is usually the case
some other head, not logging out

Maia-J:
> why does your bank website need to log out by itself irl

It checks for inactivity however

VaelophisNyx:
yes but this unfortunately is a game running on miles of shitcode that doesn't have a way to make a timer refresh instead of letting it tick down

Bibby0110:
Too hugboxy, sorry

PR #15794, opened 2022-09-17
Gives the ability for  Xenomorphs to cool down, Embryos' die with the host, Beds bucklecuff victims, and bed escape time reduced to 55 seconds

GraveHat:
When has there ever been a xenomorph chest burst from a dead body in any of the source material?

Adds a check if the host is dead that qdels the embryo, spawns a dead one in its place alongside adding in some flavor text and brute damage. Adds new cuffs under the alien subtype that are as weak as fake cuffs but are there to stop buckled individuals from using their hands that destroy themselves upon removal that uses the sprites of the unused abductor alien handcuffs. Gives the ability for xenomorphs to cooldown in space and other cold areas. The escape time has also been reduced to encourage the xenomorphs to be more attentive as you are no longer a threat because of the cuffs.

Why? Current Xeno gameplay is unfun for anyone captured being executed or dying in crit shortly after being impregnated. This pr was made to address that and other issues that may arise from simply forcing xenomorphs from spacing the room and getting free spawns from corpses to, actual xenomorph behavior of taking in live captures to convert.

# Wiki Documentation

Xenomorph embryos die alongside the host. Resin nest buckle cuff people and lets Xenomorphs can cooldown in cold areas.

# Changelog

:cl:  
rscadd: Adds resin cuffs and makes the alien nest auto bucklecuff hosts
tweak: Xenomorph embryos now die shortly after the host
tweak: Xenomorphs can cooldown now
/:cl:

### DIFF
--- a/code/game/objects/items/handcuffs.dm
+++ b/code/game/objects/items/handcuffs.dm
@@ -158,6 +158,18 @@
 /obj/item/restraints/handcuffs/alien
 	icon_state = "handcuffAlien"
 
+/obj/item/restraints/handcuffs/alien/beno
+	name = "resin handcuffs"
+	desc = "Extremely weak cuffs made to restrict movement."
+	icon_state = "cuff" // Needs sprite
+	breakouttime = 10 //Deciseconds = 1s
+	break_strength = 1
+	item_flags = DROPDEL
+
+/obj/item/restraints/handcuffs/alien/beno/dropped(mob/user)	
+	user.visible_message(span_danger("[user]'s [name] falls apart"))
+	. = ..()
+
 /obj/item/restraints/handcuffs/fake
 	name = "fake handcuffs"
 	desc = "Fake handcuffs meant for gag purposes."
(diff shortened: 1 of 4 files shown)


Marmio64:
its far too difficult to baby them especially when you need space to be immediately available to not die from the smallest bit of fire and overheating, and all the hosts want to do is kill themselves.

Mqiib:
If the embryo dies after the host does then make the xeno nest make it so that the people nested can't use their hands to do anything, because right now you can just grab a welder out of your belt and destroy the nest faster than it takes to break out, along with attacking any nearby xenos. Only way to stop this is to crit/kill people or spend a minute fullstripping and dragging away EVERY ITEM within reach. This is why xenos kill people.

RG4ORDR:
Doing this you need to make it so Xenos bleed heat if they get set on fire, one the biggest reasons why nests are often in spaced areas is to kill people and to prevent fires. One inced shotgun or god forbid a plasma flood means you will die if you're in or near the heat, there's no way to remove the heat either.

Sniblet:
That very cool engaging CM gameplay feature of lying down in pitch darkness for five minutes while a tearfully bored sentinel occasionally throws up on you (added for lore reasons)

Bibby0110:
Cucks Xenos too hard, we're not CM. In addition the cooldown code doesn't do what you think it does.

PR #15808, opened 2022-09-17
changes poppys chemicals from libital to bicardine to promote botanists actually using them

boodaliboo:
changes poppys chemicals from libital to bicardine to promote botanists actually using them
:cl:  

tweak: tchanges poppys chemicals from libital to bicardine to promote botanists actually using them

/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/flowers.dm
+++ b/code/modules/hydroponics/grown/flowers.dm
@@ -15,7 +15,7 @@
 	icon_grow = "poppy-grow"
 	icon_dead = "poppy-dead"
 	mutatelist = list(/obj/item/seeds/poppy/geranium, /obj/item/seeds/poppy/lily)
-	reagents_add = list(/datum/reagent/medicine/c2/libital = 0.2, /datum/reagent/consumable/nutriment = 0.05)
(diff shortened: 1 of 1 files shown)


RG4ORDR:
Hell yes opiate crisis in SS13

VaelophisNyx:
mmmm, this feels....unbalanced

what if we did the barotrauma thing and made it morphine

boodaliboo:
> mmmm, this feels....unbalanced
> 
> what if we did the barotrauma thing and made it morphine

i would honestly rather have it give like codeine that could resist damage slowdown slightly, ill pay someone 3$ to make that code

00ze-cyclone:
Libital heal 3 brute per tick while bicaridine is only 2, it also metabolize slower than bicaridine making it overall more effective. Botanists can get silibilin to cure the liver damage, negating the one relatively small downside of libital. That PR is making poppy less valuable instead of promoting its production.

Bibby0110 (changes requested):
Respond to the above comment

boodaliboo:
> Respond to the above comment

what about it i read it and still want bicardine

Bibby0110:
Weak/non-existent reasoning.

PR #15810, opened 2022-09-17
removes the gibbing from vending machine without removing all the other thingies

boodaliboo:
removes the gibbing from vending machine without removing all the other vending machine crushing related things
:cl:  
rscadd: removes the gibbing from vending machine without removing all the other thingies
/:cl:


JamieD1:
Dont hit the vending machine

JamieD1 (approved):

PR #15827, opened 2022-09-18
Massively Reduces Granite Spawn Rate And Ore Chance

VaelophisNyx:
ss (2022-09-17 at 11 44 55)

Reduces granite spawn chance by half, but this in turn comes out to a lot less than just half because of how cellular automa works.

also reduces the ore chances a bit

# Changelog

:cl:  
tweak: Granite now CONSIDERABLY less common
tweak: Granite now has a lower chance to yield ores
/:cl:

### DIFF
--- a/code/datums/mapgen/Cavegens/LavalandGenerator.dm
+++ b/code/datums/mapgen/Cavegens/LavalandGenerator.dm
@@ -24,7 +24,7 @@
 	var/basalt_death_limit = 3
 	var/basalt_turf = /turf/closed/mineral/random/volcanic/hard
 
-	var/initial_granite_chance = 35
+	var/initial_granite_chance = 20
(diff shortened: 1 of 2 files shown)


Maia-J:
I would much more prefer if it just lowered the spawn rate, reducing the ore chance kindoff defeats the point it was added for.

VaelophisNyx:
you still get more ore per granite than basalt so there's that, and it IS more common

but like, uranium was TWICE as common PLUS the extra ore, as an example

Maia-J:
Well I feel like nerfing both values kinda defeats the purpose. I do think it needs a nerf, but not to both its spawn rate and ore amount, but just to one of the values.

Bibby0110 (approved):

PR #15831, opened 2022-09-18
re-adds reactive pyro armor

TheRyeGuyWhoWillNowDie:
the reason this was removed in the first place, four years ago, is really shitty.

reactive pyro armor ignites people nearby when it reacts. this is quite powerful, but so are the other anomaly armors (most notably, gravitational anomaly armor fucking stunning people nearby with wallthrows) if you use an anomaly core to make reactive armor, which very few people do, you deserve a powerful item.

# Changelog

:cl:  
tweak: using a pyro core on a reactive armor shell once again results in reactive pyro armor
/:cl:

### DIFF
--- a/code/modules/clothing/suits/reactive_armour.dm
+++ b/code/modules/clothing/suits/reactive_armour.dm
@@ -10,6 +10,7 @@
 	var/static/list/anomaly_armor_types = list(
 		/obj/effect/anomaly/grav	                = /obj/item/clothing/suit/armor/reactive/repulse,
 		/obj/effect/anomaly/flux 	           		= /obj/item/clothing/suit/armor/reactive/tesla,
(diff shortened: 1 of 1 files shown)


boodaliboo:
wait this was removed? i remeber colton teaching me about these

VaelophisNyx:
on one hand I Agree because well, you put in the effort...
on the other hand I'm worried science is gonna make these all the time because they can

TheRyeGuyWhoWillNowDie:
> on one hand I Agree because well, you put in the effort... on the other hand I'm worried science is gonna make these all the time because they can

i am not going to lie to you, i have never seen anyone except for me make reactive armor in science. its ALWAYS phazons, if they do even bother with the anomaly core, which requires a random event to trigger

Bibby0110 (approved):

PR #15846, opened 2022-09-19
Embedded Syringes Inject Full Transfer Amt. Per Tick

VaelophisNyx:

Should make syringe guns loaded with things that aren't bluespace syringes not wet-noodles without explosive mixes. The main problem with them right now is they inject 5u on contact...then...1u. Per tick. Painfully weak for what they're supposed to be.

"its too OP!" then fucking remove syringe guns
what do you want them to be, because useless means that the answer is Bloat.

# Changelog

:cl:  
tweak: Embedded syringes now inject their full normal transfer amount per tick instead of 20% of that.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/syringes.dm
+++ b/code/modules/reagents/reagent_containers/syringes.dm
@@ -200,7 +200,7 @@
 	return TRUE
 	
 /obj/item/reagent_containers/syringe/embed_tick(embedde, part)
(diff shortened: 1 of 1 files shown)


Rdrazga:
This needs to be done after bomb mixes are dealt with since this allows things like black powder bomb mixes to max really fast before they go off

Bibby0110 (approved):

PR #15847, opened 2022-09-19
Reworks the poor aim quirk and all sources of poor aim to be able to stack

Moltijoe:
As the title says, before they all just piggy backed off applying and removing the poor aim quirk
Now there's a species variable that can be increased and decreased,
this allows for:
Species with innate poor aim
Variable amounts of poor aim (rather than just 25 from all sources)
Stacking poor aim

This is an indirect buff to flying fang which applied a poor aim debuff as this allows it to affect people with the poor aim quirk
It could also allow for future adjustment to the strength of the flying fang poor aim

:cl:  
bugfix: sources of spread increase stack
/:cl:


code/modules/clothing/head/helmet.dm
```diff
@@ -451,12 +451,18 @@
 	strip_delay = 60
 	flags_cover = HEADCOVERSEYES
 
-/obj/item/clothing/head/helmet/stormtrooper/equipped(mob/living/user)
-	ADD_TRAIT(user, TRAIT_POOR_AIM, CLOTHING_TRAIT)
+/obj/item/clothing/head/helmet/stormtrooper/equipped(mob/living/user, S)
+	if(ishuman(user) && S == SLOT_HANDS)
```
Bibby0110: helmet in your hands?

code/modules/clothing/head/helmet.dm
```diff
@@ -451,12 +451,18 @@
...
+		var/mob/living/carbon/human/H = user
+		H.dna.species.aiminginaccuracy += 25
+		to_chat(world, H.dna.species.aiminginaccuracy)
```
Bibby0110: ...

Bibby0110 (changes requested):
..

code/modules/clothing/head/helmet.dm
Moltijoe: oopsie

JamieD1 (approved):

PR #15871, opened 2022-09-20
Preterni/IPCs/Golems no longer valid for augmentation

VaelophisNyx:

These three species are all already inorganic, letting them get augmented just let them wiggle out of their species related downsides easier.

Disables mechanical augment surgery as it is useless with this.

# Wiki Documentation

Guide to Surgery might need mechanical augmentation removed.

# Changelog

:cl:  
rscdel: You cannot augment Preterni/IPCs/Golems
/:cl:


Sniblet:
Letting humans get augmented just let them wiggle out of their species related downsides easier

VaelophisNyx:
augmenting exists for organic beings.
No more augmented metal hydrogen golems just being a near unkillable tank

Sniblet:
So like... why not disable it only for golems
All of them but golems can just have their limbs detached and replaced anyway so this only adds an extra step

Bibby0110 (approved):

PR #15885, opened 2022-09-22
Removes the "undershirt" from the green EMT jacket

Moltijoe:
Changes the casually worn green emt jacket to have three pixels width of transparency in the middle.

It was pretty much impossible to mix and match this with any different jumpsuit as it would still cover it up with what looked like a white undershirt.
This will make it function sorta like the sci labcoat, where you can swap your jumpsuit and still see it through your jacket
The blue jacket is untouched if you prefer the undershirt look
### Before
Before
### After
image

:cl:  
imageadd: "removed" the old icon "added" the new one 
imagedel: why is there no imagetweak tag?
/:cl:


SomeguyManperson:
revert changes to icons/obj/clothing/suits.dmi that's the icon of the item on the floor. Turns out people are lazy and just copy+paste the front icon of suits into the mob icon and that undershirt isn't meant to be an undershirt it's meant to be the back of the suit

Moltijoe:
> 

ah, that actually makes sense, in that case i have a bit more tweaking to do

KillerOrcaCora:
*sigh

Moltijoe:
> *sigh

you sigh, but don't explain what's wrong

Moggram:
> > *sigh
> 
> you sigh, but don't explain what's wrong

Maybe that the side view still has white going up to the top? Its all I can think of.

Moltijoe:
> Maybe that the side view still has white going up to the top? Its all I can think of.

yeah, i thought that, but a single pixel's width isn't enough to actually see what's inside, and making the side green parts only 2 pixel width removes the colour definition which looks kinda bad, it's possible that the "undershirt" look was intentional because of these problems.

JamieD1 (approved):

PR #15887, opened 2022-09-22
Reduces Phytosian Ammonia/Saltpetre/RobHarv Healing

VaelophisNyx:

As a phytosian player, these particular chems are a bit...much, currently. This still lets you use them to heal, BUT, half as effectively as before, because as a quick example, you can chug 500-1000u of Ammonia, and dive unarmored into combat against a traitor with armor and an esword, and have close to a 50/50 of winning that fight, instead of it being heavily traitor-sided like it probably should be.

Saltpetre's burn heal isn't quite halved, but is still significantly lower

# Wiki Documentation

If any mention is made of phytosians healing from ammonia, saltpetre, and robust harvest, it needs to mention it's half as effective as it currently mentions...if it does?

# Changelog

:cl:  
tweak: Phytosians heal half as much from non-addictive botany chems
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/carbon/human/species_types/plantpeople.dm
+++ b/yogstation/code/modules/mob/living/carbon/human/species_types/plantpeople.dm
@@ -159,19 +159,19 @@
 		return 1
 
 	if(chem.type == /datum/reagent/saltpetre)
-		H.adjustFireLoss(-2.5*REAGENTS_EFFECT_MULTIPLIER)
-		H.adjustToxLoss(-1*REAGENTS_EFFECT_MULTIPLIER)
+		H.adjustFireLoss(-1.5*REAGENTS_EFFECT_MULTIPLIER)
+		H.adjustToxLoss(-0.5*REAGENTS_EFFECT_MULTIPLIER)
 		H.reagents.remove_reagent(chem.type, chem.metabolization_rate * REAGENTS_METABOLISM)
 		return 1
 
 	if(chem.type == /datum/reagent/ammonia)
-		H.adjustBruteLoss(-1*REAGENTS_EFFECT_MULTIPLIER)
-		H.adjustFireLoss(-1*REAGENTS_EFFECT_MULTIPLIER)
+		H.adjustBruteLoss(-0.5*REAGENTS_EFFECT_MULTIPLIER)
+		H.adjustFireLoss(-0.5*REAGENTS_EFFECT_MULTIPLIER)
 		H.reagents.remove_reagent(chem.type, chem.metabolization_rate * REAGENTS_METABOLISM)
 		return 1
 
 	if(chem.type == /datum/reagent/plantnutriment/robustharvestnutriment)
-		H.adjustToxLoss(-2*REAGENTS_EFFECT_MULTIPLIER)
+		H.adjustToxLoss(-1*REAGENTS_EFFECT_MULTIPLIER)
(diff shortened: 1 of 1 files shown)


Reed0506:
why not just add addiction/overdose thresholds and solve the problem rather than just making them a bit weaker but still mechanically broken

VaelophisNyx:
because that'd be a lot of really fucky code on the phytosian side, and adding overdoses to chems that only effect one race is...Weird

Reed0506:
could just add them as universal overdose thresholds and it would have practically zero effect on anyone who isnt a pod

also pod code already have its own overdose code for diethylamine and stuff you could just copypaste it

VaelophisNyx:
this is the better solution in the end.

Reed0506:
but the entire problem is based on the insane chem volume not the healing power

VaelophisNyx:
consider: if you're dead it wont process, so as long as it isn't enabling brain-off dives into combat

its fine

Reed0506:
i do not understand what is bad about balancing pod chem healing the same way all other healing chems are balanced

VaelophisNyx:
because it's three chems, that have no use outside of botany, rare reactions, and pod person healing, which would make adding proper OD effects to them incredibly fucky. This is the fast, elegant solution.

boodaliboo:
> i do not understand what is bad about balancing pod chem healing the same way all other healing chems are balanced

no one uses them ever this wont effect anything

Reed0506:
as far as i can tell all you would need to do is copypaste the 7 lines of code from diethylamine OD into the chem processing for saltpetre/ammonia/robustharvest and the chem chugging problem would be solved. i dont think diethylamine even has an OD effect other than just disabling the healing and thats completely fine for all of these.

Reed0506:
one last point, 2.5 brute 5 burn 4 toxloss of healing per life tick to 2 brute 3.5 burn 2.5 toxloss per life tick is quite insignificant in the context of combat, as when taking high damage per hit, only a few life ticks are needed to make up the difference between being crit or not. for example, lets say 30 brute damage or 30 burn damage (stetchkin/esword/weldingtool) per hit. to get past the required 20 damage healed to survive the 4th hit, brute healing would only require 2 extra life ticks (9->11) and burn would only need 1 more tick (5->6), which amounts to about 2-4 extra seconds of wiggle room before a 5th hit would be required.

also, no wiki documentation should be required if this change is made in its current state, as no specific numbers are listed for the healing effects of the chems, only vague promises.

VaelophisNyx:
actually better idea
what if  Imade it give phytosians *nutrition* and thus it'd make them fat (thus slower)

boodaliboo:
> actually better idea what if Imade it give phytosians _nutrition_ and thus it'd make them fat (thus slower)

yes

VaelophisNyx:
though that is a what-if
as this works just fine

SomeguyManperson:
Cant you just chug around 29 or whatever units pf conventional medicine and itll last for long enough in a fight to let you no thoughts head empty with whatever passive healing that provides, though on the other hand these dont particularly conflict with conventional medicines

Cobbychems remedy this slightly by causing organ damage but most of them wont have any negative effects unless taken for over 5 minutes without a break anyway

VaelophisNyx:
yeah just
in the case of these three chems, there is NO downside at all

JamieD1 (approved):

PR #15929, opened 2022-09-26
Lets Drones Understand Common + Gives Drones Internal Radios + Lets them Emote more

VaelophisNyx:

Drones understanding common is a good change for them, allowing you to more readily hear about things that need fixing or upgrading. AI's and borgs *rarely* talk on binary chat about these things as is.
Standard drones get an internal AI headset
syndrones get a syndicate headset
cogscarabs, and derelict drones get no headset
bar-drones get a common headset
polymorph drones get a common headset

Tested, drones CANNOT use this to talk over the radio, ONLY to hear.

also lets drones use all silicon related emotes, because they reasonably should be able to

for those worried about this letting drones interact easier
yeah it does
because almost no admin seems to enforce drone-law's "do not interact **period**" statement, and crew absolutely will point-spam you to do stuff
this just removes the need to point spam as you can now directly ask a drone to do something.
besides
never being allowed to even acknowledge the existence of people around you is immensely dull, and leads to situations where drones refuse to not follow an obvious antag around

# Wiki Documentation

would require small updates to the drone page to mention you can understand common and normally hear all radio channels.

# Changelog
:cl:  
tweak: Drones can understand Common
tweak: Normal Drones can hear all radio channels
tweak: drones can now access all silicon emotes
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/drone/_drone.dm
+++ b/code/modules/mob/living/simple_animal/friendly/drone/_drone.dm
@@ -118,6 +118,12 @@
 		ADD_TRAIT(src, TRAIT_PACIFISM, JOB_TRAIT)
 		ADD_TRAIT(src, TRAIT_NOGUNS, JOB_TRAIT) //love drones t. Altoids <3
 
+/mob/living/simple_animal/drone/Destroy()
+	. = ..()
+	var/atom/T = drop_location()
+	if(T && istype(radio) && istype(radio.keyslot))
+		radio.keyslot.forceMove(T)
+		radio.keyslot = null
 
 /mob/living/simple_animal/drone/med_hud_set_health()
 	var/image/holder = hud_list[DIAG_HUD]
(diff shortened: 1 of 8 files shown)


Cartlord:
Can't dornes already see atmos alarms, and maybe power alarms? That seems like enough to know when stuff needs fixed, since most methods of causing station damage result in hull breaches or power loss.

VaelophisNyx:
not always though, and that's part of the problem
but also drones are supposed to upgrade and such, and if someone wants an upgrade and calls for it, this would let you hear that and understand it.

code/modules/mob/living/simple_animal/friendly/drone/_drone.dm
```diff
@@ -95,6 +96,7 @@
 	access_card = new /obj/item/card/id(src)
 	var/datum/job/captain/C = new /datum/job/captain
 	access_card.access = C.get_access()
+	radio = new /obj/item/radio/headset/silicon/ai(src)
```
Bibby0110, line 99: you'll need to delete the radio when they die

VaelophisNyx: Yeah I kinda forgot I needed to do that
I'll get around to it soon

JamieD1:
No, drones should not be interacting with crew and have no need to. This is a rules issue as well.

PR #15938, opened 2022-09-27
Shadowling Damage From Light Is Now Clone Damage

VaelophisNyx:

Doing this after observing a chemist chug three burn healing chems to automatically heal the damage from lights, so he could walk around in the light for extreme durations for a sling. Not ok.

Also they have free clone healing in darkness so, this isn't a huge nerf. You can still chug chems to heal burns just, not ignore lights so easily. Sling DNA was removed for this very reason with antiglow, letting you 100% ignore lights.

"muh winrate!!!" get out. Irrelevant.

buffs their clone healing in darkness to 5 per tick instead of 1 so it matches their burn healing

# Changelog

:cl:  
tweak: Shadowlings and their empowered thralls take Clone damage instead of Burn damage from light
tweak: Shadowlings heal 5x faster from clone damage, making the prior change not effect their ability to recover from walking in lights
/:cl:

### DIFF
--- a/yogstation/code/game/gamemodes/shadowling/shadowling.dm
+++ b/yogstation/code/game/gamemodes/shadowling/shadowling.dm
@@ -186,7 +186,7 @@ Made by Xhuis
 			H.heal_overall_damage(5,5)
 			H.adjustToxLoss(-5)
 			H.adjustOrganLoss(ORGAN_SLOT_BRAIN, -25) //Shad O. Ling gibbers, "CAN U BE MY THRALL?!!"
-			H.adjustCloneLoss(-1)
+			H.adjustCloneLoss(-5)
 			H.SetKnockdown(0)
 			H.SetStun(0)
 			H.SetParalyzed(0)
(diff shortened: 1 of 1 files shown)


TheRyeGuyWhoWillNowDie:
Skill issue

RG4ORDR:
Why not make Slings just not process chems?

VaelophisNyx:
> Why not make Slings just not process chems?

would cripple them a little too hard

Marmio64:
This is also potentially increasing the time it takes to heal light dmg by 5x, which is a pretty damn large nerf.

VaelophisNyx:
> This is also potentially increasing the time it takes to heal light dmg by 5x, which is a pretty damn large nerf.

Hmm, I'll look into it. I don't think it's that big an issue if it's not the same damage type as lasers and such but...yeah

Sniblet:
Light damage regen is in fact a big deal because slings spend time in light on purpose sometimes
Not as much as darkspawn but too much to slow it down by a factor of 5

Marmio64:
> > This is also potentially increasing the time it takes to heal light dmg by 5x, which is a pretty damn large nerf.
> 
> Hmm, I'll look into it. I don't think it's that big an issue if it's not the same damage type as lasers and such but...yeah

you'll want to raise the amount of clone damage slings heal per tick from 1 to at least 3, but i'd prefer 5 since then its the same regen rate as before you just cant chug burn chems to heal while in the light.

Bibby0110 (approved):

PR #15953, opened 2022-09-29
Preternis chem purge also purges the effects of the chems

Moltijoe:
Didn't call on_mob_end_metabolize() before deleting the chemical from the system
Some chems that would revert the player to normal as the chem ran out would instead leave the player in whatever fucked up state they were in

:cl:  
bugfix: no more permanent rotatium angle
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/preternis.dm
+++ b/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/preternis.dm
@@ -135,6 +135,7 @@ adjust_charge - take a positive or negative value to adjust the charge level
 				to_chat(H,span_info("NOTICE: Digestive subroutines are inefficient. Seek sustenance via power-cell C.O.N.S.U.M.E. technology induction."))
 
 	if(chem.current_cycle >= 20)
+		chem.on_mob_end_metabolize()
 		H.reagents.del_reagent(chem.type)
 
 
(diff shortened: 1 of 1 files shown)


redmoogle:
pretty sure this would also give them perma meth lol

**though** del_reagent deletes the reagent

Moltijoe:
> pretty sure this would also give them perma meth lol
> 
> **though** del_reagent deletes the reagent

how? on_mob_end_metabolize for meth removes the movespeed modifier

redmoogle:
> > pretty sure this would also give them perma meth lol
> > **though** del_reagent deletes the reagent
> 
> how? on_mob_end_metabolize for meth removes the movespeed modifier

probably some broken logic

Maia-J:
Ok it would appear time is running out for me to get every effect on a preternis... Time to do it before merge day

Bibby0110:
> > pretty sure this would also give them perma meth lol
> > **though** del_reagent deletes the reagent
> 
> how? on_mob_end_metabolize for meth removes the movespeed modifier

del_reagent should also call on_mob_end_metabolize() so there's an underlying bug that needs fixing.

Moltijoe:
> del_reagent should also call on_mob_end_metabolize() so there's an underlying bug that needs fixing.

I think it's because del_reagent isn't being called during the metabolize function, it only runs on_mob_end_metabolize if metabolizing = TRUE, probably to avoid chems that get deleted before metabolization starts from applying end-metabolize effects

since this seems to be the only known problem for now, it's probably best not to risk breaking del_reagant

Bibby0110:
dont like bandaid fixes, fix the underlying pls

Moltijoe:
Aight, i'll give it a good old college try once i get home

PR #15960, opened 2022-09-29
Replaces the Bond Syndikit Special with the V Syndikit Special

Skrem7:

Replaces the very boring Bond bundle with a bundle themed around Big Boss/Venom Snake. Adds two new items to make it more unique, too.

Faded Eyepatch: An eyepatch that gives flash protection AND night vision.
Tactical Fingerless Gloves: Fingerless gloves that provide not only nitrile-level body carrying, but also a 0.9 action modifier on top of that. In addition, it reduces all intent attacks by 0.2 seconds, to a 0.6 sec CD. To compare it with the Gloves of the North Star, GotNS reduces CD by 0.6 seconds on harm intent, to a 0.2 sec CD.

The bundle with a roughly estimated cost of ~28 TC is as follows:
- Camoflauge fatigues (reskinned tactical turtleneck)
- Faded Eyepatch
- Combat Boots
- Tactical Fingerless Gloves
- Tactical Webbing
- CQC Manual
- Toy Pistol with Riot Darts
- Survival Knife
- Stealth Implant
- Premium Cigar Case
- Zippo Lighter

# Wiki Documentation

Bond Syndikit Special will have to be replaced, and the new items detailed in its description or elsewhere on the syndicate items page.

# Changelog

:cl:  
rscadd: Adds the V Syndikit Special bundle, which contains items themed around Big Boss and his basic skillset
rscadd: Adds the faded eyepatch, an eyepatch that grants basic flash protection and night vision
rscadd: Adds the tactical fingerless gloves. They lightly reduce action speed, help with carrying bodies, and slightly reduce CD on all unarmed attacks, regardless of intent
rscdel: Removes the Bond Syndikit Special bundle
/:cl:


code/modules/clothing/gloves/miscellaneous.dm
```diff
@@ -12,6 +12,33 @@
...
+
+/obj/item/clothing/gloves/fingerless/bigboss/equipped(mob/user, slot)
+	..()
+	var/mob/living/carbon/human/boss = user
+	if(slot == SLOT_GLOVES)
+		ADD_TRAIT(user, carrytrait, CLOTHING_TRAIT)
+		boss.physiology.do_after_speed *= 0.9 //Does stuff 10% faster
+		worn = TRUE
+
+/obj/item/clothing/gloves/fingerless/bigboss/dropped(mob/user)
```
SomeguyManperson, line 35: Does this trigger if you take the gloves off and put them in a pocket or somesuch

Skrem7: If you take the gloves in-hand doesn't it technically drop them on the ground for a moment?

The paramedic nanochip code is done in such a way to imply it will remove the body boost if they're removed so the code I did works under that assumption.

If I wanted to be safe I could have the do_after speed increase check for worn so no funny exploits can happen.

Marmio64:
Include a survival knife?

Skrem7:
> Include a survival knife?

Ask and you shall receive.

RG4ORDR:
There's a headband item in the code, I think.
Put it in there too, give it some minor head armor.

Skrem7:
> There's a headband item in the code, I think. Put it in there too, give it some minor head armor.

Headband traditionally gives infinite ammo so it felt a bit awkward to try and include it, I might see if I can feasibly make it unique in some way, then throw some basic cigars into the kit too.

JamieD1 (approved):

JamieD1:
Cornflakes

JamieD1:
Wait why are we removing the Bond Special anyway

Skrem7:
> Wait why are we removing the Bond Special anyway

Its total value is less than 20 TC, nor is it all that exciting in my honest opinion.

Bond can already be effectively emulated using the base 20 TC, the idea of Syndikit Specials is something more unique that you can't really get elsewhere.

It COULD be made again later but it would require a more consider approach and honestly I don't see it as much other than pistol + suppressor + ammo, suit with armor polish probably, then some fancy camouflaged gear of your choosing. There just aren't quite as many direct references to Bond while CQC and the stealth implant are pretty clear nods to Snake.

PR #15963, opened 2022-09-29
Adds the Half-Dragon Syndikit Special bundle

Skrem7:

Adds a Half-Dragon Syndikit Special bundle that has two items: Pure Dragon's Blood and a Flying Fang Tablet. The total value is 32 TC.

Pure Dragon's Blood makes a user a draconid, gives them wings, and grants them the fire breath mutator. Because it's technically giving them the mutator and not inherently granting the ability, it functions just like the mutator that can be purchased by lizard traitors. This means you can technically lose the ability with mutadone.

image
Nice awkward line at the end because of this but it's okay

# Wiki Documentation

To be detailed in Syndikit Special bundles at the very end; Pure Dragon's Blood should be explained as well

# Changelog

:cl:  
rscadd: Adds the Half-Dragon Syndikit Special bundle, which is entirely themed around (go figure) being a representation of a dragon
rscadd: Adds Pure Dragon's Blood found in the bundle, which gives a user the draconid species, fire breath, and wings.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/uplink_kits.dm
+++ b/code/game/objects/items/storage/uplink_kits.dm
@@ -142,7 +142,7 @@
 			new /obj/item/card/emag(src) // 6 tc
 
 /obj/item/storage/box/syndicate/bundle_B/PopulateContents()
-	switch (pickweight(list( "bond" = 2, "neo"=1, "ninja" = 1, "darklord" = 1, "white_whale_holy_grail" = 2, "mad_scientist" = 2, "bee" = 2, "mr_freeze" = 2, "gang_boss" = 1)))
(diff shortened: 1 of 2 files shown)


TheRyeGuyWhoWillNowDie:
the lizard obsession is unstoppable. soon we will have the council all unanimously vote yes on forcing nuke ops to be lizards

Marmio64:
i dont want to become a lizard

Skrem7:
after exactly 6 hours of deliberation after coming up with the idea in 6 minutes i agree it's dumb, could feasibly make it a potential option for lizards but that would make syndikit special code gross and icky

PR #15970, opened 2022-09-30
Nerfs Adrenaline

Skrem7:

Makes it so adrenaline only gives resistance to damage slowdown instead of total immunity. The threshold for damage needed to be taken in a second is 25 instead of 15. This should keep it in line of helping against instantly being slowed by guns while not causing lower damage weapons to suffer relatively to stamina-based options.

Especially combined with how damage slowdown is handled on a better less binary function now, adrenaline is less necessary as a concept in my opinion.

# Wiki Documentation

Ramble about adrenaline at the top of Guide to Combat will need to be updated. Don't know if it ever got added to Guide to Medicine.

# Changelog

:cl:  
tweak: Adrenaline now only reduces damage slowdown instead of removing it entirely. It now procs when you take 25 or more damage in one second, rather than 15.
/:cl:


JamieD1 (approved):

JamieD1:
Conflakes

PR #15984, opened 2022-09-30
Removes the 10% chance of getting a bluespace emag when emagging an emag with another emag

Bibby0110:

This used to be 16 TC thrown out the window versues 12 now.
Bluespace emags are just plain annoying too

# Wiki Documentation

# Changelog

:cl:  
rscdel: You can no longer gamble with your two emags, sadge.
/:cl:


ToasterBiome (changes requested):
Still throwing away a good chunk of your TC

either change the percentage lower to approx. match the investment to be the same as the old or no

PR #16000, opened 2022-10-01
Makes mortar and pestles not made of solid lead leading to tired arms in as little as one cycle

VaelophisNyx:

You can in fact, use this thing a few times now and not get too tired. Instead of...once.

# Changelog

:cl:  
tweak: Mortar and Pestle grinding costs about 1/3rd the stamina now
/:cl:


JamieD1:
1/8th?!? Why do you nerf/buff with so much force.

VaelophisNyx:
because tiny tweaks cant be felt that much
also in this case I have no frame of reference for how much you need to use this thing, so

I can drop it to 1/3rd or 1/4th

JamieD1 (approved):

PR #16015, opened 2022-10-01
re-add frag 12

TheRyeGuyWhoWillNowDie:
since frag 12 is no longer fucking taser shells that stun you for 5 seconds this is fine i think. also explosions dont double explode people anymore so shooting someone doesnt instantly kill and nugget them anymore.

from testing: unarmored assistant takes around 50 damage
HoS takes around 40 damage

the advantage of these over normal slugs is the short knockdown they inflict on victims because they explode

these require 5u glycerol (so corn oil which requires botany) 5u fluacid, and an unloaded tech shell
# Changelog

:cl:  
rscadd: frag 12 can be crafted again
/:cl:

### DIFF
--- a/code/datums/components/crafting/weapons.dm
+++ b/code/datums/components/crafting/weapons.dm
@@ -394,6 +394,17 @@
 	category = CAT_WEAPONRY
 	subcategory = CAT_AMMO
 
+/datum/crafting_recipe/frag12
+	name = "FRAG-12 Shell"
+	result = /obj/item/ammo_casing/shotgun/frag12
+	reqs = list(/obj/item/ammo_casing/shotgun/techshell = 1,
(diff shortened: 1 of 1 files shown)


Rdrazga:
Doesn't frag 12 to some extent still work as a breeching shell?

TheRyeGuyWhoWillNowDie:
> Doesn't frag 12 to some extent still work as a breeching shell?

i mean, kinda, but if you have access to frag 12 you have access to pulse slugs which work a lot better than frag 12 for this purpose

Skrem7:
techslug so super harsh to craft so I'm honest fine with it

would be interested to see AoE testing, especially against armored, if you have the time

JamieD1 (approved):

PR #16046, opened 2022-10-03
Revert "Removes gang from mode rotation. It can still be admin forceed"

boodaliboo:
i liked gangs it was one of the few gamemodes where you had to defend something and it would actually be a sizeable threat completely removing gangs kinda fucks over people wanting to change it in the future as they will never experience it, when gangs go right its beautiful and some of the most fun ive had in ss13


boodaliboo:
i honestly think removing this is a really bad idea. if someone is willing to take it i have a 10$ bounty to make gangs autoassign 7 people per team to counteract one team steamrolling

VaelophisNyx:
sadly that wouldn't cover what makes gangs truly bad; it's a flawed gamemode. Goal: Steal the nuke disk

...that means its basically Revolutionary-Ops. the intent was for them to infight all round until one side wins...but because their goal involves the nuke disk, they focus exclusively on sec and heads to get it

VaelophisNyx:
if anything we should port Families instead

N3D6:
Gang is the worst mode we have by a long shot. The mode is only enjoyed by murderboners and powergamers.

It’s very difficult to understand who is an enemy gangster and who isn’t, so 99% of gang rounds are either a two hour long stalemate because everybody is dead or hiding, or it ends in 15 minutes because one side steamrolled everybody and dominated the station.


Gangs caters to murderboners and powergamers exclusively. I’m usually on the improve don’t remove side of things but nobody ever tries to fix this mode so it’s still a shit show that’s too chaotic for anyone to ever know what’s going on.

00ze-cyclone:
Address the crippling flaws of gang before thinking about reenabling it.

MenacingManatee:
Gang is awful from an admin standpoint as well, probably the #1 worst standard round type to admin for, because it encourages so much pvp with few ways to identify who is on what team in the moment which leads to a lot of questionable killings and ahelps.

JamieD1:
No, port families instead.

PR #16047, opened 2022-10-03
NULL_ENTRY crates are now abandoned crates

ynot01:

Now requires some effort to open NULL_ENTRY crates. Traitors will now try to get to a discreet area with their goodies before opening them lest they get caught.

free 30TC worth of items with little risk is bad

# Changelog

:cl:  
tweak: NULL_ENTRY crates are now abandoned crates and need to be solved open
/:cl:


Cartlord:
An issue with this is that you actually can't reliably solve abandoned crates due to their exact mechanics, so this'd also mean a portion (I do not know exactly how many, but i'm pretty sure it'd be over 10%) of null crates are just wasted.

ynot01:
> An issue with this is that you actually can't reliably solve abandoned crates due to their exact mechanics, so this'd also mean a portion (I do not know exactly how many, but i'm pretty sure it'd be over 10%) of null crates are just wasted.

It is proved that any combination can be solved within seven turns. The average minimal game length is ~5.21 turns.

https://en.wikipedia.org/wiki/Bulls_and_Cows#The_numerical_version
https://slovesnov.users.sourceforge.net/index.php?bullscows,english
http://www.cs.nccu.edu.tw/~chaolin/papers/science3203.pdf (in Chinese)

N3D6:
Yeah this pretty much removes null crates because nobody is doing that shit

Cartlord:
I might be confusing our abandoned crates with another server's, but I thought our players only got three tries.
I'm pretty sure we don't HAVE a wiki page for abandoned crates, so I have to go off what other servers do

00ze-cyclone:
Can you add in the description at the cargo console a mention that it has a code to solve so new players know what to expect since no other cargo crate has that.
Other than that it look good.

ynot01:
> Can you add in the description at the cargo console a mention that it has a code to solve so new players know what to expect since no other cargo crate has that. Other than that it look good.

added a hint in the export description

Chubbygummibear:
i think this is pretty good. 6 tc for as many traitor lootboxes as the station budget allows is definitely questionable. the b&c puzzle isn't hard at all but is a good annoying element to slow down people from opening 5 boxes on the cargo shuttle and being kitted with god knows what in 10 seconds. all the puzzle really does is ask that you have a multitool and 30 seconds to a minute which is more than fair for the emag potentially getting you infinitely more tc

MenacingManatee:
> I might be confusing our abandoned crates with another server's, but I thought our players only got three tries. I'm pretty sure we don't HAVE a wiki page for abandoned crates, so I have to go off what other servers do

``var attempts = 10``
Players get 10 attempts on abandoned crates

Sniblet:
Null crates are stupid but so am I, I've never solved an abandoned crate before
Please do something about the turn limit to be inclusive to the mathematically handicapped

RG4ORDR:
I simply do not see a reason to do this.  

- Null-crates aren't really metashielded,
- They need cargo personnel to either be dead or totally complacent in letting the console get emagged(normally they are)
- AI can see if the console can order crates and I've seen AIs check the console often enough
- The RNG on any of the items is totally unpredictable to where you can get 3 power-fists and codespeak manuals. Or you can get X4,Ebow, and a bo-staff. 
- They already cost 20k which if cargo suddenly lost 100k they can determine that nullcrates got ordered.
- Nullcrate memes are not common enough to warrant this action. 

By all means make their cost higher, which just makes them not worth getting if you can't obtain more than one. Change their contents to be Syndicate Tactical or Special kits. This isn't a needed PR and it's making a change for the sake of making a change.
This isn't like Time-Stop,Rod,Jaunt,Hardsuit, Hammer meta-wizard or Drake doing nothing else than chemistry fuckery. It's a uncommon occurrence that happens like every 20 rounds. And maybe start punishing non-antags for letting antags just order null crates and not report it.

SomeguyManperson:
Wont this effectivelt just lock the crates behind having a solver open in another window

ynot01:
> Wont this effectivelt just lock the crates behind having a solver open in another window

At the least, yes. This still delays the traitor by a good amount of time, hugely widening the window to get caught (previously 0 seconds).

Sniblet:
So stick a timer on the lock instead?

ynot01:
> So stick a timer on the lock instead?

How does this prevent a player from just setting the timer on all of the crates at once? Your solution does not scale.

Sniblet:
I’m talking about hitting the crate with a multitool and waiting a bit, similar to, among many other things, surgery, which I believe we’ve also intentionally avoided attaching a minigame to.
If you can’t prevent them doing that to more than one crate at once (“the lock’s electronics spark out violently, your hand spams!” and you drop the multitool - or something else, like a faulty razor wire trap that does no damage, to bypass insuls - or a random variety of traps to fulfill this function, which each have some tool to bypass them, to reward preparation), you might also try forcing people to break the crate by not giving it a toggle open button. Making these into abandoned crates already forgoes the sneaky disguise aspect of their outer look, so we could make some plastitanium reskin of those wooden box crates that need a crowbar, without the crowbar interaction.
There’s lots of ways to do this that don’t significantly favor players who simply have their browser open, while punishing others by literally blowing up their 20,000 credits in their face.

JamieD1:
Like it, but not sure about the current abandoned crate solve method.

ynot01:
pending abandoned crate rework

Bibby0110:
Reopen when that rework is done

PR #16049, opened 2022-10-04
captain's antique laser gun now shoots hellfire lasers

TheRyeGuyWhoWillNowDie:
you know for a gun that only 1 exists of, and that's a grand theft objective, and one that the owner of the gun cannot take outside of extreme emergencies, this gun really is worse than the ones you can print for no cost from the armory protolathe huh

the antique fires a special version of hellfire that has the same energy cost as normal lasers so you're not losing ammo capacity

# Changelog

:cl:  
tweak: antique lasergun now shoots hellfire
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition/energy/laser.dm
+++ b/code/modules/projectiles/ammunition/energy/laser.dm
@@ -7,6 +7,9 @@
 	e_cost = 130
 	select_name = "maim"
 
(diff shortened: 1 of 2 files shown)


Marmio64:
even if its a grand theft item, i feel like giving it hellfire with normal ammo capacity and still letting it auto recharge is too powerful, especially since 80% of the time its an antag who is killing every hapless lad around with it.

PR #16066, opened 2022-10-05
IPC Lungs require much less pressure to function

VaelophisNyx:

Takes IPC lungs from requiring nearly 10x the pressure of everyone else on internals, down to about 10-15% more
~24kPa minimum on oxygen internals

Why? Because that pressure determines how fast your tank drains, and at 95 kPa emergency internals only last about a minute or two for IPCs. And IPCs aren't even breathing, it's cooling. This is however simply easier than making IPCs not actually consume their internals gases.

# Wiki Documentation

Requires an update to the Notes section on the IPC wiki page

# Changelog

:cl:  
tweak: IPCs can run on sensible amounts of pressure instead of requiring an entire emergency internals tank per 2 minutes
/:cl:

### DIFF
--- a/code/modules/surgery/organs/lungs.dm
+++ b/code/modules/surgery/organs/lungs.dm
@@ -507,7 +507,7 @@
 	var/temperature = breath.return_temperature()
 	for(var/id in breath.get_gases())
 		var/moles = breath.get_moles(id)
-		total_heat_capacity += GLOB.meta_gas_info[id][META_GAS_SPECIFIC_HEAT] * moles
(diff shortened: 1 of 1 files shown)


CMOisLing:
Im not entirely sure how the breath calculations for Ipc are done but if you gave them internals full of hyper noblium would that reduce internal requirements by 100x

VaelophisNyx:
given that this takes their minimums down to 3pKa for plasma, hypernob might just break the math entirely

CMOisLing:
Cant wait to try it out 👍

JamieD1 (approved):

PR #16091, opened 2022-10-08
[META] Reworks Meta Service to be less bad and also reworks science too a little to fit it and more stuff

ToasterBiome:

I played an entire round roleplaying in the "club" and nobody talked to me except a HOS (cool) and the bartender to refill my drink. I determined this is because ~~we are a low RP server~~ there is literally no way to see who is in this location because it's comically removed from the rest of, well everything. This takes a page out of TG's metastation (which is very nice I might add) to rework it into something better.

image
There's now a service back room, bar based off of TG, redone hydroponics to fit better, 1 maint room as removed and had to be shuffled around. Maint is kinda empty but I can add to it

image
Moves janitor closet to be where science sec used to be. Moved science sec to where the break room used to be (sorry) and turned the experimentor room on it's side to fit.

image
Artist's coven is here now.

# Changelog

:cl:  
mapping: reworks metastation service area
mapping: reworks science sec/experimentor area on meta
mapping: moves artists coven on meta
/:cl:


Cartlord:
Really not a fan of there not actually being a smartfridge in the kitchen, and not sure why there aren't two sets of Botany vendors any more. Also, did you remove one Botanist spawn?

Cartlord:
Regarding the intent of this P.R.: I think MetaStation's bar being largely ignored may be due to its placement on the station rather than any finer details of the bar layout, as you don't pass by it unless you're going south from Engineering or southeast from Sec, while anyone going to Medical, Sci, or Evac on BoxStation will pass by the bar.

ToasterBiome:
>Also, did you remove one Botanist spawn?

I actually forgot the botanist spawns which I have to add :)

>and not sure why there aren't two sets of Botany vendors any more. 

There hasn't been two since janitor's room needed to be expanded, there's still garden

ToasterBiome:
> Regarding the intent of this P.R.: I think MetaStation's bar being largely ignored may be due to its placement on the station rather than any finer details of the bar layout, as you don't pass by it unless you're going south from Engineering or southeast from Sec, while anyone going to Medical, Sci, or Evac on BoxStation will pass by the bar.

A step to help mitigate its poor placement is actually being able to see who's in it besides the very front area and only from one direction

LazennG:
good pr since the lounge is totally blocked off from other rooms and nobody has a reason to go inside it normally

Cartlord:
I would suggest moving the smartfridge to the Kitchen's side of the service hall so Chefs don't have to walk as much to get ingredients (keep in mind they don't have a plant bag by default), and maybe adding more Hydroponics trays or moving the existing ones to be closer to the smartfridge. You might also want to remove one Chef spawn - the Kitchen only has one or two of each machine, so a third one should probably only be there if one of the existing two dies or fucks off into maints, and the Head of Personnel can open another job-slot if that happens.
I'm a big fan of the service hall itself being within view of both Hydroponics and the Kitchen, since that reduces the odds of some asshole breaking in to use the lathes.
I don't know how many Botanist spawns are _supposed_ to be there, and the webmap doesn't show the spawns we have on current Meta, but whatever number you decide on, you should probably also have that exact number of lockers, just so everyone gets the same tools - and yes, I know that that would also mean more data disks, and make the Botanical Technology node even less important. _I don't think that matters_, because that node already barely matters.

VaelophisNyx:
keep in mind spawners dont dictate job slots; that's a map config thing

Cartlord:
i was not aware

VaelophisNyx:
mhm, that's why I was letting you know
it's maybe unintuitive but definitely nice since we can make 12 botanist spawns and not have 12 botanists.

RG4ORDR:
I kinda like how the theater room is separated by doors in default meta, just saying it's not a bad idea to make Bar/Kitchen/Theater better. But I think the theater being seperated by airlocks and walls makes sense.

patpol4 (approved):
In my opinion this is objectively better.

Mqiib:
image

The one tile under the crate is considered space, not part of the maint area.

SomeguyManperson:
I think having a window at the service to kitchen area would be nice so the botanista can ask the cook if they need plants easier

JamieD1:
Chefs room looks a little small, other then that, what other people said here should be taken into consideration

ToasterBiome:
I believe I have created a superior version:
image

@Cartlord I have shortened the service area to have a connecting space between the chef area and the hydroponics area.

> I think having a window at the service to kitchen area would be nice so the botanista can ask the cook if they need plants easier

See above, I think this is a better solution than adding a window




> Chefs room looks a little small, other then that, what other people said here should be taken into consideration

I have expanded the Chef area slightly, including the freezer, to better give available space for chefs to work.

JamieD1 (approved):

PR #16125, opened 2022-10-10
Makes vial boxes printable and lets them go in the medical belt

Moltijoe:
With the addition of hyposprays trying to hold all the vials for the chemicals became difficult
it usually just became a matter of using a bluespace beaker and refilling the syringes
This just lets you carry a bunch of filled vials instead of a beaker if you want.

Vial boxes bind to the first person to open them
Medical personnel, HoS, Captain, and HoP are able to open them regardless of bind
This lets not medical players use printed vial boxes

:cl:  
rscadd: Can now print vial boxes after researching them
tweak: Vial box can now go in the medical toolbelt
/:cl:


Cartlord:
In favor of printing, against making them fit in belts.

Moltijoe:
> In favor of printing, against making them fit in belts.

Should beakers be removed from belt too? Because thats currently what happens instead of this.
If you think 6 vials is too many, i could very well reduce the printed vial box size to 4 or 5

JamieD1:
Dont like how they can fit in belts, they would not fit.

Moltijoe:
> Dont like how they can fit in belts, they would not fit.

I thought of the vial boxes having a sort of clip that would latch on to the belt.
The main goal i had was to replace the "just have a beaker of the drugs in your belt".
If youre against the vial box for balancing reasons, i could reduce its size.

Moltijoe:
I'm also considering removing beakers from the medical belt, probably only if this PR gets merged though

JamieD1:
Make them printable but not able to go in belts, also clip onto the belt with lots of glass, that would be very very awkward weight wise maybe but im not sure to be honest.

Moltijoe:
> Make them printable but not able to go in belts

okay, easy fix

code/game/objects/items/storage/lockbox.dm
```diff
@@ -324,3 +324,6 @@
 /obj/item/storage/lockbox/vialbox/blood/PopulateContents()
 	for(var/i in 1 to 6)
 		new /obj/item/reagent_containers/glass/bottle/vial(src)
+
+/obj/item/storage/lockbox/vialbox/printed
+	var/registered_name = null
```
ReddicusDragon: registered_name seems to be an unused variable.

ReddicusDragon (changes requested):
Just remove the unused variable or point out where you use it and I'd say you're good to go

code/game/objects/items/storage/lockbox.dm
Moltijoe: ah, right

ReddicusDragon (approved):

Bibby0110 (approved):

PR #16130, opened 2022-10-10
Replaces Colombus Day with Indigenous Peoples Day

VaelophisNyx:

This is the direction that things are going in the USA IRL, and realistically, the way they SHOULD be going.
Kills off a reference to a colonizer, human trafficker, rapist, and tyrant; replaces that with the better option, culturally speaking.

also adds a separate instance for Canada, who has their own holiday for this earlier in the year

# Wiki Documentation

If we have a page that covers Holidays, this needs updating; if we DONT, maybe we should add one

# Changelog

:cl:  
spellcheck: Colombus Day is now Indigenous Peoples Day
/:cl:

### DIFF
--- a/code/modules/holiday/holidays.dm
+++ b/code/modules/holiday/holidays.dm
@@ -495,11 +495,15 @@
 	begin_month = OCTOBER
 	begin_weekday = MONDAY
 
-/datum/holiday/indiginous
-	name = "Indiginous Peoples Day"
-	begin_week = 2
+/datum/holiday/indigenous
+	name = "Indigenous Peoples Day in the United States"
+	begin_day = 10
 	begin_month = OCTOBER
-	begin_weekday = MONDAY
+
+/datum/holiday/indigenous
+	name = "Indigenous Peoples Day in Canada"
+	begin_day = 21
+	begin_month = JUNE
 
 /datum/holiday/mother
 	name = "Mother's Day"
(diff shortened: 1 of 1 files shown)


Rdrazga:
If the reference is a problem why not just remove it instead, it’s a specifically US holiday on a not specifically US server. This WILL cause political debate otherwise given there are people on here who will be against this change (me included)

VaelophisNyx:
we have both US and Canadian thanksgiving
we have a LOT of national holidays

removing it alone would be weird if we kept all the rest?

Rdrazga:
Then keep it as Columbus Day by that logic since it is the nationally recognized bank holiday. If you feel it is dividing remove the reference.

VaelophisNyx:
I'm overwriting it with the one that is slowly being accepted, state-by-state :|

ToasterBiome (approved):

VaelophisNyx:
do love the supposed dozens that will be against this literally never appearing

hmmmmmmm

JamieD1:
Christ alive.
Shall we remove thanksgiving too?

JamieD1:
> do love the supposed dozens that will be against this literally never appearing
> 
> hmmmmmmm

Because anyone that is against it you will call a label most likely.

VaelophisNyx:
Not really, just think it was funny that there was an assertation that "many would be against it" then just

no-one showed up.

As far as thanksgiving, it's fine mostly, it isn't directly celebrating a man that was bad enough that even the spanish crown decided "hmm maybe he's a bit TOO extreme" and had him put in prison

JamieD1:
Sorry but im not merging this due to the controversial standing currently in the united states with this, also its political, up for removing holidays but not changing them to PC issues.

PR #16110, opened 2022-10-08
reworking borg combat to be less reliant on 1 click instant kills for both sides

TheRyeGuyWhoWillNowDie:
borg combat is currently 1 click wins on both sides which is fucking awful 2011 combat that we got rid of for humans a couple years ago, this changes a bunch of borg stuff to be less binary

Flashes:
Immobilize and blur the cyborg's vision for the same time as old flashes. They can still act. Down from instantly locking them in an inescapable chainstun. Locked borgs are affected the same way they previously were

Lockdown:
Locked down borgs can now move and act instead of being trapped forever, HOWEVER, locked borgs move 4x slower, have 2 of their module slots disabled, and their remaining module breaks at the same damage threshold as the third module on unlocked cyborgs. They can only interact with stuff they're next to, and borgs with guns cannot shoot. This is down from lockdown being an instant inescapable GBJ button

Detonate:
Remote detonation of cyborgs can only be done by the cyborg's connected master AI. Down from instantly gibbing the cyborg at literally the press of a button.

EMPS:
Emps do double damage against borgs (40 on heavy, 20 on light) and their ability to pulse internal wires has been restored as lockdown is no longer a stupid infinite stun. EMPs now only disable the borg's modules (16 seconds heavy like how it was before, jesus christ, 6 seconds light) Down from being an instant kill on hit.

Stun arm:
Now has a 1.2 second cooldown between attacks, and only does 60 stamina damage.

Harm alarm:
Knocks down instead of being a fucking paralyze because ranged stuns are stupid

Self destruct:
Can't be done while locked.

Hugging module:
Shock hugs now respect insulation when electrocuting people. This is down from it being an instant inescapable chainstun for humans.

# Changelog

:cl:  
tweak: borg combat has been reworked to be a lot less binary
/:cl:

PR #16136, opened 2022-10-11
burning stun combat: fuck flashes

TheRyeGuyWhoWillNowDie:
stun combat is not fun combat
flashes are not in limited supply (robotics can make infinite) and our maps are absolutely lousy with them, some maintenance ruins even have flashes ripe to be stolen.
doing this can let us walk back from 'EVERYONE IMPORTANT NEEDS FLASH PROTECTION' and should discourage powergaming a little bit
mounted flashes unaffected
would pair best with #16110

flashes now do not instantly stun people, instead being an 8-12 second knockdown. flash confusion has been doubled to accommodate for this, flashes now also do 80 stamina. also flashes now have a 9 second cooldown so that your victim has time to flip down their welding helmet or whatever instead of just being fucked. i also preemptively prevented you from carrying 2 flashes to instantly cuck somebody because i know our playerbase would do it.

use cases this affects:
security: flash first and then stun baton instead of click once to win
heads: your flash is for self defense and disarming and knocking down whoever's attacking you is pretty good especially because you can then stamcrit them in 1 telescopic baton hit
borgs: validhunter borgs can get fucked for all i care, emagged/malf ones have their hacked modules
antags: pair with stunprod or something or use your antag abilities instead of shitty stun combat

aoe flash retains its 0 cooldown
as a side effect makes rev speedrun (convert 30 people in 2 seconds) less prevalent because the 9 second flash cooldown applies to revs converting people too

# Changelog

:cl:   
tweak: flashes are now not 1 click wins lol
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -135,6 +135,7 @@
 #define TRAIT_DISFIGURED		"disfigured"
 #define TRAIT_XENO_HOST			"xeno_host"	//Tracks whether we're gonna be a baby alien's mummy.
 #define TRAIT_STUNIMMUNE		"stun_immunity"
+#define TRAIT_FLASHED		"flashed_recently"
 #define TRAIT_STUNRESISTANCE    "stun_resistance"
 #define TRAIT_SLEEPIMMUNE		"sleep_immunity"
 #define TRAIT_PUSHIMMUNE		"push_immunity"
--- a/code/modules/assembly/flash.dm
+++ b/code/modules/assembly/flash.dm
@@ -93,13 +93,14 @@
 	else
 		return typecache_filter_list(target_loc.GetAllContents(), GLOB.typecache_living)
 
-/obj/item/assembly/flash/proc/try_use_flash(mob/user = null, nocooldown)
+/obj/item/assembly/flash/proc/try_use_flash(mob/user = null, nocooldown = FALSE)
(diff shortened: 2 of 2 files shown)


neeshacark:
maybe dont apply the cooldown on flashes to the ones revheads get

Skrem7:
Make the CD less strict then it's fine

Will inadvertently make chain flashing on headrev impossible (not bad because it's a stupid gimmick) but 9 seconds is a long, LONG time, especially if you decide to click Z with it like before.

Speaking of that click Z should have its own CD if it doesn't have one I assume it's tied to the target attack one

TheRyeGuyWhoWillNowDie:
> you

oh fuck you're right i forgot about aoe flash

TheRyeGuyWhoWillNowDie:
turdis is shitting itself on this pr and i dont know why

iamthesilly:
8-12 seconds of knockdown is too much in my opinion, i think 5-8 is more reasonable considering it also confuses and causes slowdown from stamina damage, on top of blinding you

TheRyeGuyWhoWillNowDie:
> 8-12 seconds of knockdown is too much in my opinion, i think 5-8 is more reasonable considering it also confuses and causes slowdown from stamina damage, on top of blinding you

this is changed from it hardstunning you from 8-12 seconds, id fucking remove the knockdown if i could but i'd be burnt at the stake by security players and other stun combat enthusiasts

JamieD1:
Had this discussion before, flashbangs stay the same.

PR #16139, opened 2022-10-11
Ynot has no idea how strong explosions are

Mqiib:
Returns the PDA bomber thing to just a light explosion instead of a heavy explosion (which kills you and deletes most of your equipment)

# Changelog

:cl:  
 
tweak: PDA bomber is no longer a free kill

/:cl:


SomeguyManperson:
Mr maintainer slap the i ded label on this pr right now

Maia-J:
How about instead of that we make it cost more?

Mqiib:
> How about instead of that we make it cost more?

no
You suggesting this means you don't understand WHY this is a bad thing. it's bad because currently it's a way to nuke from orbit someone and almost all of their equipment without putting yourself at risk AT ALL. There's no tracing it, there's no reacting to it, and there are pretty solid odds you'll just straight up die. it's like "What if esword but instead of being a melee weapon it was infinite range, you didn't need to see the target, it destroyed almost all of their equipment, it hard stunned them with a good chance to delimb, and it did 120 damage but oh no it only has 4 uses!!!"

00ze-cyclone:
It has the ided label, we are required to merge it now.

Bibby0110 (approved):

PR #16143, opened 2022-10-11
Embed pulling now does less damage but more wound

ynot01:

1/4 damage
attempts slash wound with the damage it used to deal + 20 wound bonus

# Changelog

:cl:  
tweak: embed pulling now does less damage and more wounding
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon_defense.dm
+++ b/code/modules/mob/living/carbon/carbon_defense.dm
@@ -146,7 +146,8 @@
 	if(!do_after(user, time_taken, needhand = 1, target = src) && !(choice in body_part.embedded_objects))
 		return
 	var/damage_amount = choice.embedding.embedded_unsafe_removal_pain_multiplier * choice.w_class
-	body_part.receive_damage(damage_amount, sharpness = SHARP_EDGED)//It hurts to rip it out, get surgery you dingus.
+	body_part.receive_damage(damage_amount * 0.25, sharpness = SHARP_EDGED)//It hurts to rip it out, get surgery you dingus.
+	body_part.check_wounding(WOUND_SLASH, damage_amount, 20, 0)
 	if(remove_embedded_object(choice, get_turf(src), damage_amount))
 		user.put_in_hands(choice)
 		user.visible_message("[user] successfully rips [choice] out of [user == src? p_their() : "[src]'s"] [body_part.name]!", span_notice("You successfully remove [choice] from your [body_part.name]."))


Mqiib:
Ok but why

milk-is-cool:
based pr

ynot01:
> Ok but why

When you have something stuck in you, it is a common trope to keep it in so you don't die of massive blood loss

The damage is already mostly done by the entry, pulling it out will typically not harm as much

githubuser4141:
> > Ok but why
> 
> When you have something stuck in you, it is a common trope to keep it in so you don't die of massive blood loss
> 
> The damage is already mostly done by the entry, pulling it out will typically not harm as much

Huh? No, you're wrong. It's a common, **incorrect** movie/media trope that you should self-remove shrapnel asap, afterwards your wound will immediately heal. _In real life, you shouldn't because it's a fact that doing so could cause serious bleeding/further damage._

On Yogstation unfortunately the devs followed the movie trope and coded it so,

Leaving shrapnel in will cause 1-10 brute damage every second regardless of what you're doing or where the shrapnel is or what IT'S doing. 
You die of 100 brute damage regardless of the source and location. E.g. 100 brute to your toe from stepping on lego is as lethal as 100 brute to your torso from gunshot wounds.

Therefore, on Yog, and the majority of SS13 servers it's ideal to remove shrap ASAP because in this video game, not doing so is actually **extremely lethal to your character.** Because video games.

Mqiib:
> > > Ok but why
> > 
> > 
> > When you have something stuck in you, it is a common trope to keep it in so you don't die of massive blood loss
> > The damage is already mostly done by the entry, pulling it out will typically not harm as much
> 
> Huh? No, you're wrong. It's a common, **incorrect** movie/media trope that you should self-remove shrapnel asap, afterwards your wound will immediately heal. _In real life, you shouldn't because it's a fact that doing so could cause serious bleeding/further damage._
> 
> On Yogstation unfortunately the devs followed the movie trope and coded it so,
> 
> Leaving shrapnel in will cause 1-10 brute damage every second regardless of what you're doing or where the shrapnel is or what IT'S doing. You die of 100 brute damage regardless of the source and location. E.g. 100 brute to your toe from stepping on lego is as lethal as 100 brute to your torso from gunshot wounds.
> 
> Therefore, on Yog, and the majority of SS13 servers it's ideal to remove shrap ASAP because in this video game, not doing so is actually **extremely lethal to your character.** Because video games.

I think what he means is that removing it is "good" in-game but bad irl so this makes it so that it's bad in-game by causing significant bleeding wounds if you just yank shit out, not that irl its a good thing to pull them out.

VaelophisNyx:
does this affect tape

Bibby0110 (approved):

PR #16145, opened 2022-10-11
Revolutionary heads now get a (hidden) flash implant rather than converting with just any old flash on the ground

SomeguyManperson:

Obliterates any and all metashield related problems with revs by giving the revheads a flash that is completely obvious in what it does but is also perfectly hidden when not in use, solving both problems of "revs is too easy to game if you catch someone with a random flash" and "revs is too heavily metashielded because the round ending over random flashes is boring", and this time they actually spawn.

This also conveniently gives us a subtype of flash for specifically revs with which we can balance cooldowns and shit around (implant flashes are already slower than normal flashes, and like headrevs using flashes, they won't burn out_

"but theos peopel will not be able to confirm rev sightings"
they already can't, this won't affect that. In any situation, the suspect item could be metashielded or hidden anyways.

"but theos htis is hubgofx"
hugbox for who, revs get their conversion tool hidden easier but also don't get the rules pampering them for carrying 50 flashes in their backpack because "well you dont know its a rev flash"

# Wiki Documentation

revheads get a flash implant in an arm  rather than normal flashes

# Changelog

:cl:  
tweak: revheads get a flash implant in an arm rather than normal flashes to do their conversions with
experimental: REVHEADS GET A FLASH IMPLANT, IT IS THE ONLY FLASH THEY CAN CONVERT WITH. DO NOT TRY TO USE OTHER FLASHES YOU WILL JUST LOOK STUPID
/:cl:


code/modules/antagonists/revolution/revolution.dm
```diff
@@ -253,17 +252,9 @@
...
-			"right pocket" = SLOT_R_STORE
-		)
-		var/where = H.equip_in_one_of_slots(T, slots)
-		if (!where)
-			to_chat(H, "The Syndicate were unfortunately unable to get you a flash.")
-		else
-			to_chat(H, "The flash in your [where] will help you to persuade the crew to join your cause.")
+		var/obj/item/organ/cyberimp/arm/flash/rev/T = new
+		T.Insert(H, special = TRUE, drop_if_replaced = FALSE)
+		to_chat(H, "The flash implant in your arm will help you to persuade the crew to join your cause.")
```
tattax: some input saying they can only use this to convert and bigger text font would help

code/modules/assembly/flash.dm
```diff
@@ -193,21 +202,24 @@
...
+		if(!H.client)
+			to_chat(user, span_warning("This mind is so vacant that it is not susceptible to influence!"))
+			return
+		if(H.stat != CONSCIOUS)
+			to_chat(user, span_warning("They must be conscious before you can convert [H.p_them()]!"))
+			return
+		if(converter.add_revolutionary(H.mind))
+			times_used -- //Flashes less likely to burn out for headrevs when used for conversion
+		else
+			to_chat(user, span_warning("This mind seems resistant to the flash!"))
```
tattax:
```suggestion
	var/datum/antagonist/rev/head/converter = user.mind?.has_antag_datum(/datum/antagonist/rev/head)
	if(!converter)
		return
	if(!H.client)
		to_chat(user, span_warning("This mind is so vacant that it is not susceptible to influence!"))
		return
	if(H.stat != CONSCIOUS)
		to_chat(user, span_warning("They must be conscious before you can convert [H.p_them()]!"))
		return
	if(converter.add_revolutionary(H.mind))
		times_used -- //Flashes less likely to burn out for headrevs when used for conversion
	else
		to_chat(user, span_warning("This mind seems resistant to the flash!"))
```

tattax: or just return on user not having mind

Cartlord:
didn't you already try to do this

VaelophisNyx:
Gus did and it didnt work

SomeguyManperson:
> didn't you already try to do this

Yes but it didnt work and i couldnt fix it at the time

JamieD1 (approved):

PR #16153, opened 2022-10-12
Makes Hacked Drones Not 100% Garbage AND Makes Hacking A Drone Take a Multitool

VaelophisNyx:

Hacked drones went from movespeed -1 to 1, just because some asshole hit them with a wrench
they also lost ventcrawl entirely?
one could argue "well that's balanced!"
maybe! but sometimes balance gets in the way of fun. A lot. In this particular case it means no one ever hacks drones, because they move at a snail's pace, cant go anywhere fast, and are VISIBLY hacked, and still have only 30 HP so they die in about 2 hits from a crowbar.

this corrects this by just not changing their speed nor movement values. If you wanna make a little tool of extreme destruction, go for it. Just be aware their laws DO NOT state they cant kill **you**

they still cant use guns or directly harm since hacking does not turn pacifism or no-gun off

also wrench? to hack an electronic device? huh?
Multitool that shit.

# Changelog

:cl:  
tweak: Hacked Drones can move at full speed and ventcrawl again
tweak: Hacking a Drone needs a Multitool instead of a wrench
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/drone/interaction.dm
+++ b/code/modules/mob/living/simple_animal/friendly/drone/interaction.dm
@@ -91,7 +91,7 @@
 		else
 			to_chat(user, span_warning("[src]'s screws can't get any tighter!"))
 		return //This used to not exist and drones who repaired themselves also stabbed the shit out of themselves.
-	else if(I.tool_behaviour == TOOL_WRENCH && user != src) //They aren't required to be hacked, because laws can change in other ways (i.e. admins)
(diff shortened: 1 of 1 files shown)


VaelophisNyx:
working on a check to prevent drones from hacking other drones, an oversight I'm not sure anyone has exploited, but would be probably after this is done if I dont

VaelophisNyx:
hang on I've been editing FIXING hacked drones!
gotta go find where they get hacked...

PR #16157, opened 2022-10-12
Cleans Up Belt Storage Allowances + Rebalances Belts

VaelophisNyx:

Long story short, belts were fucky and people were ignoring subtype pathings 100%

so now that's fixed up. Tested to be sure, nothing I removed is broken, can still put intended items in belts

onto the balance portion!:

Belts that could hold a specific type of holo barrier projector can now hold **any**
this should not be a problem in 90% of cases, the only ones I can think of being Janitorial and Sec projectors since they're solid objects. But, you...likely wont get those?

Most specific objects have been repathed into their subtypes wherever it wouldn't shatter balance. See: Sec Belts and Bandoliers can hold *any* loose ammo now, not just shotgun shells

Over-all this is a defatting of the belts file and moderately balance impactful, but even that's a stretch

# Wiki Documentation

I KNOW skrem JUST did a belt update on the wiki, or is in the middle of it, so that will require updating to this PR's changes

# Changelog

:cl:  
tweak: Belts Less Code Bloaty + Can Hold Less Strict Types Of Things
/:cl:


CMOisLing:
If im reading this right, can I have a medbelt loaded with random organs?

VaelophisNyx:
that'd be the explorer webbing
but yes

code/game/objects/items/storage/belt.dm
```diff
@@ -257,22 +244,21 @@
...
 		/obj/item/ammo_box,
 		/obj/item/storage/box/rubbershot,
 		/obj/item/storage/box/lethalshot,
 		/obj/item/storage/box/breacherslug,
 		/obj/item/storage/box/beanbag,
 		/obj/item/reagent_containers/food/snacks/donut,
-		/obj/item/kitchen/knife/combat,
+		/obj/item/kitchen/knife,
 		/obj/item/flashlight/seclite,
-		/obj/item/melee/classic_baton/telescopic,
```
JamieD1, line 270: Why is this removed?

code/game/objects/items/storage/belt.dm
```diff
@@ -356,28 +342,25 @@
 		/obj/item/stack/sheet/bone,
 		/obj/item/lighter,
 		/obj/item/storage/box/fancy/cigarettes,
-		/obj/item/reagent_containers/food/drinks/bottle,
```
JamieD1, line 360: Why is this removed

JamieD1 (changes requested):

code/game/objects/items/storage/belt.dm
VaelophisNyx: They already have /obj/item/reagent_containers/food/drinks on the list, subtype pathing works for storages.

code/game/objects/items/storage/belt.dm
VaelophisNyx: same reason as the mining belt's bottles removal; it's subtype supported and the basetype classic baton was already on the belt

ReddicusDragon (commented):
I see absolutely nothing wrong with this, nothing can go wrong

EDIT: I do see some changes that I would approve, but I have neither the brain power nor the time to review the other changes.

JamieD1 (approved):

JamieD1:
Not sure if I like the fact a mining belt can hold a medicine bag etc.

JamieD1 (changes requested):
Not sure if I like the fact a mining belt can hold a medicine bag etc.

VaelophisNyx:
I don't see the issue tbh. It occupies a slot they need for other stuff, and as such I doubt you'll see them take one on their belts; you're more likely to see them with them stuffed in their backpack, which already happens

PR #16178, opened 2022-10-16
Repaths /obj/screen into /atom/movable/screen

ChesterTheCheesy:
Does what tg did in https://github.com/tgstation/tgstation/pull/54403.

Should improve performance when loading screen objects (like damage overlays which for some reason lag you for 3-5 seconds).

I tested this on my own server, and it does make a slight difference.

# Changelog
:cl:  
tweak: should improve performance by repathing screen objects into a more basic format
/:cl:


Bibby0110:
I think this should only make a difference for memory usage? Good change either way

JamieD1 (approved):

PR #16185, opened 2022-10-17
Reduces pipe volume

SapphicOverload:

Reduces the volume of pipes from 200 liters for everything to 50 liters for straight/bent pipes, 75 liters for 3-way manifolds, and 100 liters for 4-way manifolds. This reduces the volume of most atmospherics components/machines from 200 to 50 liters per node as well.

The point of this is that currently whenever you connect a canister to anything at all at least 1/6th of the gas that was in it will be left behind in whatever it's connected to. Reducing the volume mitigates the loss significantly.

# Changelog

:cl:  
tweak: reduces the volume of pipes and most atmospherics components
/:cl:

### DIFF
--- a/code/__DEFINES/atmospherics.dm
+++ b/code/__DEFINES/atmospherics.dm
@@ -200,7 +200,7 @@
 /// (kPa) What pressure pumps and powered equipment max out at.
 #define MAX_OUTPUT_PRESSURE					4500
 /// (L/s) Maximum speed powered equipment can work at.
-#define MAX_TRANSFER_RATE					200
+#define MAX_TRANSFER_RATE					50
 /// 10% of an overclocked volume pump leaks into the air
 #define VOLUME_PUMP_LEAK_AMOUNT				0.1
 //used for device_type vars
--- a/code/game/machinery/computer/atmos_control.dm
+++ b/code/game/machinery/computer/atmos_control.dm
@@ -132,6 +132,7 @@ GLOBAL_LIST_EMPTY(atmos_air_controllers)
 	var/data = list()
 
 	data["sensors"] = list()
+	data["max_rate"] = MAX_TRANSFER_RATE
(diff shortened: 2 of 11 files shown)


ToasterBiome:
Why? Also explain what they were previously

ToasterBiome:
image
and explain the purpose of this (as in why do this)

iamthesilly:
> image and explain the purpose of this (as in why do this)

this is to keep the scrubber nodes at the same volume as they were before because reducing it would negatively impact the supermatter's cooling and filtering

iamthesilly:
my reasoning is that currently whenever you connect a canister to anything at all it's inevitable that at least a 6th of your gas gets left behind in whatever you connected it to. on top of this, having 13 one meter long pipes be the same volume as an entire 1x1x2.5 area is a bit excessive, so i decided to change it.

Mqiib:
image
BUT MY HIGH-EFFICIENCY VOLUME STORAGE STUPIDITY!!! NOOOOOOOO

CMOisLing:
Does this effect radiators?

JamieD1:
If you are literally just doing this to remove the left overs I cant support it. Too much relies on 200 L/S if you want to remove left overs think of another way.

ToasterBiome:
wouldn’t this just leave leftovers still? if you wanted a perfect system why not make vacuum connectors or something that will pull the rest of the gas

VaelophisNyx:
> If you are literally just doing this to remove the left overs I cant support it. Too much relies on 200 L/S if you want to remove left overs think of another way.

what exactly relies on 200L/s operation?

JamieD1:
> > If you are literally just doing this to remove the left overs I cant support it. Too much relies on 200 L/S if you want to remove left overs think of another way.
> 
> what exactly relies on 200L/s operation?

Every historic atmos setup ever?

PR #16192, opened 2022-10-17
hugboxxes the surgical drill because i hate medbay players

TheRyeGuyWhoWillNowDie:
reduces the crazy wounding bonuses on the surgical drill (10 and 10 bare wound bonus) because these are backpack sized and pointy so because of our crackhead wounding system they usually cause t2 puncture wounds on the first hit. medbay players know its useless but take it anyways because funny murder

# Changelog
:cl:  
tweak: the surgical drill is now less robust
/:cl:

### DIFF
--- a/code/modules/surgery/tools.dm
+++ b/code/modules/surgery/tools.dm
@@ -123,8 +123,8 @@
 	w_class = WEIGHT_CLASS_NORMAL
 	attack_verb = list("drilled")
 	sharpness = SHARP_POINTY
-	wound_bonus = 10
-	bare_wound_bonus = 10
+	wound_bonus = -15
+	bare_wound_bonus = 5
 
 /obj/item/surgicaldrill/suicide_act(mob/user)
 	user.visible_message(span_suicide("[user] rams [src] into [user.p_their()] chest! It looks like [user.p_theyre()] trying to commit suicide!"))


ToasterBiome:
Please explain what changing these numbers will result in

TheRyeGuyWhoWillNowDie:
> Please explain what changing these numbers will result in

the chance the drill will do a puncture wound every hit is reduced from "basically always" to "in line with everything else" and spamming the drill against someone wearing a helmet/armor/whatever is less likely to wound them. against unarmored people drilling the head is still robust but it wont give t2 puncture wounds (challenging to heal and make getting hit worse) every single attack

RG4ORDR:
> Please explain what changing these numbers will result in

I've been on the end of this before.
The drill has an insanely high chance to cause bleed wounds, even through armor they can cause bleed wounds with relative ease. Compared to something like the saw,spear,scapel, or whatever non-traitor/mining item one can get on the station. It's rarely used except by a couple people or someone needing a weapon desperately as the medical drill is only used in one surgery. My example is that as a changeling I was aiming to kill a paramedic with my armblade. Despite me having block chance, and much higher damage, he was able to deliver some pretty nasty wounds that ended up getting me killed in the end. 

A changeling armblade lost to a medical drill because it was able critically wound me multiple times.

VaelophisNyx:
simply go get healed up with some crafted or stolen sutures

idk this isn't a problem imo

JamieD1:
Why have you nerfed it completely the other way though instead of just reducing it.. Also its a drill

Marmio64:
way too severe of a nerf, just lower bare wound and normal wound bonus by 5 each at the most.

JamieD1:
Feel free to reopen with changes

PR #16196, opened 2022-10-18
reduce admin music default volume again to 7%

ynot01:
this is the optimal music volume % everything else is ear rape


CMOisLing:
Hugbox

JamieD1 (approved):

PR #16209, opened 2022-10-20
Add a machine to xenobio so they can actually contribute and not be worthless

MajManatee:

Adds a plort refinery machine to xenobio that turns slime cores into research points! Finally something that makes the "Research" you are doing not worthless!

# Spriting
yeah there probably will be some sprites when I get around to it, or if someone else makes them idk it would be nice :)

# Wiki Documentation
will probably have to be added to the xenobio section or something
# Changelog
:cl:  
rscadd: Added plort refinery
wip: This is work in progress  
mapping: put the thingy on the maps where it needs to be.
/:cl:

### DIFF
--- a/code/modules/research/plortmachine.dm
+++ b/code/modules/research/plortmachine.dm
@@ -1,0 +1,57 @@
+/**********************Plort Redemption Machine**************************/
+//Accepts slime "Cores" as they are definetly NOT called, and turns them into stuff. Probably money? maybe goods? idk im writing this thing before i code it.
+
+/obj/machinery/plortrefinery
+	name = "plort redemption machine"
+	desc = "A machine that accepts slime cores, and sells them to the highest bidder. This generates money and research, depending on the rarity."
+	icon = 'icons/obj/machines/mining_machines.dmi' // TODO	
+	icon_state = "ore_redemption" // TODO
+	layer = BELOW_OBJ_LAYER
+	density = TRUE
+	speed_process = TRUE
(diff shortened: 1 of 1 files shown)


ToasterBiome:
research values should probably be on the slime core instead of the machine, nice idea though
will need a cap, either total slime cores or x of each core, maybe diminishing returns

Cartlord:
I like the idea, though if it's just a matter of numbers, it might end up being optimal to just make a fuckton of greys - perhaps add diminishing returns to some of the lower-tier extracts, or a special effect for high-tier ones?

MajManatee:
you over-estimate my abilities with the special effect thing
putting it in the core is probably a good idea, if I can figure that out

code/modules/research/plortmachine.dm
```diff
@@ -0,0 +1,57 @@
+/**********************Plort Redemption Machine**************************/
+//Accepts slime "Cores" as they are definetly NOT called, and turns them into stuff. Probably money? maybe goods? idk im writing this thing before i code it.
+
+/obj/machinery/plortrefinery
+	name = "plort redemption machine"
```
drderp3635, line 5: plorts are from slime rancher, it should be called an extract redemption machine to match our vernacular

VaelophisNyx: more slimerancher = good

ReddicusDragon: I am in agreement with drderp

ReddicusDragon (changes requested):
slime rancher bad

Bibby0110:
Stale for many days, lovely idea, please make it happen :)

PR #16220, opened 2022-10-21
severely powercreeps syndicate mechs because we're charging nuclear operatives 140tc for a durand

TheRyeGuyWhoWillNowDie:
ok basically, i added 2 new mech weapons, one of them is a reskin of the LMG except it fires l6 SAW bullets and a reskin of the scattershot shotgun except it fires syndicate shotgun bullets and also fires 12 pellets per shot
dark gygax and mauler have had their ultra ACs and scattershots replaced with these new weapons, ammo bundles updated so no bamboozle.
dark gygax is currently super awful so i made it use 3/4 the energy of a normal gygax when overload walking, also it no longer fucking breaks itself while overload walking, and i increased its deflect chance to 25 from 15
also the energy tesla relays in the mechs have been replaced with an upgraded version that charges about 5x faster. this may seem like a lot, but gygax burns through this very quickly while overload walking and then has to sit out to recharge, this is mostly to stop nukie mechs from running out of battery and then getting stuck in place because operatives cant get inducers without stealing them from scientists or engineers
:cl:  
tweak: operative mechs are now less bad
/:cl:

### DIFF
--- a/code/game/mecha/combat/gygax.dm
+++ b/code/game/mecha/combat/gygax.dm
@@ -22,7 +22,6 @@
 	deflect_chance = 25
 	armor = list(MELEE = 40, BULLET = 40, LASER = 50, ENERGY = 35, BOMB = 20, BIO = 100, RAD = 100, FIRE = 100, ACID = 100)
 	max_temperature = 35000
-	leg_overload_coeff = 100
 	operation_req_access = list(ACCESS_SYNDICATE)
 	internals_req_access = list(ACCESS_SYNDICATE)
 	wreckage = /obj/structure/mecha_wreckage/gygax/dark
--- a/code/game/mecha/combat/gygax.dm
+++ b/code/game/mecha/combat/gygax.dm
@@ -6,7 +6,7 @@
 	dir_in = 1 //Facing North.
 	max_integrity = 250
 	deflect_chance = 5
-	armor = list(MELEE = 25, BULLET = 20, LASER = 30, ENERGY = 15, BOMB = 0, BIO = 100, RAD = 40, FIRE = 100, ACID = 100)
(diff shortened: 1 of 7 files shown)


00ze-cyclone:
Why did you reduce their rad and bio armor though?
Also you haven’t increased the deflect chance of the Dark Gygax it seem.

TheRyeGuyWhoWillNowDie:
> Why did you reduce their rad and bio armor though? Also you haven’t increased the deflect chance of the Dark Gygax it seem.

forgot to push a commit

Mqiib:
Note that scattershot already did more damage per pellet (16) than a regular shotgun (12) and more than even the syndicate pellets (15.5) and had no falloff (-0.45/tile) iirc. Also if it fires 10 pellets per shot then you need to increase the ammo because it'll subtract 10 shots from the ammo count on firing.

TheRyeGuyWhoWillNowDie:
> Note that scattershot already did more damage per pellet (16) than a regular shotgun (12) and more than even the syndicate pellets (15.5) and had no falloff (-0.45/tile) iirc. Also if it fires 10 pellets per shot then you need to increase the ammo because it'll subtract 10 shots from the ammo count on firing.

the syndicate buckshot seemed more thematic to me, but you're right

00ze-cyclone:
Wait so it’s actually weaker bullets than the default ones? Lmao

Mqiib:
> Wait so it’s actually weaker bullets than the default ones? Lmao

Significantly weaker per-pellet, in fact.

TheRyeGuyWhoWillNowDie:
meh i think it'll probably be fine firing 12 projectiles with falloff per shot, dark gygax is supposed to be close range and mauler has the LMG for people out of range of the shotgun

JamieD1:
The point is they are only supposed to be viable when on discount. Otherwise everyone would always just pick them.
Happy to up the discount rate slightly.

PR #16224, opened 2022-10-22
AsteroidStation minimum player count required 34 -> 40

ToasterBiome:

Asteroid is a huge map and anything under 35 (because 50% of people are ghosted) players makes it feel empty

Also there should never be a situation in the world where a population fit for GaxStation (max 40 players) is also fit for AsteroidStation (min 40 players)

# Changelog

:cl:  
tweak: AsteroidStation requires 40 players minimum instead of 34
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -18,7 +18,7 @@ map yogstation
 endmap
 
 map asteroidstation
-	minplayers 34
+	minplayers 40
 	votable
 endmap
 


MenacingManatee:
Seeing how often there's been a situation where asteroid has ~20 or fewer players due to the way maps are chosen, bumping the count up a bit seems reasonable

MenacingManatee (approved):

JamieD1 (approved):

PR #16232, opened 2022-10-23
i port gun kits from /tg/ except done in a way that makes sense for our codebase

TheRyeGuyWhoWillNowDie:
ok heres the quick rundown:
instead of being able to hit the x10 button for guns on the armory protolathe and being able to instantly replace an entire stolen/lost armory for no effort, you now print kits instead. hit a kit with the weapon it wants and wait a few seconds, and then you get the gun that the kit drops.

stuff that's affected:
teleshield, kit requires a normal riot shield to make
aeg, kit requires a normal energy gun to make
xray laser gun, kit requires a normal laser gun to make
beam rifle, kit requires an xray laser gun to make
tesla revolver, kit requires a laser gun
decloner, kit requires a laser gun
tempgun, kit requires an energy gun
ion carbine, kit requires a laser gun
ebow, kit requires a PKA
hardlight bow, kit requires an energy gun

But why?
Currently, stealing or destroying the armory is a complete waste of time and totally irrelevant, sans the 3 shotguns. Any losses to any weapons security has are replaced in seconds with a click of a button on a machine that spawns roundstart in the armory. This PR aims to help with security officers carrying an entire armory of advanced guns (all mindshield pinned, obviously) in their backpack, which is fucking despised and with good reason. Also makes cargo arming people actually matter because unlike cargo security pulls from an essentially unlimited source, that being the ore silo which will ALWAYS have enough materials to make tons of guns unless there's no shaft miners or the silo was destroyed. And, unlike Cargo, printing guns is instantaneous instead of having to wait for the shuttle. 

firing pins and weapon charge are retained before and after putting a gun through a kit because i cant fucking stand security officers draining all of the diamonds in the ore silo to make firing pins

# Changelog

:cl:  
rscadd: added gunkits to the armory protolathe, which turn into an advanced weapon when hit with their required gun
rscdel: armory protolathes now cant make infinite guns for free, rip
/:cl:

PR #16236, opened 2022-10-23
Make firing pins actually serve their intended purpose

00ze-cyclone:

Alternative to #16232

The mindshield firing pin has always been a problem in that it completely invalidated the purpose of firing pins while being necessary for sci to research before any of the good weapons. You had a strictly better firing pin and it was always unlocked when you had the possibility to print advanced eguns, xray, decloner, temp gun and hardlight bow

So now mindshield firing pins are obtained by crafting them using a regular pin and an mindshield detector which design replace the mindshield firing pin in the protolathe.
The firing pin section in the protolathe has been removed because there was literally only one item that mattered in it, the pins have been moved to be in the weapon category like attachments.

This mean you get 10 mindshield pins with what's at roundstart in the brig, this limit both expanding the arsenal of the armory and making your guns unuseable against you while not just making all the research weapons that aren't the xray gun or aegun worthless because they are very niche compared to lasers and make pins actually matter.

The price of the firing pin crate also has been increased to 6000 credits andonly contain one box of 5 pins now, making crafting a new gun cost 1200 credits after the ressources of the brig have been depleted. The crate is also sec access now so officers can still get pins for a temp gun if there's no warden or HoS, the real roadblock for obtening the guns as a secoff is access to the armory, not to a crate you can force open easily.

# Wiki Documentation

Mindshield pins can no longer be printed, instead you print a mindshield detector. The design is at the same place as the old mindshield pin in the techweb.
Mindshield pin can now be crafted with a firing pin and a mindshield detector, it's in the weapon category of the crafting menu, no tools required.
Firing pin crate now cost 6000 credits, contain only one box of 5 firing pins and is security access.

# Changelog

:cl:  
rscadd: Added mindshield upgrade for firing pins, the mindshield detector
rscdel: Removed printable mindshield pins, instead you print the mindshield detector
tweak: Firing pin crate cost 6000 credits from 2000, contain only 5 firing pins and require basic security access instead of armory access
/:cl:

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -397,12 +397,11 @@
     crate_name = "secway crate"
 
 /datum/supply_pack/security/firingpins
-	name = "Standard Firing Pins Crate"
-	desc = "Upgrade your arsenal with 10 standard firing pins. Requires Security access to open."
-	cost = 2000
-	access_view = ACCESS_ARMORY
-	contains = list(/obj/item/storage/box/firingpins,
-					/obj/item/storage/box/firingpins)
+	name = "Standard Firing Pins Pack"
+	desc = "Upgrade your arsenal with 5 standard firing pins. Requires Security access to open."
+	cost = 6000
+	small_item = TRUE
+	contains = list(/obj/item/storage/box/firingpins)
 	crate_name = "firing pins crate"
 
 /datum/supply_pack/security/justiceinbound
--- a/code/modules/projectiles/pins.dm
+++ b/code/modules/projectiles/pins.dm
@@ -245,3 +245,9 @@
 	if(gun)
 		gun.pin = null
 	return ..()
+
+/obj/item/mindshield_upgrade
+	name = "mindshield detector"
+	desc = "A small circuit compatible with electronic firing pins to make them check for mindshiel implant in their user."
+	icon = 'icons/obj/module.dmi'
+	icon_state = "cyborg_upgrade3"
--- a/code/modules/projectiles/pins.dm
+++ b/code/modules/projectiles/pins.dm
@@ -248,6 +248,6 @@
 
 /obj/item/mindshield_upgrade
 	name = "mindshield detector"
-	desc = "A small circuit compatible with electronic firing pins to make them check for mindshiel implant in their user."
+	desc = "A small circuit compatible with electronic firing pins to make them check for mindshield implant in their user."
 	icon = 'icons/obj/module.dmi'
 	icon_state = "cyborg_upgrade3"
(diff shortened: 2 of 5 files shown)


Cartlord:
Firing pins are destroyed when removed, so people still need to print new pins anyway. I'm not sure what this accomplishes.

00ze-cyclone:
> Firing pins are destroyed when removed, so people still need to print new pins anyway. I'm not sure what this accomplishes.

There’s 10 firing pins in the brig at roundstart so you can equip 10 guns with mindshield pins, those can be new you printed or those you started with like the shotgun of the warden, the egun of captain and HoS and the X-01. The idea is exactly the same as gunkits, not being able to print 666 guns with just a techfab. Except in that case I use a system we already had for this purpose and make it functional instead of adding bullshit that make every R&D gun beside aegun and xray never been made. You can still print guns, you can still put mindshield pins in your, just not to obscene levels without a trip to cargo.

Cartlord:
But... you can still print firing pins, and you can print mindshield detectors. How is that any different than being able to print mindshield firing pins?

00ze-cyclone:
You can't print regular firing pins, only the test range firing pins that am I not even sure have any use now. You can still obtain mindshield firing pins however right now you have 10 firing pins + the virtually infinite amount of mindshield pin you can print once researched, with that PR you'll have 10 pins that you can make into mindshield ones after it's been researched.

Only way to get regular electronic firing pins is cargo.

Marmio64:
i believe that mindshield pins discourage antags from just using station equipment to win rather than antag equipment, which is kind of boring and I disagree with it, so i'd like mindshield pins to remain printable directly, without having to be another upgrade for sec that only cargo can grant.

TheRyeGuyWhoWillNowDie:
> i believe that mindshield pins discourage antags from just using station equipment to win rather than antag equipment, which is kind of boring

maybe the gear security gets for free in an unlimited amount shouldnt be worse than antag gear 🤔

code/modules/projectiles/pins.dm
```diff
@@ -245,3 +245,9 @@
 	if(gun)
 		gun.pin = null
 	return ..()
+
+/obj/item/mindshield_upgrade
+	name = "mindshield detector"
+	desc = "A small circuit compatible with electronic firing pins to make them check for mindshiel implant in their user."
```
TheRyeGuyWhoWillNowDie: you forgot a d

code/modules/cargo/packs.dm
```diff
@@ -397,12 +397,11 @@
     crate_name = "secway crate"
 
 /datum/supply_pack/security/firingpins
-	name = "Standard Firing Pins Crate"
-	desc = "Upgrade your arsenal with 10 standard firing pins. Requires Security access to open."
-	cost = 2000
-	access_view = ACCESS_ARMORY
-	contains = list(/obj/item/storage/box/firingpins,
-					/obj/item/storage/box/firingpins)
+	name = "Standard Firing Pins Pack"
```
ReddicusDragon, line 400: Switch this back and you'll singlehandedly increase your chances of getting this merged. I disagree with upping the price.

ReddicusDragon (changes requested):

JamieD1:
Literally just anti qol

PR #16241, opened 2022-10-24
Extreme rooms let space and heat adapt return to comfortable temps

Moltijoe:
Currently, space adapt prevents people from cooling down if too hot
and heat adapt prevents people from heating up if too cold
This is good to balance out the power it gives you

The problem comes from always needing to use a shower or chems to fix your temp
This will allow extreme room temps to override the "no passive heating" or "no passive cooling"

**This is will make it so:**
someone with heat adapt won't freeze to death in a post-plasma fire room
And someone with space adapt won't burn to death in space

but they will still need to worry about being too warm or cold under normal circumstances

:cl:  
tweak: being in a super hot room will let heat adapt people warm up
tweak: being in a super cool room will let space adapt people cool down
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1945,7 +1945,7 @@ GLOBAL_LIST_EMPTY(mentor_races)
 		if(ismovable(human_loc))
 			var/atom/movable/occupied_space = human_loc
 			thermal_protection *= (1 - occupied_space.contents_thermal_insulation)
-		if(!HAS_TRAIT(H, TRAIT_NO_PASSIVE_HEATING) || loc_temp > BODYTEMP_NORMAL+200) //if the room is very hot, it will allow heat adapt people to warm up
+		if(!HAS_TRAIT(H, TRAIT_NO_PASSIVE_HEATING) || loc_temp > BODYTEMP_NORMAL+1000) //if the room is very hot, it will allow heat adapt people to warm up
 			if(H.bodytemperature < BODYTEMP_NORMAL) //and we're cold, insulation enhances our ability to retain body heat but reduces the heat we get from the environment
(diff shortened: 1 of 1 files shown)


alexkar598:
As the author of the pr that split them, this is very intentional. It's meant to add counter play to people having space adapt (temp gun is a big one)

Moltijoe:
> As the author of the pr that split them, this is very intentional. It's meant to add counter play to people having space adapt (temp gun is a big one)

they'll still cook alive if they're not in space, this will only change stuff regarding extreme ROOM temps

alexkar598:
It ain't that hard to get yourself in space if need be, considering this was meant to be a two fold effect:
1) Annoy people into not having space adapt be an upgrade that was always good
2) Allow for hard counters
this PR breaks both of those intentions, space is much easier to find than a shower or other temperature affecting methods breaking 1 as its no longer a major downside and breaking 2 as space can be created very easily negating the hard counter easily.

Moltijoe:
> It ain't that hard to get yourself in space if need be, considering this was meant to be a two fold effect:
> 
> 1. Annoy people into not having space adapt be an upgrade that was always good
> 2. Allow for hard counters
>    this PR breaks both of those intentions, space is much easier to find than a shower or other temperature affecting methods breaking 1 as its no longer a major downside and breaking 2 as space can be created very easily negating the hard counter easily.

that's fair it just seems dumb that space adaptation is often a death sentence and worse than not having it if the station gets plasmaflooded as it becomes impossible to get anything to cool down. for 40 stability it shouldn't basically guarantee someone's death

Moltijoe:
would you consider it to be better to just add more ways for someone to cool down/heat up without needing to get your hands on special chems or drinks?

Perhaps drinking water letting people return to comfortable temperatures?

PR #16242, opened 2022-10-24
Drinking water will slowly equalize your body temp

Moltijoe:
Currently there only ways to heat up or cool down with heat/space adapt is to use a shower or a few limited drinks/chems
This will allow drinking water to do the same, albeit much slower

Mainly to combat plasmafloods from giving a death sentence to all people with space adapt without removing it's main weakness

:cl:  
rscadd: drinking water will equalize your body temp
/:cl:


iamthesilly:
w a t e r

JamieD1 (approved):

PR #16243, opened 2022-10-24
PLASMAFLOOD POOL PARTY

Moltijoe:
If the pool is set to be at a comfortable temperature it will bring the body of people in the pool to be closer to temperature of the pool
This will let people use it to stay warm if the room is spaced, or cool if the room is plasmaflooded

Also makes the pool filter indestructible because it can't be made by people, only placed by mappers and if it breaks, the pool functionality does too

:cl:  
tweak: a comfy pool will give you a good body temperature
/:cl:

### DIFF
--- a/code/modules/pool/pool.dm
+++ b/code/modules/pool/pool.dm
@@ -302,6 +302,10 @@ GLOBAL_LIST_EMPTY(pool_filters)
 					if(!HAS_TRAIT(C, TRAIT_RESISTHEAT))
 						C.adjustFireLoss(5)
 					to_chat(M, "<span class='danger'>The water is searing hot!</span>")
+				else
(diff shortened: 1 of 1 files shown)


Marmio64:
dont think it should be totally indestructible, its not important enough to really warrant that. Just make it very tough but still breakable.

Moltijoe:
> dont think it should be totally indestructible, its not important enough to really warrant that. Just make it very tough but still breakable.

pool functionality breaks entirely if the pool filter gets destroyed, since pools can't be broken, the filter shouldn't be either

CMOisLing:
Make it super effective if you rest so you can hide from the fire like an action movie.

JamieD1 (approved):

PR #16259, opened 2022-10-25
removes the 10 day timelock for zombies completely

SlasherX01:
To be able to be a zombie you need to wait about 10 days (if you are new) i feel that this is a bit too long to be a zombie as it isn't a very difficult antagonist to play, for comparison bloodsuckers only require 3 days. It doesn't make sense for such a easy antag to have such a high timelock

# Changelog

:cl:  
rscdel: the 10 day zombie timelock has been completely removed
/:cl:

### DIFF
--- a/code/game/gamemodes/zombie/zombie.dm
+++ b/code/game/gamemodes/zombie/zombie.dm
@@ -17,7 +17,6 @@ GLOBAL_LIST_EMPTY(zombies)
 	required_players = 40
 	required_enemies = 3
 	recommended_enemies = 3
-	enemy_minimum_age = 14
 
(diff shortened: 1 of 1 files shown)


Bibby0110:
This sets it to 7, also it was originally 14?? where are you getting 10 from
also bloodsuckers doesn't have a 3 day minimum???

SlasherX01:
> This sets it to 7, also it was originally 14?? where are you getting 10 from

either way 14 is way too much, how can i lower it completely?

Rdrazga:
This code is for the old never completed zombie game mode is it not? Not the mid round zombie antag.

PR #16277, opened 2022-10-27
Makes aggressive grabs significantly less bad for the game

ynot01:

Can now use items while aggressively grabbed (no longer counts as restrained)

Can shove your captor, it will not release you but if you hit them into a wall they'll drop their active item like normal

Can grab your captor, if you manage an aggressive grab they will release their grip on you, which should result in interesting grab trades, though this results in not being able to neck grab someone who is intent on grabbing you to get out. You will now need to stun someone to reliably neck grab them.

Being thrown into a wall or person is now a knockdown (slip) instead of a paralyze (hardstun)

## Goals

Makes aggressive grab meta a lot more possible to fight against while keeping it strong for disarms and debilitating your opponent

Deletes wall chainstunning

Throwing bodies at people is still effective for sake of disarm and slow, but no longer hardstun

Martial arts have been a problem child for a long while since they have instant aggressive grabs. The reason *why* they have these is because hitting your opponent with combos is a nigh pointless task if you don't have a way to keep them still. This created a very bad balancing issue where martial arts users would just grab and use their effective RNG stun to either chainstun or simply keep hold and beat the person they were holding to death because they couldn't roll a breakout.

This change keeps that effectiveness of keeping your opponent still to land combos, but allows them to fight back with whatever they want. You still have the major advantage, forcing them into melee range, but it is no longer an instant free game-ender.

# Changelog

:cl:  
tweak: Players can now use items while under the effects of an aggressive grab instead of being hardstunned
tweak: Crashing into walls and other players is now a knockdown instead of a hardstun
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -198,7 +198,7 @@
 		changeNext_move(CLICK_CD_RANGE)
 
 /mob/living/carbon/restrained(ignore_grab)
-	. = (handcuffed || (!ignore_grab && pulledby && pulledby.grab_state >= GRAB_AGGRESSIVE))
+	. = (handcuffed || (!ignore_grab && pulledby && pulledby.grab_state > GRAB_AGGRESSIVE))
 
 /mob/living/carbon/proc/canBeHandcuffed()
 	return 0
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -198,7 +198,7 @@
 		changeNext_move(CLICK_CD_RANGE)
 
 /mob/living/carbon/restrained(ignore_grab)
-	. = (handcuffed || (!ignore_grab && pulledby && pulledby.grab_state > GRAB_AGGRESSIVE))
+	. = (handcuffed || (!ignore_grab && pulledby && pulledby.grab_state >= GRAB_NECK))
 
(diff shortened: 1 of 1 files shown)


Marmio64:
Seems too aggressive for a nerf without any testing, and seems to have nerfed aggressive grabs mainly with martial arts in mind, not really caring about doing it normally where it's harder to pull off, and not nearly as useful now.

neeshacark:
this should definitely be a testmerge first yeah

JamieD1 (approved):

Firewolf34:
Totally makes sense to nerf the martial arts stun, but I feel disarm/grab really needs a buff - unarmed, non-martial combat has become so difficult to succeed with now that you need to be quite skilled to attack without a weapon. Raises the skill barrier a lot, which will actually just make things harder on less skilled players if the pros get excellent at it.  

There should, imho, still be some mechanism to turn the tables on some guy running at you with a fireaxe even if you're 100% unarmed. If it's movement or reflexes or accuracy even a limb targetting method, there needs to be some basic (but difficult) counter. Wall-stun was the situational counter, now it's just wall slip... and the issue with that is: 1, very hard to reliably wallstun already, 2, wallstun time very short so hard to capitalize on it. Now they'll be both moving away from you *and* using an item against you while you're trying to pull it off. 

Maybe we need watered-down martial style combat combos just as default?

prisoner12349:
basically cant choke people anymore

PR #16283, opened 2022-10-27
hierophant drops loot when gibbed now

LazennG:

gibbing the hierophant with something like cuffs before it was finished with its death animation used to make it drop nothing but now it does

# Changelog
:cl:  
bugfix: fixed hiero not dropping loot when it should 
/:cl:


ShoziOffical:
Doesn't it drop loot when dead?

JamieD1 (approved):

PR #16285, opened 2022-10-28
Gives cryoxadone an OD effect, prevents Cryotube OD

Moltijoe:
Gives cryoxadone an overdose affect to prevent chugging a bunch of it and going out into space like everything's fine

blame RG4 they asked me to do this

:cl: 
tweak: gives cryoxadone an overdose affect
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
@@ -246,6 +246,9 @@
 					air1.adjust_moles(/datum/gas/pluoxium, -max(0,air1.get_moles(/datum/gas/pluoxium) - 0.5 / efficiency))
 				else 
 					air1.adjust_moles(/datum/gas/oxygen, -max(0,air1.get_moles(/datum/gas/oxygen) - 2 / efficiency)) //Let's use gas for this
+				if(occupant.reagents.get_reagent_amount(/datum/reagent/medicine/cryoxadone) >= 100) //prevent cryoxadone overdose
+					occupant.reagents.del_reagent(/datum/reagent/medicine/cryoxadone)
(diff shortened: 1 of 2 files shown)


Hislyna:
I ded

Hislyna:
How will this effect cryo cells, seems like youre gonna OD everyone who uses one.

Moltijoe:
not a clue, just following what rg4 told me would be good, i could always just make cryo cells not constantly pump more cryoxadone into people

TheBombMakerJL:
I am coming back from the dead to say that this is not a good idea, no I will not elaborate @Mqiib can do it for me

RG4ORDR:
There is no other medicine one can chug with zero side-effects and multiple ways to actively trigger. 
I'm all for ghetto cyro such as using showers, using heat adapt and lowering your body temp, ice cups,frostoil, or even space-walks. However someone chugging 600 units is insane. Just with my own experimentation you can walk around in space indefinitely with damage capping at 4 Brute, 1.5 Burn, and 3 Oxygen(if you aren't wearing internals). 

It's not an instant fix of course, but paired other means of basic healing it can become powerful, or in other cases an extreme annoyance. Most other healing meds have OD limits,caps, or cause damage to organs, or only heal specific types of damage such as toxins. Where as cryox can heal wounds, toxins, clone, brute, burn. Unless there's a cap on how much it can heal, some form of organ damage,tube only(I like ghetto cryo so no) or OD limit I just find it too strong. Expressively because one player doing the same gimmick of it a morbillion times.

Maia-J:
Just ban Drake already

Moltijoe:
> Just ban Drake already

Hes not doing anything bannable as far as I know, most rounds he abuses stuff its either as an antag or in response to a threat. Everything else is within the borders of a chemists job.

JamieD1 (approved):

PR #16292, opened 2022-10-28
makes the syndicate hardlight bow 33% cheaper because the mini ebow still exists lmao

Skrem7:

Mini ebow: 10 TC, recharges quickly, guaranteed knockdown and toxin and stamina which causes instant slowdown, can be held in one hand and doesn't have to be drawn.

Hardlight bow: 12 TC, requires both hands, does okay damage, has a neat x ray mode IF you buy thermals which means you can harass the heck out of sec (that rhymes) but it's still clunky to use and takes a while to fire.

Solution: Never nerf the mini ebow (it will never be merged) and instead make the hardlight bow cheaper (8 TC instead of 12)

# Wiki Documentation

Cost will have to be updated on Syndicate Items and Guide to Combat

# Changelog

:cl:  
tweak: Hardlight bow for Syndicate is 33% cheaper
/:cl:


ynot01:
reduce changelog by at least 12 characters

ReddicusDragon (approved):

Bibby0110 (approved):

PR #16294, opened 2022-10-29
Bag of Storing

Moltijoe:
Some jobs are pretty tight for items they need to carry for their job, a lot of them used the BoH to remove that strain
This adds a "downgrade" to the bag of holding that doesn't require an anomaly core, but also doesn't allow for larger items to be put inside it.

:cl:  
rscadd: Adds storage upgrade that doesn't "break balance"
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -37,6 +37,20 @@
 	var/datum/component/storage/STR = GetComponent(/datum/component/storage)
 	STR.max_combined_w_class = 12
 
+/obj/item/storage/backpack/storing
+	name = "bag of storing"
+	desc = "A small backpack that opens into a localized pocket of bluespace."
+	icon_state = "holdingpack"
+	item_state = "holdingpack"
+	resistance_flags = FIRE_PROOF
+	item_flags = NO_MAT_REDEMPTION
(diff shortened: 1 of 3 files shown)


cuackles:
legally distinct BOH

TheRyeGuyWhoWillNowDie:
this will break balance as long as guns are normal sized because this just soft reverts the BoH removal

forhiti:
Soft nerf is better than just making the boh a complete pain in the ass to obtain (hur dur farm the sm for cores).
I think this is a better solution.

Bibby0110:
> this will break balance as long as guns are normal sized because this just soft reverts the BoH removal

This

JamieD1:
> Some jobs are pretty tight for items they need to carry for their job, a lot of them used the BoH to remove that strain This adds a "downgrade" to the bag of holding that doesn't require an anomaly core, but also doesn't allow for larger items to be put inside it.
> 
> 🆑 rscadd: Adds storage upgrade that doesn't "break balance" /🆑

Relying on a BoH to do your job is not a viable option, we should remove these constraints then.

Moltijoe:
> Relying on a BoH to do your job is not a viable option, we should remove these constraints then.

sounds good, i'll see what else i can do then

PR #16335, opened 2022-11-02
e

warface1234455:
.


ShoziOffical:
actually there is a zombie organ that does grab ghost but theres a bug that prevents it

PR #16350, opened 2022-11-03
ruins lavaland teleport cubes 

LazennG:

not a fan of people teleporting to a safe area instantly at the first sign of danger so i'm making cubes take 3 seconds to use when you're not using one on lavaland

# Wiki Documentation

include the 3 second channel time for using a cube while off lavaland in this bit where the cubes are https://wiki.yogstation.net/wiki/Shaft_Miner#Necropolis_Chest_Loot

# Changelog

:cl:  
tweak: 3 second channel to use cube off lavaland  
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -389,6 +389,10 @@ GLOBAL_LIST_EMPTY(bloodmen_list)
 		return
 	if(teleporting)
 		return
+	if(!is_mining_level(current_location.z))
+		user.visible_message(span_danger("[user] starts fiddling with [src]!"))
+		if(!do_after(user, 3 SECONDS, user))
+			return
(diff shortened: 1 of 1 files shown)


Rdrazga:
I fully agree with this pr given how stupid that one IAA round a while ago was to watch. I think the final count was the 52 uses of the teleport cube to instantly escape a situation that would have been a loss.

JamieD1:
How about we make cubes only usable on lavaland?

LazennG:
> How about we make cubes only usable on lavaland?

didn't wanna do this because i felt it was too much.

JamieD1:
> > How about we make cubes only usable on lavaland?
> 
> didn't wanna do this because i felt it was too much.

Alright, but if its still an issue, make sure you do that.

JamieD1 (approved):

PR #16351, opened 2022-11-04
Fixes NTSL scripts not working

BurgerLUA:
This should theoretically fix tcomms by allowing more characters to be used. The proc ``reject_bad_text()`` was edited to accept more args so that characters essential to making an NTSL script (such as tab, newlines, carriage returns, < and >, / and \) can be accepted.

You can cross check the ASCII codes here: https://www.w3schools.com/charsets/ref_html_ascii.asp

# Changelog

:cl:  BurgerBB
bugfix: Fixes NTSL script sanitization being too strict.
/:cl:

### DIFF
--- a/yogstation/code/modules/scripting/IDE.dm
+++ b/yogstation/code/modules/scripting/IDE.dm
@@ -52,7 +52,7 @@
 					if(!compileerrors)
 						src << output("<b>NTSL.exe Error</b>", "tcserror")
 						src << output("<font color = red>\t><b>A fatal error has occured. Please contact your local network adminstrator.</b></font color>", "tcserror")
-					else if(istext(compileerrors)
+					else if(istext(compileerrors))
 						src << output("<b>NTSL.exe Error</b>", "tcserror")
 						src << output("<font color = red>\t><b>[compileerrors]</b></font color>", "tcserror")
 					else if(length(compileerrors))
--- a/yogstation/code/modules/scripting/IDE.dm
+++ b/yogstation/code/modules/scripting/IDE.dm
@@ -49,7 +49,10 @@
 					if(!telecomms_check(mob))
 						return
 
-					if(compileerrors.len)
+					if(!compileerrors)
+						src << output("<b>Memory Error:</b> Script ran out of memory. Please ensure that the script does not have a large file size and contains no special characters.", "tcserror")
+
+					else if(compileerrors.len)
 						src << output("<b>Compile Errors</b>", "tcserror")
 						var/i = 1
 						for(var/scriptError/e in compileerrors)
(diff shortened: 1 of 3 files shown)


Mqiib:
You can build in visual studio code using ctrl+shift+B. If you don't have that, I do recommend it, it's pretty good.

ynot01:
if Build.bat doesn't work then vsc wont fix it

BurgerLUA:
Its probably something on my end that's fucking up. I'll look more into it tomorrow.

BurgerLUA:
Tested it, and it works. I also removed a runtime detected and changed the message displayed when tcomms runs out of memory.

Valid Script:
image

Invalid Script:
image

Bibby0110:
The shown pictures aren’t current and there’s an unfortunate logic error in the new script (I think)
I’m not at home so I can’t fully investigate. DNM until resolved

yogstation/code/modules/scripting/IDE.dm
```diff
@@ -49,7 +49,10 @@
 					if(!telecomms_check(mob))
 						return
 
-					if(compileerrors.len)
+					if(!compileerrors)
+						src << output("<b>Memory Error:</b> Script ran out of memory. Please ensure that the script does not have a large file size and contains no special characters.", "tcserror")
```
BurgerLUA: What you're referring to is probably this. compilerrors returns null if the script contains bad characters. Checking null.len causes a runtime as seen in one of the screenshots above. if a script is successful, compileerrors will return list().

BurgerLUA:
Yeah I tested it again and it works. All 3 cases work (Case 1: script contains errors, Case 2: script does not contain errors, Case 3: script contains bad text)

yogstation/code/modules/scripting/IDE.dm
Bibby0110: It also returns null if it fails for any other reason (recharging, banned)

BurgerLUA: Yep. The "Out of Memory" still seems applicable for both those situations.

Bibby0110: It isn’t

BurgerLUA: I really don't mean to be rude but I think you need to be more clear in what you want and what is actually going on so it would save me more time and also save you more time.

BurgerLUA:
I made it so that the compile errors proc can return a text, which will just display the text as an error. Returning a list filled with compileerror datums will run normal behavior. Returning a blank list means that everything is fine. Returning null means something  went wrong with the code.

Bibby0110 (approved):

BurgerLUA:
Ah, seems like I forgot a parenthesis when adding the new changes, one sec

ReddicusDragon:
@TheGamerdk please remove do not merge

PR #16361, opened 2022-11-04
Rework/buff flashbangs

Simplehorror:

Flashbangs now slow down initially for 2 seconds and apply a white screen depending on the distance for a variable amount of seconds. They no longer knock down.
I believe that flashbangs should be more of a utility weapon and in their current state are not that great, a short knockdown is NOT good for defending against a crowd which is the whole point of a flashbang - crowd control! So I made it so that they are made to dispel crowds if there's revs outside of security, for example, but they're still very useful as a utility but you're just not going to make arrests with them alone.
# Wiki Documentation
Flashbangs now slow down initially for 2 seconds and apply a white screen depending on the distance for a variable amount of seconds. They no longer knock down.
# Changelog
:cl:  
tweak: Flashbangs now slow down initially for 2 seconds and apply a white screen depending on the distance for a variable amount of seconds. They no longer knock down.
/:cl:


Moltijoe:
jamie has specifically said in the past that flashbangs are not to be changed, even closed a PR about flashes because he assumed it was about flashbangs

Simplehorror:
> jamie has specifically said in the past that flashbangs are not to be changed, even closed a PR about flashes because he assumed it was about flashbangs

I will look into other AOE options for security if this PR is closed

TheRyeGuyWhoWillNowDie:
how is removing the stun a buff

Simplehorror:
you do understand how powerful a 15 second blind is right

PR #16367, opened 2022-11-05
lag compensation

Simplehorror:
Adds lag compensation to melee attacks, maybe helps feel the game a bit more responsive with high ping
# Changelog
:cl:  
rscadd: Lag compensation for melee attacks
/:cl:
wip

### DIFF
--- a/code/__DEFINES/_tick.dm
+++ b/code/__DEFINES/_tick.dm
@@ -26,3 +26,5 @@
 #define TICK_CHECK_HIGH_PRIORITY ( TICK_USAGE > 95 )
 /// runs stoplag if tick_usage is above 95, for high priority usage
 #define CHECK_TICK_HIGH_PRIORITY ( TICK_CHECK_HIGH_PRIORITY? stoplag() : 0 )
+/// Gets how many ticks have ran currently
+#define TICK_TIME (world.time / world.tick_lag)
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -79,8 +79,7 @@
 		return
 	var/end_comp = FALSE
 	if (client)
-		end_comp = TRUE
-		SSlag_compensation.begin_lag_compensation(client.avgping)
+		end_comp = SSlag_compensation.begin_lag_compensation(client.avgping)
 	var/list/modifiers = params2list(params)
 	if(modifiers["shift"] && modifiers["middle"])
 		ShiftMiddleClickOn(A)
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -77,57 +77,61 @@
 
 	if(SEND_SIGNAL(src, COMSIG_MOB_CLICKON, A, params) & COMSIG_MOB_CANCEL_CLICKON)
 		return
-
+	var/end_comp = FALSE
+	if (client)
+		end_comp = TRUE
+		SSlag_compensation.begin_lag_compensation(client.avgping)
 	var/list/modifiers = params2list(params)
 	if(modifiers["shift"] && modifiers["middle"])
 		ShiftMiddleClickOn(A)
-		return
+		goto end;
 	if(modifiers["shift"] && modifiers["ctrl"])
 		CtrlShiftClickOn(A)
-		return
+		goto end;
 	if(modifiers["middle"])
 		MiddleClickOn(A)
-		return
+		goto end;
 	if(modifiers["shift"])
 		ShiftClickOn(A)
-		return
+		goto end;
 	if(modifiers["alt"]) // alt and alt-gr (rightalt)
 		AltClickOn(A)
-		return
+		goto end;
 	if(modifiers["ctrl"])
 		CtrlClickOn(A)
-		return
+		goto end;
 
 	if(incapacitated(ignore_restraints = 1))
-		return
+		goto end;
 
 	face_atom(A)
 
 	if(next_move > world.time) // in the year 2000...
-		return
+		goto end;
 
 	if(!modifiers["catcher"] && A.IsObscured())
-		return
+		goto end;
 
 	if(ismecha(loc))
 		var/obj/mecha/M = loc
-		return M.click_action(A,src,params)
+		. = M.click_action(A,src,params)
+		goto end;
 
 	if(restrained())
 		changeNext_move(CLICK_CD_HANDCUFFED)   //Doing shit in cuffs shall be vey slow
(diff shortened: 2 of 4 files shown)


Bibby0110:
Is this a port of something?

Simplehorror:
> Is this a port of something?

no

Maia-J:
**This man!!!**
**Ided tag him!**

In all honesty, if this is done right, it would increase the enjoyment for quite a lot of people. 
Of course, the Americans with their perfect 5ms will cry about this, but don't listen to them.

Simplehorror:
i live in china so this PR is very important to me

Bibby0110:
> > Is this a port of something?
> 
> no

You’re going to have to explain in technical detail what this does, and how it does it.

Simplehorror:
> > > Is this a port of something?
> > 
> > 
> > no
> 
> You’re going to have to explain in technical detail what this does, and how it does it.

it will save the location history of all mobs in the world for 100 ticks in a rolling queue, when BeginLagCompensation is called it will rewind all mobs temporarily to their previous history and then restore them after EndLagCompensation is called

Simplehorror:
actually not sure if updating it and restoring it in the same proc would cause visual updates but if it does i'll just add a hitbox xyz that's used for distance checking in clicks

ToasterBiome:
> > > > Is this a port of something?
> > > 
> > > 
> > > no
> > 
> > 
> > You’re going to have to explain in technical detail what this does, and how it does it.
> 
> it will save the location history of all mobs in the world for 100 ticks in a rolling queue, when BeginLagCompensation is called it will rewind all mobs temporarily to their previous history and then restore them after EndLagCompensation is called



> > > > Is this a port of something?
> > > 
> > > 
> > > no
> > 
> > 
> > You’re going to have to explain in technical detail what this does, and how it does it.
> 
> it will save the location history of all mobs in the world for 100 ticks in a rolling queue, when BeginLagCompensation is called it will rewind all mobs temporarily to their previous history and then restore them after EndLagCompensation is called

What are the instances where BeginLagCompensation are called and under what conditions, who does it effect, etc.

Simplehorror:
> > > > > Is this a port of something?
> > > > 
> > > > 
> > > > no
> > > 
> > > 
> > > You’re going to have to explain in technical detail what this does, and how it does it.
> > 
> > 
> > it will save the location history of all mobs in the world for 100 ticks in a rolling queue, when BeginLagCompensation is called it will rewind all mobs temporarily to their previous history and then restore them after EndLagCompensation is called
> 
> > > > > Is this a port of something?
> > > > 
> > > > 
> > > > no
> > > 
> > > 
> > > You’re going to have to explain in technical detail what this does, and how it does it.
> > 
> > 
> > it will save the location history of all mobs in the world for 100 ticks in a rolling queue, when BeginLagCompensation is called it will rewind all mobs temporarily to their previous history and then restore them after EndLagCompensation is called
> 
> What are the instances where BeginLagCompensation are called and under what conditions, who does it effect, etc.

it's called in ClickOn if the calling mob has a client, effects every mob on the map

ToasterBiome:
> 

This just sounds like it will wildly lag everyone on the server if someone has a shitty connection

Simplehorror:
no it wouldn't, it's virtual, it should only move them for hitreg detection and no visual updates would be sent

ToasterBiome:
I do not think this will
1) work
2) have the intended effect

I think this will just make hit reg against lagging players advantageous to them.

Most importantly, this is shitcode. No goto, etc. Storing 10 seconds of locs for every mob (even every client) is bad and will cause far more lag at highpop than helping any amount of lagging players.

Simplehorror:
> I do not think this will
> 
>     1. work
> 
>     2. have the intended effect
> 
> 
> I think this will just make hit reg against lagging players advantageous to them.
> 
> Most importantly, this is shitcode. No goto, etc. Storing 10 seconds of locs for every mob (even every client) is bad and will cause far more lag at highpop than helping any amount of lagging players.

please show me a way to do what i did with goto without a massive amount of code duplication and it's not 10 seconds, yogs tickrate is 20. and it wouldn't cause CPU pressure but memory pressure and storing 4 values is not a lot of memory, and you can't say "etc" after stating one example
>I think this will just make hit reg against lagging players advantageous to them.

that's the point? to make it easier for EU players to actually play the game?

Bibby0110:
> no it wouldn't, it's virtual, it should only move them for hitreg detection and no visual updates would be sent

That’s not how that works. You’re doing a massive amount of processing and moving every time someone clicks anything, in addition to doing costly loops 10 times per second.
I don’t see how this wouldn’t have a massive performance impact.
In addition you’re opening yourself up to numerous issues by having a list stored like this. You’re hindering garbage collection of every single mob for approx 10 seconds, potentially causing a flurry of hard deletions and a list sprinkled with random and corrupted data. The same goes for the turfs (and objects) you’re storing.

While yes this is approximately how lag compensation is done in a simple FPS shooter those games have the advantage of being able to tailor their networking code to support all of this, while we’re stuck with BYOND and no way to do the same.

For some of the reasons above, the fact that you’re a new coder around here, and the fact that this would probably have been attempted sometime during the last decade and a half of SS13 development, I remain highly skeptical that this is even possible.

Feel free to correct me if I’ve misunderstood something, but I ask that you do this in a new PR that includes a thorough technical write up and heavily commented code.

Simplehorror:
there's not a massive amount of processing, the move proc is NOT called when i set their xyz so there's almost 0 CPU time used. 
the GC issue is trivial to solve since we can just use the ID of the mob instead of a reference to it. you're acting like this is something impossible when it's been trivially done in every game since the 90s and even is in other BYOND games

Simplehorror:
in fact BYOND gives you all the information you need to do lag compensation in the Click function but nobody uses it @TheGamerdk

Bibby0110:
Provide detailed technical rundown and actual tangible data showing performance improvement and I’ll believe you.

You are absolutely spending a lot of CPU time on all these loops.

Bibby0110:
> in fact BYOND gives you all the information you need to do lag compensation in the Click function but nobody uses it @TheGamerdk

Provide some examples of other BYOND implementations then

Simplehorror:
all other BYOND games are closed source

Simplehorror:
don't forget that this is a background subsystem

CMOisLing:
See if someone who lives in another continent can host so you can preform a test to prove this works and doesnt cause to much extra processing time with a lot of mobs

PR #16376, opened 2022-11-06
Re-enables the "Oh, Hi Daniel" shuttle

Simplehorror:
Believe it or not, this shuttle is actually one of the safest in the game. It has no windows which are the easiest to break and it is quite tight which means traitors can't kite 24/7 like they can on a shuttle. The reason people remember it being such a hellhole is because back in the day you couldn't rest, which means this shuttle could be filled quite easily which gave it a notorious reputation for being a grief shuttle, which it no longer is. I've increased the credit cost significantly since I assume people will want to buy it at random, however.

# Wiki Documentation
the "Oh, Hi Daniel" shuttle is enabled once more.
# Changelog

:cl:  
tweak: Re-enables the "Oh, Hi Daniel" shuttle
/:cl:

### DIFF
--- a/code/datums/shuttles.dm
+++ b/code/datums/shuttles.dm
@@ -373,8 +373,7 @@
 	description = "How was space work today? Oh, pretty good. We got a new space station and the company will make a lot of money. What space station? I cannot tell you; it's space confidential. \
 	Aw, come space on. Why not? No, I can't. Anyway, how is your space roleplay life?"
 	admin_notes = "Tiny, with a single airlock and wooden walls. What could go wrong?"
-	credit_cost = 7500
-	emag_buy = TRUE
+	credit_cost = 30000
 	movement_force = list("KNOCKDOWN" = 3, "THROW" = 2)
 
 /datum/map_template/shuttle/emergency/goon


Bibby0110 (approved):

Marmio64:
wooden walls can be cut apart with any sharp tool, it only has a single airlock, no medbay, no brig, and no command area. This is pretty onpar if not worse than scrapheap challenge but also costs you money, while scrapheap is still an emag buy and gives you credits for getting it. This shuttle should stay emag buy only, its a meme shuttle that is really bad asides from being 'cozy', and any other normal buy shuttle would be superior.

Simplehorror:
> wooden walls can be cut apart with any sharp tool, it only has a single airlock, no medbay, no brig, and no command area. This is pretty onpar if not worse than scrapheap challenge but also costs you money, while scrapheap is still an emag buy and gives you credits for getting it. This shuttle should stay emag buy only, its a meme shuttle that is really bad asides from being 'cozy', and any other normal buy shuttle would be superior.

ok, let me add these functionalities

Moja10:
Actually Daniel Shuttle was in the game when you could rest + crawl in order to fit everyone. It was just a bait shuttle for maxcaps, and no medbay/Command area mentioned above. While I do miss it (sometimes), it did cause a lot of issues. Don't know which is better, never seeing it ever since it currently needs an emag to get, or seeing it too much and it causing issues by its design.

00ze-cyclone:
Oh hi daniel shuttle is a fever dream, it’s not even a shuttle it’s the house of someone that committed suicide that is just going around in space for some reason. It should remain emag only so it can remain that complete fever dream, if it was not emag it would have to be made into a proper shuttle which would just be lame.

JamieD1:
Too small and noone likes it when its bought

PR #16384, opened 2022-11-06
Clockwork cultists now appear on the manifest and have their metatells stored in their bag instead of being noob bait

Simplehorror:

Clockwork cultists now appear on the manifest and have their metatells stored in their bag instead of being noob bait

# Changelog

:cl:  
tweak: Clockwork cultists now appear on the manifest and have their metatells stored in their bag instead of being noob bait
/:cl:

### DIFF
--- a/code/game/gamemodes/clock_cult/clock_cult.dm
+++ b/code/game/gamemodes/clock_cult/clock_cult.dm
@@ -191,6 +191,8 @@ Credit where due:
 		greet_servant(L)
 		equip_servant(L)
 		add_servant_of_ratvar(L, TRUE)
+		GLOB.data_core.manifest_inject(L)
+
 	var/list/cog_spawns = GLOB.servant_spawns_scarabs.Copy()
 	for(var/turf/T in cog_spawns)
 		new /obj/item/clockwork/construct_chassis/cogscarab(T)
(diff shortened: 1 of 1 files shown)


Bibby0110:
Why are they on the manifest?

Simplehorror:
> Why are they on the manifest?

because every metahunter will just see some assistant with insulated gloves and a belt and check their manifest and surprise surprise they're not there and then they're valid because they're not crew

have you played before?

ShoziOffical:
that's admin issue

Bibby0110:
> > Why are they on the manifest?
> 
> because every metahunter will just see some assistant with insulated gloves and a belt and check their manifest and surprise surprise they're not there and then they're valid because they're not crew
> 
> have you played before?

That's an admin issue, in addition you're not valid due to not being on the manifest.

Sniblet:
However you can be arrested, searched, and that will make you valid

Bibby0110:
> However you can be arrested, searched, and that will make you valid

Yes but that goes for every antag

Sniblet:
However, most antags are on the manifest
Question answered

Bibby0110 (changes requested):
Please change your title and changelog to be less misleading. This is not a bugfix

JamieD1 (approved):

PR #16392, opened 2022-11-06
Paperwork Printing Program for Modular Computers

ReddicusDragon:
Modular computers can now print paperwork. Gone are the days of clipboard and copier.

I DID TEST THIS, Special Thanks to Chubbygummibear, baimou, Bibby, and Chester
# Changelog

:cl:ReddsNotDead
rscadd: Added paperwork printing to modular devices. To use, download the program.
/:cl:

### DIFF
--- a/code/modules/modular_computers/file_system/programs/paperworkprinter.dm
+++ b/code/modules/modular_computers/file_system/programs/paperworkprinter.dm
@@ -15,9 +15,9 @@
 	var/list/papers = list()
 	//filter out paperwork we're not allowed to have
 	for(var/R in subtypesof(/obj/item/paper/paperwork))
-		var/obj/item/paper/paperwork/P = new R
-		if(P.printable)
-			papers += list(list("name" = P.name, "id" = P.id))
+		var/obj/item/paper/paperwork/P = R
+		if(initial(P.printable))
+			papers += list(list("name" = initial(P.name), "id" = initial(P.id)))
 	var/obj/item/computer_hardware/printer/printer
 	if(computer)
 		printer = computer.all_components[MC_PRINT]
@@ -33,10 +33,12 @@
 	//this variable stores the object of which we're actually going to print
 	if(action == "PRG_print")
 		var/paperworkToPick = params["paperworkID"]
-		for(var/obj/item/paper/paperwork/P in subtypesof(/obj/item/paper/paperwork))
-			if(P.id==paperworkToPick)
+		for(var/R in subtypesof(/obj/item/paper/paperwork))
+			var/obj/item/paper/paperwork/P = R
+			if(initial(P.id)==paperworkToPick)
 				src.print(P)
 				break
+			
 
 /datum/computer_file/program/paperwork_printer/proc/print(var/obj/item/paper/paperwork/T)
 	var/obj/item/computer_hardware/printer/printer
--- a/code/modules/modular_computers/file_system/programs/paperworkprinter.dm
+++ b/code/modules/modular_computers/file_system/programs/paperworkprinter.dm
@@ -51,8 +51,6 @@
 			src.doprint(11)
 
 /datum/computer_file/program/paperwork_printer/proc/doprint(ftype)
-	if(..())
-		return
 	var/obj/item/computer_hardware/printer/printer
 	if(computer)
 		printer = computer.all_components[MC_PRINT]
--- a/code/modules/modular_computers/file_system/programs/paperworkprinter.dm
+++ b/code/modules/modular_computers/file_system/programs/paperworkprinter.dm
@@ -11,8 +11,7 @@
 	program_icon = "clipboard-list"
 
 /datum/computer_file/program/paperwork_printer/ui_static_data(mob/user)
-		var/list/data = get_header_data()
-
+	var/list/data = get_header_data()
 	var/obj/item/computer_hardware/printer/printer
 	if(computer)
 		printer = computer.all_components[MC_PRINT]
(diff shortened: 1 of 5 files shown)


code/modules/modular_computers/file_system/programs/paperworkprinter.dm
```diff
@@ -0,0 +1,75 @@
...
+	filedesc = "Paperwork Printing"
+	category = PROGRAM_CATEGORY_MISC
+	program_icon_state = "id"
+	extended_desc = "Program for dispensing paperwork"
+	requires_ntnet = FALSE
+	size = 4
+	tgui_id = "NtosPaperworkPrinter"
+	program_icon = "clipboard-list"
+	var/obj/item/paper/paperwork/which
+	var/print = 1
```
ynot01: - why is this declared here instead of the sole scope that it's used?
- why does it start out TRUE?
- why does it use 1 instead of TRUE when it's used solely in truthy operations?

code/modules/modular_computers/file_system/programs/paperworkprinter.dm
```diff
@@ -0,0 +1,75 @@
...
+		data["have_printer"] = FALSE
+	return data
+
+/datum/computer_file/program/paperwork_printer/ui_act(action, params, datum/tgui/ui)
+	if(..())
+		return
+
+	switch(action)
+		if("PRG_print_GenReqForm")
+			which = /obj/item/paper/paperwork/general_request_form
```
ynot01: i would have the action be "PRG_print_form" or something similar and decide which paper (using a switch statement, do not blindly accept user input) to use using params instead of a unique action for each paper type

tgui/packages/tgui/interfaces/NtosPaperworkPrinter.js
```diff
@@ -0,0 +1,89 @@
+import { map } from 'common/collections';
+import { useBackend } from '../backend';
+import { Button, Section } from '../components';
+import { NtosWindow } from '../layouts';
+
+export const NtosPaperworkPrinter = (props, context) => {
+  const { act, data } = useBackend(context);
+  const {
+    have_printer,
+    manifest = {},
```
ynot01: unused and undeclared (in ui_data) variable

ynot01 (changes requested):

ynot01 (changes requested):
please include an image of the tgui interface ingame

code/modules/modular_computers/file_system/programs/paperworkprinter.dm
ReddicusDragon: It uses 1 instead of TRUE because it is functionally equivalent according to byond documentation, and is more readable in my opinion. It's a bit flag, if that makes sense. Also, it shouldn't start true. I may have either forgot to push a commit or commit some changes.

ynot01: same for the `which` variable, it has no reason to be declared here

code/modules/modular_computers/file_system/programs/paperworkprinter.dm
ReddicusDragon: I could do that, yeah. The way I was going to approach it was have the param be item type but I'm fairly certain that wouldn't work. Will have to toy around with it.

tgui/packages/tgui/interfaces/NtosPaperworkPrinter.js
ReddicusDragon: Oversight on my part. Will fix.

code/modules/modular_computers/file_system/programs/paperworkprinter.dm
ynot01: > It uses 1 instead of TRUE because it is functionally equivalent according to byond documentation, and is more readable in my opinion. 

you are correct in that they are functionally equivalent; TRUE is a macro, which means it does nothing but point to `1`; it was created solely for readability and ease of use when using truthy operations

> It's a bit flag, if that makes sense.

it doesn't because the only way you use it is in truthy operations `if(print)` and `print = !print`

further, BYOND fully supports bitflags e.g. `1<<0` `1<<1` `1<<2` and bitwise operations, see [byond ref](https://www.byond.com/docs/ref/) for further info, although i don't immediately see why you'd use them here

ReddicusDragon:
> please include an image of the tgui interface ingame

NoPrinter
Printer

code/modules/modular_computers/file_system/programs/paperworkprinter.dm
```diff
@@ -0,0 +1,80 @@
...
+	tgui_id = "NtosPaperworkPrinter"
+	program_icon = "clipboard-list"
+
+/datum/computer_file/program/paperwork_printer/ui_static_data(mob/user)
+	var/list/data = get_header_data()
+	var/obj/item/computer_hardware/printer/printer
+	if(computer)
+		printer = computer.all_components[MC_PRINT]
+
+	if(computer)
```
ynot01: redundant if

ynot01 (changes requested):

code/modules/modular_computers/file_system/programs/paperworkprinter.dm
ReddicusDragon: Issue fixed, now utilizing TRUE instead of 1 for truthy operation, further, the variable `which` has been moved to the proper place in the code.

code/modules/modular_computers/file_system/programs/paperworkprinter.dm
ReddicusDragon: will fix

ReddicusDragon:
I would like to thank baiomu, Chester, and Bibby for making this possible

JamieD1 (approved):

PR #16396, opened 2022-11-07
SuperMatter now reacts with new gases + new reactions

CMOisLing:
New gases now actually react with SM. Added some new reactions, will add more in another PR after this is merged and experimented with because im scared what atmos mains can do.

Zauker - Increases SM damage cap 10x at full ratio
Healium - Increases healing speed 5x
Halon - Greatly reduces Heat and Waste, at 15% reduces plasma o2 ratio to 50 50
Hypernoblium - Greatly increases heat resistance

:cl:  
rscadd: Adds new SM gas reactions 
bugfix: new gases now actually react with SM
/:cl:

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -13,6 +13,7 @@
 #define OXYGEN_HEAT_PENALTY 1
 #define CO2_HEAT_PENALTY 0.1
 #define PLUOXIUM_HEAT_PENALTY -1
+#define HALON_HEAT_PENALTY -1
 #define TRITIUM_HEAT_PENALTY 10
 #define BZ_HEAT_PENALTY 5
 #define NITROGEN_HEAT_PENALTY -1.5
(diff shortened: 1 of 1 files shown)


CMOisLing:
Will fix when I figure out VSC again

iamthesilly:
should replace halon with hypernoblium in my opinion

CMOisLing:
Hypernoblium already stops all in chamber and loop reactions so you can get really high temps without worrying about fires. I figured it would be overkill to add an effect as well.

iamthesilly:
> Hypernoblium already stops all in chamber and loop reactions so you can get really high temps without worrying about fires. I figured it would be overkill to add an effect as well.

it doesn't stop fires actually

iamthesilly:
for whatever reason combustion reactions aren't stopped by hypernoblium, i've been trying to fix it but haven't been able to figure out the issue

Rdrazga:
Halon should not kill power, it is useful in very specific SM reactions as it kills the oxygen the SM produces (lowers power slightly) but stops fires because of this. Halon is also stupidly hard to produce for how few moles of O2 it takes out of the air imo

CMOisLing:
closing this until I get time to work on it

JamieD1 (approved):

PR #16397, opened 2022-11-07
reduces the validhunting capabilities of syringe guns by making piercing syringes not fit into most syringe guns

TheRyeGuyWhoWillNowDie:
wow its fun to have the counter to something and it still means nothing and you get clicked once and die instantly
dart pistol can still load the piercing syringes because currently its fucking awful and there's no reason for it to exist
piercing syringes not removed because its actually helpful to be able to inject stuff into dead/dying people wearing hardsuits in space or whatever, also piercing syringe capacity changed to 15u to make it better for actually being a syringe and not a projectile

# Changelog
:cl:  
tweak: all syringe guns except the dart pistol can no longer fire piercing syringes
tweak: piercing syringes now hold 15u from 10u
tweak: bonus: syringes that are supposed to pierce armor now do so in melee
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/misc/syringe_gun.dm
+++ b/code/modules/projectiles/guns/misc/syringe_gun.dm
@@ -61,6 +61,9 @@
 /obj/item/gun/syringe/attackby(obj/item/A, mob/user, params, show_msg = TRUE)
 	if(istype(A, /obj/item/reagent_containers/syringe))
 		if(syringes.len < max_syringes)
+			if(istype(A, /obj/item/reagent_containers/syringe/piercing) && !istype(src, /obj/item/gun/syringe/syndicate))
+				to_chat(user, "<span class='notice'>[A] is designed to not be able to fit into [src].</span>")
+				return TRUE
 			if(!user.transferItemToLoc(A, src))
 				return FALSE
 			to_chat(user, span_notice("You load [A] into \the [src]."))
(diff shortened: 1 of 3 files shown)


forhiti:
wasn't the syringe gun already nerfed due to the fact that it doesn't inject all contents at once?

UltraBigRat:
honestly good idea to make station side dart guns not pierce, while letting the traitor dart pistol actually have any reason to be bought

Notamaniac:
I don't think piercing syringes penetrate hardsuits unless they're specifically fired out of a syringe gun. Correct me if I'm wrong on that.

Simplehorror:
so why are they in the fucking techfab instead of a dart pistol kit

Potototatoh:
inside job requested by meowsy

Maia-J:
Name 2 validhunting incidents which included piercing syringes and had different people firing them.

TheRyeGuyWhoWillNowDie:
> Name 2 validhunting incidents which included piercing syringes and had different people firing them.

any time any medliz gets their hands on a syringe gun, more than 2 but close enough

Maia-J:
> > Name 2 validhunting incidents which included piercing syringes and had different people firing them.
> 
> any time any medliz gets their hands on a syringe gun, more than 2 but close enough

That's not an example but what seems like a personal grudge. Just make a complaint like a normal person if it happens so often. Atleast the lizards would be gone

kooarbiter:
honestly, I'm not quite sure anyone has ever really used syringe guns for anything other than roofies or poison, why tf does medical even have syringe guns?

JamieD1 (approved):

PR #16398, opened 2022-11-07
Bodies copy damage to chairs

Moltijoe:
This removes the wonky strat of using corpses buckled to chairs as invincible walls against projectiles.
It still works, the chair just eventually breaks.

:cl:  
tweak: dealing damage to a corpse buckled to a chair also damages the chair
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1888,7 +1888,6 @@ GLOBAL_LIST_EMPTY(mentor_races)
 			H.adjustOrganLoss(ORGAN_SLOT_BRAIN, damage * hit_percent * H.physiology.brain_mod)
 	
 	if(H.buckled && istype(H.buckled, /obj/structure))//prevent buckling corpses to chairs to make indestructible projectile walls
-		to_chat(world, "[H] is buckled to [H.buckled]")
 		var/obj/structure/sitter = H.buckled
 		sitter.take_damage(damage, damagetype)
 	return 1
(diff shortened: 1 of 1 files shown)


Cartlord:
what if we just made projectiles go through corpses

Moltijoe:
> what if we just made projectiles go through corpses

but... why?

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1886,6 +1886,10 @@ GLOBAL_LIST_EMPTY(mentor_races)
 				H.adjustStaminaLoss(damage * hit_percent * H.physiology.stamina_mod)
 		if(BRAIN)
 			H.adjustOrganLoss(ORGAN_SLOT_BRAIN, damage * hit_percent * H.physiology.brain_mod)
+	
+	if(H.buckled && istype(H.buckled, /obj/structure))//prevent buckling corpses to chairs to make indestructible projectile walls
+		var/obj/structure/sitter = H.buckled
+		sitter.take_damage(damage, damagetype)
```
ynot01, line 1892: i would add a vis message to indicate that the chair is taking damage

ynot01 (changes requested):

code/modules/mob/living/carbon/human/species.dm
Moltijoe: it plays a sound effect

Marmio64:
damn, rip the one good way for clockcult to deal with x-ray beam spammers

Moltijoe:
> damn, rip the one good way for clockcult to deal with x-ray beam spammers

I don't often see that happening, if it becomes enough of a problem i can work on a less jank-ridden way to combat it

The-Sun-In-Splendour:
> > damn, rip the one good way for clockcult to deal with x-ray beam spammers
> 
> I don't often see that happening, if it becomes enough of a problem i can work on a less jank-ridden way to combat it

just make superheated clock walls reflect laser beams

Moltijoe:
> just make superheated clock walls reflect laser beams

are superheated clock walls a thing? also, i'd probably just have it stop lasers rather than reflect them.

JamieD1 (approved):

PR #16399, opened 2022-11-07
Preternis charge the Vxtvul Hammer 1 second faster than other species

Moltijoe:
With the rework, preternis attack slower than regular species, this also made their species specific hammer weaker for them than anyone else using it.
This buffs it without making it better at regular combat.

It goes from 6 seconds to 5 seconds in their hands

:cl:  
tweak: Preternis charge the Vxtvul hammer faster
/:cl:


Addust:
given that preterni were made by the civilisation that MADE the Vxtvul hammer, this also makes sense lorewise

Moltijoe:
> given that preterni were made by the civilisation that MADE the Vxtvul hammer, this also makes sense lorewise

And since it's charging a hammer, it would make sense that preterni who use electricity would be able to use some of their own to charge the hammer faster

JamieD1 (approved):

PR #16409, opened 2022-11-08
removes maint access from security officers outside of engie guards and skeleton crew

TheRyeGuyWhoWillNowDie:
makes the 'roll around maintenance roundstart searching for a valid' technique less common or fast, it's not fun for anyone when there's 6 security officers running around maintenance roundstart and killing all the valids in like 5 minutes which ruins the game for everyone. they can of course get access added to their ID but that requires cooperation between players and at the very least will slow down the maintenance rollout
engie guard gets maintenance access because maintenance is part of engineering (technically) so it makes sense
# Changelog

:cl:  
tweak: maintenance access is now restricted to engineering security or lowpop security, otherwise they dont have it
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/security_officer.dm
+++ b/code/modules/jobs/job_types/security_officer.dm
@@ -83,7 +83,7 @@ GLOBAL_LIST_INIT(available_depts_sec, list(SEC_DEPT_ENGINEERING, SEC_DEPT_MEDICA
 			accessory = /obj/item/clothing/accessory/armband/cargo
 		if(SEC_DEPT_ENGINEERING)
 			ears = /obj/item/radio/headset/headset_sec/alt/department/engi
-			dep_access = list(ACCESS_CONSTRUCTION, ACCESS_ENGINE, ACCESS_ATMOSPHERICS)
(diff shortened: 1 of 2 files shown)


Mqiib:
Wow this is a terrible idea

TosterIMC:
Best idea ever, no more patrol maint so i can find stealth antags

Hopekz:
This is a bad way to go about your goal.
If you are concerned with the beginning of the round why not just make their access not work for the first few minutes versus adding a chore they have to do every single shift.

I get that what I just proposed is a bad idea as well because the whole idea of security not having access to maintenance is pretty weird and probably shouldn't be merged.

00ze-cyclone:
WAAAAAAAAAAAH!!  SEC VALIDHUNTING ME!!
lmao

ynot01:
officers already suffer from being pretty much useless when someone is committing crimes in an area they don't have access to

restricting them out of the maintenance areas when
- the maint area is so useful that it is essentially a requirement for any antag
- it is incredibly easy for each antag to get into maint

would be a huge nerf to security

tl;dr Wow this is a terrible idea

Bibby0110:
This is a terrible idea

JamieD1:
I backup Gamer.

Simplehorror:
how are you so full of bad ideas

Hopekz:
> how are you so full of bad ideas

It's fine as long as there is someone else to filter the Good ones from the bad ones

PR #16422, opened 2022-11-09
Clamp threat level depending on the playtime of the security forces in play

Simplehorror:

this way we won't have 100 dynamic when the only security officer is some 10 hour noob who just wants to learn
# Changelog

:cl:  
tweak: Clamped threat level depending on the playtime of the security forces in play
/:cl:

### DIFF
--- a/code/game/gamemodes/dynamic/dynamic.dm
+++ b/code/game/gamemodes/dynamic/dynamic.dm
@@ -350,6 +350,15 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
 	var/relative_threat = LORENTZ_DISTRIBUTION(threat_curve_centre, threat_curve_width)
 	threat_level = round(lorentz_to_amount(relative_threat), 0.1)
 
+	var/most_hours_played = 0
+	for (var/mob/plr in GLOB.player_list)
+		var/mob/living/carbon/human/hm = plr
+		if (istype(hm))
+			var/datum/job/J = SSjob.GetJob(hm.job)
+			if (J && J.department_flag == ENGSEC)
+				if ((hm.client.prefs.exp[EXP_TYPE_SECURITY] / 60) > most_hours_played)
+					most_hours_played = (hm.client.prefs.exp[EXP_TYPE_SECURITY] / 60)
+	threat_level = clamp(threat_level, 0, clamp(most_hours_played, 25, 100000))
 	peaceful_percentage = round(LORENTZ_CUMULATIVE_DISTRIBUTION(relative_threat, threat_curve_centre, threat_curve_width), 0.01)*100
 
 /// Generates the midround and roundstart budgets


ynot01:
this is bad because 6 security officers with 500 hours will receive the same threat as 1 security officer with 500 hours

code/game/gamemodes/dynamic/dynamic.dm
```diff
@@ -350,6 +350,15 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
 	var/relative_threat = LORENTZ_DISTRIBUTION(threat_curve_centre, threat_curve_width)
 	threat_level = round(lorentz_to_amount(relative_threat), 0.1)
 
+	var/most_hours_played = 0
+	for (var/mob/plr in GLOB.player_list)
+		var/mob/living/carbon/human/hm = plr
+		if (istype(hm))
+			var/datum/job/J = SSjob.GetJob(hm.job)
+			if (J && J.department_flag == ENGSEC)
```
ynot01, line 358: why is engineering involved

Simplehorror:
> this is bad because 6 security officers with 500 hours will receive the same threat as 1 security officer with 500 hours

no, it takes the maximum hour count, not the average. unless you mean how many security officers there are. in which case it's already this way

code/game/gamemodes/dynamic/dynamic.dm
Simplehorror: engineering and security share department flag

ynot01:
> unless you mean how many security officers there are. in which case it's already this way

it "already being this way" does not mean it's good and accomplishes what your PR is trying to achieve

code/game/gamemodes/dynamic/dynamic.dm
ynot01: so an engineer with high hours is fucking over a sec team with low hours?

Simplehorror:
> > unless you mean how many security officers there are. in which case it's already this way
> 
> it "already being this way" does not mean it's good and accomplishes what your PR is trying to achieve

perfect is the enemy of good

Simplehorror:
please recommend me a formula that you believe is balanced to take into account threat level, the amount of officers and their hours

ynot01:
> please recommend me a formula that you believe is balanced to take into account threat level, the amount of officers and their hours

i don't have a formula, i just heavily suggest that you incorporate the # of officers in some way because it is just as if not more impactful than the hours of a single officer; maybe just add everyone's hours instead of selecting only the highest one

Simplehorror:
ok i think about it

ynot01:
consider adding midround threat when officers mid-join to prevent sec mains from developing a habit where they join 5 seconds in to circumvent threat calculation

Simplehorror:
yeah that might actually take some effort, i'll close this one and make a new one when i'm not tired and can actually work on it

Hopekz:
This is almost a good PR I think.
Maybe all it's missing is for security officers that are under an x amount of hours (learning) to not count towards threat calculation. That way it covers your case scenario of security only having a new officer that wants to learn basics.

I would imagine these hours would be low like the first 4 hours

PR #16423, opened 2022-11-09
Improves and doesnt remove antinobilium

Simplehorror:
Changed the TC price of antinobilium from 10 to 3. I have determined this is a fair value seeing as how it takes 1 C4 to destroy the SM, and a revolver does the same. and both have INFINITELY more utility and value. Frankly this should be a free item since it's more likely to get you killed than it is to give you a benefit which is just a station wide EMP.

# Wiki Documentation
Changed the TC price of antinobilium from 10 to 3.
# Changelog
:cl:  
tweak: Changed the TC price of antinobilium from 10 to 3.
/:cl:


Bibby0110 (changes requested):
Improve it, don’t just knee jerk reduce the price

Simplehorror:
> Improve it, don’t just knee jerk reduce the price

name one way it can be improved without making it into a completely different item, which is equivalent to removal

Simplehorror:
because frankly I can delaminate the SM for free, without being at risk of killing myself or getting caught

Bibby0110:
> > Improve it, don’t just knee jerk reduce the price
> 
> name one way it can be improved without making it into a completely different item, which is equivalent to removal

Nah that’s your job

Simplehorror:
what the hell

PR #16443, opened 2022-11-10
Announcement if AI jumps into a mech and is nuking

Simplehorror:
Announcement if AI jumps into a mech and is nuking. People are easily confused when both cores are destroyed and doomsday timer doesn't end, so this makes it obvious the AI is on the lam.
# Changelog
:cl:  
rscadd: Announcement if AI jumps into a mech and is nuking
/:cl:

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -825,6 +825,9 @@
 
 //Hack and From Card interactions share some code, so leave that here for both to use.
 /obj/mecha/proc/ai_enter_mech(mob/living/silicon/ai/AI, interaction)
+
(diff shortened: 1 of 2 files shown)


The-Sun-In-Splendour:
finally, a good simple horror pr

Bibby0110 (changes requested):
Nuke shouldn't tick while inside a mech

Simplehorror:
> Nuke shouldn't tick while inside a mech

it should, it's how it has been since 2015

Simplehorror:
if "nuke shouldn't tick while inside a mech" i'll just scrap mech domination too since that would be noob bait

PR #16446, opened 2022-11-10
Mail Changes - The Insane Rambling of a Man Cursed with Too Much Mail

ToasterBiome:

We use a define of MAX_MAIL_PER_MINUTE (3) but then when figuring out the waiting mail we multiply that value by 5 so there's actually 15 mail a minute. This is a lot. I have removed the multiplication by 5 and set the MAX_MAIL_PER_MINUTE to 9, which is lower.

I also added the ABILITY TO SEND THE MAIL BACK IF YOU DON'T WANT IT.
What if a Cargo Tech doesn't want to deliver mail? It goes IN THE TRASH!!!!!!!!!!!!!!!!!!!!!!!!!! This way, you don't have to keep going back and forth between the damn trash. 
Clockies knocking on your door and you're panic ordering Energy Gun crates? I HOPE YOU LIKE TAKING OUT THE MAIL!!!!!!!!!!!!!!!!!!!! OH YOU'RE TRYING TO BUY NULL CRATES? THE BOTANIST'S BIRTHDAY CARD FROM THEIR GRANDMA IS STUCK IN THE SPACE BRAKES, THIS SHIT AINT FLYING!

I also adjusted the normal loot pool to include some coins because coins are cool, they are slightly less chance than the equivalent spacecash because you can recycle the coins into materials.

Last thing is I made it so that if there's 3 people on the station you don't get 50000 junk mail, you get 22500 instead (55% less)

# Changelog

:cl:  
tweak: The station gets 9 mail a minute instead of 15
tweak: You can send mail back on the cargo shuttle
tweak: generic mail loot includes various coins
tweak: the chance of filling an empty space in the mail crate with junkmail was reduced from 100% to 45%
/:cl:

### DIFF
--- a/code/__DEFINES/economy.dm
+++ b/code/__DEFINES/economy.dm
@@ -2,7 +2,9 @@
 #define STARTING_PAYCHECKS_MIN 50 // Minimum roundstart money
 #define STARTING_PAYCHECKS_MAX 250 // Maximum roundstart money
 /// How much mail the Economy SS will create per minute, regardless of firing time.
-#define MAX_MAIL_PER_MINUTE 3
+#define MAX_MAIL_PER_MINUTE 6
+/// Chance to generate junk mail compared to generating nothing while packing mail crates.
+#define MAIL_JUNK_CHANCE 45 
 /// Probability of using letters of envelope sprites on all letters.
 #define FULL_CRATE_LETTER_ODDS 70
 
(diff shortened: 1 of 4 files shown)


ShoziOffical:
uhh is it me or mails go to station everytime i send back shuttle, it doesnt care about timers at all

Moltijoe:
Wait, is max mail per minute 9 now or 6? It says 6 in the code but this pr says 9.
Does cargo get anything for actually delivering the mail, or with it being optional will they often just send it back right away?

Potototatoh:
i usually deliver the mail as a device to warrant getting the department to make us shit cause the mail usually contains money

MajManatee:
i would support this, if i could commit mail fraud and burn unwanted mail.

JamieD1 (approved):

PR #16464, opened 2022-11-11
Tweaks medspray cans to be usable

Moltijoe:
With the addition of hyposprays this made medsprays very useless.
This gives them their own niche in being "consumable hyposprays".
Since they aren't refillable, and the only variants that currently exist are very limited in the chems they have, a proper hypomix is still FAR preferred to using these.

This make synthflesh sprays not a complete joke and waste of money, and possibly worth giving to mining medics.

There is still the "squirt mode" if you want to apply less per spray than 25u

:cl:  
tweak: Medsprays now hold 100u instead of 30u and apply 25u per use instead of 10u
tweak: Medsprays are no longer refillable
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/medspray.dm
+++ b/code/modules/reagents/reagent_containers/medspray.dm
@@ -14,15 +14,19 @@
 	w_class = WEIGHT_CLASS_SMALL
 	throw_speed = 3
 	throw_range = 7
-	amount_per_transfer_from_this = 10
-	volume = 30
+	amount_per_transfer_from_this = 25
+	volume = 100
 	var/can_fill_from_container = TRUE
 	var/apply_type = PATCH
 	var/apply_method = "spray"
-	var/self_delay = 3 SECONDS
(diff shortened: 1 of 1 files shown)


Burger1998:
will no longer being fillable effect the sprays in chem locker?

Moltijoe:
> will no longer being fillable effect the sprays in chem locker?

I would imagine so, this just changed the objects themselves, any instances of them will be updated

Sniblet:
The sprays in chem lockers are empty by default, so this makes all those spawns into trash items
I believe you can also print empty medical sprays from the medical protolathe

Moltijoe:
> The sprays in chem lockers are empty by default, so this makes all those spawns into trash items I believe you can also print empty medical sprays from the medical protolathe

ah, yeah, they'd be useless, i could either just remove them if they're bloat, or i could change the pr to only affect the specific prefilled ones

Sniblet:
I've never used a hypospray so I can't say what's redundant and not ¯\_(ツ)_/¯

JamieD1 (approved):

PR #16470, opened 2022-11-11
Viruses now spread as advertised

ChesterTheCheesy:
Makes it so if the virus doesn't have a fluids spread flag and gets splashed on someone, it doesn't get treated like an injection.
Getting splashed by a high transmission viruses is properly blocked by armor.

# Changelog
:cl:  
bugfix: being splashed by low transmitability virus is no longer treated as an injection.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -18,8 +18,9 @@
 			if((D.spread_flags & DISEASE_SPREAD_SPECIAL) || (D.spread_flags & DISEASE_SPREAD_NON_CONTAGIOUS))
 				continue
 
-			if((method == TOUCH || method == VAPOR) && (D.spread_flags & DISEASE_SPREAD_CONTACT_FLUIDS))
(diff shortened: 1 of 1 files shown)


Simplehorror:
This commit was created on GitHub.com and signed with GitHub’s verified signature.

Sniblet:
I believe this means that smoke and foam are no longer effective vectors for bloodborne viruses

ChesterTheCheesy:
> I believe this means that smoke and foam are no longer effective vectors for bloodborne viruses

The real problem was that you could've splashed someone with a virus, and if had low transmission (below fluids), it would force an infection. On higher transmission it worked properly and did correct armor checks for permeability before forcing an infection.

All I did was fixing a bug with low transmission, if virus only infects via blood, it actually needs to enter your bloodstream.

CMOisLing:
You know I never noticed this, gonna have to get rid of my muscle memory to chuck beakers of HIV: Prepare to Die Edition at anyone trying to attack me in viro.

JamieD1 (approved):

PR #16476, opened 2022-11-12
Xenomorphs mini rework

Moltijoe:
Xenomorphs full stripping is boring, it makes recovering people from xeno nests pointless unless all xenos are dead.
This removes their ability to strip, but also lets them manually apply facehuggers to forcefully remove the target's helmet. (so they can still facehug people with hardsuits)

This should make it at least possible for those that escape to survive as they still have most of their gear.
This also helps people locate victims by not letting xenos from just removing suit sensors.

I am receptive to other buffs to xenos that will make sure this isn't a crippling nerf

**Current buff ideas are**
- Xeno embryo take less time to burst out of a body (it currently takes 5 whopping minutes)(this has been changed to 3:20)
- Facehugger implants quicker
- Give stat buffs to xenos

Hopefully with them being easier to notice, but stronger it'll make xeno antag triggers less binary
As they are currently either found instantly and demolished, or not found until it's too late to even have a chance of doing anything

:cl:  
tweak: Xenos can no longer strip people
tweak: Xenos can now apply facehuggers by hand, removing any helmets in the way
tweak: Xenos beds now bucklecuff
tweak: Facehuggers now act like muzzles
tweak: Alien Embryos now burst in 3:20 instead of 5:00
/:cl:

### DIFF
--- a/code/game/objects/items/handcuffs.dm
+++ b/code/game/objects/items/handcuffs.dm
@@ -157,6 +157,18 @@
 
 /obj/item/restraints/handcuffs/alien
 	icon_state = "handcuffAlien"
+	
+/obj/item/restraints/handcuffs/alien/beno
+	name = "resin handcuffs"
+	desc = "Extremely weak cuffs made to restrict movement."
+	icon_state = "cuff" // Needs sprite
+	breakouttime = 10 //Deciseconds = 1s
+	break_strength = 1
+	item_flags = DROPDEL
+
+/obj/item/restraints/handcuffs/alien/beno/dropped(mob/user)	
+	user.visible_message(span_danger("[user]'s [name] falls apart"))
+	. = ..()
 
 /obj/item/restraints/handcuffs/fake
 	name = "fake handcuffs"
--- a/code/game/objects/structures/beds_chairs/alien_nest.dm
+++ b/code/game/objects/structures/beds_chairs/alien_nest.dm
@@ -60,7 +60,10 @@
 	if(has_buckled_mobs())
 		unbuckle_all_mobs()
 
+	var/mob/living/carbon/human/alienprisoner = M
 	if(buckle_mob(M))
+		alienprisoner.handcuffed = new /obj/item/restraints/handcuffs/alien/beno(alienprisoner)
+		alienprisoner.update_handcuffed()
 		M.visible_message(\
 			"[user.name] secretes a thick vile goo, securing [M.name] into [src]!",\
 			span_danger("[user.name] drenches you in a foul-smelling resin, trapping you in [src]!"),\
--- a/code/modules/mob/living/carbon/alien/humanoid/humanoid.dm
+++ b/code/modules/mob/living/carbon/alien/humanoid/humanoid.dm
@@ -21,7 +21,14 @@
 	bodyparts = list(/obj/item/bodypart/chest/alien, /obj/item/bodypart/head/alien, /obj/item/bodypart/l_arm/alien,
 					 /obj/item/bodypart/r_arm/alien, /obj/item/bodypart/r_leg/alien, /obj/item/bodypart/l_leg/alien)
 
-
+/mob/living/carbon/alien/humanoid/stripPanelEquip(obj/item/what, mob/who, where)
+	to_chat(src, span_warning("Your sharp claws fumble as you attempt to put the [what] on [who]."))
+	return FALSE
+	
+/mob/living/carbon/alien/humanoid/stripPanelUnequip(obj/item/what, mob/who, where)
+	to_chat(src, span_warning("Your sharp claws fumble as you attempt to remove [who]'s [what]."))
+	return FALSE
+	
 /mob/living/carbon/alien/humanoid/restrained(ignore_grab)
 	return handcuffed
 
--- a/code/modules/mob/living/carbon/alien/special/alien_embryo.dm
+++ b/code/modules/mob/living/carbon/alien/special/alien_embryo.dm
@@ -9,7 +9,7 @@
 	/// Are we bursting out of the poor sucker who's the xeno mom?
 	var/bursting = FALSE
 	/// How long does it take to advance one stage? Growth time * 5 = how long till we make a Larva!
-	var/growth_time = 60 SECONDS
+	var/growth_time = 40 SECONDS
 
 /obj/item/organ/body_egg/alien_embryo/Initialize()
 	. = ..()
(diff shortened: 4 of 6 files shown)


Simplehorror:
new meta: 5 meth grenades in my bag to detonate on death when i get dragged into the xenobase

Moltijoe:
> new meta: 5 meth grenades in my bag to detonate on death when i get dragged into the xenobase

any different than just cavity implanting them?
or bombing them prior to death?
or just suicide verbing with the bombs while the queen is next to you?
or just using a maxcap from literally anywhere within a decent radius?

That just sounds like problems with bombs being unfun

Simplehorror:
> > new meta: 5 meth grenades in my bag to detonate on death when i get dragged into the xenobase
> 
> any different than just cavity implanting them? or bombing them prior to death? or just suicide verbing with the bombs while the queen is next to you? or just using a maxcap from literally anywhere within a decent radius?
> 
> That just sounds like problems with bombs being unfun

you can't do any of these without this PR aside the first one

Moltijoe:
> you can't do any of these without this PR aside the first one

you can though? bombs are all over the bloody place. people ALREADY suicide bomb xenos, this doesn't change enough to incentivize people into doing it if they weren't already planning to.

You can't tell me you think xenos are in a fun place balance wise, they're built up entirely around instant stun mechanics and round removing. With a roughly 50/50 chance of either becoming an unstoppable force or getting lynched instantly

Moltijoe:
Reducing the time it takes for larvas to burst from bodies is currently my favourite idea for buffing, gets people back in the game faster, lets xenos scale up faster, lets them recover from losses faster.
I'm also disappointed by how rarely someone gets bursted during surgery, i kinda want that to happen more.

Mqiib:
Ok so important note, xenos HAVE to strip people because otherwise they can just use tools and other random items to smash the resin bed and any unfortunate xenos around them, and you cant crit them because they'll just succumb and ruin the embryo when they die. This would be acceptable if resin beds also handcuffed you while you were in it, but they don't.

Moltijoe:
> Ok so important note, xenos HAVE to strip people because otherwise they can just use tools and other random items to smash the resin bed and any unfortunate xenos around them, and you cant crit them because they'll just succumb and ruin the embryo when they die. This would be acceptable if resin beds also handcuffed you while you were in it, but they don't.

don't they bucklecuff? wasn't that a pr made within the past few months?
ah, nevermind, it was part of #15794 which had other bad takes and got closed
lemme add that real quick

also, the embryo doesn't die if the host does, that was also part of the closed pr

Sniblet:
Given that this necessitates that all xeno caps now be killed lest they A) radio or B) pull any of an incredible variety of hilariously cheesy nonsense to both free themselves and annihilate the Queen faster than anyone can react, I guess the suit sensors are the biggest issue
And it's a big one because we love having twenty borgs storm the nest about five minutes after xenomorph infestation triggers

I like the burst time, because it's always obnoxiously slow for xenos (with unremovable suit sensors being added, even more so, because lizbay guarantees that the nest will be stormed), and never oppressively fast for humans

An electronic disabling ability would work alternatively to handle both the radio and the suit sensors, but I'm not sure how you'd do that in a thematic way... copy CM's method where you can kill armor lights and cameras by slashing the corpse? Some kind of psychic EMP that only the royals can do?

Moltijoe:
> A) radio or B) pull any of an incredible variety of hilariously cheesy nonsense to both free themselves and annihilate the Queen faster than anyone can react

Facehuggers still prevent speaking so radios won't be an issue.
Also with the bed bucklecuffing that means that someone will need to break out of the bed and then the handcuffs in order to do anything cheesy

> I like the burst time, because it's always obnoxiously slow for xenos (with unremovable suit sensors being added, even more so, because lizbay guarantees that the nest will be stormed), and never oppressively fast for humans

You like the burst time being as slow as it is? or buffing burst time?

>An electronic disabling ability would work alternatively to handle both the radio and the suit sensors, but I'm not sure how you'd do that in a thematic way... copy CM's method where you can kill armor lights and cameras by slashing the corpse? Some kind of psychic EMP that only the royals can do?

I'm more inclined to make it somehow change people's sensors from tracking to binary when buckled to the resin bed or something, but making that be obvious to people doing it seems to be an undertaking

Sniblet:
> Facehuggers still prevent speaking so radios won't be an issue.

Uhhh... are you sure? I don't think I've ever seen that and I can't find it in the code.

> You like the burst time being as slow as it is? or buffing burst time?

Buffing it

Moltijoe:
> Uhhh... are you sure? I don't think I've ever seen that and I can't find it in the code.

Maybe it's just the super long sleep it applies, lemme check, if it doesn't i can make it do that

> Buffing it

Sounds good, it's an easy change, currently it's 5 ticks of 60 seconds, i'm leaning towards 40 seconds per tick or maybe less

Sniblet:
Little concerned that 200 seconds might be too fast, but we'll find out won't we? If nothing else, bursting leaves the head intact
If it does turn out to be too oppressive, the TDM servers like(d, in TGMC's case) to use mechanics where larva growth is faster under nest conditions

Moltijoe:
> Little concerned that 200 seconds might be too fast, but we'll find out won't we? If nothing else, bursting leaves the head intact If it does turn out to be too oppressive, the TDM servers like(d, in TGMC's case) to use mechanics where larva growth is faster under nest conditions

either that or stasis preventing growth

Moltijoe:
This is a very divisive pr, id appreciate those downvoting to state whether they think if the idea or the execution is flawed

Bibby0110 (changes requested):
Atomize or update your title to be less misleading

JamieD1:
I dont like the idea of people being able to insta pull out an armoury of guns to shoot aliens with, kinda dumb.

Moltijoe:
> I dont like the idea of people being able to insta pull out an armoury of guns to shoot aliens with, kinda dumb.

yeah, that's always a problem, how to make something more interactive without letting people with experience just stomp
I can always just close this, but i don't currently feel like xenos are in a fun position for either side

Moltijoe:
Is it worth test merging?

Moja10:
do not like, there are many things that xenon have to worry about people carrying, and if they are stripped on the same tile, they could still get to a backpack to open a bomb or something, so it's not like people are totally helpless 100% of time.

Moltijoe:
i still think xenos need some amount of changes, but even a 50/50 split isn't good enough to warrant adding

PR #16486, opened 2022-11-13
Make the shuttle timer depend on REALTIMEOFDAY instead of world.time

Simplehorror:
This has definitely been responsible for a few bans
# Changelog
:cl:  
rscadd: Shuttle timer now depends on real time instead of dilated time
/:cl:

### DIFF
--- a/code/modules/shuttle/shuttle.dm
+++ b/code/modules/shuttle/shuttle.dm
@@ -642,11 +642,11 @@
 		SSshuttle.request_transit_dock(src)
 
 /obj/docking_port/mobile/proc/setTimer(wait)
-	timer = world.time + wait
+	timer = REALTIMEOFDAY + wait
 	last_timer_length = wait
 
 /obj/docking_port/mobile/proc/modTimer(multiple)
-	var/time_remaining = timer - world.time
+	var/time_remaining = timer - REALTIMEOFDAY
 	if(time_remaining < 0 || !last_timer_length)
 		return
 	time_remaining *= multiple
@@ -656,7 +656,7 @@
 /obj/docking_port/mobile/proc/invertTimer()
 	if(!last_timer_length)
 		return
-	var/time_remaining = timer - world.time
+	var/time_remaining = timer - REALTIMEOFDAY
 	if(time_remaining > 0)
 		var/time_passed = last_timer_length - time_remaining
 		setTimer(time_passed)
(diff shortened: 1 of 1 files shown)


Bibby0110 (changes requested):
DNM until testmerged since it touches every shuttle

ynot01:
world.time is game time
world.timeofday is real time
they are not interchangable without consequence

Simplehorror:
> world.time is game time world.timeofday is real time they are not interchangable without consequence

that's the point dummy

Bibby0110:
> > world.time is game time world.timeofday is real time they are not interchangable without consequence
> 
> that's the point dummy

This is just going to end up being out of sync then, the timer won’t process faster than world.time anyhow

Simplehorror:
> > > world.time is game time world.timeofday is real time they are not interchangable without consequence
> > 
> > 
> > that's the point dummy
> 
> This is just going to end up being out of sync then, the timer won’t process faster than world.time anyhow

what do you mean? the whole point is that it is out of sync with dilated time since we don't want it to be delayed. too many people have been banned because time dilation slowed down the shuttle timer so they mistimed their bombs

Bibby0110:
> > > > world.time is game time world.timeofday is real time they are not interchangable without consequence
> > > 
> > > 
> > > that's the point dummy
> > 
> > 
> > This is just going to end up being out of sync then, the timer won’t process faster than world.time anyhow
> 
> what do you mean? the whole point is that it is out of sync with dilated time since we don't want it to be delayed. too many people have been banned because time dilation slowed down the shuttle timer so they mistimed their bombs

My bad misread the code.
This will however still become out of sync with the game due to time dilation, but instead of the shuttle being dilated like everything else the timer would keep running. Doesn’t seem to be solving an issue

Simplehorror:
> > > > > world.time is game time world.timeofday is real time they are not interchangable without consequence
> > > > 
> > > > 
> > > > that's the point dummy
> > > 
> > > 
> > > This is just going to end up being out of sync then, the timer won’t process faster than world.time anyhow
> > 
> > 
> > what do you mean? the whole point is that it is out of sync with dilated time since we don't want it to be delayed. too many people have been banned because time dilation slowed down the shuttle timer so they mistimed their bombs
> 
> My bad misread the code. This will however still become out of sync with the game due to time dilation, but instead of the shuttle being dilated like everything else the timer would keep running. Doesn’t seem to be solving an issue

it solves the issue of people trying to detonate their bombs in time for end of round and then getting banned because time dilaton slowed down the timer

PR #16488, opened 2022-11-13
Wirecutters now take 0.5 seconds to cut a wire

Simplehorror:
Wirecutters now take 0.5 seconds to cut a wire. It takes like 15 seconds to wire something up and 0 seconds to cut it. Why?

# Changelog

:cl:  
tweak: Wirecutters now take 0.5 seconds to cut a wire. 
/:cl:

### DIFF
--- a/code/modules/power/cable.dm
+++ b/code/modules/power/cable.dm
@@ -142,6 +142,8 @@ By design, d1 is the smallest direction and d2 is the highest
 	if(W.tool_behaviour == TOOL_WIRECUTTER)
 		if (shock(user, 50))
 			return
+		if (!do_after(user, 0.5 SECONDS, src))
+			return
 		user.visible_message("[user] cuts the cable.", span_notice("You cut the cable."))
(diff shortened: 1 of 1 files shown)


iamthesilly:
placing wires is instant, why not cutting it?

Simplehorror:
it's not instant because you have to move around to actually get it to connect, and there's a delay every time you place one

ShoziOffical:
When you wire something, you gotta make sure things go into the right spot so it kinda take a bit time, but when you cut you dont need to think

Simplehorror:
> When you wire something, you gotta make sure things go into the right spot so it kinda take a bit time, but when you cut you dont need to think

+1

CMOisLing:
Make it scale with engineering hours.
300hr engineers should be able to place wires without making mistakes.

Bibby0110:
Change that doesn’t provide anything but a mild annoyance, not what we’re looking for

PR #16504, opened 2022-11-14
Reverts IPC lungs, other tweaks included

Moltijoe:
reverts #15682 
The idea was reasonable, but once playing it, it was clear that it just isn't ideal.
- It just felt like breathing, which removed one of the defining features of the species
- IPC took burn instead of oxy, which doesn't slowly heal back up after.
- Knowing which gases actually are effective requires significantly more knowledge than should be expected from something as fundamental as breathing
- The emergency oxygen tank everyone gets just doesn't work for them, giving them a plasmeme tank while fixing the problem doesn't feel right thematically
- Doesn't actually make for interesting gameplay, it's more frustrating than fun
- Doesn't solve the "problem" of people being able to just use a firesuit in space

So i've reverted it, instead i've made
- Getting emped gives hallucination, dizziness, and jittering
- Acid mod of 1.5
- Tempmod of 2 (they heat up and cool down by 2x the regular amount)
- Heatmod of 1.5
- They also give off smoke when too warm now (sorta like ai cores)

:cl:  
rscdel: IPC no longer have janky lungs
tweak: IPC now heat up faster, as well as take more damage from both heat and acid
/:cl:

### DIFF
--- a/code/_onclick/hud/alert.dm
+++ b/code/_onclick/hud/alert.dm
@@ -115,11 +115,6 @@
 	desc = "You're not getting enough oxygen. Find some good air before you pass out! The box in your backpack has an oxygen tank and breath mask in it."
 	icon_state = "not_enough_oxy"
 
-/atom/movable/screen/alert/not_enough_oxy/ipc
-	name = "Overheating"
-	desc = "You're not able to disperse heat. Find some gas to cool back down!"
-	icon_state = "overheating"
-
 /atom/movable/screen/alert/too_much_oxy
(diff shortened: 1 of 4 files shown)


Potototatoh:
yeah it seemed like some parts of the old PR didnt even work properly ingame

MajManatee:
I like the idea of IPCs smoking in the halls

Moltijoe:
Sarah's making a cooler rework of ipc lungs and this getting merged would add annoying conflicts for her

PR #16509, opened 2022-11-14
Fixes both arm implant dropping and conjure spells

### LINKED ISSUE #16508: Tool implant is semi-usable without arms
## Test Merges: 
Not Supplied / None
## Reproduction:
You can pull tools out of the implant without arms but are stuck with them out, this happened to Vast Atlas
## Supplied Image:

SuppliedImage
 Submitted by: Vance Pants (973199365466382396)

Moltijoe:
actually does what #16393 says
turns out it was a problem with how the conjure item spell worked and not signals
so instead i just made it not call the signal if there is no item

closes #16508

:cl:  
bugfix: COMSIG_ITEM_PREDROPPED is only called if there is an item to drop
/:cl:

### DIFF
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -298,7 +298,8 @@
 //for when you want the item to end up on the ground
 //will force move the item to the ground and call the turf's Entered
 /mob/proc/dropItemToGround(obj/item/I, force = FALSE, silent = FALSE)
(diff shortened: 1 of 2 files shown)


Moltijoe:
working on a pr to fix both problems

Moltijoe:
Didn't take as long as i expected it to

ToasterBiome (approved):

PR #16524, opened 2022-11-15
Reworks Mech pilot suits to instead be a component

Moltijoe:
It was hard coded as specifically just that Mech pilot suit, with a hard coded variable ontop of that.
Now it's a clothing variable
Currently only Mech pilot suits make an effect, but this lets it be a variable used for gloves, jackets, helmet, etc
Multiple items stack multiplicatively, so no need to worry about negative speeds
This would also allow for clothing that makes you worse at piloting a mech

:cl:  
tweak: There is now a component for mech piloting speed
/:cl:


githubuser4141:
Would it be possible to rename the 'EVA'  stuff to something else? It's confusing and illogical that a non-spaceproof jumpsuit is referred to as an 'EVA suit' and that actual EVA suits don't have mech speed boosts, non that they should because that wouldn't make sense. Something like a 'exosuit underclothing' with an actually informative description would be good

Moltijoe:
> Would it be possible to rename the 'EVA' stuff to something else? It's confusing and illogical that a non-spaceproof jumpsuit is referred to as an 'EVA suit' and that actual EVA suits don't have mech speed boosts, non that they should because that wouldn't make sense. Something like a 'exosuit underclothing' with an actually informative description would be good

it's not actually called the EVA suit in game, i just keep thinking of it as the suit from evangelion
it's actually called something like "mech pilot suit" or something

Bibby0110 (changes requested):
This should just be a trait

Moltijoe:
> This should just be a trait

a trait and a clothing variable would act very differently, it wouldn't stack, it would always be a static number, it wouldn't be able to be negative. The only benefit i can see the trait having is it would make code simpler

githubuser4141:
> > Would it be possible to rename the 'EVA' stuff to something else? It's confusing and illogical that a non-spaceproof jumpsuit is referred to as an 'EVA suit' and that actual EVA suits don't have mech speed boosts, non that they should because that wouldn't make sense. Something like a 'exosuit underclothing' with an actually informative description would be good
> 
> it's not actually called the EVA suit in game, i just keep thinking of it as the suit from evangelion it's actually called something like "mech pilot suit" or something

I guess, I think it's still confusing for people that don't know all the code, because it's a naming conflict with the actual EVA suits and/or their spaceproofness. I dunno if the average person would immediately get 'oh it's a reference to an anime show with robots (there's 1000's of those)' 🤷

Moltijoe:
> I guess, I think it's still confusing for people that don't know all the code, because it's a naming conflict with the actual EVA suits and/or their spaceproofness. I dunno if the average person would immediately get 'oh it's a reference to an anime show with robots (there's 1000's of those)' 🤷

that's what i'm saying, it's not a naming conflict, no person will see them as EVA suit, that's just me using the wrong word for it here. They're actually called mech pilot suits, that's what everyone sees them as

githubuser4141:
> > I guess, I think it's still confusing for people that don't know all the code, because it's a naming conflict with the actual EVA suits and/or their spaceproofness. I dunno if the average person would immediately get 'oh it's a reference to an anime show with robots (there's 1000's of those)' 🤷
> 
> that's what i'm saying, it's not a naming conflict, no person will see them as EVA suit, that's just me using the wrong word for it here. They're actually called mech pilot suits, that's what everyone sees them as

you misinterpreted what I said, people who bump into it in the code might be confused if they're not an anime aficionado
if you say so however it's not a big deal & I'm not a maintainer

Bibby0110:
> > This should just be a trait
> 
> a trait and a clothing variable would act very differently, it wouldn't stack, it would always be a static number, it wouldn't be able to be negative. The only benefit i can see the trait having is it would make code simpler

Find a better way to do it than adding random vars to every piece of clothing. Doubt we will ever need this to be stackable

Moltijoe:
> you misinterpreted what I said, people who bump into it in the code might be confused if they're not an anime aficionado if you say so however it's not a big deal & I'm not a maintainer

it still says mech suit in the code

>Find a better way to do it than adding random vars to every piece of clothing. Doubt we will ever need this to be stackable

Best i can think of in the moment is giving each piece of clothing that's supposed to have it their own variable, and just checking to make sure the variable exists before using it.
Either that or I could limit it to just being the suit, jacket, and glothes, rather than every single clothing piece.

MajManatee:
i am already planning on finding ways to make this stackable for the best mech-speed possible
Some sorta headgear that improves your abilites too maybe...

Moltijoe:
> i am already planning on finding ways to make this stackable for the best mech-speed possible Some sorta headgear that improves your abilites too maybe...

i think the whole idea is that bibby doesn't want it to be stackable

Moltijoe:
There's no other way i can think of that would allow this to be as flexible as a clothing variable
only thing i think would be a reasonable change is to restrict it to only certain clothing slots rather than all pieces of clothing

JamieD1:
I dont like how its clothing, it means random people can steal the clothing and instantly be better.

Moltijoe:
> I dont like how its clothing, it means random people can steal the clothing and instantly be better.

that requires them taking out the mech and killing the person inside, i'd argue at that point it's fair game

I could also allow the component to be applicable to people, so there's a version that can't just be removed
that way we could also add a "mech pilot crash course" booklet to nukies that trains someone to be a better mech pilot

JamieD1:
> > I dont like how its clothing, it means random people can steal the clothing and instantly be better.
> 
> that requires them taking out the mech and killing the person inside, i'd argue at that point it's fair game
> 
> I could also allow the component to be applicable to people, so there's a version that can't just be removed that way we could also add a "mech pilot crash course" booklet to nukies that trains someone to be a better mech pilot

Clothing doesnt make you better at something, just make it on the people itself and not the clothes.

Moltijoe:
> Clothing doesnt make you better at something, just make it on the people itself and not the clothes.

i'd argue clothing doesn't make you better, but clothing _does_ help, think of it like some athletes using specific clothes to reduce drag. Proper clothing for a job helps people do that job better.

Also, if i got rid of the interaction that the evangelion mech pilot suit has, people would kill me.

I can reduce it to only the character and then jumpsuit slot if that's suitable for you

Simplehorror:
> > > I dont like how its clothing, it means random people can steal the clothing and instantly be better.
> > 
> > 
> > that requires them taking out the mech and killing the person inside, i'd argue at that point it's fair game
> > I could also allow the component to be applicable to people, so there's a version that can't just be removed that way we could also add a "mech pilot crash course" booklet to nukies that trains someone to be a better mech pilot
> 
> Clothing doesnt make you better at something, just make it on the people itself and not the clothes.

this is already in the game, it just makes the code not hacky

JamieD1 (approved):

PR #16529, opened 2022-11-15
Makes atmos holofan barriers fireproof

SapphicOverload:

Holofan barriers will no longer get destroyed by fire, which makes sense because it's meant to contain fire

# Changelog

:cl:  
tweak: atmos holofan barriers are now fireproof
/:cl:


patpol4:
You could argue the sign is meant to be fireproof, not the device

ShoziOffical:
Rename title to holo firelock barrier to prevent confusion with the holofans item

JamieD1 (approved):

PR #16545, opened 2022-11-15
Change some reinforced windows to normal windows

Simplehorror:

# Spriting

We buffed reinforced windows, but then we forgot to change some windows that don't need to be reinforced to normal windows. Why does secondary tool storage need reinforced windows which take a minute to deconstruct? If we overuse reinforced windows, they become the norm and that's unhealthy for the game.

# Wiki Documentation

Map thumbnails
# Changelog

:cl:  
mapping: Change some reinforced windows to normal windows
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -2726,7 +2726,7 @@
 /obj/effect/turf_decal/bot_white,
 /obj/machinery/smoke_machine,
 /obj/item/reagent_containers/glass/beaker/large/silver_sulfadiazine{
-	list_reagents = list(/datum/reagent/consumable/condensedcapsaicin = 100);
+	list_reagents = list(/datum/reagent/consumable/condensedcapsaicin=100);
 	name = "liquid pepper spray"
 	},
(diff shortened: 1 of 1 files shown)


Bibby0110:
You forgot to say which windows you changed.

ToasterBiome:
These are all places greyshits should not be breaking into, and if you're antag you have ways to break these windows.

patpol4:
I agree with your point but what bibby said, give a picture and explanation for each of the changes

ynot01:
Why would genetics' public facing door be a normal window but the ones facing medical are reinforced?

Simplehorror:
> Why would genetics' public facing door be a normal window but the ones facing medical are reinforced?

oversight, would fix but instaclosed


> I agree with your point but what bibby said, give a picture and explanation for each of the changes

no point now

Simplehorror:
> These are all places greyshits should not be breaking into, and if you're antag you have ways to break these windows.

how many times has someone died sitting outside of medbay lobby because they couldnt break a window

patpol4:
At this point baiomus opinion here takes priority.

PR #16563, opened 2022-11-16
Change the power sink sprite back to the old one

Simplehorror:

The current power sink sprite is visually indistinct and doesn't really make it clear what it is doing aside being an ominous circle. The old power sink sprite is very much obvious that it is evil (you know it when you see it) and I think it is an objectively better sprite from a design perspective. I made it so the energy harvester uses the new powersink sprite so they're visually distinct items instead of freaking the AI out.
# Spriting

image

# Wiki Documentation

Change the powersink sprite
# Changelog

:cl:  
imageadd: Change the power sink sprite back to the old one
imageadd: Energy harvester keeps the current sprite
/:cl:


KillerOrcaCora:
This sprite is old and ugly, rather keep it to how it is.

Simplehorror:
old one is 100x better than the generic circle we have today it screams evil

Sniblet:
Time for another day of security asking captain if I'm good to be executed for hooking up the energy harvester

PR #16570, opened 2022-11-16
Nerf reinforced windows

Simplehorror:

These fuckers are used in EVERY WINDOW ON THE MAP instead of being niche security windows, they're protecting the medbay lobby ffs. Since I'm not allowed to remove the dumb ones from the map I've gone and nerfed them overall by about 40%. (they take almost a solid minute to break)
# Spriting

# Wiki Documentation

# Changelog

:cl:  
tweak: Nerf reinforced windows so they don't take a minute to open
/:cl:

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -417,7 +417,7 @@
 			if(I.tool_behaviour == TOOL_WELDER && user.a_intent == INTENT_HARM)
 				user.visible_message(span_notice("[user] holds \the [I] to the security screws on \the [src]..."),
 										span_notice("You begin heating the security screws on \the [src]..."))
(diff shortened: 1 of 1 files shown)


Simplehorror:
To make it clear, decon time was nerfed. I want to lower the lower bound of damage for them as well but I don't think that'll get in

Moltijoe:
this cuts off 14 seconds, considering how i never see people actually deconstruct them and instead just break them, i think this should be fine. I would personally consider having the welding step be a bit longer than only 8 seconds though.

patpol4:
sounds sensible to me, i always hated rwindows

Sniblet:
yeah it's uhhhh a little weird that they're like four times as secure as solid steel

JamieD1 (approved):

PR #16590, opened 2022-11-18
Flamethrower rework

ynot01:

Flamethrowers no longer dispense their atmos to the turfs they ignite because atmos is hilariously violent and that area would be made uninhabitable for the rest of the round, making them pretty much unusable to all but the most suicidal or robust atmos techs

Flamethrowers now have a 0.8 second click delay to prevent spam (and instakilling with new damage calc)

Flamethrowers ignite and do damage to mobs, mechs, and blobs based on the moles dispensed in that (turf) ignition's moles (5% of the current tank)

Each turf that is ignited will drain 5% of the tank's moles and those moles are used for immediate damage calculation

At a 10 mole tank (0.5 mole ignition), or about roundstart unfilled (~8.5 moles), it will deal 8 damage

At a 20 mole tank (1 mole ignition), each will deal 16 damage

Damage is capped at 16

Finally applying that tank pressure minimum: Once a flame would do less than 4 damage (<0.25 moles in a flame or 5 moles in the tank), the flamethrower automatically shuts off (and stops pointlessly calculating turfs)

The more moles of flammable gas a tank has the longer it will last (with diminishing returns, of course)

The aim was to create something that experienced atmos techs could use as a robust weapon while not invalidating its use when simply popping in a plasma tank that you filled at a canister

**Maximum damage output can be achieved with controlled close range ignition, as longer flame trails will quickly drain your tank and thus reduce your on-hit damage**

Flames will not ignite if a turf does not have at least 10 moles of oxygen in it

# TL;DR

Flamethrowers no longer burn entire hallways

Flamethrowers now have immediate damage applied depending on the tank's remaining flammable gas mole count (each turf ignited takes away 5% and uses that for calculation)

Flamethrowers now shut off automatically when they don't have enough flammable gas to fire anymore

You can now use flamethrowers against blob without being banned

#

Also flamethrowers can be crafted again because they no longer delete entire hallways

# Changelog

:cl:  
tweak: Flamethrowers have been restored to their former glory
/:cl:


ShoziOffical:
so it only uses plasma gas?

ynot01:
> so it only uses plasma gas?

yes

ShoziOffical:
> > so it only uses plasma gas?
> 
> yes

make it works with hydrogen and tritium because those gases are flammable when reacted with oxygen

ynot01:
> > > so it only uses plasma gas?
> > 
> > 
> > yes
> 
> make it works with hydrogen and tritium because those gases are flammable when reacted with oxygen

Ok but first you need to tell me what `specific_heat` means on `/datum/gas` because I don't know

warface1234455:
> > > > so it only uses plasma gas?
> > > 
> > > 
> > > yes
> > 
> > 
> > make it works with hydrogen and tritium because those gases are flammable when reacted with oxygen
> 
> Ok but first you need to tell me what `specific_heat` means on `/datum/gas` because I don't know

specific_heat = 10 for trit
specific_heat = 15 for hydrogen

You know you can just look into it right..

ShoziOffical:
> > > > so it only uses plasma gas?
> > > 
> > > 
> > > yes
> > 
> > 
> > make it works with hydrogen and tritium because those gases are flammable when reacted with oxygen
> 
> Ok but first you need to tell me what `specific_heat` means on `/datum/gas` because I don't know

its heat capacity

Heat Capacity: A gasmix has heat capacity, and it is calculated by taking into account the quantity of all of the gases in the air and their specific heat. Heat capacity defines how much energy it takes to raise the temperature of a gas. The normal air mix (%30 O2, %70 N2) has a specific heat capacity of about 20 which doesn't impede heat transfer very much. Fires spreads quicker in gases with low heat capacity, and slower in gases with high heat capacity. (from wiki)

ShoziOffical:
maybe trit and hydrogen will do more burn damage

ynot01:
> maybe trit and hydrogen will do more burn damage

16 damage is as far as I'm willing to go

CMOisLing:
Change_in_Temp = Moles_of_gas * Specific_Heat * Change_in_energy

ShoziOffical:
> Change_in_Temp = Moles_of_gas * Specific_Heat * Change_in_energy

Show where this is in code

CMOisLing:
> > Change_in_Temp = Moles_of_gas * Specific_Heat * Change_in_energy
> 
> Show where this is this in code

No idea where it is but thats the irl use of specific heat. And how the code should do the calculation

JamieD1 (approved):

VaelophisNyx:
flamethrowers slaughtered

PR #16592, opened 2022-11-18
Explosive Wheelchair

Moltijoe:
A few people were saying something like this would be funny, so i threw it together real quick.

A very "stealthy" bomb since it's identical to a wheelchair in every way unless you're the one riding it.
Only martyrs can buy it, and even then they can only buy one.

gives a sound and text warning when the bomb's timer is activated, better get away quickly lest it blows your face off

:cl:  
rscadd: Adds an explosive wheelchair to traitor uplink
/:cl:


ToasterBiome:
Bye Koko

JamieD1 (approved):

PR #16594, opened 2022-11-18
Organic limbs can be disabled again

Moltijoe:
So when someone did a huge rework of the damage system, they forgot to make it possible for organic limbs to get disabled for some reason
This effectively made disarming organics with stamina damage or hell, even any damage, impossible.

This works exactly as would be expected, if someone has a problem with it, ask whomever made the original pr to fix it.

:cl:  
tweak: Organic limbs are no longer immune to disabling
/:cl:


Mqiib:
"But wounds are supposed to handle limb disabling not damage" yeah well how often have you seen a limb actually disabled by a wound? Like, single digit numbers. How often do people have max damage on a limb? A LOT! A WHOLE LOT! If limbs cant be disabled by damage then having max damage on a limb is a GOOD THING because that means you can't take any more damage on that limb! It makes you functionally immortal for no downside as long as the attacks hit that limb! 
Thank you molti!

SomeguyManperson:
Regardless of whether or not it's "good" it was an intentional part of wounds for limbs to not get disabled by physical damage so this is more of a tweak than a bugfix

Moltijoe:
> Regardless of whether or not it's "good" it was an intentional part of wounds for limbs to not get disabled by physical damage so this is more of a tweak than a bugfix

i suppose, but the problem comes when trying to attack a limb already past it's max damage cap, it still gets attacked, but doesn't take damage. this allows for limbs to soak theoretically infinite damage without you dying. This, i would argue is a bug. In contrast to a limb that gets disabled, it stops getting targeted and no damage is lost.

Assuming someone else goes through and redoes wounds or damage, they can just change what i've done here

SomeguyManperson:
> > Regardless of whether or not it's "good" it was an intentional part of wounds for limbs to not get disabled by physical damage so this is more of a tweak than a bugfix
> 
> i suppose, but the problem comes when trying to attack a limb already past it's max damage cap, it still gets attacked, but doesn't take damage. this allows for limbs to soak theoretically infinite damage without you dying. This, i would argue is a bug. In contrast to a limb that gets disabled, it stops getting targeted and no damage is lost.

Disabled limbs still get targetted they are just disabled. The only time afaik a limb wont be targetted is if its missing entirely

Wounds are meant to be applied more commonly/intensely to fully damaged limbs but in combat this might not be noticable enough, considering how reliable/unreliable wounds already are, and the wounds that do disable limbs (specifically third tier bone wounds) are highly unlikely to be encountered without the limb simply being dismembered regardless

```
/// When this bodypart hits max damage, this number is added to all wound rolls. Obviously only relevant for bodyparts that have damage caps.
var/disabled_wound_penalty = 15```

So it does work as intended (even if the intended function is stupid)

JamieD1:
As long as this doesnt bring back the bug where you could shoot a disabler at an arm and not disable the person, that was fucking dumb.

Moltijoe:
> As long as this doesnt bring back the bug where you could shoot a disabler at an arm and not disable the person, that was fucking dumb.

this makes organic species function exactly like IPC and preternis, which can be properly disabled

JamieD1:
> > As long as this doesnt bring back the bug where you could shoot a disabler at an arm and not disable the person, that was fucking dumb.
> 
> this makes organic species function exactly like IPC and preternis, which can be properly disabled

As long as people aiming for an arm can still disable someone with the same amount of shots as chest.

Moltijoe:
> As long as people aiming for an arm can still disable someone with the same amount of shots as chest.

ah, sure enough, they can't, no clue how to fix that

Mqiib:
> > > As long as this doesnt bring back the bug where you could shoot a disabler at an arm and not disable the person, that was fucking dumb.
> > 
> > 
> > this makes organic species function exactly like IPC and preternis, which can be properly disabled
> 
> As long as people aiming for an arm can still disable someone with the same amount of shots as chest.

That's an unrelated issue

SomeguyManperson:
> > > As long as this doesnt bring back the bug where you could shoot a disabler at an arm and not disable the person, that was fucking dumb.
> > 
> > 
> > this makes organic species function exactly like IPC and preternis, which can be properly disabled
> 
> As long as people aiming for an arm can still disable someone with the same amount of shots as chest.

That's an unrelated issue

It is also one of the few reasons to not target limbs, since you can disable a limb faster than a person (and limbs will generally be less armored than the head or chest)

Moltijoe:
no out of scope

SomeguyManperson:
should still be labeled a tweak since limbs not being disabled by damage was an intended change with wounds

ToasterBiome:
I trust Theos

PR #16596, opened 2022-11-18
Can no longer wrap items with nodrop

Moltijoe:
This both lets people trick people into picking up an item they can't drop
and lets people drop items they shouldn't be able to drop

:cl:  
bugfix: Can no longer wrap items with nodrop
/:cl:


MajManatee:
Finally i no longer have to cut my own arm off when i get pranked

JamieD1 (approved):

PR #16605, opened 2022-11-19
Mecha no more EMP armor

ynot01:

## EMP changes

The ion rifle tends to feel like a wet noodle against mechs damage-wise. This is largely in part due to the `ENERGY` armor most mechs possess, which affects nothing but EMPs due to mechs not respecting stamina damage.

All energy armor has been stripped from mechs and EMP damage has been boosted by 5/10 damage (light/heavy)

# Changelog

:cl:  
tweak: Stripped all mechs of EMP armor and slightly buffed EMP damage to mechs
/:cl:

### DIFF
--- a/code/game/mecha/combat/combat.dm
+++ b/code/game/mecha/combat/combat.dm
@@ -2,7 +2,7 @@
 	force = 30
 	internals_req_access = list(ACCESS_MECH_SCIENCE, ACCESS_MECH_SECURITY)
 	internal_damage_threshold = 50
-	armor = list(MELEE = 30, BULLET = 30, LASER = 15, ENERGY = 20, BOMB = 20, BIO = 0, RAD = 0, FIRE = 100, ACID = 100)
+	armor = list(MELEE = 30, BULLET = 30, LASER = 15, ENERGY = 0, BOMB = 20, BIO = 0, RAD = 0, FIRE = 100, ACID = 100)
 	mouse_pointer = 'icons/mecha/mecha_mouse.dmi'
 	destruction_sleep_duration = 40
 	exit_delay = 40
--- a/code/game/mecha/combat/durand.dm
+++ b/code/game/mecha/combat/durand.dm
@@ -6,7 +6,7 @@
 	dir_in = 1 //Facing North.
 	max_integrity = 400
 	deflect_chance = 20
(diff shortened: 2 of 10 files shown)


Simplehorror:
just make them cost more 4head, 3 mechs in a round shouldn't happen

ynot01:
> just make them cost more 4head, 3 mechs in a round shouldn't happen

3 mechs or 1 does not really matter. A good pilot can dominate the round. Making mechs is already a bit tedious, and massive material costs do nothing but create problems as I have learned in past changes. A roboticist just as well is not going to care how he impacts the material economy.

Additionally, if 3 mechs coordinate well and bust into an area, then I say more power to them. The issue currently is that you have to do several hoops just to fight ONE mech. With this, fighting multiple mechs will become a lot easier as a byproduct of making mechs generally weaker.

ToasterBiome:
I don't much agree with the slowdown but the energy armor reduction is a good idea

Moltijoe:
> I don't much agree with the slowdown but the energy armor reduction is a good idea

in contrast, i like slowdown, because it lets us add more items that speed them up like the evangelion mech pilot suit
That way, a mech can initially seem really fast and imposing, but damaging it can bring it back to reasonable numbers or even lower

side note, #16524

Mqiib:
You realize that slowing them down even more means that at ~50% integrity they're basically useless right? Have you TESTED to see how slow a mech moves when their step_in is 6 or 8? You might as well just leave the mech behind at that point! And ions should reasonably do less damage against certain types of mechs like combat mechs because they're reinforced to survive combat situations that might include ion guns. On top of that, ions deal a perfectly respectable amount of damage to mechs if you're not shooting them from the FRONT. It doesn't really need to get increased.

My suggestion: throw all this out and make EMPs apply a stacking slowdown that fades after like 5 seconds or something. That way if you blast a mech with 5 ion bolts it will have an extremely hard time doing anything to fight you and will allow you to either run behind it and shotgun it in the ass for 100 damage a pop or just run away.

Also if you're going to do slowdowns at least make it an additive slowdown not multiplicative because that screws over slower mechs like durands and maulers significantly more.

ynot01:
Atomized #16609

JamieD1 (approved):

PR #16609, opened 2022-11-19
Mecha damage slowdown

ynot01:

## Damage slowdown

At below 75% integrity, mechs will start to feel damage slowdown

The formula is 0.75 / (obj_integrity/max_integrity), capped at 2.5x slowdown (at 30% integrity)

# Changelog

:cl:  
rscadd: Added damage slowdown for mechs
/:cl:

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -616,13 +616,18 @@
 
 	var/move_result = 0
 	var/oldloc = loc
+	// Damage slowdown
+	// Slowdown starts at below 75% integrity
+	// Builds based on how damaged you are below 75%
+	// Max slowdown (2.5x slower) at 30% integrity
+	var/slowdown = clamp(0.75 / (obj_integrity/max_integrity), 1, 2.5)
 	if(internal_damage & MECHA_INT_CONTROL_LOST)
-		set_glide_size(DELAY_TO_GLIDE_SIZE(step_in * (check_eva() ? EVA_MODIFIER : 1)))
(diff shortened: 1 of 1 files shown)


Marmio64:
feel like mechs should get slightly more health/armor to balance this, since this combined with internal damages is a bit much.

Moltijoe:
> feel like mechs should get slightly more health/armor to balance this, since this combined with internal damages is a bit much.

internal damages are kinda wonky and inconsistent, i'd rather just replace them with this

githubuser4141:
> > feel like mechs should get slightly more health/armor to balance this, since this combined with internal damages is a bit much.
> 
> internal damages are kinda wonky and inconsistent, i'd rather just replace them with this

internal damage was a RNG chance to happen when a mech under it's internal threshold (35% for a gygax, 50% for a durand, ~25% for Phazon/Antag mechs).
 
So even if you were under the threshold it was only an RNG chance to happen, and if you were in a hardsuit + oxygen you could actually ignore some of them (MECHA_INT_FIRE,MECHA_INT_TEMP_CONTROL,MECHA_INT_TANK_BREACH) would only affect people w/o hardsuits, or air. 

The few important damages were, short circuiting & randomized movement, & an RNG chance for a module to be deleted.
So. This is actually a nerf to mechs if it's being kept at 75% for *any* mech because all mechs earlier had their internals threshold at 60% or above and it's not only guaranteed to happen but it's guaranteed to actually have a negative impact.

JamieD1:
When we removing their bullshit 100% acid armour?

ynot01:
can be done with a better algorithm later

PR #16621, opened 2022-11-19
Removes research access block and replaces it with worse researching if you're not a scientist

ynot01:

SKILLS ARE REAL

This is several times better than complete lockout

Read changelog

# Changelog

:cl:  
tweak: You no longer require access to research things at an R&D console
tweak: Researching things are 1.25x more costly if you're not a member of science, silicon, special role, or captain
/:cl:

### DIFF
--- a/code/modules/research/techweb/_techweb_node.dm
+++ b/code/modules/research/techweb/_techweb_node.dm
@@ -17,6 +17,7 @@
 	var/category = "Misc"				//Category
 	var/ui_x = 805 // It's location - override this in techweb_layout.dm
 	var/ui_y = 805
+	var/list/approved_jobs = list("Scientist", "Roboticist", "Research Director", "Captain", "AI", "Cyborg")
 
 /datum/techweb_node/error_node
 	id = "ERROR"
@@ -93,4 +94,9 @@
 		return research_costs
 
 /datum/techweb_node/proc/price_display(datum/techweb/TN)
-	return techweb_point_display_generic(get_price(TN))
+	var/list/price = get_price(TN)
+	price = price.Copy()
+	if(!(usr?.job in approved_jobs) && !is_special_character(usr))
+		for(var/L in price)
+			price[L] *= 1.25
+	return techweb_point_display_generic(price)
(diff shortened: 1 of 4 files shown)


Simplehorror:
but miners don't NEED to be able to research mining stuff if there's no scientist dude!!

N3D6:
just do away with the whole fucking thing, it's a shitty code change to what is very much an admin issue, still, this is a lot better than straight up locking people out of it

RG4ORDR:
Yes good

ynot01:
> it's a shitty code change to what is very much an admin issue

This is game design not a solution to an admin issue

Cartlord:
This just punishes Scientists for people breaking in.

Simplehorror:
can you please make it so it's just slower to research if you're not a scientist, like a 60 second do_mob or something instead of punishing scientists for a non-scientist action

code/modules/research/techweb/_techweb_node.dm
```diff
@@ -17,6 +17,7 @@
 	var/category = "Misc"				//Category
 	var/ui_x = 805 // It's location - override this in techweb_layout.dm
 	var/ui_y = 805
+	var/list/approved_jobs = list("Scientist", "Roboticist", "Research Director", "Captain", "AI", "Cyborg")
```
Bibby0110, line 20: Not a fan of using a list of strings like this, should piggyback off the GLOBs of positions so we’re sure they get updated in case we add a new science job. (You forgot network admin btw)

ynot01: Who

CrazyDrakkon:
Does "special role" include free golems and free miners?

Sniblet:
Transferred scientists are affected by this

ynot01:
> Transferred scientists are affected by this

Getting a new job does not make you immediately proficient at that job

code/modules/research/techweb/_techweb_node.dm
ynot01: Can you direct me towards this global that stores departmental positions if it exists

PR #16622, opened 2022-11-19
Mining medic and Brig phys pda start with lifeline and suit monitor

Moltijoe:
They're basically paramedic for lavaland or paramedic for security, so they get the same starting programs paramedic does

:cl:  
rscadd: Mining Medic and Brig physician PDA start with Lifeline and suit monitor pre installed
/:cl:


iamthesilly:
brig physician should as well

JamieD1 (approved):

PR #16628, opened 2022-11-20
Pipe bomb in your mail

Moltijoe:
A fun surprise for the whole family!
Use it once to select the target
one the target opens it, there is a 2.5 second delay before it explodes
it starts in their hand, with throw mode already one, so all they need to do is click
Rather small explosion so it doesn't do much collateral damage

Makes the lootbox that is mail more thrilling

:cl:  
rscadd: Adds a traitor cargo/qm item that's just a bomb in a letter
/:cl:

### DIFF
--- a/code/game/objects/items/grenades/letterbomb.dm
+++ b/code/game/objects/items/grenades/letterbomb.dm
@@ -71,6 +71,9 @@
 	icon_state = initial(icon_state)//there's no active icon for pipe bombs, so just force revert it to the default
 
 /obj/item/grenade/mailbomb/prime()
+	if(ishuman(loc))//special snowflake check to make it stronger than a light, but weaker than a heavy explosion. but only to the person if they can't throw it away in time
+		var/mob/living/carbon/human/H = loc
+		H.take_overall_damage(30,30)
 	update_mob()
-	explosion(src.loc,0,1,1)	//targeted, but if they can throw it away in time, little overall collateral damage
+	explosion(loc,0,0,2, flame_range = 2)	//targeted, but if they can throw it away in time, little overall collateral damage
 	qdel(src)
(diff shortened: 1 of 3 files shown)


code/game/objects/items/grenades/letterbomb.dm
```diff
@@ -0,0 +1,75 @@
...
+	icon_state = "pipebomb"
+	lefthand_file = 'icons/mob/inhands/equipment/security_lefthand.dmi'
+	righthand_file = 'icons/mob/inhands/equipment/security_righthand.dmi'
+	throw_speed = 3
+	throw_range = 7
+	flags_1 = CONDUCT_1
+	slot_flags = ITEM_SLOT_BELT
+	active = 0
+	display_timer = 0
+	det_time = 2.5 SECONDS //better throw it quickly
```
Simplehorror: way too high to be useful for 4 TC

Simplehorror:
make it so i can put a syndibomb in a crate that detonates when it arrives at sec

code/game/objects/items/grenades/letterbomb.dm
Moltijoe: good idea, i'll reduce it to 1 TC since it's not even guaranteed to be opened

Moltijoe:
> make it so i can put a syndibomb in a crate that detonates when it arrives at sec

no

UltraBigRat:
How deadly is the explosion when it explodes in someones hand?

ShoziOffical:
> How deadly is the explosion when it explodes in someones hand?

can instantly crit someone and leaves a space hole according to the code

Moltijoe:
> can instantly crit someone and leaves a space hole according to the code

yeah, 60 burn, 60 brute and shuttle damage ranges from destroying the tile, to going right down to space

honestly, if there was a step between heavy and light explosions, i'd use that instead
120 from heavy feels like too much
but 24 from light feels like too little

hold on, lemme actually reduce it to just a light explosion, but also add some flame range to include a bit of burn

Moltijoe:
yeah, no, light is too weak
lemme make a little snowflake check and see if it's better

Moltijoe:
there, now it does a total of 84 damage, without risk of spacing the station or damaging clothes

JamieD1 (approved):

PR #16645, opened 2022-11-21
Changes position of the MIX to Engine pump and pipes on GAXStation so they don't need re-piping shift start

Annoying-plant:

Removes the unnamed gas pump to engine that shared a pipe with the gas to incinerator pump and adds a mix to engine pump on the mix loop instead linked up to the SM so that atmos doesn't need re piping when something needs to be added to the SM (or a Trit mix isn't accidently piped)

this has been added where the old connector was so this has been moved to the opposite side allowing acess from a connector still as well as use of the mix pump to engine

# Changelog

:cl:  
rscadd: added new engine pipe in a position that actually makes sense 
rscadd: added connector to the opposite side 
rscdel: removed old engine pipe removed old connector
rscdel: removed old connector
/:cl:


ToasterBiome:
your description is in the wrong place and if there’s no wiki or sprite stuff you can delete all of it

ToasterBiome:
So why is this a good idea? All I see is a pipe accessible from maint.

Annoying-plant:
> 

good point actually okay im going to change its location again but its current location makes it not difficult to use

Annoying-plant:
okay so ive moved the location so its no can no longer be accessed via maints but you can now still pump mix to engine rather than pumping pure to engine down same path as incinerator

Annoying-plant:
the reason i change it is cause on all other maps you pump mix to engine not pure to engine and with it being directly on line with incinerator pump there is always a trit mix going to incinerator making it not really usefull most of the time

applepineapplepizza:
since you removed a connector to place the pipeline from mix to engine, can you add one back somewhere?
can be anywhere, maybe inbetween the freezers.

Annoying-plant:
> 

yeah sure can do will have a look

Annoying-plant:
Okay so ive now added in the connector on the opposite side so the mix pump is at the north side of atmos and the connector is at the south side

ToasterBiome (changes requested):
map diff has a weirdly large area changed, might need to look at if anything is being changed over there

Annoying-plant:
> map diff has a weirdly large area changed, might need to look at if anything is being changed over there

how do I check this I'm quite new to this sorry only my second ever map change?

Annoying-plant:
okay i think i know what ive done im going to close this for now and re open another once ive fixed the issues

PR #16654, opened 2022-11-22
fixes some bugs with digitigrade leg swapping

Moltijoe:
Can use any screwdriver like tool to swap between prosthetic leg types rather than just specifically the screwdriver item
fixes swapping them while they're in pockets or backpack making things buggy

:cl:  
bugfix: fixes some bugs with robotic digitigrade leg swapping
/:cl:

### DIFF
--- a/code/modules/surgery/bodyparts/robot_bodyparts.dm
+++ b/code/modules/surgery/bodyparts/robot_bodyparts.dm
@@ -196,7 +196,7 @@
 	if(spot == user && !user.get_inactive_held_item())
 		user.put_in_inactive_hand(prosthetic)
 	else
-		prosthetic.loc = get_turf(user)
+		prosthetic.forceMove(get_turf(user))
 	qdel(src)
 	
 /obj/item/bodypart/r_leg/robot/attackby(obj/item/W, mob/user, params)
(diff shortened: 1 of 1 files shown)


neeshacark:
oh thank mqiib my lizard feet are saved

code/modules/surgery/bodyparts/robot_bodyparts.dm
```diff
@@ -193,14 +193,14 @@
 	
 	var/spot = src.loc
 	moveToNullspace()
-	if(spot == user)
+	if(spot == user && !user.get_inactive_held_item())
 		user.put_in_inactive_hand(prosthetic)
 	else
-		prosthetic.loc = spot
+		prosthetic.loc = get_turf(user)
```
ToasterBiome: i guess technically you should be using forcemove here?

ToasterBiome (changes requested):

code/modules/surgery/bodyparts/robot_bodyparts.dm
Moltijoe: sounds good, know of any better way to do the location thing though? i'd rather being able to do it in pockets/backpack but it gets all wonky. This change just makes it drop at the floor rather than actually let it work

JamieD1 (approved):

PR #16657, opened 2022-11-22
Makes the preternis teslium interaction possibly worth using

Moltijoe:
Couple reasons it wasn't worth using
- It only healed if you were between 0 and 50 hp (impossible)
- The insane burn mod made literally any burn damage instantly kill you
- you needed to drink at least 10u for it to actually work

It now heals all the time (why not, since your health with this in you becomes binary)
the burn mod was reduced from 200 to 10 (still enough for even welders to instacrit)
and it now works regardless of how much you drink

:cl:  
tweak: Drinking teslium as a preternis is slightly less of a bad idea
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/preternis.dm
+++ b/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/preternis.dm
@@ -160,7 +160,7 @@ adjust_charge - take a positive or negative value to adjust the charge level
 		H.adjustFireLoss(-1*REAGENTS_EFFECT_MULTIPLIER,FALSE,FALSE, BODYPART_ANY)
 
 	if(H.reagents.has_reagent(/datum/reagent/teslium))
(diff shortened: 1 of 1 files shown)


JamieD1:
200 to 10?

Moltijoe:
> 200 to 10?

anything above ~15 burn mod functions the same as 200
one single point burn, instantly husked with 200 burn mod (preterni can't be husked but you get the idea)
no proper forms of damage actually did that little burn though, most do above 10, 
welding which is the most common is 15, basic laser guns do 20
Even with a burn mod of 10, that's still insta-crit for welding tools, and insta-death for laser guns
even with signficant armour reducing the damage, it'll still be enough to cause damage slowdown

this should only reduce the number of accidental deaths while under teslium, while not changing how quickly you get fucked in actual combat

JamieD1 (approved):

PR #16664, opened 2022-11-23
Traitors Among Traitors: Adds Sole Survivor objective + exchange objective tweaks

BurgerLUA:
Adds new sole survivor objective where if two (or more) traitors get hijack, the non-first traitor (or traitors) will get an objective to be the **only** survivor on a shuttle.

Exchange objective minimum traitor count requirements were reduced from 8 to 6 in hopes that it will occur more often.

# Details

Description of the objective:
_Escape on the shuttle to ensure **no one except you** escapes alive and out of custody._

If traitor A rolls for objectives and gets a hijack objective, then they will get the hijack objective.
If traitor B rolls for objectives and doesn't get a hijack objective, then they will not get the hijack objective.
if traitor C rolls for objectives and gets a hijack objective, then INSTEAD they will get a sole survivor objective.
if traitor D rolls for objectives and gets a hijack objective, then INSTEAD they will get a sole survivor objective.
If traitor E rolls for objectives and doesn't get a hijack objective, then they will not get the hijack objective.

So basically if in the rare chance that two or more traitors get hijack, the first traitor will get the regular hijack objective but the second traitor (and so on) will not get it but instead the sole survivor objective. The chance of this happening is very uncommon but it can still happen and I'm sure will be quite fun for the whole family.

The reasoning behind this change is that too many a time traitors team up with other traitors with hijack and end up stomping the entire station. It's a little lame and I think I want to counter that more by making that a little less safer than normal by making it so that sometimes, traitors can get an objective to kill EVERYONE on the shuttle (except themselves) in an attempt to hijack it.

Also inb4 someone tells me that /datum/antagonist/traitor/ and /datum/objective/ are no longer used in favor of dynamic and this PR is meaningless.

# Wiki Documentation

https://wiki.yogstation.net/wiki/Traitor needs to be updated to include the new objective.

# Changelog

:cl:  BurgerBB
rscadd: Adds new sole survivor objective where if two (or more) traitors get hijack, the non-first traitor (or traitors) will get an objective to be the only survivor on a shuttle.
tweak: Changes the exchange objective minimum traitor count from 8 to 6.
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/datum_traitor.dm
+++ b/code/modules/antagonists/traitor/datum_traitor.dm
@@ -97,7 +97,7 @@
 		is_hijacker = prob(10)
 	var/martyr_chance = prob(20)
 	var/objective_count = is_hijacker 			//Hijacking counts towards number of objectives
-	if(!SSticker.mode.exchange_blue && SSticker.mode.traitors.len >= 8) 	//Set up an exchange if there are enough traitors
+	if(!SSticker.mode.exchange_blue && SSticker.mode.traitors.len >= 4) 	//Set up an exchange if there are enough traitors. YOGSTATION CHANGE: 8 TO 4.
 		if(!SSticker.mode.exchange_red)
 			SSticker.mode.exchange_red = owner
 		else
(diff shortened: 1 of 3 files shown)


Simplehorror:
thank you burger jesus

ChuChu42:
gonna be honest thats a pretty fucking good idea

Hislyna:
seems a little sus

Simplehorror:
also dynamic isn't very "dynamic" it could be replaced with a monkey pressing buttons

Moja10:
What's wrong with two people having hijack very rarely? I would prefer if it was a chance to get sole survivor, rather than being a guarantee. Would make it more interesting.

Simplehorror:
i'd like to point out people can team up with traitors with hijack even if they don't have it

Bibby0110:
Make this compile and reopen, blessed

PR #16665, opened 2022-11-23
Heretics Among Heretics: Heretics have a higher chance to betray eachother.

BurgerLUA:
Heretics now have a 20% chance to target another Heretic per assassinate objective.

# Changelog

:cl:  
rscadd: Heretics now have a 20% chance to target another Heretic per assassinate objective.
bugfix: Fixes a bug where it was entirely possible to have two of the same assassinate or protect objectives of the same crewmember.
/:cl:

### DIFF
--- a/code/modules/antagonists/eldritch_cult/eldritch_antag.dm
+++ b/code/modules/antagonists/eldritch_cult/eldritch_antag.dm
@@ -107,8 +107,14 @@
 		ET.on_life(owner.current)
 
 /datum/antagonist/heretic/proc/forge_primary_objectives()
-	var/list/assassination = list()
-	var/list/protection = list()
+
+	//Yogstation change: Gathers heretics for the heretic betray/assassinate objective.
+	var/list/all_heretics = list()
+	for(var/datum/antagonist/heretic/H in GLOB.antagonists)
+		all_heretics += H.owner
+	//End of yogstation change.
+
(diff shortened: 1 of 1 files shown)


Simplehorror:
i don't think people go for objectives with heretic, they go for ascension. make it so you need to sacrifice another heretic to ascend

BurgerLUA:
> i don't think people go for objectives with heretic, they go for ascension. make it so you need to sacrifice another heretic to ascend

This was the original plan but I'm not really experienced with the code. Heretics have to choose their targets from a pool of 1 to 3 (Chosen randomly) of candidates to sacrifice. I would have to remove that system in order for that to work.

Simplehorror:
> > i don't think people go for objectives with heretic, they go for ascension. make it so you need to sacrifice another heretic to ascend
> 
> This was the original plan but I'm not really experienced with the code. Heretics have to choose their targets from a pool of 1 to 3 (Chosen randomly) of candidates to sacrifice. I would have to remove that system in order for that to work.

i can do it for you

UltraBigRat:
Not a fan of this honestly, heretics are already a pretty darn difficult antag with alot of ways of easily being fucked over. making them kill each other probably isnt a very good idea.

Bibby0110:
Make this compile and reopen, blessed

PR #16686, opened 2022-11-24
Remove slowdown from snow tiles

N3D6:

"What if we made a mining map, and it was a really cool and well designed mining map, BUT, we gimped all the miners with every block giving you a 2 slowdown"

Slow snowing you down sucks, and the main gripe people usually hold about ice stations is the incessant slowdown given to you by walking in snow.

# Spriting

# Wiki Documentation

# Changelog

:cl:  
tweak: snow no longer slows you down 
/:cl:

### DIFF
--- a/code/game/turfs/open.dm
+++ b/code/game/turfs/open.dm
@@ -208,7 +208,6 @@
 	icon = 'icons/turf/snow.dmi'
 	desc = "Looks cold."
 	icon_state = "snow"
-	slowdown = 2
 	bullet_sizzle = TRUE
 	footstep = FOOTSTEP_SAND
 	barefootstep = FOOTSTEP_SAND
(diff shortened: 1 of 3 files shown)


Simplehorror:
GIVE THEM SNOWSHOES!!!

N3D6:
> GIVE THEM SNOWSHOES!!!

As far as I’m aware there’s no way to do that. If you would like to try, and make shoes that take away slowdown on snow and ice only I’ll close this.

Annoying-plant:
> GIVE THEM SNOWSHOES!!!

would it be possible to make shoes with a reduction to slowdown only when your on snow tiles?
I actually really like simple horrors idea of snow shoes as someone who loves using geldakinesis to decorate areas with snow tiles.

N3D6:
> > GIVE THEM SNOWSHOES!!!
> 
> would it be possible to make shoes with a reduction to slowdown only when your on snow tiles? I actually really like simple horrors idea of snow shoes as someone who loves using geldakinesis to decorate areas with snow tiles.

I don't think we have support for that

MajManatee:
in RP our now is an inch deep, or maybe two. Either way that wouldn't slow anyone down.

JamieD1:
make it half as slow

N3D6:
> make it half as slow

Ice mining is still going to be painfully garbage then, there's no point

JamieD1 (approved):

PR #16690, opened 2022-11-24
abruptly makes augmentation kits much more expensive for "balance"

Skrem7:

Um hm hrm having nukies take 3 damage per shot is kinda problematic but you know someone should probably adjust the price of the elite suit too but for now this will do

# Wiki Documentation

Prices and gamemodes to be adjusted on Syndicate Items page

# Changelog

:cl:  
tweak: Nukies can no longer buy the default aug kit
tweak: The superior aug kit that nukies can only buy has had its price raised to 45 from 25 TC.
/:cl:


Moltijoe:
i appreciate they can still get discounted, that way we can still get the chrome nukies squad still

JamieD1 (approved):

PR #16691, opened 2022-11-24
Make the elite hardsuit 15 TC because Skrem didn't have the guts to do it

00ze-cyclone:

#16690

Elite Hardsuit is 15 TC now because it's just better hardsuit for incredibly cheap.
Since the elite is pricier than the regular now nuke ops can buy the basic syndie hardsuit for their reinforcements.

# Wiki Documentation

Elite hardsuit cost 15 TC now
Syndicate hardsuit can be bought by nuclear operatives

# Changelog

:cl:  The lzayness of balance coding
tweak: Elite hardsuit cost 15 TC now, nuke ops can buy the regular hardsuit
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1509,14 +1509,14 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			Nanotrasen crew who spot these suits are known to panic."
 	item = /obj/item/clothing/suit/space/hardsuit/syndi
 	cost = 8
-	exclude_modes = list(/datum/game_mode/nuclear, /datum/game_mode/infiltration) //you can't buy it in nuke, because the elite hardsuit costs the same while being better // yogs: infiltration
+	exclude_modes = list(/datum/game_mode/infiltration) // yogs: infiltration and nuke ops can buy their signature suit now
 
 /datum/uplink_item/suits/hardsuit/elite
 	name = "Elite Syndicate Hardsuit"
 	desc = "An upgraded, elite version of the Syndicate hardsuit. It features fireproofing, and also \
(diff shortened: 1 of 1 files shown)


TheRyeGuyWhoWillNowDie:
this is stupid because there's a big rack of syndicate hardsuits on the syndicate crusier, also makes the WT AP spam meta even better which sucks

Skrem7:
blood red hardsuit armor is 50 bullet opposed to 60 bullet, opinion ignored

Marmio64:
makes elite hardsuits completely unviable in stealthops, and makes it pale in comparison to the shielded hardsuit for anything other than fireproofness.

Skrem7:
would say "shielded price should be brought down" but unfortunately the desword exists

this thing should probably be priced at double digits but not above 12 TC; it's useful but not required but even a little more armor does go a long way when you consider how much damage is thrown at nukies and how they ideally don't want to lose momentum

00ze-cyclone:
> makes elite hardsuits completely unviable in stealthops, and makes it pale in comparison to the shielded hardsuit for anything other than fireproofness.

Elite is still half the price of shielded and provide reliable protection instead of a shield that will block 3 buckshot pellets and then abandon you as the rest of the bullets come to tear your body to shreds.
For stealthops lmao who care anyway, just kill them before they can even grab a gun to try to hurt you, you have the element of surprise and enough firepower to eliminate all opposition in the initial strike.

Simplehorror:
shielded hardsuit is fucking noob bait, easily the worst item in the game

Annoying-plant:
yeah I dont like this the hardsuit is good but not that good and this just limits peoples options more and more and makes it harder for players newer to antags/nukies to learn what items they like and make builds 8TC is already a decent costs imo

Simplehorror:
this has the same effect as removing 15 TC from warops

PR #16598, opened 2022-11-18
SHADOWLING REBALANCE 2022

Simplehorror:
# Document the changes in your pull request

## However, it is too HARD for new shadowlings to win.
I have taken this into account and designed a balance PR methodically that is to make it easier for newer shadowlings to win, as well as making it harder for competent shadowlings to steamroll.

1. Shadowlings can no longer enthrall mindshielded people. This is because I noticed that competent shadowlings would rush security officers or heads and convert them at round start instead of normal people, which was a guaranteed win if you got the Captain. Additionally, mindshields are completely useless in this gamemode because of the aforementioned issue. This makes it so they are a worthwhile buy.
2. I have decreased the cooldown on Icy Veins and Veil. This is to make escape easier, which noticeably helps newer players learn the basics, as well as letting them live for longer against a competent security force. I have seen new shadowlings die solo to an assistant in maintenance.
3. I have made "glare" auto select a target that is closest to you if you miss your shot. This is to make it easier for newer players as they may not always have the best aim, so it is equalizing aim skill. This makes it easier for newer players to get going because it has a very hefty cooldown if you miss, and if you miss the entire stealth portion of the gamemode is now over, which makes it almost impossible for newer groups to win.
4. Glare no longer works on mindshielded humans since it is a conversion tool. 
# Spriting

# Wiki Documentation

Can't enthrall mindshielded people
Glare no longer works on mindshielded humans since it is a conversion tool. 
# Changelog

:cl:  
rscadd: Glare now can auto-target if you miss in a radius
tweak: Shadowlings can't thrall mindshielded people
tweak: Veil and icy veins cooldown reduced
tweak: Glare no longer works on mindshielded humans since it is a conversion tool. 
bugfix: Fixed Destroy Engines being able to be used multiple times by different shadowlings
/:cl:

PR #16712, opened 2022-11-24
Shadowlings get a spell that extinguishes them on a 5 second cooldown

Simplehorror:

remove cheese tactic. They insta-die to dragonsbreath rounds if they get caught which is a notorious cheese tactic, and it makes shadowlings very hard to balance when it exists. They also have no existing way to remove fire outside of an extinguisher, which is just impossible to carry around anymore.

Sister PR to #16598
# Spriting

# Wiki Documentation

New shadowling spell: extinguish. 5 second cooldown
# Changelog

:cl:  
tweak: Shadowlings get a spell that extinguishes them on a 5 second cooldown
/:cl:

### DIFF
--- a/yogstation/code/game/gamemodes/shadowling/shadowling.dm
+++ b/yogstation/code/game/gamemodes/shadowling/shadowling.dm
@@ -194,7 +194,6 @@ Made by Xhuis
 	if(world.time >= charge_time+last_charge)
 		shadow_charges = min(shadow_charges + 1, 3)
 		last_charge = world.time
-	H.fire_stacks = 0
 
 /datum/species/shadow/ling/bullet_act(obj/item/projectile/P, mob/living/carbon/human/H)
 	var/turf/T = H.loc
--- a/yogstation/code/game/gamemodes/shadowling/shadowling.dm
+++ b/yogstation/code/game/gamemodes/shadowling/shadowling.dm
@@ -194,6 +194,7 @@ Made by Xhuis
 	if(world.time >= charge_time+last_charge)
 		shadow_charges = min(shadow_charges + 1, 3)
 		last_charge = world.time
+	fire_stacks = 0
 
 /datum/species/shadow/ling/bullet_act(obj/item/projectile/P, mob/living/carbon/human/H)
 	var/turf/T = H.loc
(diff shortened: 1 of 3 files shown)


ynot01:
carry a mini fire extinguisher

Simplehorror:
> carry a mini fire extinguisher

this is like if i removed emags and everyone complained but then i said "carry a toolbox"

noobs aren't gonna do this and it's another station item you're going to need for something that you should really have as a spell

Cartlord:
Where the fuck do you even GET mini extinguishers?

MajManatee:
> Where the fuck do you even GET mini extinguishers?

any autolaithe makes them

ynot01:
> this is like if i removed emags and everyone complained but then i said "carry a toolbox"

emags cost resources and are not innate, if it had no value over a toolbox then everyone would simply carry a toolbox

> noobs aren't gonna do this and it's another station item you're going to need for something that you should really have as a spell

if there is a line for hugbox i would think this crosses it, something as simple as carrying a fire extinguisher because you are vulnerable to fire is simple and not by any means arcane knowledge

i would focus fire on something else in shadowling's kit

if you're looking for suggestions:

- their ability to cross walls and move at ludicrous speeds; moving at such high speed is disorienting for new players; as strong as it is, it is hard to control. i would review this and their two jaunts, condensing it all down into one mechanic that is more in the player's face so that they are forced to acknowledge it rather than an off button, as well as being more fair to fight against (the original jaunt had a stamina damage system where you would take more damage in lit areas but it was basically inconsequential, and if you stamcrit the ability wouldn't go on cooldown)
- conversion is a pain in the ass, the spell sucks to use and it being nerfed does no favors, more experienced players work around this by crafting stun weapons but a new player would never think of this.

presently it feels like shadowling is a mess of asspull buttons rather than a coherent antag with a gameplay loop, i believe darkspawn executed it a lot better; adding one more button doesn't help new players and only serves to further overwhelm them

Sniblet:
Integrating this into veil and icy veins helps to negate button bloat, and most of the time when you're on fire it's in a situation when you'd want to turn off some lights anyway

I don't agree that forcing slings to do a quick chore at the start of every round or else die contributes to the game in a positive way

KillerOrcaCora (changes requested):
Good suggestion, mix this with the icy vines ability, that way we don’t have to have 2 icons that are probably gonna look very similar anyways

Simplehorror:
ok will do tmrw have migraine

Simplehorror:
ok not interested in this anymore

PR #16717, opened 2022-11-25
stechkins your .357 (the ammo is cheaper)

Skrem7:

I don't want to come up with a way to make .357 not miserable considering printing out 10mm boxes at the lathe is super easy to top off your magazines and printing off new .357 rounds to fill your speed loaders is much more miserable so I guess I would just rather not have the .357 exist as another TC dumb noob trap thingy because the mini ebow still exists mind you and the hardlight bow exists more now too

So all .357 speed loaders cost 1 TC now, and you can buy normal and Ironfeather in pairs

To offset this cost change some of the absurd ammos have been made less absurd while some of the less absurd ammos have been made more absurd. Namely,

- Ironfeather does 8 damage per pellet now, for a total max damage of 48 instead of 51
- Metalshock has its tesla value up to 20000 from 17500 (which is about 29 to 33 burn on first target my math might be wrong)
- Heartpiercer has 5 less AP so it takes 5 bullets to crit bulletproof armor now, still 3 for standard armor
- Wallstake does 9 more damage because wallstun doesn't exist anymore so it's pretty mid and meme-tier

.38 modified also should have less spread and fire rate because the significant fire delay kind of made .357 loading into your revolver even deader than it was before. .38 should still outclass .357 in raw DPS (not even considering stamina damage) so knock yourself off with the powergaming bullets I guess

# Wiki Documentation

TC costs to be updated on Syndicate Items, also the standard and Ironfeather to be replaced with da box
Damage values to be adjusted on Guide to Combat

# Changelog

:cl:  
tweak: Makes .357 Ironfeather not able to two shot at point blank
tweak: .357 Metalshock tesla arc does even more damage now (about 33 burn on first target)
tweak: .357 Heartpiercer has just slightly less AP to the point where it performs the same against bulletproof armor as 10mm AP HOWEVER it still 3-shots normal armor
tweak: .357 Wallstake now does 36 damage instead of 25 since wallstun no longer exists so it's more of a meme ammo now
tweak: You now purchase .357 and .357 Ironfeather speed loaders in pairs for 1 TC
tweak: All other single .357 speed loaders now cost 1 TC
tweak: .38 modified no longer has absurd spread and as slow of a fire rate; its reduced fire rate should still do less DPS than .38, even regarding lethal damage
/:cl:

### DIFF
--- a/code/game/objects/items/storage/uplink_kits.dm
+++ b/code/game/objects/items/storage/uplink_kits.dm
@@ -56,15 +56,15 @@
 			new /obj/item/flashlight/emp(src) //2 TC
 			new /obj/item/jammer(src) //5 TC
 
-		if("guns") //Total cost of 31 TC
+		if("guns") //Total cost of 29 TC
 			new /obj/item/gun/ballistic/revolver(src) //6 TC
 			new /obj/item/gun/ballistic/revolver(src) //6 TC
 			new /obj/item/gun/ballistic/automatic/pistol(src) //6 TC
 			new /obj/item/gun/ballistic/automatic/pistol(src) //6 TC
-			new /obj/item/ammo_box/a357(src) //1 TC
-			new /obj/item/ammo_box/a357(src) //1 TC
-			new /obj/item/ammo_box/a357(src) //1 TC
-			new /obj/item/ammo_box/a357(src) //1 TC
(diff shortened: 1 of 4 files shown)


TheRyeGuyWhoWillNowDie:
reskin

JamieD1 (approved):

PR #16722, opened 2022-11-25
makes .38 a fully lethal round as it should be

Skrem7:

woo boy this is a bad idea

Removes stamina damage from the .38 and instead makes it a 25 damage bullet with -40 armour piercing to represent its ineffectiveness against armour.

Why is this good? It reduces its overall effective damage while also making it weaker against people with armor so it's less good against prepared antags so detective isn't quite as much of a valid tool and it's considered to be some self defense weapon more strictly instead of a better disabler

The special rounds, as a result, have had their damage upped to 20, but they still retain the negative AP

# Wiki Documentation

Description will no doubt have to be changed on the Detective page and the Guide to Combat which I CAN DO

# Changelog

:cl:  
tweak: .38 now does 25 damage with -40 AP, it has lost its stamina damage
tweak: .38 experimental rounds raised to 20 damage as a result, though they retain the negative AP
experimental: Police brutality.
/:cl:

### DIFF
--- a/code/modules/projectiles/boxes_magazines/ammo_boxes.dm
+++ b/code/modules/projectiles/boxes_magazines/ammo_boxes.dm
@@ -53,21 +53,21 @@
 /obj/item/ammo_box/c38/hotshot
 	name = "speed loader (.38 Hot Shot)"
 	desc = "A six-shot speed loader designed for .38 revolvers. \
-			These rounds trade exhaustive properties for an incendiary payload which sets targets ablaze."
+			These rounds trade some damage for an incendiary payload which sets targets ablaze."
 	icon_state = "38hot"
 	ammo_type = /obj/item/ammo_casing/c38/hotshot
 
 /obj/item/ammo_box/c38/iceblox
 	name = "speed loader (.38 Iceblox)"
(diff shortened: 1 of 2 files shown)


ToasterBiome:
now im just gonna crit people instead, nice

Skrem7:
> now im just gonna crit people instead, nice

get some hardlight bullets if you wanna do stamina damage

ynot01:
> now im just gonna crit people instead, nice

If you're pulling that trigger, you shoot to kill. Not some petty arrest. I like this PR

RG4ORDR:
> > now im just gonna crit people instead, nice
> 
> If you're pulling that trigger, you shoot to kill. Not some petty arrest. I like this PR

Just like a real cop.
I approve.

Sniblet:
Now they can't argue when I ban them for using the gun

Rdrazga:
After years we have come back to exactly where we started.

MajManatee:
If the NT-USP gets merged, maybe the det can get a hardlight speedloader...

JamieD1 (approved):

PR #16738, opened 2022-11-26
Re-enable infiltrators

Simplehorror:

This gamemode is well-balanced, why not include it? Why do we just have code rotting around in this repo that goes unused as even admins don't play this gamemode
# Spriting

# Wiki Documentation

# Changelog

:cl:  
tweak: Re-enable infiltrators
/:cl:


Sniblet:
It was cool but also monumentally greenshifty

Simplehorror:
so are traitors and blood brothers if they don't get hijack

Simplehorror:
can give more evil objectives

Firewolf34:
Okay okay hear me out. Make it a random event

If the issue is that it's too stealthy, then just make it something happening in the background during dynamic. If you need to scale the amount of infiltrators down in the team, so be it. It definitely seems like it's too weak of a threat to be the entire focus of the round, at least.

Infiltrators needs to pull ghosts out from the pool in lategame, for a random operation the same way the ninja / revenant / lone op / etc. do. Unless that's what this change enables?

Simplehorror:
that might happen once i implement my Dungeon Master system

neeshacark:
i would not have an issue with reenabling infiltrators if we can fix it to where they have a way to get more objectives so the round isnt just a greenshift once they do their things.

VaelophisNyx:
I leave and everyone loses their damned minds it seems
they were disabled for being colossally shit

improve that shit before you bother re-enabling them

Simplehorror:
can't fix them until i see some gameplay

Bibby0110:
watch replays

> can't fix them until i see some gameplay

PR #16739, opened 2022-11-26
[AsteroidStation] Adds an Autolathe to Service Hall

Aquizit:

Adds an Autolathe to the Service Hall in Asteroid Station. I did have to remove a table to get it to fit, and I moved the rolling pin to the other cooking items table, and the bucket to the table with the key on it. This makes it equal to box as it has both types of lathes.

:cl:  
mapping: Adjusts the Service Hall on Asteroid Station to include an autolathe, removing a table and moving some items to make it fit.
/:cl:

### DIFF
--- a/_maps/map_files/AsteroidStation/AsteroidStation.dmm
+++ b/_maps/map_files/AsteroidStation/AsteroidStation.dmm
@@ -1625,11 +1625,6 @@
 "anj" = (
 /turf/closed/wall,
 /area/maintenance/starboard)
-"ank" = (
-/obj/structure/table,
-/obj/item/key/janitor,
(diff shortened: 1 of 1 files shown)


ToasterBiome (changes requested):
use the mapping tag in the changelog, all of this can be 1 line with the mapping tag

also i approve this change

Aquizit:
Like so?

ToasterBiome:
Yes but now map conflicts

Aquizit:
Ok I think I got it. Maybe not in the neatest way since I closed it but.

ToasterBiome (approved):

PR #16745, opened 2022-11-26
Ethereals are no longer dead weight darkspawns

Moltijoe:
"You are only able to divulge in darkness!"
fuck you! i am literally anti-darkness

:cl:  
bugfix: Ethereal darkspawn can now divulge
/:cl:


Burger1998:
can we not just blacklist ethereals from dark antags like bloodless races are blocked from being vamp and bloodsucker? seems kinda silly that a demon of darkness is literally the thing they hate

Moltijoe:
> can we not just blacklist ethereals from dark antags like bloodless races are blocked from being vamp and bloodsucker? seems kinda silly that a demon of darkness is literally the thing they hate

The difference between these antags is that darkspawn transforms when they "become antag"
vamp and bloodsucker don't change species so it causes fucky things

JamieD1 (approved):

PR #15408, opened 2022-08-21
PK borg's emagged shock hands are no longer an unblockable melee stun

TheRyeGuyWhoWillNowDie:
emagged PK borgs are really, really strong. INSANELY strong. their full equipment currently measures out to:
mostly unblockable instant melee stun
Flashbang
portable doorshock except it also ignores insulated gloves
a ton of poisons

Which makes them, ironically, the most combat-oriented of all the emagged cyborgs. This PR just changes the shock hands to respect shockproofing like insulated gloves, because unlike engineer cyborg's stun arm, you can't avoid this. You WILL get melee stunned.
 
Before you call this an Ided, I have never died to this, but I have gotten a ton of cheap kills, and it feels really dirty.

# Changelog

:cl:  
tweak: Emagged peacekeeper cyborg's shock-hugs now respect insulation.
/:cl:

PR #16781, opened 2022-11-27
Emagged peaceborg hug follows insulation

Moltijoe:
one click insta-stun with no counterplay is no fun
now it checks for insulation

:cl:  
tweak: Emagged peaceborg hug follows insulation
/:cl:

### DIFF
--- a/code/game/objects/items/robot/robot_items.dm
+++ b/code/game/objects/items/robot/robot_items.dm
@@ -133,7 +133,7 @@
 			if(scooldown < world.time)
 				if(M.health >= 0)
 					if(ishuman(M)||ismonkey(M))
-						M.electrocute_act(5, "[user]", safety = 1)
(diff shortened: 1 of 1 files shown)


Simplehorror:
counterplay is a flash in this case

Moltijoe:
> counterplay is a flash in this case

flashes aren't fun either

TheRyeGuyWhoWillNowDie:
#15408

Moltijoe:
> #15408

not the same, that PR lets it be negated simply by using insulated gloves
this one requires using a hardsuit or other suit slot item with a siemens coefficient of 0 or less, similar to a tesla zap
also this doesn't make the shock hug consume power

currently the only things that will prevent it are literally just hardsuits and reactive armour

insulated gloves will reduce the damage by 50%, but it's not enough to negate the shock

JamieD1 (approved):

PR #16790, opened 2022-11-27
Make mulligan more interesting

Simplehorror:

imagine the wizard dying 10 minutes in so you have to sit through a fucking 1h45m bloodsucker round. 
this should make mulligan more interesting
# Spriting

# Wiki Documentation

# Changelog

:cl:  
rscadd: Added cult, nuclear, revolution, wizard, zombie to the mulligan list
rscdel: Bloodsuckers, traitorsuckers, heretics, traitorvamp removed from mulligan list.
/:cl:

### DIFF
--- a/code/game/gamemodes/zombie/zombie.dm
+++ b/code/game/gamemodes/zombie/zombie.dm
@@ -18,6 +18,7 @@ GLOBAL_LIST_EMPTY(zombies)
 	required_enemies = 3
 	recommended_enemies = 3
 	enemy_minimum_age = 14
+	reroll_friendly = 1
 
 	announce_span = "zombie"
 	announce_text = "Some crew members have been infected with a zombie virus!\n\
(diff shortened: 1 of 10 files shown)


code/game/gamemodes/zombie/zombie.dm
```diff
@@ -18,6 +18,7 @@ GLOBAL_LIST_EMPTY(zombies)
 	required_enemies = 3
 	recommended_enemies = 3
 	enemy_minimum_age = 14
```
Bibby0110: no

Bibby0110 (changes requested):

code/game/gamemodes/zombie/zombie.dm
Simplehorror: ok

neeshacark:
i love this, except nukie which i dont think would work very well but yeah mulligans generally suck

Bibby0110 (changes requested):
these should be TRUE/FALSE

Simplehorror:
> these should be TRUE/FALSE

this is how the rest of the code does it

JamieD1:
Unfortunately those gamemodes are not always valid when wizard can spawn.

PR #16794, opened 2022-11-27
Removes the head blacklist from vampire revive since it breaks reviving, and moves the revive to below the limb reattachment

SomeguyManperson:

The vampire revive fails if the user has no head. This is probably unintended, as vampires can survive without a head otherwise, and the code is similar to changeling revives (which do the same thing)

# Changelog

:cl:  
tweak: vampire revive now replaces the head. This is important, apparently, for being revived. I assume it was like this because it was copied from changeling revive, but changelings can put their heads back on and vampires can't.
bugfix: vampire revive no longer fails if the head is not present
/:cl:

### DIFF
--- a/yogstation/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/yogstation/code/game/gamemodes/vampire/vampire_powers.dm
@@ -323,14 +323,14 @@
 	addtimer(CALLBACK(src, /obj/effect/proc_holder/spell/self/revive.proc/revive, L), 600)
 
 /obj/effect/proc_holder/spell/self/revive/proc/revive(mob/living/user)
-	user.revive(full_heal = TRUE)
-	user.visible_message(span_warning("[user] reanimates from death!"), span_notice("We get back up."))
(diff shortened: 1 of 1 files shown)


Cartlord:
I figured this was intentional, since - in at least some mythology - decapitating a vampire is supposed to be one of the things that keeps it dead.

SomeguyManperson:
it's effectively copied straight from changeling's revive (including the whole brain-not-necessary thing) except it doesn't work ??

Moltijoe:
> I figured this was intentional, since - in at least some mythology - decapitating a vampire is supposed to be one of the things that keeps it dead.

whether or not decapitation seems to kill the vampire in stories is somewhat inconsistent
sometimes it fully kills them 
sometimes it just "kills" them until the head is reattached
sometimes they can just form a new head from blood
sometimes the body can just pick the head up and slap it back on
would be interesting if different levels of vampires had different levels of this

JamieD1 (approved):

PR #16802, opened 2022-11-27
Nerfs industrial gray extracts

Moltijoe:
from the mewsy hitlist

-=Industrial Grey=-
makes 5 monkey cubes for 2 plasma, can give you monkeys easily and super fast- while this doesnt really make you "unkillable" or a combat advantage, it speeds up xeno extremely easily and early and is very strong in it's own way
-=Possible Solutions to it=-
*reduce the amount of monkey cubes it gets to 3

:cl:  
tweak: industrial gray extracts now give less monkey cubes
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/industrial.dm
+++ b/code/modules/research/xenobiology/crossbreeding/industrial.dm
@@ -55,7 +55,7 @@ Industrial extracts:
 	colour = "grey"
 	effect_desc = "Produces monkey cubes."
 	itempath = /obj/item/reagent_containers/food/snacks/monkeycube
-	itemamount = 5
+	itemamount = 3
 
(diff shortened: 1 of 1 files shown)


Simplehorror:
i don't think this will work as you want it to

Moltijoe:
> i don't think this will work as you want it to

this is exactly as mewsy stated it

Simplehorror:
right, but it won't really change anything, nor do I think this should be nerfed

Moltijoe:
> right, but it won't really change anything, nor do I think this should be nerfed

i sorta was thinking this too, i have no problem with the speed of xenobio, they already take forever to do stuff, i'd rather not make it tedious

isotope436:
this one shouldnt be nerfed

PR #16806, opened 2022-11-27
Adamantine shield now actually slows you

Moltijoe:
Mewsy hitlist

-=Burning Adamantine=-
a super good shield that is usually paired with adamantine armor, has a high high high block chance, while it's two handed, it's still really strong because you can do lots of damage with it
-=Possible Solutions to it=-
*reduce the chance it will block
*make it heavier

It already had the tag, it just didn't have a slowdown number so it's also sort of a bugfix

:cl:  
tweak: Adamantine shield now actually slows you
/:cl:


Simplehorror:
make it do 0 force, adamantine is all about damage reduction

ReddicusDragon (approved):

JamieD1 (approved):

PR #16808, opened 2022-11-27
nerfs admantine armour

Moltijoe:
Mewsy hitlist

-=Adamantine Armor=-
very very high armor, with a huge slowdown that can be negated by a stable red, basically free elite operative armor, it can be stacked with chemicals, nanites, stable purple, and being antags like bloodsuckers with their health gain, and various other things.
-=Possible Solutions to it=-
*wearing the armor now reduces healing recieved by 50-75% due to the high armor while worn, this way it's a direct nerf to the item rather than needing to affect other things that only really make this broken- most of the time atleast.
*it now still gives the slow through stable red, magically
*reduce the armor values of melee and bullet and things

went with the lowest effort solution, about 10 less on every defense except for laser which was removed because it already has a 40% chance to reflect lasers

:cl:  
tweak: nerfs admantine armour
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/_clothing.dm
+++ b/code/modules/research/xenobiology/crossbreeding/_clothing.dm
@@ -133,7 +133,7 @@ Slimecrossing Armor
 	item_state = "adamsuit"
 	flags_inv = NONE
 	obj_flags = IMMUTABLE_SLOW
-	armor = list(MELEE = 60, BULLET = 50, LASER = 30, ENERGY = 50, BOMB = 80, BIO = 100, RAD = 0, FIRE = 90, ACID = 90)
+	armor = list(MELEE = 50, BULLET = 40, LASER = 0, ENERGY = 30, BOMB = 70, BIO = 100, RAD = 0, FIRE = 70, ACID = 70)
 	slowdown = 4
 	var/hit_reflect_chance = 40
 


Simplehorror:
i think it should keep the armor value but the slowdown should be unable to be removed, so you're basically a juggernaut

Simplehorror:
noticing an issue with xenobio. all the items seem balanced in a vacuum and then you start stacking them and then it's just chaos

Moltijoe:
> noticing an issue with xenobio. all the items seem balanced in a vacuum and then you start stacking them and then it's just chaos

very

Moltijoe:
> i think it should keep the armor value but the slowdown should be unable to be removed, so you're basically a juggernaut

best i could do is add a special check to stabilized red that makes it keep the adamantine armour slowdown

Simplehorror:
> > i think it should keep the armor value but the slowdown should be unable to be removed, so you're basically a juggernaut
> 
> best i could do is add a special check to stabilized red that makes it keep the adamantine armour slowdown

didn't stabilized red just get shafted?

Moltijoe:
> didn't stabilized red just get shafted?

iunno, maybe? i don't think that PR is merged yet if it was

bruhlookatthisdood:
remove xenobio, add second bar

Moltijoe:
mqiib's solution is more elegant, not to mention stabilized red WAS nerfed

PR #16813, opened 2022-11-27
adds a delay to stabilized cerulean

Moltijoe:
Mewsy hitlist

-=Stablized Cerulean=-
if you get killed with it in your backpack, you get teleported back to your body, so basically you're unkillable permanently, and can stack stables into their inventory to make you even harder to kill, can have a bunch of them at once for infinite lives
-=Possible Solutions to it=-
*give a message nearby to people that their soul warped to a new body or something, maybe add a pointer to which way the body is
*make it unstackable somehow, make it kill you if used more than once or do damage or something
*make people have a certain thing to their person when holding the extract so people know it exists somewhere, allowing them to hunt it
*make it take like a minute or 2 before it transfers the body, allowing time for people to react rather than instantly

:cl:  
tweak: stabilized cerulean has a 30 second delay
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
+++ b/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
@@ -718,6 +718,8 @@ datum/status_effect/stabilized/blue/on_remove()
 /datum/status_effect/stabilized/cerulean/tick()
 	if(owner.stat == DEAD)
 		if(clone && clone.stat != DEAD)
+			owner.visible_message(span_warning("[owner] begins to glow with brilliant light."))
+			sleep(30 SECONDS)
 			owner.visible_message(span_warning("[owner] blazes with brilliant light, [linked_extract] whisking [owner.p_their()] soul away."),
 				span_notice("You feel a warm glow from [linked_extract], and you open your eyes... elsewhere."))
 			if(owner.mind)


code/modules/research/xenobiology/crossbreeding/_status_effects.dm
```diff
@@ -718,6 +718,8 @@ datum/status_effect/stabilized/blue/on_remove()
 /datum/status_effect/stabilized/cerulean/tick()
 	if(owner.stat == DEAD)
 		if(clone && clone.stat != DEAD)
+			owner.visible_message(span_warning("[owner] begins to glow with brilliant light."))
+			sleep(30 SECONDS)
```
SomeguyManperson, line 722: what the fuck is htis

Simplehorror:
honestly this one needs to be gutted, it's a better version of lich

code/modules/research/xenobiology/crossbreeding/_status_effects.dm
Moltijoe: the worst made PR of my rampage, by far
all it does is add a 30 second delay before it continues

Simplehorror: you need to use addtimer not sleep

Moltijoe: > you need to use addtimer not sleep

sleep is a global proc that adds delay, addtimer could also work, but then i'd need to make another proc for the timer to callback

Moltijoe:
this honestly is such a hack of a PR
this needs more work than what i'm willing to put the effort into

PR #16815, opened 2022-11-27
Removes Premalignant tumors

CMOisLing:
Malignant tumors are really the only useful ones, the others just took up space. Kept Benign and buffed it to premalignant levels so it can find its use in a multi factor healing virus.

Wiki Update
In infections Premalignant turmors should be removed and the value of Benign tumors should be updated to have a stage speed and transmission of -2

:cl:  
rscdel: Removed a tumor
/:cl:


ToasterBiome:
Noo my tumors

Firewolf34:
This requires a wiki page update

~~Imo, seems like 4 resistance boost for benign tumors, with (now) zero downsides is a bit stronk for a trait that has other bonuses of regrowing limbs and such. This puts it on par if not better than pure booster symptoms like evo-accel and etc that have similar strength yet tons of negatives associated. Maybe take 1 off this and move it to another category? Just a suggestion.~~

~~Because as you have it now, if you sort the Wiki page by resistance, Benign Tumors is near the top, and with this change will have no negative trait downsides. I don't see why you wouldn't use it in any Resistance-based virus (optimally)~~

CMOisLing:
> This requires a wiki page update
> 
> Imo, seems like 4 resistance boost for benign tumors, with (now) zero downsides is a bit stronk for a trait that has other bonuses of regrowing limbs and such. This puts it on par if not better than pure booster symptoms like evo-accel and etc that have similar strength yet tons of negatives associated. Maybe take 1 off this and move it to another category? Just a suggestion.
> 
> Because as you have it now, if you sort the Wiki page by resistance, Benign Tumors is near the top, and with this change will have no negative trait downsides. I don't see why you wouldn't use it in any Resistance-based virus (optimally)

Your right I thought I changed this like 5 hours ago but I guess I forgot to push it. Ill fix it when I get home.

Firewolf34:
Sorry @CMOisLing , my bad. I looked again, it appears to be -2 for the other traits. I guess I was looking at the wrong commit. Looks good to me!

CMOisLing:
> Sorry @CMOisLing , my bad. I looked again, it appears to be -2 for the other traits. I guess I was looking at the wrong commit. Looks good to me!

Naw its not your fault I forgot to push the commit. I guess they are timestamped so thats why it showed up before your comment even though the values were edited after you posted the comment.

JamieD1 (approved):

PR #16835, opened 2022-11-28
remove steam vent table exploit

Simplehorror:

fixes exploit where you can stack these to make them impassable
# Spriting

# Wiki Documentation

# Changelog

:cl:  
bugfix: remove steam vent table exploit
/:cl:

### DIFF
--- a/code/modules/antagonists/clockcult/clock_structures/traps/steam_vent.dm
+++ b/code/modules/antagonists/clockcult/clock_structures/traps/steam_vent.dm
@@ -13,10 +13,17 @@
 	icon_state = "steam_vent_[opacity]"
 	if(opacity)
 		playsound(src, 'sound/machines/clockcult/steam_whoosh.ogg', 50, TRUE)
-
+		START_PROCESSING(SSfastprocess, src)
 	else
 		playsound(src, 'sound/machines/clockcult/integration_cog_install.ogg', 50, TRUE)
+		STOP_PROCESSING(SSfastprocess, src)
 
+/obj/structure/destructible/clockwork/trap/steam_vent/process()
+	. = ..()
+	for (var/obj/structure/table/T in get_turf(src))
+		visible_message("\the [T] melts underneath the steam!")
+		qdel(T)
+	
 /obj/structure/destructible/clockwork/trap/steam_vent/Crossed(atom/movable/AM)
 	. = ..()
 	if(isliving(AM) && opacity)


ynot01:
break the vent

also this has no reason to be under SSfastprocess

Bibby0110 (changes requested):
use one_per_turf on the recipe instead

Moltijoe:
> break the vent

they're not supposed to be walls, they're supposed to be line of sight blockers

Simplehorror:
> use one_per_turf on the recipe instead

tables are not a recipe

Bibby0110:
> > use one_per_turf on the recipe instead
> 
> tables are not a recipe

what?

Rdrazga:
Tbf it’s only an exploit because people can’t figure out how to climb over tables, also the vent takes like 3 hits to break. (Don’t remove this please I just discovered it and I like how fun it is to do)

Simplehorror:
> > > use one_per_turf on the recipe instead
> > 
> > 
> > tables are not a recipe
> 
> what?

someone build steam vent.
steam vent not dense
someone build table on top of steam vent

Simplehorror:
i should just make it so the table appears on top

Simplehorror:
> Tbf it’s only an exploit because people can’t figure out how to climb over tables, also the vent takes like 3 hits to break. (Don’t remove this please I just discovered it and I like how fun it is to do)

this is an old one i remember doing this llol

Rdrazga:
> i should just make it so the table appears on top

fair trade

Rdrazga:
Now if you want to look at something that does actually need fixing is that the vents even when not on hide pressure plates and skewers

Rdrazga:
The tables are also prob a bit too hard to break which is what makes this good to begin with. Even without the vents the tables are just a great ‘screw you’ to crew trying to get in

Firewolf34:
>someone build steam vent.
>steam vent not dense
>someone build table on top of steam vent

This should be in the PR description, would be more clear

Bibby0110:
> > > > use one_per_turf on the recipe instead
> > > 
> > > 
> > > tables are not a recipe
> > 
> > 
> > what?
> 
> someone build steam vent. steam vent not dense someone build table on top of steam vent

i misread

Bibby0110 (changes requested):
do this without fastprocess

Simplehorror:
cant

PR #16844, opened 2022-11-29
removes the printable traitor items from the medlathe

TheRyeGuyWhoWillNowDie:
what if we stopped powercreeping implants?

these two are literally traitor items except they can't be pulled off of you and the wheelie implant is missing the one thing that makes it situational (the fact that it's VERY visible) and enabling more medliz/roboticist powergaming is not actually a good thing

# Changelog

:cl:  
rscdel: you cant print no slip implants and wheelie implants any more
/:cl:

### DIFF
--- a/code/modules/research/designs/medical_designs.dm
+++ b/code/modules/research/designs/medical_designs.dm
@@ -493,17 +493,6 @@
 	category = list("Implants", "Medical Designs")
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL
 
-/datum/design/noslipwater
-	name = "Slip Resistance Implant"
-	desc = "An implant that uses sensors and motors to detect when you are slipping and attempt to prevent it. It probably won't help if the floor is too slippery."
-	id = "noslipwater"
-	build_type = PROTOLATHE | MECHFAB
-	construction_time = 80
-	materials = list(/datum/material/iron = 3000, /datum/material/glass = 1500, /datum/material/silver = 1000, /datum/material/diamond = 1000, /datum/material/uranium = 400)
-	build_path = /obj/item/organ/cyberimp/leg/galosh
-	category = list("Implants", "Medical Designs")
-	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL
-
 /datum/design/jumpbootsimplant
 	name = "Jumpboots implant"
 	desc = "An implant with a specialized propulsion system for rapid foward movement."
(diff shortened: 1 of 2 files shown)


neeshacark:
lock them under illegal tech instead

Moltijoe:
wheelies are traitor? they come from medbot amputation arcade machine

airshoes are the traitor item, the implant for that is in the uplink

i could see wheelies being locked behind either combat or illegal if it's enough of a problem

> lock them under illegal tech instead

the noslip water one IS under illegal tech
it's literally just galosh implants

the full noslip implant is in the traitor uplink

TheRyeGuyWhoWillNowDie:
printing zero slowdown galoshes that also prevent ice slipping is not a thing that medbay players need (outside of powergaming)
getting wheelie shoes requires getting lucky with an arcade machine (and being good at arcade games) while pressing a button to get them for free does not

Moltijoe:
> printing zero slowdown galoshes that also prevent ice slipping is not a thing that medbay players need (outside of powergaming) getting wheelie shoes requires getting lucky with an arcade machine (and being good at arcade games) while pressing a button to get them for free does not

illegal implants is already the powergame implants slot anyways
the noslipwater implant being behind illegal is already enough of a restriction if you ask me. removing ice noslip seems fine i guess, but it's not like ice is encountered all that much
wheelie shoes being more restricted than where they are seems fair, but removing them entirely is overkill

neeshacark:
> printing zero slowdown galoshes that also prevent ice slipping is not a thing that medbay players need (outside of powergaming) getting wheelie shoes requires getting lucky with an arcade machine (and being good at arcade games) while pressing a button to get them for free does not

nor are most other implants, wtf is a md going to do with a cns rebooter or welding shield eyes? they get them because other jobs can get use of them. also n3d made ice not slip you when you wear mining boots/winter boots so I hardly think that's an issue. that being said apparently they ARE locked behind illegal tech anyway so i fail to see how this is at all a problem

UltraBigRat:
OH NO WHATEVER WILL WE DO, THE POWERGAMER PARAMEDIC IS IMMUNE TO THE JANITORS MOPPED FLOOR DEATH TRAP! AND THEY CAN UNCONTROLLABLY RUN INTO WALLS AT HIGH SPEEDS AND BREAK THEIR LEGS!

Simplehorror:
this is fun removal

Moja10:
When I read the title, I was expecting something like "oh no did the dentcoders add an arm implant with a 40 damage scalpel or something" but instead its slippy willies and go fast. This is a nonissue, chemists can make all their speed chems and no one cares about slipping, slipping protection really is a relic from when slips hard stunned. I say these implants are fine

ReddicusDragon (commented):
non-issue

Simplehorror:
we should probably remove soap from station seeing as it is a traitor item

JamieD1:
> lock them under illegal tech instead

Do this

MajManatee (changes requested):
see above

PR #16848, opened 2022-11-29
Change to how adjust_fire_stacks works

Moltijoe:
Apparently people can make themselves wet by doing stop drop and roll tactics, that doesn't seem right
rather than changing every thing individually, just add a threshold where it won't instantly make someone wet if they're on fire

:cl:  
tweak: tweaked adjust_fire_stacks
/:cl:


ReddicusDragon (approved):
code looks good

JamieD1 (approved):

PR #16854, opened 2022-11-29
buff the new stabilized red ever so slightly

Moltijoe:
it's currently weak enough that you can't really feel it to the point that people thought it didn't work
this is probably as much as it can be buffed as a zero downside speed boost

:cl:  
tweak: new stabilized red is ever so slightly faster
/:cl:


ReddicusDragon (approved):
goodchange

JamieD1 (approved):

PR #16875, opened 2022-11-30
black powder explosions are instant, securing a chemical grenade takes 3 seconds.

TheRyeGuyWhoWillNowDie:
makes it marginally harder to ignore grenades blowing up through emps, and I should have considered this earlier honestly also makes it so you can't ignore large grenades existing with blackpowder explosions from a bluespace beaker, which were previously delayed... like a grenade.

# Wiki Documentation
black powder explosions are now instant

# Changelog

:cl:  
tweak: it takes three seconds to secure a chemical grenade
tweak: black powder explosions are now instant
/:cl:

### DIFF
--- a/code/game/objects/items/grenades/chem_grenade.dm
+++ b/code/game/objects/items/grenades/chem_grenade.dm
@@ -58,9 +58,11 @@
 	if(I.tool_behaviour == TOOL_SCREWDRIVER)
 		if(stage == GRENADE_WIRED)
 			if(beakers.len)
-				stage_change(GRENADE_READY)
-				to_chat(user, span_notice("You lock the [initial(name)] assembly."))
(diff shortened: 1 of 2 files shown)


Moltijoe:
should probably be two different PRs

Moltijoe:
So this would mean blackpowder could only be used either to suicide or in a grenade combined with a very hot reagent?

TheRyeGuyWhoWillNowDie:
no its literally just a normal explosive chemical now instead of the built in timer bullshit, grenades heat their reagents by a bit when they go off, you just build a grenade with it like every other explosive

Moltijoe:
fair enough, don't know enough about chem to say whether or not this is balanced

JamieD1 (approved):

PR #16878, opened 2022-11-30
Makes the IPC martial art more accessible

Moltijoe:
Since it got added over a month ago, i don't think there's been a single person to use it
There's no info if it's even well balanced because no one uses it
I probably went overboard on restrictions

Couple tweaks to it in the uplink
Costs 16 TC instead of 20 TC
Can receive discounts now
No longer limited to martyr and hijack

:cl:    
tweak: IPC martial art now costs 16TC, CAN get discounted, and has no objective restrictions
/:cl:

### DIFF
--- a/yogstation/code/modules/uplink/uplink_item.dm
+++ b/yogstation/code/modules/uplink/uplink_item.dm
@@ -169,11 +169,9 @@
 /datum/uplink_item/race_restricted/ultra_violence
 	name = "Version one upgrade module"
 	desc = "A module full of forbidden techniques that will make you capable of ultimate bloodshed."
-	cost = 20
+	cost = 16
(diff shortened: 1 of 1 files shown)


ReddicusDragon (commented):
requires balance review

Moltijoe:
> requires balance review

Thats sort of the problem, there's no data on the martial art yet, even with this its still more expensive than every other martial art

LazennG:
> requires balance review

not again it doesnt

ReddicusDragon:
> > requires balance review
> 
> 
> 
> not again it doesnt

Any tc change should be reviewed for balance. Upon re-evaluating this PR I'd say it's good to go

JamieD1 (approved):

PR #16881, opened 2022-12-01
Prevent Space Heaters and Electrolyzers from operating indefinitely

Firewolf34:
image image

I noticed that the Space Heater sprite seemed to change even when the cell was depleted. I was even more surprised to find it continued to actually heat and cool without any charge. With some research I found that this is because it will operate as long as the charge is > 0. It never reaches 0 in practical use however because `cell.use()` simply returns false if you try to pull more joules than it has stored, instead of dropping to 0. This allows the machine to continue to operate indefinitely at the bottom of it's battery charge. Since in my research the Electrolyzer appears to be based off of the Space Heater code, I fixed that one as well. 

There were some minor improvements to the code - these devices automatically shut off when they run out of charge instead of having the appearance of running and yet not doing anything. And finally, JavaScript UI has been adjusted to not allow you to turn on these machines if their charge is < 0.5 (an artifact of rounding arithmetic).

Recommend checking diffs with whitespace ignored, as I changed indentation of a block.

### Testing
Tested against 2f1642e8df2035d0f2c7cc47a2f62e850177a288
Electrolyzer and Space Heater both tested at extreme ranges of cell charge level, with 3 different cell types (including yellow slime core) and also UI tested to work when depleted and when at full charge. Tested at Tier 4 and Tier 1 upgrade levels.

# Changelog

:cl:  
bugfix: Space Heaters and Electrolyzers now obey the laws of thermodynamics.
/:cl:

### DIFF
--- a/code/game/machinery/spaceheater.dm
+++ b/code/game/machinery/spaceheater.dm
@@ -58,7 +58,7 @@
 
 /obj/machinery/space_heater/update_icon()
 	if(on)
-		icon_state = "sheater-[mode]"
+		icon_state = "sheater-[on ? "[mode]" : "off"]"
 	else
 		icon_state = "sheater-off"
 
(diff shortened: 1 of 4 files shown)


code/game/machinery/spaceheater.dm
```diff
@@ -58,7 +58,7 @@
 
 /obj/machinery/space_heater/update_icon()
 	if(on)
-		icon_state = "sheater-[mode]"
+		icon_state = "sheater-[on ? "[mode]" : "off"]"
```
Moltijoe, line 61: i appreciate this change, but why still have the if/else then?

Simplehorror:
nice, will probably try to buff space heaters since they're useless

ShoziOffical:
I honestly hate electrolyzer and space heater using cell, make it uses apc power when it's wrenched down

ShoziOffical:
> nice, will probably try to buff space heaters since they're useless

Not really tho, but yeah sure

ShoziOffical:
Run .\tgui\bin\tgui --fix
then push again

Rdrazga:
> I honestly hate electrolyzer and space heater using cell, make it uses apc power when it's wrenched down

Or at the very least it would be nice even if they didn't draw from APC power they drew from cables they are wrenched on. Either would be good.

ShoziOffical:
> > I honestly hate electrolyzer and space heater using cell, make it uses apc power when it's wrenched down
> 
> Or at the very least it would be nice even if they didn't draw from APC power they drew from cables they are wrenched on. Either would be good.

Well using apc power will require less coding and more convenient

Firewolf34:
>buff space heater

So I actually thought about doing that with this PR, as they do kind of suck right now. 

 The way it works is, the Space Heater is limited by its initial `heatingPower`. It is allowed to use that many joules of energy to heat/cool per process call. This limits it to +/- 20°K delta max per process call, and 2 cell units of power consumption at stock efficiency assuming 2000J heat capacity available on tile. 

 Note the heat capacity rating, the space heater is reliant on the air within its tile to have a significant amount of heat capacity to do work. The amount of work that it does is directly translated into electrical power draw. Assuming that the amount of heat capacity for a cubic meter of air in SS13 at standard temperature and pressure is 2000J then it can max move 100K of heat per process tick, if it's not otherwise limited by it's `heatingPower` (best possible scenario).

 So really, it uses only a little energy... when you consider that the standard battery has 2,500 units, it's really not an unreasonable amount of electrical draw. The real issue seems that it can't usually acquire enough molar volume of gas to be able to have a high heat capacity and do work. 

 The issue is not that it runs out of energy too fast, the issue is that it doesn't operate on a large enough area. Working on a single tile would be fine, if it wasn't for the fact that the atmos system updates very slowly when it comes to heat propagation. Heat seems to only spread out like once per second or so. When it does cool atmo down, it only cools the tiles directly surrounding it. it takes a long time for a whole room to be cleared this way. 

 The easiest way you can buff it is to increase its heat capacity ranks for all upgrade tiers. The starting one is 40,000. You could maybe upgrade it to 75,000. But at tier 4 it's already pretty much maxing out the heat capacity of standard air. You could also improve its `efficiency` tiers, but I don't think power is the bottleneck. 

 If power was an issue, you wouldn't have even been able to notice it before. They operated with infinite power for the last half decade or so. That's what this PR sets out to fix after all. 

 And hey, it gives power cells a use. 

 **In my opinion we need to have a larger "Industrial Space Heater"**, kind of like the way we have a full tile Scrubber. This large fulltile Space Heater should be wrenchable on cable. And it should operate over a 3X3 area, giving it *9x heat capacity by default,* even without buffs!

 Then we place one of those suckers in the toxins gas storage area and in atmosia... And people that use those everyday can use the one that's wrenchable on cable tile and not have to screw with batteries.

Firewolf34:
As for the electrolyzer, *it should be modified to start with a power cell*

 Putting in the power cell into the electrolyzer is one of the un-fun rituals that a toxins scientist has to go through every shift.

ShoziOffical:
@Firewolf34 
[SpaceHeater.zip](https://github.com/yogstation13/Yogstation/files/10129414/SpaceHeater.zip)

this should fix ur lints

Firewolf34:
Thank you @warface1234455 I will take a look tomorrow at fixing the linter :) 
(on mobile rn)

ShoziOffical:
> Thank you @warface1234455 I will take a look tomorrow at fixing the linter :) (on mobile rn)

bless

ReddicusDragon (changes requested):
Code looks good. Fix lints then it'll be an approve from me.

Firewolf34:
Lint removed
image

Moltijoe (approved):

code/game/machinery/spaceheater.dm
ShoziOffical: bleh

Firewolf34: 2 reasons: Because the heater would look like it's working when it was off for at least one cycle after its battery depletes. Because its what the electrolyzer does already, so I followed the lead of that JS file

PR #16882, opened 2022-12-01
legcuffs can now be cut like handcuffs

TheRyeGuyWhoWillNowDie:
pretty strange how you can cut handcuffs with a jaws of life/wirecutters, but you can't do the same with legcuffs

Wirecutters can cut energy bolas (hard light is fragile) and regular bolas (they're literally made out of cable coil)

Jaws of Life can cut the same and every thing else, most notably including beartraps which took 30 fucking seconds to remove on top of causing guaranteed wounding if you got hit by one

also im sick of the bola meta because 1 click ranged wins are not funny

# Changelog

:cl:  
tweak: legcuffs can now be cut with wirecutters or jaws of life
/:cl:

### DIFF
--- a/code/game/objects/items/tools/wirecutters.dm
+++ b/code/game/objects/items/tools/wirecutters.dm
@@ -60,6 +60,11 @@
 		if(do_after(user, 1.5 SECONDS, C))
 			to_chat(C, span_notice("You succesfuly remove the durathread strand."))
 			C.remove_status_effect(STATUS_EFFECT_CHOKINGSTRAND)
+	else if(istype(C) && C.legcuffed && (C.legcuffed.type == /obj/item/restraints/legcuffs/bola || istype(C.legcuffed, /obj/item/restraints/legcuffs/beartrap/energy)))
+		user.visible_message(span_notice("[user] cuts [C]'s restraints with [src]!"))
+		qdel(C.legcuffed)
+		C.legcuffed = null
+		return	
 	else if(!(user.a_intent == INTENT_HARM) && attempt_initiate_surgery(src, C, user))
 		return
 	else
--- a/yogstation/code/game/objects/items/tools.dm
+++ b/yogstation/code/game/objects/items/tools.dm
@@ -46,6 +46,11 @@
 			user.visible_message(span_notice("[user] cuts [C]'s restraints with [src]!"))
 			qdel(C.handcuffed)
 			return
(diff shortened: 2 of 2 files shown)


Moltijoe:
will this let people cut their own legcuffs instantly?
it doesn't matter for handcuffs because they can't actually hold wirecutters or jaws of life while handcuffed

TheRyeGuyWhoWillNowDie:
> will this let people cut their own legcuffs instantly?

Yes

Moltijoe:
> Yes

i don't like the implications that people will carry wirecutters around just for when they deal with bolas
at least adding a 1-2 second delay might be proper

TheRyeGuyWhoWillNowDie:
and i dont like the implication that you can instantly slap someone with a 7x slowdown and immobilize them for a few seconds from across the screen, but it's w/e

Also if there was a delay this mechanic would be pointless since bolas are removed in like 3 seconds

RG4ORDR:
Would you consider arm-implants being able to rip apart handcuffs to be a valid?
A bit silly when someone has these blades capable of rending specialized armor being unable to bust hand-cuffs. Course the con is that if you do that then security is just going to lethal you at that point.

Likewise ANY ENERGY BASED MELEE WEAPON - Sword,Axe,Dagger,Katana would they get such benefits as well. Literally lightsabers.

Annoying-plant:
> Would you consider arm-implants being able to rip apart handcuffs to be a valid? A bit silly when someone has these blades capable of rending specialized armor being unable to bust hand-cuffs. Course the con is that if you do that then security is just going to lethal you at that point.
> 
> Likewise ANY ENERGY BASED MELEE WEAPON - Sword,Axe,Dagger,Katana would they get such benefits as well. Literally lightsabers.

there us an implant to break cuffs for syndies already its pretty good and cheap too like 3TC I believe

ReddicusDragon:
bad pr, make it so freedom implant breaks legcuffs too

TheRyeGuyWhoWillNowDie:
They do, but shockingly, there's more than one type of antagonist, and they can't get freedom implants.

UltraBigRat:
should also make cuffshoes escapable, instead of an unremovable bola.

Moltijoe (approved):
code's good, need someone to actually decide on balance

Bibby0110 (approved):

PR #16883, opened 2022-12-01
Gives leg implants an emp effect

Moltijoe:
Somehow i forgot the most important part of implants
getting fucked by emps

when hit by an emp you will take 5 brute and 10 stam in that leg (essentially 10 and 20 since you always have two)
there is a 50% chance for it to disable your leg for 5 seconds
if your leg was disabled, there is a further 20% chance for it to deal 20 brute and break your leg

:cl:  
rscadd: Leg implants now do someone when hit by an EMP
/:cl:

### DIFF
--- a/code/modules/surgery/organs/augment_legs.dm
+++ b/code/modules/surgery/organs/augment_legs.dm
@@ -7,6 +7,7 @@
 
 	//to determine what type of implant for checking if both legs are the same
 	var/implant_type = "leg implant"
+	COOLDOWN_DECLARE(emp_notice)
 
 /obj/item/organ/cyberimp/leg/Initialize()
 	. = ..()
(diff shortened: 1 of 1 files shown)


Sniblet:
> Leg implants now do someone when hit by an EMP

Please do not add ERP

code/modules/surgery/organs/augment_legs.dm
```diff
@@ -13,6 +13,39 @@
...
+	L.set_disabled(TRUE)	//disable the bodypart
+	addtimer(CALLBACK(src, .proc/reenableleg), 5 SECONDS, TIMER_UNIQUE|TIMER_OVERRIDE)
+
+	if(severity & EMP_HEAVY && prob(20))	//10% chance on getting hit by a heavy emp to break your leg (technically more since you have two)
+		to_chat(owner, span_warning("The EMP causes your [src] to thrash your [L] around wildly breaking it!"))	
+		var/datum/wound/blunt/severe/breakdown = new
+		breakdown.apply_wound(L)
+		L.receive_damage(20)
+	else
+		to_chat(owner, span_warning("The EMP causes your [src] to seize up, preventing your [L] from moving!"))
```
ReddicusDragon, line 43: feel like this would be cool to see as a like, 3 minute timer perhaps

Moltijoe: Ah, good call, ion rifles spamming chat doesnt seem ideal

Probably only a 30 second cooldown though

MajManatee (approved):
neat

PR #16874, opened 2022-11-30
Patches hypernob formation exploit by making it so hypernob no longer forms with all ingredients instantly, like other gas formations

Simplehorror:
# Document the changes in your pull request

Pulls it in line with other reactions, this made hypernob unknowingly too easy to make.

Closes #16902

# Spriting

# Wiki Documentation

Atmosians don't document their stuff
# Changelog

:cl:  
tweak: Hypernob no longer forms with all ingredients instantly, like other gas formations
/:cl:

PR #16902, opened 2022-12-02
Hypernoblium formation rework

SapphicOverload:

Hypernoblium is now more endothermic and has a limit to how much can form in a tick which can be increased by adding BZ and more heat.

Highly recommend merging #17092 first.

Closes #16874

# Changelog

:cl:  
tweak: hypernoblium formation is more endothermic and doesn't form all at once
/:cl:

### DIFF
--- a/code/__DEFINES/reactions.dm
+++ b/code/__DEFINES/reactions.dm
@@ -47,7 +47,7 @@
 #define STIMULUM_SECOND_RISE				0.0009
 #define STIMULUM_ABSOLUTE_DROP				0.00000335
 #define REACTION_OPPRESSION_THRESHOLD		5
-#define NOBLIUM_FORMATION_ENERGY			2e7 	//1 Mole of Noblium takes the planck energy to condense.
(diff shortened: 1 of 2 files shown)


ShoziOffical:
Kek

knightshade0:
I still don't think we actually need to change hypernob formation given how it largely exists as a novelty, but if we _do_ insist on changing it this is a solid implementation.

Heat/production curve looks solid. The other PR was bizarre (among other reasons) because physics doesn't really work in terms of arbitrary maximum values- everything works on curves. The math regarding what we'd actually do in game seems workable as well.

It makes sense conceptually and it works for gameplay so I have no opposition to it.

Simplehorror:
hey how about we don't put closes for someone else's PR?

iamthesilly:
My PR makes yours obsolete and will cause conflicts if both get merged

chemistrymain2:
> hey how about we don't put closes for someone else's PR?

their pr is better

MajManatee (approved):
those gasses sure are funny

PR #16909, opened 2022-12-03
skrem broke nukeops

TheRyeGuyWhoWillNowDie:
subtyping makes this have flags to both exclude and include nukeops which the uplink does not like

# Changelog

:cl:  
bugfix: superior augmentation kit can be purchased again
/:cl:


Moltijoe (approved):

Moltijoe (changes requested):
Remember to also set surplus = 0
Default surplus for the /implants/ datum is 50

Moltijoe (approved):

PR #16912, opened 2022-12-03
Pod people now have water for blood, and 50% chance to have a random reagent as blood instead

Simplehorror:

gives the chemist more to do, also podpeople shouldn't have human blood and I think water is a good substitute for a plant. reagent thing should bring more interactions.

# Spriting

# Wiki Documentation

no idea where blood type would be sorry
# Changelog

:cl:  
tweak: Pod people now have water for blood, and 50% chance to have a random reagent as blood instead
/:cl:


GraveHat:
This prevents pods from being pod cloned who already get water as blood after being pod cloned so this seems like an unnecessary nerf.

Simplehorror:
oh fuck that's right

Simplehorror:
closed until i fix that with a blood component

PR #16925, opened 2022-12-04
Tweaks how the speed potion works

Moltijoe:
First round with extensive testing, probably a bit strong
Now it applies an even -1 to speed so it's equally as effective on all simplemobs rather than being busted on slower simplemobs
This is pretty much a nerf to most uses of it
However, this lets it work on already really fast simplemobs like drones and spiderbots now

:cl:  
tweak: Speed potion now reduces speed by 1 rather than setting it to -1
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/xenobiology.dm
+++ b/code/modules/research/xenobiology/xenobiology.dm
@@ -898,11 +898,11 @@
 		to_chat(user, span_warning("The potion can only be used on simple animals!"))
 		return
 	var/mob/living/simple_animal/zipzoom = target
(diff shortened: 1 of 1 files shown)


UltraBigRat:
Time to reach lightspeed with my chocobo. But yeah this would be a much better way of doing it probably, since the slowest simplemobs tend to be the ones that hit like a truck.

MajManatee (approved):
Gold slime users deserve the death penalty anyways

Maia-J:
> Gold slime users deserve the death penalty anyways

Brings souls back into game
hEhE gOlD bAd

PR #15785, opened 2022-09-16
Removes mindswap from the spellbook (only)

Sniblet:
# Document the changes in your pull request

For admin reasons. Its theoretical existence causes friendly fire that is rough to deal with in tickets (or so I'm told)

This removes mindswap from the spellbook but leaves it for robeless apprentices.

# Wiki Documentation

It's not in the spellbook!

# Changelog

:cl: 
rscdel: The Wizard Federation has somehow managed to make their recordings of the spell Mindswap mindswap itself with a spirit of forgetfulness. Robeless apprentices have been able to remember it!
/:cl:

PR #16939, opened 2022-12-04
Revert "Removes mindswap from the spellbook (only)"

Simplehorror:
Reverts yogstation13/Yogstation#15785

since when is "it gets a bit confusing for admins" a reason to remove something? if it is confusing to admins, then tooling needs to be fixed and not this innocent spell

### DIFF
--- a/code/modules/antagonists/wizard/equipment/spellbook.dm
+++ b/code/modules/antagonists/wizard/equipment/spellbook.dm
@@ -155,6 +155,11 @@
 	spell_type = /obj/effect/proc_holder/spell/pointed/trigger/blind
 	cost = 1
 
+/datum/spellbook_entry/mindswap
+	name = "Mindswap"
+	spell_type = /obj/effect/proc_holder/spell/targeted/mind_transfer
+	category = "Mobility"
+
 /datum/spellbook_entry/forcewall
 	name = "Force Wall"
 	spell_type = /obj/effect/proc_holder/spell/targeted/forcewall


Bibby0110:
fix the tooling then

Simplehorror:
there's nothing wrong with it aside admins being whiny

neeshacark:
staff are perfectly capable of seeing who occupied a body at any given time, we literally have replays and logs also show ckeys and we can see all the people who occupied a body. i don't understand the mindswap removal 

if friendly fire happens it happens, causing confusion is literally the purpose of the spell

Hislyna:
As stated before, it causes confusion which ends up in rule breaks. It was removed for a reason and we do not need spells which can be used as a means for accidental or intentional grief.

Moja10:
No, it is legitimately one of the worst admin headaches caused by in-game items/abilities, the problem is not whether admins can track who swapped who, when, where. What the headache is is keeping track of PLAYERS knowing who swapped who, when, where. There could be numerous witnesses, numerous swaps, swaps missed by players, missing information all the time in these events. It is not reasonable at all to track who is keeping track of what swaps, what swaps they missed, what actions they saw per swap, all for numerous different people involved in the shenanigans. It routinely leads to people committing wrong actions on wrong information, which in some circumstances could be totally fine and SS13 classic experience, in other circumstances, the wrong people get beat to death on a witchhunt because someone swore on their mum that he really really was the wizard and needed to be beat to death. 

I would say on average, confusing for admins isn't a solid reason on its own to remove something, but in this circumstance, I think it is gone for good reason.

Simplehorror:
> As stated before, it causes confusion which ends up in rule breaks. It was removed for a reason and we do not need spells which can be used as a means for accidental or intentional grief.

right, that's the point of the spell

Moltijoe (changes requested):
It was removed because of it being unfun for victims and more so for causing admin headaches, neither of these problems have been addressed

Simplehorror:
fighting a wizard isn't always going to be fun?

Moltijoe:
> fighting a wizard isn't always going to be fun?

https://www.youtube.com/watch?v=FdxEUL18ySY

Simplehorror:
in that vein we should also remove fireball, the traps, spacetime distortion because it's not fun for the victim

Moltijoe:
> in that vein we should also remove fireball, the traps, spacetime distortion because it's not fun for the victim

different sort of unfun, those are temporary, body swap is permanent and can leave you permanently crippled for the round
a better comparison would be the staff of change

Moltijoe:
> different sort of unfun, those are temporary, body swap is permanent and can leave you permanently crippled for the round a better comparison would be the staff of change

or changeling transformation sting, which i specifically coded to be temporary

Simplehorror:
ok so make mindswap revert after so much time (this kills the spell and any fun you could have had by turning into the captain)

Moltijoe:
> ok so make mindswap revert after so much time (this kills the spell and any fun you could have had by turning into the captain)

wizard isn't the role-steal antag, that's changeling

Simplehorror:
wizard is the do whatever the fuck he wants antag until admins get mad that an antagonist is being antagonistic apparently

Bibby0110:
stale

PR #16940, opened 2022-12-04
adds a line to the syndie fedora so people know what the fuck it does

Skrem7:

People don't know what this does SO MUCH SO that a bug report was submitted! FUCKING CRAZY!

# Changelog

:cl:  
spellcheck: Syndie fedora has a description change to say you can extend its blades now
/:cl:

### DIFF
--- a/code/modules/clothing/head/jobs.dm
+++ b/code/modules/clothing/head/jobs.dm
@@ -103,7 +103,7 @@
 /obj/item/clothing/head/det_hat/evil
 	name = "suspicious fedora"
 	icon_state = "syndicate_fedora"
-	desc = "A suspicious black fedora with a red band."
+	desc = "A suspicious black fedora with a red band. It can be activated in-hand to extend or retract razor blades which cause significant damage when thrown."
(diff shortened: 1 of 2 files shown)


code/modules/clothing/head/jobs.dm
```diff
@@ -103,7 +103,7 @@
 /obj/item/clothing/head/det_hat/evil
 	name = "suspicious fedora"
 	icon_state = "syndicate_fedora"
-	desc = "A suspicious black fedora with a red band."
+	desc = "A suspicious black fedora with a red band. It can be activated in-hand to extend or retract razor blades which cause significant damage when thrown. Also heavily armored."
```
ReddicusDragon, line 106: won't this make it so that normal, non traitor players can read this description too?

Skrem7: The hat isn't metashielded and it's not at all obvious to people who don't know what it does

Moggram: It may need it, Sec let me keep it one time when I was in time out.

Skrem7: would require a lot more work and a separate pr which would have a different goal

MajManatee (approved):
hat

PR #16334, opened 2022-11-02
gutting flashes so that they're actually personal defense weapons and not the best stun in the game attempt 2

TheRyeGuyWhoWillNowDie:
# Hello jamied12! I'd like to ask you to read this section before closing this PR. This change is for handheld flashes, the melee weapon. This does NOT touch flashbangs, the ranged stun grenades. They will work how they currently do.

This affects NONE of the following:
Mounted flashes
Portable flashes

All this does is make it so handheld flashes disarm instead of hardstunning for 12 seconds. This is still good as flashes ignore shields, but it's no longer win in 1 click territory, and 'have the counter or you lose invariably' combat is not fun and being removed.
Makes the blinding stronger so that the box they come in that says they can cause blindness isn't false advertisement, also makes it possible to actually blind your attackers to get away. Being temporarily nearsighted or permanently blinded is bad but it's not instant lose scenario like being locked into a chainstun like how current flashes work. I have witnessed many many people get clicked on by this little green tumor for free instant wins and it makes me very sad.
# Changelog

:cl:  
tweak: flashes are now just a disarm, the stun is GONE
tweak: flash blindness has been doubled and confusion has been tripled to keep it competitive
/:cl:

PR #16950, opened 2022-12-05
non-antag targeted flashes against humans will burn out after one use

TheRyeGuyWhoWillNowDie:
Title.

Things this does not affect: AoE flashes (they can still burn out normally, although it will work like how it does currently) Portable flashers Mounted flashers Traitor flash cameras Revolutionary flash implants Targeted flashes against borgs because everyone would shit themselves that they can't chainstun borgs with 0 effort

Still better than stun batons, their boon is that they can't be used against you if you drop them and they're an instant hardstun, but now they're single use instead of being a direct upgrade. Security techfabs can print flashes now, so you're not at any risk of running out. Still good at arresting single targets, but the point of this is now you have to put in a tiny amount of effort into actually stunning people instead of chainflashing huge groups with no effort. Cyborgs are worse, but everyone hates cyborgs so this shouldn't be a problem, and emagged/malf cyborgs have their malf modules to stun and attack people with, so they don't lose out.

:cl:  
tweak: handheld flashes, when targeted against humans, will burn out in one use
/:cl:

### DIFF
--- a/code/modules/assembly/flash.dm
+++ b/code/modules/assembly/flash.dm
@@ -27,6 +27,8 @@
 	var/last_trigger = 0
 	///can we convert people to revolution
 	var/can_convert = FALSE
+	///will this burn out on use? only 2 that wont do this currently are the rev and camera flash
+	var/can_burn_out = TRUE
 
 /obj/item/assembly/flash/suicide_act(mob/living/user)
 	if(burnt_out)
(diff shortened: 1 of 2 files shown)


neeshacark:
this is not how you nerf flashes...

literally just make them not a hardstun

adamsong:
This is not a great idea, as it just makes flashes less than useless, and can be used as an antag test

TheRyeGuyWhoWillNowDie:
> literally just make them not a hardstun

Yes, I tried this.
#16334
#16136

MajManatee:
this causes me physical harm

TheRyeGuyWhoWillNowDie:
> This is not a great idea, as it just makes flashes less than useless, and can be used as an antag test

1. Instant hardstun in 1 click on somebody is still pretty good I think

2. Antag flashes as in the traitor camera flash and the rev flash implant, any antagonist with a normal flash will still burn out the same

Cartlord:
Make it still burn out in one use against borgs. It's better for consistency, and if borgs become unkillable because one tool ceases to hard-counter them, they need a deep inspection.

Simplehorror:
you're nerfing a ton of antags by trying to kill flashes; so many antags just use flashes for stunning

Moltijoe:
image

PR #16924, opened 2022-12-04
Engineering hardsuit gets full rad resistance

Simplehorror:
# Document the changes in your pull request

atmos hardsuit gets full fire + heat resistance... engi hardsuit gets 75% rad resistance? what a joke.
the singularity got buffed so this is required to safely go near it

# Spriting

# Wiki Documentation

change the armor values in the hardsuit page

# Changelog

:cl:  
tweak: Engineering hardsuit gets full rad resistance
/:cl:

PR #16952, opened 2022-12-05
Reduces the slowdown on radsuits

Moltijoe:
They are literally only useful for rad protection, which is already super limited
This lets non rad-immune species not get a ridiculous slowdown, worse than space suits, when needing to deal with radiation
also with #16924 merged, they're strictly worse than the hardsuit now

:cl:  
tweak: Reduces the slowdown on radsuits
/:cl:


MajManatee (approved):
makes sense

PR #16953, opened 2022-12-05
Reduces flash paralyze duration on paralyzed targets

Moltijoe:
Might as well throw my hat in the ring

This will either force people to burn out their flash faster to keep them down
Or give them a brief moment where they can escape or fight back

:cl:  
tweak: Flashes paralyze for a shorter duration on already paralyzed targets
/:cl:

### DIFF
--- a/code/modules/assembly/flash.dm
+++ b/code/modules/assembly/flash.dm
@@ -138,7 +138,10 @@
 				to_chat(M, span_userdanger("[user] blinds you with the flash!"))
 			else
 				to_chat(M, span_userdanger("You are blinded by [src]!"))
-			M.Paralyze(rand(80,120))
+			if(M.IsParalyzed())
+				M.Paralyze(rand(20,30))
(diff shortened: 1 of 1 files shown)


Simplehorror:
nice. if there must be something done about flashes for some reason I prefer this

ynot01:
10 seconds of stun is death/cuff, this doesnt do much besides making group arrests harder i guess?

Moltijoe:
> 10 seconds of stun is death/cuff, this doesnt do much besides making group arrests harder i guess?

Not like any other change is getting approved

Moltijoe:
> 10 seconds of stun is death/cuff, this doesnt do much besides making group arrests harder i guess?

also, maybe it's just me, but i don't think a single secoff should be able to easily make group arrests with just a flash

MajManatee (approved):
chain flashing was cringe anyways

PR #16954, opened 2022-12-05
Wheelies scooter is slower

Moltijoe:
No reason for it to be faster than the airshoes traitor item
Also makes the crash effect more devastating since it's easier to not crash

:cl:  
tweak: Wheelies moves slower but crashing is worse
/:cl:


Moltijoe:
eh, i wasn't super keen on this one

PR #16955, opened 2022-12-05
Increase run speed slightly

Simplehorror:

QOL life. Everyone likes going faster, but when we walk right now it's very very sluggish. Especially if you're hungry - it compounds into an awful experience to where you can't get anywhere. I believe some people have been trapped in between firelocks because they cant move fast enough for the firelock to stay open. There's no reason for speed to be THIS slow anymore ever since we did a ton of performance fixes.
# Spriting

# Wiki Documentation

Run speed goes from 2 to 1.7, walk from 6 to 5
# Changelog

:cl:  
tweak: Run speed goes from 2 to 1.7, walk from 6 to 5
/:cl:

### DIFF
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -47,8 +47,8 @@ EMOJIS
 ## To speed things up make the number negative, to slow things down, make the number positive.
 
 ## These modify the run/walk speed of all mobs before the mob-specific modifiers are applied.
-RUN_DELAY 2
-WALK_DELAY 6
+RUN_DELAY 1.7
+WALK_DELAY 5
 
 ## The variables below affect the movement of specific mob types. THIS AFFECTS ALL SUBTYPES OF THE TYPE YOU CHOOSE!
 ## Entries completely override all subtypes. Later entries have precedence over earlier entries.


Mqiib:
what if we didnt

ynot01:
tg player

would require testmerge before actual merge

ToasterBiome:
I don't like going fast, I think the movement on other servers that are faster is weird. I think yogs is a good speed. I don't know why you're so hungry all the time and suffer from the slowdown so often, just eat

Simplehorror:
our current run speed is what walking speed should be

ToasterBiome:
> our current run speed is what walking speed should be

no

Cartlord:
I'm not a fan of this idea. I've already had scenarios where people can steal shit or break into somewhere and leave before Sec can even be called, and I think this'd just make it worse.

Simplehorror:
so just report their name, if they can do it now they can still do it after this

Moltijoe:
A lot of things are currently balanced around current speeds
This would cause problems
not to mention, the game can only handle things going so quickly, so it would reduce the effectiveness of stacking speed boosts

Simplehorror:
> A lot of things are currently balanced around current speeds This would cause problems not to mention, the game can only handle things going so quickly, so it would reduce the effectiveness of stacking speed boosts

a lot of things are balanced around tg speed too

PR #16957, opened 2022-12-05
Showers splash for less

Moltijoe:
now instead of splashing for 200u per tick
they splash for 100u

this is equal to -10 fire_stacks rather than -20
this happens both on initial enter and each tick

this still lets it put out fire instantly but hopefully not have any preternis that either accidentally walks through or gets bumped into it be sentenced to death

:cl:  
tweak: Showers splish splash less
/:cl:


Moltijoe:
nah, i've changed my mind on this one, just don't fuckin get wet

PR #16964, opened 2022-12-05
38 speedloaders are normal sized

TheRyeGuyWhoWillNowDie:
no detective needs to have boxes and boxes of ammo in their bag for their emp-proof lethal revolver, and now that the bullets are exclusively lethal (thanks skrem) detectives have no excuse for this outside of trying to valid kill people for breaking into eva
note that detective shoulder holster can still hold 2 speedloaders and you can carry more in your backpack if you really need to

# Changelog

:cl:  
tweak: 38 speedloaders are normal sized
/:cl:

### DIFF
--- a/code/modules/projectiles/boxes_magazines/ammo_boxes.dm
+++ b/code/modules/projectiles/boxes_magazines/ammo_boxes.dm
@@ -49,6 +49,7 @@
 	max_ammo = 6
 	multiple_sprites = AMMO_BOX_PER_BULLET
 	materials = list(/datum/material/iron = 20000)
+	w_class = WEIGHT_CLASS_NORMAL
 
 /obj/item/ammo_box/c38/hotshot
 	name = "speed loader (.38 Hot Shot)"


LoliconSlayer:
i ded

Skrem7:
.357 speed loader
- more bullet
- scarier bullet
- still small sized

wt magazine
- MUCH more bullet
- still small sized

TheRyeGuyWhoWillNowDie:
> skrem

autorifle is a cargo weapon and the 357 revolver is only from uplinks, i'm not opposed to making it normal sized too but you caused this issue in the first place

UltraBigRat:
> .357 speed loader
> 
> * more bullet
> * scarier bullet
> * still small sized
> 
> wt magazine
> 
> * MUCH more bullet
> * still small sized

It is a bad idea to compare roundstart crew weapons to traitor and cargo armory ones

CMOisLing:
Seems like this is more of an admin issue






Also ided

Moltijoe:
i do kinda agree that detective should be focused on... detective work? and being able to stockpile ammo for their gun seems unnecessary for that. A detective should be more like the bloodhound of security, finding who to arrest, not actually getting in firefights and killing.

MajManatee:
I'm actually not opposed to this. 

I don't agree with the whole "Other ammo is small" argument, we shouldn't avoid trying to balance something just because something similar is unbalanced.

Detectives really shouldn't need more than a mag or two on them. If its a dire emergency, they would probably have a laser by that point, or something similar.

Their revolver is a last-ditch emergency weapon, and this would push it more towards that.

ynot01:
Every other magazine in the game, with rare exception (nukie heavy weaponry) is small or tiny

RG4ORDR:
> i do kinda agree that detective should be focused on... detective work? and being able to stockpile ammo for their gun seems unnecessary for that. A detective should be more like the bloodhound of security, finding who to arrest, not actually getting in firefights and killing.

If you play detective there is a very high likelyhood no matter your effort short of a full sec-team you're going to end up shooting the revolver. If you're doing your job at finding clues, snooping, so forth you're gonna end up getting attacked simply for being nosy. Detective getting punished for doing their job by reducing their ranged offense is silly billy.

TheRyeGuyWhoWillNowDie:
> Detective getting punished for doing their job by reducing their ranged offense is silly billy.

Their job is investigating. Want to valid people? Play security officer.

Skrem7:
better idea: make .38 only printable from the armory much like every lethal shotgun shell and shift the autolathe recipe to hacked autolathe

can observe how it goes down after that

would maybe invoke some sense of scarcity, obvious concern is that of det just flocking to cargo every single round (i know i did when i ran .357 far too often and didn't realize how powergame-y it was), but that actually requires det to be able to access the lathe rather than just walk back into the office and mash the lathe button (also takes longer)

forhiti:
Bad change. Just give them a disabler or something if you are so worried about how the revolver is lethal now.

TheRyeGuyWhoWillNowDie:
> better idea: make .38 only printable from the armory much like every lethal shotgun shell and shift the autolathe recipe to hacked autolathe

this is actually not a better idea because now detectives will break into cargo and deplete all of the metal on the station to make 50 speedloaders since they're expensive from the autolathe
or option 2 they build their own autolathe (powergaming) because autolathe boards can be made from the secfab and drain all the metal that way

Skrem7:
> > better idea: make .38 only printable from the armory much like every lethal shotgun shell and shift the autolathe recipe to hacked autolathe
> 
> this is actually not a better idea because now detectives will break into cargo and deplete all of the metal on the station to make 50 speedloaders since they're expensive from the autolathe or option 2 they build their own autolathe (powergaming) because autolathe boards can be made from the secfab and drain all the metal that way

1) ahelp this
2) ahelp this

pov i am going to drain all the metal either way if I print a morbillion .38 loaders

TheRyeGuyWhoWillNowDie:
> pov i am going to drain all the metal either way if I print a morbillion .38 loaders

so here's a whacky idea... what if it was less beneficial to make a morbillion speed loaders? something like making them larger and harder to carry en masse... 🤷‍♂️

Moltijoe:
Creates unnecessary inconsistencies between speed loaders
detective valid hunting is a pretty easy callout for ahelping
the bullets aren't even that good with the negative armour pen making them do almost nothing on proper antags

PR #16974, opened 2022-12-06
Revert "makes .38 a fully lethal round as it should be"

Simplehorror:
Reverts yogstation13/Yogstation#16722

i look away for ONE second and someone does this? now people are whining it is being used to validhunt and trying to nerf it, when there was 0 issues with the old one. it's supposed to be a self-defense weapon, and all of our self-defense weapons are purely non-lethal (See HOP's mini disabler)

### DIFF
--- a/code/modules/projectiles/boxes_magazines/ammo_boxes.dm
+++ b/code/modules/projectiles/boxes_magazines/ammo_boxes.dm
@@ -53,21 +53,21 @@
 /obj/item/ammo_box/c38/hotshot
 	name = "speed loader (.38 Hot Shot)"
 	desc = "A six-shot speed loader designed for .38 revolvers. \
-			These rounds trade some damage for an incendiary payload which sets targets ablaze."
+			These rounds trade exhaustive properties for an incendiary payload which sets targets ablaze."
 	icon_state = "38hot"
 	ammo_type = /obj/item/ammo_casing/c38/hotshot
 
 /obj/item/ammo_box/c38/iceblox
 	name = "speed loader (.38 Iceblox)"
 	desc = "A six-shot speed loader designed for .38 revolvers. \
-			These rounds trade some damage for a cryogenic payload which significantly reduces the body temperature of targets hit."
+			These rounds trade exhaustive properties for a cryogenic payload which significantly reduces the body temperature of targets hit."
 	icon_state = "38ice"
 	ammo_type = /obj/item/ammo_casing/c38/iceblox
 
 /obj/item/ammo_box/c38/gutterpunch
 	name = "speed loader (.38 Gutterpunch)"
 	desc = "A six-shot speed loader designed for .38 revolvers. \
-			These rounds trade some damage for an emetic payload which induces nausea in targets."
+			These rounds trade exhaustive properties for an emetic payload which induces nausea in targets."
 	icon_state = "38gut"
 	ammo_type = /obj/item/ammo_casing/c38/gutterpunch
 
--- a/code/modules/projectiles/projectile/bullets/revolver.dm
+++ b/code/modules/projectiles/projectile/bullets/revolver.dm
@@ -14,14 +14,16 @@
 
 /obj/item/projectile/bullet/c38
 	name = ".38 bullet"
-	damage = 25 //High damaging but...
-	armour_penetration = -40 //Almost doubles the armor of any bullet armor it hits
-	wound_bonus = -10
+	damage = 15 // yogs - Nerfed revolver damage
+	//knockdown = 60 //yogs - commented out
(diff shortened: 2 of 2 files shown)


Simplehorror:
ok listen guys its OK to try things but when they dont work out (like this) its also OK to revert them

ToasterBiome:
> all of our self-defense weapons are purely non-lethal (See HOP's mini disabler)

HOP's disabler has a kill mode 😳

Simplehorror:
> > all of our self-defense weapons are purely non-lethal (See HOP's mini disabler)
> 
> HOP's disabler has a kill mode 😳

(if you use the kill mode you'll get like 2 shots (and nobody ever uses this or they'll be banned))

Simplehorror:
objectively it's WORSE as a self-defense weapon because now you can't defend yourself against a bloodsucker, changeling or vampire or anyone who has healing/ballistic armor capabilities

ynot01:
detective is now a force to be reckoned with

doing stamina damage encourages follow-up i.e. an arrest, validhunting

being purely lethal means you have an excellent deterrent because, best case for them, they're going to walk away with some nasty damage instead of useless stam that wears off immediately and cant be capitalized if the detective is alone, which he often finds himself in because there's little incentive to follow around officers unless you want to validhunt with them

Simplehorror:
> detective is now a force to be reckoned with
> 
> doing stamina damage encourages follow-up i.e. an arrest, validhunting
> 
> being purely lethal means you have an excellent deterrent because, best case for them, they're going to walk away with some nasty damage instead of useless stam that wears off immediately and cant be capitalized if the detective is alone, which he often finds himself in because there's little incentive to follow around officers unless you want to validhunt with them

detective is not a fucking force to be reckoned with when anyone who has armor can't be put down in 6 shots. stamina crit = time to run away. still alive = they shoot me with their ebow again and i'm still fucking dead despite how much "damage" i put in them

Skrem7:
me when i stam crit with .38 and simply pull out my infinite ammo melee weapon to stunlock them forever (i can alternate between stamina hit and just killing them anyway) ((the baton is good enough at disabling them even without the revolver, it's arguably better than a stun baton))

how can people make the argument "detective isn't a force to be reckoned with" but also say "detective validhunt is too potent now" like I'm genuinely confused

Simplehorror:
> me when i stam crit with .38 and simply pull out my infinite ammo melee weapon to stunlock them forever (i can alternate between stamina hit and just killing them anyway) ((the baton is good enough at disabling them even without the revolver, it's arguably better than a stun baton))
> 
> how can people make the argument "detective isn't a force to be reckoned with" but also say "detective validhunt is too potent now" like I'm genuinely confused

they can validhunt traitors, not bloodsuckers. are you thick?

Skrem7:
in the past: 15 brute and 35 stamina

assume i am a traitor fighting sec with no armor because i engaged unprepared

the detective crits me in three shots (lol)

if i get armor polish or a basic vest for 30 bullet armor the stamina damage of the bullet is now 24.5 which is still super similar to a disabler at 25 base stamina (my 10 energy armor will make a disabler 22.5; the revolver still does more!)

assume current .38 round now (25 brute and -40 AP)

if i don't get armor i am crit in 4 shots (one more than before!) BUT i can use healing meds that are fairly easily obtained in prep from a variety of sources rather than trying to shake off stamina damage that isn't easy to heal mid-combat unless im specific antags which should already have a massive leg up

let's assume i do the sane thing and get armor against this lethal valid detective menace and get a vest of effectively 50 armor due to -40 AP and drop that bullet down to 12.5 damage (almost half the previous effective value)

if you go up against an officer without proper prep you should lose, if you can't disengage in this case you should lose, this has always been true and this 38 change doesn't actually alter how a lot of those encounters end

if you give the det a revolver that plinks against an antag that should body them then they might have the wonderful idea to retreat (most secoffs never do this) which is ideally what should be happening anyway rather than the det pursuing a kill with their mini beanbag thrower

Moltijoe:
This buffs detective, this much stamina damage in addition to the removal of negative armour pen makes them capable of doing far more than they should be expected

JamieD1:
Revert Rules

Moltijoe:
> Revert Rules

Sorry, the rules only said merging was off limits
I had assumed closing was fine

Hopekz:
> detective is now a force to be reckoned with

A detective shouldn't be a force to be reckoned with imo he's a detective.

Imo Detective should be given more tools to do his actual job versus being more focused towards valid hunting

ToasterBiome:
I've been having a lot of fun with the new det gun by shooting random people with it. Much more dangerous now. Thank you gods.

PR #16992, opened 2022-12-07
MC backport

fira:
# Changes in your pull request

What this does:
 * Updates MC, SSgarbage, SStimer from /tg/
 * Notably adds MC init stages, allowing firing during init for things like Statpanel
 * Adjusts Susbystems Initialization to match new scheme (Init proc returns a hint instead of using parent call)
 * Various defines required by these systems

What this potentially breaks:
 * Find Reference support isn't included as it got separated somewhat and i don't know the ins and outs of it
 * `COMSIG_PARENT_QDELETING` now fires before `Destroy` rather than incorrectly after
 * MC now requires rust-g for init timing. The functions are not present in older versions. **You want to MAKE SURE that version in dependencies.sh support them BEFORE you run this on TGS with EventScripts auto build or you WILL have a bad time.**
 * Probably more

What was tested:
 * Building, Proper Initialization with progress indicator, Game start
 * Basic game functionality such as objects interaction, movement, etc
 * Similar porting implementation on another codebase with live testmerging
 * Error log
 
What was (notably) NOT tested:
 * Atmos shenanigans
 * Webhooks (there's one moved within MC code)
 * Online Database testing
 * Any and everything else that might break randomly by side effects like Destroy change

# Changelog

:cl:  
tweak: Updated some backend systems, such as the /tg/ Master Controller
/:cl:


fira:
missed a GLOB.admins replace in #ifdef TESTING codeblock that got picked up by CI

JamieD1:
Thank you fira

MajManatee:
Excitin stuff here

JamieD1 (approved):

PR #16994, opened 2022-12-07
Medical job account age requirements

Moltijoe:
image

You know why this is being asked for

:cl:  
tweak: adds account time restrictions to all medical jobs 
/:cl:

PR #16995, opened 2022-12-07
deletes roundstart morphine (at least on box because i dont care about other maps)

TheRyeGuyWhoWillNowDie:
i dont think it even needs explaining why #16994 is a horrible idea, and you know you've fucked up if i'm the voice of reason on the topic of game design

# Changelog

:cl:  
rscdel: roundstart morphine is gone
/:cl:


TheRyeGuyWhoWillNowDie:
bibby missed the whole reasoning for this, but it's fair since it was one of my schizo rants in development public, basically molti wanted to **prevent new players from playing as ANY medical role** because one ban evader uses morphine spray to grief people. this is like carpet bombing a spider with the amount of overkill it was, and removing the morphine removes the grief tool

MajManatee (changes requested):
Change on all maps

TheRyeGuyWhoWillNowDie:
No

Hopekz:
> No

I wouldn't either

PR #16996, opened 2022-12-07
Mining medic time requirement

Moltijoe:
Mining medic is a limited job with only 1 slot
it requires knowledge of both mining and medical to be even the least bit effective
There's no reason to believe a new player would be capable of this job
It isn't even a good job for learning because of how it is just sitting around waiting for miners to get hurt

The requirement isn't intensive, 
4 days since first log in and 2 hours in medical is more so to prevent fresh logins from choosing it and being overwhelmed
:cl:  
tweak: Adds a mining medic time requirement
/:cl:

### DIFF
--- a/yogstation/code/modules/jobs/job_types/mining_medic.dm
+++ b/yogstation/code/modules/jobs/job_types/mining_medic.dm
@@ -8,6 +8,10 @@
 	spawn_positions = 1
 	supervisors = "the chief medical officer and the quartermaster"
 	selection_color = "#d4ebf2"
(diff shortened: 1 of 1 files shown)


MajManatee:
No time requirement as... being a miner?

Moltijoe:
> No time requirement as... being a miner?

we don't have combined requirements as an option, so i considered medical knowledge to be more important
we don't have job specific requirements either, so that would mean even cargo tech would count towards experience

MajManatee (approved):
I guess that makes sense

PR #16999, opened 2022-12-07
Makes spray bottle stream mode do touch application of reagents

Moltijoe:
image

it's a constant stream of reagent, why would it count as vapour, it's the equivalent of a water gun
spray mode still counts as vapour

This differentiates the modes more than just range and reagent transfer amount
Allows certain chems to apply different effects
Water makes people wet instead of putting water in their body (preternis nerf)

:cl:  
tweak: Spray bottle stream mode applies reagents via touch rather than vapour
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/spray.dm
+++ b/code/modules/reagents/reagent_containers/spray.dm
@@ -112,12 +112,12 @@
 				if(isliving(T))
 					var/mob/living/M = T
 					if((M.mobility_flags & MOBILITY_STAND) || !range_left)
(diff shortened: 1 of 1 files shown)


UltraBigRat:
does this mean its easier to spray long lines of lube, cause if so. Nice

Moltijoe:
> does this mean its easier to spray long lines of lube, cause if so. Nice

No, this only changes the short range move of the spray bottle

Bibby0110 (approved):

PR #17003, opened 2022-12-08
Shadowling veil nullifies glowy for ALL TIME

Sniblet:
I'VE HAD IT

See round 44257 for evidence that slings need SOMETHING for glowy. Mutadone requires accessing medbay, which is hard for loud antags at the best of times. This is not the best of times because medbay is full of hostile DoT fields.
It only takes one geneticist with lifeticks to checkmate any sling game.

"waa waa this is their only counter" play a sling round sometime. Glowy isn't a counter, it's a win condition, and you can reach it in 5 minutes.

closes #17004 

Veil removes the glowy mutation.

# Changelog

:cl:  
tweak: Veil removes glowy. Bring spare mutators.
/:cl:

### DIFF
--- a/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
+++ b/yogstation/code/modules/antagonists/shadowling/shadowling_abilities.dm
@@ -140,18 +140,7 @@
 		var/mob/living/carbon/M = H
 		var/datum/mutation/human/glow/G = M.dna.get_mutation(GLOWY)
 		if(G)
-			G.glowth.set_light(0, 0) // Set glowy to no light
-			if(G.current_nullify_timer)
-				deltimer(G.current_nullify_timer) // Stacks
-			G.current_nullify_timer = addtimer(CALLBACK(src, .proc/giveGlowyBack, M), 5 SECONDS, TIMER_STOPPABLE)
-
-/obj/effect/proc_holder/spell/aoe_turf/proc/giveGlowyBack(mob/living/carbon/M)
-	if(!M)
-		return
-	var/datum/mutation/human/glow/G = M.dna.get_mutation(GLOWY)
-	if(G)
-		G.modify() // Re-sets glowy
-		G.current_nullify_timer = null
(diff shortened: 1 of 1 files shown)


MajManatee:
i suggest sling enjoyers get copium from cargo. can be ordered in bulk

Sniblet:
This isn't about removing the gamemode this is about making it a competition

forhiti:
I think making it turn off glowy for long enough for conversion to happen, and IF conversion happens then it will remove glowy. This still lets glowy be counterable but also still useful.

Therandomhoboo:
No offence, awful idea. Be literally OP for shadowlings to be able to tap veil or w/e and then basically ruin people. They have a stun anyway ontop of being able to fuck light sources up

I would prefer it temporarily turned it off, but it shouldn't be for a long ass duration for conversion at least otherwise again. It'd be pointless to have glowy mutations, the whole point is to help counter the shadowlings. 

So make it a temp thing and either increase the veil CD (because at that point it will be OP) or make it that glowy comes back on after 2 seconds so at least the antag has to focus on spamming it quickly or dropping them. (I believe you can use veil/glare to keep the person still without breaking the conversion)

Therandomhoboo:
Or honestly, make the smokescreen they can drop actually block the glowy lights.

Sniblet:
Smokescreen is only unlocked if they have a lot of thralls. Glowy rushes make having more than 2 thralls and a breadcrumb impossible. Play a sling round sometime

Therandomhoboo:
> Smokescreen is only unlocked if they have a lot of thralls. Glowy rushes make having more than 2 thralls and a breadcrumb impossible. Play a sling round sometime

Yeah but at least it gives incentive to get to the smokescreen and im sure it only requires around 4 thralls (at least thats how many i needed to get it) And I've played my fair share of sling rounds thanks to know what you're doing is insanely daft

Simplehorror:
if you need a win condition as shadowling, have you tried goblin barrel?

PR #16991, opened 2022-12-07
Resprites the power fist because nothing is better then losing your weapon to the grey floors down below

KillerOrcaCora:
how was I supposed to ever know the reference it could have literally been anything. Resprites the powerfist to actually look like its reference. 

# Spriting
image

Capture
***Flash***
image
***Flash*
*Flash*
*Flash***

but we already have a powerfist! the powerfist at home - 
image

# Wiki Documentation
are they still alive? let's give them this just incase. 
thepowerfist

# Changelog

:cl:  

imageadd: Resprites that one traitor item you never use because it phases into the traitor panel color like a Vietnam solider.  

/:cl:

PR #17008, opened 2022-12-08
Revert "Resprites the power fist because nothing is better then losing your weapon to the grey floors down below"

Simplehorror:
Reverts yogstation13/Yogstation#16991

the old powerfist sprite was better and as the only person to use this we should revert it


KillerOrcaCora:
Oh…he’s actually serious…that’s a damn shame 😔

cuackles:
Lol, no, the fact that you don't like the sprite doesn't make it bad, this new sprite has more clarity , it's notable that is a fist and resembles more the original material, even if we excluded that, being "the only one that uses the item" does not give you the right to decide how something should look, i cannot even think of how you are able to see the news sprite as worst that the old one, the old one way literally a block of grey and black

cuackles:
Adding to that 
IMG_20221208_094429
Not game breaking and not 48 hours

MajManatee:
See above

PR #17011, opened 2022-12-08
Disabler now shoots twice as fast

Simplehorror:

I think the most annoying part about having to use a disabler is the fact that it shoots as fast as a taser but it still takes so many clicks to down someone. I think you should be able to fire faster because even if you have someone in a corner, it's still going to take like 10 seconds to down them because of how slow it is. 

# Spriting

# Wiki Documentation

# Changelog

:cl:  
tweak: Disabler now shoots twice as fast
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition/energy/stun.dm
+++ b/code/modules/projectiles/ammunition/energy/stun.dm
@@ -21,6 +21,7 @@
 	e_cost = 50
 	fire_sound = 'sound/weapons/taser2.ogg'
 	harmful = FALSE
+	click_cooldown_override = CLICK_CD_RAPID
 
 /obj/item/ammo_casing/energy/disabler/cyborg
 	e_cost = 100


thegoldencat413:
._.

Maia-J:
1.5x. Take it or leave it.

Simplehorror:
1.5x could be good, but I think this is the sec buff we've been needing

00ze-cyclone:
Skill issue, just use King Crimson to erase the time of the cooldown

Moltijoe:
better than them relying solely on flashes and stun batons, but 2x is probably too fast

00ze-cyclone:
> better than them relying solely on flashes and stun batons, but 2x is probably too fast

They already rely a lot on the disabler, that shit is already cosmically robust with it's ability shoot a fuckton of beams, very good damage that is not very well stopped by most armors and able to be used from the safety of behind of one of the 1237042874 reinforced windows that will eat 10 bullets before breaking.

ReddicusDragon:
solution: delete sec

Moltijoe:
they're good in their current spot, a good security officer should use different tools based on the situation

PR #17020, opened 2022-12-08
Gives brig physician nitrile gloves instead of latex gloves

Moltijoe:
Nitrile gloves are the upgraded version of latex gloves
Weird that security wouldn't be able to supply them for their physician

:cl:  
rscadd: Adds a pair of nitrile gloves to brig physician's locker
tweak: Gives brig physician nitrile gloves instead of latex gloves
/:cl:


MajManatee (approved):
every other doctor has them

Moltijoe:
> every other doctor has them

exactly

PR #17022, opened 2022-12-09
Rebalances Ethereal "eating"

Moltijoe:
Something must have happened when ethereal discharge was added
most things charge for so little that you need to constantly eat in order to recharge

So i multiplied all sources of charging by 10
Removes tedium and should make it easier to 

They also started completely full, meaning even a single point of charge started making them overfull

if it ends up being too easy to maintain charge, just tweak the ETHEREAL_CHARGE_FACTOR define

:cl:  
tweak: Ethereal fill up hunger faster
tweak: Ethereal no longer start completely full
/:cl:


Bibby0110 (changes requested):
just make a define for multiplying this in the future

JamieD1 (approved):

PR #16221, opened 2022-10-22
Remove vent clog (scrubber overflow) events from rotation

ynot01:
# Document the changes in your pull request

The performance impact of these scrubbers are wholly unsanitary and I believe they are the reason for a recent crash (43458)

Nanotrasen has since improved their scrubber cleaning systems which somehow get chemicals in them despite being atmos ports

# Changelog

:cl:  
rscdel: Removed vent clog events from rotation
/:cl:

PR #17044, opened 2022-12-10
Replaces all foam events with smoke events

warface1234455:
Foam affects server performance a lot, using it for stationwide can cause a lot of time dialations. Smoke is less impact on server performance, plus smoke doesnt break lighting anymore
Closes #16221
Replaces all foam events with smoke events

# Wiki document
Replaces all foam events with smoke events

# Changelog
🆑
tweak: Replaces all foam events with smoke events
/🆑

### DIFF
--- a/code/game/objects/effects/effect_system/effects_foam.dm
+++ b/code/game/objects/effects/effect_system/effects_foam.dm
@@ -107,6 +107,7 @@
 
 
 /obj/effect/particle_effect/foam/proc/kill_foam()
+	set waitfor = FALSE
 	STOP_PROCESSING(SSfastprocess, src)
 	switch(metal)
 		if(ALUMINUM_FOAM)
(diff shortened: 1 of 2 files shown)


JamieD1:
Why not add both

ShoziOffical:
> Why not add both

~~uhhh.. sure~~

Simplehorror:
did you copy my changelog?

ShoziOffical:
> did you copy my changelog?

no

Simplehorror:
editlog says otherwise

ShoziOffical:
> editlog says otherwise

maybe

ShoziOffical:
> Why not add both

~~wait, do you mean smoke and foam at the same time or two seperated events for smoke and foam~~

ShoziOffical:
actually foam is hella laggy, should remove it

ShoziOffical:
Note: this is not a bugfix but a tweak

Simplehorror:
the whole point of foam is so that you slip on it, this isn't going to fix the lag either

iamthesilly:
this was a thing for a while, it caused way more lag and broke the lighting system

ShoziOffical:
> this was a thing for a while, it caused way more lag and broke the lighting system

Well it did break lighting before now it isn't anymore, it's not really laggy as the foam if you run it on your private server

ShoziOffical:
i slipped my hand

PR #17051, opened 2022-12-12
Small hemophilia fix

Mqiib:

Stops TRAIT_BLOODY_MESS from increasing bleeding rate while you're dead.
I'll make a bigger change to this that doesn't count as a bug fix once the feature freeze is done that makes it REASONABLE and not "paper cuts turn you into a raisin"

# Changelog

:cl:  

bugfix: Hemophilia/heparin no longer making bleeding worse when you're already dead

/:cl:

### DIFF
--- a/code/datums/wounds/slash.dm
+++ b/code/datums/wounds/slash.dm
@@ -102,7 +102,7 @@
 
 	blood_flow = min(blood_flow, WOUND_SLASH_MAX_BLOODFLOW)
 
-	if(HAS_TRAIT(victim, TRAIT_BLOODY_MESS) && (victim.stat !+ DEAD))
+	if(HAS_TRAIT(victim, TRAIT_BLOODY_MESS) && (victim.stat != DEAD))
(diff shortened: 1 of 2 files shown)


code/datums/wounds/slash.dm
```diff
@@ -102,7 +102,7 @@
 
 	blood_flow = min(blood_flow, WOUND_SLASH_MAX_BLOODFLOW)
 
-	if(HAS_TRAIT(victim, TRAIT_BLOODY_MESS))
+	if(HAS_TRAIT(victim, TRAIT_BLOODY_MESS) && (victim.stat !+ DEAD))
```
Moltijoe: gotta let go of shift
it's != not !+

Moltijoe (dismissed):

code/datums/wounds/slash.dm
Mqiib: oops

Bibby0110 (approved):

PR #17052, opened 2022-12-12
Fix unnable to talk on departmental channel on bounce radio

warface1234455:
Technically you still can but you gotta reequip/remove the other radio for it, which is annoying. Now it will just use either  radio that has the channel access. Also during tcomms blackout, if you a bounce radio in your pocket or belt it will use that one instead of keep using the headset one.

Also clean up codes

**In case of you wondering if this is tested, yes i did and it is now working better than before no more wonky radio shit**

Fix unnable to talk on departmental channel on bounce radio

# Changelog

:cl:  

bugfix: Fix unnable to talk on departmental channel on bounce radio
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/say.dm
+++ b/code/modules/mob/living/carbon/human/say.dm
@@ -65,42 +65,17 @@
 		return .
 
 	if(message_mods[MODE_HEADSET])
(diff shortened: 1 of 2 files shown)


ToasterBiome (approved):
Wow, say code sure is shitcode.

PR #17068, opened 2022-12-14
Disables Rotatium if a fix is not found for exploiters

JamieD1:
Disables Rotatium if a fix is not found for exploiters

:cl:  
rscdel: Rotatium
/:cl:

PR #17089, opened 2022-12-15
Revert "Disables Rotatium if a fix is not found for exploiters"

Simplehorror:
Reverts yogstation13/Yogstation#17068

i have a fix for this


Bibby0110:
Pr the fix

PR #17094, opened 2022-12-16
buffs 10mm incind and emp

Skrem7:

they cost the same as the glorious 10mm ap so why do they suck as much as they do (I'm to blame)

adds 5 damage to both and makes it so the emp bullet actually heavy emps its tile and does a light emp at a range of 1

does this absolutely delete ipcs yes but so do emp grenades or the emp implant

# Wiki Documentation

10mm incind damage to 25 from 20
10mm EMP damage to 25 from 20, also maybe note its emp strength

# Changelog

:cl:  
tweak: 10mm incind and EMP now both do 25 damage instead of 20
tweak: 10mm EMP heavy EMPs the tile it hits and does a light EMP in the tile around rather than just a light EMP on its own tile so it actually counters sec
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets/pistol.dm
+++ b/code/modules/projectiles/projectile/bullets/pistol.dm
@@ -48,13 +48,13 @@
 
 /obj/item/projectile/bullet/incendiary/c10mm
 	name = "10mm incendiary bullet"
-	damage = 20
+	damage = 25
 	fire_stacks = 2
 
 /obj/item/projectile/bullet/c10mm/emp
 	name = "10mm EMP bullet"
-	damage = 20
+	damage = 25
 
 /obj/item/projectile/bullet/c10mm/emp/on_hit(atom/target, blocked = FALSE)
 	..()
-	empulse(target, -1, 0) //Only EMPs whatever's hit
+	empulse(target, 0, 1) //Heavy EMP on target, light EMP in tiles around


Simplehorror:
make it so EMP bullets emp along the path they came, i'd use it like that

UltraBigRat:
Does incend still do the thing hwere it makes a flame trail, cause if it does it should probably be adjusted so you cant  walk into the fire of your own bullets and can actually use them, or remove it altogether because the bullet will stop anyways when you hit the first target. (if this isnt the case ignore this, i just remember the one time i used incend rounds they made a flame trail that did nothing other then light me on fire.

Also very based to now have a good ranged option against validhunter borgs.

MajManatee (approved):
i hate IPCs and preternis.

Skrem7:
> Does incend still do the thing hwere it makes a flame trail, cause if it does it should probably be adjusted so you cant walk into the fire of your own bullets and can actually use them, or remove it altogether because the bullet will stop anyways when you hit the first target. (if this isnt the case ignore this, i just remember the one time i used incend rounds they made a flame trail that did nothing other then light me on fire.
> 
> Also very based to now have a good ranged option against validhunter borgs.

Incend rounds inherently leave hotspot turf in their wake, it's a property of all of them minus .38 Hotshot, since it just straight ignites on hit.

PR #17095, opened 2022-12-16
Switching to semi-automatic sets accuracy to default gun accuracy

Skrem7:

Super super niche code but I remember someone mentioning it being nice once. Realistically on most weapons losing out on the lethality versus your accuracy doesn't matter, but I've got a little something in the works which will make this more of a viable thing. Plus it's just... generally quality and semi auto kinda sux

Specifically coded so that it should be agreeable with attachments and laser sights, tested it on the tommy gun, seemed to work fine

# Changelog

:cl:  
tweak: If a gun is more inaccurate than most other guns AND it fires in bursts, setting it to semi-automatic sets its inaccuracy to normal (for any gun)
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -41,7 +41,8 @@
 	var/firing_burst = 0				//Prevent the weapon from firing again while already firing
 	var/semicd = 0						//cooldown handler
 	var/weapon_weight = WEAPON_LIGHT
-	var/spread = 5						//Spread induced by the gun itself.
+	var/spread = 5						//Spread induced by the gun itself. SEE LINE BELOW.
+	var/default_spread = 5				//MUST be equal to the value above; used to calculate adjustments for if semi-auto is used on a burst weapon.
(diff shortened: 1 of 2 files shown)


ynot01:
im pretty sure spread_difference is not required but regardless you need to go and set it for every gun that doesnt have the default 5 spread

Skrem7:
> im pretty sure spread_difference is not required but regardless you need to go and set it for every gun that doesnt have the default 5 spread

So how it works is that it SHOULD value a value of zero for any weapon that has normal accuracy, even on its burst mode

If a weapon does have spread, it takes the difference of the gun's normal spread and the default spread, so when it makes the check to set it to semi auto, it sets it to default spread. This means that a laser sight should still work properly.

Basically the issue is that if you check the current spread and someone was using a laser sight on a burst weapon with inherent inaccuracy (such as the L6) setting it to semi-auto wouldn't set its spread to -1 as intended if the variable were not created initially

MajManatee (approved):
finally i can one deag again

PR #17111, opened 2022-12-18
Emagged defibs knockdown instead of paralyze

Mqiib:
It's an instant 10 second stun that you can't really mitigate like you can batons. A knockdown still makes it useful, just not quite the instant win it was before. This is specifically the disarm intent effect.

# Wiki Documentation

Emagged/combat defibs knockdown instead of stun.

# Changelog

:cl:  

tweak: Emagged/combat defibs knockdown instead of stun

/:cl:

### DIFF
--- a/code/game/objects/items/defib.dm
+++ b/code/game/objects/items/defib.dm
@@ -473,7 +473,7 @@
 	M.visible_message(span_danger("[user] has touched [M] with [src]!"), \
 			span_userdanger("[user] has touched [M] with [src]!"))
 	M.adjustStaminaLoss(50)
-	M.Paralyze(100)
+	M.Knockdown(100)
 	M.updatehealth() //forces health update before next life tick //isn't this done by adjustStaminaLoss anyway?
 	playsound(src,  'sound/machines/defib_zap.ogg', 50, 1, -1)
 	M.emote("gasp")


Rdrazga:
I would recommend making this not affect syndie combat defibs since nukie things should remain powerful

Rdrazga:
Thought 2 on this PR: how about make it do stamina damage instead so it can still stun but can be countered at least a little easier.

TheRyeGuyWhoWillNowDie:
if you're going to do this then re-add EMPs hacking defibs

UltraBigRat:
> Thought 2 on this PR: how about make it do stamina damage instead so it can still stun but can be countered at least a little easier.

this would probably be better, making them basically just makeshift batongs that are a bit bulkier.

CMOisLing:
i ded this man

Mqiib:
> i ded this man

This was a request from @cowbot92 iirc

CMOisLing:
> > i ded this man
> 
> This was a request from @cowbot92 iirc

i ded that man

MajManatee (approved):
sounds like a skill issue but ok

Moltijoe (commented):
It should really just be shorter duration rather than a knockdown, it's literally shocking someone with ridiculous amounts of power, that's bound to cause more than just a knockdown.
Make it consume more batter if need be.

SomeguyManperson:
> Thought 2 on this PR: how about make it do stamina damage instead so it can still stun but can be countered at least a little easier.

it already  does

PR #17128, opened 2022-12-20
Prevent people that are unable to speak from opening the Say prompt

ynot01:

This is mostly just to prevent people from showing chat bubbles while unable to speak (like living statues which aren't meant to show signs of living) but also lets people (especially new players) know that they're muted so they don't waste time typing

# Changelog

:cl:  
tweak: Things that are unable to speak or are muted no longer get access to the Say prompt
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/statue.dm
+++ b/code/modules/mob/living/simple_animal/hostile/statue.dm
@@ -137,8 +137,11 @@
 
 // Cannot talk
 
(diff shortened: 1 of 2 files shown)


Sniblet:
this prevents them from doing asterisk emotes

UltraBigRat:
yeah since emotes and *me are a thing maybe send the you are muted message when they open the menu, instead of disabling it alltogther (maybe with no speech bubble)

PR #17129, opened 2022-12-20
Only shows the "You may call the shuttle whenever you find it appropriate." part of the end of shift announcement if the shuttle is idle(not called)

ThatLing:

It always bothered me, this improves RP because CC are not idiots, they are going to know that the shuttle is called and theyre not going to tell you to feel free to call it if its already called

# Changelog

:cl:  
tweak: CentCom won't tell you to call the shuttle if it's already been called
/:cl:


alexkar598:
Since when are you a thing, its been a while

MajManatee (approved):
yeah that never made sense

PR #17110, opened 2022-12-18
Targeted black powder nerf

Mqiib:
Increases the required explosion temperature to 737 K (464 C, the actual ignition temperature) and increases the metabolization rate of black powder to 1/tick. This should effectively neuter semi-instant-death black powder bomb syringes without dunking on black powder grenades or monkey bombs. Black powder syringes should be a lot easier to survive now if you act quickly.

Yes I realize this is made during the feature freeze, I can wait. It's just a number change.

# Wiki Documentation
See changelog

# Changelog

:cl:  
tweak: Black powder metabolization rate increased (0.05 -> 1) 
tweak: Black powder explosion temp required increased (474 -> 737)
/:cl:

PR #17132, opened 2022-12-21
Rebalances Black Powder Explosion

BurgerLUA:

- Increases blackpowder's strength divisor from 6 to 12.
- Increases blackpowder's modifier from 1 to 5.

# Wew

Ever since I've joined Yogs I keep seeing people complain about black powder. The whole issue of black powder seems more political than actual real-world debates considering there is like 4 people who play chemistry and actively exploit the strength of the chemical. I haven't seen the chemical in action until this round where I unintentionally ran towards a deathmix foam grenade I was maybe within 3 tiles of it, briefly exposed, and it killed me in a reasonable amount of time for what a deathmix grenade should do. No issue there.

However, I somehow exploded like maybe 6-7 seconds after the explosion. The explosion was strong enough to put a hole in the station and expose a room to space, as well as destroying all my equipment. Despite never encountering black powder before, I absolutely and immediately knew what it was based on the fact that it was delayed (bodytemp needed to rise) and it was stupidly strong.

/tg/ is a good place to get balance ideas from because they're a playground of sweats. People push the envelop of the game and given they have like 4 servers with very decent pop there is a lot of rounds to get samples from, on top of their extremely active dev team and strict standards for balance. I am porting their values for gun powder from their because of that.

# Changelog

:cl:  BurgerBB
tweak: Rebalances Black Powder so it is actually balanced.
/:cl:


Mqiib:
#17110 should mostly solve this issue of it nuking you when you ingest it because it'll already significantly reduce the power of any explosions it makes by way of reducing the amount of it in you (metabolizes faster) AND increase the time it takes to reach the point at which it reacts because it requires a higher temp. This pr reducing the explosion power of black powder ALSO significantly impacts its usefulness in grenades as actual explosives as well as dunking on cherry bombs (which are pretty rare but still) which is not desirable.

BurgerLUA:
> #17110 should mostly solve this issue of it nuking you when you ingest it because it'll already significantly reduce the power of any explosions it makes by way of reducing the amount of it in you (metabolizes faster) AND increase the time it takes to reach the point at which it reacts because it requires a higher temp. This pr reducing the explosion power of black powder ALSO significantly impacts its usefulness in grenades as actual explosives as well as dunking on cherry bombs (which are pretty rare but still) which is not desirable.

Nothing wrong with both. On /tg/, the metabolism rate of black powder is actually higher on there too.

ynot01:
i like the temperature change better, maybe higher metab too, but this will just make it so no one makes black powder anymore for any reason

TheRyeGuyWhoWillNowDie:
> i like the temperature change better, maybe higher metab too, but this will just make it so no one makes black powder anymore for any reason

One or the other works fine, makes meth and black powder distinct.
Meth for large craters, black powder for funny syringes or quick and small explosions.

JamieD1 (approved):

PR #17141, opened 2022-12-21
Slugs now have damage falloff

Skrem7:

To accompany the pellet armor nerf (yes these are just universal shotgun nerfs!)

Each shotgun slug now loses 2 damage per tile traveled, and/or 1.5 stamina damage per tile traveled. Probably not the most statistically significant, but it's a start. Incind slug + dragon's breath remain unaffected because they're pathed under incendiary and I don't want to mess with them

To make pellets slightly more appealing as well their damage falloff has been very very slightly reduced

Reorganizes the main shotgun projectile file to make more sense and to actually categorize slugs versus pellets

# Wiki Documentation

Slug properties should probably be changed on Guide to Combat

# Changelog

:cl:  
tweak: Shotgun slugs now lose 2 damage and 1.5 stamina damage per tile traveled
tweak: Pellet damage falloff is minorly less bad
/:cl:


MajManatee:
the beginning of the end of shotgun combat...

PR #17153, opened 2022-12-23
Removes frenzy dash throw

Mqiib:
This is NOT the ABILITY jump, this is the short range (from user) click-to-teleport-and-slap that frenzy holoparas ALSO get that for some reason flings you A QUARTER OF THE Z-LEVEL on a 3-second timer. No massive flinging for free, lets leave that to the actual ability.

# Changelog

:cl:  

rscdel: Non-spell frenzy dash no longer sends you into orbit

/:cl:


Moltijoe (changes requested):
Man really wants to sneak sidewinder in

UltraBigRat:
could  swap the fling with something else or weaken it instead of removing it altogether? cause this combined with the cooldown puts the ability in an even odder place then before where its just a random ranged attack that with frenzys high speed might aswell just be used for beating up people behind windows

Mqiib:
> could swap the fling with something else or weaken it instead of removing it altogether? cause this combined with the cooldown puts the ability in an even odder place then before where its just a random ranged attack that with frenzys high speed might aswell just be used for beating up people behind windows

These PRs affect two different things, they just HAPPEN to both be almost identical and from the same holoparasite ability.

PR #17158, opened 2022-12-24
Moves .38 speed loaders to armory lathe, autolathe .38 now only prints single bullets when hacked

Skrem7:

The .38 is exclusively lethal now rather than less-than-lethal, why is it in the normal lathe?

Same goes for it being simply available at autolathes. To mirror the .357 and to prevent detectives from hoarding .38 loaders without warden/hos clearance, you can only print single bullets now which are much harder to carry (OR you could do it when you need to refill bullets)

I realize this makes that one bounty much harder to do however bounties are stupid and the "security" bounty should theoretically require input from security personally I would just tear up the whole system but it's w/e, there are much better and easier ways to make money

# Wiki Documentation

Significant changes to the Detective page should be made regarding the acquisition and usage of ammunition for the revolver

# Changelog

:cl:  
tweak: Protolathe .38 speed loader design moved to armory lathe
tweak: .38 printed at an autolathe can only be done while hacked, also only prints single bullets
/:cl:

### DIFF
--- a/code/modules/research/designs/autolathe_designs.dm
+++ b/code/modules/research/designs/autolathe_designs.dm
@@ -628,14 +628,6 @@
 	build_path = /obj/item/ammo_casing/shotgun/rubbershot
 	category = list("initial", "Security")
 
-/datum/design/c38
-	name = "Speed Loader (.38)"
-	id = "c38"
-	build_type = AUTOLATHE
-	materials = list(/datum/material/iron = 20000)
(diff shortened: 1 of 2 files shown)


UltraBigRat:
could put like one or two extra roundstart speedloaders in the detective closet maybe? much better then the box full of speedloaders they normally make but better then only having two and thats it, also makes it still somewhat viable for antags to raid.

JamieD1 (approved):

PR #17177, opened 2022-12-26
Removes Pyroclastic Anomaly as its just pain for our atmos system

JamieD1:
Removes Pyroclastic Anomaly as its just pain for our atmos system
:cl:  
rscdel: Removes Pyroclastic Anomaly as its just pain for our atmos system
/:cl:

### DIFF
--- a/code/modules/events/anomaly_pyro.dm
+++ b/code/modules/events/anomaly_pyro.dm
@@ -2,8 +2,8 @@
 	name = "Anomaly: Pyroclastic"
 	typepath = /datum/round_event/anomaly/anomaly_pyro
 	
-	max_occurrences = 5
-	weight = 20
+	max_occurrences = 0
+	weight = 0
 
 /datum/round_event/anomaly/anomaly_pyro
 	startWhen = 3


Moja10:
a widdle tiny plasmafire is a pain on our system? I'm surprised our atmos works at all then, if a single anomaly is such a pain. don't like this change at all

JamieD1:
> a widdle tiny plasmafire is a pain on our system? I'm surprised our atmos works at all then, if a single anomaly is such a pain. don't like this change at all

It makes every room nearby unusable for 20 years. Because of the shitty way our temp system works

Bibby0110 (approved):

Marmio64:
We're down to like 2 anomalies now...

JamieD1:
it will be added back when auxmos is merged

ShoziOffical:
> > a widdle tiny plasmafire is a pain on our system? I'm surprised our atmos works at all then, if a single anomaly is such a pain. don't like this change at all
> 
> It makes every room nearby unusable for 20 years. Because of the shitty way our temp system works

It can easily be fixed by just spacing the room (technically, the room would already be spaced by the fire) and refill it with a new fresh air

PR #17187, opened 2022-12-26
Fix bloodsucker blood examines (masquerade and such)

ynot01:

Wrapping my head around this seemingly nonsensical code was a chore

The way it used to work is it would basically just always return the bloodsucker's true blood. The code was very misdirecting returning `BLOODSUCKER_HIDE_BLOOD` when that meant to show true blood and when it wasn't returning `BLOODSUCKER_HIDE_BLOOD` it was returning normal blood examines based on your true blood. So basically it did nothing useful and was just confusing.

The way it works now is as follows:
Show true blood volume on examine when returning `BLOODSUCKER_SHOW_BLOOD` (not bloodsucker or not masqueraded)
Show 100% blood volume on examine when returning `BLOODSUCKER_HIDE_BLOOD` (When masqueraded and not low blood or frenzied)
Returning anything else will be treated as a custom message, for when we port Nosferatu or want to do anything else with this regarding custom blood examines

# Changelog

:cl:  
bugfix: Fix bloodsucker blood examines with masquerade
/:cl:

### DIFF
--- a/code/__DEFINES/misc.dm
+++ b/code/__DEFINES/misc.dm
@@ -492,6 +492,7 @@ GLOBAL_LIST_INIT(pda_styles, list(MONO, VT, ORBITRON, SHARE))
 
 /// Whether we have succesfully hidden out blood level
 #define BLOODSUCKER_HIDE_BLOOD "hide_blood_volume"
+#define BLOODSUCKER_SHOW_BLOOD "show_blood_volume"
 /// 1 tile down
 #define ui_blood_display "WEST:6,CENTER-1:0"
 /// 2 tiles down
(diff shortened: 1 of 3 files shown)


Moltijoe:
> Wrapping my head around this seemingly nonsensical code was a chore

you weren't kidding, i'm trying to figure this out myself

Moltijoe (commented):
Definitely janky code in the previous version
- Extra commented out return line
- Multiple duplicate text lines in different locations
- Returning complicated strings
- Seemingly wanting to be a binary value but also not

Only thing of note is that `else if(bloodDesc != BLOODSUCKER_HIDE_BLOOD)` doesn't seem like it will ever trigger. Just want to make sure this won't be a problem

ynot01:
> Only thing of note is that `else if(bloodDesc != BLOODSUCKER_HIDE_BLOOD)` doesn't seem like it will ever trigger. Just want to make sure this won't be a problem

> Returning anything else will be treated as a custom message, for when we port Nosferatu or want to do anything else with this regarding custom blood examines

Moltijoe (approved):
looks good

PR #17193, opened 2022-12-27
Makes hydrogen burns less stupid

Mqiib:
Important note: tritium and hydrogen burns are still really stupid and could be made less stupid but I don't want to mess with the underlying mechanics of the COOL gas too much.

Basically, this makes hydrogen burns mimic tritium burns except with significantly lower temperature because it now also makes a metric buttload of water vapor while burning. In testing, temperatures seemed to reach around 12,000-13,000 degrees with a 50/50 mix (compared to tritium's >100,000). Toxins mains can now DESPAIR because their funny super high temperature burn stupidity has been made REASONABLE! Use exotic gases to make funny bombs not the most abundant element in the universe! Why are we even researching plasma if hydrogen is better?!

# Wiki Documentation

Uhhhh something about hydrogen being just worse non-radioactive tritium (or tritium being spicier hydrogen i guess)

# Changelog

:cl:  

tweak: Hydrogen is now just shittier tritium as it should be
bugfix: hydrogen/tritium burns will now actually consume oxygen instead of fabricating it out of thin air (haha get it because it's a gas)

/:cl:


Ethan4303:
If its shittier tritium, then why do i need to make tritium to get hydrogen??????? These are the real questions

JamieD1 (approved):

PR #17200, opened 2022-12-28
Adds colour-coded nodes for gas mixers

ThatLing:
Adds colours to gas mixer inputs so it's easier to track which node is which

# Spriting
image

# Changelog

:cl:  
rscadd: Gas mixers nodes are now colour-coded
/:cl:


warface1234455 (changes requested):
Should keep it as "Node 1" and "Node 2" because when analyzing mixer it gives gas details on Node 1 and 2. So this won't confuse players which one is node 1 or 2

ThatLing:
> Should keep it as "Node 1" and "Node 2" because when analyzing mixer it gives gas details on Node 1 and 2. So this won't confuse players which one is node 1 or 2

I changed it back, it also didn't make much sense with the T-shaped mixers since there's no "main" input node.

JamieD1 (approved):

PR #17205, opened 2022-12-28
Buffs the Atmos Backpack Firefighter Tank so it isn't noob bait

ToasterBiome:

The backpack firefighter tank sucks ass, and is terrible noob bait on how both unruly and expensive (in terms of using it) is. I buffed it so that potentially once in a blue moon Atmospheric Technicians could _potentially maybe_ fix a gassed up room with it.

**Volume from 200 to 300**: This will give you an increase on resin balls from 2 to 6. This is a massive buff but I believe a needed one. You cannot reasonably fight anything with 200 units of water, and I would have increased it curther if I didn't also reduce the cost for the resin launcher. This does mean you get 50% more extinguishing power, but for an atmos item specifically made for fighting fires (and doesn't even anything more special than water) I think we can handle this.

**Cost for resin launcher from 100 units to 50 units**: You shoot it twice (and accidentally smack your foam that's already there with the second one) and you're done. Go home. You're expected to drag around a tank while ditching your backpack to also carry around your holofan, advanced extinguishers, normal tools, and potentially the fireaxe to get into areas you don't have access to? I can see why atmos techs just space rooms or don't even come out of incinerator.

**Cooldown on resin launcher from 10 seconds to 5 seconds**: Makes it so you can actually use it. The cooldown is CBT and prevents things that are time sensitive (um, plasmafire) from being dealt with.

**Cooldown on resin foamer from 10 seconds to 5 seconds**: This will make it faster to actually fight 900000K rooms instead of waiting around for you to actually be able to do anything.

> But baiomu, some greyshit can steal this and grief with it!

I'll ban them lol

# Spriting

Also changes the sprites from non-sensical icons to what the different modes stand for. (extinguisher was the only one that made sense):
image

# Wiki Documentation

backpack firefighter tank isn't even on the wiki. Here's the image to upload:
backpack_firefighter_tank

Add this to here: https://wiki.yogstation.net/wiki/Atmospherics_items
```
{{anchor|Backpack Firefighter Tank}}
{{Item
 |bgcolor1 = #ccccee
 |bgcolor2 = #ddddff
 |name = Backpack Firefighter Tank
 |image = backpack_firefighter_tank.png
 |foundin = [[Atmospherics]]
 |usedfor = Fighting fires, cleaning up plasmafloods
 |strategy = A backpack water tank that can switch between firefighting modes as the situation calls for it.
 |description = A backpack water tank used for firefighting. It can switch between three different modes:
* '''Extinguisher''': Functions like a normal extinguisher.
* '''Resin Launcher''': Shoots balls of resin that expand into firefighting resin foam. Good for clearing entire rooms. Costs 50 units of water to launch on a 5 second cooldown.
* '''Resin Foamer''': Makes individual tiles of firefighting resin foam where you press. Good for getting into tight spaces and navigating more complex room setups. The tank starts with 5 charges and regenerates 1 every 5 seconds.
Holds 300 units of water.
 }}
```

# Changelog

:cl:  
tweak: Backpack Firefighter Tank volume increase from 200u -> 300u, resin launcher cost from 100u to 50u, resin launcher and foamer cooldown from 10 seconds to 5 seconds.
/:cl:


iamthesilly:
Would be nice if the resin was passable (with slowdown obviously) so people wouldn't ruin all your hard work by breaking it to get through, and so you don't have to undo all your work because you missed a spot

JamieD1 (approved):
Epic

PR #17221, opened 2022-12-29
Makes AI Download 50% Faster

JamieD1:
Makes AI Download 50% Faster

At the moment it takes ages to download the AI if you are a traitor for example and have the steal AI objective. Which forces you to either have to make your own or find a different way of doing this. Failing this, maybe we could have an item which traitors can get which makes it silent or perhaps doubles the speed?

:cl:  
tweak: Makes AI Download 50% Faster
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/ai/ai.dm
+++ b/code/modules/mob/living/silicon/ai/ai.dm
@@ -110,7 +110,7 @@
 	//Can we (simple) examine humans?
 	var/canExamineHumans = FALSE
 	//Reduces/Increases download speed by this modifier
-	var/downloadSpeedModifier = 1
+	var/downloadSpeedModifier = 1.5
 
 	var/login_warned_temp = FALSE
 


ToasterBiome:
Why?

Hopekz:
> Why?

Ye kind of sucks he never puts his reasonings in a PR

Bibby0110:
Missing rationale

JamieD1:
Added Rationale

Bibby0110 (changes requested):
Change it in the defines

ThatLing:
I think it would be better to add an item for a few TC that downloads the AI instant (or a few seconds)

Bibby0110:
> I think it would be better to add an item for a few TC that downloads the AI instant (or a few seconds)

Sounds like a horrible idea.
At best the item should reduce the time it takes by 50% or so, max

ThatLing:
> > I think it would be better to add an item for a few TC that downloads the AI instant (or a few seconds)
> 
> Sounds like a horrible idea. At best the item should reduce the time it takes by 50% or so, max

Who's gonna spend TC to get a measly 50% download speed increase? If you're gonna spend TC, it should be worth it

Bibby0110:
> > > I think it would be better to add an item for a few TC that downloads the AI instant (or a few seconds)
> > 
> > 
> > Sounds like a horrible idea. At best the item should reduce the time it takes by 50% or so, max
> 
> Who's gonna spend TC to get a measly 50% download speed increase? If you're gonna spend TC, it should be worth it

It should be worth it, but not that worth it

Moltijoe:
Created my own that fixes the problem bibby has

PR #17224, opened 2022-12-29
more cool stuff for the engineer and medic uplink erts

TheRyeGuyWhoWillNowDie:
engineer and medic roles for uplink ert are pretty barebones and don't really scale to fit the strength of the ert, so i added some new stuff for some of the stronger erts for them

Engineers can buy a reskinned and renamed CE toolbelt, which has all of the advanced tools instead of the boring ones in the 1wc belt

Engineers can buy X4 bricks in addition to C4 bricks

Engineers can buy ARCDs (RCD but it can work at range) if they really want to

Medics can buy a combat injector containing 150u of restorative nanites for fast healing

Medics can buy a one-use injector containing reagents that will revive and fully heal whoever they injected, but it only works on corpses.

# Spriting
resurrector nanite serum
image

# Changelog

:cl:  
rscadd: more cool stuff for engineer and medic uplink erts
/:cl:


Marmio64:
image
make the nanite serum look like this instead for a reference. Maybe change it from 'nanite' to 'mech' too.

TheRyeGuyWhoWillNowDie:
> image make the nanite serum look like this instead for a reference. Maybe change it from 'nanite' to 'mech' too.

it makes more sense as nanites because mechanites are just rimworld nanites and this isn't rimworld

Moltijoe (approved):
Was wondering why the defines were undefined in the original PR, glad it was quickly changed to this.
Appreciate getting more items for medical ert, since it's not common that the medical ert actually knows medical. So them getting simple items to use with strong effects should greatly increase their effectiveness.

PR #17226, opened 2022-12-30
Revert "what if we put maint access to cloning on box that would be weird 😳"

ToasterBiome:
Reverts yogstation13/Yogstation#17212

Not 24 hours + I put the wrong access + this is a bad idea


Aquizit:
Honest question because i was pretty neutral about this suggestion in the first place but why did it turn out to be a bad idea?

00ze-cyclone:
Just put the correct access and keep it pls 🥺

ToasterBiome:
https://github.com/yogstation13/Yogstation/pull/17262

PR #17113, opened 2022-12-18
Forces Drone to Stay on the Station

JamieD1:
Forces Drone to Stay on the Station

:cl:  
rscadd: Forces Drone to Stay on the Station
/:cl:

PR #17227, opened 2022-12-30
Revert "Forces Drone to Stay on the Station"

Cartlord:
Reverts yogstation13/Yogstation#17113 because it was selfmerged with missing rationale and overwhelmingly negative reception after Jamie removed the "Missing/Inadequate Rationale" and "Closing after 24h" tags that Bibby placed

:cl:
tweak: Drones can be off-station again
/:cl:


JamieD1:
The rationale is that drones do not need to be on the shuttle and everytime they are on the shuttle they just create issues. They are supposed to fix the station, every single time there are drones on the shuttle there are active things on the station they could be doing, you want to get on the shuttle, be a human or borg.

PR #17229, opened 2022-12-30
Changes Station to Area because people dont know that station implies everything except the shuttle but oh well ill change it just for those people who need everything spelled out for them and need leeway. Its literally just so you dont go on shuttle.

JamieD1:
Station to Area in Drone laws

:cl:  
tweak: Station to Area in Drone laws
/:cl:


Sniblet:
Look how much less people complain when they understand why something is happening
But how are we supposed to know what an "assigned area" is

PR #17234, opened 2022-12-30
Backports several SSair improvements (more startup speed & random lavaland atmos)

ThatLing:

The SSair performance improvement might not work correctly with Monstermos. I tested this locally and didn't notice any obvious issues, however I would still recommend to test-merge first. If stuff breaks, I might remove the performance improvements and just keep the random atmos.

Notable features:
- Faster startup time
- Randomized lavaland atmos

List of PRs:
- https://github.com/tgstation/tgstation/pull/44427
- https://github.com/tgstation/tgstation/pull/69697
- https://github.com/tgstation/tgstation/pull/69831

# Changelog

:cl: LemonInTheDark & ninjanomnom
rscadd: Lavaland atmosphere is now randomized every round
tweak: Atmos startup speed increased
/:cl:


Mqiib:
Lavaland atmos isn't processing properly. There are tiles that become 0 kPa and suck all the air around them into a black hole and the planetary atmos isnt fixing the holes itself. My guess is it has to do specifically with planetary atmos and not atmos in general because everything else works fine.

Running fixair on the lavaland z-level changes all the lavaland tiles to 0 kPa but doing it in a radius works properly.

Also the max pressure needs to be like 30-40 kPa for kinetic accelerators to work.

Oh and there's a metric fuckton of megafauna for some reason.

neeshacark:
Please add a way for admins to force the lavaland atmos to anything so if we encounter issues like mqiib is describing we can fix them.  One of the PRs ported claimed they did that but mqiib and I can't find anything like that in the PR or ingame.

ThatLing:
@Mqiib Thanks for the heads-up, it seems to be an issue with moving air across adjacent tiles, since only the pressure around the mining station scrubbers is gone.

Will look into fixair

Megafauna spawning fixed in https://github.com/yogstation13/Yogstation/pull/17257

ThatLing:
This will break (with) auxmos/katmos port, which is higher priority than this, I might revive this at a later data.

PR #17237, opened 2022-12-30
tend wounds is removed??? lmao

TheRyeGuyWhoWillNowDie:
After a lot of shit flinging in admin channels, Jamie has decided that he wants to make cloning someone actually something that happens, and make losing your memories through dying more frequent. Note that Patch Burns and Patch Bruises, the surgeries that consume gauze, are not gone, (at least for now, at Jamie's bequest). Note that synthflesh is also not gone. Tend wounds is just a boring cureall that consumes no resources and removes all damage very fast.

image
image
I'm not about to pass up the opportunity to get a Jamie sticker.

# Wiki Documentation

tend wounds is gone lol

# Changelog

:cl:  
rscdel: tend wounds is gone lol
/:cl:


Hislyna:
This sounds so stupid

bruhlookatthisdood:
make burn and brute patches heal damage when dead nerd

Mqiib:
lol lmao funny pr

JamieD1:
im scared. I will let someone see if its sane before I put my sticker on it.

TheRyeGuyWhoWillNowDie:
I GOT BAITED!!! I DON'T EVEN GET A STICKER!!

Moltijoe:
But, why? all this does is make it harder to bring people back into the game, and makes everyone use the cloner.

Hislyna:
Clone Station and Cryo Tube Station LETS GOOOOOOOO

Hopekz:
This is removing a semi-large part of current medical treatment and replacing it with nothing.
I think rather than this feature just being erased; it should instead be modified to fit.

There's so many things that can be done with this vs deletion.
You can simply change coming back from the dead make you lose memory regardless of method; you can change the difficulty of this procedure. You can add more steps to it (outside of surgery).
You can make it only work on living people.

Just anything is better than a full on removal imo.
This PR is the worst outcome for a feature like this.

TheRyeGuyWhoWillNowDie:
> Clone Station and Cryo Tube Station LETS GOOOOOOOO

no different than hemostat station, you just lose your memory more because you get cloned (the whole point)

KillerOrcaCora:
You Gus has been baited into making yet another low effort bad PR, though this is not entirely your fault but still your PR

JamieD1:
> You Gus has been baited into making yet another low effort bad PR, though this is not entirely your fault but still your PR

Cloning should be more viable and tend wounds less catch all and op. Gus is not wrong.

KillerOrcaCora:
You both are smoking, also wait for the Medbay mains to have their say before we say the TG player who tried to make miner pens not work on Z level not work for no reason be thought right

Xantam-13:
Lol

ReddicusDragon:
this is SO FUCKING STUPID

Aquizit:
Most of us medbay mains like surgery. It's actually something active that we get to do that's multiple steps and showcases knowledge of the health/wound system and procedure. While tend brute/burn will still exist, and is used but a fair amount of us when a body has only one type of damage that needs extensive tending, usually if we're doing the mix, we need something like synthflesh anyway for a dehusking.

Secondly, the players don't like slow. Cloning is slow. If there's a line cloning is slower. Post cloning generally gets a cryo visit for the cellular damage if there hasn't been a cloner upgraded. So a just alive character once again gets control and awareness removed as they're put to sleep for the cellular damage.

Medbay won't like it. Players won't like it.

Edit: I suppose here I'm more arguing against the philosophy of "more cloning" rather than this specific PR. Oops?

LazennG:
average gus pr

TheRyeGuyWhoWillNowDie:
>Aquizit

this doesnt remove any surgery except the one where you click once with a hemostat and tab out

Aquizit:
Yes, but so... you can't just repeat the step in treat/patch as far as I'm aware. You need to do the whole surgery every time. So yeah, that will definitely push more people into the cloning tube. And as such the rest of my argument still stands (I hope).

Cloning has its place so I don't necessarily want to see it removed. But I think more would prefer an expansion of memory loss rather than more cloning.

Sniblet:
I for one am all for replacing slightly stimulating play with click-dragging people into a box, hitting a button and leaving

TheRyeGuyWhoWillNowDie:
> I for one am all for replacing slightly stimulating play with click-dragging people into a box, hitting a button and leaving

no different than click once with a hemostat and wait

Moltijoe:
Gutting surgery isnt the solution to the problem of people only ever using surgery. Memory retention being as important as it is, is the most likely cause. Removing tend wounds will only lead to people spending more time dead rather than playing and enjoying themselves.

PR #17238, opened 2022-12-31
Resting now only will avoid projectiles from adjacent shooters

Mqiib:
I'm getting tired of people resting and just easily dodging projectiles all the time mid-combat. If you're within MELEE range sure but a tile away should be enough to shoot you.

# Wiki Documentation

If it's mentioned, it only applies to adjacent people now.

# Changelog

:cl:  

tweak: Resting only allows you to avoid projectiles from adjacent shooters

/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -598,7 +598,7 @@
 	else
 		var/mob/living/L = target
 		if(!direct_target)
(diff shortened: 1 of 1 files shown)


KillerOrcaCora:
This reeks of skill issue

Marmio64:
feel like resting should still give you a chance to duck below bullets at least

TheRyeGuyWhoWillNowDie:
yes because getting low to the ground and making yourself hard to hit to avoid getting shot is completely unrealistic

PR #17241, opened 2022-12-31
Tritium/Hydrogen combustion tweaks and bugfixes

SapphicOverload:

1) Low-energy tritium combustion now correctly consumes oxygen instead of more tritium

2) Fixes high-energy tritium/hydrogen combustion consuming 18x as much oxygen as intended

3) Hydrogen now burns with half as much energy as tritium instead of 1/10 and burns twice as many moles per tick

# Changelog

:cl:  
bugfix: fixes low-energy tritium combustion not consuming oxygen
bugfix: fixed high-energy trit/h2 combustion consuming 18x more oxygen that it's supposed to
tweak: hydrogen burns with half the energy of tritium while consuming twice as many moles
/:cl:

### DIFF
--- a/code/__DEFINES/reactions.dm
+++ b/code/__DEFINES/reactions.dm
@@ -7,7 +7,7 @@
 #define PLASMA_MINIMUM_OXYGEN_NEEDED		2
 #define PLASMA_MINIMUM_OXYGEN_PLASMA_RATIO	30
 #define FIRE_CARBON_ENERGY_RELEASED			100000	//Amount of heat released per mole of burnt carbon into the tile
-#define FIRE_HYDROGEN_ENERGY_RELEASED		560000  //Yogs -- Amount of heat released per mole of burnt hydrogen and/or tritium(hydrogen isotope). Doubled due to Consv. of Mass fix causing shittier fires
+#define FIRE_HYDROGEN_ENERGY_RELEASED		2800000  //Yogs -- Amount of heat released per mole of burnt hydrogen and/or tritium(hydrogen isotope). Increased significantly due to a bugfix leading to much lower burn temperatures.
 #define FIRE_PLASMA_ENERGY_RELEASED			3000000	//Amount of heat released per mole of burnt plasma into the tile
 //General assmos defines.
 #define WATER_VAPOR_FREEZE					200
@@ -31,7 +31,7 @@
 #define MINIMUM_TRIT_OXYBURN_ENERGY 		2000000	//This is calculated to help prevent singlecap bombs(Overpowered tritium/oxygen single tank bombs)
 //hydrogen reaction
 #define HYDROGEN_BURN_OXY_FACTOR			100
-#define HYDROGEN_BURN_H2_FACTOR				10
+#define HYDROGEN_BURN_H2_FACTOR				5		//Burns faster and with half the energy of tritium
 #define MINIMUM_H2_OXYBURN_ENERGY 			2000000	//This is calculated to help prevent singlecap bombs(Overpowered hydrogen/oxygen single tank bombs)
 //ammonia reaction
 #define AMMONIA_FORMATION_FACTOR			250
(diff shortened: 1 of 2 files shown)


Mqiib:
So you... multiplied hydrogen burn energy by 5 and then also multiplied regular hydrogen burn energy output by another 5? I'm not seeing much that would offset that absolutely insane increase in energy here.

Did you test it?

iamthesilly:
yes i tested it, it makes hydrogen burns about half the temperature of tritium burns which remain the same and hydrogen toxins bombs about half the size of tritium toxins bombs which are also the same as before

also the lower H2 factor decreases the burn energy while increasing how much of it burns each reaction tick, so with this change it has a total net energy gain per tick the same as tritium but twice as much water vapor reducing the net temperature gain significantly and uses up twice as much fuel and oxygen

Mqiib:
I am of the opinion that it shouldn't be even close to tritium in terms of effectiveness. The PR I made was INTENTIONALLY reducing the power of hydrogen, it just also happened to fix (and add, oops) a few bugs with it. I don't want hydrogen to be as powerful as it was, it burning at like 10000K ish from my quick and dirty testing is fine.

iamthesilly:
it's really not that much harder to get tritium than it is to get hydrogen, and any time you have access to one to you almost almost certainly have access to the other

Mqiib:
Hydrogen is a hell of a lot easier to make large volumes of than tritium without access to atmos, and even with access to atmos you can make orders of magnitude more hydrogen than you can tritium. I still don't think hydrogen SHOULD be immediately burning at 50K+ degrees because it's invisible and extremely dangerous that way. The level it got stuck to with my last nerf/bugfix pr is where it really should be, not an insanely hot burning gas that is significantly better than the exotic stuff the station is researching.

iamthesilly:
not quite sure what you're talking about, i can make tritium in toxins with less effort than hydrogen and at pretty much the same rate if the electrolyzer hasn't been upgraded

Mqiib:
> not quite sure what you're talking about, i can make tritium in toxins with less effort than hydrogen and at pretty much the same rate if the electrolyzer hasn't been upgraded

When is the last time you've seen a proper tritium flood from toxins? I've seen hydrogen spew out and incinerate science far more often than I've seen pure tritium do that. 99% of cases, hydrogen is easily produced in massive quantities and I still hold that it just being _better_ than plasma for generating heat and burning stuff is stupid.

iamthesilly:
In toxins, tritium is made in the burn chamber where you just stop the burn after you've made enough (not more than a few hundred moles 99% of the time and most if not all of it ends up in tanks for bombs) whereas hydrogen is made in an electrolyzer chamber where you just add the water vapor and it just keeps building up over time. Since only a few hundred moles of hydrogen get used and the chamber builds up into the thousands there's almost always enough for a devastating flood if/when it breaks, which can happen easily.

Also because hydrogen is far less common than plasma I don't think it's that big of a deal for it to be able to reach hotter maximum burn temperatures.

Mqiib:
High-power hydrogen makes doing toxins braindead easy and THEMATICALLY the most abundant material in the universe shouldn't so easily outperform the exotic and rare stuff. In-game yeah there's a ton of plasma because the station starts with it and atmos has infinite plasma but thats just because the station was mapped with a bunch of plasma, not hydrogen. On top of that, hydrogen you just keep making basically forever with a teeny tiny plasma burn making h2o and the electrolyzer pumping out a bajillion moles of the stuff.

iamthesilly:
> High-power hydrogen makes doing toxins braindead easy and THEMATICALLY the most abundant material in the universe shouldn't so easily outperform the exotic and rare stuff. 

This PR keeps hydrogen at HALF the power of tritium making it far less useful for both bombs and fire. Also to me the exotic part of plasma burns isn't their raw energy output from the plasma itself but rather its ability to break down into another fuel (tritium) when properly oxygenated allowing it to reach much higher temperatures than hydrogen. And plasma is used for a LOT more than just as a fuel.

Also due to water vapor's high heat capacity and the temperature of tritium burns, the amount of cooling it takes to make it not immediately set your hydrogen chamber on fire makes hydrogen production from a plasma burn less effective than just making tritium, which has to exist in the first place for it to burn which means you might as well be collecting that instead. And you should, because as I mentioned before tritium is still TWICE as strong, takes fewer steps to make if you aren't just using H2O from a canister, and has a fraction of the heat capacity of water vapor making it easier to collect and cool from a plasma burn.

Mqiib:
Atmospherics operates more on a logarithmic scale most of the time than a linear scale. A full order of magnitude stronger is a difference. Twice as strong when you're going from 50k to 100k is not a significant enough difference.

iamthesilly:
50k vs 100k is the difference between whether or not you're able to make nitryl, stimulum, zauker, metal hydrogen, and a few other things

also 50k is hot enough to burn through the hull but still survivable with an atmos hardsuit/firesuit

MajManatee (approved):
funny gasses. Mib killed my hydrogen bomb

JamieD1 (approved):

PR #17244, opened 2022-12-31
Makes syndi prisoner bag not bad & broken

ynot01:
Changes tested locally
## Part 1: Made it not horrible to use

Two changes were made here
- The bag now only takes 4 seconds to zip instead of 20
  - Holy shit. 20 fucking seconds. If you have someone that refuses to move for 20 seconds, they're either dead or already asleep or AFK
- The bag now only costs 2TC from 7TC
  - Literal 7TC handcuffs & mute that takes longer to apply and is very obvious. The main attraction is from the fact that they will never get up (you can forget about them as long as they won't get found) and that you can step over them while taking them places. This does not make it worth anywhere near 7TC. I'm split between 2 and 3 TC, but I'd like to place this in 2TC heaven and see where people take it.

## Part 2: Fixed a bug where you die

Yeah it would just kill the guy

# Changelog

:cl:  
bugfix: Syndicate prisoner bag no longer kills its victims
tweak: Syndicate prisoner bag costs 2TC
tweak: Syndicate prisoner bag zips up much quicker
/:cl:

### DIFF
--- a/code/datums/diseases/advance/advance.dm
+++ b/code/datums/diseases/advance/advance.dm
@@ -296,21 +296,12 @@
 		var/res = clamp(properties["resistance"] - (symptoms.len / 2), 1, advance_cures.len)
 		if(res == oldres)
 			return
-		
-		if(prob(res * 7)) // Double cure
-			var/list/not_used = advance_cures[res].Copy()
-			cures = list(pick_n_take(not_used), pick_n_take(not_used))
-			// Get the cure name from the cure_id
-			var/datum/reagent/D1 = GLOB.chemical_reagents_list[cures[1]]
-			var/datum/reagent/D2 = GLOB.chemical_reagents_list[cures[2]]
-			cure_text = "[D1.name] and [D2.name]"
-		else // Single cure
-			cures = list(pick(advance_cures[res]))
-			// Get the cure name from the cure_id
-			var/datum/reagent/D = GLOB.chemical_reagents_list[cures[1]]
-			cure_text = D.name
-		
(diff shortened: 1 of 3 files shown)


Moltijoe:
should _probably_ atomize the bugfix from the other changes

ynot01:
> should _probably_ atomize the bugfix from the other changes

are there changes you disagree with?

Moltijoe:
> are there changes you disagree with?

No, incentivising keeping people alive rather than killing them is nice, though it is sorta dangerous in that it can round remove more easily than killing. A paramedic will look for people that are either dead or disappear from sensors, but wont ever think to look for someone that is "missing" but still on sensors.

Aside from that, antag related balance is beyond my expertise.

Edit: unless that 8 minute break out time means that if the bag is left alone, the victim can break out.

ynot01:
> unless that 8 minute break out time means that if the bag is left alone, the victim can break out.

the 8 minute timer is only relevant if the victim is not asleep which would only happen if they had internals on or didn't breathe or were otherwise unable to sleep

Moltijoe:
> the 8 minute timer is only relevant if the victim is not asleep which would only happen if they had internals on or didn't breathe or were otherwise unable to sleep

ah i see now, it's n2o they're forced to breathe... yeah that's pretty rough, makes it easy to round remove someone without raising too much suspicion. I still prefer it over death round removing though since it still gives them a chance to be found

Moltijoe:
I think this is a fair buff, it still costs 2 tc, and simply allows for other options of disposal

MajManatee (approved):
i like it

PR #17216, opened 2022-12-29
Rips out the old keybind system and ports a better version from /tg/

ThatLing:
# Document the changes in your pull request

At the start of the last rebase, I made a shitty keybind system that allowed you to rebind actions. 4 years later, it's about time to upgrade it.

This PR completely removes the old system and ports the new version used by /tg/. It adds more actions and a better way to assign keys.

I tested each keybind to ensure it works correctly, however there might be some issues still. Since it's such a big change, I decided not to convert existing keybinds to the new system. As such, people will need to re-assign their keys.

List of ports below, full credits to the original authors
- https://github.com/tgstation/tgstation/pull/47003
- https://github.com/tgstation/tgstation/pull/47626
- https://github.com/tgstation/tgstation/pull/47670
- https://github.com/tgstation/tgstation/pull/47750
- https://github.com/tgstation/tgstation/pull/47804
- https://github.com/tgstation/tgstation/pull/47868
- https://github.com/tgstation/tgstation/pull/48316
- https://github.com/tgstation/tgstation/pull/49116
- https://github.com/tgstation/tgstation/pull/52475
- https://github.com/tgstation/tgstation/pull/53709
- https://github.com/tgstation/tgstation/pull/53771
- https://github.com/tgstation/tgstation/pull/53773
- https://github.com/tgstation/tgstation/pull/53925
- https://github.com/tgstation/tgstation/pull/54035
- https://github.com/tgstation/tgstation/pull/47750

# Wiki Documentation

No clue if there is documentation on the wiki, but this system should work largely similar

# Changelog

:cl:  
rscadd: Added a new keybind system, your old keybinds have been reset
rscadd: You can now bind a key to a specific emote
/:cl:

PR #17249, opened 2022-12-31
MC changes on correct SSinput

fira:

previous MC PR incorrectly made changes to SSinput Init Stage / Initialize in its disabled files instead of the ones in `yogstation/`

this fixes that by letting it properly early-init and suppressing the warning


ThatLing:
As a heads up, this file will be deleted by https://github.com/yogstation13/Yogstation/pull/17216

JamieD1 (approved):

PR #17254, opened 2023-01-01
Moves insuls and belt back onto clockies when they spawn in

Skrem7:

#16384 was a good PR for reducing meta (bad) in adding people to the manifest however it seems moving the bulky belt into the bag where it isn't supposed to fit (!) means it just isn't spawning right now. Check belt code, and the belt still exists there, so it's just not properly manifesting in the bag.

This PR moves the gloves and belt back on to the cultist, anyone with a brain can just take them off if they desire, belt contents can be poured into the backpack. If random search and seeing an assistant with a bag is a problem, that's meta in of itself plus this highlights the problem of having sec level go to blue roundstart (screws over heretics too, also if a funky nuke core box spawns in your backpack then GG if you're caught in the hallway as soon as you get it)

# Changelog

:cl:  
tweak: Insuls and belt back in the bag
/:cl:

### DIFF
--- a/code/game/gamemodes/clock_cult/clock_cult.dm
+++ b/code/game/gamemodes/clock_cult/clock_cult.dm
@@ -279,8 +279,10 @@ Credit where due:
 	shoes = /obj/item/clothing/shoes/sneakers/black
 	back = /obj/item/storage/backpack
 	ears = /obj/item/radio/headset
+	gloves = /obj/item/clothing/gloves/color/yellow //Take them off if you want
+	belt = /obj/item/storage/belt/utility/servant //Take this off and pour it into a toolbox if you want
 	backpack_contents = list(/obj/item/storage/box/engineer = 1, \
-	/obj/item/clockwork/replica_fabricator = 1, /obj/item/stack/tile/brass/fifty = 1, /obj/item/paper/servant_primer = 1, /obj/item/clothing/gloves/color/yellow = 1, /obj/item/storage/belt/utility/servant = 1)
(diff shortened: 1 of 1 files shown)


Hopekz:
Why not just make the belt they spawn in be a child and have the child be able to be stored in the backpack?
So they'd basically have unique belts that allows them to not be meta'd at the start

Skrem7:
> Why not just make the belt they spawn in be a child and have the child be able to be stored in the backpack? So they'd basically have unique belts that allows them to not be meta'd at the start

A belt that can be stored in the backpack can be meta'd if examined.

Hopekz:
> A belt that can be stored in the backpack can be meta'd if examined.

Wdym people don't see the inside of your backpack when they look at you which is what the point of #16384 was.

Also how does your change support your argument; you get found out when you get searched with your clockie items anyway whether they're in a belt or not

Skrem7:
> Also how does your change support your argument; you get found out when you get searched with your clockie items anyway

right now the belt isn't spawning in the backpack mainly is why this pr was made

Hopekz:
> right now the belt isn't spawning in the backpack mainly is why this pr was made

Ah; well if this is a bug fix label it as such.
The way it is packaged up right now it seems like you want to go back to pre- #16384


As far as bug fixes go reverting a change to fix it isn't good imo.
The way to fix this would be to make the belt work with the backpack to make it spawn.

MajManatee (approved):
I hate clockies but they kinda needed this

PR #17258, opened 2023-01-01
grand core buff to let its moves just affect any dead thing 

LazennG:

changes grand core to let you make bloodmen out of any corpse because finding dead people specifically is annoying and i can't remember what reasoning i had for having it affect solely people specifically if any

:cl:  
tweak: tweaked grand core to be slightly more usable
/:cl:


Moltijoe (approved):
Weird a lavaland item would specifically require human corpses

PR #17259, opened 2023-01-01
Allow demos to be disabled via config

ThatLing:

Basically ports https://github.com/tgstation/tgstation/pull/50615

While replays have proven to be beloved tool by our admins, it adds no value when developing. Having to wait an extra ±20 seconds everytime I want to start YogStation is pretty annoying and it starts using a lot of disk space after a while.

This PR simply disables demos by default to not bother developers while it can stay enabled on the server. I don't know if the server config is synced with GitHub (it wasn't back in my day), if it is I can change it to be enabled by default.

### DIFF
--- a/config/config.txt
+++ b/config/config.txt
@@ -125,6 +125,9 @@ LOG_VIRUS
 ## log cloning actions
 LOG_CLONING
 
+## Enable the demo subsystem
+# DEMOS_ENABLED
+
 ##Log camera pictures - Must have picture logging enabled
 # PICTURE_LOGGING_CAMERA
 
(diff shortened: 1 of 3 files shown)


config/config.txt
```diff
@@ -125,6 +125,9 @@ LOG_VIRUS
 ## log cloning actions
 LOG_CLONING
 
+## Enable the demo subsystem
+# DEMOS_ENABLED
```
JamieD1:
```suggestion
DEMOS_ENABLED
```

JamieD1 (approved):

PR #17262, opened 2023-01-02
Flips Boxstation Cloning and Genetics

ToasterBiome:

image

image

The hallway inclusion that leads to xenobio did one extra thing: it made genetics a fortress, lacking maint access.

Every other map has maint access to genetics/cloning, but box was lacking now. 

Flips them so genetics is in a better position to be accessed from maintenance (and more susceptible to sabotage)
Also buffs cloning by making it closer to treatment and having it's "EMP cloner in maint" be the little area above it instead.

# Wiki Documentation

Genetics lab and cloning on BoxStation are flipped.

# Changelog

:cl:  
mapping: Flips genetics and cloning on BoxStation
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -123,13 +123,6 @@
 /obj/structure/grille,
 /turf/open/space/basic,
 /area/space/nearstation)
-"aaW" = (
-/obj/item/reagent_containers/food/snacks/grown/banana,
-/obj/structure/window/reinforced{
-	dir = 4
-	},
-/turf/open/floor/grass,
-/area/medical/genetics)
 "aaZ" = (
 /turf/closed/wall/r_wall,
 /area/ai_monitored/security/armory)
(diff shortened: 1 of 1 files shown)


patpol4 (changes requested):
1. double wire 
image
2. All the doors except the one leading to rnd corridor have a lot of different accesses that will only cause confusion when someone wants to give access to this area, keep just the proper ones for the proper door. So cloning for cloning and genetics for genetics: https://github.com/yogstation13/Yogstation/blob/master/code/__DEFINES/access.dm
3. as an addition, maybe add an rd c_tag for the genetics camera?
4. considering maybe removing this thindow, it looks weird next to a wall: 
image



Aside from these I think this is fine

ToasterBiome:
> 1. double wire
> 2. All the doors except the one leading to rnd corridor have a lot of different accesses that will only cause confusion when someone wants to give access to this area, keep just the proper ones for the proper door. So cloning for cloning and genetics for genetics: https://github.com/yogstation13/Yogstation/blob/master/code/__DEFINES/access.dm
> 3. as an addition, maybe add an rd c_tag for the genetics camera?
> 4. considering maybe removing this thindow, it looks weird next to a wall:
> 
> Aside from these I think this is fine

1. Fixed
2. These match the same doors as were on there previously. Most importantly, the 3 different accesses on the door between them is for the weird situation where someone has access to gene but not cloning. Working as intended
3. Genetics camera already has a c_tag, unless you mean adding it the RD network, then I will do that.
4. 
image
Flipped this so that the glass can be outside.

Hopekz:
@wejengin2 🐐

patpol4 (approved):
i guess

PR #17263, opened 2023-01-02
substantially reduces foam lag

TheRyeGuyWhoWillNowDie:
hmm today i will create HUNDREDS of effects and then have them on FAST PROCESS and have them FUCK WITH ATMOS every single tick <- WORDS OF A FUCKING PSYCHOPATH

from testing this makes the atmos SS go from like a permanent 300ms generated after stationwide foam to about 25ms, as well as obviously eliminating a ton of work from the fast processing SS

# Changelog

:cl:  
tweak: foam lags the server less
/:cl:


JamieD1:
again for the tenth time, lets fix the issue. which we are doing.

JamieD1:
Which this actually does.

PR #17264, opened 2023-01-02
Removes the Curator's space suit

LoliconSlayer:

Title self explanatory
The curator is there to curate the library, they shouldn't be going on space walks every round and leaving their precious books unattended.

As usual, I have no idea what I'm doing and this code is untested

# Spriting

N/A

# Wiki Documentation

I'm pretty sure we don't mention that the curator gets a voidsuit anywhere on the wiki

# Changelog

:cl:  
rscdel: Curators now have to curate their library instead of being Space Indiana Jones
/:cl:

### DIFF
--- a/code/game/objects/items/miscellaneous.dm
+++ b/code/game/objects/items/miscellaneous.dm
@@ -98,15 +98,6 @@
 	new /obj/item/clothing/shoes/workboots/mining(src)
 	new /obj/item/melee/curator_whip(src)
 
-/obj/item/storage/box/hero/astronaut
-	name = "First Man on the Moon - 1960's."
-
-/obj/item/storage/box/hero/astronaut/PopulateContents()
-	new /obj/item/clothing/suit/space/nasavoid(src)
-	new /obj/item/clothing/head/helmet/space/nasavoid(src)
-	new /obj/item/tank/internals/oxygen(src)
-	new /obj/item/gps(src)
-
 /obj/item/storage/box/hero/scottish
 	name = "Braveheart, the Scottish rebel - 1300's."
 


00ze-cyclone:
Whip and claymore have no purpose beside validhunting anyway. Kill the beacon entirely since now it’s just a weapon dispenser.

BurgerLUA:
Wouldn't the curator just ask the HoP for EVA access? I mean I agree with the suit removal because them getting a suit is a little dumb, but I don't think it will solve the problem.

Also you'll have to remove any references to ``/obj/item/storage/box/hero/astronaut`` to code for the PR to actually work as it is likely still being references in whatever gives it (don't know what gives it. vendor? beacon?)

00ze-cyclone:
> Wouldn't the curator just ask the HoP for EVA access? I mean I agree with the suit removal because them getting a suit is a little dumb, but I don't think it will solve the problem.
> 
> Also you'll have to remove any references to `/obj/item/storage/box/hero/astronaut` to code for the PR to actually work as it is likely still being references in whatever gives it (don't know what gives it. vendor? beacon?)

Someone interacting with the HoP to give paperwork guy something to do is a net plus imo.

ToasterBiome:
What's the problem with curators leaving? I see no good reason to remove the suit

Moltijoe (dismissed):
almost 100% certain this doesn't work, you've deleted the item, but it's still in the map files

Moltijoe (commented):
nevermind, went to actually look at the code, works sorta like the null rod where it is just one item that can select a number of others. still, you should test it before making the pr.

LoliconSlayer:
> Wouldn't the curator just ask the HoP for EVA access? I mean I agree with the suit removal because them getting a suit is a little dumb, but I don't think it will solve the problem.
> 
> Also you'll have to remove any references to `/obj/item/storage/box/hero/astronaut` to code for the PR to actually work as it is likely still being references in whatever gives it (don't know what gives it. vendor? beacon?)

This PR is removing it from the beacon, it's just a box with specified content, `/obj/item/storage/box/hero/astronaut` doesn't exist anywhere else. See #3063


> almost 100% certain this doesn't work, you've deleted the item, but it's still in the map files

As far as I can tell, this just removes the option from the beacon and the box doesn't actually exist in any map files, not that I could find anyway. Likewise, see #3063


> What's the problem with curators leaving? I see no good reason to remove the suit

You might consider it "inadequate" rationale, but I don't think people should be treating curator as the dedicated space ruins job. I especially find it problematic if they're just fucking off and leaving the station behind on gamemodes that require contribution from the entire crew such as Nuke Ops and cults.

Based on my observation in the Discord, a non-insignificant number of people share the same or similar opinion. I'd appreciate it if you actually gave this PR a chance instead of immediately marking it to close after 24 hours.

Hopekz:
Tbh if we're going this direction I'd like to see a dedicated job for space exploration at some point.

DukeBobRoss:
> Wouldn't the curator just ask the HoP for EVA access? I mean I agree with the suit removal because them getting a suit is a little dumb, but I don't think it will solve the problem.

good make them interact with somebody for five seconds before they vanish

Maia-J:
> good make them interact with somebody for five seconds before they vanish

Ironically, most curator mains want to interact with someone....
Unfortunately, it's always the alien queen.

KillerOrcaCora:
Curator going to space and finding relics for their display always seemed like a intended bit, I expect to see a actual role for this implemented before taking this lazy route here that does really nothing to stop the thing mentioned except annoying curator players for no real reason.

JamieD1:
Eh, Curator is already boring as shit, no point is cucking them harder.

PR #17276, opened 2023-01-03
Moves Syringe Gun to Armoury and Rapid to Armoury Lathe

JamieD1:
This will be controversial, the issue is the syringe gun is 99% of the time used by valid hunters, so i'm moving to to sec who are allowed to valid hunt.

:cl:  
tweak: Moves Syringe gun to Armoury and Rapid to Armoury Lathe
/:cl:

### DIFF
--- a/_maps/map_files/AsteroidStation/AsteroidStation.dmm
+++ b/_maps/map_files/AsteroidStation/AsteroidStation.dmm
@@ -2741,6 +2741,15 @@
 /obj/structure/flora/rock/jungle,
 /turf/open/floor/grass,
 /area/maintenance/starboard/fore)
+"avZ" = (
+/obj/structure/rack,
+/obj/effect/turf_decal/bot_red,
+/obj/item/circuitboard/machine/techfab/department/security,
+/obj/item/reagent_containers/syringe/lethal/execution,
+/obj/item/gun/syringe,
+/obj/item/gun/syringe,
+/turf/open/floor/plasteel/dark,
+/area/ai_monitored/security/armory)
 "awb" = (
 /turf/closed/wall,
 /area/hallway/primary/port)
@@ -18924,22 +18933,6 @@
 	},
 /turf/open/floor/wood,
 /area/hallway/secondary/service)
-"fxj" = (
-/obj/machinery/newscaster{
-	pixel_x = -3;
-	pixel_y = 32
-	},
-/obj/machinery/light_switch{
-	pixel_x = 10;
-	pixel_y = 25
-	},
-/obj/effect/turf_decal/trimline/blue/filled/line{
-	dir = 5
-	},
-/obj/structure/closet/secure_closet/medical3,
-/obj/item/gun/syringe,
-/turf/open/floor/plasteel/white,
-/area/medical/storage)
 "fxq" = (
 /obj/effect/turf_decal/tile/yellow{
 	dir = 8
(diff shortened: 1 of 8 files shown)


ToasterBiome:
This is certainly interesting to say the least

iamthesilly:
could maybe put one in the CMO locker too

Moltijoe:
probably for the best rather than nerfing them into the ground

Skrem7:
> could maybe put one in the CMO locker too

CMO has their flash and baton AND instant hypospray, why do they need the syringe gun?

Mantlecrawler:
This gets a medical lizard sticker of approval

Moltijoe:
> CMO has their flash and baton AND instant hypospray, why do they need the syringe gun?

i see it less as giving cmo a syringe gun and more as restricting the doctor's syringe guns to only CMO.
that way assuming something happens, the cmo can always either use it or provide it to a trusted medbay member without needing access to the armoury.

also, adds another reason for people to break into the cmo office

00ze-cyclone:
> > CMO has their flash and baton AND instant hypospray, why do they need the syringe gun?
> 
> i see it less as giving cmo a syringe gun and more as restricting the doctor's syringe guns to only CMO. that way assuming something happens, the cmo can always either use it or provide it to a trusted medbay member without needing access to the armoury.
> 
> also, adds another reason for people to break into the cmo office

Yeah but it’ll still probably only be used for validhunting by non sec in that case and that’s what we want to stop

Moltijoe:
> Yeah but it’ll still probably only be used for validhunting by non sec in that case and that’s what we want to stop

if it's in the cmo's office, that means for it to be used in valid hunting,
either someone needs to break into the office. 
(if they're valid hunting that means they're not an antag and this could very easily be bwoink or even ban worthy)

or the CMO authorized it's used
(which any sane CMO would not allow unless the situation actually calls for it)

or the CMO themselves is using it
(Which, if they're valid hunting as a command role that isn't HoS, could end up as a command ban)

SomeguyManperson:
Just remove it entirely?

Notamaniac:
I think it's of very limited utility to sec, considering they have access to reagent shotgun darts, and it's in a rare blue moon that security and chemistry ever cooperate together. To the point that I would say remove it entirely- but then we're progressing down the route of removing every single dangerous thing from every department because it could be used to valid hunt.

I don't know if hugbox station is what we're going for but it's certainly setting a precedent for it.

Sniblet:
The only source of useful syringe gun chemicals immediately available to security is the lethal injection syringes
This is an excellent idea

Moja10:
I don't think anything needs to be changed about syringe guns at all, but if you are channeling aggressive Jamie autism and are hellbent on this, give it to the CMO in CMO so he can distribute it as needed, rather than this change. This makes literally no sense, why would literally ANYONE in security bother with a syringe gun when they have shotguns with slugs, buckshot, incendiaries. If those wouldn't satisfy them and they want to work with another department for meme results, why would they bother with a syringe gun over tech rounds in a shotgun with pulse beams, frag-12, more incendiaries, and meteor-shot. 

Also, imagining security validhunting with a syringe gun, the literal only application I can realistically think of is that they validhunt someone with the lethal syringes they get, which sounds pretty horrible, and doesn't improve the round in the slightest.

ToasterBiome (approved):
map is fine, this is not a code approval

PR #17279, opened 2023-01-03
lets preternis and ethereals be affected by trait_eat_more and trait_eat_less

Moltijoe:
Preternis use their own janky form of hunger, this means traits that speed up or slow down hunger don't affect them
Ethereals don't get affected by these either because they rely on a weird stomach variable

This means efficient metabolism doesn't work on either of them, and a gluttony demon doesn't get "hungry" faster

IPC _does_ get affected by these because they just use nutrition

Changing either of these to rely on nutrition instead would take significant rewrites of their code and would have little benefits to my knowledge as of this moment.

Not entirely sure if this is a bugfix or a tweak

:cl:  
bugfix: lets preternis and ethereals be affected by trait_eat_more and trait_eat_less
/:cl:

### DIFF
--- a/code/modules/surgery/organs/stomach.dm
+++ b/code/modules/surgery/organs/stomach.dm
@@ -169,7 +169,14 @@
 
 /obj/item/organ/stomach/ethereal/on_life()
 	..()
-	adjust_charge(-ETHEREAL_CHARGE_FACTOR)
+	var/chargemod = 1
+	if(HAS_TRAIT(owner, TRAIT_EAT_LESS))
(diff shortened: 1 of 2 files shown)


ReddicusDragon:
some day we're going to have to standardize how food works across species, making them rely on nutrition probably. That day is not today.

Moltijoe:
> some day we're going to have to standardize how food works across species, making them rely on nutrition probably. That day is not today.

maybe some time during the summer after college term is over

MajManatee (approved):
the first of many preternis nerfs im sure

PR #17317, opened 2023-01-04
what the guh??? mechs for uplinked erts???

TheRyeGuyWhoWillNowDie:
may or may not have been requested by a certain ynot01 (i cant tell you because that would be leaking)

adds mechs to the engineer so they actually have interesting stuff to do as an alternative to being bob the builder
the mechs have no equipment as it must be purchased separately
most mech weapons are 4wcs, except the eZ-13 which is 10
marauder is 12wc (making it accessible to blues, although barely and when combined with ammo and a weapon leaves you with no wc) and seraph is 30wc so its really only deathsquad accessible
all ammos are 1wc per ammo box

yes these are pretty cheap but they are also more or less useless against non xenomorphs because melted in 0.5 seconds with emps (uncounterable) also there's a huge gap between red ert and deathsquad which is pretty difficult to accommodate 

# Changelog

:cl:  
rscadd: mechs for uplink erts
/:cl:

### DIFF
--- a/code/game/mecha/combat/marauder.dm
+++ b/code/game/mecha/combat/marauder.dm
@@ -55,8 +55,12 @@
 	force = 50
 	max_equip = 5
 
+/obj/mecha/combat/marauder/seraph/unloaded
+
 /obj/mecha/combat/marauder/seraph/Initialize()
 	. = ..()
+	if(istype(src,/obj/mecha/combat/marauder/seraph/unloaded))
+		return
 	var/obj/item/mecha_parts/mecha_equipment/ME = new /obj/item/mecha_parts/mecha_equipment/weapon/energy/pulse(src)
 	ME.attach(src)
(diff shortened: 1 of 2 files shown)


KillerOrcaCora:
In a few more PRs, uplink ERT will just be nukies type power but NT-aligned.

TheRyeGuyWhoWillNowDie:
> In a few more PRs, uplink ERT will just be nukies type power but NT-aligned.

dont tell him about the 'Open mech bay' button when creating erts

UltraBigRat:
> > In a few more PRs, uplink ERT will just be nukies type power but NT-aligned.
> 
> dont tell him about the 'Open mech bay' button when creating erts

To be fair  that is never used as the mech bay is full of several death squad level mechs

MajManatee (approved):
awooga awooga

PR #17318, opened 2023-01-04
remove ion rifle emp protection

TheRyeGuyWhoWillNowDie:
adds counterplay for mechs (people will actually use the heavy ion cannon now) as well as punishing you for rushing in and emping yourself

# Changelog

:cl:  
tweak: ion rifle (and carbine) are no longer fully immune to all emps
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -12,9 +12,6 @@
 	flight_x_offset = 17
 	flight_y_offset = 9
 
-/obj/item/gun/energy/ionrifle/emp_act(severity)
-	return
-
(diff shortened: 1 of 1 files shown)


Moltijoe:
have you tested this?
any time i've tried using the ion rifle as ipc/pre-rework preternis i've emped myself by just walking forward at all while shooting

ThatLing:
> have you tested this? any time i've tried using the ion rifle as ipc/pre-rework preternis i've emped myself by just walking forward at all while shooting

feature for playing a snowflake race

Moltijoe:
> feature for playing a snowflake race

no, what i mean is that the gun will fuckin emp itself

PR #17322, opened 2023-01-05
Lava staff is no longer also a one-handed lava axe

Mqiib:
Stop hitting people with it, it's a magic staff not a melee weapon!
It has 8 force now so people have less incentive to use it as a SWORD

# Wiki Documentation

It doesn't have 25 force any more so it's not a very good melee weapon (except against plants)

# Changelog

:cl:  

tweak: Lava staff force reduced significantly (25 -> 8)

/:cl:

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -1189,7 +1189,7 @@ GLOBAL_LIST_EMPTY(bloodmen_list)
 	icon = 'icons/obj/guns/magic.dmi'
 	slot_flags = ITEM_SLOT_BACK
 	w_class = WEIGHT_CLASS_BULKY
-	force = 8
+	force = 12
 	damtype = BURN
 	resistance_flags = LAVA_PROOF | FIRE_PROOF | ACID_PROOF
 	hitsound = 'sound/weapons/sear.ogg'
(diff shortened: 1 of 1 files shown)


GraveHat:
I like watching people die.

UltraBigRat:
it needs a buff to its actual lava placing then, cause for a magic staff it kinda sucks. could maybe make it slower and more obvious where lava is being placed but allow more to be placed.

Moltijoe (approved):
way too much melee damage for something that can cover the station in lava tiles

ANAME593:
> way too much melee damage for something that can cover the station in lava tiles

i don't think you understand really. 1. it has a cooldown (a pretty long one) 2. its rare for miners  to utilize lava staffs (because if you were to block a hall with this you'd get in alot of trouble)
3. the ideal option would be reducing it from 25 damage (okay that's too much) to about 15 damage 8 removes its effectiveness as a melee weapon it's nothing but paper. 15 damage is not alot of damage and it does not have any AP

hopefully it can remain as a weak melee weapon but not a weapon incapable of doing any kind of damage

code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -1189,7 +1189,7 @@ GLOBAL_LIST_EMPTY(bloodmen_list)
 	icon = 'icons/obj/guns/magic.dmi'
 	slot_flags = ITEM_SLOT_BACK
 	w_class = WEIGHT_CLASS_BULKY
-	force = 25
+	force = 8
```
SomeguyManperson:
```suggestion
	force = 12
```
Toolbox

MajManatee:
goofy fr

PR #17340, opened 2023-01-05
Replaces chasm destruction spawn from tendrils with lava

Moltijoe:
Round removal is cringe

:cl:  
tweak: Necropolis tendrils now spawn lava instead of chasms
/:cl:


Moja10:
Those are optional, you are not required to shoot the tendril for gaymer loot. 

With that being said, I do understand not wanting to run into them if someone else opened one. In that case, I think it just needs a slight sprite tweak to be a little more clear where the edges are. They don't need to be entirely replaced with lava.

Moltijoe:
> With that being said, I do understand not wanting to run into them if someone else opened one. In that case, I think it just needs a slight sprite tweak to be a little more clear where the edges are. They don't need to be entirely replaced with lava.

lava is already pretty much an insta-kill unless you're prepared for it. 
It being replaced with lava also lets people that die and get turned into a legion by the tendril mobs have a chance to be saved rather than condemned to the pit.

Sniblet:
no i want miners who are new or lagging to lose their rounds irrevocably as punishment for not doing enough WASD inputs in time

neeshacark:
Do jaunters still save you from them with this change?

Moltijoe:
> Do jaunters still save you from them with this change?

Jaunters only do that for chasms

MajManatee:
Skill issue, just dont go into the big hole

JamieD1:
Cool, lets try it.
image

PR #17343, opened 2023-01-05
Plasmeme can no longer get husked

Moltijoe:
QoL for the poor plasmeme that dies and gets cloned and husked infinitely
Realistically, how does a skeleton even get husked?

:cl:  
tweak: Plasmeme can no longer get husked
/:cl:


00ze-cyclone:
Can we stop calling buffs QoL please?
Other than that seem like a decent idea.

Moltijoe:
> Can we stop calling buffs QoL please? Other than that seem like a decent idea.

it IS a quality of life, it lets dent head medical actually have a chance to fix them
Quality of life and buffs are not mutually exclusive

MajManatee (approved):
Skeletons dont have skin?? What is there to husk?

PR #17353, opened 2023-01-05
Plasmeme Martial Art Buff + Bugfix?

Moltijoe:
Not only does it rely on your taking your suit off and setting yourself on fire, it also has just regular self damage built in
Gives them heat resistance so they can be a firey ball of death without actually dying

I also somehow fixed the help verb bug?

:cl:  
tweak: Plasmaman martial art now has fire and heat immunity
bugfix: Plasmaman martial art help verb should now work(?)
/:cl:


MajManatee (approved):
i forgot this was a thing

PR #17347, opened 2023-01-05
Reinforces Armoury

JamieD1:
image

Way too easy to break into, fed up of randoms just going FUCK IT ARMOURY

Alternative to https://github.com/yogstation13/Yogstation/pull/17347

Ill update the other maps depending on which passes

:cl:  
rscadd: Reinforces Armoury
/:cl:

PR #17361, opened 2023-01-06
move vents and add windoors to armory interior

Marmio64:
#17347 

a personal idea, just adding two windoors and moving the vents so that it's slightly harder to get to the good stuff without being overly distracting

image

the windoors have armory access requirement. They shouldn't do much, especially if they come prepared with an ID or an emag, but could stop them long enough for warden to actually respond.

(i moved the vents so any ventcrawler people or whatever dont have an even easier time dealing with beepsky)

# Changelog

:cl:  
rscadd: Adds windoors to safeguard armory equipment better, and moves vents around them.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -1448,12 +1448,6 @@
 	},
 /turf/open/floor/plasteel/dark,
 /area/ai_monitored/security/armory)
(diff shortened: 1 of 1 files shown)


patpol4:
also a good idea

patpol4:
Partially incorporated into https://github.com/yogstation13/Yogstation/pull/17379

PR #17363, opened 2023-01-06
Adds a cap to stamina damage slowdown

SapphicOverload:

Adds a cap to stamina damage slowdown. The cap is based off of max health, so that any stamina damage beyond what would cause you go into crit will not further increase the slowdown. The reason for this cap is because with anti-stuns you can easily accumulate hundreds of stamina damage causing you to be practically immobilized, which kinda defeats the whole purpose of antistuns. So, I've added a limit to how much it can slow you that I believe to be reasonable.

NOTE: This cap will NOT affect anyone that isn't stun-immune, you'll go into stamina crit before the point that it actually starts to matter.

# Changelog

:cl:  
tweak: caps slowdown from stamina damage
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1409,7 +1409,7 @@ GLOBAL_LIST_EMPTY(mentor_races)
 			if(I.item_flags & SLOWS_WHILE_IN_HAND)
 				. += I.slowdown
 		if(!HAS_TRAIT(H, TRAIT_IGNOREDAMAGESLOWDOWN))
-			var/health_deficiency = max(H.maxHealth - H.health, H.staminaloss)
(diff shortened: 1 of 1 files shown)


Moltijoe:
That would explain why stun immune seemed to do nothing for the ipc martial art

iamthesilly:
i'm tempted to reduce stamina slowdown or remove it entirely for anyone who's stun-immune but idk if that will go over well

Moltijoe:
> i'm tempted to reduce stamina slowdown or remove it entirely for anyone who's stun-immune but idk if that will go over well

Probably better to just have another trait to reduce or remove it

Moltijoe (approved):
Weird that slow kept stacking beyond crit, really did make stun immune useless against stamina damage

PR #17364, opened 2023-01-06
Doctor/Paramedic/CMO spawn with nitrile gloves

Moltijoe:
They already use them, but there aren't enough to go around in maps, nor any replacements after a loss.

Also added nitrile gloves to the plasmeme outfits + brig phys and mining medic (i forgot that plasmemes use entirely different outfits)

:cl:  
rscadd: Doctor/Paramedic/CMO spawn with nitrile gloves
bugfix: Plasmamen BrigPhys and MiningMed properly get nitrile gloves
/:cl:


MajManatee (approved):
hand socks

PR #17376, opened 2023-01-06
Reduce the chance that an EMP'd leg implant will break your leg

Moltijoe:
Yeah, i probably should have put it into a calculator rather than just guesstimating a good number.
20% chance to be near permanently crippled when hit by an emp is ridiculous
i've reduced it to 5% chance, since often death is preferable over wounds with how bad they currently are.

:cl:  
tweak: Leg Implants are less likely to break your leg if EMP'd
/:cl:


Burger1998:
yay less implants being noob bait to get you deleted when your breathed on by an emp

MajManatee:
cripples are dumb

MajManatee (approved):

PR #17410, opened 2023-01-08
Fixes air shoes implant exploit using boost while hardcrit or dead

ynot01:

Can no longer use it in hardcrit
Can still use it in softcrit bc im nice

# Changelog

:cl:  
bugfix: Can no longer use air shoes jump boost in hardcrit or death
/:cl:

### DIFF
--- a/code/modules/surgery/organs/augment_legs.dm
+++ b/code/modules/surgery/organs/augment_legs.dm
@@ -214,7 +214,8 @@
 	holder = user
 
 /datum/action/innate/boost/IsAvailable()
-	if(COOLDOWN_FINISHED(src, dash_cooldown) && holder.stat < UNCONSCIOUS)
+	. = ..()
+	if(COOLDOWN_FINISHED(src, dash_cooldown))
 		return TRUE
 	else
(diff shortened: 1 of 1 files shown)


Moltijoe:
Hold up, isnt that what the ability check flags do?

Edit: Oh, probably because im overwriting the original isavailable proc.
In that case, the other abilities probably dont respect check flags either

Moltijoe:
> Edit: Oh, probably because im overwriting the original isavailable proc. 

Exactly what happened

code/modules/surgery/organs/augment_legs.dm
```diff
@@ -214,7 +214,7 @@
 	holder = user
 
 /datum/action/innate/boost/IsAvailable()
-	if(COOLDOWN_FINISHED(src, dash_cooldown))
```
Moltijoe, line 217: just add 
`. = ..()`
above this line instead

Moltijoe (changes requested):

Moltijoe (approved):

ynot01:
sir if you write . = ..() and then return a value after that it's going to return the value not ..()

PR #17416, opened 2023-01-09
Should fix Spacebar automatically signing you up for every alert under the sun

JamieD1:
Should fix Spacebar automatically signing you up for every alert under the sun

:cl:  
bugfix: Fixes spacebar making you accept everything
/:cl:


ynot01 (approved):
awful

PR #17425, opened 2023-01-09
Fixes PDA reply button not working until opening the program once

ynot01:

What: Condensed instances of finding the computer into a helper proc that just sets the `computer` variable so it can be used elsewhere (like the Topic proc)

Why: `computer` is normally not set until the program is opened once by a computer so it is considered NULL and `usr.shared_ui_interaction(computer)` fails

# Changelog

:cl:  
bugfix: fixed PDA reply button not working until opening the program once  
/:cl:


Moltijoe (approved):
tested and works

PR #17429, opened 2023-01-09
PR template and github issue template updates

ReddicusDragon:

Opening an issue on github now asks a user if the issue they are experiencing is occurring on a non-Yog hosted server or a local server instead of the actual server.

PR template now mentions changelogs shouldn't exist if there are non-player facing changes


JamieD1:
Why do we care if its on a non yog server

JamieD1 (approved):

PR #17204, opened 2022-12-28
Changeling adrenal chem now reduces stamina damage (1->30)

ynot01:
# Document the changes in your pull request

The hero reads a most unsettling passage

I did a similar thing in #8254 but missed this

Lingkt adrenals are now good for resisting disabler spam

# Changelog

:cl:  
tweak: Changeling adrenals now heal more stamina damage
/:cl:

PR #17458, opened 2023-01-12
Revert "Changeling adrenal chem now reduces stamina damage (1->30)"

SomeguyManperson:
Reverts yogstation13/Yogstation#17204

Adrenals already provides a massive stamina heal on being used, and with the chem buff this can just be used several times so having constant passive stamina regeneration isn't particularly necessary or limited


Bibby0110:
PR freeze

PR #17463, opened 2023-01-12
Stops Supermatter Shards being spawned in the bluespace locker..

JamieD1:
Fucking horrible round, this is cheese which should not be allowed to happen again.

:cl:  
bugfix: Stops Supermatter Shards being spawned in the bluespace locker..
/:cl:

### DIFF
--- a/code/game/objects/items/miscellaneous.dm
+++ b/code/game/objects/items/miscellaneous.dm
@@ -41,7 +41,9 @@
 	var/choice = input(M,"Which item would you like to order?","Select an Item") as null|anything in display_names
 	if(!choice || !M.canUseTopic(src, BE_CLOSE, FALSE, NO_TK))
 		return
-
+	if(is_centcom_level(src.z))
+		visible_message(M, span_notice("You are unable to call this here."))
+		return 
 	spawn_option(display_names[choice],M)
 	uses--
 	if(!uses)


Maia-J:
Borg does something else than plasmaflooding on do not allow organics to escape.
> This gets created

(Also ided)

JamieD1:
> Borg does something else than plasmaflooding on do not allow organics to escape.
> 
> > This gets created
> 
> (Also ided)

Those objectives should not be given to a TRAITOR ai, that will be in a new pr.

iamthesilly:
> Those objectives should not be given to a TRAITOR ai, that will be in a new pr.

then make that PR instead

Moltijoe:
Hold up. What happened for this PR to be made?

Moja10:
ided?

JamieD1:
> > Those objectives should not be given to a TRAITOR ai, that will be in a new pr.
> 
> then make that PR instead

Both

UltraBigRat:
with how difficult this would be to pull off needing both finding the bluespace locker and  having access to cargo and opening the sm shard crate, i really dont see the need to remove an interesting tactic like this other then boohoo ided.

RG4ORDR:
> Hold up. What happened for this PR to be made?

What I understand is someone created a portable SM engine in the BS locker.
Apparently the AI was also Malf and found out about this so... I dunno why this would need to be PR'd. There's a multitude of factors and means a player has to actively go through to even attempt this.

Maia-J:
The main issue was reportedly the anomalies, making it effectively a death chamber. Why not disable those is beyond me. (Also you can use the crystaliser to get a sm shard there anyway)

RG4ORDR:
> 

So what.
This even a bug or even an OP thing someone can do. You need to do specific set ups and collaborate with multiple departments to make a bluespace locker engine in the first place.

iamthesilly:
Guh

ShoziOffical:
Guh indeed

UltraBigRat:
This is one of the dumbest self merges ive seen so far.

PR #17465, opened 2023-01-12
Make Clockies no longer low pop and have a more agressive start.

airlines7:
This PR is to make clock cult no longer a low pop gamemode. The current clockcult can run at 24 people on the server and that kinda sucks if you plan to do your own thing during off hours of the server. [Original Clock cult on /vg/ ](https://github.com/vgstation-coders/vgstation13/pull/5236/files#diff-ed81548292d914a8f553e9ab58886503bd0286c47a3d6ab5689ad829513e8bb9)was never intended for a server with over 50 people on at a time, this is to adapt it to yogstation's pop better because we can easily break 65 people on at once. I've also noticed much of the player base doesn't like the current low pop for clockies with people complaining they have no chance because of a victory (without resorting to bombing the fuck out of reebe which is a problem of its own).

All I'm doing is just changing seven number so hopefully that works.
Min pop requirement is now 40 people
Starting clockies is 5
More clockies per 5 people after 50 pop with a max of 7 (this might be changed if this number includes drones in calculating)

:cl: 
 
tweak: Ratvar is no longer interested in low populated stations.

/:cl:


Moltijoe:
sorry, feature freeze, unless jamie give approval you need to wait until it ends to open. When that happens just ask any of the maintainers to reopen this.

code/game/gamemodes/clock_cult/clock_cult.dm
```diff
@@ -160,12 +160,12 @@ Credit where due:
...
 	if(CONFIG_GET(flag/protect_assistant_from_antagonist))
 		restricted_jobs += "Assistant"
-	var/starter_servants = 4 //Guaranteed four servants
+	var/starter_servants = 5 //Guaranteed five servants
 	var/number_players = num_players()
 	roundstart_player_count = number_players
-	if(number_players > 30) //plus one servant for every additional 10 players above 30
-		number_players -= 30
-		starter_servants += round(number_players / 10)
+	if(number_players > 40) //plus one servant for every additional 5 players above 40
```
MajManatee: Yeah maybe revert this.
Cant wait to have 10 fucking clockies shift start because its high pop and they are now unbeatable.

ThatLing:
> The current clockcult can run at 24 people on the server and that kinda sucks if you plan to do your own thing during off hours of the server.

Shitty reasoning, you can also get killed by a traitor during lowpop

airlines7:
Clock cult forces you to fight, traitor doesn't. If you're good you can run from a traitor, but once gamma hits on clock cult you are required to fight. Also traitors 80% of the time don't have you either gibbed or controlled, something clock cult does with vitality matrices and conversion.

JamieD1:
We are thinking of getting a different map for reebee instead

airlines7:
Well I'm no map maker, but if you're thinking of redoing the map, it could be a good approach to try Beestation's reebe or something similar, other than that I'm no map maker so I'll close the pr since you have a plan with clock cult.

PR #17469, opened 2023-01-13
Unlocks the keys to reality

ynot01:
# Documentation

Tachyon-doppler array will no longer reject bombs small enough to significant net point gain

# Changelog
:cl:  
bugfix: The tachyon-doppler array will now properly respect sufficiently sized negative bombs
/:cl:

### DIFF
--- a/code/game/machinery/doppler_array.dm
+++ b/code/game/machinery/doppler_array.dm
@@ -108,10 +108,10 @@ GLOBAL_LIST_EMPTY(doppler_arrays)
 
 	/*****The Point Calculator*****/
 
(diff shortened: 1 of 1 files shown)


iamthesilly:
YES

ynot01:
This is a bugfix because it says it's not enough for research but in fact it gives a lot to research

Moltijoe:
... but why? do negative bombs even do anything?

iamthesilly:
Negative bombs are mostly theoretical at the moment but would be able to produce a near-infinite amount of research points

ReddicusDragon (approved):

PR #17483, opened 2023-01-14
removes bio bag compatibility w/ reproductive crossbreed because it doesn't work

ReddicusDragon:
you can no longer use bio bags with crossbreeds

# Changelog
:cl:  
rscdel: Removed slime crossbreed bio bag compatibility because it doesn't work  
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/reproductive.dm
+++ b/code/modules/research/xenobiology/crossbreeding/reproductive.dm
@@ -8,7 +8,7 @@ Reproductive extracts:
 	desc = "It pulses with a strange hunger."
 	icon_state = "reproductive"
 	effect = "reproductive"
-	effect_desc = "When fed monkey cubes it produces more extracts. Bio bag compatible as well."
+	effect_desc = "When fed monkey cubes it produces more extracts."
 	var/extract_type = /obj/item/slime_extract/
 	var/cubes_eaten = 0
 	var/last_produce = 0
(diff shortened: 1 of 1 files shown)


Moltijoe:
is it possible to make it work?

ReddicusDragon:
> is it possible to make it work?

5 year old code, you're welcome to try

code/modules/research/xenobiology/crossbreeding/reproductive.dm
```diff
@@ -18,15 +18,6 @@ Reproductive extracts:
 	if((last_produce + cooldown) > world.time)
 		to_chat(user, span_warning("[src] is still digesting!"))
 		return
-	if(istype(O, /obj/item/storage/bag/bio))
-		var/list/inserted = list()
-		SEND_SIGNAL(O, COMSIG_TRY_STORAGE_TAKE_TYPE, /obj/item/reagent_containers/food/snacks/monkeycube, src, 1, null, null, user, inserted)
-		if(inserted.len)
-			var/obj/item/reagent_containers/food/snacks/monkeycube/M = inserted[1]
```
ynot01, line 25: is it perhaps due to this line that it assumes the top item is a monkey cube? have you tried it with only monkey cubes in the bag?

Moltijoe: i could have sworn it's worked for me in the past, so that might be the issue

MajManatee (commented):
I mean, id prefer if it was made to work, but I also have no issued with doing away with it either.

MajManatee:
Author taking a break, feel free to re-open when you're back.

ynot01:
if you plan on retouching on this, please make it so it finds the monkey cube instead of assuming the topmost item is a monkey cube, it's a really simple fix and we shouldn't be tossing features like this

PR #17485, opened 2023-01-14
clock scarabs no longer screw over clock cults

TheRyeGuyWhoWillNowDie:
scarabs being spawned in can bring the threshold of cult size above 20% which can cause the slow teleport thing to activate before any cultists can place brass tiles or convert people

# Changelog

:cl:  
bugfix: clockwork scarabs being created will not interfere with the clock cult's teleportation time
/:cl:


Bibby0110:
Pr freeze

PR #17487, opened 2023-01-14
Fixes a bunch of runtime problems (probably)

Moltijoe:
Look at me ma! i'm doin a maintain!

all of them are just "can't read null.'something' "

:cl:  
bugfix: fixes a bunch of runtime errors (players don't care about this)
/:cl:

### DIFF
--- a/code/datums/traits/negative.dm
+++ b/code/datums/traits/negative.dm
@@ -355,11 +355,11 @@
 	
 /datum/quirk/poor_aim/add()
 	var/mob/living/carbon/human/H = quirk_holder
-	H.dna.species.aiminginaccuracy += 25
+	H?.dna?.species?.aiminginaccuracy += 25
 
 /datum/quirk/poor_aim/remove()
 	var/mob/living/carbon/human/H = quirk_holder
-	H.dna.species.aiminginaccuracy -= 25
+	H?.dna?.species?.aiminginaccuracy -= 25
 
 /datum/quirk/prosopagnosia
 	name = "Prosopagnosia"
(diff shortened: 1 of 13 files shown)


code/datums/traits/negative.dm
```diff
@@ -355,11 +355,11 @@
 	
 /datum/quirk/poor_aim/add()
 	var/mob/living/carbon/human/H = quirk_holder
-	H.dna.species.aiminginaccuracy += 25
+	H?.dna?.species?.aiminginaccuracy += 25
```
Bibby0110: Stuff like this we probably want to runtime

code/modules/modular_computers/file_system/programs/medical/crew_monitor.dm
```diff
@@ -36,7 +36,7 @@
 	var/z = computer.z
 	if(!z)
 		var/turf/T = get_turf(computer)
-		z = T.z
+		z = T?.z
```
Bibby0110: This too

code/datums/traits/negative.dm
Moltijoe: Honestly, i think it was just removal of the quirk that runtimed, i just did it to both parts.

code/modules/modular_computers/file_system/programs/medical/crew_monitor.dm
Moltijoe: Fair enough, never considered that keeping a runtime is better because it can point to a bigger problem

Moltijoe:
```suggestion
		z = T.z
```

code/datums/traits/negative.dm
Moltijoe:
```suggestion
	H.dna.species.aiminginaccuracy += 25
```

Moltijoe (commented):
Removals

JamieD1 (approved):

PR #17490, opened 2023-01-14
[Asteroid] Fix warden being unable to press his shutter buttons

Aquizit:

Moves the tile the two shutter control buttons are actually on before pixel-shifting so the Warden can again access them from his desk. Image shows source to destination tiles.

image

# Changelog

:cl:  
bugfix: Moves the Warden shutter button access from the brig hallway to their office. 
/:cl:

### DIFF
--- a/_maps/map_files/AsteroidStation/AsteroidStation.dmm
+++ b/_maps/map_files/AsteroidStation/AsteroidStation.dmm
@@ -8202,6 +8202,32 @@
 	},
 /turf/open/floor/plasteel,
 /area/crew_quarters/bar)
+"bId" = (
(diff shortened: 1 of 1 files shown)


patpol4 (approved):
also moves the light switch

Hopekz:
But then people from the brig can't press it even though it looks like it is in the brig

Unless your screenshot isn't the end result

Moltijoe:
> But then people from the brig can't press it even though it looks like it is in the brig
> 
> Unless your screenshot isn't the end result

They arent intended to be able to press it though

Bibby0110:
PR freeze. Not game breaking

PR #17499, opened 2023-01-16
removes unintended contents from gax and meta armories

TheRyeGuyWhoWillNowDie:
box is the most popular map but maintaining meta & gax is probably good i think

# Changelog

:cl:  

mapping: Removes unintended contents (the syringe guns) from gax and meta armories
/:cl:


MajManatee:
ok but what are the unintented contents?

ToasterBiome (approved):

PR #17502, opened 2023-01-16
Gives Krokodil Addicts proper traits

Moltijoe:
It was probably a derivative of zombie in the past and piggy backed off their traits.
This should make it so losing your skin no longer makes you immune to all wounds and properly gives the agender tag.

Doesn't fix the sprite draw error, that will be a different PR

:cl:  
bugfix: Krokodil addict no longer lacks all traits
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/zombies.dm
+++ b/code/modules/mob/living/carbon/human/species_types/zombies.dm
@@ -92,6 +92,9 @@
 	id = "goofzombies"
 	limbs_id = "zombie" //They look like zombies
 	sexes = FALSE
+	species_traits = list(HAS_FLESH, HAS_BONE, AGENDER)
+	inherent_traits = list(TRAIT_EASILY_WOUNDED) //you have no skin
+	inherent_biotypes = list(MOB_UNDEAD, MOB_HUMANOID) //pretty much just rotting flesh, somehow still "technically" alive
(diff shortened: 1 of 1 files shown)


ReddicusDragon:
i wonder what the balance implications are, if any. Not sure what mob_undead does

Moltijoe:
> i wonder what the balance implications are, if any. Not sure what mob_undead does

main balance implications are that they used to be immune to wounds because they had neither flesh nor bones.
The only things mob_undead currently does is make them immune to the bonechill ability of the bone golem

MajManatee (approved):
yeah uh idk I see this used like once a month. usually accidentally shooting the person so.

code/modules/mob/living/carbon/human/species_types/zombies.dm
```diff
@@ -92,6 +92,9 @@
 	id = "goofzombies"
 	limbs_id = "zombie" //They look like zombies
 	sexes = FALSE
+	species_traits = list(HAS_FLESH, HAS_BONE, AGENDER)
+	inherent_traits = list(TRAIT_EASILY_WOUNDED) //you have no skin
```
ynot01, line 96: this (TRAIT_EASILY_WOUNDED) was not present before and im inclined to say this is not a part of the bugfix, but otherwise i approve

ynot01 (approved):

ReddicusDragon:
2 maintainer approvals im merging this

PR #17506, opened 2023-01-16
Glockroach properly keep their distance

Moltijoe:
They have a gun, why do they enter melee range?

:cl:  
tweak: Glockroaches now keep their distance
/:cl:


MajManatee:
same reason they hold the glock sideways, they mean buisness duh

JamieD1 (approved):

PR #17504, opened 2023-01-16
Fixes augmented genders

Moltijoe:
I was initially just removing the genders, but then someone pleaded me to not do it
image
Head mentor ladies and gents

Apparently the female sprites weren't tailored to the female model so there were stray pixels everywhere
Also makes genderless species just default to the male icon so they don't turn invisible

default (also engineering because that one just used default)
regular
mining
mining
security
security

:cl:  Papi, Mqiib
bugfix: Genderless species no longer turn invisible when augmented
bugfix: Female characters no longer have too many pixels when augmented
/:cl:

PR #17508, opened 2023-01-16
Rebuffs Ethereal

Moltijoe:
With the lighting change ethereals got nerfed back to previous numbers
Since lighting works differently this time i've only buffed the radius and not the power
Currently the light range is pitifully small, i've also made it linked with health more so a full health ethereal is notably brighter than a hurt ethereal.

:cl:  
tweak: Rebuffs Ethereals
/:cl:


MajManatee (approved):
all (null) etherial mains can rejoice

Moltijoe:
> all (null) etherial mains can rejoice

not true, there is a single ethereal main, they were telling me about how they noticed #17504 like 600 rounds ago but didn't think to mention it

PR #17509, opened 2023-01-16
Adjusts the bluespace bodybag breakout timer

Majkl-J:

Removes the ability of SOME people to surprise bodybag people to easily take them out of fights.

The bluespace bodybag is effectively a 20 second stun, this lowers it to a much more reasonable value of 5 seconds and hopefully fixes the dropping bug.

Yes this is an ided, tag it.

# Changelog

:cl:  
tweak: The bluespace bodybag breakout timer is significantly shorter now
/:cl:

### DIFF
--- a/code/game/objects/items/bodybag.dm
+++ b/code/game/objects/items/bodybag.dm
@@ -82,6 +82,11 @@
 		return
 	user.changeNext_move(CLICK_CD_BREAKOUT)
 	user.last_special = world.time + CLICK_CD_BREAKOUT
+	to_chat(user, span_notice("You claw at the fabric of [src], trying to tear it open..."))
+	to_chat(loc, span_warning("Someone starts trying to break free of [src]!"))
+	if(!do_after(user, 5 SECONDS, src, stayStill = FALSE))
+		to_chat(loc, span_warning("The pressure subsides. It seems that they've stopped resisting..."))
+		return
 	loc.visible_message(span_warning("[user] suddenly appears in front of [loc]!"), span_userdanger("[user] breaks free of [src]!"))
 	qdel(src)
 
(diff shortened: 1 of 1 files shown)


ToasterBiome (changes requested):
look at the proc definition of do_after and you can utilize the stayStill variable to make it so dropping it wont cancel.

code/game/objects/items/bodybag.dm
```diff
@@ -84,7 +84,7 @@
 	user.last_special = world.time + CLICK_CD_BREAKOUT
 	to_chat(user, span_notice("You claw at the fabric of [src], trying to tear it open..."))
 	to_chat(loc, span_warning("Someone starts trying to break free of [src]!"))
-	if(!do_after(user, 20 SECONDS, src))
+	if(!do_after(user, 5 SECONDS, src, stayStill = FALSE))
```
ReddicusDragon, line 87:
```suggestion
	if(!do_after(user, 10 SECONDS, src, stayStill = FALSE))
```
5 feels too low.

Maia-J: That would make it as long as the stunbaton, too much for a "weapon" any med staff can print 20 mins in.

ReddicusDragon: > That would make it as long as the stunbaton, too much for a "weapon" any med staff can print 20 mins in.

I refuse to give thumbs up unless it is 10 seconds. 5 is way too short.

JamieD1:
the whole point of the body bag is that you keep corpses fresh, annoying its being used to gamer shit which was not intended at all

JamieD1 (approved):

PR #17510, opened 2023-01-16
Lets Revenants see like the ghosts they are

Moltijoe:
Revenant is not a strong antag in any manner, their entire thing relies on them being at the right place at the right time, but being unable to see beyond the room they are in just makes it feel tedious and unfun.
Not to mention the flavour of them being just an angry ghost is lost when for some reason becoming angry makes them lose the ability to see through walls.

:cl:  
tweak: Gives revenant X-ray vision
/:cl:

### DIFF
--- a/code/modules/antagonists/revenant/revenant.dm
+++ b/code/modules/antagonists/revenant/revenant.dm
@@ -24,7 +24,7 @@
 	layer = GHOST_LAYER
 	healable = FALSE
 	spacewalk = TRUE
-	sight = SEE_SELF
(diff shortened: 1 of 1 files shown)


Burger1998:
can we let them see other ghosts? or is that too far because its really frustrating trying to direct a revenant in deadchat in the correct direction(where do you think left is motherfucker we're on a 2d plane)

MajManatee:
maybe dont use left and right when you can say "Top right of bar" or "Maints below xenobio"

MajManatee (approved):
also yeah i think that works

PR #17512, opened 2023-01-16
Buffs Holotool, lets it start surgery, and fit in a toolbelt

Moltijoe:
Holotool is weirdly bad for how it's supposed to be the RD equivalent of the CMO hypospray which is amazing.
I've buffed it's speed by quite a lot and let it actually start surgery.

:cl:  
tweak: Buffs Holotool, lets it start surgery, and fit in a toolbelt
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -78,7 +78,8 @@
 		/obj/item/jawsoflife,
 		/obj/item/shuttle_creator, //Yogs: Added this here cause I felt it fits
 		/obj/item/barrier_taperoll/engineering,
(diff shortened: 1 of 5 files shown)


MajManatee (approved):
Yeah this looks fine.

Also can you make it fit in a toolbelt? That would be a game changer there.

ReddicusDragon (commented):
Tested?

Moltijoe:
> Tested?

curse you, my shitcode nearly got it.
Yeah i tested it, but for some reason despite being able to start surgery, it immediately starts to cancel the surgery rather than do the step

I'll fix it in a bit, just busy for the time being

MajManatee:
maintainter btw

PR #17513, opened 2023-01-16
People drop held brains/heads when gibbed

Moltijoe:
Round removal is kinda cringe.
This removes the jank situation where someone tries to bring a head to cloning and it just gets deleted because the holder gets gibbed.

:cl:  
tweak: Getting Gibbed drops all held brains and heads
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/death.dm
+++ b/code/modules/mob/living/carbon/death.dm
@@ -29,10 +29,17 @@
 			M.gib()
 		else
 			visible_message(span_danger("[M] bursts out of [src]!"))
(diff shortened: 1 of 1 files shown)


MajManatee (approved):
sounds like someone got trolled in a round once

Moltijoe:
> sounds like someone got trolled in a round once

nah, just me being an incompetent mining medic accidentally round removing someone

milk-is-cool:
Does this work for carded AIs cause I've had that happen before too

Moltijoe:
> Does this work for carded AIs cause I've had that happen before too

probably not

PR #17517, opened 2023-01-16
Bluespace locker no longer has infinite power

ThatLing:
It's stupid that the bluespace locker gives you an area with infinite power for free. You will now have to provide an external power source (e.g. P.A.C.M.A.N generator) if you want to use machines inside.

I added an APC inside the locker and removed the ``requires_power = FALSE`` flag so stuff actually requires power. The APC spawns with 90% charge like default, however the 2 lights inside will use some of it over time.

image

# Changelog

:cl:  
tweak: Bluespace locker no longer has infinite power, it's powered by an APC now and requires an external power source.
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom.dmm
+++ b/_maps/map_files/generic/CentCom.dmm
@@ -18929,7 +18929,7 @@
 	dir = 4
 	},
 /obj/machinery/door/airlock/centcom{
-	armor = list("melee" = 90, "bullet" = 90, "laser" = 90, "energy" = 90, "bomb" = 90, "bio" = 100, "rad" = 100, "fire" = 100, "acid" = 90);
+	armor = list("melee"=90,"bullet"=90,"laser"=90,"energy"=90,"bomb"=90,"bio"=100,"rad"=100,"fire"=100,"acid"=90);
 	name = "Test Chamber Blast Doors"
 	},
 /obj/structure/fans/tiny,
@@ -19943,7 +19943,7 @@
 	dir = 8
 	},
 /obj/machinery/door/airlock/centcom{
-	armor = list("melee" = 90, "bullet" = 90, "laser" = 90, "energy" = 90, "bomb" = 90, "bio" = 100, "rad" = 100, "fire" = 100, "acid" = 90);
+	armor = list("melee"=90,"bullet"=90,"laser"=90,"energy"=90,"bomb"=90,"bio"=100,"rad"=100,"fire"=100,"acid"=90);
 	name = "Test Chamber Blast Doors"
 	},
 /obj/structure/fans/tiny,
@@ -20580,7 +20580,7 @@
 	dir = 8
 	},
 /obj/machinery/door/airlock/centcom{
-	armor = list("melee" = 90, "bullet" = 90, "laser" = 90, "energy" = 90, "bomb" = 90, "bio" = 100, "rad" = 100, "fire" = 100, "acid" = 90);
+	armor = list("melee"=90,"bullet"=90,"laser"=90,"energy"=90,"bomb"=90,"bio"=100,"rad"=100,"fire"=100,"acid"=90);
 	name = "Test Chamber Blast Doors"
 	},
 /obj/structure/fans/tiny,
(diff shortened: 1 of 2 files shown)


patpol4 (changes requested):
We don't use the autoname apc's, please use one of the manual ones to set the area and name by hand.

patpol4:
My personal opinion is that this is needlesly limiting, bluespace locker is supposed to be this magical fun place where you can make crazy projects without having to worry about things such as no power for your machines.

MajManatee:
I agree with Wej

ReddicusDragon:
this is stupid. Why ruin player fun? People love using the bluespace locker as a place to bullshit and build funny small projects

TheRyeGuyWhoWillNowDie:
Mid

RG4ORDR:
Absolutely not.
I wont even give an explanation. Just bad all round.

ThatLing:
> My personal opinion is that this is needlesly limiting, bluespace locker is supposed to be this magical fun place where you can make crazy projects without having to worry about things such as no power for your machines.

Why should BS locker be "magical" and have infinite power? If you want to build crazy projects without worrying about power, we have both aux base construction and normal construction area on Yogstation, with the former even giving you your own solar for power.


> this is stupid. Why ruin player fun? People love using the bluespace locker as a place to bullshit and build funny small projects

God forbid people have to actually put in a little effort to get power inside the locker. You're acting like this will make the locker useless when in reality there's a dozen ways to power it, such as a PACMAN or simply changing out the power cell.

JamieD1 (approved):

Moltijoe:
Would be nice if the APC inside the locker could charge from the APC of the room the locker is in. That way its not infinite power, but doesnt gimp making cool non power creating setups inside the locker.

TheLinuxOS:
This is a horrible change, not only does it detract from the magic of the bluespace locker- It also removes 1 of the precious tiles by making you build a generator in there that could go completely against the aesthetic.

adamsong:
If you are going to make it draw power, make it draw from the area in which the outside is

ThatLing:
> This is a horrible change, not only does it detract from the magic of the bluespace locker- It also removes 1 of the precious tiles by making you build a generator in there that could go completely against the aesthetic.

Or you can just replace the power cell in the APC and still use all of the tiles, crazy right? I hope you realize that the bluespacer having infinite power because "magic" is an insanely lame excuse.

patpol4 (approved):
map fine

TheLinuxOS:
> > This is a horrible change, not only does it detract from the magic of the bluespace locker- It also removes 1 of the precious tiles by making you build a generator in there that could go completely against the aesthetic.
> 
> 
> 
> Or you can just replace the power cell in the APC and still use all of the tiles, crazy right? I hope you realize that the bluespacer having infinite power because "magic" is an insanely lame excuse.

and I hope you realize how insanely lame it is to try to limit the bluespace locker

ThatLing:
> > > This is a horrible change, not only does it detract from the magic of the bluespace locker- It also removes 1 of the precious tiles by making you build a generator in there that could go completely against the aesthetic.
> > 
> > 
> > Or you can just replace the power cell in the APC and still use all of the tiles, crazy right? I hope you realize that the bluespacer having infinite power because "magic" is an insanely lame excuse.
> 
> and I hope you realize how insanely lame it is to try to limit the bluespace locker

What a cope reply

Moltijoe:
why are people trying to balance around the bluespace locker? it's tedious as hell to find it, if someone finds it only to figure out all the fun gimmicks were killed, that's kinda cringe.

JamieD1:
> why are people trying to balance around the bluespace locker? it's tedious as hell to find it, if someone finds it only to figure out all the fun gimmicks were killed, that's kinda cringe.

Its hardly tedious to find, people go assistant just to find the locker and they find it before like 20-30 mins. If anything it needs to be made harder to find.

ThatLing:
> why are people trying to balance around the bluespace locker? it's tedious as hell to find it, if someone finds it only to figure out all the fun gimmicks were killed, that's kinda cringe.

You're legitimately delusional if you think this "kills" all fun gimmicks you can do

UltraBigRat:
Biggest question here, why? whats the reasoning here other then bluespess locker bad i nerf? i really dont see how this is good for the game whatsoever.

CrazyDrakkon:
The APC power/status stays the same when the locker is destroyed.

PR #17529, opened 2023-01-16
Laser scalpel no longer makes you look like a spessman-shaped nuclear fuel rod

Mqiib:
Changed the base power from 1 to 0.2 so now you don't glow weirdly

image
image

# Changelog

:cl:  

tweak: Scalpel mode for the laser scalpel now significantly less bright

/:cl:

### DIFF
--- a/code/modules/surgery/tools.dm
+++ b/code/modules/surgery/tools.dm
@@ -414,7 +414,7 @@
 		tool_behaviour = TOOL_SCALPEL
 		to_chat(user, span_notice("You lower the power, it can now make precise incisions."))
 		set_light_range(1)
-		set_light_power(0.3)
+		set_light_power(0.2)
 		force -= 1
 		icon_state = "scalpel_a"
 
--- a/code/modules/surgery/tools.dm
+++ b/code/modules/surgery/tools.dm
@@ -397,6 +397,7 @@
 	light_system = MOVABLE_LIGHT
 	light_range = 1
 	light_color = LIGHT_COLOR_GREEN
+	light_power = = 0.2	//Barely glows on low power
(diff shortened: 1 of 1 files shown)


Moltijoe:
Yeah, with the new lighting system most things feel a lot brighter than before

code/modules/surgery/tools.dm
```diff
@@ -397,6 +397,7 @@
 	light_system = MOVABLE_LIGHT
 	light_range = 1
 	light_color = LIGHT_COLOR_GREEN
+	light_power = = 0.2	//Barely glows on low power
```
ThatLing:
```suggestion
	light_power = 0.2 //Barely glows on low power
```

ThatLing:
> Yeah, with the new lighting system most things feel a lot brighter than before

I kept the same range for all light sources, it just feels larger because it looks different. However, feel free to make a PR to change the numbers

MajManatee:
is that not what hes doing?

Moltijoe (approved):
SlingBuff

code/modules/surgery/tools.dm
Moltijoe:
```suggestion
	light_power = 0.2	//Barely glows on low power
```

Moltijoe (changes requested):
you reverted the fix

ThatLing:
> is that not what hes doing?

I was replying to Molti, who isn't the author of this PR

Moltijoe (approved):
is less glowy, it works

code/modules/surgery/tools.dm
Mqiib: oops lmao

PR #17533, opened 2023-01-17
Sprayed runes and graffiti are now more convincingly cult runes and gang graffiti

ynot01:

`is_alpha()` returned true no matter what you were drawing (unless it was a number) because it only respects the first letter and because that was the first check it would always default to `letter`. Now it is the last check.

Also removed a snowflake check that would rename gang graffiti to make it obvious that it was gang graffiti.

Cult runes and gang sprays should now be identical to sprayed runes except for the fact that they are perfectly centered which can be achieved by other means.

# Changelog

:cl:  
bugfix: Fixed a bug where every drawn graffiti would show up as a letter instead of rune, graffiti, number, symbol, etc
tweak: Cult runes are no longer obvious about being made of blood
tweak: Gang sprays are no longer named gang graffiti
/:cl:

### DIFF
--- a/code/game/objects/items/crayons.dm
+++ b/code/game/objects/items/crayons.dm
@@ -315,11 +315,9 @@
 			drawing = pick(all_drawables)
 
 
-	var/temp = "rune"
-	if(is_alpha(drawing))
-		temp = "letter"
-	else if(is_digit(drawing))
-		temp = "number"
+	var/temp = "scribble"
+	if (drawing in runes)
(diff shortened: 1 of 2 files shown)


Moltijoe:
hmm, interesting cult buff, no longer instantly outed by someone wandering maints

patpol4:
Good, no meta

Moltijoe (approved):
Can't find any problems with the code

PR #17547, opened 2023-01-17
The syndicate has ordered newer, more accurate sniper rifles from space australia

Mqiib:
Now they have 0 innate spread because for a gun that fires once every FOUR SECONDS (that's 10x the base fire delay of guns), missing one shot because of RNG can really cause problems. Also it's a special nukie-only weapon (when seen on the station) so it being a sniper rifle that can miss is a bad thing.

# Changelog

:cl:  

tweak: Syndicate 50 cal sniper rifle base inaccuracy reduced from 5 to 0

/:cl:


MajManatee:
absurd... when will the powercreep end...

Moltijoe (approved):
number change is easy

ynot01 (approved):
sniper rifle should be the last thing to be inaccurate

PR #17548, opened 2023-01-17
Adds cloaker belt to uplink

jachlompsky:

Puts the cloaker belt in the uplink. Cloaker belt exists in the code already, I found it under yogstation/code/game/objects/items/devices/traitordevices.dm. Searched through PRs and discord and never saw it mentioned so I guess its fine to implement.

The belt charges, and drains charge, faster the darker or brighter it is. Only charges and drains while the belt is active. At full charge, you stay invisible in a well-lit area for about 5-6 seconds, then begin slowly fading into full visibility. The invisibility is not affected by moving or interacting with things.

I decided to make it 10 TC, 2 more than the stealth implant. Though the belt is more conspicuous and has limited charge, you aren't restricted to being in a box. Uses the toolbelt sprite, but is still called "cloaker belt," because metashielding it would be too strong and giving it a unique sprite would be giving your stealth item a "valid me" sign. Edit: Nukies can purchase it for 20 tc and i might change the price again

Hopefully, this should provide a way for traitors to be stealthy, but still make an impact on the round; It encourages you to sabatoge lights and power to sneakily move around the station.

Edit: Ok I double checked and it has to be activated to charge in the dark. I don't think it really matters and maybe that can get changed later but yeah
Nukies get it for 20 tc this is open to debate sorry I keep changing things

# Wiki Documentation

Add cloaker belt to the uplink items page.

# Changelog

:cl:  
rscadd: Added Cloaker Belt to the uplink
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1380,14 +1380,14 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 
 /datum/uplink_item/stealthy_tools/shadowcloak
 	name = "Cloaker Belt"
-	desc = "Renders the wearer invisible while active. Has a short charge that is refilled in darkness."
+	desc = "Renders the wearer invisible while active. Has a short charge that is refilled in darkness; only charges when in use.
 	item = /obj/item/shadowcloak
 	cost = 10
 	exclude_modes = list(/datum/game_mode/nuclear)
 
 /datum/uplink_item/stealthy_tools/nuclearshadowcloak
 	name = "Cloaker Belt"
-	desc = "Renders the wearer invisible while active. Has a short charge that is refilled in darkness."
+	desc = "Renders the wearer invisible while active. Has a short charge that is refilled in darkness; only charges when in use.
 	item = /obj/item/shadowcloak
 	cost = 60
 	include_modes = list(datum/game_mode/nuclear)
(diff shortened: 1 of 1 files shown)


ynot01:
give it to the nukies for like 60 tc lol

MajManatee:
i love invisible nukies

code/modules/uplink/uplink_items.dm
```diff
@@ -1378,6 +1378,20 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
...
+	item = /obj/item/shadowcloak
+	cost = 10
+	exclude_modes = list(/datum/game_mode/nuclear)
+
+/datum/uplink_item/stealthy_tools/nuclearshadowcloak
+	name = "Cloaker Belt"
+	desc = "Renders the wearer invisible while active. Has a short charge that is refilled in darkness; only charges when in use."
+	item = /obj/item/shadowcloak
+	cost = 60
+	include_modes = list(datum/game_mode/nuclear)
```
Hopekz:
```suggestion
	include_modes = list(/datum/game_mode/nuclear)
```

UltraBigRat:
isnt 60 tc a bit much for nukies? They already can get stealth implant, and making it 60 tc just creates more confusion with prices and how they actually correlate to the value of the item.

jachlompsky:
> isnt 60 tc a bit much for nukies? They already can get stealth implant, and making it 60 tc just creates more confusion with prices and how they actually correlate to the value of the item.

Extra arm used to cost more for nukies than traitors because it was more useful for them (and then got removed from nukies for being too useful), so I'm applying the same reasoning here. You can still run around and shoot people while cloaking, whereas the stealth implant can't do that. 60 tc is definitely a ton, I just wanted some more feedback on nukie balancing before I lowered it. I think 20 or so tc would be good, though that could change eventually if it ends up too strong or weak.

UltraBigRat:
Yeah thats probably better, 60tc just sounds like bait for new nukies to buy and spend half their tc on something they dont need that much.

Moltijoe (approved):
Can't find anything wrong with the code

PR #17555, opened 2023-01-18
Update guardian.dm to remove duplicated proc

ReddicusDragon:

Fixes a duplicated proc


MajManatee:
code didnt work without it

ReddicusDragon:
> code didnt work without it

It compiles. You can take it up with @JamieD1 if that’s the case, I’m taking a month or two away from the code.

MajManatee (approved):
idk it probably wont break horribly

PR #17559, opened 2023-01-18
Makes Preternis also considered mob_robotic

Moltijoe:
This does nothing to change gameplay currently, but should increase consistency in the future as preternis are both organic AND robotic.

:cl:  
tweak: Nothing you need to worry about as a player
/:cl:


MajManatee:
sounds kinda snowflakey...

Moltijoe:
> sounds kinda snowflakey...

the opposite of such, it removes the need for snowflakey code when making something that affect robots and preternis

PR #17568, opened 2023-01-19
riot shotguns need 2 hands to shoot

TheRyeGuyWhoWillNowDie:
riot shotguns need two hands to shoot. 

does not affect combat shotguns so they're actually competitive versus riot shotguns. this is to kill the shield + shotgun shit on everything meta, makes several gamemodes less unenjoyable (primarily darkspawn)

# Changelog

:cl:  
tweak: riot shotguns are two-handed
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/shotgun.dm
+++ b/code/modules/projectiles/guns/ballistic/shotgun.dm
@@ -40,6 +40,7 @@
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/riot
 	sawn_desc = "Come with me if you want to live."
 	can_be_sawn_off  = TRUE
+	weapon_weight = WEAPON_HEAVY
 
 // Breaching Shotgun //
 


MajManatee:
i hate darkspawns so this is a bad change

patpol4:
But he can do it

TheRyeGuyWhoWillNowDie:
> But he can do it 

that's a lever action shotgun, these are pump action

Marmio64:
> > But he can do it 
> 
> that's a lever action shotgun, these are pump action

make pumping require an empty hand? Let us still one hand rapid fire sawed off shotguns like flintlock pistols
(in seriousness i would rather have them still be usable in one hand somewhat but be restricted.)

ReddicusDragon:
interesting

kooarbiter:
I vote we replace all riot shotguns in armory with sawn off double barrels that only load buckshot

Bibby0110 (approved):

00aquir:
> does not affect combat shotguns so they're actually competitive versus riot shotguns. 

wait what? combat shotguns are much better than riot ones, you can shoot faster with them, you don't need to do any pumping action, and yet they won't be affected? It would make sense the other way around.
There are also syndicate shotguns which can shoot quickly like combat ones.

Issue #17561, opened 2023-01-18
 Interrogation Room Monitor on Box is hooked up wrong


Yogbot-13:
## Test Merges: 
Not Supplied / None
## Reproduction:
Description: it shows the brig phys' camera instead of the interrogation room's

 Submitted by: Davos (555869314784428032)

Aquizit:
#17562

patpol4:
#17569

PR #17569, opened 2023-01-19
Box interrogation camera fix

Aquizit:

As the title says. Replaced the camera in brig phys area, which was a copy-paste of the one in Interrogation, which was causing it to show up incorrectly when trying to monitor Interrogation. (Used the version from yogsmeta, just with a different dir so it's on the wal properly.)

Fixes issue #17561 

# Changelog

:cl:  
bugfix: The monitor outside interrogation now shows the correct view again
bugfix: Brig Infirmary camera now shows up correctly on the standard camera console
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -54724,23 +54724,6 @@
 	icon_state = "platingdmg2"
 	},
 /area/maintenance/aft)
-"sig" = (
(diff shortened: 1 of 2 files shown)


patpol4 (changes requested):
No need for the meta changes, the cams on meta actually use the "-" while box does not.

patpol4 (approved):
thanks

PR #17577, opened 2023-01-20
(port) adds disfiguration

boodaliboo:
original pr:https://github.com/tgstation/tgstation/pull/45297
"Disfiguration: Makes the host disfigured, causing them to appear as Unknown. (I originally said prosopagnosia but it turns out that trait doesn't affect what you see while mousing over mobs, so it is kind of pointless as a symptom.)" IT IS A LEVEL 5 SYMPTOM
:cl:  
rscadd: Added new viro symptom Disfiguration: Makes the host disfigured, causing them to appear as Unknown. (I originally said prosopagnosia but it turns out that trait doesn't affect what you see while mousing over mobs, so it is kind of pointless as a symptom.)
/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/disfiguration.dm
+++ b/code/datums/diseases/advance/symptoms/disfiguration.dm
@@ -1,0 +1,46 @@
+/*
+//////////////////////////////////////
+Disfiguration
+	Hidden.
+	No change to resistance.
+	Increases stage speed.
+	Slightly increases transmittability.
+	Critical Level.
+BONUS
+	Adds disfiguration trait making the mob appear as "Unknown" to others.
+//////////////////////////////////////
+*/
+
+/datum/symptom/disfiguration
+
+	name = "Disfiguration"
+	desc = "The virus liquefies facial muscles, disfiguring the host."
+	stealth = 2
+	resistance = 0
+	stage_speed = 3
+	transmittable = 1
+	level = 5
+	severity = 1
+	symptom_delay_min = 25
+	symptom_delay_max = 75
+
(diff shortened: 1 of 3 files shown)


ThatLing (changes requested):
Have you tested this? Seems like you have to actually add the symptom in other places to actually use it. From quickly at the original PR, seems like ``code/modules/antagonists/disease/disease_abilities.dm`` is the approriate file.

boodaliboo:
> Have you tested this? Seems like you have to actually add the symptom in other places to actually use it. From quickly at the original PR, seems like `code/modules/antagonists/disease/disease_abilities.dm` is the approriate file.

fuck let me check my other branches and see if i changed that shit their

code/datums/diseases/advance/symptoms/disfiguration.dm
```diff
@@ -0,0 +1,46 @@
...
+			M.visible_message("<span class='warning'>[M]'s face appears to cave in!</span>", "<span class='notice'>You feel your face crumple and cave in!</span>")
+		else
+			M.visible_message("<span class='warning'>[M]'s face begins to contort...</span>", "<span class='notice'>Your face feels wet and malleable...</span>")
+
+
+/datum/symptom/disfiguration/End(datum/disease/advance/A)
+	. = ..()
+	if(!.)
+		return
+	if(A.affected_mob)
```
ThatLing, line 46: You're missing a part of your file mate

ThatLing (changes requested):

MajManatee:
Does not seem to intend to fix issues, nor make functional. Re-open when fixed.

PR #17578, opened 2023-01-20
(ports) adds organ preservation nanite program

boodaliboo:
"Added a new nanite program, Organ Preservation. While the host is dead, it consumes 0.4 volume per tick to prevent organ decay, like formaldehyde would."

"This PR makes formaldehyde prevent organs from decaying in dead bodies and adds 3u formaldehyde to the epinepherine medipens." i chose not to add formaldehyde to epi pens i will if asked to
Original PRS
https://github.com/tgstation/tgstation/pull/54542
https://github.com/tgstation/tgstation/pull/47770
:cl:  
rscadd: Added formaldehyde preserving bodies and a nanite program that does aswell
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -230,6 +230,8 @@
 #define TRAIT_BADMAIL			"badmail"	//Your mail is going to be worse than average
 #define TRAIT_SHORT_TELOMERES	"short_telomeres" //You cannot be CLOONED
 #define TRAIT_LONG_TELOMERES	"long_telomeres" //You get CLOONED faster!!!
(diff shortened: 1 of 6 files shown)


boodaliboo:
apparently theos already did the formaledehyde thing in a different way so i may be fucked with my lack of coding skills to fix it

CMOisLing:
> apparently theos already did the formaledehyde thing in a different way so i may be fucked with my lack of coding skills to fix it

The formaledehyde works through the organ preservation trait. You should probably just make it so the nanites and formaledehyde both check for the trait for their tick and add it if you dont have it already so they dont mess with each other when one goes away and removes the trait.

MajManatee:
Does not seem to intend to fix issues, nor make functional. Re-open when fixed.

PR #17579, opened 2023-01-20
(port) "Adds chromosome effects to Cryokinesis, Shocktouch, Chameleon"

boodaliboo:
this is likely going to have merge errors due to changes in code ill attempt to fix them but, i may of fucked up because im not sure if their shock touch is ranged and i made it for the wrong one any help would be appreciated
i didnt touch any of the instabilities
"Cryokinesis instability up by 10 (from 20 to 30)
Cryokinesis base cooldown up by 1 second (from 15 to 16) now accepts energetic chromosomes (8 seconds when applied)

Shock touch instability up by 5 (from 30 to 35) now accepts both energetic and power chromosomes

Shock touch cooldown up by 2 (from 10 to 12) (cooldown of 6 when energetic is applied)

If a power chromosome is applied Shock touch does a weak 7500 power (I`ve seen it hit from 5-7 damage) tesla shock with range of 7 when used on a viable target (extra damage does not apply to target, no you cant use it on yourself, no you cant use it on someone that is shock immune, yes you have to put yourself in harms way to apply the tesla shock.)

(The vars for the tesla shock can be edited by admins for events/abuse)

Chameleon now accepts power chromosomes to fully stealth about 3 seconds faster."
:cl:  
rscadd: Added new special effects if power chromosome is used on shock touch
tweak: adds the abilty to use more chromosomes on mutations
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -230,6 +230,8 @@
 #define TRAIT_BADMAIL			"badmail"	//Your mail is going to be worse than average
 #define TRAIT_SHORT_TELOMERES	"short_telomeres" //You cannot be CLOONED
 #define TRAIT_LONG_TELOMERES	"long_telomeres" //You get CLOONED faster!!!
+#define TRAIT_ORGAN_PRESERVATION	"organ_preservation" //Prevents organ decay while dead
+
 
 /// This person is crying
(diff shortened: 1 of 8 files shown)


MajManatee:
Does not seem to intent to fix issues, nor make functional. Re-open when fixed.

PR #17585, opened 2023-01-20
[ADMIN] admin logs to stuff that should have it

ynot01:

Added an admin log to viewing variables, which is weird that there wasn't one

Added a `message admins` & admin log to renaming something via view variables panel, which is weird that there wasn't one

### DIFF
--- a/code/modules/admin/view_variables/view_variables.dm
+++ b/code/modules/admin/view_variables/view_variables.dm
@@ -80,6 +80,8 @@
 			names += V
 	sleep(0.1 SECONDS)//For some reason, without this sleep, VVing will cause client to disconnect on certain objects.
 
+	log_admin("[key_name(usr)] viewed the variables of [D].")
+
 	var/list/variable_html = list()
 	if (islist)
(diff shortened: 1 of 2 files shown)


code/modules/admin/view_variables/view_variables.dm
```diff
@@ -80,6 +80,8 @@
 			names += V
 	sleep(0.1 SECONDS)//For some reason, without this sleep, VVing will cause client to disconnect on certain objects.
 
+	log_admin("[key_name(usr)] viewed the variables of [D].")
```
patpol4, line 83: this will spam

ynot01: no more than other admin logs, it is important for admin investigations

ynot01: resolved in discord

patpol4: thought it would message each time, logs are good

JamieD1 (approved):

PR #17588, opened 2023-01-20
slugs have even more damage falloff and overall shotgun pellet reworks

Skrem7:

hm i wonder why buckshot was where it was before

makes shotguns less obnoxious damage-wise considering they're the most lethal weapon in sec's arsenal yes i understand this "nerfs" cargo antags and the gimmick of stealing armory however have you also considered using antag gear or a mosin or a wt (you're next bucko) or a "good enough" laser gun or a bow (I'll be making them not shit soon™)

ADDENDUM: Pellets are now more accurate (not by a MASSIVE margin) because the idea is to make them not just a club of a melee weapon and they should still be effective in close quarters. So! If you're more accurate, you hit more! AKA they work a lil more like they do in real life (but not by much!)

also slugs have more falloff of both damage types now and now slugs won't qdel themselves if they go out of range bcuz uh that might unintentionally slap some slug types that go kaboom or zappy or whatever the hell

# Wiki Documentation

DU slug does 35 instead of 25 damage now
Buckshot does 11 per pellet instead of 12 for a total of 66 instead of 72
ILLEGAL buckshot does 14.5 instead of 15.5 for a total of 87 instead of 93
Flechette does 12 instead of 13 now for a total of 72 instead of 78
Rubbershot does 3/13 instead of 3/14.5 for a total of 18/78 instead of 18/87
Improv pellets no longer have random range

# Changelog

:cl:  
tweak: Pellet damage nerfed across the board, mostly
tweak: Pellets are more tightly clumped in every pellet-based shell
tweak: Improv pellets now don't have a random range because why printing buckshot is so easy
tweak: Slugs have more damage falloff but they don't delete themselves if they don't have damage
tweak: DU slug damage from 25 to 35
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition/ballistic/revolver.dm
+++ b/code/modules/projectiles/ammunition/ballistic/revolver.dm
@@ -11,7 +11,7 @@
 	desc = "A .357 Ironfeather shell that contains six pellets."
 	projectile_type = /obj/item/projectile/bullet/pellet/a357_ironfeather
 	pellets = 6
-	variance = 25 //Same spread and pellets as buckshot
+	variance = 20 //Same spread and pellets as buckshot
 
 /obj/item/ammo_casing/a357/nutcracker
 	name = ".357 Nutcracker bullet casing"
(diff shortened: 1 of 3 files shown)


Moltijoe (approved):

MajManatee:
DU slugs my beloved

PR #17589, opened 2023-01-20
kicks .38 wound chance

Skrem7:

less likely to wound still less likely to wound on bare naked targets but should be still reputable when doing so

bullets should bleed people out i believe however t3 wound per shot is a little not poggers

ran the numbers it should be better? it might need more tweaking based on how it performs

# Changelog

:cl:  
tweak: .38 overall worse at wounding however the difference between wounding with/without wound armor is greater
/:cl:


Moltijoe (approved):
yeah, det shouldn't be killing people with the wounds if they're armoured

PR #17591, opened 2023-01-20
makes hardlight bow recharge rates reasonable + makes their arrows even better

Skrem7:

I came here expecting to reduce the recharge a little of the bows and buff up the projectiles and discovered the Syndicate bow has a recharge of TWO WHOLE SECONDS so I just kinda halved that and made the universal charge time one second, because it was obnoxious

this is technically a clockie buff too because they get to fire their arrows more often

Buffs up all the arrows too damage-wise (EXCEPT RATVAR) (it's probably a mistake) because I would rather overtweak them then run it back with how much of a meme of a weapon it is right now so we'll see? Burn and x-ray got negative wound_bonuses because NONE of the arrows have any wound stuff so uh yeah. HOWEVER, Ratvar arrows got a +5 wound bonus because it was pretty low and I have decided that I hate people today

# Wiki Documentation

Burn arrow now does 40 burn instead of 32
Stamina arrow now does 50 instead of 48
X-ray arrow now does 30/500 for burn/irradiate instead of 21/400

# Changelog

:cl:  
tweak: All energy bows have had their recharge_time reduced to 1 from 1.5. For the syndicate bow it is going from TWO to ONE (HALF the time) because it had a TWO SECOND RECHARGE
tweak: All arrows (except ratvar!!) do even MORE damage (x-ray has MORE irradiate)
tweak: Burn and x-ray arrows now have negative wound chance, ratvar arrow had a very slight wound chance increase
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/bow.dm
+++ b/code/modules/projectiles/guns/ballistic/bow.dm
@@ -160,7 +160,7 @@
 	mag_type = /obj/item/ammo_box/magazine/internal/bow/energy
 	no_pin_required = FALSE
 	draw_slowdown = 0
-	var/recharge_time = 1.5 SECONDS
+	var/recharge_time = 1 SECONDS
 
 /obj/item/gun/ballistic/bow/energy/update_icon()
 	cut_overlay(arrow_overlay, TRUE)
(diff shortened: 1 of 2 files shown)


ynot01 (approved):
Does what is documented

iamthesilly:
Great PR, bows are cool

PR #17593, opened 2023-01-20
Fixes runtime when firing suppressed weapons without suppressors

SomeguyManperson:

I think this works

:cl:  
bugfix: fixes runtime when firing a gun that is suppressed, but doesn't have a suppressor attached. Probably no visible changes
/:cl:


ynot01:
This would make weapons with `suppressed = TRUE`, such as the ebow, display `user.visible_message(span_danger("[user] fires [src]!"), null, null, COMBAT_MESSAGE_RANGE)`

ynot01 (approved):

PR #17601, opened 2023-01-21
Reduces touch of death cooldown and makes it deal up to 200 brute damage

Notamaniac:

Reduces the cooldown of touch of death and causes it to deal up to 200 brute damage to prevent easy defibs.

# Wiki Documentation

Touch of death cooldown is now 40/35/30/25/20

# Changelog

:cl:  
tweak: Touch of death cooldowns are now 40/35/30/25/20 (20 second reduction from before) and now deals up to 200 brute damage (on top of killing it's victim).
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/godhand.dm
+++ b/code/modules/spells/spell_types/godhand.dm
@@ -226,5 +226,6 @@
 		qdel(suit)
 		new /obj/effect/decal/cleanable/molten_object(M.loc)
 		return ..()
+	M.adjustBruteLoss(max(200-(M.getOxyLoss() + M.getToxLoss() + M.getBruteLoss() + M.getFireLoss()),0))		
 	M.death(FALSE)
 	return ..()	//yogs end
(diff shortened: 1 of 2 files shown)


BurgerLUA:
This is probably a nitpick but I don't know if I agree with the 200 value as most doctors assume that 200 means suicide.

RG4ORDR:
> This is probably a nitpick but I don't know if I agree with the 200 value as most doctors assume that 200 means suicide.

Any person that plays medbay can just look at the body and start fixing it and try to revive them.
200 damage doesn't mean it's a suicide.

UltraBigRat:
Better then nothing, still is just tend wounds into defib though

Moltijoe (approved):
numbers easy

MajManatee:
if only we had a similar spell that just did... MORE...
maybe gibbing someone so it doesnt take 30 seconds to revive them, idk just an idea.

Moltijoe:
> if only we had a similar spell that just did... MORE... maybe gibbing someone so it doesnt take 30 seconds to revive them, idk just an idea.

gibbing is far more than just death, under certain conditions that's the equivalent of round removal.

MajManatee:
good thing there isnt like 6 types of cloning that can save you.

PR #17605, opened 2023-01-21
punctuation fix for drone law

ktlwjec0:
what the title says, had a comma at the end of a sentence

also adds a missing space to hypnosis text so i dont have to make a brand new pr


Moltijoe (approved):
Should probably have atomized this, but it's a lot of extra work just for one additional space

PR #17606, opened 2023-01-22
aggressive grabs are significantly worse for the game

TheRyeGuyWhoWillNowDie:
fun fact there is a good reason why you couldn't shove or grab while aggressive grabbed, and the change made it more or less impossible to upgrade grab levels since you could just be shove-spammed

this change doesn't make it impossible to use items while grabbed, however you must now actually put in effort to escape grabs.

:cl:  
bugfix: you can no longer spidermonkey people to escape grabs
tweak: you can't grab while aggro grabbed
tweak: you can't shove while aggro grabbed
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -965,7 +965,9 @@
 
 //src is the user that will be carrying, target is the mob to be carried
 /mob/living/carbon/human/proc/can_piggyback(mob/living/carbon/target)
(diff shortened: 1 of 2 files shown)


bruhlookatthisdood:
what in gods name is a spider monkey

UltraBigRat:
Woo grabs no longer suck

MajManatee (approved):
No longer can i piggy back my attackers to get away, this is so sad

PR #17607, opened 2023-01-22
Fixes some mech punch weirdness

Mqiib:
This pr does two things:
1: technically buffs mech punches to throw + knockdown if the mech has 20 force rather than greater than 20 force (enabling rocket fist + non-combat mech shenanigans for tator robos)
2: makes standard mech punches respect the low damage no-throw/knockout limitation, so odysseuses an ripleys and stuff shouldn't hard stun borgs now

# Changelog

:cl:  

tweak: Reduces mech knockdown threshold on punches to 20 (from 21)
bugfix: Low damage mech punches no longer stun and throw silicons and other non-human mobs
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -391,7 +391,7 @@
 			var/dmg = rand(M.force/2, M.force)
 			switch(M.damtype)
 				if(BRUTE)
-					if(M.force > 20)
+					if(M.force >= 20)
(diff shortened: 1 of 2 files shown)


MajManatee (approved):
Might need more approval since it is TECHNICALLY a buff

Mqiib:
> Might need more approval since it is TECHNICALLY a buff

It's technically a mech buff for specifically the weaker/underpowered melee weapon available to the crew, the fist.
It's more consistently a mech _nerf_ since regular mechs no longer hard stun borgs and non-carbon/humans

PR #17616, opened 2023-01-22
Removes Cringe Metashield from IPC Martial Arts - Wings are obvious as fuck

JamieD1:
image

Wings are now on examine and if they are martial arts wings they are obvious via hacked message.

:cl:  
rscadd: Examine on Wings
tweak: IPC Martial Arts is now Obvious
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -116,7 +116,7 @@
 	if("wings" in dna.species.mutant_bodyparts)
 		var/badwings = ""
 		if(mind?.martial_art && istype(mind.martial_art, /datum/martial_art/ultra_violence))
-			badwings = "Weaponised "
+			badwings = "Weaponized "
 		. += "[t_He] [t_has] also have a pair of [span_warning(badwings)][(dna.features["wings"])] wings on their back"
 
 	var/appears_dead = 0
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -114,10 +114,10 @@
 			. += span_warning("[t_He] [t_is] twitching ever so slightly.")
 
 	if("wings" in dna.species.mutant_bodyparts)
-		var/hacked = ""
+		var/badwings = ""
 		if(mind?.martial_art && istype(mind.martial_art, /datum/martial_art/ultra_violence))
-			hacked = ionnum() + ionnum() + " "
-		. += "[t_He] [t_has] also have a pair of [span_warning(hacked)][(dna.features["wings"])] wings on their back"
+			badwings = "Blood Soaked"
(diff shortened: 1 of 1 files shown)


Hopekz:
Yes because the average crew member is expected to know the nuances of wing type.

Imo it'll be better to make them all valid or not valid vs this "pick and choose" type implementation.

I'm talking about it in a metagame aspect because now all of a sudden people are expected to know the difference in-charecter (at least that is what is implied by your reasoning)

JamieD1:
> Yes because the average crew member is expected to know the nuances of wing type.
> 
> Imo it'll be better to make them all valid or not valid vs this "pick and choose" type implementation.

Im happy to remove wings from shitty ipc martial art, its the only martial art that makes a physical difference, kinda dumb imo

Moltijoe:
> Im happy to remove wings from shitty ipc martial art, its the only martial art that makes a physical difference, kinda dumb imo

I initially didnt include them, but people begged me to because the wings are also an ultrakill reference

ynot01 (approved):

Hopekz:
> Im happy to remove wings from shitty ipc martial art, its the only martial art that makes a physical difference, kinda dumb imo

Your choice on what direction to go with in order to commit to one direction but it is an inconsistency ATM

Not going to pretend the wings aren't cool af though

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -113,6 +113,12 @@
 		if(100 to 200)
 			. += span_warning("[t_He] [t_is] twitching ever so slightly.")
 
+	if("wings" in dna.species.mutant_bodyparts)
+		var/hacked = ""
+		if(mind?.martial_art && istype(mind.martial_art, /datum/martial_art/ultra_violence))
+			hacked = ionnum() + ionnum() + " "
+		. += "[t_He] [t_has] also have a pair of [span_warning(hacked)][(dna.features["wings"])] wings on their back"
```
SomeguyManperson: Blood drenched wings perhaps? Hacked is something strange to say someone just sees

ynot01:
Ultra Violence has existed this long being a meta-questionable "I am valid" sign and I think it's clear that the martial art basically encourages that playstyle, so this will turn out fine I think. This may provoke validhunters, but in a perfect world, as long as the IPC is not attacking randoms or killing in broad daylight, randoms shouldn't be interfering.

Moltijoe:
> Ultra Violence has existed this long being a meta-questionable "I am valid" sign and I think it's clear that the martial art basically encourages that playstyle, so this will turn out fine I think. This may provoke validhunters, but in a perfect world, as long as the IPC is not attacking randoms or killing in broad daylight, randoms shouldn't be interfering.

if someone's validhunting, they probably don't care too much about meta in the first place. nice side effect is that lets IPCs get wings from other sources and not immediately be lynched

Moltijoe (approved):
no more trying to be stealthy with the bloodsoaked murderbone martial art

PR #17626, opened 2023-01-23
Chemical explosions now respect z-level bomb modifiers

ynot01:

Lavaland - 2.5x
Station - 1x
Reebe - 0.6x

# Changelog

:cl:  
tweak: Chemical explosions now respect z-level bomb modifiers
/:cl:


Cartlord:
This won't make the bomb cap affect them, right?

Moltijoe:
> This won't make the bomb cap affect them, right?

correct, they still bypass the cap, but are just dampened or strengthened based on the z-level

2.5x in lavaland, looking forward to drake deleting all of lavaland one round

Moltijoe (commented):
based, but the worry is that it also affects admin bombs, is that something that should be worried about?

Moltijoe (approved):
i suppose admins don't drop bombs often enough for it to matter, and the only reason they'd drop a bomb with ignore cap is if they're trying to blow shit up anyways

PR #17634, opened 2023-01-24
Security nerf: Mindshield firing pins no longer printable

ynot01:

Being able to print firing pins contributes to the printable-gun-crisis

Security must now work with their limited (10) roundstart firing pins, which creates a sense of scarcity and importance, and have to order more from cargo

Promotes security-cargo interaction

# Changelog

:cl:  
tweak: Security can no longer print mindshield firing pins
tweak: Firing pins crate now require security access instead of armory access to open
tweak: Mindshield firing pins are now available at cargo for 3000 credits
/:cl:

### DIFF
--- a/code/game/objects/items/storage/boxes.dm
+++ b/code/game/objects/items/storage/boxes.dm
@@ -670,6 +670,15 @@
 	for(var/i in 1 to 5)
 		new /obj/item/firing_pin(src)
 
+/obj/item/storage/box/secfiringpins
+	name = "box of mindshield firing pins"
+	desc = "A box full of mindshield firing pins, to allow newly-developed firearms to operate."
+	illustration = "id"
+
+/obj/item/storage/box/secfiringpins/PopulateContents()
+	for(var/i in 1 to 5)
+		new /obj/item/firing_pin/implant/mindshield(src)
+
 /obj/item/storage/box/lasertagpins
 	name = "box of laser tag firing pins"
 	desc = "A box full of laser tag firing pins, to allow newly-developed firearms to require wearing brightly coloured plastic armor before being able to be used."
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -419,9 +419,8 @@
 
 /datum/supply_pack/security/secfiringpins
 	name = "Mindshield Firing Pins Crate"
-	desc = "Upgrade your arsenal with 10 mindshield firing pins. Requires Armory access to open."
+	desc = "Upgrade your arsenal with 10 mindshield firing pins. Requires Security access to open."
 	cost = 2000
-	access_view = ACCESS_ARMORY // next to firing pins for ease of access
 	contains = list(/obj/item/storage/box/secfiringpins,
 					/obj/item/storage/box/secfiringpins)
 	crate_name = "firing pins crate"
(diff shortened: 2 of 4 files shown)


Moja10:
Not a bad idea, since mindshield pins are better for security than the regular pins, since unshielded plebs can't use the weapons. and God help us when sec starts replacing their guns roundstart pins with mindshield pins. maybe give sec 2 mindshield pins roundstart

MajManatee:
Initially I was against this, but it would help prevent officers unnecessarily upgrading their equipment, and stopping every antag from using it.

RG4ORDR:
This going to end up like the baton situation where the Captain,HOS, or Warden is going to strut into cargo and buy it.  There's not going to be any cooperation because any idiot can justify needing the firing pins as security. At best this is a soft removal which you can probably just gate behind more research,give them their own tree, or even have them locked to armory only. They cost diamonds as well, and eat into resources quicker than you'd think.

I only see cargo getting abused by this because it means you can just buy them instead of needing to get weapons tech research. As well as not have to worry about material cost.

00ze-cyclone:
> I only see cargo getting abused by this because it means you can just buy them instead of needing to get weapons tech research. As well as not have to worry about material cost.

The tech doesn’t matter because every R&D weapon except electronic weaponry is behind adv weaponry that unlock mindshield pins. Mindshield pins always have been available when you could print guns and the material cost is irrelevant, I **CONSTANTLY** see a stack of 10 mindshield pins on the armory and sec techfab made beforehand by some random seccie/captain/HoS.

Printable mindshield pins also make the pins in the brig just completely irrelevant, making it armory only wouldn’t matter because most R&D weapons are already armory only so if you can print an xray gun you can print a mindshield pin.

ynot01:
> This going to end up like the baton situation where the Captain,HOS, or Warden is going to strut into cargo and buy it. There's not going to be any cooperation because any idiot can justify needing the firing pins as security.

cargo being useful isn't why people storm in and do their job for them

if you want people to stop doing that then you need some incentives for them to not do so

as it stands there is great incentive to barge into cargo and order the crates yourself because relying on 5 hour cargo techs is slow as shit and inconsistent

on the same note people don't really enjoy playing cargo because
1. they don't get a lot of orders
2. they get stepped on when they become relevant

so if you find a way to mechanically make cargo techs useful then maybe people will start playing them again

Marmio64:
me, i'd like if mindshield pins also would have the option to be done like gunkits and take a normal pin and a R&D fabricated kit to upgrade.

JamieD1:
> > This going to end up like the baton situation where the Captain,HOS, or Warden is going to strut into cargo and buy it. There's not going to be any cooperation because any idiot can justify needing the firing pins as security.
> 
> cargo being useful isn't why people storm in and do their job for them
> 
> if you want people to stop doing that then you need some incentives for them to not do so
> 
> as it stands there is great incentive to barge into cargo and order the crates yourself because relying on 5 hour cargo techs is slow as shit and inconsistent
> 
> on the same note people don't really enjoy playing cargo because
> 
> 1. they don't get a lot of orders
> 2. they get stepped on when they become relevant
> 
> so if you find a way to mechanically make cargo techs useful then maybe people will start playing them again

Also maybe if the cargo people didnt just break into every crate and arm themselves

MajManatee (approved):

ReddicusDragon (approved):

PR #17648, opened 2023-01-25
The L6 is no longer just a better minigun

MajManatee:

Yeah yeah get to complaining

The L6 bullets have had a dmg reduction by 5 across the board
The L6 now fires in bursts of 2 instead of 3.

Currently, the L6 is literally the minigun but does more damage, and with no slowdown in exchange for a slightly smaller magazine. 
If its not enough can be toned down more.

I considered increasing the fire delay from 1 to 2, but decided instead for a burst reduction.

In an attempt to make it more appealing, the minigun slowdown has been reduced from 2 to 1.5

# Wiki Documentation

Dont know what section its under but probably needed updating in the nukies page or something.

# Changelog

:cl:  
tweak: L6 is no longer just a better minigun
tweak: Minigun slowdown decrease
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/automatic.dm
+++ b/code/modules/projectiles/guns/ballistic/automatic.dm
@@ -233,6 +233,7 @@
 	var/cover_open = FALSE
 	can_suppress = FALSE
 	fire_delay = 1
+	burst_size = 2
 	spread = 7
 	pin = /obj/item/firing_pin/implant/pindicate
 	bolt_type = BOLT_TYPE_OPEN
--- a/code/modules/projectiles/guns/ballistic/minigun.dm
+++ b/code/modules/projectiles/guns/ballistic/minigun.dm
@@ -109,7 +109,7 @@
 	lefthand_file = 'yogstation/icons/mob/inhands/weapons/minigun_inhand_left.dmi'
 	righthand_file = 'yogstation/icons/mob/inhands/weapons/minigun_inhand_right.dmi'
 	flags_1 = CONDUCT_1
-	slowdown = 2
+	slowdown = 1.5
 	slot_flags = null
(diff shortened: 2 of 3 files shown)


MajManatee:
The minigun does 36 dmg per bullet btw

Cleverslim:
this is fucking stupid the problem is not the l6 saw its the fact we have 2 weapons that fill the same role as a machine gun aka remove one (the minigun)

TheBombMakerJL:
1875F303-52DF-4ACD-A839-8BDECCE77B22

This reduces damage output of the L6 by 43%. Just because you died to the L6 once without any headgear protection doesn’t mean it needs a nerf. Buff the mini gun so it’s used more often, don’t make a trash ided pr and make one of the few nukie weapons that are actually viable less so

Moltijoe:
easy, nerf the L6 and buff everything else.
Large scale balance pass on syndicate guns.
shouldn't take forever to do.

Marmio64:
LMG could perhaps do with a small nerf, either the damage or the burst, but not both. It's more of the second primary that is closest to it's level, the m90gl, could do with a buff to make it more competitive. The c20r is quite a bit worse than either but it is also super damn cheap yet still gets the job done so it's more of a reinforcement's weapon.

TheRyeGuyWhoWillNowDie:
the l6 isn't too good (takes two hands to fire preventing you from getting block chance, you basically fire telecrystals every time you click with how fucking expensive the ammo is, there's no way to store the damn thing) it's just that the minigun, which requires your back slot, therefore preventing you from taking like anything else with you AND slows you down AND overheats is really garbage, plus this doesn't adjust the price of the l6 so now you're spending like sixty telecrystals for a cumbersome mgl90

Bibby0110 (approved):

PR #17655, opened 2023-01-26
remove kinetic accelerator recoil

TheRyeGuyWhoWillNowDie:
who merged this garbage

:cl:  
tweak: kinetic accelerators have no recoil
/:cl:


ToasterBiome (changes requested):
remove the inaccuracy too

ynot01:
-1 miners should be dusted on spawn and not allowed to take any ghost roles

MajManatee:
I think because all guns were given recoil, it also applied to the KA.

ToasterBiome (approved):

PR #17531, opened 2023-01-16
kills grand core for new item called cursed cane

LazennG:
# Document the changes in your pull request

introduces a new item called 'cursed cane,' a small object which is kinda like the grand core but not hastily made to replace something extremely stupid/op. Using the cane inhand summons small friendly legion skulls that scale in number to the limit of the cane which starts at 3 on a 6 second cooldown. Hitting a target below 20% hp will kill them and replace them with a legion that's friendly to anyone who has held the cane, also following the same limit rule as the skulls. The legion and legion skulls only bother attacking targets who have more than 10% of their hp left. You can also have a skull near a corpse or hit it with a cane to revive it as a minion but it can only be done once before they become invalid for that kind of revival again. The skulls do 12 damage to lavaland fauna and 2 otherwise.
   Finishing off a megafauna needs them to be below 10%hp and will increase the limit by 1. Doing this will kill the megafauna and make a bigger minion called an attendant which is pretty much a legionnaire post-tumor sharding. It's got 200hp and does 15 damage while also being friendly, only ramping up to 30 damage if the target is a lavaland fauna. Additionally, the attendant is rideable for the sake of crossing lava and unlike the other mobs it doesn't stop attacking if they're about to die.
   All the minions bigger than tiny skulls can be called to the user on a 30 second cooldown.
   People can also be finished off but they'd need to have at most 13hp remaining before they crit, and when they're finished they get sealed inside a legion and barely put into crit.
   Lastly, the cane can teleport the user to wherever it's thrown as long as it hits something and the distance from the thrower to the endpoint isn't further than 10 spaces. If the cane hits a target valid for finishing off, it gets put in the user's hand.

oh also the minions are just legions and legionnaire but painted purple

# Spriting
image

# Wiki Documentation
https://wiki.yogstation.net/wiki/Megafauna#Loot_4
replace the line with the grand core with something like

Drops a cursed cane image which, while doing no damage on its own, can finish off dying targets to make minions. Finishing off stronger fauna will in turn make stronger minions. Minions made from corpses cannot be revived again. Additionally, the cane will teleport the user to it if thrown as long as it hits something.

# Changelog

:cl: @Chubbygummibear 
rscadd: Adds cursed cane
rscdel: Removed grand core
imageadd: added cane sprites
/:cl:
big thanks to @Chubbygummibear for coding help

PR #17658, opened 2023-01-26
I dont think this is crashing it, but lets try anyway

JamieD1:
Reverts yogstation13/Yogstation#17531


ThatLing (approved):

MajManatee:
jamie just HATES when miners get a new thing to play with, so he prematurely removes it before any fun can be had smh

PR #17660, opened 2023-01-26
cursed cane shouldn't crash the server anymore

LazennG:

cane now checks to see if more minions can be added to the list of them sooner so no more infinite production via corpses
alternative to #17658 
also now aides are summoned near the user instead of right on them
# Changelog

:cl:  
bugfix: fixed cane crashing the server probably
tweak: aides no longer spawn right on you when summoned
/:cl:


MajManatee (approved):
idk if this is the solution but its still good fixes.

PR #17663, opened 2023-01-26
Science Security & HoS can now enter robotics & xeno

ynot01:

Gives sci officers & HoS robotics and xenobio

#

The scientist officer exists for a few reasons

1. Look after robotics and make sure they aren't building giga mechs
2. Look after toxins and make sure they aren't destroying the station
3. Look after xenobiology and make sure they aren't idk powergaming too hard

HOWEVER they don't have access to robotics or xenobiology, so they can't actually do anything besides yell and bang on the door (or ask the captain, RD, or AI to open it, if there is any and they are cooperative), especially if the shutters are closed and they can't see if anything is happening inside.

#

Problem: Officers will start making mechs and blowing borgs

Solution: Only let them use the airlocks, which is implemented here

ACCESS_ROBOTICS is now ACCESS_ROBO_CONTROL

A new ACCESS_ROBOTICS was created that only gives access to doors and buttons

# Changelog

:cl:  
tweak: The security department has been granted robotics and xenobio access
/:cl:


ToasterBiome (changes requested):
the way that this works is backwards. Make the generic robo access give you access to robo (inside) and then the new access splits off for things like blowing borgs and making mechs. Yes this is slightly more involved with work but that’s the way we have it already with science access and actually researching

MajManatee:
Willing to give them xenobio access too? Or are we leaving them screwed over there?

ynot01:
Xenobio access not needed, as an officer it is very low priority and dangerous situations are less urgent and distinct as a robo making a combat mech. The point of having robo access is a preventative measure to stop the roboticist before they finish their ultradeath machine. In xenobio, it will almost always be "wtf" but that is their job and you can't bust down their door and make arrests because that is the norm. You will, most of the time, only know a xenobiologist is bad when it is too late.

MajManatee:
Ill remember this when im listening to someone scream "Help xenobio" for the morbillionth time and i cant get in

JamieD1:
> Xenobio access not needed, as an officer it is very low priority and dangerous situations are less urgent and distinct as a robo making a combat mech. The point of having robo access is a preventative measure to stop the roboticist before they finish their ultradeath machine. In xenobio, it will almost always be "wtf" but that is their job and you can't bust down their door and make arrests because that is the norm. You will, most of the time, only know a xenobiologist is bad when it is too late.

Probably useful to stop them releasing idiot monsters or aliens or helping kill slimes that escaped

code/game/mecha/mech_fabricator.dm
```diff
@@ -8,7 +8,7 @@
 	idle_power_usage = 20
 	active_power_usage = 5000
 	
-	req_access = list(ACCESS_ROBOTICS)
+	req_access = list(ACCESS_ROBO_CONTROL)
```
Maia-J, line 11: This does nothing, you need to re-add the lines removed from ui_interact some time back.

	if(!allowed(user) && !combat_parts_allowed && !isobserver(user) && !hacked)
		to_chat(user, span_warning("You do not have the proper credentials to operate this device!"))
		return

code/game/mecha/mecha_control_console.dm
```diff
@@ -3,7 +3,7 @@
 	desc = "Used to remotely locate or lockdown exosuits."
 	icon_screen = "mecha"
 	icon_keyboard = "tech_key"
-	req_access = list(ACCESS_ROBOTICS)
+	req_access = list(ACCESS_ROBO_CONTROL)
```
Maia-J, line 6: This probably does nothing too for about the same reason, but unlike the mechfab I didn't test it and am too lazy to launch my testing server again

MajManatee (approved):
idk what biome is on about. Seems like if sec is printing mechs then thats an admin issue to me.

Maia-J:
> idk what biome is on about. Seems like if sec is printing mechs then thats an admin issue to me.

Not that this disalows them from doing it so I guess its fine

ToasterBiome (approved):
maps fine

ReddicusDragon:
coder and mapping approved, merging this

PR #17672, opened 2023-01-27
Medibot will now accurately insult medbay for having dead patients

ynot01:

What kind of medbay is this? Everyone's dropping like flies.

will only trigger if medibot sees anyone that is dead

# Changelog

:cl:  
tweak: Medibot will only say that everyone is dropping like flies if they are dropping like flies
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/medbot.dm
+++ b/code/modules/mob/living/simple_animal/bot/medbot.dm
@@ -390,7 +390,15 @@
 				var/list/i_need_scissors = list('sound/voice/medbot/fuck_you.ogg', 'sound/voice/medbot/turn_off.ogg', 'sound/voice/medbot/im_different.ogg', 'sound/voice/medbot/close.ogg', 'sound/voice/medbot/shindemashou.ogg')
 				playsound(src, pick(i_need_scissors), 70)
 			else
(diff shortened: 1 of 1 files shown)


Moltijoe (commented):
instead of doing math and then overwriting the list, why not just add it to the list?

ynot01:
> instead of doing math and then overwriting the list, why not just add it to the list?

just add it to the list..? like how i took it out of the list?? what are you asking here

Moltijoe:
> just add it to the list..? like how i took it out of the list?? what are you asking here

as in, if someone is dead, just add the new sound to the list with +=

ynot01:
> > just add it to the list..? like how i took it out of the list?? what are you asking here
> 
> as in, if someone is dead, just add the new sound to the list with +=

This is an option but by using the current method we don't have run the loop every time, only when it would theoretically roll for that message

Moltijoe:
> This is an option but by using the current method we don't have run the loop every time, only when it would theoretically roll for that message

huh, yeah, actually i didn't think about that
not sure if the performance impact will be even close to noticeable, but it definitely looks like an improvement

MajManatee (approved):
Finally, this game breaking bug is fixed

PR #17673, opened 2023-01-27
PDA bombs are now guaranteed to break PDA screens

ynot01:

Could also break (not destroy) the hard drive, if the target is not in a computer, I suppose

# Changelog

:cl:  
tweak: PDA bombs are now guaranteed to break PDA screens
/:cl:


MajManatee (approved):
Im not gonna sugarcoat it
explosion(source, -1, 0, 3, 4)

PR #17677, opened 2023-01-28
Preternis water damage tweak

Moltijoe:
Currently it does too little damage when applied in small amounts, making less "standard" forms of wetness almost useless

When reworking preternis I wavered between flat and scaling damage
I for some reason never considered doing both
This will make small splashes do more damage, but large splashes will still do the same amount over it's duration

:cl:  
tweak: Preternis take more water damage at small amounts, but less at high amounts
/:cl:


UltraBigRat:
water bottle based warfare, nice.

MajManatee (approved):
me locking and loading my 5 dollar squirt gun for my assassinate sarah objective.

PR #17678, opened 2023-01-28
Nanite Heart no longer helps itself be deleted

Moltijoe:
A person already loses nanites when emp'd
Nanite hearts made you lose extra nanites when emp'd
This often resulted in all nanites being removed instantly deleting your heart

Literally the only time anyone used it is in conjunction with vein muscle membrane which removed the downside of your heart getting deleted

Your heart will still be stopped (it already did this but no one knew because their heart was instead deleted)

this probably won't make it good enough to be worth using, but it's something

:cl:  
tweak: Nanite heart no longer makes it's owner lose additional nanites from an emp
/:cl:


MajManatee (approved):
yeah thats kinda goofy.

PR #17694, opened 2023-01-29
Current Station Alerts no longer pop up for non-silicons

ynot01:

Current Station Alerts

# Changelog

:cl:  
bugfix: fixed Current Station Alerts popping up for non silicons
/:cl:


MajManatee (approved):
If this means what i think this means, that would be epic.

PR #17695, opened 2023-01-29
Syndicate bombs now always detonate when exploded and active as opposed to sometimes

ynot01:

Follow-up to #17168

When blowing up lightly directly on top of it I could get it to detonate but when trying explosions next to it it wouldn't detonate and when I dev exploded it it just deleted itself

Now it will always try to detonate and will succeed if the bomb is counting down

# Changelog

:cl:  
tweak: Syndicate Bombs will now detonate always if counting down and exploded
/:cl:

### DIFF
--- a/code/game/machinery/syndicatebomb.dm
+++ b/code/game/machinery/syndicatebomb.dm
@@ -38,6 +38,10 @@
 	if(.)
 		payload.detonate()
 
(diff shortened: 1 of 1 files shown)


MajManatee (approved):
didnt know there was a chance it DIDNT explode

PR #17713, opened 2023-01-31
You can now wag an implanted tail

ynot01:

Used to check if the species started with a tail and didn't actually care if you had a tail or not so it also would silently fail if you were missing your tail

# Changelog

:cl:  
bugfix: You can now wag an implanted tail
/:cl:


MajManatee (approved):
the PR of a man in denial.
Also surely theres a better way to do it than making a case for every type of tail?

ShoziOffical:
Fetish

ynot01:
> Also surely theres a better way to do it than making a case for every type of tail?

Can't make it apply to every tail as there are nonwagging tails (i.e. polysmorph tail)

MajManatee:
clearly the only option is to let polysmorphs wag their tails

MajManatee:
or something like 
var/waggable = true
and on polysmorphs set it to false

ChuChu42:
ynot's fursona is a goat and everybody knows that

PR #17714, opened 2023-01-31
Heretic runes will no longer eat more items than is required to transmutate

ynot01:

Eats 3 knives, makes 1 blade

Now eats 1 knife, makes 1 blade and can be done 3 times if you have 3 knives

# Changelog

:cl:  
bugfix: Heretic runes will no longer eat more items than is required to transmutate
/:cl:


MajManatee (approved):
id rather that, if given 3 knives, it makes 3 blades so i dont have to spam click and deafen myself. But this is better than current.

PR #17715, opened 2023-01-31
Gem buffs and Gem Modularization

GraveHat:
Worth more now drop more gem clean up gem

- All gems point values are doubled, and minor gems increased to 500
- Increased value of Diamond and Auric gem bounty by 2000 credits, reduced amount of gems necessary to complete bounty by 1 gem from 3
- Aligns Bluespace Data Crystal with the rest of the gems as it is technically not a gem
- Increases the likelihood of Gold Grubs, Ice, and Magmawing watchers dropping their respective gems from 10% to 75%
- Removes redundant analyzed code

After all this time I noticed that it isn't really worth it to haul all the gems back to cargo from Lavaland and that you never really get to see rupees, frost diamonds, or calcified auric even if they do drop because of the rarity of the mobs. 

This has been lightly tested and no glaring bugs have been detected.

# Modularization WIP

- Modularizing point value system so that new gems don't need a point_value var dumped at the end of it every time maybe
- Get welding feature to function
- Fix selling bounties and receiving points from ID

# Wiki Documentation

Stat changes above

# Changelog

:cl:  
bugfix: Aligns Bluespace Data Crystal with the rest of the gems as it is technically not a gem 
bugfix: Removes redundant analyzed code
tweak: All gems point values are doubled, and minor gems increased to 500
tweak: Increased value of Diamond and Auric gem bounty by 2000 credits, reduced amount of gems necessary to complete bounty by 1 gem from 3
tweak: Increases the likelihood of Gold Grubs, Ice, and Magmawing watchers dropping their respective gems from 10% to 75%
experimental: Modularization
/:cl:

### DIFF
--- a/code/datums/components/gems.dm
+++ b/code/datums/components/gems.dm
@@ -1,0 +1,63 @@
+/// Components to list other things as gems so it would have sweeping changes across all gems
+/datum/component/gems
+	///owning ID, used to give points when sold
+	var/obj/item/card/id/claimed_by = null
+	///How many points we grant to whoever discovers us
+	var/point_value = 100
+	///what's our real name that will show upon discovery? null to do nothing
+	var/true_name
+	///the thing that spawns in the item.
+	var/sheet_type = null
+	//shows this overlay when not claimed
+	var/image/shine_overlay
+
+/datum/component/gems/Initialize()
+	. = ..()
+	///the thing that makes it a parent 
+	var/atom/parent = src.parent
+
+	RegisterSignal(parent, COMSIG_PARENT_EXAMINE, .proc/examine)
+	RegisterSignal(parent, COMSIG_PARENT_ATTACKBY, .proc/attackby)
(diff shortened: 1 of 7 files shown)


MajManatee (approved):
I didnt know those gems even existed.
Anyways yeah they are pretty much never worth your time to grab.
Hope this can change that now that its 1-1 for gem to survival pen.

ToasterBiome (changes requested):
don't restate all the gem code just for the stalwart thing. This situation calls for a 🤢 (allah forgive me) component.

move the gem code to it's own component, along with the "selling" looking for this component instead of a gem. Gems will simply be fancy items that do nothing except get a component added to them on initialize. You can then also add the component to the ai_cpu, so it's all in one place, and in the far future when all of us are gone, people don't have to change two places when they want to buff gems even more.

MajManatee:
Over a week of no activity, Please re-open when work continutes

PR #17729, opened 2023-02-01
Adds more info to cloaker belt

ynot01:

Cloaker belt now gives you information about itself and your environment

Also examine tells you that it only charges while turned on

# Changelog

:cl:  
rscadd: Added some info to cloaker belt
/:cl:


MajManatee (approved):
looks probably fine

PR #17730, opened 2023-02-01
adds a admin only pacman disease and a admin only reagent that can cause this disease

boodaliboo:
added pacman disease
:cl:  
rscadd: Added pacman reagent 
/:cl:


Bibby0110:
qAh yes the classic make the pr again when it's another headcoder

This was previously denied for being bloat

PR #17760, opened 2023-02-02
Fixes tgui bluescreen

adamsong:
TGUI errors now direct players to our github, instead of tg's github.

# Changelog

:cl:  
bugfix: TGUI bluescreens no longer point to a place where help is not given
/:cl:


MajManatee (approved):
I was much more excited for a second when i thought you just fixed bluescreening

PR #17613, opened 2023-01-22
Bluespace Locker can now also be a crate

JamieD1:
Bluespace Locker can now also be a crate,

Makes it harder/easier to find the locker, also people have to search crates now, haha.

:cl:  
rscadd: Bluespace Locker can now also be a crate
/:cl:

PR #17761, opened 2023-02-02
Revert "Bluespace Locker can now also be a crate"

Cartlord:
Reverts yogstation13/Yogstation#17613 because I think it's a stupid idea, the locker's power has already been reduced and I think that making it possibly be ANY CRATE when it's already hard to find just makes it completely unreasonable for anyone to find it - especially because, frankly, I don't think anyone enjoys climbing onto crates, just to close them and check if it's the bluespace crate.
Since the locker now has to have an energy supply brought to it, I think that ALSO making it rarer is too severe a nerf.

### DIFF
--- a/yogstation/code/controllers/subsystem/bluespace_locker.dm
+++ b/yogstation/code/controllers/subsystem/bluespace_locker.dm
@@ -16,7 +16,7 @@ SUBSYSTEM_DEF(bluespace_locker)
 	// basically any normal-looking locker that isn't a secure one
 	var/list/valid_lockers = typecacheof(typesof(/obj/structure/closet) - typesof(/obj/structure/closet/body_bag)\
 	- typesof(/obj/structure/closet/secure_closet) - typesof(/obj/structure/closet/cabinet)\
-	- typesof(/obj/structure/closet/cardboard)\
+	- typesof(/obj/structure/closet/cardboard) - typesof(/obj/structure/closet/crate)\
 	- typesof(/obj/structure/closet/supplypod) - typesof(/obj/structure/closet/stasis)\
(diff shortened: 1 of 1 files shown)


ChuChu42:
another issue: bluespace CRATE is exepctionally buggy and very hard to get out/in, it can sometimes glitch making the crate non collideable to mobs regardless of being open or not, can also sometimes delete mobs and objects

agree on reverting

MajManatee:
revert the power change instead

Moja10:
cargo likes to mail their roundstart crates, god forbid a cargo tech doesn't spend 2 minutes crawling into every one of them to see if it's bluespac3

ThatLing:
> cargo likes to mail their roundstart crates, god forbid a cargo tech doesn't spend 2 minutes crawling into every one of them to see if it's bluespac3

The locker gets re-rolled when destroyed, this isn't an issue

Cartlord:
Maj, I TRIED to revert that already, and it was denied - apparently there were talks among the balance council, and they agreed that they had a direction in mind for the Bluespace locker that involved making it not be self-powered.

Bibby0110:
We've discussed this internally once again, and still agree that diluting the chance of finding the locker by adding crates to the mix is desired.
The mainstay of the maintainer team agrees that the bluespace locker is meant to be a rare phenomenon, that might not happen even if someone purposefully looks for it.

Cartlord:
I recognize I'm unlikely to receive an answer, but I'd still like to ask: what DO you want the bluespace locker to be? You've reduced the chance of it spawning while also making it not able to generate its own power, thus lowering the chance of _and_ point to finding it - and it seems like it wasn't in a bad spot, at all, before those changes were made.

PR #17766, opened 2023-02-03
L6 re-tweak (we're rolling it back (kinda))

Skrem7:
# Document the changes in your pull request

I should blindly nerf a weapon with zero consideration

BASICALLY makes this thing lethal again but armor should ACTUALLY help against AP (not a one-shot but still a LOT better against armor) (you're basically dead but you don't eat an instant crit unless you have a funky brute mod in which case you were going to die during nukies anyway)

also makes it more inaccurate because you're fucking running around two-handing a high-caliber LMG I would do something about how you could take time to set up a bipod to have accurate fire but I'm lazy and this will do for now

I should probably make L6 AP cheaper but I don't care to rn

Closes https://github.com/yogstation13/Yogstation/pull/17771

# Wiki Documentation

Damage values and inaccuracy should be noted in Guide to Combat

# Changelog

:cl:  
tweak: L6 has three round burst again
tweak: All L6 bullet damage has been increased by 5
tweak: L6 AP has 45 AP, down from 60
tweak: L6 has 14 spread, up from 7
/:cl:

PR #17771, opened 2023-02-03
L6 Tweak but its not just literally a revert.

MajManatee:

Alternative to #17766
Should bibby have merged the other PR I made? Definitely not.
Does that mean there was bad ideas there? Also no, i stand by my changes.
Instead of reverting everything and calling it a tweak, actually has trade offs

Gives the L6 back its 3 round burst
Gives the L6 a longer delay between shots because how are you holding that MF on target
Also adds a bit of spread which should REALLY only be noticeable if you are blind firing down a hallway.
Takes away some AP from the AP mag because 60 is kinda absurd yeah
Slightly reduces the cost of the gun and its ammo to compensate because you are all big babies

# Changelog

:cl:  
tweak: L6 3 round burst is back
tweak: Fire delay increased by 0.6
tweak: Slight increase of spread from 7 to 12
tweak: AP reduction on AP mag (nobody buys this anyway)
tweak: Gun cost from 20TC to 18
tweak: Ammo cost from 6 to 5 per box (9 to 8 for AP)
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/automatic.dm
+++ b/code/modules/projectiles/guns/ballistic/automatic.dm
@@ -234,7 +234,7 @@
 	can_suppress = FALSE
 	fire_delay = 1.6
 	burst_size = 3
-	spread = 14
+	spread = 12
 	pin = /obj/item/firing_pin/implant/pindicate
 	bolt_type = BOLT_TYPE_OPEN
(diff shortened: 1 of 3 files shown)


Skrem7:
Effectively a revert PR.

PR #17774, opened 2023-02-03
blue burger uses patty not steak

ktlwjec0:
all the other ones use patties, think the blue burger was just missed. no wiki changes.

:cl:  Ktlwjec
tweak: Blue burger is crafted with a patty instead of a steak.
/:cl:


Moltijoe (approved):
burgr

PR #17784, opened 2023-02-04
True semi-auto shotguns + racking medium/heavy ballistics requires both hands + removes outdated stechkin animations

Skrem7:

The first part is pretty simple; adjusts the code for all semi-auto shotguns to actually exist as semi-auto guns, with a charging handle and all. A simple set of new sprites has been added to give feedback to whether the chamber is open or not (if the gun is out), except for the cycler because it didn't look like it could use one + it's admin-only so I don't really support its existence (its code is terrible)

The second part adds a check to the rack() action specifically (NOT the drop_bolt() action) so that it can't be done while your off-hand is full for any ballistic that cannot be dual-wielded. This means that cycling the chamber while it has a spent or live round in it cannot be done with two hands (primarily targeting shotguns, here). I didn't want to have it hit weight_light firearms such as the stechkin even if it's a little non-sensical how you can rack it with your other hand occupied

As a result, the riot shotgun is no longer weapon_heavy, but you just won't be able to rack it while your off-hand is filled.

Also removes the stechkin animation sprites that are outdated, since we have a new sprite for the gun, and the animations not only don't fit but also slow down the performance of the gun (lumped into this PR to avoid cornflakes).

# Spriting

Combat Shotgun
Original:
image
Chamber open:
image

Compact Combat Shotgun
Original:
image
Chamber open:
image

Breaching Shotgun
Original:
image
Chamber open:
image

# Wiki Documentation

I can go ahead and update the descriptions for semi-auto shotguns + ballistics with how they rack now on the Guide for Combat

# Changelog

:cl:  
tweak: Semi-auto shotguns are now actually semi-auto
imageadd: Adds bolt_state icons for semi-auto shotguns
tweak: Racking a ballistic (interacting with it in-hand while a round is chambered, be it live or not) of medium or heavy size requires your off-hand to be empty
tweak: Riot shotgun is no longer weapon_heavy
imagedel: Removes outdated stechkin animation sprites
/:cl:


code/modules/projectiles/guns/ballistic.dm
```diff
@@ -257,6 +257,10 @@
 /obj/item/gun/ballistic/proc/rack(mob/user = null)
 	if (bolt_type == BOLT_TYPE_NO_BOLT) //If there's no bolt, nothing to rack
 		return
+	if (weapon_weight != WEAPON_LIGHT) //Can't rack it if the weapon doesn't permit dual-wielding and your off-hand is full
+		if (user.get_inactive_held_item())
+			to_chat(user, span_notice("You cannot rack the [bolt_wording] of \the [src] while your off-hand is full!"))
```
tattax:
```suggestion
	if (weapon_weight != WEAPON_LIGHT)
		if (user.get_inactive_held_item()) //Can't rack it if the weapon doesn't permit dual-wielding and your off-hand is full
			to_chat(user, span_warning("You cannot rack the [bolt_wording] of \the [src] while your off-hand is full!"))
```
i'd rather have it say other hand instead of off hand but you do you

tattax: also does the text in the to chat come out right

tattax:
idea: monkey has tail therefore can reload with tail ???? (purely scientific)

Skrem7:
> idea: monkey has tail therefore can reload with tail ???? (purely scientific)
> 
> 

che

code/modules/projectiles/guns/ballistic.dm
Skrem7: yes

Skrem7: > i'd rather have it say other hand instead of off hand but you do you

yeah sure I'll get it done later probably

cuackles (approved):

ReddicusDragon (approved):

PR #17796, opened 2023-02-05
Updates inventory procs to allow silent item pickups

ReddicusDragon:
now when calling put_in_hands, put_in_active_hand, or put_in_hand you can use `no_sound = TRUE` to make the pickup noise silent.

### DIFF
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -226,7 +226,7 @@
 //Puts the item our active hand if possible. Failing that it tries other hands. Returns TRUE on success.
 //If both fail it drops it on the floor and returns FALSE.
 //This is probably the main one you need to know :)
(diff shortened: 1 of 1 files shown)


Hopekz:
I think this is backwards normally control variables are "if yes then something happens" vs "if yes then it doesn't happen"

So the normal way around would be to make a "does it make sound" variable that is set to true by default and can be set to false

ReddicusDragon:
> I think this is backwards normally control variables are "if yes then something happens" vs "if yes then it doesn't happen"
> 
> So the normal way around would be to make a "does it make sound" variable that is set to true by default and can be set to false 

I did think of that, however it was easier to use pre-existing code. The equipped proc of an item has a parameter called initial that, when set to false, allows items to make noise. Hence why our new parameters are called no_sound and default to false.

PR #17799, opened 2023-02-05
Bloodsucker tasks will no longer lie to you about their requirements

ynot01:

A task is generated. It tells you "Go suck 123 units of blood." You would expect `123` to get stored somewhere, right? It didn't. It would just generate a random number each time you checked based on your bloodsucker rank and compare that against how much task blood you had gotten. This means you could be told to suck 190 units of blood and only need to suck 110, or vice versa.

Additionally, if you had dared ranked up in the time since you got the task, you could be in a whole new bracket, so instead of 100-200 blood, it would roll for 200-300 blood, but still tell you that you needed to suck 123 units of blood to complete it.

Coded by chimps

# Changelog

:cl:  
bugfix: fixed bloodsucker blood altar tasks randomly generating themselves at will
/:cl:


MajManatee (approved):
ooh ee oh che ah!

PR #17804, opened 2023-02-06
Adds box of shotgun darts to the uplink for 2TC

ynot01:

Stealthy item, disguised as beanbag slugs in a beanbag slug box

Cannot be examined for reagents

Does not drop syringe when it hits a non-carbon like normal darts

People will still figure out the syringe in their arm though

# Changelog

:cl:  
rscadd: Added box of shotgun darts disguised as beanbag slugs to the uplink for 2TC
/:cl:


MajManatee (approved):
ya sure

PR #17808, opened 2023-02-06
Rechargeable hardlight shotgun and shells (for normal shotguns)

MajManatee:

Since according to certain admins, using rubber ammo is considered "Power gaming" and "Over escalating", we need a NEW contender in the world of less than lethal armaments.

Adds two new ammo types for sec to play with. Hardlight buckshot and slugs.

The buckshot is just... rubber shot basically, without the wound bonus and sharpness. (for now, unless I get a good idea)
The slug fires TWO slugs at once, each doing half the total damage, and in a tight arc (haven't tested it yet, but it should be cool), and does a bit more than a beanbag.

This PR ALSO now adds a hardlight shotgun, Using rechargeable magazines and able to switch between buckshot and slugs on a whim. Each mag holds a max of 8 rounds, and can be recharged same as any other.

Ideally id put one in the armory locker with the riot shotguns, and another in the same place with the disablers and the USPs.
They will be orderable from cargo probably.

They are all locked behind the newfound hardlight tech, and require glass in addition to metal to print.

# Spriting

TBD

# Wiki Documentation

Sprites to be added. 
Important info:
Buckshot is basically just slightly worse rubber shot with a total of 72 stamina as opposed to 78
Slugs are actually TWO slugs in a tight arc, and do 30 stam each, for a total of 5 more than beanbags (a total of 55)

also TBD when im a bit closer to being finished

# Changelog

:cl:  
experimental: HARDLIGHT SUPREMACY!!
rscadd: Hard light Buckshot and Slug, Printable with hard light tech from armory lathe.
/:cl:


Moltijoe:
new hardlight tech is always nice, will take a look at the actual code in a bit

MajManatee:
linter is mad because of my sprites being noexistant

cuackles:
Good PR but only if you rename the laser pointers to soft light pointers

Rdrazga:
Like the idea of some more hardlight stuff but would think this would make more sense as it’s own shotgun with a rechargeable pack like the pistol, rather then just another shell for regular shotgun that makes shotgun fill every single role.

MajManatee:
Scope is expanding, hang on

cuackles:
post sprites please

ynot01 (changes requested):
Balance council says no

Some things to be salvaged in this or another PR:
- Hardlight (magfed) shotguns can be moved to ERT
- Dual shot slugs are a very cool concept, worth keeping and adjusting for normal shotgun use

PR #17809, opened 2023-02-06
Nanite heart processes nanites on life tick

Moltijoe:
Nanite heart is still pretty dangerous to have, rather than reducing the downside, this makes the upside more enticing

Nanites tick more often than life, so this is only ~30% increase to tick frequency
Making it run nanites on every life tick means it will boost healing, and generation of nanites, but will also increase consumption and the speed of any negative program that the nanites might have.

:cl:  
tweak: Nanite heart runs your nanites on each life tick
/:cl:

### DIFF
--- a/yogstation/code/modules/surgery/organs/heart.dm
+++ b/yogstation/code/modules/surgery/organs/heart.dm
@@ -13,6 +13,11 @@
 
 /obj/item/organ/heart/nanite/on_life()
 	. = ..()
+	if(owner)//makes nanites tick on every life tick, only about 33% speed increase
+		var/datum/component/nan = owner.GetExactComponent(/datum/component/nanites)
+		if(nan)
+			nan.process()
+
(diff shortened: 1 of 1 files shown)


iamthesilly:
this might almost make me consider using it for once

MajManatee:
Yeah I'm still never going to get nanites. Basically a death sentence.

Maybe if the consoles were ID locked, or required special access then I would, but right now you are rolling the dice on instant death every time.

Moltijoe:
> Yeah I'm still never going to get nanites. Basically a death sentence.
> 
> Maybe if the consoles were ID locked, or required special access then I would, but right now you are rolling the dice on instant death every time.

Id locking is kinda cringe, giving nanites their own room that's a bit less accessible might be something at least.
Something worth considering is making the weaponized nanite programs less of an insta-kill and more of a "oh god oh fuck, get these out of me"
Maybe making a chemical that can purge nanites is worth considering too

JamieD1 (approved):

PR #17810, opened 2023-02-06
Nerfs Preternis because I want to be the only one that plays them

Moltijoe:
I just played a round where there were FOUR preterni, this is UNACCEPTABLE as everyone knows this is my personal snowflake species that no one else should play.

Jokes aside
I initially went and overtuned them a bit initially, just to get a feeling of how they'd play and perhaps get actual feedback rather than no one playing them still.
Hoped they wouldn't be needing of a bit of a nerf, seems I was wrong.
The downside of attacking slower doesn't actually feel noticeable, so i've increased it by 0.1 (might increase it by another 0.1 in the future)
Also, reduces leg health, but increases arm health. makes them easier to bring to the ground, but not easier to nugget.

:cl:  
tweak: Preterni attack 0.1 second slower
tweak: Preterni arms have 5 more health (40)
tweak: Preterni legs have 5 less health (30)
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/preternis.dm
+++ b/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/preternis.dm
@@ -26,7 +26,7 @@ adjust_charge - take a positive or negative value to adjust the charge level
 	burnmod = 1.1 //The plasteel has a really high heat capacity, however, if the heat does get through it will REALLY burn the flesh on the inside
 	coldmod = 3 //The plasteel around them saps their body heat quickly if it gets cold
 	heatmod = 2 //Once the heat gets through it's gonna BURN
-	tempmod = 0.1 //The high heat capacity of the plasteel makes it take far longer to heat up or cool down
+	tempmod = 0.15 //The high heat capacity of the plasteel makes it take far longer to heat up or cool down
 	stunmod = 1.1 //Big metal body has difficulty getting back up if it falls down
 	staminamod = 1.1 //Big metal body has difficulty holding it's weight if it gets tired
 	action_speed_coefficient = 0.9 //worker drone do the fast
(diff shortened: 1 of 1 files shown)


MajManatee (approved):
Lizard supremacy

Moggram:
>
>>I just played a round where there were FOUR preterni, this is UNACCEPTABLE as everyone knows this is my personal snowflake species that no one else should play.

This is why I miss my old glass cannon robo man who has hides in the dark to live.

TheBombMakerJL:
why do preternis need an attack slowdown

Moltijoe:
> why do preternis need an attack slowdown

they're a species designed for work, not combat
think of it like "big attack throws them off balance and they take longer to readjust"

TheBombMakerJL:
> > why do preternis need an attack slowdown
> 
> they're a species designed for work, not combat think of it like "big attack throws them off balance and they take longer to readjust"

Maybe have the preternis martial art remove the slowdown?

Moltijoe:
> Maybe have the preternis martial art remove the slowdown?

They already had a slowdown, this just makes it stronger, that you didn't even notice it proves my point.
any martial art removes it

Uristthedorf:
I think the attack slowdown probably matters most when trying to get through something like a window or door, where there's no running around time between hits to buffer your attacks and you're attacking the door quite a lot before it goes down.

iamthesilly:
> I think the attack slowdown probably matters most when trying to get through something like a window or door, where there's no running around time between hits to buffer your attacks and you're attacking the door quite a lot before it goes down.

i agree with this, attack speed matters more for breaking things than actual combat

Moltijoe:
> i agree with this, attack speed matters more for breaking things than actual combat

true, but i also find it can play into finishing someone off if they're down.
Considering they have access to the hammer, limiting that seems to be for the best

iamthesilly:
> true, but i also find it can play into finishing someone off if they're down.

If you're finishing someone off they're probably going to die anyway and won't be fighting back

iamthesilly:
the legs only having 5 more max damage than surplus prosthetics seems way too low, i think 30 for the legs and 40 for the arms would be better

Moltijoe:
> the legs only having 5 more max damage than surplus prosthetics seems way too low, i think 30 for the legs and 40 for the arms would be better

i suppose that's fair

ynot01 (approved):

PR #17811, opened 2023-02-06
Reduces body temp increase from being on fire

Moltijoe:
It had a base of 30, with an additional `Fire_Stacks * 12`

it takes just 50 temp to start taking additional damage from the heat rather than just the fire
This meant that with even 2 fire stacks, one would instantly start taking damage in a single tick
Also, having 20 fire stacks mean that each life tick, bodytemp would increase by 270, this basically makes a corpse not worth reviving since after a bit they'll be so hot that it will take forever for them to cool down unless you have a shower.

Reducing it to `Fire_Stacks * 2` means that the victim will take 3 life ticks to be at max heat damage from a full fire, but 5 from minor ones rather than being instant anywhere above 10 fire stacks.

I also considered making it `15 + (Fire_Stacks * 6)` or something similar, with reduced flat heat, but increase scaling, will change the PR if someone can convince me

:cl:  
tweak: Fire is less hot
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -2106,7 +2106,7 @@ GLOBAL_LIST_EMPTY(features_by_species)
 		if(thermal_protection >= FIRE_SUIT_MAX_TEMP_PROTECT && !no_protection)
 			H.adjust_bodytemperature(11)
 		else
-			H.adjust_bodytemperature(BODYTEMP_HEATING_MAX + (H.fire_stacks * 12))
+			H.adjust_bodytemperature(BODYTEMP_HEATING_MAX + (H.fire_stacks * 3))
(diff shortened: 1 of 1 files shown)


MajManatee:
maybe now space adapt wont make me fry from the inside out because I was within 1sq km of a lit match

Moltijoe:
> maybe now space adapt wont make me fry from the inside out because I was within 1sq km of a lit match

probably won't help too much unless you put yourself out almost instantly

Moltijoe:
found a miner's corpse after them being on fire for about 5 minutes
they were ~20000 degrees
it took close to 4 minutes for me to cool them down to fix them

at that point, it would have been faster to bring them up on the station, cut their head off, and clone them
or even find a random legion corpse and do a brain transplant

MajManatee (approved):
you know i like it...
much less hot

PR #17819, opened 2023-02-07
Makes zombie scaling infection chance simpler and more consistent

Mqiib:
Current zombie infection chance is overcomplicated and janky and actually gives you really really low chance of being infected by a bite even if you have no armor. This pr changes the scaled_infection_chance path to just effectively the normal armor block to the power of 1.5 which gives technically better infection blocking than the linear one while also making single attacks still have decent chance of infecting people. Really high armor will of course make you a lot harder to infect but the chance is still there if you get hit a lot.

# Wiki Documentation

Infect chance for scaled infection is now (1 - armor block fraction) ^ 1.5 for scaled infection, if the math is relevant.

# Changelog

:cl:  

tweak: Scaling infection chance for zombie bites is now less overcomplicated and more consistent

/:cl:

### DIFF
--- a/code/modules/zombie/items.dm
+++ b/code/modules/zombie/items.dm
@@ -18,12 +18,7 @@
 	damtype = "brute"
 	var/inserted_organ = /obj/item/organ/zombie_infection
 	var/infect_chance = 100 //Before armor calculations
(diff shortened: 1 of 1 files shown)


MajManatee:
Maybe the game mode will be playable now for the zombies.

Moltijoe (approved):

PR #17820, opened 2023-02-07
Adds a syringe and hand labeler to the medical area of mining base

Moltijoe:
The idea of the vial box is that you take a blood sample of all the miners to pod clone in case a megafauna deletes them
Slight problem, there's no syringe to take the blood with
Also, adds a labeler to make it possible to keep track of bloods
image

:cl:  
mapping: Adds a syringe and hand labeler to mining base
/:cl:


patpol4 (approved):
cluttered but ok

MajManatee:
there is definitely already a hand labeler. And 99% of people just grab a charcoal syringe, shoot it up (because it does nothing) and then problem solved.

Moltijoe:
fair, labeler is in morgue, and i suppose the charcoal syringe works

PR #17828, opened 2023-02-07
IPC martial buffs

Moltijoe:
Renames a couple of defines, makes a code a bit nicer in general

It still never gets used, and when it does the user gets bodied by sec nearly instantly since the wings make them instantly valid.
couple tweaks:
Immune to stamina damage - apparently neither stunimmune nor damageslowdownimmune prevent stamina slowdown
Gives a 0.1 default speed boost - makes them the same as ethereals, this will probably be barely noticeable
removes hardcrit - I didn't want to just give them more health, this fits thematically since crit isn't in ultrakill anyways

Also adds a warning to the uplink that it will remove push and grab so people know what they're getting into.

:cl:  
tweak: IPC martial now makes you immune to stamina damage and hard crit
tweak: IPC martial now gives a 0.1 movementspeed boost at all times
/:cl:

### DIFF
--- a/code/datums/martial/ultra_violence.dm
+++ b/code/datums/martial/ultra_violence.dm
@@ -290,6 +290,7 @@
 	H.dna.species.punchstunthreshold += 4
 	H.dna.species.staminamod = 0 //my god, why must you make me add all these additional things, stop trying to disable them, just kill them
 	H.dna.species.speedmod -= 0.1
+	update_movespeed(TRUE)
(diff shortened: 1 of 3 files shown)


MajManatee:
just a better version of hunter fu

ynot01 (changes requested):
Balance council: Make it so users with this martial art can only use the guns that it grants them and not disablers or other guns

Moltijoe:
> Balance council: Make it so users with this martial art can only use the guns that it grants them and not disablers or other guns

they already have no_stun_weapons, i'll add the no other guns thing

PR #17842, opened 2023-02-08
Removes appendix from darkspawn

Moltijoe:
They can't have surgery done on them to remove it

:cl:  
rscdel: Darkspawn appendix
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/shadowpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/shadowpeople.dm
@@ -15,6 +15,7 @@
 
 	mutanteyes = /obj/item/organ/eyes/night_vision
 
+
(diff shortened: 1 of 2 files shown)


tattax:
you should make the event ignore their antag datums (/datum/round_event/spontaneous_appendicitis/start())


appendix gang

tattax:
or if you really want to.. you could make a no appendix trait.. since we have NOLIVER and NOSTOMACH already

tattax:
you could also just add trait_nohunger to darkspawns since it removes appendix
and hunger (which sling has but dspawn doesn't)

Moltijoe:
> you could also just add trait_nohunger to darkspawns since it removes appendix and hunger (which sling has but dspawn doesn't)

huh, didn't know nohunger did that, that's probably better so dspawns don't need to eat either

PR #17844, opened 2023-02-08
Moves locking and unlocking lockboxes to ctrl so they can be accessed with alt

Moltijoe:
Accessing the contents of the lockbox is highly irritating
This makes it not that

One can still drag lockboxes that are on the ground, it just locks or unlocks them

:cl:  
tweak: Lockboxes are locked and unlocked with ctrl
tweak: Can access the inventory of a lockbox with alt
/:cl:

### DIFF
--- a/code/game/objects/items/storage/lockbox.dm
+++ b/code/game/objects/items/storage/lockbox.dm
@@ -70,7 +70,7 @@
 	open = TRUE
 	update_icon()
 
-/obj/item/storage/lockbox/AltClick(mob/user)
+/obj/item/storage/lockbox/CtrlClick(mob/user)
 	..()
 	if(!user.canUseTopic(src, BE_CLOSE))
 		return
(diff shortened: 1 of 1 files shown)


tattax:
alt click is consistent with lockers and crates, i just suggest making it unlock with altclick only when you hold it or you could also do it on the ground. same works for ctrl where you could check if loc == user for making it less annoying when ctrl clicking with it

Moltijoe:
> alt click is consistent with lockers and crates, i just suggest making it unlock with altclick only when you hold it or you could also do it on the ground. same works for ctrl where you could check if loc == user for making it less annoying when ctrl clicking with it

you mean, make it unlock with alt, but otherwise access it if it's unlocked?

JamieD1:
> Accessing the contents of the lockbox is highly irritating This makes it not that
> 
> One can still drag lockboxes that are on the ground, it just locks or unlocks them
> 
> 🆑 tweak: Lockboxes are locked and unlocked with ctrl tweak: Can access the inventory of a lockbox with alt /🆑

what if you want to pull it?

Moltijoe:
> what if you want to pull it?

it still lets you pull it, it just toggles lock.
unlike other things that can be locked, this one is an item that can be picked up, so it's not so common that someone will drag it.

PR #17845, opened 2023-02-08
Lets Overwatch be installed on any device with access

Moltijoe:
Unlike other secureye programs, this one is more of a paramedic tool than a shitsec tool
Also lets mining medic keep an eye on miners whenever they have to pop station side for tools

:cl:  
rscadd: Mining medic starts with overwatch installed
tweak: Overwatch can be installed on any device with access
/:cl:


Moltijoe:
Gonna give them a laptop with the programs preloaded instead

PR #17846, opened 2023-02-08
Chasm maze ruin now actually gives loot

Moltijoe:
carefully walking through the maze to not die
at the end > no loot

:cl:  
bugfix: Chasm maze ruin now gives loot
/:cl:


ToasterBiome (approved):

patpol4 (approved):
ah, right

PR #17848, opened 2023-02-08
Adds barricading of airlocks and doors

Majkl-J:

Allows barricading of airlocks and doors, same as with windows. I believe this is fairly reasonable seeing as a few ruins already have that as a thing and putting a few planks across a door is a thing that should be possible.

(Tested, however not tested on each special instance of doors and airlocks)

Adds some small tweaks to already existing stuff too.

# Changelog
:cl:  
rscadd: Airlocks and doors can now be barricaded
tweak: The crude barricade is now faster to crowbar than harm intenting it with it
bugfix: It is no longer possible to put multiple barricades on the same tile
/:cl:

### DIFF
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -100,7 +100,7 @@
 	if(I.tool_behaviour == TOOL_CROWBAR && user.a_intent != INTENT_HARM)
 		user.visible_message("[user.name] starts prying [src.name] apart.", \
 							span_notice("You start prying the barricade apart"))
-		if(I.use_tool(src, user, 190, volume=50))
+		if(I.use_tool(src, user, 100, volume=50))
 			to_chat(user, span_notice("You disassemble the barricade."))
 			new /obj/item/stack/sheet/mineral/wood(user.loc, 5)
 			qdel(src)
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -100,7 +100,7 @@
 	if(I.tool_behaviour == TOOL_CROWBAR && user.a_intent != INTENT_HARM)
 		user.visible_message("[user.name] starts prying [src.name] apart.", \
 							span_notice("You start prying the barricade apart"))
-		if(I.use_tool(src, user, 100, volume=50))
+		if(I.use_tool(src, user, 10 SECONDS, volume=50))
 			to_chat(user, span_notice("You disassemble the barricade."))
 			new /obj/item/stack/sheet/mineral/wood(user.loc, 5)
 			qdel(src)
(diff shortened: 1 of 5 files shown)


MajManatee (changes requested):
as funny as it would be, please add a check to see if there is already a baracade there. 

I can imagine the rage when someone tries to break into cargo only to right click on the door and see 
baracade
baracade
baracade
baracade
baracade
baracade
baracade
baracade
baracade
baracade
baracade
baracade
baracade
baracade
Airlock

code/game/machinery/doors/airlock.dm
```diff
@@ -952,6 +952,18 @@
 				return
 	add_fingerprint(user)
 
+	if(istype(C,/obj/item/stack/sheet/mineral/wood))
+		var/obj/item/stack/sheet/mineral/wood/W = C
+		if(W.amount < 5)
+			to_chat(user, span_warning("You need at least five wooden planks to barricade the airlock!"))
```
MajManatee:
```suggestion
			to_chat(user, span_warning("You need at least five wooden planks to barricade the window!"))
```

Maia-J:
> as funny as it would be, please add a check to see if there is already a baracade there.
> 
> I can imagine the rage when someone tries to break into cargo only to right click on the door and see baracade baracade baracade baracade baracade baracade baracade baracade baracade baracade baracade baracade baracade baracade Airlock

The glorious shitcode of the barricade actually seems to already allow this on windows for some reason, fixing time

MajManatee (changes requested):

code/game/machinery/doors/airlock.dm
MajManatee: im dumb

MajManatee:
did you test on windows too??

Maia-J:
> did you test on windows too??

Windows can already be barricaded, and yes, they have that issue on live

MajManatee:
💀 
halves the removal time...

Maia-J:
> 💀 halves the removal time...

That makes it about 20% faster than hitting them with the crowbar on harm and I am not snowflaking that part of code nor giving them more durability

MajManatee (approved):
should be fine

code/game/machinery/deployable.dm
```diff
@@ -100,7 +100,7 @@
 	if(I.tool_behaviour == TOOL_CROWBAR && user.a_intent != INTENT_HARM)
 		user.visible_message("[user.name] starts prying [src.name] apart.", \
 							span_notice("You start prying the barricade apart"))
-		if(I.use_tool(src, user, 190, volume=50))
+		if(I.use_tool(src, user, 100, volume=50))
```
Moltijoe:
```suggestion
		if(I.use_tool(src, user, 10 SECONDS, volume=50))
```

Moltijoe: might as well update it to use the defines

Moltijoe (changes requested):

Moltijoe (approved):
❤️

PR #17854, opened 2023-02-08
Gives mining medic a laptop to keep track of bodycams

Moltijoe:
Apparently brig physician gets a laptop with secureye, probably to keep track of secoff bodycams
So i've given one to mining medic to let them keep track of their miner's bodycams

:cl:  
rscadd: Mining medic now gets a laptop with overwatch installed
tweak:  Brig physician laptop also starts with lifeline and crew monitor
/:cl:

### DIFF
--- a/code/modules/modular_computers/computers/item/laptop/laptop_presets.dm
+++ b/code/modules/modular_computers/computers/item/laptop/laptop_presets.dm
@@ -25,7 +25,13 @@
 	. = ..()
 
 /obj/item/modular_computer/laptop/preset/paramedic/mining_medic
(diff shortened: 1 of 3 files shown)


nmajask:
May want to consider giving the mining medic laptop a better network card, otherwise it won't be able to connect to the NTnet while on lavaland.

Moltijoe:
> May want to consider giving the mining medic laptop a better network card, otherwise it won't be able to connect to the NTnet while on lavaland.

Shows what I know about modular computer components. Thanks for the info

MajManatee (approved):
like brig phys but not

PR #17868, opened 2023-02-10
Removed dilithium cost from super power cell, increased glass cost to match that of hyper power cell, added silver and gold cost to half that of hyper power cell cost.

Uristthedorf:

Removes dilithium cost from super power cell, and increases glass cost to match that of hyper power cell, added silver and gold cost to half that of hyper power cell cost. As of now, super power cells are the only T2 part that cannot be produced with just iron and glass. This PR allows people to get some cell upgrades even if miners are dead or non-existent. This also makes it easier for technophile chaplain to get favor without bankrupting the ore silo of its dilithium, making every else mad.

I would have tested my changes but Build.bat doesn't seem to work for me. (Even when I don't have any files changed.)

# Wiki Documentation

I don't believe the cost of super power cells (or other cells) are mentioned anywhere on the wiki.

# Changelog

:cl:
tweak: Removed super power cell dilithium cost, increased super power cell glass cost to match that of hyper power cell, added silver and gold cost to half that of hyper power cell cost.  
/:cl:

### DIFF
--- a/code/modules/research/designs/power_designs.dm
+++ b/code/modules/research/designs/power_designs.dm
@@ -29,7 +29,7 @@
 	desc = "A power cell that holds 20 MJ of energy."
 	id = "super_cell"
 	build_type = PROTOLATHE | MECHFAB
-	materials = list(/datum/material/iron = 700, /datum/material/glass = 70, /datum/material/dilithium = 250) //Yogs: added dilithium
+	materials = list(/datum/material/iron = 700, /datum/material/glass = 80)
 	construction_time=100
 	build_path = /obj/item/stock_parts/cell/super/empty
 	category = list("Misc","Power Designs")
--- a/code/modules/research/designs/power_designs.dm
+++ b/code/modules/research/designs/power_designs.dm
@@ -29,7 +29,7 @@
 	desc = "A power cell that holds 20 MJ of energy."
 	id = "super_cell"
 	build_type = PROTOLATHE | MECHFAB
-	materials = list(/datum/material/iron = 700, /datum/material/glass = 80)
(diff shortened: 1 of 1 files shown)


ThatLing:
Dilithium was a stupid addition

> I would have tested my changes but Build.bat doesn't seem to work for me. (Even when I don't have any files changed.)

If the build.bat doesn't work, you can also compile TGUI and DM seperately.

code/modules/research/designs/power_designs.dm
```diff
@@ -29,7 +29,7 @@
 	desc = "A power cell that holds 20 MJ of energy."
 	id = "super_cell"
 	build_type = PROTOLATHE | MECHFAB
-	materials = list(/datum/material/iron = 700, /datum/material/glass = 70, /datum/material/dilithium = 250) //Yogs: added dilithium
+	materials = list(/datum/material/iron = 700, /datum/material/glass = 80)
```
MajManatee: so in exchange for 250 dilithium, it now costs... 10 more glass...?
250 of basically bluespace for 10 glass?

that dont sound right

MajManatee (changes requested):

code/modules/research/designs/power_designs.dm
MajManatee: /datum/material/iron = 700, /datum/material/gold = 75, /datum/material/silver = 75, /datum/material/glass = 80

so its a cheaper hyper cell

Uristthedorf: Done

Uristthedorf:
> > I would have tested my changes but Build.bat doesn't seem to work for me. (Even when I don't have any files changed.)
> 
> If the build.bat doesn't work, you can also compile TGUI and DM seperately.
>
How would you compile them separately? DM in particular.

MajManatee (approved):
sounds fine to me

PR #17873, opened 2023-02-10
seismic arm's overcharged emitter flight path fix

LazennG:

fixes a bug with the victim of the overcharged emitter attack straying from the path if the user moves around after hitting the target

# Changelog
:cl:  
bugfix: fixed a seismic arm bug  
/:cl:


MajManatee:
should have been an intended feature, that sounds bad ass

LazennG:
i originally called it a feature until seeing that fauna would sometimes just noclip into rocks instead of flying into them

Moltijoe (approved):

PR #17875, opened 2023-02-10
Kindle range increase

MajManatee:

kindle is uhh 3 tiles, and thats kinda crazy short.
Increases it to 5 tiles because thats not as bad

Stun WAS going to be increased but its 10 seconds already, only lowing if you hurt the person. so uh... dont stab them i guess.

# Changelog

:cl:  
tweak: Kindle range 3 -> 5 tiles
/:cl:


Moltijoe (approved):
number change

JamieD1:
image

Moltijoe:
General consensus was that a melee touch version of kindle would be better

PR #17880, opened 2023-02-10
Makes vote weight and preferred map actually work

adamsong:

Vote weights are currently not considered during the voting process. Now they are.

# Changelog

:cl:  
bugfix: fixed vote weights not working
bugfix: fixed preferred map pref not working
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -42,12 +42,7 @@ SUBSYSTEM_DEF(vote)
 /datum/controller/subsystem/vote/proc/get_result()
 	//get the highest number of votes
 	var/greatest_votes = 0
-	var/total_votes = 0
-	for(var/option in choices)
-		var/votes = choices[option]
-		total_votes += votes
-		if(votes > greatest_votes)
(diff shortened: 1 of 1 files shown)


patpol4:
yes we can

MajManatee:
alternate idea
every 10 rounds, it picks each map to be played twice , and whatever map wins the vote gets played an additional two times, then it randomly rotates through the 10 map list until expended, and re-votes.

adamsong:
Sounds like a meme

Marmio64:
i personally disagree with vote weight in the first place. People already vote for other maps and they get played, why do we still need a bias against box? Only argument I could see that would make logical sense would be because a lot of players would keep the default option as their preferred map (box), but could you not just have it pick a random default map if it already doesn't do something like that?

patpol4:
Its literally just a bugfix, this was decided long ago

ynot01 (approved):
Voting weights are a thing and have been for a long time (with staff and coders assuming that they worked), if you disagree with them in their current form then we can disable the weights in another PR

PR #17887, opened 2023-02-11
Actually removes janitors access to the engine that the job had for the last two months

patpol4:
title
changes it to construction access instead so they can still go into engineering foyer.

# Changelog

:cl:  
bugfix: removes janitor access to engine
/:cl:


Moggram:
You can get in, you just have to use the back door. Seemed like a cool way to do it.

ToasterBiome (approved):

PR #17892, opened 2023-02-11
Removes 'id in pda' preference, makes it the default behaviour

ThatLing:
My previous PR didn't work (testing is for losers). I will probably have to add some spaghetti to get it to work correctly, so instead I propose to just remove the preference. This fixes some preference bloat and adds a sane default behaviour for all players.

# Changelog

:cl:  
rscdel: Removed 'id in pda' preference, it is now the default behaviour
/:cl:


Marmio64:
there's going to be so many players mhelping 'where is my id?'

ToasterBiome (approved):

Simplehorror:
no, i always throw my PDA away, now I will have to take it out of my ID before throwing it away

Cartlord:
This will lead to new players not knowing the ID and P.D.A. are seperate.
The thing is, we also don't actually have much reason for them to be TREATED seperately, so I don't think that matters much... you only ever take your card out of your P.D.A. to use a wallet, get an access change, or go Unknown.

PR #17900, opened 2023-02-12
Moves some config settings around which should make debug builds from github run faster

Bibby0110:
hon hon

also turns replays back on for main


JamieD1 (approved):

ynot01:
this shit better be merged before next round

PR #17919, opened 2023-02-13
[s] fixes recharger exploit

ynot01:
# Changelog

:cl:  
bugfix: Fixed an exploit regarding rechargers
/:cl:


Moltijoe (approved):
i take it that the exploit is "it keeps charging even after taken off the charger"?

PR #17926, opened 2023-02-14
More Ethereal colours

Moltijoe:
Ethereal colours are based off stars, and named as such
Slight problem, don't see any green on this spectral colouring graph
image

They also don't have either white or orange despite them being prominent star colours

Adds light purple, orange, light red and white
Makes dark green slightly more green
Makes yellow less muddy
image

:cl:  
rscadd: More Ethereal colours
tweak: Ethereal colours are (sorta) properly classified
/:cl:


Moja10:
The reason you don't see green is because stars are not green, they are only the colors on that spectrum due to physics

Moltijoe:
> The reason you don't see green is because stars are not green, they are only the colors on that spectrum due to physics

that's the point, why are ethereals green when stars can't be.

MajManatee (approved):
game changing

PR #17927, opened 2023-02-14
Thralled or Veiled Ethereals create a light void

Moltijoe:
Ethereals are notoriously difficult to thrall or veil
If a sling or dspawn is suicidal enough to try and pulls it off, might as well reward them
Ethereals that are thralled or veiled create giant dark zones for their masters to walk around in safely
image

:cl:  
rscadd: Thralled or Veiled Ethereals succ light
/:cl:


Moltijoe:
Only half works, can't figure how to make it full work, someone that knows the lighting system better would need to do this.

PR #17935, opened 2023-02-15
Malfunctioning AIs no longer need to kill androids or synthetics

SapphicOverload:

The objective to prevent organics from escaping now requires all shuttle occupants to have the MOB_ROBOTIC biotype and not have the MOB_ORGANIC biotype. Cyborgs, IPCs, androids, synths, and clockwork golems will be allowed to escape without causing the objective to fail. The doomsday device will also work similarly.

# Wiki Documentation

Might be worth mentioning on the guide to malfunction
https://wiki.yogstation.net/wiki/Guide_to_malfunction

# Changelog

:cl:  
tweak: AI no longer needs to kill androids or synthetics
tweak: AI doomsday will now spare androids and synthetics
/:cl:

### DIFF
--- a/code/game/gamemodes/objective.dm
+++ b/code/game/gamemodes/objective.dm
@@ -463,7 +463,7 @@ GLOBAL_LIST_EMPTY(objectives)
 	if(SSshuttle.emergency.mode != SHUTTLE_ENDGAME)
 		return TRUE
 	for(var/mob/living/player in GLOB.player_list)
-		if(player.mind && player.stat != DEAD && !(issilicon(player) || isipc(player)))
+		if(player.mind && player.stat != DEAD && ((MOB_ORGANIC in player.mob_biotypes) || !(MOB_ROBOTIC in player.mob_biotypes)))
 			if(get_area(player) in SSshuttle.emergency.shuttle_areas)
 				return FALSE
 	return TRUE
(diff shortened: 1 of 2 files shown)


Snellie12:
Makes sense lore-wise; might even be able (if this passes) to add antag roles adding into that; making it so they have to protect the AI. 

Although gameplay wise...I can see Ye Ol' Classic Plasmeme Flood just being used and borgs not giving a damn because 'Not borg, not AI = Death'.

Moltijoe:
preterni feeling singled out with that !MOB_ORGANIC.
jokes aside, this is honestly just a better solution than my addition of just IPC was

Moltijoe (approved):

wonderinghost:
Me and I think tank were talking about when a Borg throws a inorganic like ipc into the factory they would get mindslaved. And with ipc it would turn them into a tankier v1 ipc

iamthesilly:
> Me and I think tank were talking about when a Borg throws a inorganic like ipc into the factory they would get mindslaved. And with ipc it would turn them into a tankier v1 ipc

I can make a separate PR for this

yacabo:
Wouldn't you want Preterni in as well? They might be partially organic but so are cyborgs with MMIs

Moltijoe:
> Wouldn't you want Preterni in as well? They might be partially organic but so are cyborgs with MMIs

probably best not to since all preterni are at least a bit organic, where as most cyborgs are posibrains rather than MMIs

Uristthedorf:
Would it be better if malf AI could spare inorganic mobs like golems? It would make sense because the objective text only mentions organics can't escape so explicitly inorganic species like golems should be able to live.

PR #17953, opened 2023-02-16
Moves antimagic from nullrod to chaplain

Moltijoe:
Lets a late-join chaplain with a stolen null rod still have antimagic
Lets a chaplain actually swap between items without risking getting bopped by magic
Miners can get extra nullrods (the spirit and armblade ones), this makes it so cults and wizards only ever need to worry about one antimagic person

# Wiki Documentation

Remove text about nullrod giving antimagic
add text about chaplain having antimagic

:cl:  
rscadd: Adds antimagic to chaplain
rscdel: Removes antimagic from nullrod
/:cl:

### DIFF
--- a/code/game/objects/items/holy_weapons.dm
+++ b/code/game/objects/items/holy_weapons.dm
@@ -253,10 +253,6 @@
 	var/reskinned = FALSE
 	var/chaplain_spawnable = TRUE
 
-/obj/item/nullrod/Initialize()
-	. = ..()
-	AddComponent(/datum/component/anti_magic, TRUE, TRUE, FALSE, null, null, FALSE)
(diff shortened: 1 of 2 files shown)


00ze-cyclone:
Wizard and cult will still have to worry about more than one person because now every chaplain is antimagic instead of only one, also mean if there’s no chaplain you can get fucked to obtain antimagic unless you can manage to have botany work.
Also I kinda prefer jobs relying on tools rather than having things innately because it mean you can’t remove them but also you can’t replace them.

Moltijoe:
aight, sounds like i'll just make more nullrod options for pacifist chaplains then

PR #17955, opened 2023-02-16
Adds a pacifism check to the cross nullrod

Moltijoe:
It hurts people, it shouldn't bypass pacifism

:cl:  
rscadd: Cross nullrod now checks for pacifism
/:cl:

### DIFF
--- a/code/game/objects/items/holy_weapons.dm
+++ b/code/game/objects/items/holy_weapons.dm
@@ -896,6 +896,9 @@
 	if(held_up)
 		unwield(user)
 		return
+	if(HAS_TRAIT(user, TRAIT_PACIFISM))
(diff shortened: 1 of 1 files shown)


Hislyna:
Hows about, if it marks you as a pacifist, it will only do the stam damage

Moltijoe:
> Hows about, if it marks you as a pacifist, it will only do the stam damage

i suppose that's fair

ThatLing (approved):

MajManatee:
go back to original idea because of #17977 ?

Moltijoe:
> go back to original idea because of #17977 ?

that's still waiting on balance review

Moltijoe:
covered in a different PR

PR #17956, opened 2023-02-16
Chaplain nullrod shows 11 items per page instead of 7

Moltijoe:
there's like 5 pages of nullrods, that's kinda dumb
image

:cl:  
tweak: Chaplain nullrod now shows 11 items per page instead of 7
/:cl:


Moltijoe:
converting it to tgui instead, this will cause conflicts

PR #17866, opened 2023-02-10
Unfucks darkspawn veils but they're still weaker

Mqiib:
Darkspawn thralls now get -40 max hp. This means they take effectively ~167% damage from all sources. Also you are no longer screwed over after being dethralled with -30 max hp. Finally, makes the effect apply when it's supposed to apply (cloning would entirely circumvent the health increase/loss.)

# Wiki Documentation

Darkspawn thralls lose 40 max hp. Deconversion reverses this.

# Changelog

:cl:  

bugfix: Cloning no longer circumvents darkspawn veil health loss
tweak: Darkspawn veils now lose 40 max hp rather than 65 and get it ALL back upon deconversion  

/:cl:

PR #17964, opened 2023-02-17
Re-un-fucks darkspawn veils because apparently veil mind is wonky fuck wucky and doesn't call all the procs it is supposed to or something

Mqiib:
[#17866] part 2: veil mind boogaloo
I'll test this when I'm not slapping rats

# Changelog

:cl:  
 
bugfix: Fixes veil mind max HP buggery probably

/:cl:


ynot01 (changes requested):
lints

adamsong (changes requested):
Looking into why the original didn't work, don't want a bandaid if it can be avoided

adamsong:
Need to fix properly

PR #17975, opened 2023-02-18
fully_heal() removes changeling succ BADDNA trait

Moltijoe:
Calling revive with full heal will revive someone with BADDNA and remove the husk
if they proceed to die again they can be revived normally with a defib
However, they won't be cloneable and can't use dna injectors

:cl:  
tweak: Aheal removes changeling succ BADDNA trait
/:cl:


ynot01 (approved):
Does what it says

PR #17977, opened 2023-02-18
Gold Cross null rod damage revamp

ynot01:

The rod was pretty much universally used as a "stunlock" weapon because it applied stamina damage quickly

Stamina damage is now capped at 65

Pacifists can only apply stamina damage

# Changelog

:cl:  
tweak: Golden Crucifix null rod stamina damage is now capped at 65
bugfix: Pacifists can no longer apply lethal damage with the Golden Crucifix
/:cl:

### DIFF
--- a/code/game/objects/items/holy_weapons.dm
+++ b/code/game/objects/items/holy_weapons.dm
@@ -942,6 +942,11 @@
 		COOLDOWN_START(src, holy_notification, 0.8 SECONDS)
 		notify = TRUE
 
+	var/mob/user = loc
+	var/damage = TRUE
+	if(HAS_TRAIT(user, TRAIT_PACIFISM))//no hurting as a pacifist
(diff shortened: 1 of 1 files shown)


Maia-J:
Make it cap the stam damage it deals at whatever the slowdown value is, so we can still have the reference work

ynot01:
> Make it cap the stam damage it deals at whatever the slowdown value is, so we can still have the reference work

im not sure how this breaks the reference, the cross never slowed down enemies it kills them

Hislyna:
Well it does technically slow down certain enemies when theyre struck, others just run or take damage and move as normal

Moltijoe:
Balance council changes requested:
Add a cap to the stamina damage instead
Add a pacifism check to remove the burn entirely

Moltijoe (approved):

PR #17979, opened 2023-02-18
holy light now only goes on cooldown if it actually healed someone

SomeguyManperson:

blah blah skill issue blah blah

:cl:  
tweak: holy light bible healing will no longer go on cooldown if you try to heal someone who is already at full health
/:cl:


Hislyna:
Skill Issue

Hopekz:
Imo it would probably be cool if the cooldown was proportional to the amount of the full healing you have done.
So if you do 10% of the healing you were able to do, your cool down would only be 10% of the max

That way you can top up a larger amount of people that only need like one to two points of healing

Moltijoe (approved):

PR #17985, opened 2023-02-19
transcendent olfaction nerf

ynot01:

image

Sanguine Olfaction unchanged

Transcendent Olfaction & Follow The Scent now has a 60 second cooldown

Smell lines now only last 2 seconds, leading you in the correct direction but not being able to track them down perfectly

Bloodsuckers can no longer be smelled

## Rationale

Transcendent Olfaction makes antag play agonizing. Antagonists, once caught or suspected by security, are forced to play this ring-around-the-rosie around the station while security perfectly tracks them 24/7. Once you're caught, it's effectively game over. Unless you murder everyone with the mutation and delete it from records, you WILL be caught. 

This makes re-entering stealth impossible, which is awfully unbalanced and an absolute migraine to play against.

Everything I just said is doubly true for bloodsuckers, who are meant to establish a **hidden** lair and use it as a safe-grounds. 

Unfortunately for them, once security catches the scent of anything the bloodsucker touched, the bloodsucker will never be able to hide again.

Once Sol inevitably comes around, the bloodsucker will either be busy fighting security due to being unable to flee and hide or they will run to their lair, cover up their tracks, and security will be led straight to their lair where they can pry open the coffin and kill them via Sol or wait out Sol and kill the bloodsucker who is now cornered.

# Changelog

:cl:  
tweak: Transcendent Olfaction now has a 60 second cooldown
tweak: Transcendent Olfaction now vanishes much faster
tweak: Transcendent Olfaction can no longer track bloodsuckers
/:cl:

### DIFF
--- a/code/modules/antagonists/bloodsuckers/powers/olfaction.dm
+++ b/code/modules/antagonists/bloodsuckers/powers/olfaction.dm
@@ -63,10 +63,10 @@
 		if(tracking_flags & TRACKING_SCENT)
 			scents |= samples.return_scents()
 	for(var/mob/living/carbon/C in GLOB.carbon_list)
-		if(IS_BLOODSUCKER(C)) // Bloodsuckers have no scent, and finding their lair with this during Sol would be OP
-			continue
 		if(blood_samples.Find(C.dna.unique_enzymes) && !possible.Find(C))
 			possible |= C
+		if(IS_BLOODSUCKER(C)) // Bloodsuckers have no scent, and finding their lair with this during Sol would be OP
+			continue
 		if(scents[md5(C.dna.uni_identity)] && !possible.Find(C))
 			var/datum/job/J = SSjob.GetJob(C.job)
 			if(!J)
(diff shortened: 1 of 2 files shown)


Chubbygummibear:
did you check how it looks using the power with the scent trails set to 2 seconds? i worry people would be stuck in grayhound smell vision with nothing to look at for a significant amount of time

ynot01:
> did you check how it looks using the power with the scent trails set to 2 seconds? i worry people would be stuck in grayhound smell vision with nothing to look at for a significant amount of time

Grey vision lasts 20 seconds, and they will be able to "follow" the scent trail briefly as each trail tile is 0.1 seconds apart

ThatLing:
balance council speedrun

Moltijoe (approved):
number changes are easy

code/modules/antagonists/bloodsuckers/powers/olfaction.dm
```diff
@@ -137,6 +139,8 @@
 		Selecting one of those scents will grant you a new ability to follow that scent. This will create a visible trail to follow to the owner of that scent.\n\
 		WARNING: it will be difficult to see around you while following a scent trail due to the the color being drained from all your vision except for the trail and your target."
 
+	cooldown = 60 SECONDS
```
tattax, line 142: why 60 secondes and not 1 minutes

ynot01: it is likely to be moved around in seconds rather than minutes in future balancing, for example `1.1 MINUTES` is unreadable

PR #17993, opened 2023-02-20
Robotic species now have flavoured examines

Moltijoe:
Robotic species don't bruise and burn, but they do dent and char.
Should help with telling robotic and non-robotic species apart with just examines (i'm looking at you medbay lizards putting preterni in cryotubes)
Also changes bodypart damage text when using `change_bodypart_status()` all this affects is self examine, which people don't use

:cl:  
rscadd: Robotic species now have flavoured examines
/:cl:


MajManatee:
when the lizards put the man with a prothetic arm in the tube for an hour and dont even blink at it.

JamieD1 (approved):

PR #17998, opened 2023-02-20
Mad scientist kit buff

ynot01:

Wow these kits suck

Mad scientist kit now gets:

- A restricted list of moderately useful cluster busters for a syndicate instead of being able to roll things like smoke or cleaner grenades
- EZ Clean grenade box

This sets it to being worth around 20-30 TC depending on your rolls whereas before it was nigh useless always

# Changelog

:cl:  
tweak: Mad scientist special syndie kit buffed with better grenades and a box of EZ Clean grenades
/:cl:


patpol4 (commented):
Are you sure you want to name it /syndie/, wouldnt full /syndicate/ be better since that is what most of other objects use

patpol4 (approved):
Map fine

ynot01:
> Are you sure you want to name it /syndie/, wouldnt full /syndicate/ be better since that is what most of other objects use

patpol4:
must have been a dream

MajManatee (approved):
Re-paths all syndicate cluster grenades. Probably wont cause problems in the future. I hope. Would like to see that change documented in the PR.

otherwise it looks fine.

we should really go through every single kit and change them up because my god, a lot of these suck.

PR #18002, opened 2023-02-21
Buffs changeling chameleon

ynot01:

Felt silly that they just get the regular mutation that anyone else can get but have to pay for it

Changeling Chameleon:
- Acts twice as fast as normal chameleon
- Cannot be mutadoned or duplicated
- Only removes partial invisibility when moving
- Still removes invisibility entirely if you click on something

# Changelog

:cl:  
tweak: Buffed changeling chameleon ability to work faster
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/chameleon_skin.dm
+++ b/code/modules/antagonists/changeling/powers/chameleon_skin.dm
@@ -1,7 +1,7 @@
 /datum/action/changeling/chameleon_skin
 	name = "Chameleon Skin"
 	desc = "Our skin pigmentation rapidly changes to suit our current environment. Costs 25 chemicals."
-	helptext = "Allows us to become invisible after a few seconds of standing still. Can be toggled on and off."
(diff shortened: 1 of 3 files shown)


code/modules/antagonists/changeling/powers/chameleon_skin.dm
```diff
@@ -1,7 +1,7 @@
 /datum/action/changeling/chameleon_skin
 	name = "Chameleon Skin"
 	desc = "Our skin pigmentation rapidly changes to suit our current environment. Costs 25 chemicals."
-	helptext = "Allows us to become invisible after a few seconds of standing still. Can be toggled on and off."
+	helptext = "Allows us to become invisible after a few seconds of standing still. Can be toggled on and off. Better than the mutation."
```
Hopekz, line 4: This is weirdly worded imo.
Relies on the user having knowledge on something they might have never seen yet.

Moltijoe (approved):
Code works checks out, but yeah, should probably have better helptext

PR #18013, opened 2023-02-21
Makes jellypeople do clone damage.

Uristthedorf:

This makes all jellyperson variants deal clone damage while punching instead of brute. This is because normal slimes deal cellular damage while attacking so why shouldn't slime people be able to?

If this is too OP, jellypeople could have their punch damage reduced to compensate. Or maybe a subspecies could be made that gets this and no other powers.

I have tested my change.

# Wiki Documentation

This change could be added to https://wiki.yogstation.net/wiki/Guide_to_races. Only change is that jellypeople do clone damage with punches.

# Changelog

:cl:  
rscadd: All jellypeople variants do clone damage with their punches.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/jellypeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/jellypeople.dm
@@ -13,6 +13,7 @@
 	damage_overlay_type = ""
 	var/datum/action/innate/regenerate_limbs/regenerate_limbs
 	liked_food = MEAT
+	attack_type = CLONE
 	coldmod = 6   // = 3x cold damage
 	heatmod = 0.5 // = 1/4x heat damage
 	burnmod = 0.5 // = 1/2x generic burn damage


CMOisLing:
> why shouldn't slime people be able to?

Because its annoying as fuck to heal

patpol4:
Love it

ReddicusDragon:
Closing in 24 hours. Not only is this annoying to heal, it's a bitch for most people to deal with. I can almost guarantee you balance won't agree with this.

Maia-J:
Based and slime pilled

MajManatee:
yeah naw thats dumb as hell

Uristthedorf:
Maybe a better idea would be to make a jellyperson martial art that can deal cellular damage. Making it syndicate only and cost lots of TC would be more balanced. (You could make it deal less raw damage and harder to stun with than other martial arts to compensate for the damage type being so hard to heal.)

ReddicusDragon:
24 hrs

PR #18019, opened 2023-02-22
Psych back office access

Aquizit:

Adds new access define: ACCESS_PSYCH -- Psych and CMO get it. It opens the windoor into the back area where all the psych toys are. General medical access is still for the front door.
ALSO on [Asteroid] changed the windoor from regular to reinforced because of course asteroid is wrong at something.
~~ALSO ALSO set the psych's locker to psych access too~~ they don't even get their own locker it's just a generic secure one smh

# Changelog

:cl:  
rscadd: ACCESS_PSYCH define  
mapping: Sets psych back room door to psych access
mapping: [Asteroid] Makes psych back door reinforced like the others
experimental: [TO DO] Perhaps look into coding psych their own locker
/:cl:


patpol4:
The locker was just put there because of lack of any other, if they get their own one with their own special contents, that would great

patpol4 (approved):
Maps fine

PR #18030, opened 2023-02-22
Grenadier belt changes

MajManatee:

Removes the completley useless flashbang and 4 smokes that you would just throw out anyways in exchange for 2 more frags that you actually WILL use.

# Changelog

:cl:  
tweak: Useless grenades removed from grenadier belt
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -677,11 +677,9 @@
 
 /obj/item/storage/belt/grenade/full/PopulateContents()
 	var/static/items_inside = list(
-		/obj/item/grenade/flashbang = 1,
-		/obj/item/grenade/smokebomb = 4,
 		/obj/item/grenade/empgrenade = 1,
 		/obj/item/grenade/empgrenade = 1,
-		/obj/item/grenade/syndieminibomb/concussion/frag = 10,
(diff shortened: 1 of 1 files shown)


00ze-cyclone:
change the uplink description to tell nuke ops they get 23 grenades in it instead of 26 now.

JamieD1:
Smoke can be useful, kinda feels like you are forcing your preference onto other people here.

Mantlecrawler:
I have never seen a nukie use the flashbangs or smoke grenades asides from myself and even then that was one time.

MajManatee:
@JamieD1 do a poll then and we can find out. I'm willing to bet everyone would rather have more killing grenades than some smokes. Especially on nukies.

ReddicusDragon:
I'm gonna slap awaiting action HD here since Jamie seems to be considering this PR

ynot01 (changes requested):
> change the uplink description to tell nuke ops they get 23 grenades in it instead of 26 now.

ynot01:
no point in holding this PR for HD

Moja10:
why not 1 smoke and 1 flashbang on top of the two new frags? they can have niche use, syndies didn't need multiple though

MajManatee:
i dont care enough to change it

PR #18031, opened 2023-02-22
Explosives at a cheap price: Frag grenades for 3TC

ynot01:

Satisfactory

#

For a very long time I have wanted minibombs to be cheaper, but I was always told that they were set to a high price because people could use them to gib people. WHat? Who uses minibombs like that? 6TC for a gibber? Why wouldn't I just do something FUN with that 6TC instead?

So here is the compromise. Frag grenades are here to replace minibombs at half the price. The difference? No gibbing.

**Frag grenades are identical to minibombs, but have 1 extra range of flames and no devastation range.**

(Frag grenades previously had one less light range but in my opinion the explosion seemed piss small without it so they were bumped up one to match minibombs)

# Changelog

:cl:  
tweak: Frag Grenades are now in the traitor uplink for 3TC which are minibombs but without gibbing
tweak: Syndicate Minibombs are now nukie-locked
/:cl:

### DIFF
--- a/code/game/objects/items/grenades/syndieminibomb.dm
+++ b/code/game/objects/items/grenades/syndieminibomb.dm
@@ -18,7 +18,7 @@
 
 /obj/item/grenade/syndieminibomb/concussion/prime()
 	update_mob()
-	explosion(src.loc,0,2,3,flame_range = 3)
+	explosion(src.loc,0,2,4,flame_range = 3)
 	qdel(src)
 
 /obj/item/grenade/syndieminibomb/concussion/frag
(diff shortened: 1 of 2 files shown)


Moja10:
You do not need to lock minibombs to nukie, otherwise good change.

ynot01:
> You do not need to lock minibombs to nukie, otherwise good change.

Why would we keep them?

For these two items to exist in the same ecosystem and the minibomb not be a noob trap the description would have to include "identical to frags but gib people" which is just shitty to me

MajManatee:
minibombs are bigger than frags, no?
if not, then they should be.

00ze-cyclone:
why not just making minibombs not gib so you can make them 3 TC? Frags are kinda ugly

Moltijoe (approved):

PR #18037, opened 2023-02-22
slightly expands box maint

Marmio64:
this just adds a couple extra maint rooms to box, really minor, just to slightly expand it and give a few more places to hide/build evil things in. Some space (such as specifically the extra room in arrivals maint) doesn't have a door so you can weld it open+hide it's entrance and hide things in it without it looking out of place.

This is kind of rushed+ugly and all, but I personally think box could just use with a few extra hiding spots and whatever, so these are just some small things I came up with.

hiding3
hiding2
hiding1

# Changelog

:cl:  
rscadd: Adds a little bit more box maint room.
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -22332,6 +22332,15 @@
 	},
 /turf/open/floor/plasteel,
 /area/quartermaster/storage)
+"eOA" = (
+/obj/machinery/door/airlock/maintenance_hatch,
+/obj/effect/mapping_helpers/airlock/abandoned,
+/obj/machinery/door/firedoor/border_only,
+/obj/machinery/door/firedoor/border_only{
+	dir = 1
+	},
+/turf/open/floor/plating,
+/area/maintenance/port/aft)
 "eOX" = (
 /obj/structure/bed,
 /obj/item/bedsheet/medical/virology,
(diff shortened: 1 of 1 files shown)


ToasterBiome (dismissed):
I like the ideas presented, but some concerns:

- The biggest concern is that Asteroid may have implied approval of making maint rooms accessible from weird angles that aren't just pure cardinal directions. In the first image, at the very least the doors should be centered, will have to take a bit more planning. 
- We don't do sectioned off pieces of maint like the middle picture, don't know how I like the sectioned off sectioned off piece of maint either.
- Some indication that there's a room or something would be nice for the last picture instead of it being a closed off box. Below science maint we have tinted walls to imply there's something there instead of "Oops guess you'll never know".
I like the idea of adding more maint and this doesn't add too much bloat.

Marmio64:
> I like the ideas presented, but some concerns:
> 
> * The biggest concern is that Asteroid may have implied approval of making maint rooms accessible from weird angles that aren't just pure cardinal directions. In the first image, at the very least the doors should be centered, will have to take a bit more planning.
> * We don't do sectioned off pieces of maint like the middle picture, don't know how I like the sectioned off sectioned off piece of maint either.
> * Some indication that there's a room or something would be nice for the last picture instead of it being a closed off box. Below science maint we have tinted walls to imply there's something there instead of "Oops guess you'll never know".
>   I like the idea of adding more maint and this doesn't add too much bloat.

i can do the other things but i'm not sure why we couldnt have some sectioned off maints, as it helps a great deal with random maint goers just accidently stumbling into antag lairs

Marmio64:
adjusted the new maint areas

less sectioned off and centered doors
hiding1

centered doors and marginally more space
hiding2

adds tinted window to cue the existence of the maint box
hiding3

patpol4 (changes requested):
The sectioned off parts are ok i guess even though they look a bit stupid, its nice to have a path everywhere but we'll see how it goes
However:
this one has no entry
image

bottom circle has a space turf on it, upper one need to be moved to the right position
image
also maybe add a lattice or two on the space side so it doesnt look as blocky

Marmio64:
> The sectioned off parts are ok i guess even though they look a bit stupid, its nice to have a path everywhere but we'll see how it goes However: this one has no entry image
> 
> bottom circle has a space turf on it, upper one need to be moved to the right position image also maybe add a lattice or two on the space side so it doesnt look as blocky

fixed the space tile and stuff
Screenshot 2023-02-23 153707

but uh, the no entry area has been like that, and i didn't change that. I'm of the opinion that it is fine since it's still pretty obvious something is inside and that it's really not in the 'scope' of this PR which is just expanding it, so I am going to leave it unless you really want it changed in this PR?

patpol4 (approved):
Ah ok its fine

PR #17833, opened 2023-02-07
Darkspawn is now a midround instead of a gamemode

ynot01:
# Document the changes in your pull request

Unfortunately darkspawn rounds go way too quick. Rounds are decided in minutes to either side. Lumping them into other gamemodes should make their flow go a little smoother instead of everyone realizing it's a darkspawn round and spamming lights and validhunting.

If there is 5 or more security members present, 2 darkspawn will spawn, otherwise only 1 will spawn

Players will be forcefully divulged upon spawning

Nightmare is also disabled

# Changelog

:cl:  
tweak: Darkspawn has been moved to a midround event instead of a gamemode
rscdel: Nightmare has been removed from midround rotation
/:cl:

PR #18070, opened 2023-02-25
Revert "Darkspawn is now a midround instead of a gamemode"

Cartlord:
Reverts yogstation13/Yogstation#17833 because Darkspawn are a round-ender, they shouldn't be midround because every crew-member and antag has to work around them. That's a gamemode, as far as I'm concerned.

:cl:
tweak: Darkspawn are no longer a midround.
/:cl:


Cartlord:
just realized i'd probably have to undo the other pull request too, whoops

PR #18079, opened 2023-02-25
Adds tiny fans to Arrivals shuttles

Aquizit:

As title states. This allows for people who are arriving to the station not to have a shuttle full of toxic gases, or no air whatsoever.

image

# Changelog

:cl:  
mapping: Tiny fans added to all arrivals shuttle airlocks so people can breathe when late-joining.
/:cl:

### DIFF
--- a/_maps/shuttles/arrival_box.dmm
+++ b/_maps/shuttles/arrival_box.dmm
@@ -5,20 +5,6 @@
 "b" = (
 /turf/closed/wall/mineral/titanium,
 /area/shuttle/arrival)
-"c" = (
-/obj/machinery/door/airlock/titanium{
-	name = "Arrivals Shuttle Airlock"
-	},
-/obj/machinery/door/firedoor/border_only{
-	dir = 1
-	},
-/obj/machinery/door/firedoor/border_only,
-/obj/effect/turf_decal/stripes/line,
-/obj/effect/turf_decal/stripes/line{
-	dir = 1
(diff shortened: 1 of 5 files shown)


MegaEmpirical:
Nice.

Uristthedorf:
Another issue I see with arrivals shuttle is that if arrivals is depowered, people who arrive have to break the window to get out, anything you could do about that?

Aquizit:
> Another issue I see with arrivals shuttle is that if arrivals is depowered, people who arrive have to break the window to get out, anything you could do about that?

Unfortunately no, as that has to do with station power and how that all works. It's not the shuttle doors in that case that are the issue, but rather the station airlocks.

patpol4:
> Another issue I see with arrivals shuttle is that if arrivals is depowered, people who arrive have to break the window to get out, anything you could do about that?

Thats why they have crowbars on the shuttle

MegaEmpirical:
Isn't the crowbar only some of the time? Depending on the oxygen closest to randomly spawn one.

neeshacark:
i thought we already did this, wtf happened?

patpol4:
The crowbar is inside the shuttles

patpol4 (approved):
And we didnt do this yet, no

neeshacark:
it was because i made a forum post ages ago which is why this pr was made, i was thinking of that

PR #18081, opened 2023-02-25
fake bloodsucker lair 3x3 random maint

Marmio64:
A fake bloodsucker lair. Has a coffin, some of the building materials for a persuasion rack, and a bloodbag. Both terrifies maint goers and gives bloodsuckers some useful stuff if they explore maint and find it.
lair

# Changelog

:cl:  
rscadd: Adds bloodsucker lair random maint 3x3 template
/:cl:

### DIFF
--- a/_maps/RandomRuins/StationRuins/maint/3x3/3x3_lair.dmm
+++ b/_maps/RandomRuins/StationRuins/maint/3x3/3x3_lair.dmm
@@ -12,6 +12,12 @@
 /obj/structure/table_frame,
 /turf/open/floor/plating,
 /area/template_noop)
+"o" = (
+/obj/structure/closet/crate/coffin/metalcoffin{
+	anchored = 1
+	},
+/turf/open/floor/plating/broken/three,
+/area/template_noop)
 "E" = (
 /obj/item/restraints/handcuffs/cable,
 /obj/item/stack/sheet/mineral/wood,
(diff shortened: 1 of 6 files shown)


ynot01:
needs variations or else people will recognize this instantly, see if you can lower the weight so you can put a bunch without making it the only ruin

Marmio64:
> needs variations or else people will recognize this instantly, see if you can lower the weight so you can put a bunch without making it the only ruin

only idea i'd have for making it weighted would be to assign every other 3x3 template a higher weight, unless you can assign smaller than 1 weight values?
im not exactly sure on how no weight listed chances go, but i'd assume each one would have a 1 weight chance to get chosen? Or if it automatically calculates the percent chance each one has of getting picked and assigns that weight to them.

ynot01:
> only idea i'd have for making it weighted would be to assign every other 3x3 template a higher weight, unless you can assign smaller than 1 weight values?

stationroom templates uses pickweight() which supports non-integer weights

list items without values default to 1 so you can just add `= 0.1` to your variations for example

Marmio64:
lair2
lair3
lair4

including the first, there are now 4 lair variations. Each one has a 0.25 weight so the chance of getting a lair ruin is the same, but each variation has a 25% chance of being chosen pretty much.

ynot01:
Are the coffins anchored?

Marmio64:
> Are the coffins anchored?

no, i can do that too

tattax:
good pr

ToasterBiome (approved):
lol, lmao even.

ynot01 (approved):

PR #18087, opened 2023-02-26
Adds a magboot multisurgeon to CE's bag if they're digitigrade

Skrem7:

Not actually a bugfix for magboot implants (they're broken right now), but this multisurgeon works perfectly fine

# Changelog

:cl:  
rscadd: Adds a single-use multisurgeon to the CE's bag to grant magboots, in case the CE's legs aren't compatible with the advanced magboots (much like how the CMO gets the MEDhud autosurgeon in-case they're nearsighted)
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/engineering.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/engineering.dm
@@ -30,7 +30,6 @@
 	new /obj/item/storage/lockbox/medal/eng(src)
 	new /obj/item/barrier_taperoll/engineering(src)
 	new /obj/item/analyzer/ranged(src)
-	new /obj/item/multisurgeon/magboots(src)
 
 /obj/structure/closet/secure_closet/engineering_electrical
 	name = "electrical supplies locker"
--- a/code/game/objects/structures/crates_lockers/closets/secure/engineering.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/engineering.dm
@@ -30,6 +30,7 @@
 	new /obj/item/storage/lockbox/medal/eng(src)
 	new /obj/item/barrier_taperoll/engineering(src)
 	new /obj/item/analyzer/ranged(src)
+	new /obj/item/multisurgeon/magboots(src)
 
 /obj/structure/closet/secure_closet/engineering_electrical
 	name = "electrical supplies locker"
--- a/code/modules/jobs/job_types/_job.dm
+++ b/code/modules/jobs/job_types/_job.dm
@@ -268,9 +268,6 @@
 	/// What shoes digitgrade crew should wear
 	var/digitigrade_shoes
 
-	/// In case the crew member needs to spawn with something additional if they're digitigrade, often something to bypass their inability to wear normal shoes
-	var/digitigrade_backpack_contents
-
 /datum/outfit/job/pre_equip(mob/living/carbon/human/H, visualsOnly = FALSE)
 	switch(H.backbag)
 		if(GBACKPACK)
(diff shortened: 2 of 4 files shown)


Moltijoe:
> Not actually a bugfix for magboot implants (they're broken right now), but this multisurgeon works perfectly fine

PR to fix them has been open for over a week now
the absolute state of yogdev

Skrem7:
> > Not actually a bugfix for magboot implants (they're broken right now), but this multisurgeon works perfectly fine
> 
> PR to fix them has been open for over a week now the absolute state of yogdev

Didn't even notice it was opened lmao

Bibby0110 (dismissed):
Should only spawn if relevant for the CE

Skrem7:
> Should only spawn if relevant for the CE

Impossible to do considering midround joining, unless you want to have the autosurgeon on the CE themselves (bad design)

Bibby0110:
> > Should only spawn if relevant for the CE
> 
> Impossible to do considering midround joining, unless you want to have the autosurgeon on the CE themselves (bad design)

Have the auto surgeon spawn on the CE in case of them being a dumb species.
Nothing bad about that

Skrem7:
> Have the auto surgeon spawn on the CE in case of them being a dumb species. Nothing bad about that

if you say so

Skrem7:
@TheGamerdk there you go

goes without saying but i tested it just to be sure

Cartlord:
why do you keep buffing nonhumans

Skrem7:
> why do you keep buffing nonhumans

Could you please illustrate as many reasons as you can why you care?

Cartlord:
Why do you care so much about removing every meaningful downside from nonhumans? Is it because you play them?

Skrem7:
> Why do you care so much about removing every meaningful downside from nonhumans? Is it because you play them?

You're correct, once polysmorph CE's have a vulnerability to EMP roundstart, I plan to move on to actually giving non-humans more upsides. First, I'm going to make lizards do more damage with their fists, move faster, and have innate species armor (lore-accurate).

Next, I'm going to make it so moths can't have their wings burnt off. This way, they won't have downsides, too, so that makes three species (lizard, polysmorph, and moth) have zero total downsides (as they're just better than humans in every way). Not sure yet what I'm going to do for the rest.

And all of these terrible changes that threaten the playerbase catastrophically? It's because of me. My PRs. My lore. My terrible infatuation with non-humans. The same non-human characters that I play the game with (just like you play the game too, right?)

Moja10:
>coders add in snowflake feet that don't fit any shoes
>now we have to debate whether to add magboot autosurgeons because otherwise how can a person choose snowflake feet and not have their cake and eat it too

the yogswest has fallen

iamthesilly:
having them start in the locker was probably better, the CMO gets a medical HUD autosurgeon in their locker in case they're nearsighted so i don't see why the CE magboot implant should work any differently

Moltijoe:
> having them start in the locker was probably better, the CMO gets a medical HUD autosurgeon in their locker in case they're nearsighted so i don't see why the CE magboot implant should work any differently

and it'll end up used exactly like the CMO medhud, just given to a random engineer if not needed, and i think that's quite all right. it would also let a considerate CE give magboots to either a poly or lizard engineer that has digitigrade

iamthesilly:
alternatively, instead of adding an implant you could make it so you could adjust the advanced magboots to a digitigrade configuration and back with a wrench or screwdriver, though new sprites would be needed for it

Moltijoe (changes requested):
please change it back to the locker, this is no different than the CMO's medhud

Skrem7:
> please change it back to the locker, this is no different than the CMO's medhud

done

Moltijoe (approved):

PR #18115, opened 2023-02-27
Stops certain healing viruses from killing all jellypeople on the station

Moltijoe:
Starlight Condensation, Plasma Fixation, Radioactive Resonance, all of these virus symptoms heal toxin damage.
main problem is that a good healing virus that gets released upon the station can quickly kill all jelly people
this makes it just not do anything tox related to those species.

:cl:  
tweak: certain healing virus symptoms no longer kill jelly people
/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/heal.dm
+++ b/code/datums/diseases/advance/symptoms/heal.dm
@@ -86,7 +86,8 @@
 	if(M.getToxLoss() && prob(5))
 		to_chat(M, span_notice("Your skin tingles as the starlight seems to heal you."))
 
-	M.adjustToxLoss(-(4 * heal_amt)) //most effective on toxins
+	if(!TRAIT_TOXINLOVER in M?.dna?.species?.inherent_traits)
+		M.adjustToxLoss(-(4 * heal_amt)) //most effective on toxins
 
 	var/list/parts = M.get_damaged_bodyparts(1,1, null, BODYPART_ORGANIC)
 
@@ -416,7 +417,8 @@
 		if(prob(5))
 			to_chat(M, span_notice("You feel warmer."))
 
-	M.adjustToxLoss(-heal_amt * 2)
+	if(!TRAIT_TOXINLOVER in M?.dna?.species?.inherent_traits)
+		M.adjustToxLoss(-heal_amt * 2)
 
 	var/list/parts = M.get_damaged_bodyparts(1,1, null, BODYPART_ORGANIC)
 	if(!parts.len)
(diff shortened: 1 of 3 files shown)


Hopekz:
Why not have this as a down-side of the species?

Moltijoe:
> Why not have this as a down-side of the species?

there's a difference between down-side and accidental genocide, one over eager virologist making an airborne starlight condensation virus shouldn't kill people

Hopekz:
> there's a difference between down-side and accidental genocide

Won't be accidental now that it is a known quirk.
Will make viro have to think about the way they spread their virus vs just making the same virus over and over and having it spread on its own

Moltijoe:
> Won't be accidental now that it is a known quirk. Will make viro have to think about the way they spread their virus vs just making the same virus over and over and having it spread on its own

if someone doesn't think about something like that, i'd rather it not involve random people dying

Uristthedorf:
Slime people are probably the most powerful race there is, the weaknesses they do have shouldn't be mitigated. The toxin healing symptoms typically aren't too hard to avoid (If you know what's happening,) and you can go to medbay for a cure. Harmful diseases are too rare anyways.

Moltijoe:
eh, fair enough, so long as they make sure to put on suit sensors.
goes in a similar category as janitors needing to know that water vapour fucks up preterni

Hopekz:
> if someone doesn't think about something like that, i'd rather it not involve random people dying

It's a game the goal isn't to save lives 🤠
As long as people know why they're dying then it should be a non-issue. The only death I find cheap is if there is no indication on why you are taking damage and in this case it's about as clear as pod people taking damage from drinking alcohol.

In this case they start taking damage as soon as they get the disease so it should be easy to know that the virus caused it

Moltijoe:
> It's a game the goal isn't to save lives 🤠 As long as people know why they're dying then it should be a non-issue. The only death I find cheap is if there is no indication on why you are taking damage and in this case it's about as clear as pod people taking damage from drinking alcohol.
> 
> In this case they start taking damage as soon as they get the disease so it should be easy to know that the virus caused it

problem is, it's not always intuitive, sometimes a virus needs to progress before it heals enough to be dangerous, or it only heals under certain conditions.
sometimes, the virus spreads so much that they catch it by passing a random assistant while in maints rather than by drinking a virus bottle outside of medbay

Moltijoe:
eh, i'll split the tumors and the viruses into two PRs

MajManatee:
id like to see everyone that makes an airborne virus have a very real risk of getting banned from virologist, so I am against such a change

Moltijoe:
i might just nerf starlight fixation to no longer heal toxins, and leave the other two alone since they're more rare

Hopekz:
> i might just nerf starlight fixation to no longer heal toxins

Ye probably the better way since just making it behave differently for a single species is just inconsistent game design imo.

Someone could be like "okay my target is a jelly person. X virus does Y which means They don't like X virus. I'll make a really powerful one to assassinate them for my objective"

And then when they actually implement their thought process they learn that there's weird inconsistent hugbox behavior.

Moja10:
shitty ass jellypeople shouldn't be a roundstart race if it means removing good things just to compensate for their shitty existence

PR #18125, opened 2023-02-27
Tumors are no longer infinite healing to jelly people

Moltijoe:
tumors are ridiculously good on jelly people since they heal brute and burn, converting it into toxin 
but with their inverted toxin damage that just means all the tumors do is heal

:cl:  
tweak: Tumors no longer convert damage for toxin lovers
/:cl:

### DIFF
--- a/code/modules/surgery/organs/tumors.dm
+++ b/code/modules/surgery/organs/tumors.dm
@@ -37,8 +37,8 @@
 	if(!(src in owner.internal_organs))
 		Remove(owner)
 	if(helpful)
-		if(owner.getBruteLoss() + owner.getFireLoss() > 0)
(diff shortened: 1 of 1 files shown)


MajManatee:
Perk of the race innit? Same way they die to the normal viruses.

ThatLing:
> tumors are ridiculously good on jelly people since they heal brute and burn, converting it into toxin

Could we make it not convert for toxin lovers instead? Forcing toxin damage seems a weird interaction imo when your species is advertised as "toxin heals u"

Moltijoe:
> Could we make it not convert for toxin lovers instead? Forcing toxin damage seems a weird interaction imo when your species is advertised as "toxin heals u"

i suppose that works

ThatLing (approved):

PR #18126, opened 2023-02-27
Starlight fixation tweak + Toxolysis buff

Moltijoe:
Starlight fixation is used in a lot of viruses, it heals every damage type with no real downside aside from needing to be close to a window.
Toxin damage usually tends to be harder to heal, so being able to trivialize it just by taking a sip of a virus culture is a bit much
This moves toxin healing over to toxolysis (i have literally never seen this used)

also the removal of toxin healing from starlight fixation lets it be used without killing all jelly people on the station

:cl:  
tweak: Toxolysis heals toxin damage
tweak: Toxolysis resistance threshold now increases toxin heal rather than chem purge speed
tweak: Starlight fixation no longer heals toxin damage
tweak: Starlight fixation heals 2x as much non-toxin
/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/heal.dm
+++ b/code/datums/diseases/advance/symptoms/heal.dm
@@ -111,7 +111,7 @@
 	var/food_conversion = FALSE
 	desc = "The virus rapidly breaks down any foreign chemicals in the bloodstream. It also heals toxin damage."
 	threshold_descs = list(
-		"Resistance 7" = "Increases chem removal speed.",
+		"Resistance 7" = "Increases toxin healing speed.",
 		"Stage Speed 6" = "Consumed chemicals feed the host.",
 	)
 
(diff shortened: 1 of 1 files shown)


TheBombMakerJL:
This just makes it into tissue hydration but for space
Toxin healing is what made starlight condensation good, and the drawbacks of toxolysis will mean it’s never used

Moltijoe:
> This just makes it into tissue hydration but for space Toxin healing is what made starlight condensation good, and the drawbacks of toxolysis will mean it’s never used

after your discord comment, i tweaked it to just be a universal healing buff
i've also made toxolysis not purge chems faster with resistance, and instead just heal toxin faster

ThatLing (approved):

Misanthropy-AU:
have my concerns about changing toxolysis to a constantly ticking toxins heal. throwing this in with tumors is going to be exceptionally powerful whilst removing any downside the tumors had, as it would very easily heal the toxins dealt. statwise is a non-issue, could easily fit toxolysis, rapid protein synth or self respiration and tumors all in one while maintaining stat thresholds via other symptoms. just personally feel it's going to be too powerful.

PR #18127, opened 2023-02-27
Bible healing tweaks (nerfs?)

Moltijoe:
Them healing every single bodypart for a set amount means it can heal between 10 and 60 depending on how injured the person is in each limb, this makes it stupidly op against damage that either hits every part equally or hits random bodyparts, while being almost useless against focused damage.

This is made all the worse for stronger bibles like capitalism (30 per BP) or holy light (20 per BP) where it's enough to bring someone from crit to full health if the damage is distributed properly.

It has been changed to instead apply an overall body heal for a set amount, increasing consistency of the heal and allowing for more precise healing balance tweaks.

if this ends up being too much of a nerf to the holy light bible favour gain, changes can be made.

:cl:  
tweak: Bibles now heal overall damage rather than per bodypart
tweak: Default bible heals 20 brute and burn overall instead of 10 per bodypart
tweak: Tech bible heals 20 brute and burn overall overall instead of 10 to a single targeted bodypart
tweak: Plant bible heals 20 brute and burn overall instead of 10 per bodypart
tweak: Candle bible heals 40 BURN overall instead of 10 per bodypart
tweak: Clown bible heals 20 brute and burn overall instead of 10 per bodypart
tweak: Capitalism bible heals 60 brute and burn overall instead of 30 per bodypart
tweak: Holy light bible heals 40 brute and burn overall instead of 20 per bodypart
tweak: Holy light bible never fails to heal if it's off cooldown
/:cl:

### DIFF
--- a/code/game/objects/items/storage/book.dm
+++ b/code/game/objects/items/storage/book.dm
@@ -105,14 +105,11 @@ GLOBAL_LIST_INIT(bibleitemstates, list("bible", "koran", "scrapbook", "burning",
 			to_chat(user, span_warning("[src.deity_name] refuses to heal this metallic taint!"))
 			return 0
 
-	var/heal_amt = 10
-	var/list/hurt_limbs = H.get_damaged_bodyparts(1, 1, null, BODYPART_ORGANIC)
+	var/heal_amt = 20
 
-	if(hurt_limbs.len)
-		for(var/X in hurt_limbs)
-			var/obj/item/bodypart/affecting = X
-			if(affecting.heal_damage(heal_amt, heal_amt, null, BODYPART_ORGANIC))
-				H.update_damage_overlays()
+	if(H.getBruteLoss() > 0 || H.getFireLoss() > 0)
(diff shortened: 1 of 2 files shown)


MajManatee:
why does capitalism heal more than the holy light literally healing sect??

Moltijoe:
> why does capitalism heal more than the holy light literally healing sect??

good question, it already did so i left it as is.
i assume it was because it consumes money and has a long cooldown

MajManatee:
It has a 15 second cool down (apparently normal healing doesn't have one at all) and you have to have at least 1 credit in your account for being healed. That DEFINETLY should be so high.

Moltijoe:
> (apparently normal healing doesn't have one at all)

that's the thing that threw me off, because i wanted things like tech, clown, plant, and candle to be stronger, but when i couldn't find cooldowns for them, i reduced the healing since otherwise it would be too much

Bibby0110 (approved):

PR #18128, opened 2023-02-27
Improves improved Lizard and Polysmorph lisp

Moltijoe:
X doesn't always make an "ecks" sound, it it's the first letter of a sentence it usually makes a "z" sound
i also removed an s so it has the same number of character as the s lisp, just with one being a k

:cl:  
tweak: Lizards and Polys only replace x with kss if it's in the middle of a sentence
tweak: Lizards and Polys x lisp now has 1 less s
/:cl:


MajManatee (approved):
Thank god

PR #18147, opened 2023-03-01
Lets the holotool repair mechs

Moltijoe:
They don't actually have fuel to use

:cl:  
bugfix: Lets the holotool repair mechs
/:cl:


ynot01 (changes requested):
Override /obj/item/holotool/tool_use_check() instead

ynot01 (approved):

PR #15069, opened 2022-07-30
Disabler Nerf Attempt #2

Skrem7:
# Document the changes in your pull request

Reduces the damage of the disabler considering security is getting a lot of other toys and gadgets and theoretically less-lethals should maybe be getting more buffs down the line (rubber pellets even better, mayhaps).

Plus the thing has TWENTY SHOTS. You can use an ebola or something else there are COUNTLESS other options and disabler doing more damage than a laser is still good.

"Why don't I just use a laser" my guy you are security it is your job to detain stop being so unrobust and learn to aim you LITERALLY get a shield on your belt you can just use a baton if you have to, they're still good. Sec has been in one of the best spots it's been in a while I would like to see antags get to exist

Only thing is that it makes armor actually matter more lmao

# Wiki Documentation

Disabler beam now does 25, instead of 30. Most relevant for disabler and egun.

# Changelog

:cl:  
tweak: Disabler beam does 25 stamina instead of 30
/:cl:

PR #18149, opened 2023-03-01
Revert "Reverts #15069 properly this time because im a stupid idiot commiting to master"

jachlompsky:

Disablers go from 30 stamina -> 25 stamina
Also, NT-USP goes from 25 stamina -> 22 stamina

Someone else made a PR to do this a while ago but it got closed due to feature freeze, so:
The reasoning behind the most recent 25 -> 30 change was "the amount needed to down someone before was way too high."
I saw people mentioning a bug causing a delay before stamcrit stun around that time, which is likely the cause, because 30 to 25 is virtually unnoticeable. Unarmored people still take 4 shots, and there are only 5 armors in the entire game (that I counted) that require more than 5 disabler shots, WITH the 25 damage change:
-Mk IV Sec vests (6 shots)
-Elite Syndicate Hardsuit (6 shots)
-Heretic Robes (6 shots)
-Reflective Jacket (7 Shots, more if beams get reflected)
-ERT Hardsuits (7 shots)

Considering disablers resolve against energy armor, nearly always the lowest armor value, this isn't even that much of a change. Unless you have the equipment above, any armor is going to let you take only one extra hit. By comparison shotgun slugs, fired point blank, crit someone in an elite hardsuit in the same number of shots as a 25 damage disabler. 30 damage disablers take LESS shots than slugs. There is no reason security should be able to do this much effective damage, 20 times before recharging, at long distance. Regardless I'm only going to push for 25 damage right now so that this has a chance of being merged.

Images for reference, against elite syndicate hardsuit, with shotgun slugs (46 damage, point-blank) and 30 damage disabler.
image
image

# Changelog

:cl:  
tweak: Disablers go from 30 stamina -> 25 stamina
tweak: NT-USP go from 25 stamina -> 22 stamina
/:cl:

### DIFF
--- a/code/modules/projectiles/boxes_magazines/external/rechargable.dm
+++ b/code/modules/projectiles/boxes_magazines/external/rechargable.dm
@@ -86,7 +86,7 @@
 	flag = ENERGY
 	damage = 2 //ouch ouch my skin ouchie
 	damage_type = BURN
-	stamina = 20
+	stamina = 22
 	speed = 0.55
 
 //NT-USP Lethal Clip
--- a/code/modules/projectiles/boxes_magazines/external/rechargable.dm
+++ b/code/modules/projectiles/boxes_magazines/external/rechargable.dm
@@ -86,7 +86,7 @@
 	flag = ENERGY
 	damage = 2 //ouch ouch my skin ouchie
 	damage_type = BURN
-	stamina = 25
+	stamina = 20
 	speed = 0.55
 
 //NT-USP Lethal Clip
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -85,7 +85,7 @@
 /obj/item/projectile/beam/disabler
 	name = "disabler beam"
 	icon_state = "omnilaser"
-	damage = 30
+	damage = 25
 	damage_type = STAMINA
 	flag = ENERGY
 	hitsound = 'sound/weapons/tap.ogg'


ynot01:
Not that I agree with any of this, but you'll want to modify the NT-USP aswell

jachlompsky:
> Not that I agree with any of this, but you'll want to modify the NT-USP aswell

thought the USP was 20 stamina cause thats what the PR for it said, but looks like even in that initial PR the code said 25. Just changed that too.

Mqiib:
Note that stamina damage and regular damage don't really stack, so the USP does effectively just 20 damage or 2 damage, not 22.

MegaEmpirical:
You monster

ynot01:
Reverts must go through HD

ynot01 (changes requested):
Balance council says make NT-USP damage 22

jachlompsky:
> Balance council says make NT-USP damage 22

Changed

ynot01 (approved):

JamieD1:
Just going to say if you are saying "its not even that much of a change" then why are you needing to change it?

jachlompsky:
> Just going to say if you are saying "its not even that much of a change" then why are you needing to change it?

Because having 25 damage means that it does its job the same against typical crew, but it takes 1-2 more shots against someone armored or using stamina chems, so their armor isnt just useless. Besides, it would never get merged if it was a bigger change, and the previous 25-damage-disabler PR must've been enough change cause it was reverted.

ToasterBiome:
HDs don't feel this is needed. Balancing sec vs antagonists is a delicate balance and the added 1-2 hits gives a bit too much extra advantage to antagonists in the current state of the game.

This is something that can be looked towards in the future once we fix some blatant problems that are currently occurring with sec vs antagonist balancing.

PR #18159, opened 2023-03-02
Toxolysis requires toxins to heal at full power

Mqiib:
Giving free, perpetual, passive toxin healing causes problems (see: podpeople and radiation/zombie tumors).

This changes toxolysis to only heal 0.1 (0.2 with threshold) toxin per heal tick UNLESS there are toxins present in the person, in which case it heals the full 1 or 2 damage. Thank you to triskund for bringing the issue to my attention.

# Wiki Documentation

Toxolysis has very weak passive healing that increases significantly when toxins are present in the blood stream.

# Changelog

:cl:  

tweak: Toxolysis passive healing is significantly reduced unless toxins are present in the bloodstream.

/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/heal.dm
+++ b/code/datums/diseases/advance/symptoms/heal.dm
@@ -125,14 +125,18 @@
 		power = 2
 
 /datum/symptom/heal/chem/Heal(mob/living/M, datum/disease/advance/A, actual_power)
-	var/heal_amt = actual_power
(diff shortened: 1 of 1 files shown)


TheBombMakerJL:
are you taking into consideration the fact that toxins in your system give you toxin damage

Mqiib:
> are you taking into consideration the fact that toxins in your system give you toxin damage

No, it never occurred to me that toxins will do toxin damage to you.

Bibby0110 (approved):

PR #18162, opened 2023-03-03
adds the Emperor stand

tattax:
ftpXaGFV78

image

it's kinda sad we don't have many objects stands despiting them having potential to be cool.
just like hol horse this little buddy of yours has the advantage of having infinite ammo, but it's susceptible to balancing thanks to stats and obviously has the liability of getting a bad player as your holo.

hol can go into gun form at will, i can add a cooldown if it's too much of a problem.

also if you think you can shoot bullets through yourself remember:

<details>
    <summary>Spoiler</summary>

bundle adds a reusable holo injector + mindslave (if your current partner dies)

image

</details>

if you spawn the item by itself it'll be just basically a better .357 (infi ammo)

# Spriting

<details>
    <summary> </summary>

image

image

image

image

image

image

</details>

# Wiki Documentation
New stand: Precision (Emperor) costs 3 points, 
X being our stat level
* bullet damage | AP scales with damage, (10\*X | 5\*X, respectively)
* bullet penetration and type scales with defense, (X-1 | rounddown(X / 2), respectively)
* run away boost and cooldown from recalling scales with potential, (-0.25\*X | 20s / X, respectively)
* run away max duration and fire delay scales with speed, (2s\*X | 15 / X, respectively)
* range gives the gun more bullet decay range (7.5\*X)

# Changelog

:cl:  Baiomu, tattax
rscadd: new Emperor stand type
rscadd: adds hol horse to the game (clothes of)
rscadd: yellow cowboy bundle for special syndikit
/:cl:


Moltijoe:
the entire horse?

MajManatee:
@cowbot92

UltraBigRat:
Could maybe give the bundle more stuff? since right now its basically just stuff you could buy with 20tc anyways and unarmored  drip. (could maybe give them two holopara injectors and a mindslave implant because hol horse always has a partner.) So its not just a syndi kit you could buy individually with 20tc

Moja10:
Good PR, however it would have been hilarious if the gun was controlled by the user, and the bullet was controlled by the stand. Is it possible that this loop gets added?
1. user fires gun, gun goes on CD
2. stand moves from holopara mob to possessing bullet, can move as bullet
3. upon the bullet impacting a wall or person, stand client moves back to stand
4. user can fire gun again

tattax:
> Good PR, however it would have been hilarious if the gun was controlled by the user, and the bullet was controlled by the stand (...)

that was the original plan but i think it would feel too sluggish and disorientating for both sides

yogstation/code/modules/guardian/abilities/major/precision.dm
```diff
@@ -0,0 +1,118 @@
...
+	3 = -1,
+	4 = -1.25,
+	5 = -1.5
+)) //doing this for consistency's sake
+
+/mob/living/simple_animal/hostile/guardian/ClickOn(atom/A, params)
+	if(istype(loc, /obj/item/gun/ballistic/revolver/emperor))
+		var/obj/item/gun/ballistic/revolver/emperor/gun = loc
+		var/turf/turf_to_throw_gun = get_ranged_target_turf(gun, get_dir(A, gun), 1)
+		if(gun.process_fire(A, src, TRUE, params) && range > 1) //melee cannot go further if they get dropped on the ground
```
ynot01: > range gives the gun more potential to shoot by itself

This is the only place that the `range` stat is considered and it appears to be able to fire itself just fine without any range? What exactly does this do anyways? Throws the gun at what it fires at? Seems very weird

ynot01:
After brief review, this is WAY too strong.

Putting A in damage means you get a unlimited ammo gun that fires 50 damage with 25 AP

Putting C in potential brings you below the 0.4 ranged click CD, meaning it's strictly better than the revolver and still never reloading

After spending 3 points on the ability itself and then 6 points on the above stats, you still have 6 more points to work with for other utility like bluespace pad, or defense for mob penetrations, or speed for more free speed boosts

The syndie-kit is also forced and pointless, as it presents nothing of value that you cannot normally get, just a holoparasite injector (12TC), adrenals (8TC), and syndie smokes (2TC)

This needs some serious adjustments

#

Other than that, since the holoparasite speed stat affects how fast the guardian HITS and not MOVES, speed should be affecting fire delay, and I would suggest moving the rest of the "run away" stats to potential

Range should affect how far the bullet goes, this should be fairly easy to implement and if you make it short ranged, it will make this as a whole much more balanced, as by default bullets travel up to 50 tiles

yogstation/code/modules/guardian/abilities/major/precision.dm
tattax: if you have no range, you are bound to your summoner, therefore if you are dropped in the ground with F range you won't be able to shoot yourself out of that range

ynot01:
Please update PR comment with changes you made; it would also be helpful if you listed the actual stats and numbers instead of "it gets stronger as damage gets stronger"

tattax:
> Please update PR comment with changes you made; it would also be helpful if you listed the actual stats and numbers instead of "it gets stronger as damage gets stronger"

i will update when i'm done

cuackles (approved):

bruhlookatthisdood:
give it aimbot

ChuChu42:
empor

ynot01 (changes requested):
Balance council says too strong; just a really strong unlimited-ammo revolver

Moltijoe:
unless you have plans to nerf it a good bit, it'll be closed in 24 hours

Moltijoe:
unrelated to the balance council, my personal opinion is that if you make it based off the original idea (where the holopara player controlled the bullet) that would allow it to be more powerful because it would be harder to use properly.
it would also give it a different niche than regular guns or a ranged holoparasite

tattax:
if balance council denies it there's nothing i can do then
i could tweak numbers but that's how far i'm willing to go at the moment.
i'll guess i'll reopen when i can work on this or/and if balance council decides to reevaluate

Hislyna:
Couldn't this work as a requiem stand? I know its not in Jojo's, but that could make it balanced enough perhaps?

GraveHat:
> Couldn't this work as a requiem stand? I know its not in Jojo's, but that could make it balanced enough perhaps?

It would work as a specific upgrade from the nerfed one yeah, the problem is if this is already syndicate only then getting an actual arrow to upgrade it would be absurdly rare and would be kinda lame if it was just a stat increase

PR #18168, opened 2023-03-03
Removes the Xenomorph eggsac from meteors because I died to it once.

BurgerLUA:

Xeno eggsacs cannot spawn from meteors anymore.

# Justification

Too many a time xeno fetish players insert the organs into themselves and lay a bunch of eggs and then act surprised when the station results in a xeno takeover. There is literally no reason to do this as a non-antag, and as an antag no one does this anyways because a shuttle hijacking requires you to be on it and the xenos will not be aligned to you so antags don't do it either.

# Changelog

:cl:  BurgerBB
rscdel: Xeno eggsacs cannot spawn from meteors.
/:cl:

### DIFF
--- a/code/game/gamemodes/meteor/meteors.dm
+++ b/code/game/gamemodes/meteor/meteors.dm
@@ -324,7 +324,7 @@ GLOBAL_LIST_INIT(meteorsC, list(/obj/effect/meteor/dust)) //for space dust event
 	meteorgibs = /obj/effect/gibspawner/xeno
 
 /obj/effect/meteor/meaty/xeno/Initialize()
-	meteordrop += subtypesof(/obj/item/organ/alien)
+	meteordrop += (subtypesof(/obj/item/organ/alien) - /obj/item/organ/alien/eggsac) //Yogstation change: No more eggsac.
 	return ..()
 
 /obj/effect/meteor/meaty/xeno/ram_turf(turf/T)


ThatLing:
>There is literally no reason to do this as a non-antag

admin issue

Moja10:
The intention is not to have someone deliberately and consciously get infected or infect others, the intention is to literally just recreate the plot of the 1976 Alien movie, where incidentally and unknowingly get infected, and that causes a big incident on the ship/station. Noob bait but programmed into the game to cause station wide destruction. Fully intended and admin approved. 

Not admin approved to deliberately infect others and cause a hijacking unless you have hijack/die glorious, but if we were to code out things based on people doing them without the objectives to justify, the first thing to go would be plasmaflooding, as that's 100x worse, 100x easier, and 100x more common than this, and as you can tell, it hasn't been coded out of the game, because admins do just fine making sure that shenanigans of hijack with no hijack obj don't happen. Code really doesn't need to change to assist this goal at the moment whatsoever.

BurgerLUA:
Except the difference here is that a traitor with hijack infecting others with xenomorph stuff wouldn't work out well because Xenomorphs are supposed to hijack the shuttle to ensure no non-xenomorphs are on the ship unless you plan to singlehandedly exterminate all the xenos and the queen somehow.

With plasmaflooding, there is less of a barrier to that.

BurgerLUA:
As for the Aliens movie remark, when you implant yourself with egg sacs to lay eggs, you are deliberately choosing to infect crew. This isn't like a random egg spawn or a lavaland incident; this is a crewmember choosing to basically have a potential xeno round by implanting egg-sacs into them.

Moja10:
Ah okay, I misread the PR. You are talking about the organ being found in space, not the whole egg object that spits out a hugger. Ignore my previous point.

Bibby0110:
Admin issue. Just because you have something doesn’t mean you can use it

PR #18072, opened 2023-02-25
Handheld flashes knockdown carbons instead of hardstun

ynot01:
# Document the changes in your pull request

Revival of #17036

Flashes now apply 8-12 seconds of knockdown or 2-3 seconds if already knocked down/paralyzed. Flashes still apply 15 confusion stacks and blind on hit.

Clicking someone once and them not being able to at least struggle is bad gameplay

# Changelog

:cl:  
tweak: Handheld flashes now knockdown instead of paralyze
/:cl:

PR #18192, opened 2023-03-06
Revert "Handheld flashes knockdown carbons instead of hardstun"

stwordcitizen:
Reverts yogstation13/Yogstation#18072
this pr had a negative like to dislike ratio
It also makes security a pain in the ass for no reason at all and now the brig procedure isn't correct and batons can be easily taken and used against you while you are trying to brig someone. The old way to brig someone was to flash them to take their cuffs off while they're buckled to the bed and now you have to baton someone over and over to safely take their cuffs off which can easily get you killed if you make one mistake which is making sec less appealing to newer players/those not mechanically gifted at spaceman game
This pr should be reverted because its bugging me irl thank you

### DIFF
--- a/code/modules/assembly/flash.dm
+++ b/code/modules/assembly/flash.dm
@@ -148,10 +148,10 @@
 					trauma.on_flash(user, M)
 			else
 				to_chat(M, span_userdanger("You are blinded by [src]!"))
-			if(M.IsParalyzed() || M.IsKnockdown())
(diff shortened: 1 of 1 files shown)


Marmio64:
you can still use pepperspray, i still dont know how i feel about flashes otherwise

Moltijoe:
shoving a person on the ground stuns them

Bibby0110:
48 hour rule

PR #18193, opened 2023-03-06
Adds perk descriptions to Preterni in the preference menu

Moltijoe:
Should give people an idea of what they're getting into when playing a preternis without needing to look at the wiki

that's a lotta perks
image

:cl:  
rscadd: Preterni preference perk descriptions
/:cl:


ThatLing (changes requested):
Also add ``TRAIT_RADIMMUNE``

Moltijoe:
> Also add `TRAIT_RADIMMUNE`

ah, good call, i'll also add that to inherent so ipc and plasmemes get it

ThatLing (approved):

PR #18194, opened 2023-03-06
Fixes a mech melee bug that I didn't actually make sure was fixed

Mqiib:
You shouldn't be able to stab open turfs any more. This means the spear and katana are a LOT more accurate because they wont keep going on cooldown if you misclick.

bugfix: Significantly increases accuracy of precise-only mech melee weapons  

/:cl:


isotope436:
mqiib breaking stuff like usual

Moltijoe (approved):

PR #18195, opened 2023-03-06
Actually needs 5u for vaccines rather than 5.01

Moltijoe:
\> != >=

:cl:  
bugfix: needs 5u for vaccines rather than 5.01
/:cl:


yacabo:
Comedy

Bibby0110 (approved):

PR #18204, opened 2023-03-07
Scrubs blindfold of telepathy from the code

Moltijoe:
They don't work, the blind overlay covers the x-ray vision, meaning all they do is blind you and let you see people if you scrub your face against a wall
I can't for the life of me figure out how to make them work in a way that isn't disastrous to performance
even then, they don't mesh well with what miners do, they need to see to mine
they aren't even worth handing off to someone station-side because all they're really good for is validhunting (assuming they worked)

if i can figure out how to make them work i'll re-add them, probably as a nullrod instead though.

:cl:  
rscdel: blindfold of telepathy
/:cl:


MajManatee (approved):
sucked anyways.

PR #18205, opened 2023-03-07
Hand teleporter portals now take 4 seconds to dispel

SomeguyManperson:

Prevents people from simply instantly teleporting 200 tiles and removing any way to easily follow them outside walking, which basically allows a risk-free get out of jail free card when in a fight you don't want to be in
now the exit location has to actually be defensible or whatever rather than just a long distance

# Wiki Documentation

Undocumented, probably, hand teleporter portals can be hit with the hand teleporter to dispel them. This takes four (4) seconds.

# Changelog

:cl:  
tweak: hand teleporter portals now take 4 seconds to dispel, up from 0.
/:cl:

### DIFF
--- a/code/game/objects/items/teleportation.dm
+++ b/code/game/objects/items/teleportation.dm
@@ -148,9 +148,11 @@
 
 /obj/item/hand_tele/proc/try_dispel_portal(atom/target, mob/user)
 	if(is_parent_of_portal(target))
-		qdel(target)
-		to_chat(user, span_notice("You dispel [target] with \the [src]!"))
-		return TRUE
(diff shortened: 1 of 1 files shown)


MajManatee:
like 50% of the playerbase didnt know this was a feature

ynot01:
next make entering the portal take 2 seconds so you cant play scooby doo with the portal(s) going back and forth to avoid combat until it dissipates

Therandomhoboo:
Four seconds seems a bit extreme honestly, I would suggest putting it down to One or two seconds cause at least it gives a fair chance to both sides.

Moltijoe (approved):

PR #18208, opened 2023-03-07
Cerulean extract no longer saves from dusting or gibbing

Moltijoe:
it was supposed to do this already but sometimes the ticks would line up such that it would still work
adds a 0.4 second delay to account for such events
this means walking into a SM or dying with a holopara will still kill you

:cl:  
tweak: Cerulean extract no longer saves from dusting or gibbing
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
+++ b/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
@@ -716,9 +716,14 @@ datum/status_effect/stabilized/blue/on_remove()
 		C.real_name = O.real_name
 		O.dna.transfer_identity(C)
 		C.updateappearance(mutcolor_update=1)
+		RegisterSignal(owner, COMSIG_GLOB_MOB_DEATH, .proc/dead)
 	return ..()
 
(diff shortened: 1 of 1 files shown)


ChuChu42:
>dying with a holopara

this PR was made right after a round where a hijack IAA was saved from dusting by cerulean, this is a ided PR
bugfix or not.

Moltijoe:
> this PR was made right after a round where a hijack IAA was saved from dusting by cerulean, this is a ided PR bugfix or not.

It was actually a personal request made by the hijacking IAA player

Maia-J:
> > this PR was made right after a round where a hijack IAA was saved from dusting by cerulean, this is a ided PR bugfix or not.
> 
> It was actually a personal request made by the hijacking IAA player

Proof or it didn't happen

Moltijoe:
> Proof or it didn't happen

bet
image
image

MajManatee:
I mean it probably makes sense, since the slime itself would be destroyed so the clone is also destroyed.
That being said, I hate it

ToasterBiome (approved):

PR #18210, opened 2023-03-08
Slime extract grenades now make sense

MajManatee:
Imagine the following
you have spent an hour making slimes
you go to make a great fun grenade with your slimes
you pull the pin throw it, wait 5 seconds...
and the core pops out... and you wait another 5 seconds.
your target have all ran away and your moment is LONG gone.

no longer.

Large grenades now behave as you would expect, and begin the reaction when you prime it.
They are now also printable from the RnD Protolaithe

# Changelog

:cl:  
tweak: Slime core grenades no longer take twice as long as they should have.
tweak: Large grenades now printable in RnD
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes/slime_extracts.dm
+++ b/code/modules/reagents/chemistry/recipes/slime_extracts.dm
@@ -699,7 +699,7 @@
 		S.icon_state = "[S.base_state]_active"
 		S.active = TRUE
 		var/obj/item/slime_extract/M = holder.my_atom
-		addtimer(CALLBACK(S, /obj/item/grenade.proc/prime), rand((M.react_time - 3),(M.react_time + 2)))
+		addtimer(CALLBACK(S, /obj/item/grenade.proc/prime), rand((M.react_time - 35),(M.react_time + 10)))
 		qdel(holder.my_atom) //deleto
 	else
 		var/mob/living/simple_animal/slime/random/S = new (get_turf(holder.my_atom))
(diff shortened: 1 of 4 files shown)


code/modules/reagents/chemistry/recipes/slime_extracts.dm
```diff
@@ -712,7 +718,8 @@
 	S.visible_message(span_danger("Infused with slime jelly, the core begins to expand uncontrollably!"))
 	S.icon_state = "[S.base_state]_active"
 	S.active = TRUE
-	addtimer(CALLBACK(S, /obj/item/grenade.proc/prime), rand(15,60))
+	var/obj/item/slime_extract/M = holder.my_atom
+	addtimer(CALLBACK(S, /obj/item/grenade.proc/prime), rand((M.react_time - 3),(M.react_time + 2)))
```
ynot01: This results in rand(47, 52) not rand(15,60)

ynot01 (dismissed):

code/modules/reagents/chemistry/recipes/slime_extracts.dm
MajManatee: Zamn

UltraBigRat:
should probably also give sci lathe  the abilty to make large nades since currently even if they worked only med can actually make nades that fit    slime cores i believe.

Moltijoe (approved):

PR #18212, opened 2023-03-08
Bee kit buff

jachlompsky:

Improves the bee kit by improving the Stinger™ and beesease. 

The Stinger™ currently does a measly 7 damage, but has a whopping 85 armor penetration, fast attack speed, and injects 4u histamine each hit. The issue is, a fast attack isn't much help against a moving and dodging target. Also histamine sucks and is useless.

Now the Stinger™, instead of injecting 4u histamine, will now inject 2u each of sulfuric acid, tirizene, and cryptobiolin. The acid because that's what bees inject and it does minor toxin; tirizene causes high, but decaying stamina damage, so your target is slowed down after a bit but doesn't get totally disabled; and the cryptobiolin causes dizziness/confusion so your target can't avoid you/fight back as easily. Trying to make it debilitating without actually stunning, not sure how balanced this is.

The Stinger™ can also now be placed in the exosuit slot of the bee costume instead of just taking up the belt slot, so you can actually wear the yellow fanny pack you get if you really want to.

Finally, beesease has its probability to spawn a bee changed from 1 to 3, and the stomach buzz message is a bit less common and no longer in huge span_userdanger text (because it's annoying.) Beesease is cured super easily, its negative effect occurs rarely, and at worst you'll get a bunch of sulfuric acid from bee stings and have to head to medbay for the 1 toxin/tick you're receiving. A probability of 3 is still pretty negligible, and normal bees aren't very dangerous to the crew, but it should give you a few more bee friends. The carp kit has 9 carp-producing items so I think the bee kit should have a few more bees buzzing around.

# Wiki Documentation

I don't think the wiki specifies info about anything this changes

# Changelog

:cl:  
tweak: beesease messages less annoying and bee barf probability increased
tweak: the Stinger now injects sulfuric acid+tirizene+cryptobiolin instead of histamine and can be stored on the bee costume
/:cl:

### DIFF
--- a/code/datums/diseases/beesease.dm
+++ b/code/datums/diseases/beesease.dm
@@ -27,12 +27,12 @@
 					affected_mob.adjustToxLoss(2)
 					affected_mob.updatehealth()
 		if(4)
-			if(prob(10))
+			if(prob(7))
 				affected_mob.visible_message(span_danger("[affected_mob] buzzes."), \
-												span_userdanger("Your stomach buzzes violently!"))
+												span_danger("Your stomach buzzes violently!"))
 			if(prob(5))
 				to_chat(affected_mob, span_danger("You feel something moving in your throat."))
-			if(prob(1))
+			if(prob(3))
 				affected_mob.visible_message(span_danger("[affected_mob] coughs up a swarm of bees!"), \
 													span_userdanger("You cough up a swarm of bees!"))
 				new /mob/living/simple_animal/hostile/poison/bees(affected_mob.loc)
--- a/code/game/objects/items/melee/misc.dm
+++ b/code/game/objects/items/melee/misc.dm
@@ -165,7 +165,9 @@
 	user.changeNext_move(CLICK_CD_RAPID)
 	if(iscarbon(target))
 		var/mob/living/carbon/H = target
-		H.reagents.add_reagent(/datum/reagent/toxin/histamine, 4)
+		H.reagents.add_reagent(/datum/reagent/toxin/acid, 2)
+		H.reagents.add_reagent(/datum/reagent/toxin/staminatoxin, 2)
+		H.reagents.add_reagent(/datum/reagent/cryptobiolin, 2)
 
 /obj/item/melee/beesword/suicide_act(mob/living/user)
 	user.visible_message(span_suicide("[user] is stabbing [user.p_them()]self in the throat with [src]! It looks like [user.p_theyre()] trying to commit suicide!"))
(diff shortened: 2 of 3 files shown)


jachlompsky:
Also, the bee kit is still only around 20 tc if you consider the stinger 6-8 tc or so (the comments in the code claim beenades are 15 tc, this is not true.) I could add a dart pistol or something to the kit, since it's sorta thematic, if this still isnt enough to make it decent. My stinger change is kinda wacky tbh

namjay1170:
I feel this is relevant and that this PR could be improved by the addition of yet more bees.

https://youtu.be/-kWW9luMD4M

UltraBigRat:
honestly this kit either needs removal or far more, even with the weapon being a bit better it has significantly less utility then any other kit, having no real range or defense and a couple pretty meh grenades.

Uristthedorf:
My thoughts,
Would sulfuric acid+tirizene+cryptobiolin fit the theme of the bee kit? Maybe venom would be better or just increasing the amount of histamine.
Beesease could get a harder cure like honey, that would require some effort to get. You could make the bees themselves spread the disease.
You could make the bee costume they get better by armoring it or giving it some fancy property. (Or give them armor polish or adamantine dust.)
Could give a DNA injector for some mutation, or autosurgeon for some organ/implant. Can't think of anything that would fit the theme though. (Maybe wings?)
Maybe could give them a weak spell/ability? Would have to code one.

Burger1998:
> My thoughts, Would sulfuric acid+tirizene+cryptobiolin fit the theme of the bee kit? Maybe venom would be better or just increasing the amount of histamine. Beesease could get a harder cure like honey, that would require some effort to get. You could make the bees themselves spread the disease. You could make the bee costume they get better by armoring it or giving it some fancy property. (Or give them armor polish or adamantine dust.) Could give a DNA injector for some mutation, or autosurgeon for some organ/implant. Can't think of anything that would fit the theme though. (Maybe wings?) Maybe could give them a weak spell/ability? Would have to code one.

adamantine dust is only 1 bullet, laser and energy and 2 melee your better off just stealing a discarded explorer suit from mining

UltraBigRat:
An idea ive seen floating around before is giving the stinger a ranged attack it can do, which could help with bees lack of  range, which combined with the chems might make it somewhat more able to deal with a secoff

MajManatee:
Id love to see honey as the cure for the bee sease, and MORE bees. like a 10-15% chance for bee if not higher, maybe releasing MORE bees instead of just one too.

jachlompsky:
I'll open another PR with this feedback. Looks like this one was not ambitious enough :)

PR #18217, opened 2023-03-08
Beekit buff 2: electric beegaloo

jachlompsky:

Improves the bee kit by making some of its gear better and buffing beesease.

THE STINGER:
New version of the stinger added, as a subtype of the spellblade. Same stats as the old stinger, but it now injects 2u of venom instead of 4u histamine, and fires projectiles that also inject 2u venom. Projectiles recharge quickly and only do 1 brute damage on hit, like the bee simplemob, because venom itself is a pretty nasty poison. Has a "buzzes" message when empty instead of "whizzes." Also fixed a bug having to do with the stinger's reagents not injecting how they were supposed to.

THE BEE COSTUME:
New "authentic" bee costume added and used in the syndikit. 20 armor for basic damage types and gives a -0.35 speedmod. Protects from bee stings like the regular costume, and you can put your stinger in the suit storage slot instead of having to use your belt slot (now you can wear your yellow fannypack!)

BEESEASE:
Beesease has had its bee-spawning increased from a probability of 1 to 3, and you now cough up 3 bees at a time, so now the "You cough up a swarm of bees!" text is more accurate and you have more friends!

All tested and should work unless I forgot to do something when making this pr

# Wiki Documentation

change "bee costume" to "authentic bee costume" in syndikit item list

# Changelog

:cl:  
rscadd: Added an improved version of the Stinger that fires projectiles
rscadd: Added an improved version of the bee costume
bugfix: fixed an issue with the Stinger's poison injection
tweak: Bee syndikit contains improved stinger & bee costume
tweak: Beesease makes more bees
/:cl:

### DIFF
--- a/code/datums/diseases/beesease.dm
+++ b/code/datums/diseases/beesease.dm
@@ -27,13 +27,15 @@
 					affected_mob.adjustToxLoss(2)
 					affected_mob.updatehealth()
 		if(4)
-			if(prob(10))
+			if(prob(7))
 				affected_mob.visible_message(span_danger("[affected_mob] buzzes."), \
-												span_userdanger("Your stomach buzzes violently!"))
+												span_danger("Your stomach buzzes violently!"))
 			if(prob(5))
 				to_chat(affected_mob, span_danger("You feel something moving in your throat."))
-			if(prob(1))
+			if(prob(3))
 				affected_mob.visible_message(span_danger("[affected_mob] coughs up a swarm of bees!"), \
 													span_userdanger("You cough up a swarm of bees!"))
+				new /mob/living/simple_animal/hostile/poison/bees(affected_mob.loc)
+				new /mob/living/simple_animal/hostile/poison/bees(affected_mob.loc)
 				new /mob/living/simple_animal/hostile/poison/bees(affected_mob.loc)
 	return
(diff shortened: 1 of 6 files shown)


MajManatee:
Still only 3%? I guess thats like 3% per tick or something so its probably good.

jachlompsky:
> Still only 3%? I guess thats like 3% per tick or something so its probably good.

Figured it's better to have more bees per cough than more coughs so people arent fighting off an endless stream of individual bees. If you want lots of bees you can always infect a buncha monkeys.

Uristthedorf:
Would it be better to have the name of the antag bee costume be the same as the normal beesuit? So that someone who examines you when you have a bee suit doesn't immediately know you are a valid.

jachlompsky:
> Would it be better to have the name of the antag bee costume be the same as the normal beesuit? So that someone who examines you when you have a bee suit doesn't immediately know you are a valid.

Anyone examining you would also see you're carrying the stinger, but I get that maybe you wouldn't take it out of the box yet or something so yeah I'll make it have the normal bee suit name

ChuChu42:
BIG ISSUE

because of how lights work, the mime mech emits a green light which makes it easy to see as long as your mech light is off

and if your light is on, its easy to see, its basically easy to see both ways

so this ruins the entire point of the mech
i request fixing before merging

jachlompsky:
> BIG ISSUE

wrong PR. merge my PR!

MajManatee (approved):
mm bees

PR #18220, opened 2023-03-08
Removes conversion based antags from midround events + Other changes.

BurgerLUA:

Darkspawn is back in starting round rotation, with a weight of 3.
Changed the weight of Shadowling from 6 to 3 as it is very similar to Darkspawn.
Conversion based antags, such as Darkspawn, Shadowling, Clockwork Cult, and Blood Cult can no longer be rolled midround.
Wizard can no longer be rolled midround.

# Justification

Big and chaotic gamemode types, like the conversion gamemodes and Wizard, should absolutely not roll midround because of the massive effect it has on the round.  Not only is it unfair to security to deal with one mainline antagonist type, but it is also unfair to the original antagonist types for dynamic to effectively go "hmmm no you're actually cult now, focus on that instead of what you were before :^)" if they manage to get converted.

On the topic of dynamic, dynamic is not a good system in its current state and it's absolutely not good enough to make decisions about if a round should turn into a hybrid clusterfuck featuring changelings and clockwork cultists.

# Changelog

:cl:  BurgerBB
rscdel: Darkspawn, Shadowling, Clockwork Cult, Blood Cult, and Wizard can no longer be rolled midround.
rscadd: Darkspawn can roll at the start again, with a weight of 3.
tweak: Changed the weight of Shadowling from 6 to 3 as it is very similar to Darkspawn.
/:cl:


isotope436:
comedy i ded pr but honestly has some marrit i dont think all of the conversion tags should be removed like revs but definietly any with powers like dspawn

BurgerLUA:
> comedy i ded pr but honestly has some marrit i dont think all of the conversion tags should be removed like revs but definietly any with powers like dspawn

I did not die to darkspawn. ~~Only converted.~~

tattax:
i haven't checked if this removes threat wizard or bad wizard but both should stay as they are anyhow. 
1) if it has enough threat why not let it roll
2) bad wizard has pre loadouts anyways

BurgerLUA:
> i haven't checked if this removes threat wizard or bad wizard but both should stay as they are anyhow.
> 
> 1. if it has enough threat why not let it roll
> 2. bad wizard has pre loadouts anyways

The logic here is that you're allowed to powergame if it's wizard.

But if it's wizard, and there is another antag type, wouldn't powergaming in the name of wizard also affect the other antagonist type?

tattax:
> The logic here is that you're allowed to powergame if it's wizard.
> 
> But if it's wizard, and there is another antag type, wouldn't powergaming in the name of wizard also affect the other antagonist type?

admin issue + if it's dynamic black orbit there's nothing you could do

namjay1170:
Are zombies a conversion antag?

Darkspawn being midround I feel is validly explained in the PR that made it a midround #17833

BurgerLUA:
> Are zombies a conversion antag?
> 
> Darkspawn being midround I feel is validly explained in the PR that made it a midround #17833

"Rounds are decided in minutes to either side." is generally an argument to rework or remove the antagonist type and not straight up add them to mid-round events.

RG4ORDR:
> Wizard can no longer be rolled midround.

There's no justification for this.
Midround wizard either is Shitty Wizard or very rarely an actual full blown Wizard. This doesn't occur often enough to necessitate this.

namjay1170:
> > Are zombies a conversion antag?
> > Darkspawn being midround I feel is validly explained in the PR that made it a midround #17833
> 
> "Rounds are decided in minutes to either side." is generally an argument to rework or remove the antagonist type and not straight up add them to mid-round events.

The balance council also OK'd it aswell as the straight up disabling of nightmare when dspawn became a midround as it replaced it. Darkspawn is ass please dont make it a round type

MajManatee (changes requested):
Atomize these changes.

MajManatee:
From a non code point of view
Darkspawn is dogshit, was disabled for a reason.
Wizard midround is already incredibly rare, if it happens suck it up and deal with it, as antag or crew.
I can understand clock cult midround not occurring, just because they have timers and such that forces the round to end after a certain time.
Bloodcult however, I see no issue with remaining as a midround.
Slings also suck, so I'm happy to see them less common, however atomization.
I would argue that zombies ARE a conversion antag.

Overall, I don't see much rationale behind these changes besides "I don't like chaos as sec"

BurgerLUA:
So far I've seen "Darkspawn/Shadowling is a bad gamemode" multiple times yet its somehow good enough to have it as a midround role that very likely interrupts other antagonists types, which is worse than it being on its own because at least antagonists aren't affected by it and the crew isn't super weakened by already existing antagonists.

I've seen "Wizard is rare enough so just deal with it." just now which is hilarious because it has the starting weight of both changeling and bloodsucker combined. It also has the midround weight of 1, which is only 1 less than blob.

I'll atomize the changes, starting with Darkspawn.

namjay1170:
Here are 3 legit ways to become a real wizard midround and they're generally less rare than bad wizard existing in the first place and getting spells that arent ass. 2/3 can happen within 10 minutes of round start.

Below is the probability of winning the slot machine/finding it in maint it and getting a nat 20 for wizard (not factoring the rng needed for these to be spawned in via dynamic weight or lavaland budget+demon ship being the right one or even the demon of sin being greed and not a different one)

Greed demon of sin 0.5%  (10% win chance/per play)
Lavaland greed ship 0.25%  (5% win chance/per play)
Maint loot seems to be 0.006% if im interpreting the spreadsheet and recent changes correctly but doubt its very accurate

Bad wiz has 9 loadouts 5 of which have spells that arent practically useless and only 1 having an offensive spell with nothing else so 11% of a loadout that can actually threaten the round

I have watched too many rounds where the disk was never touched and no loneop spawned this gives me an inference that bad wiz which has a much smaller weight of 5 compared to the ever rising disk is incredibly rare though I lack statistics as dynamic is kinda a black box.

Bad wizard has a max occurrence of 1 same as the lavaland greed ship but demon of sin could make the entire station wizards with enough rng so greed demon of sin is a conversion antagonist here.

PR #18230, opened 2023-03-09
Gives jelly people a「LORE ACCURATE™」reduction to pay

Moltijoe:
I completely forgot this when adding them as a species
as we know, only humans can be paid 1.0

:cl:  
tweak: Jelly people are paid 40% less like the subhumans they are
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/jellypeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/jellypeople.dm
@@ -17,6 +17,7 @@
 	coldmod = 6   // = 3x cold damage
 	heatmod = 0.5 // = 1/4x heat damage
 	burnmod = 1 // = regular burn damage unlike other slimes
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/species_types/jellypeople.dm
```diff
@@ -17,6 +17,7 @@
 	coldmod = 6   // = 3x cold damage
 	heatmod = 0.5 // = 1/4x heat damage
 	burnmod = 1 // = regular burn damage unlike other slimes
+	payday_modifier = 0.6 //literally a pile of toxic ooze walking around, definitely a health hazard
```
SomeguyManperson, line 20:
```suggestion
	payday_modifier = 0.05 //literally a pile of toxic ooze walking around, definitely a health hazard
```

Moltijoe: this isn't THAT kind of meme PR

MajManatee (approved):
Society.

PR #18231, opened 2023-03-09
Buffs stechkin + supressor price

solwashere:

The stechkin being the same price as the 357 makes no sense, especially considering how weak it is in comparison. This changes the price of the stechkin from 6 tc to ~~4~~ 5 tc, and the price of the suppressor from 3 tc to ~~2~~ 1 tc so you can have a weaker but stealthier weapon for the same price

# Spriting

# Wiki Documentation

# Changelog

:cl:  

tweak: Stechkin costs 5 now
tweak: Suppressor costs 1 tc now 

/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -555,7 +555,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	desc = "A small, easily concealable handgun that uses 10mm auto rounds in 8-round magazines and is compatible \
 			with suppressors."
 	item = /obj/item/gun/ballistic/automatic/pistol
-	cost = 6
+	cost = 4
 	exclude_modes = list(/datum/game_mode/nuclear/clown_ops)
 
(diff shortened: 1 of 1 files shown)


DeadlyBag:
Hey!
We have been talking (Balance council) about the prices of the Stechkin & suppressor.
Since theres another Stechkin buff going through soon having the Stechkin be as expensive as a dart pistol doesnt make much sense as its roughly on the same terms as the revolver. 
Suppressors never get used since theres drawbacks to using them (size) and they are also quite expensive for what you are getting.
We think that the Steckin being lowered to 5 TC and the suppressor to 1 TC would be the best thing going forward.

solwashere:
Aye I'll agree, I'll go ahead and change that in a few hours

solwashere:
updated

Moltijoe:
Just replace the changelog rather than making it a strikethrough

ynot01 (dismissed):
Just replace the changelog rather than making it a strikethrough

MajManatee:
i like the strikethrough

solwashere:
1984, strikethroughs are based

ToasterBiome (approved):

PR #18232, opened 2023-03-09
Stechkin Ammo Buff + Revolver ammo tweaks

Skrem7:

I should spend 6 TC on a better gun instead of 6 TC on the worse gun

Makes stechkin more relatively unique to the revolver by making it pew-pew its smoler bullets more (ten round magazines)

Revolver alternative ammos have generally been made slightly more appealing so they still do SOMETHING even if you don't use them in their very tight niches

Ironfeather box should actually give ironfeather now because I am very bad at coding. Also ironfeather has a little AP now because it suffered from the negative AP of pellets and I didn't like this for how little damage it does more than the base bullet PLUS the base bullet has AP

Technically tinkers with some wound_chances but they're very minor and not really worth noting

# Wiki Documentation

Guide to combat and syndicate items and all that will need the new magazine values and also damage values

# Changelog

:cl:  
bugfix: .357 Ironfeather box in the uplink should now actually give Ironfeather
tweak: Stechkin now has ten rounds per magazine
tweak: .357 Nutcracker and .357 Metalshock have had their damage values increased
tweak: .357 Ironfeather now has 10 AP, up from -20 of default pellets
/:cl:


UltraBigRat:
Stetchkin and revolver just dont really feel like they fit on the same power level, cost adjustments like in the other pr would probably be better, giving mroe room for the stetchkin to be used and give it more room to combo with other items rather then trying to balance them with buffs and nerfs. Would also probably be better to up the revolvers cost by one or two tc rather then making it another ballistics weapon that armor gobbles up.

Moja10:
I think the revolver is perfectly balanced, it tells sec to fuck off by its mere presence, unlike the myriad of other traitor weapons that they just kind of ignore you using and just run at you to batong you. Revolver says "respect my space" during combat. I supported your other PR to make the sketchkin the cheap option and the revolver the expensive option more than making the revolver worse, I say leave its AP as is.

MajManatee (approved):
looks fine

cuackles (approved):

PR #17905, opened 2023-02-12
Changes Antag Coefficient to increase traitor, changeling and blood brother population relative to total station population (Increases total number of atnags

stwordcitizen:
# Document the changes in your pull request

This pull request changes the antag cooeficient for the antags listed above in an attempt to increase total number of antags in a round, this is because I believe that there are too few antags for the population on the station. This has been echod many times in the Discord server.
image
image
Many more times in the discord, my search function is broken.
# Wiki Documentation

Maybe this information was on the wiki, I could not find it.
# Changelog

:cl:  
tweak: Changes antag coefficient
/:cl:

PR #18234, opened 2023-03-09
Revert "Changes Antag Coefficient to increase traitor, changeling and blood brother population relative to total station population (Increases total number of atnags"

SomeguyManperson:
Reverts yogstation13/Yogstation#17905
It sucked

:cl:
tweak: blood brothers and changelings scale slower
/:cl:

### DIFF
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -247,9 +247,9 @@ SHUTTLE_REFUEL_DELAY 12000
 ## Variables calculate how number of antagonists will scale to population.
 ## Used as (Antagonists = Population / Coeff)
 ## Set to 0 to disable scaling and use default numbers instead.
-TRAITOR_SCALING_COEFF 5
-BROTHER_SCALING_COEFF 8
-CHANGELING_SCALING_COEFF 7
+TRAITOR_SCALING_COEFF 7
+BROTHER_SCALING_COEFF 10
+CHANGELING_SCALING_COEFF 8
 
 ## Variables calculate how number of open security officer positions will scale to population.
 ## Used as (Officers = Population / Coeff)
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -247,7 +247,7 @@ SHUTTLE_REFUEL_DELAY 12000
 ## Variables calculate how number of antagonists will scale to population.
 ## Used as (Antagonists = Population / Coeff)
 ## Set to 0 to disable scaling and use default numbers instead.
-TRAITOR_SCALING_COEFF 7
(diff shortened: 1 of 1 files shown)


Moja10:
Its been great, you are smoking crack, don't full revert if you think its too high. Just tweak the numbers a little. Previous was way too low and resulted in greenshifts if sec was successful whatsoever at their job.

MegaEmpirical:
More antags bad

solwashere:
If I have to suffer through one more round with 10 antagonists on like 30-40 pop I'm going to lose it

MajManatee:
if I have to suffer through one more greenshift on 60 pop I'm going to go postal

PibeCantina:
It doesnt need to be reverted, just need to lower the rates a bit, before it was too little, now it's too much, we need to find the sweetspot.

Therandomhoboo:
The only people that dislike the revert are the ones that antag-roll go figure

stwordcitizen:
> If I have to suffer through one more round with 10 antagonists on like 30-40 pop I'm going to lose it

I'd be ok with raising the traitor coefficient by like one or just keeping it unchanged because the chaos is great imo

Burger1998:
> The only people that dislike the revert are the ones that antag-roll go figure

sir aren't you banned?

MegaEmpirical:
I'd rather see events increase rather than antags. Maybe increase as antags go down or something. Plus new events would be nice.

namjay1170:
> I'd rather see events increase rather than antags. Maybe increase as antags go down or something. Plus new events would be nice.

I love greytide and meteors lets get more of that, lowpop didn't have nearly enough issues to deal with.

Moja10:
Antag spawn rates have been so low and lower than what it was historically for a bug that ynot fixed that was persistent for years, and now people who have only ever known bugged antag spawn rates are saying there are too many antags. 

This is actually close to what the spawning rate was historically, if only a tad bit more. Arguably, Its spawning things like BB in too high numbers, but that can be tweaked and its honestly spawning things like traitors, heretics in fine rates. 

Absolutely no more 1 heretic for 50 pop, that's a 20x bigger issue than a little chaos and spicier rounds. Spicier rounds are only an issue if they get literally too spicy for crew/sec to handle, but that's not really been the case.

UltraBigRat:
Better to have a round that is too chaotic then one where theres two antags that get caught immediatly or do absolutely nothing of note as the crew  dies of boredom.

Hislyna:
It should scale with sec, not pop.

PR #18236, opened 2023-03-10
Immovable rod vs the unstoppable... wait hold on...

MajManatee:
Renames the immovable rod to the unstoppable rod. Because it cannot be stopped. that's the whole point. Its not immobile its very clearly moving, unless I'm missing the deeper ROD LORE

# Changelog

:cl:  
experimental: The Rod is no longer immovable. Its unstoppable.
/:cl:


neeshacark:
the rod is not moving, the station is

ynot01:
the rod is not moving, the station is

isotope436:
man doesnt know about the rod lore smh

ToasterBiome:
the rod is not moving, the station is

Mantlecrawler:
the rod is not moving, the station is

MajManatee:
erm if the rod isnt moving then how do two rods pass eachtother?

LoliconSlayer:
the rod is not moving, the station is

Moltijoe:
the rod is not moving, the station is
\+ meme PR

MajManatee:
not a meme pr but ok

Moltijoe:
very well

RG4ORDR:
The Rod knows where it is at all times. It knows this because it knows where it isn't. By subtracting where it is from where it isn't, or where it isn't from where it is (whichever is greater), it obtains a difference, or deviation. The unmoving system uses deviations to generate corrective commands to not move the Rod from a position where it is to a position where it isn't, and arriving at a position where it wasn't, it now is. Consequently, the position where it is, is now the position that it wasn't, and it follows that the position that it was, is now the position that it isn't.
In the event that the position that it is in is not the position that it wasn't, the system has acquired a variation, the variation being the difference between where the Rod is, and where it wasn't. If variation is considered to be a significant factor, it too may be corrected by not moving. However, the Rod must also know where it was.
The Rod Not Moving computer scenario works as follows. Because a variation has modified some of the information the Rod has obtained, it is not sure just where it is. However, it is sure where it isn't, within reason, and it knows where it was. It now subtracts where it should be from where it wasn't, or vice-versa, and by differentiating this from the algebraic sum of where it shouldn't be, and where it was, it is able to obtain the deviation and its variation, which is called error.

Cartlord:
Since you seem committed to this, let me try to explain: the "Immovable Rod" we have is actually a reference to a 2008 4chan post about Dungeons & Dragons - see, in D. & D., one of the many items you can acquire is an "Immovable Rod", a magical item which can be activated to hang it in place, even defying gravity; the post in question mentioned a Dungeon Master who treated them as "truly immovable", meaning that - since all stars, planets, and entities in the universe are in motion at all times - activating one would (from the user's perspective) result in it rocketing off at incredibly high speeds, tearing through anything in its path. The poster who shared this also mentioned that they later abused this property to destroy a castle.
The idea of the rod event is that, if these rods were a common item possessed by Space Wizards, they'd likely have scattered a lot through the universe - effectively biding their time until the classic
**CLANG-
"What the _fuck_ was that!?!"**
occurs when something bumps into them.
i'd attach the pics but github keeps throwing file processing errors

Skrem7:
the rod is not moving, the station is

Aquizit:
the station is moving, the rod is not

TheBombMakerJL:
the rod is not moving, the station is

namjay1170:
I happen to recall a few objects that exist/used to exist that would actually stop the rod, the nuke if hit directly on could do this.

Bibby0110:
the rod is not moving the station is

PR #18237, opened 2023-03-10
Holotool can repair mechs (for real this time)

Moltijoe:
Needed another check to be forced = TRUE to bypass fuel

:cl:  
bugfix: Holotool welder can repair mechs
/:cl:

### DIFF
--- a/yogstation/code/game/objects/items/holotool/holotool.dm
+++ b/yogstation/code/game/objects/items/holotool/holotool.dm
@@ -50,6 +50,9 @@
 			return TRUE
 	. = ..()
 
+/obj/item/holotool/use(used)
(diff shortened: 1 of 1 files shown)


MajManatee:
@Moltijoe next make it so the holo tool welder can cut through the reinforced plating on doors

Moltijoe:
> @Moltijoe next make it so the holo tool welder can cut through the reinforced plating on doors

unless the airlocks have snowflake code, this should also allow that, this just lets it bypass the fuel check, meaning anything that looks for TOOL_WELDER tool behaviour should work.

MajManatee (approved):
surely this wont go wrong

PR #18242, opened 2023-03-10
pickpocketing puts the item in your hand

jachlompsky:

taking an item from someones pocket puts it in your hand
no more sneakily taking someones stuff just for it to drop onto the floor in front of them

only affects pockets because stripping other stuff is obvious already, and putting something in your hand resets item-strip progress so doing it for everything would make fullstrips have to be done one item at a time

Edit: video example

https://user-images.githubusercontent.com/74697056/224428157-78d673ba-70d6-4f1a-a2f9-d575411a3a60.mp4

# Changelog

:cl:  
tweak: pickpocketing puts the item in your hand
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -300,6 +300,7 @@
 			if(pocket_item)
 				if(pocket_item == (pocket_id == SLOT_R_STORE ? r_store : l_store)) //item still in the pocket we search
 					dropItemToGround(pocket_item)
+					usr.put_in_hands(pocket_item)
 			else
 				if(place_item)
 					if(place_item.mob_can_equip(src, usr, pocket_id, FALSE, TRUE))


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -300,6 +300,7 @@
 			if(pocket_item)
 				if(pocket_item == (pocket_id == SLOT_R_STORE ? r_store : l_store)) //item still in the pocket we search
 					dropItemToGround(pocket_item)
+					usr.put_in_hands(pocket_item)
```
Moltijoe, line 303: should probably remove `dropItemToGround(pocket_item)` to make it truely stealthy, otherwise the drop animation will still play

jachlompsky: > should probably remove `dropItemToGround(pocket_item)` to make it truely stealthy, otherwise the drop animation will still play

I didn't see the item show when I tested this, and when I originally tried this without having dropItemToGround it created a copy of the item that acted like the target's pocket and my hand were now paradox bags holding that item.

Moltijoe: weird, fair enough, byond code is more jank than i thought i guess.

Moltijoe (approved):

TheBombMakerJL:
A sec main coder or something will have this happen to them and will make a PR preventing it so maybe make it so that there’s a small message that shows it’s happening if they have sufficient chest melee armor

jachlompsky:
> A sec main coder or something will have this happen to them and will make a PR preventing it so maybe make it so that there’s a small message that shows it’s happening if they have sufficient chest melee armor

I know this is a possibility, but I don't want to concede special protections to security right away just cause it _could_ happen. I don't see this as nearly so much of an issue as cult stuns being used in combat, for example. Worst that could happen is the secoff puts his only cuffs in his pocket instead of belt (skill issue), doesn't notice them disappear (skill issue), and then has to arrest someone so far away from the brig or any source of cable coils that all his many methods of stunning become exhausted first. Non-antagonists aren't allowed to antagonize security/steal security items anyways so this shouldn't be something that happens constantly. None of this might matter in the face of a sec main coder, but I'm going to hope for the best.

Moltijoe:
> A sec main coder or something will have this happen to them and will make a PR preventing it so maybe make it so that there’s a small message that shows it’s happening if they have sufficient chest melee armor

isn't there already text that pops up when someone tries to strip you?

TheBombMakerJL:
> > A sec main coder or something will have this happen to them and will make a PR preventing it so maybe make it so that there’s a small message that shows it’s happening if they have sufficient chest melee armor
> 
> isn't there already text that pops up when someone tries to strip you?

Only if you move while it’s happening

PR #18244, opened 2023-03-11
Give mining medics something to do

Moltijoe:
Every miner seems to get the implants now, because points aren't _that_ rare, and an autosurgeon removes any hassle from getting the implants over the boots
The implants will still be enticing because keeping the storage space or using the drip shoes will be pretty nice.
This just adds a small step that encourages miners to actually rely on their medics for something (or just go to medbay i guess)

:cl:  
tweak: jumpboot implants cost half as much from the mining vendor
tweak: jumpboot implants from the mining vendor no longer come with a free autosurgeon
/:cl:

### DIFF
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -62,7 +62,7 @@
 		new /datum/data/mining_equipment("GAR Meson Scanners",			/obj/item/clothing/glasses/meson/gar,								500, VENDING_EQUIPMENT),
 		new /datum/data/mining_equipment("Meson Health Scanner HUD",	/obj/item/clothing/glasses/hud/health/meson,						1000, VENDING_EQUIPMENT),
 		new /datum/data/mining_equipment("Jump Boots",					/obj/item/clothing/shoes/bhop,										2500, VENDING_EQUIPMENT),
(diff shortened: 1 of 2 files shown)


neeshacark:
I still think we should've merged the PR that added a small botany area to mining for this very reason.

Moltijoe:
> I still think we should've merged the PR that added a small botany area to mining for this very reason.

giving mining medic a rudimentary "chemistry" setup, that mainly tailors to lavaland plants and stuff like grub guts might be an intersting idea

MajManatee:
I feel as though this small change will make the usage of them drop to pretty much zero. People that can buy the shoes can, and still will, but people that cannot wear shoes simply wont bother. there either isnt a mining medic, there is a mining medic but they suck and dont know how to do it, or they are afk. Nobody wants to waste 15 minutes to go find a medic in the real medbay.

Moltijoe:
> I feel as though this small change will make the usage of them drop to pretty much zero. People that can buy the shoes can, and still will, but people that cannot wear shoes simply wont bother. there either isnt a mining medic, there is a mining medic but they suck and dont know how to do it, or they are afk. Nobody wants to waste 15 minutes to go find a medic in the real medbay.

better than it seemingly being 100% using the implants
they already need to get people to implant the arm if they want it, needing to get these implanted should be no different
miners don't even really do mining after a certain point, so it's not like they're limited on time

ToasterBiome (approved):

PR #18245, opened 2023-03-11
Moves wheelie and jumpboot implants behind combat cybernetics

Moltijoe:
These implants are too easy to get for how impactful they can be, especially when things are going south in a round.
Them being on-par research wise with anti-drop and CNS-rebooter seems reasonable for their power.

Also renames the ids to fit with the standard naming convention

:cl:  
tweak: wheelie and jumpboot implants are moved behind combat cybernetics
/:cl:


ReddicusDragon (approved):
I see no issues

PR #18248, opened 2023-03-11
updates Ptech vending machine to new pdas

Moltijoe:
wouldn't you know it, old pdas were removed a while ago

:cl:  
tweak: updates Ptech vending machine to new pdas
/:cl:

### DIFF
--- a/code/modules/vending/cartridge.dm
+++ b/code/modules/vending/cartridge.dm
@@ -5,14 +5,7 @@
 	product_slogans = "Carts to go!"
 	icon_state = "cart"
 	icon_deny = "cart-deny"
-	products = list(/obj/item/cartridge/medical = 10,
-					/obj/item/cartridge/engineering = 10,
-					/obj/item/cartridge/security = 10,
-					/obj/item/cartridge/janitor = 10,
-					/obj/item/cartridge/signal/toxins = 10,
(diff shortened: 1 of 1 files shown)


00ze-cyclone:
There's already the modular computer vendor to dispense the new PDA, a vendor with only one item also feel a little useless. If we can't find anything to put in it maybe it should just be removed.

Moltijoe:
> There's already the modular computer vendor to dispense the new PDA, a vendor with only one item also feel a little useless. If we can't find anything to put in it maybe it should just be removed.

Its for the HoP to give out new pdas to those that lost them

00ze-cyclone:
In that case maybe a box of PDAs in their locker would be better

MajManatee:
the old PDAs have the bluespace paper printer
no other PDA's have a printer installed by default, making it really annoying to dispense paperwork.

ReddicusDragon:
> the old PDAs have the bluespace paper printer
> 
> no other PDA's have a printer installed by default, making it really annoying to dispense paperwork.
> 
> 

Actually, there's a paperwork program for new pdas and head PDAs come with a printer installed. I made this PR a bit ago, I think you just need to add paper.

ToasterBiome (approved):
goodbye, might be useful for more stuff to be added to this but, such is life

PR #18251, opened 2023-03-12
[s] Soil has no electricity

ThatLing:

Soil has no electricity


RG4ORDR:
The Shamans summon the power of the necropolis and geothermal vents.

PR #18265, opened 2023-03-15
Bloodsucker punches now care about armor a little more

Mqiib:
Brawn punch damage is now reduced by armor so you cant get free high-level bone wounds on people wearing riot armor.
Tzimisce dice ability now only dismembers if the targeted bodypart has less than 40 melee armor. Otherwise it just does a bunch of damage. I didn't make the damage reduced by armor because its a special ability for a single midround-only clan that is kinda themed to be pretty slicey and precise.

# Wiki Documentation

Dice can't dismember living humans with significant armor.

# Changelog

:cl:  

tweak: Bloodsucker brawn ability now respects armor (with 20 AP)
tweak: Tzimisce bloodsucker dice ability can no longer instantly dismember targets with significant armor
/:cl:

### DIFF
--- a/code/modules/antagonists/bloodsuckers/powers/targeted/brawn.dm
+++ b/code/modules/antagonists/bloodsuckers/powers/targeted/brawn.dm
@@ -136,7 +136,8 @@
 		playsound(get_turf(target), 'sound/weapons/punch4.ogg', 60, TRUE, -1)
 		user.do_attack_animation(target, ATTACK_EFFECT_SMASH)
 		var/obj/item/bodypart/affecting = target.get_bodypart(ran_zone(target.zone_selected))
-		target.apply_damage(hitStrength, BRUTE, affecting)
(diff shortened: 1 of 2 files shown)


Marmio64:
ided, but reasonable. Though i must add that rank 3 to 6 dismember will still instantly disable limbs regardless of armor, while rank 6+ will only deal 30 damage without disabling, meaning the middle rank dismember is technically better than the high rank up against specifically highly armored individuals. Perhaps have the high rank dismember deal 30 brute AND 20 stam, or even 25 stam and 25 brute like the middle rank

Mqiib:
> ided, but reasonable. Though i must add that rank 3 to 6 dismember will still instantly disable limbs regardless of armor, while rank 6+ will only deal 30 damage without disabling, meaning the middle rank dismember is technically better than the high rank up against specifically highly armored individuals. Perhaps have the high rank dismember deal 30 brute AND 20 stam, or even 25 stam and 25 brute like the middle rank

Made the middle rank do 30/20 stamina/brute and the failed max rank 20/30 stamina/brute.

ToasterBiome (approved):

PR #18267, opened 2023-03-15
Removes black market wing dealers

Mqiib:
Anything that wasn't a valid species could get wings by just drinking less than 5 units of wing juice because the logic was fucked. This is ILLEGAL and now it doesn't work hahaha I hate fun.

# Changelog

:cl:  Mqiib, @Chubbygummibear 

bugfix: Species that can't get wings now actually can't get wings

/:cl:

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -689,15 +689,18 @@ GLOBAL_LIST_EMPTY(aide_list)
 /datum/reagent/flightpotion/reaction_mob(mob/living/M, method=TOUCH, reac_volume, show_message = 1)
 	if(iscarbon(M) && M.stat != DEAD)
 		var/mob/living/carbon/C = M
-		var/holycheck = ishumanbasic(C)
-		if(!(holycheck || islizard(C) || ismoth(C) || isskeleton(C) || ispreternis(C) || isipc(C) || (reac_volume < 5))) //humans (which are holy?), lizards, skeletons, and preterni(ises?) can get wings
+		var/valid_species = (ishumanbasic(C) || islizard(C) || ismoth(C) || isskeleton(C) || ispreternis(C) || isipc(C))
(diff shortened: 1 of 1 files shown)


namjay1170:
NOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO

UltraBigRat:
Just let all species get wings

MajManatee:
so much for the 0.01u pills

code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -689,15 +689,18 @@ GLOBAL_LIST_EMPTY(aide_list)
 /datum/reagent/flightpotion/reaction_mob(mob/living/M, method=TOUCH, reac_volume, show_message = 1)
 	if(iscarbon(M) && M.stat != DEAD)
 		var/mob/living/carbon/C = M
-		var/holycheck = ishumanbasic(C)
-		if(!(holycheck || islizard(C) || ismoth(C) || isskeleton(C) || ispreternis(C) || isipc(C) || (reac_volume < 5))) //humans (which are holy?), lizards, skeletons, and preterni(ises?) can get wings
+		var/valid_species = (ishumanbasic(C) || islizard(C) || ismoth(C) || isskeleton(C) || ispreternis(C) || isipc(C))
```
ToasterBiome, line 692: ew this check sucks

ToasterBiome (approved):

code/modules/mining/lavaland/necropolis_chests.dm
Mqiib: It's basically the original check but moved out of the if statement. I could rework it into more than one line of code and make it prettier but why would I bother when any performance gains would be pretty much nonexistant from how often this gets called.

namjay1170:
what if you just didnt do this and instead we all called it a feature

isotope436:
no this is bad

PR #18295, opened 2023-03-18
Singularity hotfix

Kaithyl:
Hotfix for 754d553 breaking player lighting effects. Accidentally included a line that was clearing all filters from the lighting layer so I removed it. Also noticed in testing that objects that inherit from singularity (ie. narsie) also inherit the effects so I created a new obj class singularity/gravitational to differentiate.

# Changelog

:cl:  
bugfix: Fixes lighting, other objs that inherit from singularity
tweak: Singulo generator and sword's creation type changed to new singulo class
/:cl:


cuackles:
Question, does the distortion range match the gravitational pull range?

Kaithyl:
> Question, does the distortion range match the gravitational pull range?

It does for every stage except 4 and 5, where the pull radius remains the same but the effect increases in size. Otherwise its correct for stage 1, which has a pull radius of 4 tiles.

cuackles (approved):
Good stuff

PR #18298, opened 2023-03-18
Should fix replacing your eyes not removing blindness from eye damage

Mqiib:
More relevant for robot eyes since they have a tendency to die on you from someone spitting a little iron/uranium your way. This should make replacing your eyes actually work for fixing blindness.

# Changelog

:cl:  

bugfix: Replacing your eyes should actually cure eye damage blindness now

/:cl:


MajManatee:
mib is such a slacker smh
re-open when work continutes

PR #18300, opened 2023-03-18
Several AltClick improvements

ThatLing:

- Preternis will no longer get the ``X is too far away`` error message when alt-clicking anything
  - Made the error message a bit clearer
  - Uses component signal now instead of overriding ``/AltClickOn``
- Windows/windoors/railing will only print error message when alt-click if you are adjacent to them
- Converted a few old yogs component signal names to newer ones
  - Removed daniel component since it's unused and old code
  - Cleaned up ``/mob/proc/AltClickOn`` a bit
- Moved changeling sting over to component signal instead of overriding ``/AltClickOn``
- Added ``IN_GIVEN_RANGE`` macro

### DIFF
--- a/code/modules/antagonists/changeling/changeling.dm
+++ b/code/modules/antagonists/changeling/changeling.dm
@@ -378,19 +378,24 @@
 		add_new_profile(C)
 
 /datum/antagonist/changeling/apply_innate_effects()
+	var/mob/living/living_mob = owner.current
+	if(!isliving(living_mob))
+		return
+
+	RegisterSignals(living_mob, list(COMSIG_MOB_MIDDLECLICKON, COMSIG_MOB_ALTCLICKON), .proc/on_click_sting)
+
 	//Brains optional.
-	var/mob/living/carbon/C = owner.current
-	if(istype(C))
-		var/obj/item/organ/brain/B = C.getorganslot(ORGAN_SLOT_BRAIN)
-		if(B)
-			B.organ_flags &= ~ORGAN_VITAL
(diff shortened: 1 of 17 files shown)


ynot01:
i love signals!!!

code/modules/antagonists/changeling/changeling.dm
```diff
@@ -519,6 +524,31 @@
...
+		return
+
+	if(!isliving(ling) || clicked == ling || ling.stat != CONSCIOUS)
+		return
+
+	// sort-of hack done here: we use in_given_range here because it's quicker.
+	// actual ling stings do pathfinding to determine whether the target's "in range".
+	// however, this is "close enough" preliminary checks to not block click
+	if(!isliving(clicked) || !IN_GIVEN_RANGE(ling, clicked, sting_range))
+		return
```
Hopekz, line 545:
```suggestion

 // Checks if the target is within the sting range of the changeling.
/proc/in_changeling_sting_range(mob/living/ling, mob/living/target, range)
	return (get_dist(ling, target) <= range)

// Using a more specific proc for the changeling sting range check.
// Actual ling stings do pathfinding to determine whether the target is "in range".
// This is a "close enough" preliminary check to not block click.
if(!isliving(clicked) || !in_changeling_sting_range(ling, clicked, sting_range))
	return
```

Might as well just do it in a non hack way

ThatLing: This is literally what it's already doing but as a proc instead

Moltijoe (approved):

PR #18309, opened 2023-03-18
Lets people crawl while nuggeted

Moltijoe:
Full nuggeting making it impossible to move can allow for some pretty boring events where a nuggeted antag just has to sit around waiting for sec, or a nuggeted secoff just has to lie there waiting for someone to find them. 

The default speed for having no limbs was a bit too fast, so i made missing arms matter more for crawling
this should only make a slight difference for those using a wheelchair when they lose an arm

:cl:  
tweak: Can crawl while nuggeted
/:cl:

PR #18310, opened 2023-03-18
EMP only removes limbs on heavy emp + augmented humanoids with ligament hook

Moltijoe:
only merge if #18309 does

it's pretty damning to lose a limb, light emps are super common too, limits it to only heavy emps.
also makes it remove limbs from those that got ligament hook and then augmented themselves
makes it not remove the head, both because becoming deaf, blind, and mute from a single emp was dumb, and because that would instakill an augmented person with ligament hook

also moves the IPC easydismember trait to inherent from species (like it should have been)(this will make it show up in preferences)
also uses the defines rather than the "text" that the defines stand for

:cl:  
tweak: EMP now also checks for ligament hook to remove limbs
tweak: EMP only removes limbs on heavy rather than any EMP strength
tweak: EMP no longer removes head 
/:cl:


milk-is-cool:
this wont affect skeleton easydismember limbs right?

Moltijoe:
> this wont affect skeleton easydismember limbs right?

not unless you get augmented

wonderinghost:
I think this still happens when you got prosphetics. Cause I remember losing both of them when I got emp'd by cultist

Moltijoe:
> I think this still happens when you got prosphetics. Cause I remember losing both of them when I got emp'd by cultist

It shouldnt, unless you were an IPC, in which case it doesnt matter between regular limbs and prosthetics

wonderinghost:
No it's a thing. Had that happen when my character would magically change to a changeling and I managed to emp my prosphetics off

iamthesilly:
> seemingly never happens

About 23% chance that you'll lose at least one

Moltijoe:
other emp PR preferred

PR #18313, opened 2023-03-19
Buffs plastitanium glass windows

Moltijoe:
Fun fact, plastitanium glass windows are weaker than plasma glass windows to bullets
Doubled it's integrity
halved it's melee armor so it's just as easy to break by melee

:cl:  
tweak: Buffs plastitanium glass windows
/:cl:


MajManatee (approved):
mmm window

PR #17706, opened 2023-01-30
Prevents mechs from shotgun swapping

ynot01:
# Document the changes in your pull request

No more ultrakill

# Changelog

:cl:  
tweak: Mecha weapon cooldowns are now shared
/:cl:

PR #18328, opened 2023-03-20
Revert "Prevents mechs from shotgun swapping"

Majkl-J:
Reverts yogstation13/Yogstation#17706

This made multiple mechs with long cooldown equipment (Namely HONK) completely unusable. Also it blocks melee from happening aswell. 

 # Changelog

🆑 
tweak: Mecha weapon cooldowns are no longer shared
/🆑

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -563,11 +563,6 @@
 		if(!target)
 			return
 
-	// No shotgun swapping
-	for(var/obj/item/mecha_parts/mecha_equipment/weapon/W in equipment)
-		if(!W.equip_ready)
-			return
-
 	var/mob/living/L = user
 	if(!Adjacent(target))
 		if(selected && selected.is_ranged())


Hopekz:
Why not just code an exception?
What I mean is add a new variable that bypasses the cool down then give the tools or weapons that you need that variable so it bypasses that cool down

Maia-J:
> Why not just code an exception? What I mean is add a new variable that bypasses the cool down then give the tools or weapons that you need that variable so it bypasses that cool down

I do have my own thing planned but this is the lazy way of doing it and there are 498 issues waiting for me

MajManatee (changes requested):
Rather than give mechs broken abilites again, just fix the issue where needed.

Maia-J:
> Rather than give mechs broken abilites again, just fix the issue where needed.

Oh boy the community is gonna hate me for what I am about to do 😏

PR #18330, opened 2023-03-20
The alert level no longer is raised to blue on shift start.

BurgerLUA:

The alert level is no longer raised to blue on shift start.

(Extended shifts still do the special message. )
(The "Antagonist report" is still generated.)
(Of course, war declaration still is raised to gamma.)

# Justification

This is a bit of a security nerf because it means security can't do random searches on green (they can still do justified searches) or openly carry armory guns.
It's an antagonist buff because it means that heads setting to blue means the crew might be on to them.

(This PR should probably be testmerged to see the actual effect on the round.)

# Changelog

:cl:  BurgerBB
rscdel: The alert level is no longer raised to blue on shift start.
/:cl:

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -310,7 +310,13 @@
 
 	intercepttext += generate_station_goal_report()
 
-	print_command_report(intercepttext, "Central Command Status Summary")
+	if(CONFIG_GET(flag/auto_blue_alert))
+		print_command_report(intercepttext, "Central Command Status Summary", announce=FALSE)
+		priority_announce("A summary has been copied and printed to all communications consoles.\n\n[generate_station_trait_announcement()]", "Enemy communication intercepted. Security level elevated.", ANNOUNCER_INTERCEPT)
+		if(GLOB.security_level < SEC_LEVEL_BLUE)
+			set_security_level(SEC_LEVEL_BLUE)
+	else
+		print_command_report(intercepttext, "Central Command Status Summary")
 		
 /*
  * Generate a list of station goals available to purchase to report to the crew.
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -310,10 +310,7 @@
 
 	intercepttext += generate_station_goal_report()
 
-	print_command_report(intercepttext, "Central Command Status Summary", announce=FALSE)
-	priority_announce("A summary has been copied and printed to all communications consoles.\n\n[generate_station_trait_announcement()]", "Enemy communication intercepted. Security level elevated.", ANNOUNCER_INTERCEPT)
-	if(GLOB.security_level < SEC_LEVEL_BLUE)
-		set_security_level(SEC_LEVEL_BLUE)
+	print_command_report(intercepttext, "Central Command Status Summary")
 		
 /*
  * Generate a list of station goals available to purchase to report to the crew.
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -54,6 +54,8 @@
 
 /datum/config_entry/flag/everyone_has_maint_access
 
+/datum/config_entry/flag/auto_blue_alert
+
 /datum/config_entry/flag/sec_start_brig	//makes sec start in brig instead of dept sec posts
 
 /datum/config_entry/flag/force_random_names
--- a/code/game/gamemodes/dynamic/dynamic.dm
+++ b/code/game/gamemodes/dynamic/dynamic.dm
@@ -302,11 +302,14 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
 	. += generate_station_goal_report()
 
 	desc += "\n\n[generate_station_trait_announcement()]"
-
-	print_command_report(., "Central Command Status Summary", announce=FALSE)
-	priority_announce(desc, title, ANNOUNCER_INTERCEPT)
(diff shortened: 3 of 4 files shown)


code/game/gamemodes/game_mode.dm
```diff
@@ -310,10 +310,7 @@
 
 	intercepttext += generate_station_goal_report()
 
-	print_command_report(intercepttext, "Central Command Status Summary", announce=FALSE)
-	priority_announce("A summary has been copied and printed to all communications consoles.\n\n[generate_station_trait_announcement()]", "Enemy communication intercepted. Security level elevated.", ANNOUNCER_INTERCEPT)
-	if(GLOB.security_level < SEC_LEVEL_BLUE)
-		set_security_level(SEC_LEVEL_BLUE)
+	print_command_report(intercepttext, "Central Command Status Summary")
```
tattax: summary should still have a message telling people about it. possibly just saying "Possible threats located in your region" or something

BurgerLUA: The classified command report already says that. The intercept text contains that information.

Bibby0110: you forgot about dynamic

Bibby0110 (changes requested):
this should be a config setting

BurgerLUA:
I'll make it a config setting this week.

code/game/gamemodes/game_mode.dm
tattax: ah

Moja10:
this change has been attempted like two other times but didn't make it in. At those times I was iffy, but I think it could be worth trying since so many roundstarts might as well be greenshifts with traitor giving delay, people taking time before doing antag actions, etc

burn1cycle:
Been tried before. I don't think you realize you want to change a complicated set of rules that have wide ranging effects and undesirable results. 

It's a hassle that people will forget to set to blue and admins will have to play "Aha! Got U!" while at the same time fending off rule lawyers who'll complain their traitor capture is unfair because green alerts 

It will be for very little gain, in most staffed/experienced command rounds command will flip to blue the moment something goes wrong because that's what blue is. On understaffed rounds where nobody can or knows to flip to blue it will just punish players for other players' inexperience.


 For the record I believe the balance issue comes from Yogstation insisting(rightfully) that we have responsible security players and if they aren't they don't last long and get job banned so we've ended up with a pool of very talented security players who've survived the bans to be robust unbalanced players. That's just my opinion maybe I'm wrong but I want people to think on that.

BurgerLUA:
Unless I'm mistaken, just because it is green alert doesn't mean that antags can't be caught or searched. Green alert just means that you need probable cause to search someone. Here is the text for the alert levels.

Green: "All threats to the station have passed. Security may not have weapons visible, privacy laws are once again fully enforced."

Blue: "The station has received reliable information about possible hostile activity on the station. Security staff may have weapons visible, random searches are permitted."

Red: "There is an immediate serious threat to the station. Security may have weapons unholstered at all times. Random searches are allowed and advised."

MegaEmpirical:
1) I am not a fan of this.

2) If this was to happen then you need to remove the round start report as well as that basically just makes it auto-blue by proxy.

Hislyna:
This is way too much of an antag buff imo. If people really want an "Antag Setup Safezone", change the alert to be 10-15 mins into the round instead.

MegaEmpirical:
> 1. I am not a fan of this.
> 2. If this was to happen then you need to remove the round start report as well as that basically just makes it auto-blue by proxy.

I've decided that I'd be cool with this but I would want to see two things happen:

1) we get species foot prints and maybe something with shoe sizes for detectives.
2) MORE folders for the detectives.

Buff the detective.

Hislyna:
> > 1. I am not a fan of this.
> > 2. If this was to happen then you need to remove the round start report as well as that basically just makes it auto-blue by proxy.
> 
> I've decided that I'd be cool with this but I would want to see two things happen:
> 
>     1. we get species foot prints and maybe something with shoe sizes for detectives.
> 
>     2. MORE folders for the detectives.
> 
> 
> Buff the detective.

As Lyis said, if we get a rework on how investigating is done, and not just an antag buff, this could work

BurgerLUA:
I kind of wanted to do a detective rework where detectives get a special "detective vision" ability to see a black-and-white visual of evidence in a crime scene. Problem is that it's a lot of work.

MegaEmpirical:
> I kind of wanted to do a detective rework where detectives get a special "detective vision" ability to see a black-and-white visual of evidence in a crime scene. Problem is that it's a lot of work.

Definitely work. I would personally favor just expanding ways to collect and hide evidence which would be simpler and I think more engaging.

When Hisa and I were talking about foot prints things that crossed our minds were like shape by species, size (which could work like blood types, randomish), and even a traitor tool to fake foot prints but only of your same size or larger.

Similar ideas could be applied elsewhere. I love playing detective and investigating.

ynot01:
Still not a config setting

BurgerLUA:
I am waiting for the council to vote before doing any changes. It will get a config setting if council approves it. No point in working on it if it's not approved.

Necrosium:
I'm up for seeing this in just to try, finally sec won't have an excuse to """random search""" every person they see

ToasterBiome (approved):

ToasterBiome (dismissed):
Make config setting + account for dynamic

BurgerLUA:
Time to work on this again.

Bibby0110 (approved):

BurgerLUA:
wow doing the work for me nice

PR #18341, opened 2023-03-21
EMP limb damage tweaks and bugfixes

SapphicOverload:

Distributes damage across all limbs so that EMPs will no longer be able to disintegrate your limbs or destroy jumpsuits.

Light EMPs will do an overall 10 burn and 30 stamina damage instead of 5 burn damage per limb and the maximum possible stamina damage it can sustain. Limbs will not be dismembered from light EMPs anymore.

Heavy EMPs will do an overall 20 burn and 60 stamina damage instead of 10 burn damage per limb and as much stamina as it can sustain. If either the head or chest are augmented, the stamina damage will be replaced by a 5-second paralyze.

Also, IPCs will now do the new *warn emote (blares an alarm!) when EMP'd, because I thought it would be funny.

# Changelog

:cl:  
bugfix: moves TRAIT_EASYDISMEMBER to the correct place in the IPC species file
bugfix: fixes EMPs disintegrating limbs and destroying jumpsuits
tweak: EMP damage to robotic limbs is now distributed evenly across all robotic limbs (up to 10 burn 30 stamina on light, up to 20 burn 60 stamina on heavy plus 5 second paralyze if fully augmented or IPC)
tweak: only heavy EMPs can dismember
tweak: IPCs do the *warn emote (*blares an alarm!) when EMP'd
/:cl:


Marmio64:
if i'm reading this right, this makes the cost of having a singular robotic limb worse than having multiple since you won't be able to spread out the damage. Doesn't make sense that having 3 or something robotic limbs would be better/the same if you're EMP'd compared to just one.

iamthesilly:
made it scale with how many parts you have

Moltijoe (approved):

PR #18353, opened 2023-03-22
[Gax] Real Evac Fix Since There Were 2 Problems

Aquizit:

Expands Evac Hall area to sec holding room for two reasons:
1) The door bordering the other area made the air alarm think that both were exterior doors. Which was the other half of the evad door problem.

2) More in line with all the other stations where this room is in the Evac area. The only reason it wasn't set originally this way is because it is directly adjacent to the Evac Sec Checkpoint

# Changelog

:cl:  
bugfix: Fixes the other of the evac door issues that wasn't diagnosed the first time since it causes the same issue.
/:cl:

### DIFF
--- a/_maps/map_files/GaxStation/GaxStation.dmm
+++ b/_maps/map_files/GaxStation/GaxStation.dmm
@@ -1176,6 +1176,15 @@
 	},
 /turf/open/floor/plasteel/dark,
 /area/science/robotics/lab)
+"aCU" = (
+/obj/structure/chair{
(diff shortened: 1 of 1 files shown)


MajManatee:
no pictures?

Aquizit:
> no pictures?

Didn't think of one since there are no structural changes. Just behind the scenes stuff like area redrawing and adding airlock helper arrows.

ToasterBiome (approved):

PR #18358, opened 2023-03-23
NT-USP no longer causes burn damage

SapphicOverload:

The NT-USP will no longer cause real damage. The burn damage gave silicons a valid justification to interfere with arrests and prevented it from being usable by pacifists, so I decided to make it completely non-lethal because it's already worse than the disabler in almost every way.

# Changelog

:cl:  
tweak: NT-USP is now fully non-lethal
/:cl:


MajManatee:
Let 'em burn

Moltijoe (approved):

PR #11984, opened 2021-08-09
Fixes the access on the spare ID cabinet and bridge fire axe cabinet

adamsong:
# Github documenting your Pull Request

The bridge fire axe cabinet, and spare ID cabinet will now have captain access, instead of atmos tech access.

# Changelog

:cl:  
bugfix: fixed the access on the spare ID cabinet
bugfix: fixed the access on the bridge fire axe cabinet
/:cl:

PR #18359, opened 2023-03-23
Give CE access to the bridge fireaxe cabinet

hedgehog1029:
# Changes

Prior art: https://github.com/yogstation13/Yogstation/pull/11984

Changes the fireaxe cabinet in the Bridge to give the Chief Engineer access. They already have access to the Atmospherics one, and the bridge itself, so seems a bit weird that they don't have access to this.

### wait but why

Obviously in most cases the CE can just ask the captain to unlock the cabinet for them.
But say for the sake of argument, there's nobody with captain access, and:

- Atmos is currently filled with space vines/fire/giant spiders/wizards, and therefore that fireaxe is not reachable
- The other fireaxe is in space in the hands of some poor atmos tech
- etc.

I don't think any other heads need access - HoS and RD likely have better options, CMO should probably be healing rather than hurting, HoP could give themselves all access anyway and probably will take over if there's no captain. The CE has legitimate reasons to hold a fireaxe (i.e. firefighting) wheras the other heads don't.

I'm told this will need a balance council review.

# Wiki Documentation

The access isn't currently documented on the wiki as far as I can see, but maybe it should be?

# Changelog

:cl:  
tweak: Gave the CE access to the bridge fireaxe cabinet
/:cl:

### DIFF
--- a/yogstation/code/game/objects/structures/fireaxe.dm
+++ b/yogstation/code/game/objects/structures/fireaxe.dm
@@ -3,7 +3,7 @@
 	var/datum/effect_system/spark_spread/spark_system	//the spark system, used for generating... sparks?
 
 /obj/structure/fireaxecabinet/bridge
-	req_access = list(ACCESS_CAPTAIN)
+	req_access = list(ACCESS_CAPTAIN, ACCESS_CE)
 
 /obj/structure/fireaxecabinet/Initialize()//<-- mirrored/overwritten proc
 	. = ..()
--- a/yogstation/code/game/objects/structures/fireaxe.dm
+++ b/yogstation/code/game/objects/structures/fireaxe.dm
@@ -3,7 +3,7 @@
 	var/datum/effect_system/spark_spread/spark_system	//the spark system, used for generating... sparks?
 
 /obj/structure/fireaxecabinet/bridge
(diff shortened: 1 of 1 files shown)


Aquizit:
Ah I was going to make it all heads, not just CE. But I mean that can be decided before merging.

I was the one who suggested it needed Balance Council review, by the by, since the original PR of locking the bridge fire axe to Captain only seemed intentional and it was written by a council member for whatever weight that's worth (I'm not sure if it's worth any). #11984

(If I'm also wrong about Adam being council in August 2021 then oops.)

yogstation/code/game/objects/structures/fireaxe.dm
```diff
@@ -3,7 +3,7 @@
 	var/datum/effect_system/spark_spread/spark_system	//the spark system, used for generating... sparks?
 
 /obj/structure/fireaxecabinet/bridge
-	req_access = list(ACCESS_CAPTAIN)
+	req_access = list(ACCESS_CAPTAIN, ACCESS_CE)
```
Moltijoe:
```suggestion
	req_one_access = list(ACCESS_CAPTAIN, ACCESS_CE)
```

Moltijoe (changes requested):

Moltijoe (approved):

PR #18366, opened 2023-03-24
Switches Lavaland Syndicate hardsuit to modified Elder Atmosian harsuit

Aquizit:

As the title states, the syndi suit loot from the crate is now a modified elder atmosian. Modified how you say? I'm glad you asked. It has no hardsuit slowdown. It keeps the armor values for it too, which while different from the syndi suit, are not drastic except for a couple (like bomb) and in that case it is the atmosian suit that is better.

# Changelog

:cl:  
tweak: Swaps the red hardsuit for the blue one in lavaland loot.
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/hardsuit.dm
+++ b/code/modules/clothing/spacesuits/hardsuit.dm
@@ -234,6 +234,11 @@
 	flags_1 = RAD_PROTECT_CONTENTS_1
 	slowdown = 0.6
 
+/obj/item/clothing/suit/space/hardsuit/elder_atmosian/lavaland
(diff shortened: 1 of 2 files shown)


MajManatee:
ok but why?

MajManatee:
also it doesnt work at all so

namjay1170:
> ok but why?

because I made a complaint against them for forcing their admin powers on a round that included a red hardsuit and this is their response to it

Aquizit:
It's not like I did it without consulting with council or anything but whatever. It's not worth this.

PR #18393, opened 2023-03-27
The presence of a Xeno Queen no longer delays the shuttle because I had to wait for the next round to start once.

BurgerLUA:

Xeno Queens can no longer delay the shuttle.
Xeno Queens no longer trigger a shuttle call 30 minutes after they spawn.

# Justification

It is pretty easy for a Xeno Queen to make a base to hold up in and delay the round near indefinitely to get enough Xenomorphs, even if the crew is mostly dead or hiding in pods. With this PR, they can still do that, but they are actually put on a timer (if the crew chooses) to Hijack the shuttle when it is called. This prevents rounds going on longer than they should.

# Wiki Documentation

https://wiki.yogstation.net/wiki/Xenos

# Changelog

:cl:  BurgerBB
rscdel: Xeno Queens can no longer delay the shuttle.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/alien/humanoid/queen.dm
+++ b/code/modules/mob/living/carbon/alien/humanoid/queen.dm
@@ -25,9 +25,6 @@
 	var/datum/action/small_sprite/smallsprite = new/datum/action/small_sprite/queen()
 
 /mob/living/carbon/alien/humanoid/royal/queen/Initialize()
-	if(!is_centcom_level(get_turf(src)))
-		SSshuttle.registerHostileEnvironment(src) //yogs: aliens delay shuttle
-		addtimer(CALLBACK(src, .proc/game_end), 30 MINUTES) //yogs: time until shuttle is freed/called
 	//there should only be one queen
 	for(var/mob/living/carbon/alien/humanoid/royal/queen/Q in GLOB.carbon_list)
 		if(Q == src)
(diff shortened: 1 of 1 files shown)


namjay1170:
it can only delay for 10 minutes

BurgerLUA:
> it can only delay for 10 minutes

Actually it was 30 unless I misread it.

namjay1170:
> > it can only delay for 10 minutes
> 
> Actually it was 30 unless I misread it.

you're most likely correct im going off info an admeme said and they arent always reliable

Maia-J:
Make it 10 so the queen cant get completely fucked just by crew instacalling

BurgerLUA:
> > > it can only delay for 10 minutes
> > 
> > 
> > Actually it was 30 unless I misread it.
> 
> you're most likely correct im going off info an admeme said and they arent always reliable

It says 30 minutes in the code I deleted, so unless there is some fuckery going on, I'm right.

Moltijoe:
the 30 minute timer is started when the queen is first created, reducing it to 10 might make it too short, but 30 is perhaps too long, if a queen is living for 30 minutes straight, nearly everyone is already dead.
removing the timer outright probably isn't the right move, just reduce the number or something

Bibby0110:
Just go kill the Queen?

tattax:
blob can no longer delay the shuttle.

Justification
It is pretty easy for a blob to make a base to hold up in and delay the round near indefinitely to get enough tiles, even if the crew is mostly dead or hiding in pods. With this PR, they can still do that, but they are actually put on a timer (if the crew chooses) to get enough tiles when the shuttle when it is called. This prevents rounds going on longer than they should.

BurgerLUA:
> blob can no longer delay the shuttle.
> 
> Justification It is pretty easy for a blob to make a base to hold up in and delay the round near indefinitely to get enough tiles, even if the crew is mostly dead or hiding in pods. With this PR, they can still do that, but they are actually put on a timer (if the crew chooses) to get enough tiles when the shuttle when it is called. This prevents rounds going on longer than they should.

hey thats a good idea let me make the pr

MajManatee:
not even ided, this is a skill issue pr

Marmio64:
30 minutes is kinda long but not having anything is terrible. 10 minutes would also be too short in my opinion and it would better as like 20.

Burger1998:
1. xeno gets called out
2. red alert shuttle calls
3. like 2 xenos burst in that 5 minutes
4. crew sail into the sunset while the rare antag spawn go play ftl or someshit idk.

Bibby0110:
Go kill the Queen

PR #18395, opened 2023-03-27
Blazing oil takes 30% more damage from extinguishers

Notamaniac:

Blazing oil takes 30% more damage from extingiushers. Currently blazing oil is immune to burn damage (i.e. lasers), and extinguisher damage was also nerfed by 33% as well. This doesn't restore the old numbers, but it should hopefully make extinguishers actually viable again. 

# Changelog

:cl:  
tweak: Fire extinguishers deal 30% more damage to blazing oil.
/:cl:


Notamaniac:
Something more fundamental needs to be done, using an extinguisher on the blob deals burn damage, which in turn gives the blob resources. With an immunity to lasers, your only options are an extremely risky melee, the beam rifle (which deals burn damage maybe?) or bombs (that don't kill the core).

PR #18396, opened 2023-03-27
Removes the completely undocumented and unexplained change of fire damage giving Blazing Oil blobs resource points.

BurgerLUA:

PR #14956 added a feature that made it so that blazing oil blobs are immune to fire damage. This was documented and fine.
This PR also added a feature, unlisted in the changelog or the PR itself, that made it so that blazing oil blobs gain resource generation from taking burn damage. This can cause a massive amount of generation to occur if people use lasers against it.

# Wiki Documentation

Nothing needs to be changed because this was undocumented lmao.

# Changelog

:cl:  BurgerBB
rscdel: Removes the completely undocumented and unexplained change of fire damage giving Blazing Oil blobs resource points.
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
+++ b/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
@@ -18,7 +18,6 @@
 
 /datum/blobstrain/reagent/blazing_oil/damage_reaction(obj/structure/blob/B, damage, damage_type, damage_flag)
 	if(damage_type == BURN && damage_flag != ENERGY)
-		var/mob/camera/blob/O = overmind
 		damage = 0 //completely and entirely immune to burn damage!
 		for(var/turf/open/T in range(1, B))
 			var/obj/structure/blob/C = locate() in T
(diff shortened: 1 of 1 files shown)


Burger1998:
*cough "Blazing Oil blob is now COMPLETELY immune to burn damage. Welders, laser guns, PA rifles. All of them do nothing.
In fact, they give the blob a little bit of points back"

**they give the blob a little bit of points back**

Burger1998:
also dunno if im remembering correctly but i think it says somewhere ingame that blazing oil getting hit with heat feeds blob growth? might be a tip or something so your gonna need to get rid of that aswell.

Maia-J:
I believe the heart of the issue is the abysmal damage of water (1 per extinguish with hard blobs having 150 HP), and not point regeneration.

BurgerLUA:
I strongly believe that one of the issues is the insane regeneration you can get from clueless crew. This is a yogstation exclusive change and I doubt a lot of people knew about the blob being able to regen resources from burn damage before today.

There shouldn't be a mechanic that punishes the crew because an inexperienced player shot the blob with a laser rifle.

The-Sun-In-Splendour:
> 

Yes there should be.

Bibby0110:
Not undocumented, open a PR with a less biased accusation or update the explanation blurb

PR #18397, opened 2023-03-27
Extinguishers no longer burn oil blobs

SegaCD:
I don't know why the previous person didn't check the code for this really obvious issue

I changed the tag for what damage extinguishers do to the Blazing Oil strain to BRUTE, ENERGY

# Changelog

:cl:  
bugfix: Changed Blazing oil extinguisher damage tags to BRUTE, ENERGY

/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
+++ b/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
@@ -14,7 +14,7 @@
 	reagent = /datum/reagent/blob/blazing_oil
 
 /datum/blobstrain/reagent/blazing_oil/extinguish_reaction(obj/structure/blob/B)
-	B.take_damage(1, BURN, ENERGY)
+	B.take_damage(1, BRUTE, ENERGY)
 
 /datum/blobstrain/reagent/blazing_oil/damage_reaction(obj/structure/blob/B, damage, damage_type, damage_flag)
 	if(damage_type == BURN && damage_flag != ENERGY)


Notamaniac:
Blobs have an innate 50% brute resistance, which probably will effect the amount of damage extinguishers do, due to to this change.

You might need to double the damage done to compensate.

SegaCD:
Would having it just do energy damage work better?

00ze-cyclone:
> Would having it just do energy damage work better?

ENERGY isn’t a damage type, it define what armor it resolve against.

SegaCD:
> > Would having it just do energy damage work better?
> 
> ENERGY isn’t a damage type, it define what armor it resolve against.
So could I just remove BRUTE and ENERGY from the code or do I need to specify something?

code/modules/antagonists/blob/blobstrains/blazing_oil.dm
```diff
@@ -14,7 +14,7 @@
 	reagent = /datum/reagent/blob/blazing_oil
 
 /datum/blobstrain/reagent/blazing_oil/extinguish_reaction(obj/structure/blob/B)
-	B.take_damage(1, BURN, ENERGY)
+	B.take_damage(1, BRUTE, ENERGY)
 
 /datum/blobstrain/reagent/blazing_oil/damage_reaction(obj/structure/blob/B, damage, damage_type, damage_flag)
 	if(damage_type == BURN && damage_flag != ENERGY)
```
Maia-J, line 20: It already checks to specifically exclude them so this PR's rationale is invalid

PR #18398, opened 2023-03-28
xenoqueen QoL

tattax:
image

xeno hosts don't count as alive anymore so you don't have to kill all hosts when on the shuttle

xeno queen gets a shuttle block timer tell to see when you need to leave

:cl:  
rscadd: xeno QoL: shuttle timer and host mercy
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/alien/humanoid/queen.dm
+++ b/code/modules/mob/living/carbon/alien/humanoid/queen.dm
@@ -48,8 +48,9 @@
 
 /mob/living/carbon/alien/humanoid/royal/queen/get_status_tab_items()
 	. = ..()
-	. += ""
-	. += "Blocked Shuttle Timer: [round(timeleft(time_to_shuttle) / 600, 1)] minutes" //weird conversion but works
+	if(time_to_shuttle)
+		. += ""
+		. += "Blocked Shuttle Timer: [round(timeleft(time_to_shuttle) / 600, 1)] minutes" //weird conversion but works
 
 /mob/living/carbon/alien/humanoid/royal/queen/proc/kill_shuttle_timer()
 	SSshuttle.clearHostileEnvironment(src)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/alien/humanoid/queen.dm
```diff
@@ -45,6 +46,16 @@
 	smallsprite.Grant(src)
 	return ..()
 
+/mob/living/carbon/alien/humanoid/royal/queen/get_status_tab_items()
+	. = ..()
+	. += ""
+	. += "Blocked Shuttle Timer: [round(timeleft(time_to_shuttle) / 600, 1)] minutes" //weird conversion but works
```
ynot01: this may runtime when the timer is deleted

tattax: it doesn't runtime but it is true it shouldn't show up

code/modules/mob/living/carbon/alien/humanoid/queen.dm
```diff
@@ -45,6 +46,17 @@
 	smallsprite.Grant(src)
 	return ..()
 
+/mob/living/carbon/alien/humanoid/royal/queen/get_status_tab_items()
+	. = ..()
+	if(time_to_shuttle)
+		. += ""
+		. += "Blocked Shuttle Timer: [round(timeleft(time_to_shuttle) / 600, 1)] minutes" //weird conversion but works
```
BurgerLUA, line 53: You should probably just make this like the normal shuttle timer as it is more consistent with what already exists, plus you don't have to deal with the funny "1 minutes" grammar issue.

tattax: minutes are fine as this is not a type of info you'd need live updates on, only minutes, plus timeleft does the work just fine

Moltijoe (approved):

PR #18401, opened 2023-03-28
buster style no longer totally negated by soap and flashes

LazennG:

forgot to do this last pr but the old buster attacks were usable if the user was resting and it would be inconsistent for that to not still apply, also the "65% damage reduction" in the recall teachings blurb was fixed to be 50% like it actually is

# Changelog

:cl:  

tweak: tweaked buster arm use conditions
spellcheck: fixed incorrect number in buster arm's description

/:cl:

### DIFF
--- a/code/datums/martial/buster_style.dm
+++ b/code/datums/martial/buster_style.dm
@@ -43,7 +43,7 @@
 /datum/martial_art/buster_style/can_use(mob/living/carbon/human/H)
 	var/obj/item/bodypart/r_arm/robot/buster/R = H.get_bodypart(BODY_ZONE_R_ARM)
 	var/obj/item/bodypart/l_arm/robot/buster/L = H.get_bodypart(BODY_ZONE_L_ARM)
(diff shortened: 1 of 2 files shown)


code/datums/martial/buster_style.dm
```diff
@@ -472,10 +472,10 @@
 	destroys it but uses up the attack. Attacking a living target uses up the attack and sends them flying and dismembers their limb if its damaged enough. Has a 15 second \
 	cooldown."
 
-	combined_msg +=  span_warning("You can't perform any of the moves besides megabuster if you're lying down or have an occupied hand. Additionally, if your buster arm should become \
-	disabled, so shall your moves.")
+	combined_msg +=  span_warning("You can't perform any of the moves besides if you have an occupied hand. Additionally, if your buster arm should become disabled, so shall\
```
Moltijoe:
```suggestion
	combined_msg +=  span_warning("You can't perform any of the moves if you have an occupied hand. Additionally, if your buster arm should become disabled, so shall\
```

Moltijoe (changes requested):

Moltijoe (approved):

PR #18402, opened 2023-03-28
Fixes some nitrium-related oversights

SapphicOverload:

Changes nitrium production to be closer to what stimulum production used to be due to them being very similar in effect:
- Requires twice as much nitrogen
- Makes 0.05x as much nitrium, the same as stimulum used to be
- Heat scale is 8x higher, use fusion if you want to make a significant amount of it.

Nitrium decomposition will now make antinoblium instead of duplicating nitrium.

Also fixes the tesla generator recipe in the crystallizer.

# Changelog

:cl:   
bugfix: fixes nitrium decomposition duplicating nitrium
bugfix: fixes tesla generator recipe requiring two separate instances of nitrium
tweak: makes nitrium formation more like what stimulum formation used to be
tweak: nitrium decomposition now makes antinoblium instead of duplicating itself
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/gas_recipe_machines/atmos_machines_recipes.dm
+++ b/code/modules/atmospherics/machinery/components/gas_recipe_machines/atmos_machines_recipes.dm
@@ -106,7 +106,7 @@ GLOBAL_LIST_INIT(gas_recipe_meta, gas_recipes_list())
 	max_temp = 12000
 	energy_release = -350000
 	dangerous = TRUE
-	requirements = list(/datum/gas/nitrium = 1200, /datum/gas/freon = 500)
+	requirements = list(/datum/gas/nitrium = 500, /datum/gas/freon = 500, /datum/gas/tritium = 500)
 	products = list(/obj/machinery/the_singularitygen/tesla = 1)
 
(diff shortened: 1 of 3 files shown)


code/modules/atmospherics/machinery/components/gas_recipe_machines/atmos_machines_recipes.dm
```diff
@@ -106,7 +106,7 @@ GLOBAL_LIST_INIT(gas_recipe_meta, gas_recipes_list())
 	max_temp = 12000
 	energy_release = -350000
 	dangerous = TRUE
-	requirements = list(/datum/gas/nitrium = 500, /datum/gas/freon = 500, /datum/gas/nitrium = 800)
```
ThatLing, line 109: 1200 mols is quite a lot for a stupid tesla gen. Perhaps something like 500-700mols of nitrium and 100-500mols of BZ?

iamthesilly: changed to 500 moles nitrium, 500 moles freon, and 500 moles tritium

ThatLing (approved):

JamieD1 (approved):

PR #18416, opened 2023-03-29
Attempts to fix problems with the battle royale gamemode

Moltijoe:
Workin on it
it works perfectly fine on my test server so i have no clue what's causing it

:cl:  
bugfix: attempts to make the battle royale gamemode not break 66% of the time
/:cl:

### DIFF
--- a/yogstation/code/game/gamemodes/battle_royale/battleroyale.dm
+++ b/yogstation/code/game/gamemodes/battle_royale/battleroyale.dm
@@ -28,6 +28,8 @@ GLOBAL_LIST_EMPTY(battleroyale_players) //reduce iteration cost
 	if(A)
 		var/turf/T = safepick(get_area_turfs(A)) //Move to a random turf in arrivals. Please ensure there are no space turfs in arrivals!!!
 		new /obj/structure/battle_bus(T)
+	else //please don't ever happen
+		message_admins("Something has gone terribly wrong and the bus couldn't spawn, please alert a maintainer or someone comparable.")
 	for(var/mob/L in GLOB.player_list)//fix this it spawns them with gear on
 		if(!L.mind || !L.client)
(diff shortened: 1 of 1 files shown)


MajManatee:
ah yes the only thing needed to fix it
adding elses

Moja10:
If possible, could the crates be sent to a garbage collector after like 3 - 5 minutes of being open? They clog up the halls, especially when more spawn, making fighting really hard. Other than that, awesome that you are fixing the BR mode!

Moltijoe:
> If possible, could the crates be sent to a garbage collector after like 3 - 5 minutes of being open? They clog up the halls, especially when more spawn, making fighting really hard. Other than that, awesome that you are fixing the BR mode!

i could probably do that, i'm just wanting to make sure it works first though.

yogstation/code/game/gamemodes/battle_royale/battleroyale.dm
```diff
@@ -64,9 +69,14 @@ GLOBAL_LIST_EMPTY(battleroyale_players) //reduce iteration cost
 	var/list/royalers = list()
 	if(GLOB.player_list.len <= 1) //It's a localhost testing
 		return
+	if(!GLOB.battleroyale_players.len) //sanity check for if this gets called before people are added to the list somehow
```
ToasterBiome: use length(x) proc

ToasterBiome (changes requested):

yogstation/code/game/gamemodes/battle_royale/battleroyale.dm
Moltijoe: lazylen or length?

Moltijoe: also, should i change the other .len uses?

ToasterBiome (approved):

PR #18301, opened 2023-03-18
Refactor internals code

ThatLing:
# Document the changes in your pull request

Reverts https://github.com/yogstation13/Yogstation/pull/17992 and ports https://github.com/tgstation/tgstation/pull/70750

Closes https://github.com/yogstation13/Yogstation/issues/18287

PR #18418, opened 2023-03-30
Fixes an update_internals runtime

Mqiib:
Code checked for internal.loc without checking for internals in the first place and threw a runtime when they didn't exist. If it exists it should have a location probably.

# Changelog

:cl:  

bugfix: fixes an unpdate_internals runtime

/:cl:


tattax:
https://github.com/yogstation13/Yogstation/pull/18301

PR #18422, opened 2023-03-30
Makes beanbag slugs fully non-lethal

Moltijoe:
This follows the similar logic as #18358 
This also lets warden fully non-lethally detain people in the brig area without relying on a baton or disabler
Also lets bartender be a bit more trigger happy in RP situations without causing damage and potentially wounding their target (getting them thrown in jail, or worse)

:cl:  
tweak: Beanbag slugs no longer do regular damage
tweak: Hidden reagent darts no longer do regular damage (to match the thing they're hiding as)
/:cl:


MajManatee:
warden  will not use beanbags, are you mental?

RG4ORDR:
Beanbags are not non-lethal.
They are LESS-THAN-LETHAL.

Therandomhoboo:
Aww what this was a good idea

PR #18425, opened 2023-03-30
Makes less-lethals more lethal

Skrem7:

I have decided that instead of pacifying less-lethals we should make them more lethal

Main point of concern I have is the bartender, but I still question why we give a service role a shotgun capable of being sawn off and then loaded with lethal shot. More damaging beanbags matter less if they just stamcrit you (then kill you). Almost like this will happen with less-lethals anyway.

NOTE: All slugs lose 3 damage and 2.25 stamina per tile traveled.
All pellets lose 0.4 damage and 0.3 stamina per tile traveled.

Beanbag slugs now do 18 damage up from 5, but they have lost their positive wound bonus
Rubbershot pellets now do 5.5 damage up from 3
4.6x30mm rubber bullets now do 8 damage up from 5
5.56mm rubber bullets (you will never see these) now do 20 damage up from 10

ADDED BONUS: Dragonsbreath pellets now do 7 damage up from 5

# Wiki Documentation

Values should be updated on Guide to Combat

# Changelog

:cl:  
tweak: Rubber bullets, rubber shot, and beanbags should now all actually hurt significantly more
tweak: Dragonsbreath pellets now do slightly more damage per pellet (notable jump in overall damage)
tweak: Less-lethal bullets that were sharp are now properly blunt
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets/rifle.dm
+++ b/code/modules/projectiles/projectile/bullets/rifle.dm
@@ -17,8 +17,9 @@
 
 /obj/item/projectile/bullet/a556/rubber
 	name = "5.56mm rubber bullet"
-	damage = 10
+	damage = 20
 	stamina = 47
+	sharpness = SHARP_NONE
 
 // .308 (LWT-650 DMR)
 
--- a/code/modules/projectiles/projectile/bullets/shotgun.dm
+++ b/code/modules/projectiles/projectile/bullets/shotgun.dm
@@ -23,7 +23,7 @@
 
 /obj/item/projectile/bullet/incendiary/shotgun/dragonsbreath
 	name = "dragonsbreath pellet"
-	damage = 5
+	damage = 7
 
 /obj/item/projectile/bullet/shotgun/slug/stun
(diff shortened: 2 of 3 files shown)


wonderinghost:
Now we need something for ghetto weapons

Marmio64:
Dragonsbreath is already a lethal round, why does it need a damage increase?

Skrem7:
> Dragonsbreath is already a lethal round, why does it need a damage increase? 

Because I didn't want rubber doing more damage than an ammo type you basically only see nukies use, especially considering syndicate buckshot/slugs

Therandomhoboo:
Only issue i see is that this makes bartenders more viable for arrests/ai validing. Or others that arent sec

00ze-cyclone:
Finally I can stun and execute with the same type of round.
Killing innocents with lethal rounds is just plain and boring murder, true police brutality is shooting someone with rubber until they are pulped. We can expect an increase in the quality of sec RP: no longer operative wannabe, true mall cop violence now.

PR #18426, opened 2023-03-31
Fixes hypernoblium crystals

SapphicOverload:

Hypernoblium crystals no longer makes your clothes cover parts of the body they shouldn't which caused problems like allowing helmets' armor to protect the rest of your body.

Also allows using the crystals on jumpsuits, this doesn't make them spaceproof but still prevents them from being destroyed by fire like any other clothes you use it on.

And also changed the message slightly for when you use it and made the name not a proper noun.

:cl:  
tweak: hypernoblium crystal can be applied to jumpsuits and has a slightly different text when you use it
bugfix: hypernoblium crystals no longer cause armor to protect parts of the body they shouldn't
spellcheck: hypernoblium crystal no longer a proper noun
/:cl:


MajManatee:
ive never seen anyone with hypernoblium, let alone crystals of it

iamthesilly:
> ive never seen anyone with hypernoblium, let alone crystals of it

Yeah this went unnoticed for so long because no one has ever made them until recently

Moltijoe (approved):

PR #18432, opened 2023-03-31
DeathNettle, less bang for your buck

Therandomhoboo:

Look, botanists speedrun death nettles as a some weird power fantasy. So reduction to the damage output should either force the botanists to stop doing that or commit fully to it. 

# Changelog

:cl:  

tweak: Deathnettle does up to 19 max now,

/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/nettle.dm
+++ b/code/modules/hydroponics/grown/nettle.dm
@@ -97,11 +97,11 @@
 
 /obj/item/reagent_containers/food/snacks/grown/nettle/death/add_juice()
 	..()
-	force = round((5 + seed.potency / 4), 1) // Max 30 dmg, esword level, with max potency and buffed force reduction, should down unarmored in 3-4 hits
+	force = round((5 + seed.potency / 5.5), 1) // Max 19 dmg, with max potency and buffed force reduction, should down unarmored in 3-4 hits
 
 /obj/item/reagent_containers/food/snacks/grown/nettle/death/pickup(mob/living/carbon/user)
 	if(..())
-		if(prob(50))
+		if(prob(30))
 			user.Paralyze(100)
 			to_chat(user, span_userdanger("You are stunned by the Deathnettle as you try picking it up!"))
 
--- a/code/modules/hydroponics/grown/nettle.dm
+++ b/code/modules/hydroponics/grown/nettle.dm
@@ -97,11 +97,11 @@
 
 /obj/item/reagent_containers/food/snacks/grown/nettle/death/add_juice()
 	..()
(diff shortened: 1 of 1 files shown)


Moja10:
what warranted this at all? deathnettles aren't really an issue for being abused or killing too easy. They degrade in damage per hit so that 19 damage will become 14, to 10, to basically nothing so really you are just making them worse than their roundstart hatchet or a toolbox. They only get the 30 damage on the first hit.

Therandomhoboo:
> what warranted this at all? deathnettles aren't really an issue for being abused or killing too easy. They degrade in damage per hit so that 19 damage will become 14, to 10, to basically nothing so really you are just making them worse than their roundstart hatchet or a toolbox. They only get the 30 damage on the first hit.

When does anyone use deathnettles round start or at anything less than potency 100? Its extremely easy to max out potency with a few minutes of botany. As well as no one uses just 1 nettle, this basically forces people that want to spam it to spend more time making them and stops people getting heavily fucked by it. Its been done in the past and has started to creep up again.

I would say at least see how it goes, if it really screws d.nettles over. I can reduce it by 0.5 making it 21-22damage instead.

iamthesilly:
just give them negative AP instead since it's a fragile plant that wouldn't be able to get through armor very well

Therandomhoboo:
I'll see how it goes through balance review. If the damage is too much, then I'll make it negative ap. To save the tears from botany mains for for some reason need to spam produce d.nettles.

stwordcitizen:
> what warranted this at all? deathnettles aren't really an issue for being abused or killing too easy. They degrade in damage per hit so that 19 damage will become 14, to 10, to basically nothing so really you are just making them worse than their roundstart hatchet or a toolbox. They only get the 30 damage on the first hit.

The issue with nettles isn't hitting people with them its throwing them which causes a huge amount of acid and other chems to enter their bloodstream and acid to melt their gear

Therandomhoboo:
> > what warranted this at all? deathnettles aren't really an issue for being abused or killing too easy. They degrade in damage per hit so that 19 damage will become 14, to 10, to basically nothing so really you are just making them worse than their roundstart hatchet or a toolbox. They only get the 30 damage on the first hit.
> 
> The issue with nettles isn't hitting people with them its throwing them which causes a huge amount of acid and other chems to enter their bloodstream and acid to melt their gear

It is both, death nettle is one of the stronger melee weapons for some bizarre reason. There is already a PR reducing its acid and this one is to bring it down to some reasonable levels so it doesn't act like an esword since it really shouldn't.

Moltijoe:
the other nerf to reduce acid content was deemed to be plenty in terms of nerfing deathnettle

PR #18439, opened 2023-04-01
Prevents using the buster arm if you already have a martial art

Moltijoe:
No more getting clowned on by overwriting your existing martial art

:cl:  
tweak: Prevents using the buster arm if you already have a martial art
/:cl:

### DIFF
--- a/code/game/objects/items/devices/busterarm/buster_limb.dm
+++ b/code/game/objects/items/devices/busterarm/buster_limb.dm
@@ -47,6 +47,9 @@
 		return
 	if(!ishuman(L))
 		return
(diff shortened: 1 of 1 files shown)


ynot01:
Wouldn't this prevent chefs from using the buster arm?

Marmio64:
> Wouldn't this prevent chefs from using the buster arm?

looks like they'll need to cook something else up

Moltijoe:
> Wouldn't this prevent chefs from using the buster arm?

I'll check, they might just need to leave the kitchen to use it
edit: nope, i'll need to add a special check

ToasterBiome (approved):
ok

PR #18457, opened 2023-04-03
Disable Lavaland Beach Dome Ruin

Aquizit:

Every time this ruin spawns currently it's glitched out with large areas just pitch black. Right clicking on the tile says it's what it is supposed to be, but the tile just isn't appearing.

image

:cl:  
tweak: Disables Lavaland Beach Dome ruin until the visual glitches on it can be fixed.
/:cl:

### DIFF
--- a/config/lavaruinblacklist.txt
+++ b/config/lavaruinblacklist.txt
@@ -6,7 +6,7 @@
 ##BIODOMES
 #_maps/RandomRuins/LavaRuins/lavaland_surface_biodome_winter.dmm
 #_maps/RandomRuins/LavaRuins/lavaland_surface_biodome_winter_inn.dmm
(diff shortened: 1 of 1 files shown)


namjay1170:
Would you kindly delete the snow dome too? originaly it had an antique laser but someone removed that and replaced it with NOTHING, its just a waste of lavaland budget (same with that dogshit one with only a combat knife like why would anyone map and pr that wtf)

ToasterBiome (changes requested):
could u see why this happens instead maybe uwu

Aquizit:
I do not know how to do that. I don't even know how to begin figuring that out

Aquizit:
Alright well if this is gonna languish I'll just close it. I dunno how to fix it as I dunno what's going wrong otherwise I would have tried for the record.

PR #18460, opened 2023-04-04
You no longer heed a helmet to be space-proof if you don't have a head

SapphicOverload:

Why would you need a helmet if you don't need a head?

# Changelog

:cl:  
tweak: you no longer heed a space helmet if you don't have a head
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -68,6 +68,10 @@
 		var/obj/item/clothing/CH = head
 		if (CS.clothing_flags & CH.clothing_flags & STOPSPRESSUREDAMAGE)
 			return ONE_ATMOSPHERE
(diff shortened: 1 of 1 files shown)


wonderinghost:
Heed the helmet

Marmio64:
most suits probably are designed for people with heads so the top of it would just be entirely open unless you cover the hole to stop your body from being exposed.

Bibby0110:
hello how are you going to seal the suit???

MajManatee:
they got a point

githubuser4141:
> most suits probably are designed for people with heads so the top of it would just be entirely open unless you cover the hole to stop your body from being exposed.

maybe nanotrasen are inclusive enough to put some spare material at the neck cuff to seal the suit if you don't have a head?
it is 25th century after all :)

wonderinghost:
It's slime fuckery

iamthesilly:
Mainly made this for IPCs because they can live without a head but can't wear a helmet  without it so if they lose it (which is somewhat common) they have no way to be spaceproof whatsoever but making it so space helmets can be used without a head would break a lot of stuff

Moltijoe (approved):

PR #18452, opened 2023-04-03
Lets IPCs use vaccines

Moltijoe:
They can get specialized diseases, might as well let them use the specialized vaccine

:cl:  
tweak: Lets IPCs use vaccines
/:cl:

PR #18486, opened 2023-04-07
tweaks IPC viruses (system cleaner and immunity redux pt2 FINAL (probably))

Moltijoe:
reverts #18452 (it's been over 48 hours since the merge, please don't close this bibby)

Vaccines no longer work on IPC because, why would they
Lets system cleaner instead be used to remove viruses (gives it a purpose more so than just for the 10% chance for toxin damage from emp)
Makes them ignore most virus symptoms that would make little sense for them to be affected by

this has better flavour since most symptoms mention targetting a specific organ in the body
lets IPC act mostly as virus carriers since most viruses target organic processes

This restores one of IPC's main identity of being nigh completely virus resistant, while not entirely ignoring the system like they did before

works using the same flag system as reagents do

:cl:  
tweak: Vaccines no longer work on IPCs
tweak: IPC can purge viruses with system cleaner
tweak: most virus symptoms no long affect mobs with synthetic chemical processing 
/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/deafness.dm
+++ b/code/datums/diseases/advance/symptoms/deafness.dm
@@ -32,7 +32,7 @@ Bonus
 		"Resistance 9" = "Causes permanent deafness, instead of intermittent.",
 		"Stealth 4" = "The symptom remains hidden until active.",
 	)
-	affectrobot = TRUE
+	process_flags = ORGANIC | SYNTHETIC
 
 /datum/symptom/deafness/Start(datum/disease/advance/A)
 	. = ..()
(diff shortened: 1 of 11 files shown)


Notamaniac:
so when are we addressing the fact that borgs can also catch viruses

Moltijoe:
> so when are we addressing the fact that borgs can also catch viruses

Never
Fuck, they can?
I assumed it had a check to only let carbons catch diseases

Moltijoe:
Whoops

MajManatee:
Synthetic /= Robotic, that is a concern

Moltijoe:
> Synthetic /= Robotic, that is a concern

that's how it's used for reagents, i'm just being consistent

ToasterBiome (approved):

PR #18493, opened 2023-04-07
[Asteroid] Opening Podbay doors will no longer suck you into space.

Aquizit:

Sets Asteroid Station podbay as its own area like it's supposed to be, with its own APC and such. Also adds the podbay green bar thingy that acts as the barrier that keeps air getting sucked into space when the doors open.

# Changelog

:cl:  
bugfix: Correctly marks Podbay as its own area, adds the APC for it, and fixes the podbay doors so the forcefield is there.
/:cl:


yacabo:
Open the pod bay doors, Hal

ToasterBiome (approved):

PR #18495, opened 2023-04-08
fixes a decal that is annoying me

ktlwjec0:
image

now i did both these doors
image


ToasterBiome (changes requested):
add the actual warning line that should be there in front of the door instead

ToasterBiome:
also why is there hundreds of line changes

ToasterBiome (approved):

PR #18503, opened 2023-04-09
Increase bluespace activity maximum throw range to 3 tiles

SomeguyManperson:

2 tiles is almost melee range, 3 tiles should be a "thrown" weapon without being able to effectively outrange an actual gun/thrown weapon

This is just under half the default 15x15 view range

# Spriting

# Wiki Documentation

# Changelog

:cl:  
tweak: bluespace activity maximum throw range now 3 from 2
/:cl:


Cartlord:
Honestly, I wouldn't mind having it reset to normal, especially since the trick that let people easily get full Potency every time got removed.

MajManatee:
People still get full potency every shift consistently, but this seems reasonable given how small a 2 tile range is.

Moltijoe (approved):

PR #18510, opened 2023-04-10
Lets pacifists throw the possessed sword

Moltijoe:
it doesn't actually do any damage because throw force is never used
this just lets pacifists throw the sword to put it into mob form

:cl:  
bugfix: Lets pacifists throw the possessed sword
/:cl:


MajManatee (approved):
he probably (didnt) test this

PR #18511, opened 2023-04-11
nerfs nitrium D:

Notamaniac:

Nitrium now deals 1 tox damage and makes you jitter.

Currently nitrium is easy to make compared to stimulum (which at least requires fusion temps) and has no downsides if inhaled in a low enough concentration with pluxonium (plux also being easy to make with the trit fire needed to make nitrium). This adds a downside, which shouldn't be too extreme, but enough to at least make huffing it 24/7 not an option anymore. The jitter is more to add a tell that said guy is stun immune, though I might also consider adding an examine text.

Just a reminder, nitrium is basically stimulum. (Which makes you stun immune and sleep immune and heals stamina damage faster.)

# Wiki Documentation

Nitrium deals 1.5 tox damage for every 2 points of stamina damage it heals and makes you jittery

# Changelog
:cl:  
tweak: Nitrium deals 1.5 tox damage per tick for every 2 points of stamina damage it heals and makes you jittery.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -1309,6 +1309,8 @@
 
 /datum/reagent/nitrium_low_metabolization/on_mob_life(mob/living/carbon/M)
 	M.adjustStaminaLoss(-2 * REM, FALSE)
+	M.adjustToxLoss(1 * REM,FALSE)
+	M.Jitter(15)
 	return ..()
 
 /datum/reagent/nitrium_high_metabolization
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -1308,8 +1308,9 @@
 	return ..()
 
 /datum/reagent/nitrium_low_metabolization/on_mob_life(mob/living/carbon/M)
-	M.adjustStaminaLoss(-2 * REM, FALSE)
-	M.adjustToxLoss(1 * REM,FALSE)
+	if(M.getStaminaLoss() > 0)
+		M.adjustStaminaLoss(-2 * REM, FALSE)
+		M.adjustToxLoss(1.5 *REM, FALSE)
 	M.Jitter(15)
 	return ..()
 


SomeguyManperson:
> you can't really rush an actual significant ammount of nitrium at roundstart anymore, a recent PR made it require harder temps. You can only get good ammounts at 25 mins into shift at the minimum. Ided PR too. also you can just use lethals.

Found the atmos main

Notamaniac:
> you can't really rush an actual significant ammount of nitrium at roundstart anymore, a recent PR made it require harder temps. You can only get good ammounts at 25 mins into shift at the minimum.

You do realise that a single mole is all you need right?

>also you can just use lethals.

The swarmer shoots the atmos tech with his non lethal disabler to stun him. He fails and he is told - just use lethals bro
The horror uses his tentacle to try and stun the atmos tech. He fails and he is told - just use lethals bro
The clown tries to slips the angry atmos tech chasing him with a fire axe. He fails and he is told - just use lethals bro

Just use charcoal bro :D

iamthesilly:
Sadly due to the nature of breath code this will make it near-useless, as nitrium takes up to 8 seconds to have an effect (sometimes even more depending on server lag) and does not clear existing stuns when it does, and most fights with someone using a stun weapon like a disabler or baton will be over before it has a chance to kick in. You need to be breathing it constantly for it to be useful, and this will crit you in less than two minutes. You cannot use charcoal to counter this because it just purges the nitrium.

Notamaniac:
> Sadly due to the nature of breath code this will make it near-useless, as nitrium takes up to 8 seconds to have an effect (sometimes even more depending on server lag) and does not clear existing stuns when it does, and most fights with someone using a stun weapon like a disabler or baton will be over before it has a chance to kick in. You need to be breathing it constantly for it to be useful, and this will crit you in less than two minutes. You cannot use charcoal to counter this because it just purges the nitrium.

Okay I made it so you only take tox damage when you have stamina damage. You should still be able to huff it, but over relying on the stun immunity will kill you sooner or later.

iamthesilly:
> Okay I made it so you only take tox damage when you have stamina damage. You should still be able to huff it, but over relying on the stun immunity will kill you sooner or later.

I like this idea, I think you forgot to commit the change though

DylonBR:
> Okay I made it so you only take tox damage when you have stamina damage. You should still be able to huff it, but over relying on the stun immunity will kill you sooner or later.

this is actually a great idea. would be cool if charcoal didn't purge chems that you get from internals. but good anyways

Moltijoe (approved):
code checks out

code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -1308,7 +1308,10 @@
 	return ..()
 
 /datum/reagent/nitrium_low_metabolization/on_mob_life(mob/living/carbon/M)
-	M.adjustStaminaLoss(-2 * REM, FALSE)
+	if(M.getStaminaLoss() > 0)
+		M.adjustStaminaLoss(-2 * REM, FALSE)
+		M.adjustToxLoss(1.5 *REM, FALSE)
```
ThatLing, line 1313: Why did you change it from 1 to 1.5?

```suggestion
		M.adjustToxLoss(1 * REM, FALSE)
```

ThatLing (dismissed):

code/modules/reagents/chemistry/reagents/other_reagents.dm
Notamaniac: Since the tox damage only occurs when you take stamina damage, I thought to up the damage a little to compensate. Since 3 was a little too extreme, and 1 was a little too low, I went with 1.5.

PR #18515, opened 2023-04-11
Makes gas masks fireproof

Moltijoe:
both the fire fighter suit and fire fighter helmet are fireproof, why not the mask that literally comes with them

this also lets plasmamen that get the martial art have an actual option for not suffocating when they take their helmet off to do more damage

:cl:  
tweak: Makes basic gas masks fireproof
/:cl:


MajManatee:
thick rubber does tend to not catch fire that easily

MajManatee (approved):
i suppose this is ok

PR #18516, opened 2023-04-11
Adds synth blood as preternis blood

Moltijoe:
Basically universal donor blood (that also includes exotic bloodtypes)

image

Made via a reaction of 
1 unstable mutagen (since i can't use blood because of synthmeat recipe)
1 cryoxadone
1 water (can be swapped for something else to balance)(probably should, to prevent saline invalidation)

also removes the safety check that vampires have for blood sucking exotic blood since bloodsuckers didn't have one
this will let them succ preterni again, but they need to actually think for even a moment before trying to suck an ethereal or poly
also, it's funny watching them suck bad blood

:cl:  
rscadd: Adds new reagent Synthblood that acts like all blood types (exotic too)
tweak: Changes preternis exotic blood from stable plasma to synthblood
tweak: Removes safety check for vampires trying to drink exotic blood (bloodsuckers didn't have one)
/:cl:

### DIFF
--- a/code/game/objects/effects/decals/cleanable/humans.dm
+++ b/code/game/objects/effects/decals/cleanable/humans.dm
@@ -14,6 +14,13 @@
 			C.bloodiness += bloodiness
 	return ..()
 
+/obj/effect/decal/cleanable/blood/synthetic //mainly for use in recolouring
+	name = "unusual blood"
+	desc = "It's unpleasant and gooey. Perhaps it's the chef's cooking?"
+	bloodiness = 0
+	icon_state = "genericsplatter1"
+	random_icon_states = list("genericsplatter1", "genericsplatter2", "genericsplatter3", "genericsplatter4", "genericsplatter5", "genericsplatter6")
+
 /obj/effect/decal/cleanable/blood/old
 	name = "dried blood"
 	desc = "Looks like it's been here a while.  Eew."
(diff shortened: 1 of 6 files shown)


neeshacark:
might be wise to remove the virology cheese to make blood duplicate if you do this, not sure how you'd work that out without fucking up viro though.

00ze-cyclone:
Maybe make it able to carry viruses and when the virology machine make culture bottles it’s of that synthblood, because the culture bottles already say that the content is synthetic blood.

Moltijoe:
> might be wise to remove the virology cheese to make blood duplicate if you do this, not sure how you'd work that out without fucking up viro though.

i'm sorry, what?
oh, you mean unstable mutagen + blood to make more blood?
i don't think that's a problem, since that just creates more of the same blood type
this is more of a replacement/upgrade to saline than anything

Moltijoe:
> Maybe make it able to carry viruses and when the virology machine make culture bottles it’s of that synthblood, because the culture bottles already say that the content is synthetic blood.

i'd rather just change the description instead of letting virologists print blood

00ze-cyclone:
> > Maybe make it able to carry viruses and when the virology machine make culture bottles it’s of that synthblood, because the culture bottles already say that the content is synthetic blood.
> 
> i'd rather just change the description instead of letting virologists print blood

They already print blood and there's more effective ways that are right in medbay to print blood anyway.

Moltijoe:
> They already print blood and there's more effective ways that are right in medbay to print blood anyway.

i'll see what i can do for virus transfer

Moltijoe:
blood code is too intertwined with everything, i'm just going to give them a universal bloodtype rather than whatever this jank is

PR #18519, opened 2023-04-11
[admin] Record player living playtime in notes

adamsong:

Makes it easier for admins to determine frequency of rule breaks. 

image

# Changelog

:cl:  
rscadd: Admin notes now record the living playtime of a player when they were noted. 
/:cl:

### DIFF
--- a/SQL/database_changelog.txt
+++ b/SQL/database_changelog.txt
@@ -2,12 +2,16 @@ Any time you make a change to the schema files, remember to increment the databa
 
 The latest database version is 5.11; The query to update the schema revision table is:
 
-INSERT INTO `schema_revision` (`major`, `minor`) VALUES (5, 11);
+INSERT INTO `schema_revision` (`major`, `minor`) VALUES (5, 12);
 or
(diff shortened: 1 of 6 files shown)


alexkar598 (commented):
I'd rather this be an SQL TRIGGER instead so that all services compute the playtime in the same way in a single place ngl

adamsong:
> I'd rather this be an SQL TRIGGER instead so that all services compute the playtime in the same way in a single place ngl

Mysql does not allow for modification of the new row in the trigger.

alexkar598:
Worked for me on mariadb 10.11
 ```sql
DELIMITER $$
CREATE TRIGGER add_playtime
BEFORE INSERT
ON erro_messages FOR EACH ROW
BEGIN 
    SET NEW.playtime = (SELECT minutes FROM erro_role_time rt WHERE rt.ckey = NEW.targetckey AND rt.job = 'Living');
END$$
DELIMITER ;
```

alexkar598 (approved):
Looks poggers

PR #18535, opened 2023-04-12
Nullrod Recall Check

### LINKED ISSUE #18517: Recalling possessed blade without an open hand deletes it
## Test Merges: 
Not Supplied / None
## Reproduction:
I was experimenting with the possessed blade and decided to recall it while holding onto a poster and a jacket, covering both my hands. I was quick to realize that this caused the blade to disappear from the game. I suspect this edge case didn't get caught during testing because of the relative rarity of situations in which you'll be pressing the recall button without intending to use the blade immediately after.
## Supplied Image:

SuppliedImage
 Submitted by: Flikr (168806901457354753)

MajManatee:
Fixes  #18517 

# Changelog

:cl:  
bugfix: Adds a check to make sure you have an empty hand before recalling your null rod.
/:cl:

### DIFF
--- a/code/game/objects/items/holy_weapons.dm
+++ b/code/game/objects/items/holy_weapons.dm
@@ -942,8 +942,8 @@ it also swaps back if it gets thrown into the chaplain, but the chaplain catches
 	action_icon_state = "swordrecall"
 
 /obj/effect/proc_holder/spell/targeted/recallnullrod/cast(list/targets, mob/user)
-	if(!owner.get_empty_held_indexes())
-		to_chat(usr, span_warning("You need an empty hand to call forth your [sword.name]!"))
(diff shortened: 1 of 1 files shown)


Moltijoe:
Put the check on the catch part instead, where it just drops to the ground instead of catching.
If you do it this way, someone can throw the sword at you while your hands are full to delete the sword

MajManatee:
Not exclusive with #18543 and both should probably be merged to cover all bases.

Moltijoe:
> Not exclusive with #18543 and both should probably be merged to cover all bases.

I want people to call it when their hands are full and have it smack them in the face though

MajManatee:
understandable

PR #18445, opened 2023-04-02
Fixes some inventory stuff

SapphicOverload:
# Document the changes in your pull request

When taking off a jumpsuit, items will no longer fall off of you if you're able to equip them in their current slot without said jumpsuit.

# Changelog

:cl:  
bugfix: fixes items falling off when removing a jumpsuit when you're still able to equip them in their current slot without one
/:cl:

PR #18541, opened 2023-04-13
Fixes a bug involving dropping pocket items or belts when stripped

Moltijoe:
Closes: #18540

Fixes a bug introduced by #18445 where no one would drop pockets when stripped, not just the intended cases

Gotta remove the uniform before checking to remove items from pockets

:cl:  
bugfix: Fixes a bug involving dropping pocket items or belts when stripped
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/inventory.dm
+++ b/code/modules/mob/living/carbon/human/inventory.dm
@@ -163,6 +163,10 @@
 				update_inv_w_uniform()
 			update_inv_wear_suit()
 	else if(I == w_uniform)
+		w_uniform = null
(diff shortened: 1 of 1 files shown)


BurgerLUA:
I think what the bug also didn't point out is that I think it applies to belts too. Does that fix it too?

Moltijoe:
> I think what the bug also didn't point out is that I think it applies to belts too. Does that fix it too?

ah, yeah, it dropped my toolbelt too, but the pockets sounded like the main focus
lemme just add that to the changelog

ToasterBiome (approved):

PR #18544, opened 2023-04-13
Changes nullrod hammer to work nicer and throw a bit further

Moltijoe:
it was coded in a janky pseudo throw way that made it difficult to throw the target farther than a single tile

now it just uses a regular throw, but gives the person being thrown a temporary trait that removes the impact effect of the throw

increased the throw distance from only 1 to between 1 and 5 tiles

also has a 1% chance to throw someone 50 tiles (very funny)

:cl:  
tweak: Nullrod hammer hit now throws between 1 and 5 tiles (still no knockdown)
/:cl:


ReddicusDragon (approved):
molti "I tested this" joe

PR #18582, opened 2023-04-17
Makes resonators replicating to ALL adjacent tiles a feature rather than a bug so the admins cant ban me for using it

SomeguyManperson:

Allows the resonator to break rocks extremely quickly in the case there's a mineral vein, regardless of its contents, giving it a small but hopefully notable niche in the pre-plasmacutter early-game

Also it might potentially not stack resonator bubbles on eachother now. I think.

# Changelog

:cl:  
tweak: resonators replicating to all nearby rocks is no longer a bug
bugfix: probably stops resonator bubbles from overlapping when replicating because it is annoying
/:cl:

### DIFF
--- a/code/modules/mining/equipment/resonator.dm
+++ b/code/modules/mining/equipment/resonator.dm
@@ -121,5 +121,5 @@
 	if(!istype(M) || !M.mineralType) // so we don't end up in the ultimate chain reaction
 		return
 	for(var/turf/closed/mineral/T in orange(1, M))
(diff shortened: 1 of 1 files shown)


ToasterBiome:
someone is ambiguous

TheBombMakerJL:
this was how it was before, so bugfix or it was literally never intended which is quite funny

ToasterBiome (approved):

PR #18585, opened 2023-04-17
Nitrium now deals a small amount of liver damage while active instead of a pitiful amount of toxin damage

SomeguyManperson:

Having a passive full antistun with 0 downsides is bullshit, if it's annoying to use proactively that's fine because it's a full immunity to stun weaponry

# Wiki Documentation

nitryl now deals liver damage instead of toxin damage, and no longer has the check for stamina damage

# Changelog

:cl:  
tweak: nitryl now deals liver damage instead of toxin damage
tweak: nitryl's damage is now constant, rather than only occuring if it starts healing stamina damage. No more ignoring beepsky, sorry.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -1342,6 +1342,7 @@
 
 /datum/reagent/nitrium_low_metabolization/on_mob_life(mob/living/carbon/M)
 	M.adjustOrganLoss(ORGAN_SLOT_LIVER, 1)
+	M.adjustStaminaLoss(-2 * REM, FALSE)
 	M.Jitter(15)
 	return ..()
 
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -1341,9 +1341,7 @@
 	return ..()
 
 /datum/reagent/nitrium_low_metabolization/on_mob_life(mob/living/carbon/M)
-	if(M.getStaminaLoss() > 0)
-		M.adjustStaminaLoss(-2 * REM, FALSE)
-		M.adjustToxLoss(1.5 *REM, FALSE)
+	M.adjustOrganLoss(ORGAN_SLOT_LIVER, 1)
 	M.Jitter(15)
 	return ..()
 


DylonBR:
This will make it useless because it does not clear any current stuns and takes around 8 seconds to actually apply the chem after you turn on internals. Also why does every single atmos/xenobio thing get nerfed to the ground after people use it for their intended purposes?

DylonBR:
A person that is stun immune is also not invincible. They still get slowed by stamina damage and normal damage. You also KNOW that they have nitrium or some other anti stun cause they will be jittery

SomeguyManperson:
> This will make it useless because it does not clear any current stuns and takes around 8 seconds to actually apply the chem after you turn on internals. Also why does every single atmos/xenobio thing get nerfed to the ground after people use it for their intended purposes?

no downside permanent invulnerability to standard security equipment being an "intended behavior" sounds like bad faith design, especially with how fast it can be produced and how easily it can be distributed.

> A person that is stun immune is also not invincible. They still get slowed by stamina damage and normal damage. You also KNOW that they have nitrium or some other anti stun cause they will be jittery

it's not the only thing that can cause jittery. Security players won't be able to get off lethalling someone for, say, getting shocked on a door, because "he had the antistun gas!!"

DylonBR:
Can you make it so that the chem is applied instantly when you turn on internals then? Also theres no way to heal toxin damage without purging the chem so you would eventually have to stop to take charcoal.

SomeguyManperson:
> theres no way to heal toxin damage without purging the chem

This isn't true and you know it, come on

code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -1341,9 +1341,8 @@
 	return ..()
 
 /datum/reagent/nitrium_low_metabolization/on_mob_life(mob/living/carbon/M)
-	if(M.getStaminaLoss() > 0)
-		M.adjustStaminaLoss(-2 * REM, FALSE)
-		M.adjustToxLoss(1.5 *REM, FALSE)
+	M.adjustOrganLoss(ORGAN_SLOT_LIVER, 1)
```
ThatLing, line 1344: This will always deal damage and thus also affect high metabolization, revert the ``getStaminaLoss`` check.

ThatLing (changes requested):

ThatLing:
Description and changelog are wrong and doesn't mention the ``getStaminaLoss`` check being removed

code/modules/reagents/chemistry/reagents/other_reagents.dm
SomeguyManperson: updated changelog

ThatLing: This review isn't about the changelong

SomeguyManperson: the cost of having an antistun prepared all the time 🤷

DylonBR:
> > theres no way to heal toxin damage without purging the chem
> 
> This isn't true and you know it, come on

Honestly have no clue what chem you could possibly use to heal the toxin damage without also purging the chem

tattax:
ignore atmos mains kill atmos mains ignite atmos mains eat atmos mains

Moltijoe:
balance council has deemed that while a nerf is within reason, this isn't how it should be nerfed

PR #18592, opened 2023-04-18
Wire splicing from Eris

ToasterBiome:

Ports wire splicing from Eris. They are random wire splice piles that spawn in maint. (although no maps have them) Moving over them has a chance to shock you depending on their level from 1-10. Each level increases the chances to be shocked every Cross by 10%. You cannot make them yourself. using cable coil on them will increase how large the pile is, along with the chance that you will actually get shocked by them. 
Wirecutting will remove the entire thing. 
Attempting to cut or attempting to add wires will attempt to shock you.

Also adds a spawner so you can pop them on maps, one that's 100% spawn and one that's only 30% spawn. You can place them within a group of wires and if it's not already above a wire it will look for an available place to move within 3 tiles, if you want it to be a bit more dynamic.

# Spriting
dreamseeker_eGCdvFjexE

Spawned wire splicings of various stages.

# Wiki Documentation

# Changelog

:cl:  ToasterBiome, NanakoAC
rscadd: Adds wire splicing from eris, and the associated random spawner
/:cl:


Cartlord:
They seem remarkably small, so I don't like the idea of not being to see them in dark-ass maints. On top of that, it seems that gloves are what block the shocks, which doesn't make sense unless everyone canonically walks on their hands.

N3D6:
toasterbiome added 23conflicts 2 years ago

solwashere:
> They seem remarkably small, so I don't like the idea of not being to see them in dark-ass maints. On top of that, it seems that gloves are what block the shocks, which doesn't make sense unless everyone canonically walks on their hands.

The gloves part I agree on but I don't see what's wrong with unbabyfying maint a bit
Why would you be running around maints without a flashlight anyways?

ToasterBiome:
> They seem remarkably small, so I don't like the idea of not being to see them in dark-ass maints. On top of that, it seems that gloves are what block the shocks, which doesn't make sense unless everyone canonically walks on their hands.

The problem here with just the electrical resistent design in general in SS13 is that functionally all of the electrical resistance is placed solely on the insulated gloves. No, really. I COULD add this snowflake unique rubber shoes or just make it so that wearing any shoe cancels the effect but I don’t think it’s worth it. Making a rubber soled shoe etc etc will just break the unifying “ok if you’re wearing insuls you’re okay”, which I’m not keen on changing. Hell, you even become immune to SM shocks if you’re wearing insuls.

Secondly I don’t agree with making them functionally ignorable with shoes. We already have a mechanic like this in maint (glass shards) and I don’t see a point in adding another. 99% of the time you will be in shoes, so 99% of this time this entire mechanic will just be ignored.

For the size, yeah it’s supposed to be kinda hard to see. As they get stronger they get more obvious (along with the wire being blue anyway) so Indont believe this is a problem. Honestly if you’re running into dark maints anyway I think you should rethink your choices, who actually does that?

MajManatee:
I will rewire maint to be blue and make it even harder to see

cuackles (approved):

adamsong (approved):
lgtm

PR #18601, opened 2023-04-19
Makes fairy boots nullrod more faithful to the source material

Moltijoe:
Not only is 100% consistent movespeed strong
but it's not very faithful to the terraria fairy boots that the nullrod is based on

Now it has a 0.3 slowdown than gets reduced by 0.05 per step down to a minimum of -0.4
this speed boost resets if you stand still for at least 0.8 second(s?)
this makes it worse for all-purpose situations, but better for long-distance

the visual effect only starts appearing once it starts speeding you up
also adds a sound effect once above that same speed
https://user-images.githubusercontent.com/108117184/232969150-a02672bc-d760-4c52-8328-17ef26646f06.mp4

:cl:  
tweak: fairy boots nullrod now starts with slowdown and builds up speed as you walk continuously
soundadd: adds a sound effect to being at high speeds with the fairy boots
/:cl:


ChuChu42:
so it speeds up the more you run, just like terraria, and also has the sound effect, pretty nice

MajManatee:
Yes... that's what the description says. Glad we have people that can read.

ToasterBiome (approved):

PR #18613, opened 2023-04-21
Ocular wardens no longer destroy bibles if the target has antimagic

Moltijoe:
I think the idea is that it was supposed to be temporary protection from the wardens
however, if the target already has permanent protection via a nullrod or holy water it will still destroy the bible
this makes holy light chaplains literally useless against clock cultists

YES i'm salty
YES this is an ided

:cl:  
tweak: Ocular wardens no longer destroy bibles if the target has antimagic
/:cl:


Marmio64:
Just carry extra bibles on you

ToasterBiome (approved):

PR #18615, opened 2023-04-21
Makes the possessed sword huge (no more pocket size) + invocation removal

Moltijoe:
Oversight from when i changed it from being a subtype of scythe, i forgot to give it a proper size

:cl:  
tweak: possessed sword is now huge (no more pockets)
bugfix: possessed sword drop ability no longer has an invocation chant
/:cl:


ToasterBiome (approved):
wtf

PR #18616, opened 2023-04-21
The RD's Bag of Holding is a steal objective

Aquizit:

This slides the old BoH steal objective from the ninja list into the regular traitor steal objective list. It also differentiates the RD's bag from the others that can be made later on in the round. Really all I took was the already coded rd version (with the indestructible flag etc.) and added some flair text to it to differentiate it.

image

image

# Wiki Documentation

New name of "prototype bag of holding." Nothing else that I can think of.

# Changelog

:cl:  
rscadd: Adds RD's specific Bag of Holding as steal objective
rscdel: Removes BoH steal from old ninja steal list (in case it's ever reactivated)
tweak: Made RD's BoH its own thing, different from ones made using a shell and anomaly core
/:cl:

### DIFF
--- a/code/game/gamemodes/objective_items.dm
+++ b/code/game/gamemodes/objective_items.dm
@@ -89,6 +89,12 @@
 	difficulty = 5
 	excludefromjob = list("Research Director")
 
+/datum/objective_item/steal/boh
+	name = "the Research Director's prototype bag of holding."
+	targetitem = /obj/item/storage/backpack/holding/rd
+	difficulty = 10
+	excludefromjob = list("Research Director")
+
 /datum/objective_item/steal/documents
 	name = "any set of secret documents of any organization."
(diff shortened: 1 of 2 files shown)


MajManatee:
It already was, then it got removed. please dont bring this dumb obj back

Aquizit:
It was a ninja objective. Then it wasn't when ninja tweaks happened, then BoHs became mostly unique because of the added difficulty to make them now. This is just like most any other Command steal item now, and I think it already counted as Grand Theft which is why I did this PR in the first place.

MajManatee:
To my knowledge its barely grand theft if at all. It was added as a traitor objective for a little while when the bag rework was done, and then removed. The issue is that having it as an objective means you have to be going loud and removes the stealthy playstyles. Theres no way to hide the BoH in a box or a bag, so sec instantly jumps you

Aquizit:
People who don't want this are bothering me more than people who would like this change.

PR #18626, opened 2023-04-23
Fix the vending machine chef's compartment

hedgehog1029:

Fix vending machines not displaying or vending any items in the chef's compartment.

The vending UI now also shows icons for items in the chef's compartment.

image

I have tested these changes.

# Wiki Documentation

Now the stuff about the chef's compartment on the wiki is correct!

# Changelog

:cl:  
bugfix: Fix the vending machine chef's compartment  
rscadd: Add icons to custom items in the vending UI  
/:cl:

### DIFF
--- a/code/modules/vending/_vending.dm
+++ b/code/modules/vending/_vending.dm
@@ -624,7 +624,8 @@ GLOBAL_LIST_EMPTY(vending_products)
 	if(!vending_machine_input[name])
 		vending_machine_input[name] = new /datum/data/vending_custom_product(I)
 
-	vending_machine_input[name].amount++
+	var/datum/data/vending_custom_product/P = vending_machine_input[name]
+	P.amount++
 	loaded_items++
 
 /obj/machinery/vending/exchange_parts(mob/user, obj/item/storage/part_replacer/W)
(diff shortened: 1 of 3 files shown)


code/modules/vending/_vending.dm
```diff
@@ -830,6 +859,65 @@ GLOBAL_LIST_EMPTY(vending_products)
 			R.amount--
 			SSblackbox.record_feedback("nested tally", "vending_machine_usage", 1, list("[type]", "[R.product_path]"))
 			vend_ready = TRUE
+		if("dispense")
```
ynot01: The other action being "vend" and this one being "dispense" are basically synonyms in this context and does not clearly indicate that this one is for custom products and the other is for default products

ynot01 (changes requested):

code/modules/vending/_vending.dm
hedgehog1029: ngl I only used that because it was [already present](https://github.com/yogstation13/Yogstation/blob/master/tgui/packages/tgui/interfaces/Vending.js#L69) on the tgui side (this whole thing was partially implemented):
```js
onClick={() => act('dispense', {
  'item': product.name,
})
```
but i'm happy to update it. while i'm at it, maybe i should refactor out the checking and price code into their own procs, to reduce the amount of duplicated code?

hedgehog1029:
Very minor behaviour change with the refactor: vending machines now check for age requirement even if the mob has `ignore_capitalism` enabled

ynot01 (approved):

PR #18627, opened 2023-04-23
ooc spam prevention filter thing

Bibby0110:
Wouldn't want to be left behind


Bibby0110:
dont think we're going to need this permanently

Maia-J:
> dont think we're going to need this permanently

Just testmerge it for a month until the spammer gets bored

adamsong (approved):
Been tm'd forever, seems to be working just fine

PR #18635, opened 2023-04-23
Blazing oil takes 30% more damage from extinguishers, and 50% extra brute damage

Notamaniac:

Current blazing oil... is overpowered okay I'm just outright saying it.

Extingiushers were nerfed to ineffectiveness, and blazing oil still has it's innate 0.5x brute modifier (on most standard tiles, I think strong blob takes even less).

Lasers are ineffective, and sometimes crew will absolutely refuse to listen to people saying not to shoot the blob, even though it gives it resources.

# Wiki Documentation

Blazing oil takes 50% extra brute damage.

# Changelog
:cl:  
tweak: Blazing oil takes 30% more damage from extinguishers, and 50% extra brute damage.
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
+++ b/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
@@ -3,9 +3,9 @@
 /datum/blobstrain/reagent/blazing_oil
 	name = "Blazing Oil"
 	description = "will do medium burn damage and set targets on fire."
-	effectdesc = "is immune to, and will gain blob points from burn damage. Will also release bursts of flame when burnt, but takes damage from water."
+	effectdesc = "is immune to, and will gain blob points from burn damage, but takes extra brute damage. Will also release bursts of flame when burnt, but takes damage from water."
 	analyzerdescdamage = "Does medium burn damage and sets targets on fire."
-	analyzerdesceffect = "Releases fire when burnt and will gain power when exposed to heat, but takes damage from water and other extinguishing liquids."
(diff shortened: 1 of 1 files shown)


Marmio64:
Starting small with nerfs is almost always better than starting big. 20% more extinguisher damage and 33% more brute damage seems like a more reasonable first step.

Notamaniac:
> Starting small with nerfs is almost always better than starting big. 20% more extinguisher damage and 33% more brute damage seems like a more reasonable first step.

No, sorry to be bull headed on this one but I'm not backing down. Blazing oil is immune to lasers and welders. Brute and extingiushers MUST be a viable weapon by logical necessisty. They cannot be buffed 'by a small amount'. They MUST be useful to counter this complete immunity.

Please note, miner KA's were nerfed stationside by 0.5, dramatically lessening their effective, and ballistic weapons are limited in their ammo (and lack easy reload methods like dragging chargers over, you can't drag an autolathe/protolathe weapon.)

ShadowDeath6:
I haven't had a chance to see how blazing oil blob plays out now after extinguisher nerfs but I remember that one blast from an extinguisher would shred ~6 blobs, which was the reason why I didn't really want to add a brute weakness to begin with. If extinguishers have been nerfed to the point they aren't as much of a hard counter to blazing oil as they used to be, then I fully support this PR. 

I will say that brute damage is fairly easy to come by in the form of spears and whatnot, so you should be aware of damage thresholds since spears one-shotting normal blob tiles is probably too much.

Moltijoe (approved):

PR #18640, opened 2023-04-24
Makes holy water splashes "inject" a percentage into the target

Moltijoe:
This makes it work similarly to unholy water, 
though at a weaker rate of 1/10 instead of the 1/4 of unholy water

a full bluespace beaker of it can convert a cultist if nothing is done
Brings back holy water foam grenades as anti-cult weaponry

:cl:  
rscadd: Splashing with holy water injects some into the target
/:cl:


ChuChu42:
what mix of chems does one need for holy water foam

PR #18641, opened 2023-04-24
Changeling armblade door pry takes 6 seconds instead of 10

Moltijoe:
This makes it on-par with dual mantis blades instead of literally 2x the time of jaws of life
also didn't have a check for welded doors

:cl:  
tweak: Changeling armblade door pry takes 6 seconds instead of 10
bugfix: Changeling armblade can no longer pry open welded doors
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/mutations.dm
+++ b/code/modules/antagonists/changeling/powers/mutations.dm
@@ -205,7 +205,7 @@
 			user.visible_message(span_warning("[user] jams [src] into the airlock and starts prying it open!"), span_warning("We start forcing the airlock open."), //yogs modified description
 			span_italics("You hear a metal screeching sound."))
 			playsound(A, 'sound/machines/airlock_alien_prying.ogg', 100, 1)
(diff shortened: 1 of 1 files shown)


githubuser4141:
I think it makes sense that a purpose built steel tool that's designed to force doors is actually superior at this then a makeshift bone that -technically- can be used. Realistically if a armblade can be used, that means -any- item could be used dependant on the user's strength and if they could get it around the airlock creases.

Moltijoe:
> I think it makes sense that a purpose built steel tool that's designed to force doors is actually superior at this then a makeshift bone that -technically- can be used. Realistically if a armblade can be used, that means -any- item could be used dependant on the user's strength and if they could get it around the airlock creases.

consider this however, one is an easily printable tool that anyone can get
the other is a specialized antag weapon
also, most people don't even use jaws of life, they just figure out which wires do what and hack it in 2 seconds

ToasterBiome (changes requested):
too short make it 6

ToasterBiome (approved):

PR #18658, opened 2023-04-25
Nitroglycerin no longer explodes inside of bodies

SapphicOverload:

Nitroglycerin will no longer explode when inside a body.

# Wiki Documentation

Might be worth mentioning on the guide to chemistry.

# Changelog

:cl:  
tweak: nitroglycerin no longer explodes inside of bodies
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
+++ b/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
@@ -33,6 +33,7 @@
 	results = list(/datum/reagent/nitroglycerin = 2)
 	required_reagents = list(/datum/reagent/glycerol = 1, /datum/reagent/toxin/acid/fluacid = 1, /datum/reagent/toxin/acid = 1)
 	strengthdiv = 2
+	mob_react = FALSE
 
 /datum/chemical_reaction/reagent_explosion/nitroglycerin/on_reaction(datum/reagents/holder, created_volume)
 	if(holder.has_reagent(/datum/reagent/stabilizing_agent))
@@ -46,6 +47,7 @@
 	required_reagents = list(/datum/reagent/nitroglycerin = 1)
 	required_temp = 474
 	strengthdiv = 2
+	mob_react = FALSE
 
 
 /datum/chemical_reaction/reagent_explosion/potassium_explosion


milk-is-cool:
why

iamthesilly:
> why

because being able to instantly kill people with no counter is bad? this is already the case with meth explosions for the same reason

wonderinghost:
its steady progress to stop ark's nettles

milk-is-cool:
> > why
> 
> because being able to instantly kill people with no counter is bad? this is already the case with meth explosions for the same reason

the problem i have with it is that it still doesnt solve that issue, it was still a one shot kill without the nitro and still will be after this. its too indirect to fix the one shot kill problem because there are still many other ways of doing it. the real problem is with the method of reagent transfer. just like with chem and removing the syringe gun, removing the ability to inject chems into people at all is what would be the real solution. its not the solution people will want but its not like everyone was happy about the syringe gun being removed either.

iamthesilly:
> the problem i have with it is that it still doesnt solve that issue, it was still a one shot kill without the nitro and still will be after this. its too indirect to fix the one shot kill problem because there are still many other ways of doing it. the real problem is with the method of reagent transfer. just like with chem and removing the syringe gun, removing the ability to inject chems into people at all is what would be the real solution. its not the solution people will want but its not like everyone was happy about the syringe gun being removed either.

we can do more than one thing

milk-is-cool:
> > the problem i have with it is that it still doesnt solve that issue, it was still a one shot kill without the nitro and still will be after this. its too indirect to fix the one shot kill problem because there are still many other ways of doing it. the real problem is with the method of reagent transfer. just like with chem and removing the syringe gun, removing the ability to inject chems into people at all is what would be the real solution. its not the solution people will want but its not like everyone was happy about the syringe gun being removed either.
> 
> we can do more than one thing

i am also bias towards chem so i am just generally against nerfs. for chemistry this also prevents things like monkey bombs which i feel deserve to be possible because of how much more work you need to put into those. also black powder still reacts in the body so that is still an alternative to do the same thing, just not near as powerful. 

in short: i dont like it because its also an indirect chemistry nerf on something that is already very difficult to obtain.

wonderinghost:
well what about botany though

milk-is-cool:
> well what about botany though

like i said before its not the chems that are the issue, its the method of reagent transfer. aka the hypodermic needles trait.

wonderinghost:
> > well what about botany though
> 
> like i said before its not the chems that are the issue, its the method of reagent transfer. aka the hypodermic needles trait.

hypodermic needles is too powerful. increasing potency in a plant just makes it have more chems in it and upon getting hit once you would be in crit while also burning to death by sulfuric acid.

milk-is-cool:
> > > well what about botany though
> > 
> > 
> > like i said before its not the chems that are the issue, its the method of reagent transfer. aka the hypodermic needles trait.
> 
> hypodermic needles is too powerful. increasing potency in a plant just makes it have more chems in it and upon getting hit once you would be in crit while also burning to death by sulfuric acid.

yes and thats just one example of many which is why the hypodermic needles trait is the problem and not nitro

wonderinghost:
ide be down if the traits got debuffs. like hypodermic needles having a regent leak or the chances of them injecting while thrown is far less

wonderinghost:
however the problem is that botany can just mass produce these. And there isnt a good way of destroying them without a conveyor that warps them to lavaland into a pool of lava

milk-is-cool:
> however the problem is that botany can just mass produce these. And there isnt a good way of destroying them without a conveyor that warps them to lavaland into a pool of lava

just throwing a death nettle on a stack of plants will cover them all in acid and burn them into goo

wonderinghost:
> > however the problem is that botany can just mass produce these. And there isnt a good way of destroying them without a conveyor that warps them to lavaland into a pool of lava
> 
> just throwing a death nettle on a stack of plants will cover them all in acid and burn them into goo

yeah but then there is also the plant and the seeds that you also have to destroy. and if cargo isnt brain dead and doesnt activate their conveyers then botany can just take the trash and start the cycle anew

Moltijoe (approved):

PR #18659, opened 2023-04-25
New donator item for Cark

Moltijoe:
New cark donator item
it's a toggleable "hoodie"

item, sleeves down, sleeves up
image

:cl:  @cuackles @Moltijoe 
rscadd: Adds new donator item for Cark
imageadd: adds sprites for Cark's donator item
/:cl:


ToasterBiome:
what country is this

CrazyDrakkon:
> 

noone tell them

Moltijoe:
> what country is this

no clue, very patriotic though

ToasterBiome (approved):
still don't know what flag this is

adamsong:
So this is making cark's existing item (the bedsheet) no longer obtainable, and adding a new item locked to cark, in its place?

Is the intent for someone in future to be able to use this for other purposes, or if someone opens a pr that uses it, will it get shot down because "that used to be cark's item" because if the latter is the case, I would prefer the item removed entirely.

Moltijoe:
> So this is making cark's existing item (the bedsheet) no longer obtainable, and adding a new item locked to cark, in its place?
> 
> Is the intent for someone in future to be able to use this for other purposes, or if someone opens a pr that uses it, will it get shot down because "that used to be cark's item" because if the latter is the case, I would prefer the item removed entirely.

Nah, this is a 2nd donator item, it gives them a box
Were the bedsheet to be removed as a donator item id just incorporate it into the game

adamsong:
Under the new policy, each person is only allowed one donator item

PR #18674, opened 2023-04-27
[admin] Disables midround OOC

Bibby0110:
idk haha


DylonBR:
why

neeshacark:
there is zero need to do this, if anything make it only be disabled when admins aren't present but even that is dumb

PR #18691, opened 2023-04-29
Beepsky does stamina damage instead of instant stun

SapphicOverload:

Beepsky now does 54 stamina damage instead of 10 seconds of instant paralysis

# Changelog

:cl:  
tweak: beepsky does 54 stamina damage to you instead of 10 seconds of instant paralysis
tweak: increased beepsky's health to compensate and removes its damage coefficients
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/secbot.dm
+++ b/code/modules/mob/living/simple_animal/bot/secbot.dm
@@ -5,9 +5,8 @@
 	icon_state = "secbot"
 	density = FALSE
 	anchored = FALSE
-	health = 40
-	maxHealth = 40
-	damage_coeff = list(BRUTE = 0.5, BURN = 0.7, TOX = 0, CLONE = 0, STAMINA = 0, OXY = 0)
+	health = 75
+	maxHealth = 75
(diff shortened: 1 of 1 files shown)


TheBombMakerJL:
Buff it’s HP

solwashere:
Dislike the decision to buff the HP and damage so much, it might as well be a hardstun again cause you're not gonna be able to get out of that once it hits you

RG4ORDR:
If you're getting robusted by Beepsky that's your own fault. Bad PR.

iamthesilly:
Alright, now it downs you in 3 hits if you have a bit of armor and 2 hits if you don't, and tweaked its health and damage values a bit more. Also makes it actually respect armor and block chance as it should. Damage is lower so it should be possible to escape after being hit once without armor, and you'll be able to kill it with a proper melee weapon (eswords, vxtvul hammer, sec vibro blade, fire axe, etc) if you do have armor.

neeshacark:
a legend's end

JohnFulpWillard:
Bad day for good sec

Moltijoe (approved):

PR #18692, opened 2023-04-29
Adds a disabler arm implant to alien research

Moltijoe:
Moja was wanting "the disabler arm implant" to be made available instead of admin only
well that arm implant is actually a taser/disabler hybrid, so i just made a new arm implant

can only be printed in security once researched

:cl:  
rscadd: Adds a disabler arm implant to alien research
/:cl:


Moltijoe:
balance council doesn't like un-disarmable sec weapons (why am i telling myself this?)

PR #18695, opened 2023-04-29
adds upgrade tiers to hand teleporters

TheRyeGuyWhoWillNowDie:
adds upgrade tiers to hand teleporters. each upgrade tier decreases the dispel time by 1 second and adds 1 extra max portal at a time. it's upgraded like a laser pointer and uses micro manipulators

# Wiki Documentation

T1 manipulator, 1 portal, 4 seconds dispel time
T2 manipulator, 2 portals, 3 seconds dispel time
T3 manipulator, 3 portals, 2 second dispel time
T4 manipulator, 4 portals, 1 second dispel

# Changelog

:cl:  
tweak: hand teleporters can be upgraded
/:cl:

### DIFF
--- a/code/game/objects/items/teleportation.dm
+++ b/code/game/objects/items/teleportation.dm
@@ -173,7 +173,7 @@
 
 /obj/item/hand_tele/proc/try_dispel_portal(atom/target, mob/user)
 	if(is_parent_of_portal(target))
-		var/dispel_time = 4 - manipulator.rating
+		var/dispel_time = 5 - manipulator.rating
 		if(dispel_time == 0)
 			qdel(target)
 			to_chat(user, span_notice("You dispel [target] with \the [src]!"))
--- a/code/game/objects/items/teleportation.dm
+++ b/code/game/objects/items/teleportation.dm
@@ -134,12 +134,37 @@
 	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | ACID_PROOF
 	cryo_preserve = TRUE
 	var/list/active_portal_pairs
-	var/max_portal_pairs = 3
 	var/atmos_link_override
+	var/obj/item/stock_parts/manipulator/manipulator
 
 /obj/item/hand_tele/Initialize()
 	. = ..()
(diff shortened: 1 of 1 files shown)


solwashere:
interesting idea but I think that t4 should still have a dispel time to prevent abuse.

TheRyeGuyWhoWillNowDie:
> interesting idea but I think that t4 should still have a dispel time to prevent abuse.

the dispel time was added because people were cheesing nukeops. you are not getting t4 research and the rare materials required for t4 parts in under 20 minutes on warops, so it's not an issue

solwashere:
> > interesting idea but I think that t4 should still have a dispel time to prevent abuse.
> 
> the dispel time was added because people were cheesing nukeops. you are not getting t4 research and the rare materials required for t4 parts in under 20 minutes on warops, so it's not an issue

What about traitors grabbing the hand tele late into the round and using it to infinitely evade security?

TheRyeGuyWhoWillNowDie:
> What about traitors grabbing the hand tele late into the round and using it to infinitely evade security?

identical to crushing bluespace crystal. if they're using beacons just deconstruct teleporters? hand teleporters were never an issue outside of nukeops before the dispel timer was added

solwashere:
> > What about traitors grabbing the hand tele late into the round and using it to infinitely evade security?
> 
> identical to crushing bluespace crystal. if they're using beacons just deconstruct teleporters? hand teleporters were never an issue outside of nukeops before the dispel timer was added

That is fair, however repeatedly using bluespace crystals causes nausea stacks does it not?

TheRyeGuyWhoWillNowDie:
> That is fair, however repeatedly using bluespace crystals causes nausea stacks does it not?

there are only 2 hand teleporters. them having some kind of advantage over the thing literally anyone can do seems fair

Moltijoe:
Balance council has requested that the dispel times end up with T4 being 1 second
please note as per the balance council's request, if no response is made within 24 hours, the PR will be closed

Moltijoe (approved):
well done gus, you managed to turn the vote around with a simple change

PR #18697, opened 2023-04-29
kills special .38, adds rubber .38

Skrem7:

The existence of exotic .38 rounds encourages detective validhunting because they're fun goofy toys that are quite frankly obnoxious (looking at you Iceblax) that, while they almost never see the light of day, I hate the fact that they exist and occupy the "oh these ammos are cool" for a gun that shouldn't have them, since the detective shouldn't be using their gun, period. So I scrubbed exotic .38 entirely. Techshells have been scooted back to exotic ammo so the node actually has something ~~because I don't have a .38 handgun to introduce yet with printable magazines like the WT~~

Adds .38 rubber to be loaded into the detective's revolver and shoulder holster by default roundstart as well as making it a much more accessible speed loader because printing off single bullets was a nightmare and not really exhilarating and detective valid hunting has been clamped down on fairly well, as far as I've observed. So you can print full lethal speed loaders if you hack the lathe now. The armory lathe can still print them, too. .38 rubber is less lethal than old .38 (15/35) and generally sucks more, but it can be printed at autolathes and the security lathe as normal, because, less-than-lethal. Still shouldn't be using it for validhunting + it sucks against armor (no longer is it a better disabler).

Also makes .38 do less damage (21 from 25) at the cost of having less bad negative AP (-30 from -40) because I hate binary performance.

Minor note, the cargo bounty of .38 speedloaders should actually be semi-possible now with the autolathe change, you're welcome

# Wiki Documentation

Detective page and Guide to Combat should have the new ammo reflected (mainly the old stuff killed)
Designs pages (autolathe and protolathe) should be rectified with the new stuff
Research should include how exotic ammo only gives you techshells now, moved from experimental ammo

# Changelog

:cl:  
rscadd: Adds .38 rubber and makes it the default detective ammunition.
rscdel: Removes .38 Hotshot, Iceblax, and Gutterpunch from the game wholly
tweak: Messed around with detective mail options (lazily
tweak: .38 lethal does less damage now, but less bad AP
tweak: Hacked autolathes can print full .38 speed loaders again.
tweak: All rubber rounds are now blunt
imageadd: Caldwell now faces the right way
spellcheck: Fixed some crate descriptions
/:cl:


TheRyeGuyWhoWillNowDie:
here and there and back again

Moltijoe:
Trims the number of words in each changelog line please

cuackles (approved):

ToasterBiome (approved):

PR #18698, opened 2023-04-30
lessened ethereal colour washout

Moltijoe:
With the new lighting system power is a lot more impactful
This means an ethereal with full health having a light power of 2 kinda just overwrites the colours of everything nearby
This colour change makes it nearly impossible to wear cool clothes or have distinct body colours as the light will just wash everything out

Ethereal light range is untouched

:cl:  
tweak: Ethereal light power is lowered to reduce colour problems
/:cl:


Marmio64:
is it really necessary to gut the whole thing? Seeing their color dim as they died was pretty cool, why not just lessen how much of an effect health has or something.

Moltijoe:
> is it really necessary to gut the whole thing? Seeing their color dim as they died was pretty cool, why not just lessen how much of an effect health has or something.

fair, it can go from 0-1 instead of static 1
either way, range still scaled

ToasterBiome (approved):
these are really annoying

PR #18699, opened 2023-04-30
Buffs explosive fist

Moltijoe:
It's a species martial art, it deserves to be cooler

- detonate has 1 size larger fire explosion
- detonate also did damage only to yourself instead of the enemy for some reason
- immolate creates hotspots within the damage radius
- immolate ignites mobs within the damage radius, rather than in a radius 1 tile smaller for some reason
- harm act create a hotspot on target and ignites them
- lifeforce grab starts at 4 succ damage instead of 2 and gets 50% more damage per channel stacking multiplicatively
- lifeforce grab no longer ends when the target goes unconscious
- removes damage slowdown because of self damage

explosion sound effects now pick between a couple different sound effects

reorders code to make it easier to read

:cl:  
tweak: Buffs plasmeme martial art
tweak: detonate explosion has a 1 tile larger fire explosion
bugfix: detonate no longer does damage to yourself instead of the target
tweak: Immolate creates hotspots and ignites within the damage radius
tweak: Harm act creates a hotspot on the target and ignites them
tweak: Lifeforce grab starts with higher base drain and scales exponentially per tick
tweak: plasmeme martial art also gives immunity to damage slowdown
/:cl:


KingSuperBOB:
Would be interesting if explosive fist gave no breathe (since most of the times your internals mask will burn off and youll choke to death)

Moltijoe:
> Would be interesting if explosive fist gave no breathe (since most of the times your internals mask will burn off and youll choke to death)

I made regular gas masks fireproof a while ago

ynot01 (approved):

ynot01 (changes requested):
Merge conflicts

ynot01 (approved):

PR #18649, opened 2023-04-24
Removes bluespace body bag cheese

ynot01:
# Document the changes in your pull request

# Changelog

:cl:  
tweak: It now takes 2.5 seconds to fold a bluespace body bag when a mob is inside of it
/:cl:

PR #18701, opened 2023-04-30
Requires people to be lying down to enter a bodybag + closet flags

Moltijoe:
it's a bodybag
bodybags hold bodies
bodies are lying down
why can bodybags just kinda encompass people that are standing up

changes closets to use a flag system for checking whether or not things can go into it (removes 2 variables from all closets)

:cl:  
tweak: Requires people to be lying down to enter a bodybag (can be closed with someone standing on it)
/:cl:


solwashere:
Good change

ToasterBiome:
Would prefer this over https://github.com/yogstation13/Yogstation/pull/18649 (instead of both)

Moltijoe:
> Would prefer this over #18649 (instead of both)

personally, same, which is sorta why i made this

ToasterBiome (approved):

PR #18711, opened 2023-04-30
removes craftable pneumatic cannons

TheRyeGuyWhoWillNowDie:
Pneumatic cannon is never used except for making instakill shotguns. Nerfing it would make it useless, so it is being removed instead.

# Changelog

:cl:  
rscdel: you can't make pneumatic cannons anymore
/:cl:

### DIFF
--- a/code/datums/components/crafting/weapons.dm
+++ b/code/datums/components/crafting/weapons.dm
@@ -327,17 +327,6 @@
 	category = CAT_WEAPONRY
 	subcategory = CAT_WEAPON
 
-/datum/crafting_recipe/improvised_pneumatic_cannon //Pretty easy to obtain but
-	name = "Pneumatic Cannon"
-	result = /obj/item/pneumatic_cannon/ghetto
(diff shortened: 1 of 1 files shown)


DylonBR:
maybe make it syndi craftable only?

TheRyeGuyWhoWillNowDie:
> maybe make it syndi craftable only?

"it's in 4 gamemodes only" != justification for literal instant kill

DylonBR:
then how come you can still buy a dart pistol as a tot chemist?

TheRyeGuyWhoWillNowDie:
> then how come you can still buy a dart pistol as a tot chemist?

because chemicals have some delay before you die, whereas you stuff 50 cult daggers or edaggers or exploding plants into a pneumatic cannon and then left click and whoever you hit is now dead

RG4ORDR:
Swap it to makeshift crafting, that makes logical sense because generally the crew doesn't need to make this. It's not balanced out by being ass like the gauss rifle.

Moltijoe:
>improve don't remove

PR #18712, opened 2023-04-30
removes the 100 everything stat 10 minute rollout from botany

TheRyeGuyWhoWillNowDie:
botany is ridiculous and braindead easy, you can get max stats everything in like 20 minutes, which is braindead op when combined with stuff like pneumatic cannons and explosive plants

# Wiki Documentation

stats can only be interchanged between the same plant species. (genes like slippery skin are unaffected) 
eg, you can make potency 100 bananas and transfer the potency 100 to bluespace bananas, or the potency 100 from bluespace bananas to regular bananas. you cannot transfer the potency 100 from the bananas to anything non-banana.

# Changelog

:cl:  
tweak: plant stats in botany can only be inserted into plants of the same species
/:cl:

### DIFF
--- a/code/modules/hydroponics/gene_modder.dm
+++ b/code/modules/hydroponics/gene_modder.dm
@@ -124,6 +124,12 @@
 			target = null
 			interact(user)
 			return
+		if((operation == "replace" && !((istype(disk.seed, seed)) || (istype(seed, disk.seed)))))
+			say("WARNING! PLANT STATS CAN ONLY BE INSERTED INTO PLANTS OF A SIMILAR TYPE!")
+			operation = ""
+			target = null
+			interact(user)
+			return
 
 		dat += "<div class='line'><h3>Confirm Operation</h3></div>"
 		dat += "<div class='statusDisplay'>Are you sure you want to [operation] "
@@ -320,7 +326,11 @@
 				if("extract")
 					if(disk && !disk.read_only)
 						disk.gene = G
+						qdel(disk.seed)
+						disk.seed = null
 						if(istype(G, /datum/plant_gene/core))
+							disk.seed = seed
+							seed.forceMove(disk)
 							var/datum/plant_gene/core/gene = G
 							if(istype(G, /datum/plant_gene/core/potency))
 								gene.value = min(gene.value, max_potency)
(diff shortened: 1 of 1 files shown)


GregTheEgg42069:
oh god.

ToasterBiome:
we should instead look at the reason for extremely strong stats in the first 20 minutes in the first place

TheRyeGuyWhoWillNowDie:
> we should instead look at the reason for extremely strong stats in the first 20 minutes in the first place

Because every stat has at least one seed in the seed vendor with 50% or higher in that category that you can then splice into every other plant

RG4ORDR:
> we should instead look at the reason for extremely strong stats in the first 20 minutes in the first place

The way botany is works is that you can just for Saltpetre from chem, you need 4 bottles if not less to get 100 potency off it. Which you apply to grass, get more grass seeds, plant 2-3 trays of grass and you're basically fine on biogen stuff so you can shit out mutagen in which if you start apply 1 unit of mutagen to plants you'll random roll their stats. That and once you get gaia it goes in the grass and you do this all in the span of less than 10 minutes. 

That and the same player keeps doing the same exact strategy in botany over and over and over as antag.

ToasterBiome:
> > we should instead look at the reason for extremely strong stats in the first 20 minutes in the first place
> 
> Because every stat has at least one seed in the seed vendor with 50% or higher in that category that you can then splice into every other plant

So what about moving those downs in stats?

> > we should instead look at the reason for extremely strong stats in the first 20 minutes in the first place
> 
> The way botany is works is that you can just for Saltpetre from chem, you need 4 bottles if not less to get 100 potency off it. Which you apply to grass, get more grass seeds, plant 2-3 trays of grass and you're basically fine on biogen stuff so you can shit out mutagen in which if you start apply 1 unit of mutagen to plants you'll random roll their stats. That and once you get gaia it goes in the grass and you do this all in the span of less than 10 minutes.
> 
> That and the same player keeps doing the same exact strategy in botany over and over and over as antag.

So what if we reworked the recipe (or chamged the chem)? or made it so no chem gave potency on plants?

RG4ORDR:
> > > we should instead look at the reason for extremely strong stats in the first 20 minutes in the first place
> > 
> > 
> > Because every stat has at least one seed in the seed vendor with 50% or higher in that category that you can then splice into every other plant
> 
> So what about moving those downs in stats?
> 
> > > we should instead look at the reason for extremely strong stats in the first 20 minutes in the first place
> > 
> > 
> > The way botany is works is that you can just for Saltpetre from chem, you need 4 bottles if not less to get 100 potency off it. Which you apply to grass, get more grass seeds, plant 2-3 trays of grass and you're basically fine on biogen stuff so you can shit out mutagen in which if you start apply 1 unit of mutagen to plants you'll random roll their stats. That and once you get gaia it goes in the grass and you do this all in the span of less than 10 minutes.
> > That and the same player keeps doing the same exact strategy in botany over and over and over as antag.
> 
> So what if we reworked the recipe (or chamged the chem)? or made it so no chem gave potency on plants?

So the funny thing is that outside of a DNA machine/mutagen the only way to boost yield higher is ammonia. The thing with ammonia is that you needed to actually dump 100 units into it constantly. Not 50 unit beakers, not 300 from a bluespace, but you had to exactly dump 100 units of ammonia into the dirt/planter to increase the yield. This is an idea where you can't boost planet potency w/o chemistry but reduce the overally amount of potency that plants get from saltpeter.

Another idea is removing chemical reactions from the plants themselves to prevent people from doing chemistry with them. That's the main issue with the plants - most people either make a healing plant, or in Rune's case he's using the fact he can do nitro due the fact nitro only needs deathnettles and coil to produce the reaction. Or change the recipie for nitro to require chemical heater and you probably fix the issue that comes with it. Botany will still retain their rather strong niche of funny OP plants but you're not shooting 1HK bombs at people.

Moltijoe:
after 24h
solve the problem in a way that doesn't make botany terrible

PR #18718, opened 2023-05-01
Jellypeople squoosh easier

Mqiib:
Jellypeople are secretly an incredibly gamer race with the downside of being cold hurts you a lot except you can just carry coffee to completely nullify that. Now they have an ACTUAL downside.

# Wiki Documentation

Jellypeople now are more prone to being injured by basically everything because they're held together by ???? and not bone and muscle like everyone else.

# Changelog

:cl:  

tweak: Jellypeople now have -10 armor by default
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/jellypeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/jellypeople.dm
@@ -17,6 +17,7 @@
 	coldmod = 6   // = 3x cold damage
 	heatmod = 0.5 // = 1/4x heat damage
 	burnmod = 1 // = regular burn damage unlike other slimes
+	armor = -10	//Squishy
 	payday_modifier = 0.6 //literally a pile of toxic ooze walking around, definitely a health hazard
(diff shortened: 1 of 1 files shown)


DylonBR:
jelly racism

Marmio64:
should probably also fix the comments regarding their cold and heat damages as they were made on the basis of other slime races, which take less burn damage. Also maybe make their cold damage be reasonable like 3x rather than 6x, so if you happen to not be carrying coffee its not an instant death sentence. Or not to keep this atomized and what not.

Mqiib:
> should probably also fix the comments regarding their cold and heat damages as they were made on the basis of other slime races, which take less burn damage. Also maybe make their cold damage be reasonable like 3x rather than 6x, so if you happen to not be carrying coffee its not an instant death sentence. Or not to keep this atomized and what not.

Even using the temperature gun on them doesn't actually do all that much damage, especially since it's delayed damage and can effectively be countered by easily-obtainable chems. They die to frost oil pretty quick, sure, but otherwise being cold doesn't happen particularly often. I can probably fix the comments though.

MajManatee:
I love racism

Moltijoe (approved):

PR #18764, opened 2023-05-02
Fixes loophole for Bibby

JamieD1:


adamsong:
"I don't like being bound by my own rules, so I won't be"

PR #18775, opened 2023-05-03
Makes "disabled" limbs more likely to get wounded

Moltijoe:
Since apparently #16594 is "fixing" something that isn't a bug
This just makes wounds more likely when a limb should otherwise be disabled

:cl:  
tweak: Heavily damaged limbs are more likely to get wounds
/:cl:


ynot01 (approved):
code is shrimple

PR #18793, opened 2023-05-05
fixes meta chem shutter button

solwashere:
requested by @milk-is-cool, fixes a chem shutter button (it is no longer on the outside)
~~please help they keep requesting things in VC~~

# Spriting

# Wiki Documentation

# Changelog

:cl:  

mapping: fixes chem shutter button on meta

/:cl:


solwashere:
Don't merge, Ezhan pointed out an issue that I'll fix when I get home

Ethan4303 (changes requested):
you have put the button onto the wall directly so now people on both sides of the wall can hit the button,
image
its gotta be in the room, then pixel shifted onto the wall like the intercomm

Ethan4303 (approved):

PR #18800, opened 2023-05-06
Fixes mapmerge2 not working under python 3.11

alexkar598:
im not filling out this essay of a template, close the pr if you care

makes mapmerge compatible with python 3.11 without having to build from source

there's a pygit2 1.12.0 but it has more breakage potential than 1.11.1 so 1.11.1 it is


alexkar598:
@adamsong

alexkar598:
Ideal fix would probably be tgstation/tgstation#55658 but its way more work

Bibby0110 (approved):

PR #18804, opened 2023-05-06
semicircle of cogs

azzzertyy:
Clockies have a hard time defending ever since we became circle of cogs instead of square of cogs.
This changes that by returning coggies to their eternal roots, giving them access to a single rectangular wall within their grand establishment, turning it into a semicircle of cogs.

This also changes some of the interior sections of the city, making it so there is a second camera room and merges the southwest room with the northwest room.

This also changes the glass in the interior section near the ark to be normal walls, this was too easy to just shoot through and I dont know why anyone thought it was good.

# Wiki Documentation
Image of reebee will need changing

image

# Changelog
:cl:  
mapping: Changes city of cogs to be a semicircle of cogs, giving clockies a buff.
/:cl:


Moltijoe:
Can it be a semicircle in the other direction?

azzzertyy:
> Can it be a semicircle in the other direction?

why

Moltijoe:
> why

No real reason aside from it feeling weird to attack downwards ty 👍

Aquizit:
Yeah looking up and not down into your UI just feels better.

azzzertyy:
ok but reframe it so that it feels weird to defend upwards and were in the same boat

Aquizit:
A fair point. Can make it face east or west then and there's more parity.

Moltijoe:
> ok but reframe it so that it feels weird to defend upwards and were in the same boat

Id argue you dont need to see as much to defend since you arent likely moving towards something

RG4ORDR:
Holy fuck finally.
Clockies now will be able to make defendable rebees

Hislyna:
Does this include a spawnpoint for deathsquaddies?

neeshacark:
i also agree in flipping the circle,  but maybe flip it like so instead. best of both worlds, nobody has to fight into their hud
236610210-bd4d8db0-6601-478f-abdf-d591cccdde97

Aquizit (changes requested):
Turn it sideways so it's facing east or west, I think is the consensus here.

Aquizit:
Should also be made bigger. Will get very cramped in there during high pop hours.

00ze-cyclone:
> Turn it sideways so it's facing east or west, I think is the consensus here.

Make two map and it randomly either face west or east

azzzertyy:
> Should also be made bigger. Will get very cramped in there during high pop hours.

the point is to buff coggies, making it bigger reverts the entire point of the pr

Aquizit:
> 

There is a difference between an entire circle and 4 angles of attack, and adding like 2 more tiles of width to the surface area of the semicircle.

azzzertyy:
> > 
> 
> There is a difference between an entire circle and 4 angles of attack, and adding like 2 more tiles of width to the surface area of the semicircle.

its not just about reducing angles of attack, its reducing the amount of possible walls they have to build to defend themselves, making it bigger increases this. I don't think it'l get crowded, considering its the size of like 4 box hallways stacked on top of eachother

Aquizit:
Alright, well let's just start by flipping it sideways, then we'll testmerge it and see how it goes.

azzzertyy:
Fixed.
image

(joke, see updated pr)

Aquizit:
~~I would prefer these to be identical (I don't care which design although I think the first one just looks cleaner). But we can testmerge with this.~~

Omg shut up I missed the joke part fuck me sideways

Aquizit:
Ok so from the one round I saw -- it's missing the 3 cogscarab spawns at the beginning of the round. 
Apparently there are places along the barrier that clockies aren't supposed to be able to go through that the eminence can slip through.
Those are the two things that I was made aware of.

azzzertyy:
> Ok so from the one round I saw -- it's missing the 3 cogscarab spawns at the beginning of the round. Apparently there are places along the barrier that clockies aren't supposed to be able to go through that the eminence can slip through. Those are the two things that I was made aware of.

I have no idea what could be causing the eminence slip through.
The tiles are identical to all the others, so theres nothing special.

Aquizit:
There is now a new issue where this PR has 7 files changed. It should only be the map file being changed. Need to revert the stuff to the other 6.

azzzertyy:
> There is now a new issue where this PR has 7 files changed. It should only be the map file being changed. Need to revert the stuff to the other 6.

if you look at the changes they are literally nothing, literally just a git error and i have no idea why it exists, but its literally zero changes

Aquizit:
> if you look at the changes they are literally nothing, literally just a git error and i have no idea why it exists, but its literally zero changes

Not a me thing, this demand comes from on high.

azzzertyy:
closing, my branch is royally fucked, im going to make a new pr because holy shit this is stupid

PR #18809, opened 2023-05-06
Adds gravitational lensing to the Ark

azzzertyy:

Whenever the Ark is shot now, it will give a gravitational lensing affect based entierly on its damage, (0.1 + damage_amount/80)
It will then shrink back down as it recovers.
It also adds a lens effect to the ark once its destroyed, as well as when its completed. Currently this is unfinished, as I have yet to add a check for damaging it, meaning the lensing will get reset once you shoot it again.
Makes shooting the ark way more interesting, and has the potential to cause a little bit more chaos in the Ark room.

https://user-images.githubusercontent.com/42524344/236644511-439b1e12-4331-4181-a686-18cb10258966.mp4

# Changelog
:cl:  
rscadd: Added Gravitation Lensing to the Ark
/:cl:

### DIFF
--- a/code/modules/antagonists/clockcult/clock_structures/ark_of_the_clockwork_justicar.dm
+++ b/code/modules/antagonists/clockcult/clock_structures/ark_of_the_clockwork_justicar.dm
@@ -49,13 +49,9 @@
 	START_PROCESSING(SSprocessing, src)
 	if (!grav_lens)
 		grav_lens = new(src)
-		grav_lens.transform = matrix().Scale(0.5)
+		grav_lens.transform = matrix().Scale(0.15)
 		grav_lens.pixel_x = -240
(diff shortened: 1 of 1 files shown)


Moltijoe:
looks far too harsh, try messing with the alpha of the effect

also the final explosion should expand for longer

cuackles (changes requested):
this is kinda visual so i asume i have a say in this as the art director but i think it would look better if when damaged it did a similar effect to the when dead , so like every hit make a wave or make it so just the dying one

azzzertyy:
Will take a further look into this one a bit later.

ynot01:
> it would look better if when damaged it did a similar effect to the when dead , so like every hit make a wave or make it so just the dying one

agreed

PR #18816, opened 2023-05-07
IPC martial art nerfs

Moltijoe:
removes some unfun insta-win mechanics
as well as reducing passive mobility

:cl:  
rscadd: Adds a verb for ipc martial to fix click intercept
tweak: both blood burst and gun hand comboes require 1 less intent to pull off
tweak: blood burst heals for half as much on execute
tweak: gun hand now uses ironfeather shots as ammo, making it deal close to half damage
tweak: pocket revolver ammo has 25 less AP and 15 less wound bonus (same as regular 357 now)
tweak: reduces strength of pocket revolver speed boost by more than half, but lasts 4 seconds longer
tweak: Ultra Violence no longer gives passive movespeed
tweak: passive blood healing heals for 30 adapatively instead of 20 brute 20 burn
tweak: removes the stun from the emp tesla counter
/:cl:


Therandomhoboo:
So, still doing an execute will reward you with a good heal (70 if I'm correct), I can't speak for other martial arts cause I dont know if they do heals, but I still think it is really high. I think they should either get one or the other but not both for healing considering executing people will always splatter blood on yourself.

Moltijoe:
> So, still doing an execute will reward you with a good heal (70 if I'm correct), I can't speak for other martial arts cause I dont know if they do heals, but I still think it is really high. I think they should either get one or the other but not both for healing considering executing people will always splatter blood on yourself.

The execute isnt quite that common, is it? If it is i can always reduce it further, but a loss of 50 healing is pretty substantial

Therandomhoboo:
> > So, still doing an execute will reward you with a good heal (70 if I'm correct), I can't speak for other martial arts cause I dont know if they do heals, but I still think it is really high. I think they should either get one or the other but not both for healing considering executing people will always splatter blood on yourself.
> 
> The execute isnt quite that common, is it? If it is i can always reduce it further, but a loss of 50 healing is pretty substantial

Honestly, might of been that one round. I would have to see more rounds with the martial arts to tell you how often it happens but if you think it's not common then we'll just see how it goes really. All in all, another PR can be done if its still strong.

Moltijoe:
Completely forgout about the ricoshot, i'll just readd the special bullet and nerf it instead

Marmio64:
It still seems pretty powerful, maybe even more so now since the 2 tc discount now opens up more stuff, including the all providing emag. Should receive additional nerfs if it's gonna now allow you a lot more versatility in purchases.

wonderinghost:
> So, still doing an execute will reward you with a good heal (70 if I'm correct), I can't speak for other martial arts cause I dont know if they do heals, but I still think it is really high. I think they should either get one or the other but not both for healing considering executing people will always splatter blood on yourself.

ide say otherwise. mainly due to lasers. 1.5 times burn is still really rough. along with being conductive. you can easily crit ipc with lasers in a few shots (without martial) however that gets bumped up to 4 since martial art ipc dont go into crit

Moltijoe:
This martial art also has literally 0 CC to reduce damage taken, in contrast to most other martial arts which either are full CC or are relatively quiet to avoid conflicts in the first place

code/datums/martial/ultra_violence.dm
```diff
@@ -60,13 +61,18 @@
 	check_streak(A,D)
 	return FALSE
 
-/datum/martial_art/ultra_violence/proc/speed_boost(mob/living/carbon/human/A, duration, tag)
-	A.add_movespeed_modifier(tag, update=TRUE, priority=101, multiplicative_slowdown = -0.5, blacklisted_movetypes=(FLOATING))
-	addtimer(CALLBACK(src, .proc/remove_boost, A, tag), duration, TIMER_UNIQUE|TIMER_OVERRIDE)
+/datum/martial_art/ultra_violence/proc/speed_boost(mob/living/carbon/human/A, strength, tag)
+	A.add_movespeed_modifier(tag, update=TRUE, priority=101, multiplicative_slowdown = strength, blacklisted_movetypes=(FLOATING))
+	addtimer(CALLBACK(src, .proc/remove_boost, A, tag), 6 SECONDS, TIMER_UNIQUE|TIMER_OVERRIDE)
```
tattax:
```suggestion
	addtimer(CALLBACK(src, PROC_REF(remove_boost), A, tag), 6 SECONDS, TIMER_UNIQUE|TIMER_OVERRIDE)
```

Moltijoe: ah right, i completely forgot to update this part, ty

ynot01 (approved):

PR #18823, opened 2023-05-08
EMPing guns takes less charge but disables them for 5-10 seconds

ynot01:

EMPs take 25/50% instead of 50/100% of a gun's cell now, but also disable the weapon for 5-10 seconds to retain its usefulness in combat.

Currently, laying an EMP on someone who relies on energy weapons (like a security officer) immediately puts them out of the fight unless they want to take their chances at melee stun combat.

Now, EMPs are still advantageous but not so much that it wins the fight outright by disabling your opponent's ranged options for the rest of the engagement.

Yes it works for NT-USP too

# Changelog

:cl:  
tweak: EMPing an energy gun now takes 25/50% charge instead of 50/100%
tweak: EMPing an energy gun now disables it for around 5 or 10 seconds
/:cl:

### DIFF
--- a/code/modules/projectiles/boxes_magazines/external/rechargable.dm
+++ b/code/modules/projectiles/boxes_magazines/external/rechargable.dm
@@ -80,7 +80,8 @@
 			if(!G.magazine == src)
 				return
 			G.semicd = TRUE
-			var/unjam_time = ((2/severity) + (rand(-20,20)*0.01)) SECONDS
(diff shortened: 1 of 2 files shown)


Hopekz:
You're balancing these changes for neo from the matrix or something because this doesn't retain their usefulness in combat at all how are people supposed to take advantage of a gun being unusable for 1 second compared to what we previously had of making the gun not usable with 100% of the ammo being gone.

Moltijoe (approved):

PR #18838, opened 2023-05-09
Antagonists can tend wounds faster if they are severe or critical

SomeguyManperson:

Self-tending is almost necessary if an antag gets wounded so it shouldn't be as horrible compared to crew who can just go to medical and get treated by the like 15 people standing around there, so anyone with an antagonist role gets reduced penalties for using ghetto tools/self-tending wounds

Moderate wounds are exempt from this since they are both common and mostly harmless, and someone magically healing their papercut slightly faster would be a meta tell I can see people being shits about

# Wiki Documentation

Antagonists don't suffer the self tend penalty when tending wounds with normal tools, and have a boosted tending speed for using ghetto tools when tending wounds

# Changelog

:cl:  
tweak: Antagonists don't suffer the self tend penalty when tending severe or critical wounds with normal tools, and have a boosted tending speed for using ghetto tools when tending wounds
/:cl:


00ze-cyclone:
I feel like we should just make those wounds actually possible to heal by yourself, even if we buff self wound tending it'll remain inferior to medbay so crew will keep going to medbay.

Moltijoe (approved):

Necrosium:
Somebody give this guy a trophy

PR #18841, opened 2023-05-09
forcing open airlock with jaws or jaws-like does damage to the airlock

Skrem7:

except the airlock jimmy which should probably be in more parts of the game

is this easily repaired YES however i was requested to do it

# Changelog

:cl:  
tweak: Forcing open an airlock with jaws now damages the airlock
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -1315,6 +1315,8 @@
 				prying_so_hard = TRUE
 				if(do_after(user, time_to_open, src))
 					open(2)
+					if(!istype(I,/obj/item/jawsoflife/jimmy)) //You get to be special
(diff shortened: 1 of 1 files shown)


BlueishTsunami:
What about something like help intent is a slower open with no damage, and harm intent opens quicker but causes damage?

SomeguyManperson:
> What about something like help intent is a slower open with no damage, and harm intent opens quicker but causes damage?

the point is I believe to make jaws less harmless and closer to par with hacking the door

ToasterBiome (approved):

Hopekz:
> except the airlock jimmy which should probably be in more parts of the game

When you code it (I was thinking about doing this before when I made it just never got around to it). Maybe something for HoS or a higher tier upgrade for sec otherwise idk

can probably fit in some antag kits like the where's waldo chaser people etc

PR #18849, opened 2023-05-10
Ghetto IPC revival board

Moltijoe:
Adds an option for antags to revive dead IPC teammates without needing to raid robotics specifically
there's more alternative ways to revive humans so it's not really needed otherwise
reviving an ipc with this causes brain damage (enough for a trauma)

behold the codersprite
image

:cl:  
rscadd: Adds a ghetto ipc revival board
/:cl:


wonderinghost:
Based

ChuChu42:
i whould suggest reducing to 40-50 because 60 seems too much.

Moltijoe:
> i whould suggest reducing to 40-50 because 60 seems too much.

it takes 200 to die from it, this is just enough to cause a basic trauma
anything less is basically a non-issue

cuackles:
2 things, add the art add to the changelog and probably add an igniter or something like that to the recipe too

iamthesilly:
> > i whould suggest reducing to 40-50 because 60 seems too much.
> 
> it takes 200 to die from it, this is just enough to cause a basic trauma
> anything less is basically a non-issue

If the intent is to cause a trauma, you could just add the trauma manually instead of setting brain damage to 60

Moltijoe:
> If the intent is to cause a trauma, you could just add the trauma manually instead of setting brain damage to 60

the intent is to cause damage that can be repaired via other sources that also causes a negative effect until repaired
simply adding a trauma manually would work i suppose, but it just doesn't feel right to me

cuackles (changes requested):
think you could make the sprite less visually loud ?, a bit more clean, i know its suposed to be something makeshift but i still think the pieces it has should have some sense on where they are

wonderinghost:
You plug 4 different long boards into a machine. Only difference is the name. I'm fine with the sprites and that kinda makes sense

Moltijoe:
> think you could make the sprite less visually loud ?, a bit more clean, i know its suposed to be something makeshift but i still think the pieces it has should have some sense on where they are

Aight guess im closing this then since i have no clue what that means

PR #18763, opened 2023-05-02
Revert "Beepsky does stamina damage instead of instant stun"

JamieD1:
Reverts yogstation13/Yogstation#18691

He can be one tapped, skill issue

PR #18857, opened 2023-05-11
Makes beepsky do stamina damage instead of an instant stun, for real this time

SapphicOverload:

I couldn't properly revert #18763 due to conflicts caused by another PR made after it, so here we are. This is basically a revert of that PR with minor changes.

This makes beepsky take 2 hits to down an unarmored person and 3 hits to down an armored person, instead of instantly stunning on hit. Also fixes the attack not caring about armor or block chance, and increased the health while adding flat damage reduction to make it harder to kill.

Specifics:
Health of 50 (up from 25) and 6 flat damage reduction, welding tools deal 9 damage to it (6 hits to kill) and toolboxes deal 7 to it (8 hits to kill). Lasers will deal 14 damage to it (4 shots to kill). Anything weaker than a mosin will not be able to one-shot it anymore and trying to melee it with anything less than esword-level damage or anti-stuns will not be successful.
Attack does 60 stamina damage, 2 hits to down an unarmored person and 3 hits against someone with at least 17 energy armor.

# Changelog

:cl:  
tweak: beepsky is harder to kill but takes 2-3 hits to stun you
bugfix: fixed beepsky attack ignoring armor and block chance
/:cl:


Moltijoe (approved):

ToasterBiome:
Same PR that was reverted - 25 HP and +6 innate damage reduction

Therandomhoboo:
I got mixed feelings on this, however I say at least test-merge it and see how it goes really.

I know its because we are trying to get out of the stun2win era so I can understand why this has been made, but I also get why people would be against it as well.

iamthesilly:
> Same PR that was reverted - 25 HP and +6 innate damage reduction

I would have reverted it normally but I couldn't because of conflicts caused by another PR made since then, I decided to make a few minor adjustments because I felt it would be better to make it stronger against weak weaponry without making it too strong against high damage per hit weapons like mosins or hardlight bows.

Also would like to point out the original was reverted before 48 hours passed, and we didn't actually get to see how it actually affected the game.

MajManatee:
cringe

BurgerLUA:
> cringe

image

ToasterBiome:
Dragging speed was reduced, if it needs more adjustment in the future this will be merged.

PR #18859, opened 2023-05-11
Murders the old prefixed SQL template

alexkar598:
The idea behind prefixing SQL tables is that if youre using shared hosting, you can share your database with wiki software and forums software or whatever because your hosting provider is too cheap to give you more than 1 db.

Noone will ever host ss13 on shared hosting, and this is being a pain in the ass. NOBODY should be setting up new prefixed production or development environments in 2023. This slows down dev speed because I have to update 3 files every SQL change for no reason. 

@yogstation13/head-coders 
@adamsong


adamsong (approved):
lgtm

ToasterBiome (approved):

PR #18867, opened 2023-05-12
Legions can now move through their own skulls

SapphicOverload:

Allows legions (and hivelords which they're a subtype of) to move through their own skulls. This doesn't mean the skulls can pass through each other and stack up, just the legions themselves can pass through the skulls they create.

Why? My goals are beyond your understanding.

# Changelog

:cl:  
tweak: legions can pass through their own skulls
/:cl:


ynot01 (approved):
Code seems like it might only allow skulls to pass through the mother, but it's fine either way

PR #18873, opened 2023-05-12
Flying fang tweaks

SapphicOverload:

Flying fang now lets lizards move faster when crawling, and leap while lying down. Also reduces the self-stuns. Leap now has a 5 second cooldown if you miss.

# Changelog

:cl:  
tweak: flying fang lets you crawl faster
tweak: reduced flying fang self-stuns
tweak: adds 5 second cooldown to leap attack
/:cl:


code/datums/martial/flying_fang.dm
```diff
@@ -196,14 +200,15 @@
 	if(linked_martial.leaping)
 		return
 	linked_martial.leaping = TRUE
+	linked_martial.next_leap = world.time + 5 SECONDS
 	A.Knockdown(5 SECONDS)
-	A.Immobilize(30 SECONDS) //prevents you from breaking out of your pounce
+	A.Immobilize(3 SECONDS, TRUE, TRUE) //prevents you from breaking out of your pounce
```
SomeguyManperson, line 205: For reference, this exists solely to prevent moving while the pounce is active and is reset to a more reasonable amlunt in leap_end, because being able to control or interrupt your own attack by holding a movement key would be problematic

iamthesilly: I am aware, I set it to 3 seconds instead of 30 because leaps last less than one second and having it immobilize you for 30 seconds can cause problems if you accidentally leap into space or something and never hit anything to end the leap (this happened to me with ultra violence a few times)

iamthesilly: Also set it to ignore antistuns so they wouldn't let you cancel it

code/datums/martial/flying_fang.dm
```diff
@@ -179,6 +180,9 @@
 		return FALSE
 
 /datum/action/innate/lizard_leap/Activate(silent)
+	if(world.time < linked_martial.next_leap)
+		to_chat(owner, span_warning("You aren\'t ready to pounce again yet!"))
+		return
```
Moltijoe: use cooldown declares

Moltijoe (changes requested):

code/datums/martial/flying_fang.dm
iamthesilly: okay

iamthesilly:
@Moltijoe

Moltijoe (approved):
unfortunate lizard W

PR #18876, opened 2023-05-13
Mime drinks, Silencer + Blank Page Buff

Therandomhoboo:

We care about diversity, so with that in mind. Mime drinks are no longer exclusive for those pesky human mimes. As well as Silencer and Blank page should do more healing than their previous ones because they require "Nothing" to make them both. 

[EDIT] I was wrong about the humans

# Changelog

:cl:
  
tweak: Silencer and Blank Page now do more healing than Nothing because why wouldn't it?

/:cl:


Mqiib:
ishuman(M) returns true for any mob/living/carbon/human, which includes all the different species, so like lizards and moths too. ishumanbasic(M) is what would specifically check for the human species.

Therandomhoboo:
> ishuman(M) returns true for any mob/living/carbon/human, which includes all the different species, so like lizards and moths too. ishumanbasic(M) is what would specifically check for the human species.

Ah I see, I looked at some of the other job related ones (like hooch) and saw no ishuman(M). Should I revert these parts so its less hassle or would it be fine to leave it how it is?

PR #18900, opened 2023-05-17
Removes cark's donator bedsheet from the list of randomizable bedsheet

Moltijoe:
It's a ckey locked donator item, it shouldn't be a random spawn

:cl:  
bugfix: ckey locked bedsheet no longer spawns as a random bedsheet
/:cl:


code/game/objects/structures/bedsheet_bin.dm
ynot01, line 1: avoid infinite loop: get a list of all possible bedsheets (`typesof(/obj/item/bedsheet) - /obj/item/bedsheet/random`), use the `shuffle()` proc to shuffle it, then iterate on that list until you find a randomizable bedsheet (and throw an error if you find none)

ynot01 (changes requested):

code/game/objects/structures/bedsheet_bin.dm
```diff
@@ -280,8 +281,16 @@ LINEN BINS
...
 /obj/item/bedsheet/random/Initialize()
 	..()
-	var/type = pick(typesof(/obj/item/bedsheet) - /obj/item/bedsheet/random)
-	new type(loc)
+	var/list/sheets = list(typesof(/obj/item/bedsheet) - /obj/item/bedsheet/random)
+	sheets = shuffle(sheets)
+	var/obj/item/bedsheet/sheet
+	for(var/i in sheets)
+		sheet = new i(loc)
+		if(sheet.randomizable)
```
ynot01: typecast `i` `as anything` and use `initial(i.randomizable)` and create the bedsheet on the `break` instead of creating and qdeling unneeded sheets

```js
for(var/obj/item/bedsheet/sheet as anything in sheets)
  if(initial(sheet.randomizable)
    break
```

ynot01 (changes requested):

code/game/objects/structures/bedsheet_bin.dm
Moltijoe: huh, didn't realize you could check values of type paths

ynot01 (approved):

ynot01 (approved):

PR #18905, opened 2023-05-17
Reworks hydrogen and tritium fires to actually make sense

SapphicOverload:

Hydrogen and tritium fire burn rate is now dependent on both fuel and oxygen concentration instead of just the fuel amount and a binary "is the oxygen above 50% concentration" check.

This will make single tank bombs unable to reach the bomb cap without adding unnecessary checks or removals.

# Changelog

:cl:  
tweak: hydrogen/tritium fires make sense now, no more making timed singletanks with them
/:cl:

### DIFF
--- a/code/__DEFINES/reactions.dm
+++ b/code/__DEFINES/reactions.dm
@@ -44,15 +44,11 @@
 #define NITRO_BALL_HEAT_SCALE 100000
 
 //tritium reaction
-#define TRITIUM_BURN_OXY_FACTOR				100
 #define TRITIUM_BURN_TRIT_FACTOR			10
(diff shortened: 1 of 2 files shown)


DylonBR:
Cringe

MajManatee:
sarah single handedly making toxins worse and removing fun every day

iamthesilly:
> sarah single handedly making toxins worse and removing fun every day

This will not affect toxins

iamthesilly:
Actually it will affect toxins, you will be able to use an exact 50/50 mix of hydrogen/tritium and oxygen without worrying about it being significantly weaker than 51% oxygen just because it hadn't crossed the oxygenation threshold yet

ToasterBiome (approved):

ShoziOffical:
oh so this is why i couldnt do single tank bomb to explode clockies anymore

PR #18910, opened 2023-05-18
semicircle of cogs 2 electric boogaloo

azzzertyy:
Clockies have a hard time defending ever since we became circle of cogs instead of square of cogs.
This changes that by returning coggies to their eternal roots, giving them access to a single rectangular wall within their grand establishment, turning it into a semicircle of cogs.

This also changes some of the interior sections of the city, making it so there is a second camera room and merges the southwest room with the northwest room.

This also changes the glass in the interior section near the ark to be normal walls, this was too easy to just shoot through and I dont know why anyone thought it was good.

This also makes the entire thing a few tiles larger, the actual combat area has a 2 tile wider circumference.

# Wiki Documentation
Image of reebee will need changing

image

# Changelog
:cl:  
mapping: Changes city of cogs to be a semicircle of cogs, giving clockies a buff.
/:cl:


azzzertyy:
sorry for having to close the last one, with a bit of luck ive fixed whatever the git rebase demon wanted with me

Aquizit:
Still missing the 3 cogscarab spawns.

Aquizit (changes requested):
Ok, had another test go -- the whole thing needs to be widened out some. It's too easy to make everything into 1-tile wide chokes. Which I understand is part of the point but with cultists about to fall back and revive it's a bit too much.

I would like to see it widened out 2 tiles all around.

Mqiib:
Fix the outside windows

azzzertyy:
> Fix the outside windows

they already are

Xoxeyos:
why god why, why can't you just make this a vertical battlefield that goes North to South, instead of East to West?
I assume most users have a 16:9 monitor or something similar to it, it would be much more efficient for construction, combat, etc, since the mouse has less distance to move.

azzzertyy:
> why god why, why can't you just make this a vertical battlefield that goes North to South, instead of East to West? I assume most users have a 16:9 monitor or something similar to it, it would be much more efficient for construction, combat, etc, since the mouse has less distance to move.

because the suggestions on the pr were full of east to west instead of north to south
i did not choose this i wanted it north to south

Aquizit (approved):

KonKonBreezeFox:
i dont like the fact that it is pointed to the right, isnt left way better

PR #18911, opened 2023-05-18
MURDERS buster arm

LazennG:

tweaked AP for the normal buster style moves since they were originally that high due to how tough the moves were to land but implementing buster style made it so that isn't the case anymore. 

AP was reduced from 35 to 15, so on a normal officer, the damage values are:
Slam doing 15
point blank mop the floor doing 30
lobbing them without hitting anything doing 11 and hitting something doing 16
wire slam doing 14

megabuster damage values are unchanged because it's still tricky to land in a real fight

# Changelog

:cl:  

tweak: tweaked buster style AP

/:cl:


Necrosium:
Why, at last! Finally off the buster arm tyranny!

ynot01 (approved):

PR #18915, opened 2023-05-18
Modifies gax engineering

SomeguyManperson:

image

emitters are placed in line with eachother like every other map so they are able to be turned on/off easier
emitters also have the goofy girder+grille behind them removed since it just gets in the way of turning them off without walking directly in front of them

Missing oxygen canister (I assume it's missing?) added to the hardsuit area since there's a tank dispenser, an air pump, but no oxygen canister to fill the air pump with

# Wiki Documentation

updated map image

# Changelog

:cl:  
mapping: gax engineering slightly more intuitive
/:cl:


ToasterBiome:
image

ToasterBiome (changes requested):
add broken grates behind the leftmost two emitters, to at least remind of what you took from me ;-;

ToasterBiome (changes requested):
image

Should only be like 50 lines changed

ToasterBiome (approved):

PR #18924, opened 2023-05-19
Delets unused binaries

alexkar598:
Part of replays 2.0, which was reverted never to be talked about ever again


adamsong:
@monster860 Any reason to keep these?

monster860:
If I revive the project I'll just put the binaries back so no

adamsong (approved):

PR #18925, opened 2023-05-19
Kills libvg

alexkar598:
All of this has been a massive no-op for years, the linux binary is unavailable and on windows its not being used since 513, which isn't even supported for yogs anymore.


adamsong (approved):
lgtm

PR #18928, opened 2023-05-19
Deletes the mirrorbot

alexkar598:
Lol

Same story, broken script, not supported, will likely break


adamsong (approved):
Look how the massacred my boy (this was before my time)

yacabo:
It served us well

Well not really but still F

PR #18929, opened 2023-05-19
Delete yarn.lock

alexkar598:
This file is for tgui, except its empty and not located in the right folder.

It's unneeded and even if some program could find it, it would be actively harmful


adamsong (approved):
lgtm

MajManatee:
Why does this yarn have to be so complicated?

PR #18930, opened 2023-05-19
Deletes appveyor

alexkar598:
We never use appveyor, we inherited it from a (failed?) TG project, this stuff isn't able to build master anyways


adamsong (approved):
lgtm

PR #18931, opened 2023-05-19
Removes docker

alexkar598:
This dockerfile hasn't been functional in eons


adamsong (approved):
lgtm, will cause merge conflicts with the 515 pr

PR #18945, opened 2023-05-20
Stamina damage can disable limbs again

SomeguyManperson:

the one damage type that should still be able to do this

:cl:  
tweak: stamina damage can disable limbs again
/:cl:


ynot01 (approved):
strange that it was done this way

PR #18947, opened 2023-05-20
Stamina rework - Dynamic Regeneration

ynot01:

In addition to clearing all stamina damage after 10 seconds of not taking damage, instances of stamina damage† will now regenerate independent of other sources of stamina damage. Once damage is received, 10 seconds later you will heal for 65% of that damage.

Example:
Two disabler beams (30 dam each) hit a player, two seconds apart, giving the player 60 stamina damage. After 10 seconds, the player will only have 40.5 stamina damage, because the first shot had "worn off". After 12 seconds, the player had not taken any damage for 10 seconds and thus would have 0 stamina damage.

## †Chemicals

All "continuous" forms of stamina damage, like being applied by chemicals clear your regeneration. You will NOT regenerate stamina while this is happening.

# Goal

The goal of this rework is primarily to weaken stamina weaponry in prolonged engages.

- Getting hit by stray disabler shots while dodging most (i.e. getting hit every 10 seconds, resetting your stamina timer) is less punishing.

Energy weapons are high damaging and hard to armor against, but are meant to wear off. If you want to "slowly wear someone down" you should use real damage.

## UPDATE 6/27/2023

Stamcrit now clears all regen, people will no longer get up mid-stamcrit from regeneration

# Changelog

:cl:  
experimental: Instances of stamina damage now partially regenerate individually unless in stamcrit. After 10 seconds of receiving a disabler shot, you will heal for a portion of that damage, regardless if you were shot more or not.
/:cl:


azzzertyy:
bad day for ai players

Moja10:
Would be interested in a test merge, don't think its possible to judge either way until a test

Therandomhoboo:
I don't like the current push we are doing towards forcing more lethal scenarios. I think this just will make sec roll out with lethals at any chance of not being to get the antag down which will stress everyone out. However, like Moja just said no way to really judge unless a test happens.

ToasterBiome:
“work to keep them down” 

best stun in the game is death

ynot01:
> best stun in the game is death

lethals are generally less potent than non-lethals

additionally, if you invite security to a lethal fight you should expect to suffer the consequences

code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
```diff
@@ -1892,6 +1892,9 @@ All effects don't start immediately, but rather get worse over time; the rate is
 /datum/reagent/consumable/ethanol/fanciulli/on_mob_metabolize(mob/living/M)
 	if(M.health > 0)
 		M.adjustStaminaLoss(20)
+		if(iscarbon(M))
+			var/mob/living/carbon/C = M
+			C.clear_stamina_regen()
```
JohnFulpWillard: ``clear_stamina_regen`` should probably be on the living level with ``adjustStaminaLoss``, they are both for stamina so it makes no sense for them to be on different levels. iscarbon check every time looks terrible.

solwashere:
Don't like this anymore now that it's back at someone instantly winning when they use a disabler

Moltijoe (approved):

PR #18966, opened 2023-05-23
Mime drinks aren't the same v2.0 no more stacks

Therandomhoboo:

Essentially, all mime drinks give the same heal. I've tweaked it so that each drink tier heals more
Nothing = 1.0 heal
Silencer = 1.5 heal
Blank Paper = 2.5 heal
The reasoning behind this is that Silencer needs Nothing to make, and Blank paper requires both Nothing & Silencer to make.
As suggested too, I've put in a way to prevent mimes from stacking heals, therefore only one drink may be in the mime system at all times. 

I.E, if they drank Nothing first and then blank paper & silencer, the nothing will cancel out blank paper & silencer till Nothing reaches 0u's. The same apply for the rest.

# Changelog

:cl:  
rscadd: Mimes can't stack their job drinks now and only one will be in their system at any time  
tweak: Silencer heals for 1.5, and blank paper heals for 2.5
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
@@ -1216,9 +1216,9 @@ All effects don't start immediately, but rather get worse over time; the rate is
 	glass_desc = "A drink from Mime Heaven."
 
 /datum/reagent/consumable/ethanol/silencer/on_mob_life(mob/living/carbon/M)
-	if(ishuman(M) && M.job == "Mime")
+	if(M.mind?.assigned_role == "Mime")
 		M.silent = max(M.silent, MIMEDRINK_SILENCE_DURATION)
-		M.heal_bodypart_damage(1,1)
+		M.heal_bodypart_damage(1.5,1.5)
 		. = 1
 	return ..() || .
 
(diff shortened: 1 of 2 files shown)


Therandomhoboo:
I want to add this on too. If people want something different, then I'm open for suggestions, but how I see it. It makes sense for the healing increase for the fact it uses previous drinks to make it up. There is also other drinks/medicine that just do the same heal but better that I haven't touched.

Moltijoe:
balance council doesn't seem to like the idea of it just being a stronger healing drink

Therandomhoboo:
> balance council doesn't seem to like the idea of it just being a stronger healing drink

Literally uses itself to do make the higher drinks and then it uses two drinks (Nothing and silencer) to make the final drink, with that logic Quint sec should be gutted and half the chems should be gutted since there is ones that heal for less and some that heal for more.

Moltijoe:
> Literally uses itself to do make the higher drinks and then it uses two drinks (Nothing and silencer) to make the final drink, with that logic Quint sec should be gutted and half the chems should be gutted since there is ones that heal for less and some that heal for more.

quint sec is a lot harder to make because it require bananium ore, but it's for an entire department and one that healing from a drink would thematically make sense for
there's no thematic reason that mimes should have drinks that heal them
chems get a pass because that's what they were designed for

PR #18972, opened 2023-05-23
Removes refined dragons blood from the uplink

Mqiib:
I brought it into this world, I can take it out! 
Some have voiced concerns about how it's sort-of-but-not metashielded because it COULD be lavaland loot but also it never is and that's kind of annoying to deal with because it's a big fat meta slap in the face. Draconids should be much rarer now.

Dragon's blood is no longer in the uplink, instead there's a 1/4 chance of getting refined blood rather than regular blood from an ash drake so now you know if you're about to turn into a lizard or not. The draconid roll was removed from regular blood, and the refined blood now works on any species.

# Wiki Documentation

Refined dragons blood no longer in the uplink.
Ash drake loot now includes a 1/4 chance of refined blood rather than regular blood.

# Changelog

:cl:  

rscdel: Refined dragons blood is no longer available to lizard traitors
tweak: Refined blood can now drop instead of normal blood (25% chance) from ash drakes
tweak: Refined blood can be used by any species now

/:cl:


code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -942,7 +942,11 @@ GLOBAL_LIST_EMPTY(aide_list)
 		if(3)
 			new /obj/item/dragon_egg(src)
 		if(4)
-			new /obj/item/dragons_blood(src)
+			loot = rand(1,4)
+			if(loot == 4)	//Still same chance but now you know if you're turning into a lizard (ew)
+				new /obj/item/dragons_blood/refined(src)
+			else
+				new /obj/item/dragons_blood(src)
```
Moltijoe: just use prob(25) with an else if you're only caring about a single value of rand()

Moltijoe (changes requested):

Moltijoe (approved):

PR #18978, opened 2023-05-24
Fixes an oversight causing the singularity to eat incoming particles, providing heavily reduced or no energy

SomeguyManperson:

# Spriting

# Wiki Documentation

# Changelog

:cl:  
bugfix: thw singularity no longer obliterates particles from the particle accelerator before recieving power from them
/:cl:


Moltijoe (commented):
should hitting the singularity delete it after applying energy then
sort of like what happens when hitting a blob structure in the bump proc?

Moltijoe (approved):
ah, singularities are dense, so it shouldn't matter

PR #18979, opened 2023-05-24
Fixes the hand teleporter gettign a little quirky when failing to dispel portals, further proving the theory that yogsdev debugging is "let theos use it"

SomeguyManperson:

Removes extra dispel on afterattack since dispel gets called both on pre attack and after attack for... some reason. Thid would allow a second attempt at dispelling a portal if you moved away from it during the first one, rather than;
Failing to dispel a portal no longer forces you to teleport through it, since for some reason we decided to pass whether the dispel was successful as whether or not we should hit the portal normally

# Spriting

# Wiki Documentation

# Changelog

:cl:  
bugfix: the hand teleporter no longer acts quirky when failing to dispel portals, which would result in attempting to dispel it a second time or teleporting you through it
/:cl:


SomeguyManperson:
Yes, I did die to this

ToasterBiome (approved):

PR #18625, opened 2023-04-23
space dragon qol, refactoring and misc

tattax:
ports what was leftover in the original pr in respect to refactoring, QOLs and etc, together with other various bugfixes
since the original one felt like something was missing from the antag (it was)

mostly from : https://github.com/tgstation/tgstation/pull/68988

closes: https://github.com/yogstation13/Yogstation/issues/18871
closes: https://github.com/yogstation13/Yogstation/issues/18921
raises weight from 5 to 7, since 5 is really low considering space dragon doesn't immediately ends round like blob (weight 8)

### gust is now activated with _ALT-CLICK_, instead of being an action

dragon roar SFX: 

https://vocaroo.com/1cRq1jBzGDD1

dragon with color (name is also customizable):
image

:cl: IndieanaJones, Timberpoes, tatax
tweak: space dragon qol: allows you to choose color and name, adds nightvision, increases dragon weight from 5 to 7
soundadd: space dragon roar
/:cl:

PR #19000, opened 2023-05-26
Fixes Space Dragon Rifts not Progressing Whatsoever

dcxgame:

As it turns out, every check a space dragon's rift does hinges on time_charged being an even number, which can never happen if delta_time is always 2 and it starts at 1. This one-line change makes the space dragon midround function exactly as intended.

I tested the event after making this change, and can assure you that no other bugs appear to be present related to the rift mechanics.

It should be noted that space dragons actually end the round if they're allowed to succeed, so this may have a large impact on people's playstyles when the event is called, and that may warrant adjusting its weight in the future, at your discretion.

# Changelog

:cl:  
bugfix: space dragon can again place more than one rift, rifts again alert the crew when they're about to finish charging, and carps again can enter through the rift more than once during charging
/:cl:


tattax:
https://github.com/yogstation13/Yogstation/pull/18625

ToasterBiome:
Solved in https://github.com/yogstation13/Yogstation/pull/18625

PR #19021, opened 2023-05-27
[admin] resolving a ticket now sets you as its owner if there is none

ynot01:
image


adamsong (changes requested):
Do this for closing a ticket as well

adamsong (approved):
lgtm

PR #19042, opened 2023-05-28
Kills permeability coefficient

SapphicOverload:

Chemical and disease permeability is now based on bio armor, which permeability coefficient has been replaced with. Every 15 bio armor reduces permeability by half (15 is 0.5, 30 is 0.25, 60 is 0.0625, etc) and 100 bio armor grants complete protection. This makes things like foam actually work as intended, makes bio armor actually relevant, and allows you to know roughly how well something will protect you from chemical spills and disease by examining it, making the entire system more intuitive for people who don't code-dive.

This also makes certain touch chems that should be affected by permeability actually take it into account, fixes xeno neurotoxin only working on targets WITHOUT protection instead of targets WITH protection, and changes the xeno neurotoxin spit to care about bio armor instead of bullet armor.

I might have missed a few things so I highly recommend test-merging this first before actually merging it, just in case.

# Wiki Documentation

Don't think there's anything about permeability coefficient but if there is, remove it.

# Changelog

:cl:  
bugfix: make permeability actually work
bugfix: fixes xeno neurotoxin only working on targets with protection from it
tweak: toxin damage from neurotoxin spit cares about bio armor instead of bullet armor
tweak: reagent/disease permeability now uses bio armor instead of a permeability coefficient
tweak: certain touch chems now take permeability into account when they previously did not
tweak: changed the bio armor on lots of stuff so they have the same or similar intended permeability as before
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -747,7 +747,10 @@
 // permeability: now slightly more sane and probably functional!
 /mob/living/carbon/get_permeability(def_zone)
 	if(def_zone)
-		return 1 / (2**(getarmor(def_zone, BIO) / 15))
+		var/permeability = 1 / (2**(getarmor(def_zone, BIO) / 15))
+		if(permeability <= 0.01)
+			return 0
+		return permeability
 
 	var/total_bodyparts = 0
 	var/total_permeability = 0
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -753,7 +753,7 @@
 	var/total_permeability = 0
 	for(var/obj/item/bodypart/BP in bodyparts)
 		total_bodyparts++
-		var/protection = getarmor(BP.body_zone)
+		var/protection = getarmor(BP.body_zone, BIO)
 		if(protection < 100)
 			total_permeability += 1 / (2**(protection / 15)) // every 15 bio armor reduces permeability by half (15 is 0.5, 30 is 0.25, 60 is 0.0625, etc)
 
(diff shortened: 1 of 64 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -744,17 +744,24 @@
...
-				tally["[zone]"] = max(1 - I.permeability_coefficient, target_zones["[zone]"])
-	var/protection = 0
-	for(var/key in tally)
-		protection += tally[key]
-	protection *= INVERSE(target_zones.len)
-	return protection
+// permeability: now slightly more sane and probably functional!
+/mob/living/carbon/get_permeability(def_zone)
+	if(def_zone)
+		return 1 / (2**(getarmor(def_zone, BIO) / 15))
```
Moltijoe: this one doesn't return 0 if below 0.01?

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -744,17 +744,24 @@
...
+
+	var/total_bodyparts = 0
+	var/total_permeability = 0
+	for(var/obj/item/bodypart/BP in bodyparts)
+		total_bodyparts++
+		var/protection = getarmor(BP.body_zone)
+		if(protection < 100)
+			total_permeability += 1 / (2**(protection / 15)) // every 15 bio armor reduces permeability by half (15 is 0.5, 30 is 0.25, 60 is 0.0625, etc)
```
Moltijoe: no reference to bio armour?

iamthesilly: weird how it still worked without that, not sure how i missed this considering it's probably the most important part of the pr

Moltijoe (approved):
can't seem to find any other problems, but this is a big enough PR that i'd like a second opinion

Moltijoe:
aight, it's marinated enough, time to pull the trigger and hope nothing breajs

PR #19047, opened 2023-05-28
[PORT] Adds the delamination counter from /tg/

azzzertyy:

Ports https://github.com/tgstation/tgstation/pull/70099

Should add a number whenever the SM doesnt delaminate in a round, and reset back to 0 when it does.

Currently I haven't tested the persistence, I'm unsure how to test this and I am assuming it does not work, so some help here would be nice.

# Spriting
image
image

# Changelog
:cl:  
rscadd: Added the SM delam counter from tg
/:cl:


Aquizit:
So as far as I know, tg *only* runs SM, right? This will be counting rounds we don't actually have a supermatter engine? I mean it's still *technically* correct in that case but.

azzzertyy:
> So as far as I know, tg _only_ runs SM, right? This will be counting rounds we don't actually have a supermatter engine? I mean it's still _technically_ correct in that case but.

i think thats fine

azzzertyy:
Also it only is on maps with an SM.

ynot01:
a singularity loose detector would go nicely with this, there is an admin alert that says when singularities/teslas spawn and there's no containment field, could use similar logic

azzzertyy:
> a singularity loose detector would go nicely with this, there is an admin alert that says when singularities/teslas spawn and there's no containment field, could use similar logic

this would count singularities on charlie

Aquizit (approved):
Map stuff is fine.

cuackles (approved):

PR #18909, opened 2023-05-18
Traitor: Episode III - Revenge of the Desword

ynot01:
# Document the changes in your pull request

Baby, I'm sorry; can you ever forgive me?

# Changelog

:cl:  
rscadd: Added the double-bladed energy sword back to the traitor uplink
/:cl:

PR #19065, opened 2023-05-30
Revert "Traitor: Episode III - Revenge of the Desword"

solwashere:
Reverts yogstation13/Yogstation#18909

This was a bad idea in the first place, I think it would honestly be better if maybe you could craft two eswords together for a makeshift desword with flat 75 block chance and nothing else (which is still pretty goofy)
Desword lets you completely negate all energy and laser attacks and has a 75 block chance for "everything else" and generally completely destroys any semblance of balance regarding trying to pick a strategy and it being counterable with another strategy.
Which is melee and bullet, with bulletproof armor being two rwalls or one cargo order away, making bullets laughably bad and melee your only viable option against a desword user (good luck with that), and this isn't even factoring in any other TC that you can get from bounties, contractor, and other traitors. The only TRULY viable strategy is bomb but security can't go around chucking singlecaps can they?

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -412,7 +412,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	item = /obj/item/twohanded/dualsaber
 	player_minimum = 25
 	cost = 16
-	exclude_modes = list(/datum/game_mode/nuclear/clown_ops, /datum/game_mode/infiltration) // yogs: infiltration
+	include_modes = list(/datum/game_mode/nuclear)
 
 /datum/uplink_item/dangerous/doublesword/get_discount()
 	return pick(4;0.8,2;0.65,1;0.5)


hwseo7:
Please wait I didn't roll traitor yet

MajManatee:
Lame -1

neeshacark:
isnt it meant to be hijack only?

solwashere:
> isnt it meant to be hijack only?

no, it's pop locked to 25

Moltijoe (approved):
code is ez

Moja10:
poop change, it's fine, but if you are hellbent I would rather have dewsword split into traitor and nukie versions and rhe traitor versi9n slightly nerfed than  outright removal.

hwseo7:
> poop change, it's fine, but if you are hellbent I would rather have dewsword split into traitor and nukie versions and rhe traitor versi9n slightly nerfed than outright removal.

This actually sounds like nice solution, maybe craftable desword with two esword that is not fucking immune to laser?

solwashere:
> > poop change, it's fine, but if you are hellbent I would rather have dewsword split into traitor and nukie versions and rhe traitor versi9n slightly nerfed than outright removal.
> 
> This actually sounds like nice solution, maybe craftable desword with two esword that is not fucking immune to laser?

 This was a bad idea in the first place, I think it would honestly be better if maybe you could craft two eswords together for a makeshift desword with flat 75 block chance and nothing else - me 9 days ago on this PR

hwseo7:
> 

I stole your idea :)

N3D6:
yeah

RG4ORDR:
So after having gotten to use it. The new meta for IAA and teaming traitors, I'm pretty confident in saying this thing needs to fucking stay with Nukies only. We currently have no system nor no other solution that doesn't make the Desword entirely useless or entirely unfair. 
We can remove the deflection but it's not the deflection that makes it strong it's the 75% block chance. 
We can't reduce the block chance because anything lower than 75 means that you're better off buying a singular esword at the cost of slightly reduced damaged.
It needs to be taken off traitors, there is no bandaid fix for this.

solwashere:
closed because #19242

PR #19066, opened 2023-05-30
plain burger needs plain patty to craft

ktlwjec0:
you could make a plain burger with a plain patty, corgi patty, human patty etc. seems an oversight, this fixes that.

:cl:  ktlwjec
bugfix: plain burger can only be crafted with a plain patty
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/recipes/tablecraft/recipes_burger.dm
+++ b/code/modules/food_and_drinks/recipes/tablecraft/recipes_burger.dm
@@ -73,7 +73,7 @@
 /datum/crafting_recipe/food/burger
 	name = "Burger"
 	reqs = list(
-			/obj/item/reagent_containers/food/snacks/patty = 1,
+			/obj/item/reagent_containers/food/snacks/patty/plain = 1,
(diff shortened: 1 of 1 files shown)


neeshacark:
i think calling this a fix is entirely wrong.  im pretty confident this was fully the intent.

unless this is a joke

neeshacark:
KTL ARE YOU /SRS OR /J!!!?!!! ARE YOU /SRS OR /J!!??

ktlwjec0:
> i think calling this a fix is entirely wrong. im pretty confident this was fully the intent.

I have literally got no clue how you think this is intended. tg's original change that added patties had this as a plain patty, and currently has it still with the burger named slightly. The other burgers that don't use special meats all have it as a plain patty (super/big bite, colored). Why should I be able to take a piece of human, xeno, etc meat and turn it into a completely normal burger? Especially human that is supposed to share the victim's name (currently broken, but intended), yet I can bypass this completely by just crafting a plain burger. This was very obviously an oversight when we ported across the change

Moltijoe (approved):

PR #19068, opened 2023-05-30
Meteors no longer explode space turfs

SomeguyManperson:

yes, this is stupid

Tested on a private server by spawning 20 meteor events of varying levels, seemed to reduce meteor lag but I can't tell without a testmerge

# Changelog

:cl:  
experimental: meteors will no longer explode empty space tiles, which should impact nothing since the tile would be empty anyways, but may possibly reduce the amount of stress they put on everything else
/:cl:


Moltijoe (approved):

SomeguyManperson:
Tested and works

PR #19080, opened 2023-05-31
Moves AI's data core display to AI upload in all maps except gax.

azzzertyy:

just makes sense innit

image
image
image

:cl:  
mapping: Moved ai core display to upload on all maps except gax
/:cl:

### DIFF
--- a/_maps/map_files/AsteroidStation/AsteroidStation.dmm
+++ b/_maps/map_files/AsteroidStation/AsteroidStation.dmm
@@ -5738,6 +5738,10 @@
 /obj/effect/turf_decal/tile/white{
 	dir = 8
 	},
+/obj/structure/table,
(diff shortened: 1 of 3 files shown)


Aquizit:
My one question is: why move it? Why not just add this one and keep the other one as well?

KingSuperBOB:
Honestly thats a good question why move it instead of making more

azzzertyy:
> My one question is: why move it? Why not just add this one and keep the other one as well?

maybe, ill consider

Aquizit:
Need a commit either adding the Sat ones back or a comment you aren't going to otherwise I don't know when I should be reviewing this.

MajManatee:
Id perfer to see one in both locations

Aquizit (approved):
Well it's fine. Someone else (I might later) can just add the second one in later.

PR #19082, opened 2023-05-31
glasses color toggle text is now correct

ktlwjec0:
the text was switched around, so it would say "You will no longer see glasses colors." when you toggle it on and "You will now see glasses colors." when you toggle it off.

:cl:  
bugfix: glasses color toggle text is now correct
/:cl:


JohnFulpWillard (approved):

JohnFulpWillard:
Do you not have a Changelog for this however?

This goes for all your other PRs too, please add the player-facing changes to them.

PR #19087, opened 2023-06-01
fixes hypospray vendor being free because i hate medbay

azzzertyy:
to clarify, it showed as costing money to medical doctors when it wasnt, and to people who werent medical doctors it showed as costing money and didn't charge you when you bought it, which it should.

:cl:  
bugfix: fixed hypospray vendor being free
/:cl:

### DIFF
--- a/code/modules/vending/medical_wall.dm
+++ b/code/modules/vending/medical_wall.dm
@@ -32,28 +32,28 @@
 	tiltable = FALSE
 	density = FALSE
 	// No default products, all of this shit costs money
-	premium = list(	/obj/item/hypospray = 5,
-					/obj/item/hypospray_upgrade/quickload = 5,
-					/obj/item/reagent_containers/glass/bottle/vial/libital = 10,
-					/obj/item/reagent_containers/glass/bottle/vial/aiuri = 10,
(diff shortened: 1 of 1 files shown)


ynot01:
elaborate? and why does everything now have 10x its normal stock? I think you misunderstand vending code greatly

MajManatee:
Fairly certain this was intentional, and not a bug too.

azzzertyy:
> Fairly certain this was intentional, and not a bug too.

it absolutely was not intentional, comment even says so, "all of this shit costs money" and they all had their prices labled in-game but did not charge, it should be free to meddoctors out of the medical budget but not to other crewmembers

> elaborate? and why does everything now have 10x its normal stock? I think you misunderstand vending code greatly

believe youre correct, woops, my bad, but yes. everything previously was free because vending machines NEED non premium stock to work, you cant just set everything they sell to premium because that causes it to not work properly

hedgehog1029:
this was broken by my vending custom compartment PR when I refactored the vending code into multiple procs. i've fixed the underlying issue in #19092, premium-only vending machines work fine again

MajManatee:
no longer relevant

PR #19089, opened 2023-06-01
alcohol tolerance mutation has instability

ktlwjec0:
Makes alcohol tolerance cost 5 instability (same as glowy). It's a minor positive mutation but it is still a positive mutation, so I think it needs some instability. It's never made sense to me why you can just get it for free.

I'll change wiki for this.

:cl:  Ktlwjec
tweak: Alcohol Tolerance mutation instability is now 5.
/:cl:


Therandomhoboo:
Why you driving away all the barkeeps customers? They need as many as they can get.

Moltijoe (approved):

PR #19101, opened 2023-06-02
Pulling bots slows you down

MajManatee:
For some reason molti made this just apply to sec-bots.
so now it applies to all bots (including mules, but those aren't pull able.)
No more outsourcing your jobs Paramedics with medbots, Sec with Secbots, Janitors with Cleanbots, Engineers with floorbots? 
Whatever. 
Do your jobs, and let the bots do theirs.

# Changelog

:cl:  
tweak: ALL bots apply a slowdown, not just secbots
/:cl:

### DIFF
--- a/code/modules/mob/living/living_movement.dm
+++ b/code/modules/mob/living/living_movement.dm
@@ -51,7 +51,7 @@
 		if(drag_slowdown && !(L.mobility_flags & MOBILITY_STAND) && !L.buckled && grab_state < GRAB_AGGRESSIVE && !istype(L, /mob/living/simple_animal))
 			add_movespeed_modifier(MOVESPEED_ID_PRONE_DRAGGING, multiplicative_slowdown = PULL_PRONE_SLOWDOWN)
 			return
-		else if(drag_slowdown && istype(L, /mob/living/simple_animal/bot/secbot))//snowflake proc to stop sec from dragging beepsky everywhere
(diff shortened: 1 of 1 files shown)


Lichton:
Cleanbots and floorbots do NOT need this nerf. Secbots are extremely powerful against antags, and medibots could be argued to be powerful enough healing... but not really, since they can just heal damage types and not wounds and shit. I can see this extending to medibots, maybe, but even then they're usually just sitting still and having people dragged to them, so this nerf is really just damaging people MOVING medbots to locations they're supposed to be. If you want them to not be used to replace jobs, remove the setting to make them not path to different locations.

MajManatee:
naw

Hopekz:
Molti's change was intentional he was targeting the secbot specifically.
You can see the intent if you actually understood the comment you are replacing in the PR.

neeshacark:
exactly what lichton said. medbots... i mean sure? i rarely see anyone moving them around for healing and really its not that good compared to just having your own medkit

there is zero reason for the other bots to have this.  the cleanbot and floorbot are EXTREMELY slow compared to their human counterparts.  in the time it takes for a cleanbot to do one tile, a human can do at least nine. 

 i rarely even SEE floorbots outside of the rare two or three robotics makes and then they fuck off into maint never to be seen again.  just like cleanbots they are significantly slower than humans doing it and nobody in their right mind would use them as a substitute.

Moltijoe:
Balance council: the nerf was specifically for secbots because theyre kinda silly, the rest of the bots arent nearly that much of a balance threat

PR #19104, opened 2023-06-02
gives liquid electricity blood packs a name

ktlwjec0:
I've tested as much as I can, works. Instead of it just being named blood pack it's named blood pack - LE. No more reaching for an empty blood pack and screaming when it's just a prank as it's full of ethereal blood.

:cl:  Ktlwjec
rscadd: Liquid Electricity blood packs are named with the blood type now.
/:cl:


MajManatee:
How much testing is needed for this sorta thing???

Moltijoe (approved):

PR #19105, opened 2023-06-02
fix techwebs runtime

ktlwjec0:
there were two /datum/techweb_node/exotic_ammo

:cl:  Ktlwjec
bugfix: Experimental Ammunition techweb works.
/:cl:


MajManatee (approved):
Your honor, i plead oopsie dasies

PR #19111, opened 2023-06-03
All Heads of Staff can set Blue/Green Alert and Emergency Maint Access

Aquizit:

As stated in title. Also added an error text for if someone attempts to change the alert level while it's on cooldown.

Reason for this is for the no auto-blue at round start, it's needed for rounds without Captains. Emergency Maint is just QoL if no AI or Cap for rad storm/blob/etc.

# Changelog

:cl:  
tweak: Heads of Staff can use Communications Console to change security level and emergency maint access.
/:cl:

### DIFF
--- a/code/game/machinery/computer/communications.dm
+++ b/code/game/machinery/computer/communications.dm
@@ -136,9 +136,6 @@
 			SSshuttle.requestEvac(usr, reason)
 			post_status("shuttle")
 		if ("changeSecurityLevel")
-			if (!authenticated_as_silicon_or_captain(usr))
-				return
-
 			if (!COOLDOWN_FINISHED(src, important_action_cooldown))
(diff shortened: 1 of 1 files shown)


JohnFulpWillard:
Does this remove silicon from being able to use it?

Aquizit:
No they're still able to. I assume since even though I took out the first check of silicon_or_captain there's another separate issilicon check later down the condition checks

TheRyeGuyWhoWillNowDie (changes requested):
Emergency maintenance access is a keycard authenticator action, so I'd suggest getting rid of the keycard authenticator action or not making it available for every heard

Aquizit:
It has always been on the comms console as well. Why is having it in both places bad?

ToasterBiome (approved):

PR #19134, opened 2023-06-06
Makes Hypospray quickload module innate

Moltijoe:
It's basically free from the vendor and all it does is QoL

:cl:  
tweak: Hyposprays have quickload by default
/:cl:


MajManatee (approved):
i honestly have no clue if this works

PR #19135, opened 2023-06-06
Removes some of the stronger chemicals from hypospray vendor

Moltijoe:
Being able to make rather strong hypomixes with just the vendor was reducing the need for chemists to actually do their job 
also, why buy something like perflouro or charcoal when you are also being offered salbutamol or calomel

:cl:  
tweak: Sal acid, Oxandrolone, calomel, salbutamol have all been removed from the hypospray vendor
/:cl:


JohnFulpWillard (approved):
forgot to hit approve first

PR #19146, opened 2023-06-07
Makes healium actually useful

SapphicOverload:

Healium no longer bypasses anti-sleep effects, instead if you're awake it will heal significantly less and cause jittering, disgust, and slowdown. However if you ARE asleep, you get healed faster depending on how cold you are, making it actually worth using in cryo and giving atmos a reason to supply medbay with some. Also, grinding healium crystals will actually make healium.

# Changelog

:cl:  
tweak: healium is actually useful now
tweak: grinding healium crystals makes healium
/:cl:

### DIFF
--- a/code/game/machinery/electrolyzer.dm
+++ b/code/game/machinery/electrolyzer.dm
@@ -95,10 +95,16 @@
 	if(!removed)
 		return
 
-	var/proportion = min(removed.get_moles(/datum/gas/water_vapor), (3 * delta_time * workingPower)) //Works to max 12 moles at a time.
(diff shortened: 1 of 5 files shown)


iamthesilly:
My bad, pushed a commit to the wrong PR, reverted it

MajManatee:
I dont like it being just an off brand cryoxadone

iamthesilly:
> I dont like it being just an off brand cryoxadone

It's not off-brand, it's a direct upgrade that's harder to make.

JohnFulpWillard:
Doesn't this need a wiki update

ToasterBiome (approved):

PR #19151, opened 2023-06-07
Ultra Violence Rework

SapphicOverload:

This PR includes a number of changes intended to make ultra violence more skill based. IPCs with Ultra Violence will die instantly when they reach the damage required to enter hard crit, and taking damage will temporarily reduce the maximum health you can heal to. The speed boost on successful combo has been replaced with a style system.

- SSSTYLE
Style is a measure of how "stylish" you are. It grants a slight speed boost (replacing the speed boost on successful combo) depending on the rank and decays over time. Style increases when using a variety of methods to deal damage, when executing and parrying, and when successfully landing a shot that has ricocheted at least once. Dealing damage with the same method will give less style the more often it is used until it eventually gives no style, incentivizing the use of multiple strategies instead of just one. Taking damage reduces style.

- PARRYING
Clicking in a direction with an empty hand on grab intent will parry projectiles on the adjacent tile in the direction of your click, with a cooldown to prevent spamming it. The projectile needs to be on the adjacent tile at the exact moment of the parry in order to be parried. Successful parries increase style.

- HARD DAMAGE
Taking damage will apply "hard damage", which reduces the maximum health you can heal to. For example, if you have 50 brute damage with 10 hard damage, you will only be able to heal down to 10 brute damage and will not be able to heal the rest until after the hard damage decays, and then you can heal the remaining damage. This makes tanking damage just to heal back no longer a viable strategy, making it more important to avoid incoming damage. Hard damage decays over time.

- OTHER CHANGES
-You will now die instantly when entering hard crit instead of being immune to it, effectively reducing maximum health by almost half
-Removed speed boost on successful combo, replaced with the style speed bonus
-Coin ricochets increase the speed of the projectile, making them slightly more viable
-Adds sleep immunity because they already have stun immunity and sleep is a stun that bypasses that

# Spriting

HUD icons for current style rank
image
image

# Changelog

:cl:  
tweak: adds a whole bunch of new mechanics to ultra violence
tweak: ultra violence now makes you die instantly in hard crit instead of being immune to it
tweak: projectiles ricocheted off of coins are faster
bugfix: makes coin ricochets able to actually hit their intended target when off-center
/:cl:


MajManatee:
this has had more changes made to it than mining has like jesus.

cuackles (approved):

ToasterBiome (approved):

PR #19152, opened 2023-06-07
EMPs do reduced damage to already damaged positronic brains

SapphicOverload:

IPCs will take less brain damage from EMPs if they already have a lot of brain damage already. Also wearing a tin foil hat can prevent the damage now.

# Changelog

:cl:  
tweak: EMPs do reduced damage to already damaged IPC brains
tweak: tinfoil hat protects brain from EMP
/:cl:

### DIFF
--- a/code/modules/mob/living/brain/brain_item.dm
+++ b/code/modules/mob/living/brain/brain_item.dm
@@ -262,7 +262,11 @@
 	if(prob(25))
 		return
 
-	owner.adjustOrganLoss(ORGAN_SLOT_BRAIN, 50/severity)
+	var/obj/item/clothing/head/foilhat/hat = owner.get_item_by_slot(SLOT_HEAD)
+	if(hat && istype(hat))
+		return
+
+	owner.adjustOrganLoss(ORGAN_SLOT_BRAIN, (50/severity) * (maxHealth - damage) / maxHealth)
 	owner.adjust_drugginess(40/severity)
 	switch(severity)
 		if(1)
--- a/code/modules/mob/living/brain/brain_item.dm
+++ b/code/modules/mob/living/brain/brain_item.dm
@@ -262,8 +262,8 @@
 	if(prob(25))
 		return
 
-	var/obj/item/clothing/head/foilhat/hat = owner.get_item_by_slot(SLOT_HEAD)
-	if(hat && istype(hat))
+	var/obj/item/clothing/head/hat = owner.get_item_by_slot(SLOT_HEAD)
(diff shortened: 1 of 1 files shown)


MajManatee:
Why...? Nobody else takes less brain damage when already damaged? Seems like its just an IPC buff.

iamthesilly:
> Seems like its just an IPC buff.

That's the point, IPCs currently die of brain damage from just 4 EMPs which is a bit harsh. So instead of just making it do less brain damage overall, I made it do less damage the more brain damage you already have, so one or two EMPs will be just as effective against you but spamming it doesn't just kill you instantly.

Rdrazga:
Me waiting to laugh at all the people who will see this pr and put on tinfoil hats not knowing what they do

code/modules/mob/living/brain/brain_item.dm
```diff
@@ -262,7 +262,11 @@
 	if(prob(25))
 		return
 
-	owner.adjustOrganLoss(ORGAN_SLOT_BRAIN, 50/severity)
+	var/obj/item/clothing/head/foilhat/hat = owner.get_item_by_slot(SLOT_HEAD)
```
ToasterBiome: type stuffing is GROSS

iamthesilly: what does this mean

Moltijoe: probably that you're assigning the hat as a foilhat no matter what the hat actually is

iamthesilly: i mean there's no functional difference between doing this and using istype() and just assigning it as a normal hat and then checking if it's the foilhat type right?

iamthesilly: changed it i guess

ToasterBiome:
Why is this good for the game?

iamthesilly:
> Why is this good for the game?

Instakilling IPCs by spamming 4 EMPs at once isn't good for the game

ToasterBiome:
> > Why is this good for the game?
> 
> Instakilling IPCs by spamming 4 EMPs at once isn't good for the game

Explain further please

iamthesilly:
> Explain further please

The problem is EMPs cover a large area, go through walls, and can't be blocked unless you have a specific implant only available in the uplink, and if you EMP 4 times in the vicinity of an IPC, they die instantly without any way to counter it without being a specific antagonist with a specific item.

I don't see how this should be that severe considering the multitude of other effects EMPs already have on IPCs, including actual burn damage and a knockdown/stun.

Moltijoe (approved):

PR #19158, opened 2023-06-07
Hypospray QoL fixes

SapphicOverload:

Hypospray quickload actually swaps the vials properly instead of dropping to the ground, and trying to load one on harm intent no longer deletes the reagents in the vial

# Changelog

:cl:  
bugfix: hypospray quickload actually swaps the vials instead of dropping one onto the floor
bugfix: trying to load a hypospray on harm intent no longer deletes the vial's contents
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/glass.dm
+++ b/code/modules/reagents/reagent_containers/glass.dm
@@ -84,7 +84,7 @@
 		to_chat(user, span_notice("You fill [src] with [trans] unit\s of the contents of [target]."))
 
 	else if(reagents.total_volume)
-		if(user.a_intent == INTENT_HARM)
+		if(user.a_intent == INTENT_HARM && !istype(target, /obj/item/hypospray))
 			user.visible_message(span_danger("[user] splashes the contents of [src] onto [target]!"), \
 								span_notice("You splash the contents of [src] onto [target]."))
 			reagents.reaction(target, TOUCH)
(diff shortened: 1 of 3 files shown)


code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -84,7 +84,7 @@
 		to_chat(user, span_notice("You fill [src] with [trans] unit\s of the contents of [target]."))
 
 	else if(reagents.total_volume)
-		if(user.a_intent == INTENT_HARM)
+		if(user.a_intent == INTENT_HARM && !istype(target, /obj/item/hypospray))
```
JohnFulpWillard: What is this meant to accomplish? If hyposprays should be fillable by beakers, they should be passing ``is_refillable`` and end here. If they aren't, then maybe the proper fix would be to make sure you can't splash contents onto things in your inventory?

iamthesilly: I didn't want to entirely remove the ability to splash reagents onto an item just to fix hyposprays, so I made an exception because they're the only item I'm aware of that you load with reagent containers

JohnFulpWillard: Then should they not be passing is_refillable and therefore never get to this point?

iamthesilly: That's the thing, they don't because hyposprays aren't directly refillable with reagents

JohnFulpWillard: I see what you mean, but this is not the proper fix for it. There's also a var on reagent containers ``spillable`` that you can use to prevent them from spiling vials at all, since the intent for it seems to be to load it up in a hypospray only.

iamthesilly: That works too I guess, I wasn't intending to make hypospray vials never able to splash their contents on anything but it's not that big of a deal.

JohnFulpWillard (approved):

PR #19165, opened 2023-06-08
Removes the non-bitfield slot flags

JohnFulpWillard:

Replaces instances of non-bitfield slot flags with the proper bitfield instead.

Essentially, this is a port of https://github.com/tgstation/tgstation/pull/47742

This PR is important because non-bitfield slot flags are a pain to deal with for contributors. You have to choose one to use, appropriately translate it to the bitflag if necessary, and also if it breaks it's hard to tell if it's because it's comparing the bitfield to the non-bitfield. This tech debt is not worth keeping around for something that gives no benefits whatsoever.

This PR will conflict with my ComponentInitialize one, but I'd prefer if that one was merged first.

cool screenshots
image
image

(gorilla)
image

image

# Changelog

No player-facing changes.


Moltijoe (approved):
always sorta wondered why we had both

PR #19168, opened 2023-06-08
Removes the desword from the traitor uplink (again)

Mqiib:
Desword is now nukie-only because it was REMOVED FOR A REASON and we had to let it reappear to reinforce this idea.

# Changelog

:cl:  

rscdel: Desword removed from the traitor uplink (again)
experimental: Told you so.

/:cl:


Mqiib:
fuck

Moja10:
I'm still fine with it and glad it's back. If you are hellbent on this, I would rather nukie desword and traitor desword split and the traitor one nerfed. I don't support it's removal at all

Moltijoe:
> I'm still fine with it and glad it's back. If you are hellbent on this, I would rather nukie desword and traitor desword split and the traitor one nerfed. I don't support it's removal at all

that can create confusion between which does what when there's two items that look similar but have wildly different power levels

PR #19171, opened 2023-06-08
Ignore that last PR, i'm merging adrenaline together with determination instead

Moltijoe:
# reasoning for merging the both

They're two very similar effects, an effect that kicks in while in combat in an attempt to draw the combat out longer
Problem is there's not really a visual indicator for adrenaline so you can never tell
You probably thought determination gave the damage slowdown resistance, nope, it's adrenaline

Adrenaline is the one being removed as it risks sharing it's name with adrenals (which is something else entirely)

The main balance point is that this means getting wounds will give slowdown resistance
and you'll get minor healing to your wounded limbs sometimes even if you didn't just get a new wound

# reasoning for moving "adrenaline" off of the heart
When redmoogle added adrenaline as an interesting way to make fights not end instantly upon taking enough damage to enter slowdown, they put it directly on the heart organ.
While that works, it also means any species without a heart, or someone that otherwise lost their heart without dying simply won't benefit from the mechanic

This change will allow people with vein muscle membrane + no heart to still have adrenaline
as well as plasmamen who have a different organ in place of their heart

however, as adrenaline requires organic reagent processing, IPCs will still not benefit from adrenaline

:cl:  
tweak: Adrenaline no longer requires a heart to generate
tweak: Adrenaline and determination are combined
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -18,6 +18,12 @@
 #define THERMAL_PROTECTION_HAND_LEFT	0.025
 #define THERMAL_PROTECTION_HAND_RIGHT	0.025
 
+#define DETERMINATION_THRESHOLD 25
+
+/mob/living/carbon/human
+	var/lasthealth
+	COOLDOWN_DECLARE(determination_cooldown)
+
 /mob/living/carbon/human/Life(times_fired)
 	set invisibility = 0
 	if (notransform)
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -18,6 +18,12 @@
 #define THERMAL_PROTECTION_HAND_LEFT	0.025
 #define THERMAL_PROTECTION_HAND_RIGHT	0.025
 
+#define DETERMINATION_THRESHOLD 25
+
+/mob/living/carbon/human
+	var/lasthealth
+	COOLDOWN_DECLARE(determination_cooldown)
```
tattax, line 25: why are you putting this here

tattax: also auto doc this

Moltijoe: > why are you putting this here

there's no function difference as far as i know, and this keeps important information together

>also autodoc this

what?

PR #19173, opened 2023-06-08
Possibly fixes dusting leaving invisible corpses

Moltijoe:
fucked if i know, i couldn't re-create the bug no matter what form of dusting i used
so i tweaked it so the QDEL_IN part is called before the dust animation to hopefully fix any runtime breaking the proc and not calling QDEL_IN

:cl:  
bugfix: Dusting leaving invisible corpses might be fixed now
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/death.dm
+++ b/code/modules/mob/living/carbon/human/death.dm
@@ -6,17 +6,19 @@
 			new /obj/effect/temp_visual/gib_animation(loc, "gibbed-r")
 
 /mob/living/carbon/human/dust(just_ash, drop_items, force)
+	death(TRUE)
+
 	if(drop_items)
 		unequip_everything()
 
 	if(buckled)
 		buckled.unbuckle_mob(src, force = TRUE)
 
-	Stun(100, TRUE, TRUE)//hold them still as they get deleted so they don't fuck up the animation
 	notransform = TRUE
+
 	dust_animation()
-	spawn_dust(just_ash)
-	QDEL_IN(src, 20) // since this is sometimes called in
+	QDEL_IN(src, 2 SECONDS) // since this is sometimes called in
+	addtimer(CALLBACK(src, PROC_REF(spawn_dust), just_ash), 1.5 SECONDS)//spawn the dust after the body disappears
 
 /mob/living/carbon/human/dust_animation()
 	var/obj/effect/dusting_anim/dust_effect = new(loc, ref(src))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/death.dm
```diff
@@ -12,11 +12,11 @@
 	if(buckled)
 		buckled.unbuckle_mob(src, force = TRUE)
 
-	Stun(100, TRUE, TRUE)//hold them still as they get deleted so they don't fuck up the animation
 	notransform = TRUE
```
JohnFulpWillard, line 15: Why are we setting notransform?
Would it not be better to call ``death()`` on the mob at the start of the proc?

code/modules/mob/living/carbon/human/death.dm
```diff
@@ -12,11 +12,11 @@
...
+
+	QDEL_IN(src, 2 SECONDS) // since this is sometimes called in
 	dust_animation()
-	spawn_dust(just_ash)
-	QDEL_IN(src, 20) // since this is sometimes called in
+	addtimer(CALLBACK(src, PROC_REF(spawn_dust), just_ash), 1.5 SECONDS)//spawn the dust after the body disappears
```
JohnFulpWillard: This is adding a timer to something that is expected to be qdel'ed, not a good idea.
You should keep how it was earlier, I don't think it was the problem here.

code/modules/mob/living/carbon/human/death.dm
Moltijoe: because it fucks with the animation if they die immediately, rather than standing normally and fading away, they slowly fall over while fading away

code/modules/mob/living/carbon/human/death.dm
```diff
@@ -12,11 +12,11 @@
 	if(buckled)
 		buckled.unbuckle_mob(src, force = TRUE)
 
-	Stun(100, TRUE, TRUE)//hold them still as they get deleted so they don't fuck up the animation
 	notransform = TRUE
+
+	QDEL_IN(src, 2 SECONDS) // since this is sometimes called in
 	dust_animation()
 	spawn_dust(just_ash)
-	QDEL_IN(src, 20) // since this is sometimes called in
```
JohnFulpWillard, line 19: qdel in should still be the last thing to go

Moltijoe: Well then i have no clue what might be breaking it and theres no reason for this PR

PR #19178, opened 2023-06-08
Makes an antag only version of the reviver implant

Moltijoe:
People were talking about the reviver implant, i noticed it was in the uplink, which looks kinda sad compared to other items

Similar to the CNS rebooter, makes an uplink only version that might actually be worth it's cost in the uplink
Main difference is that it heals twice as much
Also removes the heart attack emp effect from this version because its pretty damning

:cl:  
rscadd: Adds a syndicate reviver implant that replaces the regular one in the uplink
/:cl:


MajManatee (approved):
looks fine probably

DylonBR:
God damn 8 whole TC. You telling me this is as useful as an esword?

PR #19199, opened 2023-06-09
Suit sensors update slower.

Code611:
Suit sensors have been the talk of a slowdown for quite a while now, so this is not coming from an ided perspective. 
WHICH IT IS! However, it's coming from many ideds and not a spur-of-the-moment rage ided, but I'm not going to lie; I've certainly had experience with the main issues of suit sensors.

**Keep in mind that I don't see 15 seconds as the final value, and I'd like both medical and non-medical players to have input on duration.**

glort
In the image, you have
- 2 Paramedics
- 1 Brig Phys
- 2 Security
- 1 random artist (???)

However, the issue isn't going to be sucking job slots from medbay; the issue is that mainly this situation happened within 10 seconds of me doing anything that wasn't EMP -> High damage/STUN combo. 
Mainly suit sensors have become problematic due to Yogstation's active medbay player base. It'd be silly to balance antags around dealing with 5+ sec ganks, paramedics, 1-2 valid hunters, and any medical staff that's bored enough, so generally, it's best to slow down suit sensors.
That being said, too high of duration would make suit sensors useless, which I wish to avoid.
__It takes around 30 seconds to navigate from escape to arrivals__; additionally, since medbay is a bit closer to the "center" of most maps, like the most played map(box), response times are quicker. 
I believe this creates an antag loop where you must instantly or nigh instantly round remove someone or be locked to a boring playstyle of dropping an EMP into a slowdown.

While *YES* you could argue that "just get emps bro; everybody has a department lathe,"  this goes back to a boring gameplay loop of huge dmg dumps before you get detected and trigger three paramedics spawning eight tiles away. 

On an ided level, I want this to give a bit more breathing room to antags.
On a gameplay level, I want this to allow antags to do things other than boring combos by at least slowing down the rate paramedics/sec will fly to your location.
ON **all.** levels I want this to be discussed to ensure suit sensors don't become useless while still preventing silent round-removals.

# Wiki Documentation
Suit sensors update every 15 seconds if we even note that. We probably don't.
# Changelog

:cl:  
tweak: suit sensors update every 15 seconds instead of 10.
/:cl:

### DIFF
--- a/code/game/machinery/computer/crew.dm
+++ b/code/game/machinery/computer/crew.dm
@@ -1,4 +1,4 @@
-#define SENSORS_UPDATE_PERIOD 100 //How often the sensor data updates.
+#define SENSORS_UPDATE_PERIOD 150 //How often the sensor data updates.
 
 /obj/machinery/computer/crew
 	name = "crew monitoring console"


SauceCat9:
I think 20 seconds would actually be better, 15 seconds still feels too short to kill and strip someone without the medical swarm noticing.

ToasterBiome:
I think 30 seconds would actually be better.

Moja10:
The medical swarm gets ridiculous because if you killed someone, and now you have the entire medbay team including CMO on your ass, you are basically left with the choices "let them have the body + the body tattles on you = arrested/ded" or "Blast every single one of them back to hell with a weapon and then 1. get in trouble with admins potentially because le invalid 2. they go gamer and now have justification to stab you back or 3. this scenario hasn't involved sec team presence yet but there's usually 3 officers behind the medical brigade too." Suit sensors have been wack since the TGUI update on them.

Code611:
This is one of the main issues I hope to address in this. As I said, we can't really balance antags to deal with medical swarms, and I'm not going on a holy crusade to lower paramedic job slots (see how that went with lowering MD slots).  All I believe can be done is add ways to AVOID the gank squad from pulling up on you and putting you in checkmate of "die" and "get banned."

I see most inputs say to raise it more, and if possible when the balance council is reviewing this I'd like for them to convene on a valid time for it to be. I don't want them approving 15 seconds when most players want it to be around 20-30. 

> The medical swarm gets ridiculous because if you killed someone, and now you have the entire medbay team including CMO on your ass, you are basically left with the choices "let them have the body + the body tattles on you = arrested/ded" or "Blast every single one of them back to hell with a weapon and then 1. get in trouble with admins potentially because le invalid 2. they go gamer and now have justification to stab you back or 3. this scenario hasn't involved sec team presence yet but there's usually 3 officers behind the medical brigade too." Suit sensors have been wack since the TGUI update on them.

Moltijoe:
Balance council:
>Because of the nature of how suit sensors refresh, all this does is create annoyance. In order to use it effectively as a traitor, you'd need to watch suit sensors to know when the update happens, otherwise they could update after 5 seconds of starting the engagement. A better direction would be to add more ways to combat suit sensors

a personal idea i have is an item that lets you spoof someone's suit sensors signal

PR #19203, opened 2023-06-10
Adds grep for mapload and var in Args

JohnFulpWillard:

Adds CI checks for ``mapload`` missing in Initialize args, and one for ``var/`` being in them, both of which should not be the case.

This will fail CI because I did not fix any of the instances in code, I'm posting this now to prove they work.

Regexes used

vars in Args
``\/(datum|obj|mob|turf|area|atom).+\((var).*\)``

# Changelog

Nothing player-facing.


JohnFulpWillard:
image
we win

ToasterBiome (approved):

PR #19208, opened 2023-06-10
Brig Phys spawns with a medical belt

Moltijoe:
Just like mining medic, they don't have one just lying in their room like how it is for the medbay jobs, 
however, unlike mining medic, they don't spawn with one so they're forced to use their "spare" as their first one

:cl:  
rscadd: Brig physician spawns with an empty medical belt worn
/:cl:


MajManatee (approved):
neat

PR #19218, opened 2023-06-10
Holy Light rod now provides favor upon healing.

Code611:
Never made sense to me that the sect about **healing people** can get a thing that **heals people** and also **reduces your favor** because you wanted to **heal people** that wasn't slapping them with a bible 50 times.

I don't really think the chap should become worse at gaining favor if they want to help the general public by giving them a rod.

My only worry is that if the favor gain is too much (which is rare, but could happen) you could see rods being spammed, but if this is a large concern I can just make the rod favor less and less the more exist.
# Wiki Documentation
Rod of White should be noted that it gives favor equal to what it heals.

# Changelog

:cl:  
tweak: Rod of White now provides favor when healing others.
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -482,7 +482,6 @@
 			else if(isanimal(L))
 				var/mob/living/simple_animal/SM = L
 				SM.adjustHealth(-3.5 * efficiency, forced = TRUE)
-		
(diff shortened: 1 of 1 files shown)


Code611:
oh its the fucking antag rebase. I'll reopen this when I unfuck it.

PR #19223, opened 2023-06-10
Holy Light rod now gives favor upon healing.

Code611:
Never made sense to me that the sect about **healing people** can get a thing that **heals people** and also **reduces your favor** because you wanted to **heal people** that wasn't slapping them with a bible 50 times.

I don't really think the chap should become worse at gaining favor if they want to help the general public by giving them a rod.

My only worry is that if the favor gain is too much (which is rare, but could happen) you could see rods being spammed, but if this is a large concern I can just make the rod favor less and less the more exist.

# Wiki Documentation
Rod of White should be noted that it gives favor equal to what it heals.

# Changelog

:cl:  
tweak: Rod of White now provides favor (at a 20% conversion) when healing others.
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs/buffs.dm
+++ b/code/datums/status_effects/buffs/buffs.dm
@@ -463,6 +463,9 @@
 			itemUser.adjustCloneLoss(-0.5 * efficiency) //Becasue apparently clone damage is the bastion of all health
 		//Heal all those around you, unbiased
 		for(var/mob/living/L in view(7, owner))
+			if(ispath(rod_type, /obj/item/rod_of_asclepius/white)) //Used for adjusting the Holy Light Sect Favor from white rod healing.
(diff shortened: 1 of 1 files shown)


tattax:
i really don't like this playstyle that the white light has introduced of being rewarded for doing nothing, most rounds you'll just go around to the chapel and no one is there except the altar with the white light sect selected. Chaplain should be atleast required to have minimal effort to acquire favor than just being a poststand

Code611:
> i really don't like this playstyle that the white light has introduced of being rewarded for doing nothing, most rounds you'll just go around to the chapel and no one is there except the altar with the white light sect selected. Chaplain should be atleast required to have minimal effort to acquire favor than just being a poststand

Discussed. I'm going to lower the point gain to 20% so it doesn't contest with "active" healing methods such as the bible.

Moltijoe (approved):

PR #19228, opened 2023-06-10
modular holotools :3

TheRyeGuyWhoWillNowDie:
the holotool is cool and stuff, so this PR basically hypospray's it

starting at industrial engineering, you can print holotools. however, holotools now need tool cards in order to use specific tools. the first holotool unlocked is super shitty, with only a capacity of 5, and all the cards at that level are 25% slower than normal tools.

with high efficiency parts, the advanced holotool is unlocked with 10 capacity and advanced tool cards, which work at the same speed as normal tools.

with mini bluespace, elite holotools are unlocked, with 15 capacity and elite tool cards, which only take 80% as long to complete tasks

the RD's holotool is unaffected, with special RD cards that have the same tool modifier, so nothing is different

cards can be taken out of the holotools with a screwdriver

The steal objective is still the RD's specific holotool, no other holotools will work

# Wiki Documentation

basic cards and holotool unlocked with industrial engineering, tools have 1.25 speed modifier
advanced cards and holotool unlocked with high efficiency parts, tools have 1 speed modifier
elite cards and holotool unlocked with minibluespace, tools have 0.4 speed modifier

# Changelog

:cl:  
rscadd: modular holotools :3
/:cl:

### DIFF
--- a/yogstation/code/game/objects/items/holotool/holotool.dm
+++ b/yogstation/code/game/objects/items/holotool/holotool.dm
@@ -1,6 +1,6 @@
 /obj/item/holotool
-	name = "experimental holotool"
-	desc = "A highly experimental holographic tool projector."
+	name = "basic holotool"
+	desc = "A device capable of holographically projecting various tools. Fit with upgrade cards to access more tools."
 	icon = 'yogstation/icons/obj/holotool.dmi'
 	icon_state = "holotool"
 	slot_flags = ITEM_SLOT_BELT
@@ -8,6 +8,8 @@
 	usesound = 'sound/items/pshoom.ogg'
 	lefthand_file = 'yogstation/icons/mob/inhands/lefthand.dmi'
 	righthand_file = 'yogstation/icons/mob/inhands/righthand.dmi'
+	var/maximum_toolage = 5
+	var/current_toolage = 0
 	actions_types = list(/datum/action/item_action/change_tool, /datum/action/item_action/change_ht_color)
 	resistance_flags = FIRE_PROOF | ACID_PROOF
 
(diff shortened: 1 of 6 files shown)


yogstation/code/game/objects/items/holotool/holotool.dm
```diff
@@ -17,11 +19,60 @@
...
+	new /obj/item/holotool_module/screwdriver/rd(src)
+	new /obj/item/holotool_module/crowbar/rd(src)
+	new /obj/item/holotool_module/multitool/rd(src)
+	new /obj/item/holotool_module/wrench/rd(src)
+	new /obj/item/holotool_module/snips/rd(src)
+	new /obj/item/holotool_module/welder/rd(src)
+	current_toolage = 12 //I can count :D
+
+/obj/item/holotool/attackby(obj/item/I, mob/U)
+	if(istype(I, /obj/item/holotool_module))
```
tattax, line 56: when the U uses the I on the /obj/item/holotool

yogstation/code/game/objects/items/holotool/holotool.dm
```diff
@@ -17,11 +19,60 @@
...
+		var/obj/item/holotool_module/card = I
+		if(current_toolage + card.toolage_use > maximum_toolage)
+			to_chat(U, span_warning("The " + src.name + " can't hold any more module cards!"))
```
tattax, line 59:
```suggestion
			to_chat(U, span_warning("\The [src] can't hold any more module cards!"))
```

ynot01:
i do not see the purpose of this feature? i do not see why physical tools have to compete with these printable tiny sized toolbelts

iamthesilly:
Why do we need to be able to print pocket-sized toolbelts?

TheRyeGuyWhoWillNowDie:
> Why do we need to be able to print pocket-sized toolbelts?

because sometimes it's useful for people to have tools

iamthesilly:
> because sometimes it's useful for people to have tools

We had toolbelts/backpacks/pockets for that, use those to put your tools in

PibeCantina:
I mean, i like this but why make normal tools at all when i can have this? I think this should be much slower than normal tools if you want them to work together

wonderinghost:
Why the fuck are you turning science into power rangers

DylonBR:
What next? Engineering getting magboot upgrades? Feature bloat

TheRyeGuyWhoWillNowDie:
> What next? Engineering getting magboot upgrades?

Sure

Moltijoe:
Balance council:

>There isn't much reason for this to exist, it just makes the RD's holotool less cool without serving an interesting function. It's effectively just a printable toolbelt, but the jobs that need a toolbelt already have one.

RG4ORDR:
Toolset arm implant exists for this exact purple.

Moltijoe:
was waiting on 24h previously
i apologize gus, as it looks like you've spent a good amount of time on this

PR #19233, opened 2023-06-11
Jaws of life tweaks

SapphicOverload:

Jaws of life are now normal-sized and slower to use for non-engineers. Also fixes a bug that prevents you from using it on a shocked door even with insulated gloves on.

# Changelog

:cl:  
tweak: jaws of life are normal-sized
tweak: non-engineers take longer to pry open doors with jaws of life
bugfix: fixes not being able to pry open shocked doors with insulated gloves on
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -1269,8 +1269,8 @@
 
 	if(istype(I, /obj/item/jawsoflife) || istype(I, /obj/item/mantis/blade))
 		if(isElectrified())
-			shock(user,100)//it's like sticking a fork in a power socket
-			return
+			if(shock(user,100))//it's like sticking a fork in a power socket
+				return
 
 		if(istype(I, /obj/item/mantis/blade))
 			var/obj/item/mantis/blade/secondsword = user.get_inactive_held_item()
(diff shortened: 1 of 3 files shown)


code/game/machinery/doors/airlock.dm
```diff
@@ -1313,9 +1313,9 @@
 
 				playsound(src, 'sound/machines/airlock_alien_prying.ogg', 100, TRUE) //is it aliens or just the CE being a dick?
 				prying_so_hard = TRUE
-				if(do_after(user, time_to_open, src))
+				if(I.use_tool(src, user, time_to_open))
 					open(2)
-					if(!istype(I,/obj/item/jawsoflife/jimmy)) //You get to be special
+					if(!(istype(I,/obj/item/jawsoflife/jimmy) || IS_ENGINEERING(user))) //You get to be special
```
JohnFulpWillard: Why is this better for the game?

code/game/machinery/doors/airlock.dm
```diff
@@ -1293,7 +1293,7 @@
 			to_chat(user, span_warning("The airlock won't budge!"))
 			return
 
-		var/time_to_open = 7 SECONDS * I.toolspeed
+		var/time_to_open = 9 SECONDS
```
JohnFulpWillard, line 1296: Why do we not want toolspeed to be taken into account?

iamthesilly: It's already taken into account in use_tool() which it now uses instead of do_after()

UltraBigRat:
while a jaws nerf is fine, not a fan of items working differently for specific jobs

JohnFulpWillard (approved):

PR #19239, opened 2023-06-12
Traitors can no longer deflect energy projectiles with the desword

ynot01:

Only nukies can deflect energy projectiles because they are true masters of the sord

The desword is getting removed if this isn't merged I don't make the rules

# Changelog

:cl:  
tweak: Only nukies may utilize the double-bladed energy sword to deflect energy projectiles now
/:cl:

### DIFF
--- a/code/game/objects/items/twohanded.dm
+++ b/code/game/objects/items/twohanded.dm
@@ -531,7 +531,8 @@
 
 /obj/item/twohanded/dualsaber/IsReflect()
 	if(wielded && isliving(loc))
-		return is_nukeop(loc)
+		var/mob/living/holding_mob = loc
+		return is_nukeop(holding_mob)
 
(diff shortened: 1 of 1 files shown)


Marmio64:
it definitely needs a price decrease if half of it's appeal is getting removed for traitors

ToasterBiome:
> The desword is getting removed if this isn't merged I don't make the rules

says who

ynot01:
> says who

the shadow council

MajManatee:
but why is this a good change (it isnt)

Moja10:
would rather have reflection than 300 morbillion damage, security can deal with a traitor who is reflecting shit but not also doing 4 hit kills in the process

Moltijoe:
Ynot knows why it's denied, but for everyone else:
>This isn't the way we should go about handling the desword being a problem

PR #19241, opened 2023-06-12
Changes suit sensor EMP

Moltijoe:
EMPing suit sensors is currently so random that it's unreliable as an actual stealth tactic
This changes it to be a bit more deterministic
it will randomly pick a mode that is less precise than the currently selected mode

:cl:  
tweak: suit sensors EMP change (less random)
/:cl:


MajManatee (approved):
I like this one.

PR #19242, opened 2023-06-12
Adds a makeshift desword

Moltijoe:
Clearly some people want the desword and others don't
Rather than balancing the existing desword so it works different depending on who holds it, adds a new type of desword that is specifically for traitors

only has a 50% chance to deflect because it's both incredibly powerful at 100%, but also too iconic to be removed outright
30 force instead of 34 since it's literally just two eswords duct taped together
should probably get it's own sprite at some point, but i'm a coder, not a spriter

:cl:  
rscadd: Added new makeshift desword that can be made by using two eswords together
rscdel: Removed desword from the traitor uplink
/:cl:

### DIFF
--- a/code/datums/components/crafting/antag.dm
+++ b/code/datums/components/crafting/antag.dm
@@ -4,6 +4,17 @@
 ///////////////////////////////////////////////////
 
 // Weapons
+/datum/crafting_recipe/dualsaber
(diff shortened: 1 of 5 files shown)


MajManatee:
Removes the whole point of the sword

MajManatee (approved):
poggers not 0% deflection

PR #19246, opened 2023-06-12
Kills off /obj/item/onetankbomb

SapphicOverload:

This moves the singletank bomb code from /obj/item/onetankbomb to /obj/item/tank, allowing the assembled bomb to retain more of the functions and properties of the tank it was assembled from. Also lets you use hydrogen and tritium as fuel.

# Changelog

:cl:  
bugfix: assembled single-tank bombs retain the functions and properties of the tank used to make it (until it explodes)
tweak: hydrogen and tritium work in single-tank bombs
/:cl:


DylonBR:
how much trit can a tank hold at 400 C? will we finally have maxcap singletanks?

iamthesilly:
> how much trit can a tank hold at 400 C? will we finally have maxcap singletanks?

The same as plasma, since all gases here have the same pressure at equal temperatures.

DylonBR:
but the explosion size is based off how much strength right? and more moles means more strength right? can't you fit more moles of trit than plasma in a tank?

iamthesilly:
> but the explosion size is based off how much strength right? and more moles means more strength right? can't you fit more moles of trit than plasma in a tank?

No, all gases take up the same amount of space in this atmospheric system.

DylonBR:
Maybe make antinoblium or something be stronger? it would be cool to have something be stronger than plasma for singletank explosions.

code/game/objects/items/tanks/tanks.dm
```diff
@@ -300,3 +309,160 @@
...
+	if(QDELETED(src))
+		return
+	if(status)
+		ignite()	//if its not a dud, boom (or not boom if you made shitty mix) the ignite proc is below, in this file
+	else
+		release()
+
+//Assembly / attached device memes
+
+/obj/item/tank/Crossed(atom/movable/AM as mob|obj) //for mousetraps
```
JohnFulpWillard: ``as mob|obj`` is invalid. ``/atom/movable`` is already only mob or obj anyway.

JohnFulpWillard (commented):
I'm concerned on what the balance changes to this could be. Wouldn't this mean that any ol-tank could be boobytrapped?

iamthesilly:
> I'm concerned on what the balance changes to this could be. Wouldn't this mean that any ol-tank could be boobytrapped?

You could already do that, and it also still shows the assembly overlay on the tank so you know it's probably a bomb

code/game/objects/items/tanks/tanks.dm
iamthesilly: Done

code/game/objects/items/tanks/tanks.dm
```diff
@@ -81,6 +85,11 @@
 	if(air_contents)
 		qdel(air_contents)
 
+	if(tank_assembly)
+		tank_assembly.forceMove(drop_location())
```
JohnFulpWillard: Should we be deleting it instead?

code/game/objects/items/tanks/tanks.dm
```diff
@@ -300,3 +309,160 @@
...
+
+/obj/item/tank/update_icon()
+	cut_overlays()
+	if(tank_assembly)
+		add_overlay(tank_assembly.icon_state)
+		copy_overlays(tank_assembly)
+		add_overlay("bomb_assembly")
+
+/obj/item/tank/wrench_act(mob/living/user, obj/item/I)
+	to_chat(user, span_notice("You disassemble the bomb!"))
```
JohnFulpWillard: this would likely be better as a balloon alert

code/game/objects/items/tanks/tanks.dm
```diff
@@ -300,3 +309,160 @@
...
+		tank_assembly = null
+	throw_speed = initial(throw_speed)
+	status = FALSE
+	update_icon()
+	return TRUE
+
+/obj/item/tank/welder_act(mob/living/user, obj/item/I)
+	. = FALSE
+	if(status)
+		to_chat(user, span_notice("[src] already has a pressure hole!"))
```
JohnFulpWillard: Around here as well, balloon alerts would be better

code/game/objects/items/tanks/tanks.dm
```diff
@@ -300,3 +309,160 @@
...
+		ignite()	//if its not a dud, boom (or not boom if you made shitty mix) the ignite proc is below, in this file
+	else
+		release()
+
+//Assembly / attached device memes
+
+/obj/item/tank/Crossed(atom/movable/AM) //for mousetraps
+	..()
+	if(tank_assembly)
+		tank_assembly.Crossed(AM)
```
JohnFulpWillard, line 360: All of these can be signals that ``tank_assembly`` will register onto the tank.

code/game/objects/items/tanks/tanks.dm
```diff
@@ -23,6 +23,10 @@
 	var/volume = 70
 	/// Mob that is currently breathing from the tank.
 	var/mob/living/carbon/breathing_mob = null
+	/// Attached igniter assembly, used to ignite its contents
+	var/obj/item/assembly_holder/tank_assembly = null
+	/// Whether or not it can explode when it receives a signal
+	var/status = FALSE
```
JohnFulpWillard: This needs a more descriptive var name for something meant to represent something so destructive.

code/game/objects/items/tanks/tanks.dm
```diff
@@ -23,6 +23,10 @@
 	var/volume = 70
 	/// Mob that is currently breathing from the tank.
 	var/mob/living/carbon/breathing_mob = null
+	/// Attached igniter assembly, used to ignite its contents
+	var/obj/item/assembly_holder/tank_assembly = null
```
JohnFulpWillard: This starts as null by default, you don't need to declare it.

code/__DEFINES/assembly.dm
```diff
@@ -0,0 +1,6 @@
+#define WIRE_RECEIVE		(1<<0)
+#define WIRE_PULSE			(1<<1)
+#define WIRE_PULSE_SPECIAL	(1<<2)
+#define WIRE_RADIO_RECEIVE	(1<<3)
+#define WIRE_RADIO_PULSE	(1<<4)
```
JohnFulpWillard:
```suggestion
#define WIRE_RECEIVE (1<<0)
#define WIRE_PULSE (1<<1)
#define WIRE_PULSE_SPECIAL (1<<2)
#define WIRE_RADIO_RECEIVE (1<<3)
#define WIRE_RADIO_PULSE (1<<4)
```

code/game/objects/items/tanks/tanks.dm
iamthesilly: I'll see what I can do

iamthesilly: Having some issues getting it to work with signals (also there's no signal for on_found apparently) so hopefully it's not too big of an issue to keep it the way it is

JohnFulpWillard (approved):

PR #19253, opened 2023-06-12
Fixes polysmorph preferences tgui 

Mqiib:
Should make the gender option not show for species with locked genders (namely polysmorph).
Doesn't mess with currently selected preferences so TECHNICALLY the preview will show male polysmorphs if someone already had it set to that but I doubt anyone will notice because they're skinny anyways and mostly visually identical.

TO DO:
- [x] Actually works
- [ ] Maybe fix the preferences thing but I really don't want to cause a problem and fuck over everyone's characters by accident for such a small bug that only might appear and only might be noticed

# Changelog

:cl:  Chubbygummibear

bugfix: TGUI character preferences should now respect forced genders of species

/:cl:

### DIFF
--- a/code/modules/client/preferences/gender.dm
+++ b/code/modules/client/preferences/gender.dm
@@ -7,6 +7,20 @@
 /datum/preference/choiced/gender/init_possible_values()
 	return list(MALE, FEMALE, PLURAL)
 
+/datum/preference/choiced/gender/create_informed_default_value(datum/preferences/preferences)
+
+	var/species_type = preferences.read_preference(/datum/preference/choiced/species)
+	var/datum/species/S = new species_type
+	
+	if(!S.sexes || (AGENDER in S.species_traits))
+		return list(PLURAL)
+	else if(FGENDER in S.species_traits)
+		return list(FEMALE)
+	else if(MGENDER in S.species_traits)
+		return list(MALE)
+
+	return list(MALE, FEMALE, PLURAL)
+
 /datum/preference/choiced/gender/apply_to_human(mob/living/carbon/human/target, value)
 	var/datum/species/S = target.dna.species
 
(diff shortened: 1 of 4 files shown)


UltraBigRat:
does this remove agender polys too?

Mqiib:
> does this remove agender polys too?

yes

UltraBigRat:
> > does this remove agender polys too?
> 
> yes

Why not keep in they/them polys though?

Mqiib:
> > > does this remove agender polys too?
> > 
> > 
> > yes
> 
> Why not keep in they/them polys though?

Because doing that would require even MORE work trying to figure out this horrible awful bad tgui system. I barely pulled this off as-is.

You're welcome to do it yourself if you feel strongly about it.

UltraBigRat:
> > > > does this remove agender polys too?
> > > 
> > > 
> > > yes
> > 
> > 
> > Why not keep in they/them polys though?
> 
> Because doing that would require even MORE work trying to figure out this horrible awful bad tgui system. I barely pulled this off as-is.
> 
> You're welcome to do it yourself if you feel strongly about it.

fair nuff, byond is hell

JohnFulpWillard (approved):

code/modules/client/preferences/gender.dm
```diff
@@ -7,6 +7,20 @@
...
+/datum/preference/choiced/gender/create_informed_default_value(datum/preferences/preferences)
+
+	var/species_type = preferences.read_preference(/datum/preference/choiced/species)
+	var/datum/species/S = new species_type
+	
+	if(!S.sexes || (AGENDER in S.species_traits))
+		return list(PLURAL)
+	else if(FGENDER in S.species_traits)
+		return list(FEMALE)
+	else if(MGENDER in S.species_traits)
```
JohnFulpWillard:
```suggestion
	var/datum/species/species_type = preferences.read_preference(/datum/preference/choiced/species)
	
	if(!S.sexes || (AGENDER in initial(species_type.species_traits)))
		return list(PLURAL)
	else if(FGENDER in initial(species_type.species_traits))
		return list(FEMALE)
	else if(MGENDER in initial(species_type.species_traits))
```

We don't need to make a brand new species each time for no reason

PR #19258, opened 2023-06-13
blacklists bounty consoles from the improvised emag

TheRyeGuyWhoWillNowDie:
access to syndicate bounties is supposed to cost 6tc, not 2. anyone who isn't a traitor with access to the improvised emag won't have an uplink, so this doesn't affect them

if this doesn't fix the problem the next step will be to put syndicate bounties behind a dedicated cargo-only traitor item, and role restricted items are boring and lame so i don't want to do that

# Changelog

:cl:  
tweak: makeshift emag does not work on bounty consoles
/:cl:

### DIFF
--- a/code/game/objects/items/cards_ids.dm
+++ b/code/game/objects/items/cards_ids.dm
@@ -102,6 +102,10 @@
 				M.ignite_mob()
 				to_chat(user, span_danger("The card shorts out and catches fire in your hands!"))
 			log_combat(user, target, "attempted to emag")
+			if(istype(target, /obj/machinery/computer/bounty)) //we can't have nice things
+				to_chat(user, span_notice("The cheap circuitry isn't strong enough to subvert this!"))
+				emagging = FALSE
+				return
 			target.emag_act(user)
 		emagging = FALSE
 


ynot01:
the cope of all time

DylonBR:
what problem does this solve though?

KingSuperBOB:
> what problem does this solve though?

It stops people from spending 18 TC on items to kil everyone then make a EMAG with just 2 TC and gain 30 TC

DylonBR:
people are still just gonna get a normal emag then buy a desword after doing one bounty. syndi bounties are honestly more overpowered than null crates. you had to actually get money for null crates and it was always a bunch of low value items

Moja10:
> people are still just gonna get a normal emag then buy a desword after doing one bounty. syndi bounties are honestly more overpowered than null crates. you had to actually get money for null crates and it was always a bunch of low value items

this, nothing in null crates, even if you bought 10+, compares to just getting more tc and being able to buy what you want, including combos not possible when you were hard limited to 20 tc. Gaining TC just results in absurd combos and it will still be a problem on IAA.

Moltijoe (approved):

PR #19263, opened 2023-06-13
[BOX] redesigned Port Bow and Fore Maintenance

ToasterBiome:

Trying to shake it up

image

# Wiki Documentation

These two maint areas look different

# Changelog

:cl:  
mapping: redesigned Port Bow and Fore Maintenance on BoxStation
/:cl:


Marmio64:
This gives more maint space and changes up the place but some changes hurt antag's abilities to use this maint area pretty bad, which isn't great when box maints can already be pretty hostile since everybody knows them so well. Main things i've found:

-Room above checkpoint having a door hurts it slightly in viability for jaunting into/using as a lair. Especially hurts slings.

-Doorless room in in the top right having space facing windows really hurts its viability for a lair since people can peep in unless you spraypaint the windows (and make it obvious someones using it as a lair). Also you can no longer weld into it from the space inside the 2 exterior doorlocks

-small room good for hiding that was next to the space airlock above vacant commissary is gone

Moja10:
Needs at least one secret room or compartment, something to actually escape security because as it stands, there's no where to go that isn't a random generated room, a maint hallway, or a room exposed by glass, aka this is just a place for people to think they can hide, and really they can't hide anywhere.

ToasterBiome:
> This gives more maint space and changes up the place but some changes hurt antag's abilities to use this maint area pretty bad, which isn't great when box maints can already be pretty hostile since everybody knows them so well. Main things i've found:
> 
> -Room above checkpoint having a door hurts it slightly in viability for jaunting into/using as a lair. Especially hurts slings.
> 
> -Doorless room in in the top right having space facing windows really hurts its viability for a lair since people can peep in unless you spraypaint the windows (and make it obvious someones using it as a lair). Also you can no longer weld into it from the space inside the 2 exterior doorlocks
> 
> -small room good for hiding that was next to the space airlock above vacant commissary is gone

If I wanted it to be exactly the same mechanically for antags then I wouldn't change it at all. Find new strategies.



> Needs at least one secret room or compartment, something to actually escape security because as it stands, there's no where to go that isn't a random generated room, a maint hallway, or a room exposed by glass, aka this is just a place for people to think they can hide, and really they can't hide anywhere.

image

Removed the windows + two doors for a "secret room".

Aquizit (changes requested):
Just this, as stated in discord

image

Aquizit (approved):

PR #19183, opened 2023-06-08
Memory loss upon death now can occur after the brain has decayed for a period of time

solwashere:
# Document the changes in your pull request

Makes it so that if someone is dead for (currently) 5 minutes, upon revival a message will play telling the player that they have memory loss, either 5 or 10 minutes would likely be good, I'll let council decide that.
Message is tied to brain decay, so switching brains wouldn't work, and it should let you preserve the brain in order to preserve someone's "memories".

This will need a council vote.

# Why this is good for the game
This makes hiding targets a lot more viable, especially witnesses, while not removing the risk entirely of someone yelling your name. In theory this will mean that less people will get round removed if they don't need to be

# Wiki Documentation

Likely will need to be mentioned in any sort of rule guidelines on the wiki if there are any.
# Changelog

:cl:  

tweak: being dead for a while will make your memory of said death go bye bye 

/:cl:

PR #19272, opened 2023-06-14
Revert "Memory loss upon death now can occur after the brain has decayed for a period of time"

ToasterBiome:
Reverts yogstation13/Yogstation#19183


DylonBR:
why revert what happened

PR #19275, opened 2023-06-14
CE starts with their gloves on them instead of in their bag

SapphicOverload:

Mainly for QoL, CE now starts with their gloves on them instead of in their bag.

# Changelog

:cl:   
tweak: CE starts with their gloves on
/:cl:


MajManatee (approved):
As it should have been grr

PR #19209, opened 2023-06-10
fixes rhumba beat so the cure plasma works and also makes it so it actually tells you the cure as that clue makes no sense

boodaliboo:
the title
:cl:  
 
tweak: tweaked  rhumba beat virus and also fixed it so its cure works someone should check the other viruses to see if any still use chemical tags

/:cl:

PR #19284, opened 2023-06-14
Fixes rhumba beat cure reagent (not bood and actually compiles edition)

Moltijoe:
Knowing bood, #19209 is gonna sit in limbo for the next week as they never fix it
so i've made another PR

:cl:  
bugfix: rhumba beat is possible to cure
/:cl:


JohnFulpWillard (approved):
we should unit test this

PR #19297, opened 2023-06-16
ports update_appearance

tattax:
just makes it easier to update things, murders update_icon

main pr: https://github.com/tgstation/tgstation/pull/55468

this is more brute force so nothing is fully updated to properly use update_desc, icon, name, but it doesn't really break anything

just encourages people to change it if they find it set up improperly (please)

:cl:  
imageadd: adds sus airlocks
/:cl:


code/_onclick/hud/lavaland_elite.dm
```diff
@@ -5,7 +5,7 @@
 	..()
 	pull_icon = new /atom/movable/screen/pull()
 	pull_icon.icon = ui_style
-	pull_icon.update_icon()
+	pull_icon.update_appearance(updates = ALL)
```
JohnFulpWillard: ``updates=ALL`` is the default so I think the args should just be empty when you do so. But this is update_icon so it should be ``update_appearance(UPDATE_ICON)``

JohnFulpWillard (changes requested):
You have all instances of update_icon set to ``update_appearance(ALL)``, which is wrong, it should be UPDATE_ICON

code/_onclick/hud/screen_objects.dm
```diff
@@ -181,7 +181,7 @@
 	cut_overlay(object_overlays)
 	object_overlays.Cut()
 
-/atom/movable/screen/inventory/update_icon()
+/atom/movable/screen/inventory/update_appearance(updates = ALL)
```
JohnFulpWillard: We should never be overwriting update_appearance at all, this should still be update_icon

tattax:
> You have all instances of update_icon set to `update_appearance(ALL)`, which is wrong, it should be UPDATE_ICON

not necessarily, no

many items configure overlays and whatnot in update_appearance

JohnFulpWillard:
> many items configure overlays and whatnot in update_appearance

That's wrong, it should be in ``update_overlays()``, which is part of UPDATE_ICON

tattax:
> That's wrong, it should be in `update_overlays()`, which is part of UPDATE_ICON

exactly

cuackles (approved):

code/__DEFINES/layers.dm
```diff
@@ -30,13 +30,13 @@
 #define ABOVE_NORMAL_TURF_LAYER 2.08
 #define LATTICE_LAYER 2.2
 #define DISPOSAL_PIPE_LAYER 2.25
-#define GAS_PIPE_HIDDEN_LAYER 2.35 //layer = initial(layer) + piping_layer / 1000 in atmospherics/update_icon() to determine order of pipe overlap
+#define GAS_PIPE_HIDDEN_LAYER 2.35 //layer = initial(layer) + piping_layer / 1000 in atmospherics/update_appearance(updates = ALL) to determine order of pipe overlap
```
JohnFulpWillard: this comment seems done accidentally

code/_onclick/hud/alert.dm
```diff
@@ -529,11 +529,45 @@ or shoot a gun to move around via Newton's 3rd Law of Motion."
...
 	icon_state = "emptycell"
 
+/atom/movable/screen/alert/emptycell/Initialize(mapload)
+	. = ..()
+	update_appearance(updates=UPDATE_DESC)
+
+/atom/movable/screen/alert/emptycell/update_desc()
+	. = ..()
+	desc = initial(desc)
+	if(length(GLOB.roundstart_station_borgcharger_areas))
```
JohnFulpWillard, line 543: Is this feature already in or are you adding it here?
Also remove the text from the desc since it's replaced with this.

Same for the other cell alerts

code/__DEFINES/layers.dm
tattax: i mean it works as a link since it got updated anyways

code/_onclick/hud/alert.dm
tattax: good call

code/_onclick/hud/lavaland_elite.dm
JohnFulpWillard: erhm this should still be updating just icon

ktlwjec0:
i think it broke pancake stacking but id have to double check, i put a cinnamon pancake on another earlier and it just acted as if the second one never existed

wow wrong pr lmao

PR #19308, opened 2023-06-17
everything is ruined because of you

TheRyeGuyWhoWillNowDie:
Syndicate bounties? 😃
Every single traitor buying an emag, bum rushing every department for their bounty items and then selling them for 40tc? 😭

changes the syndicate bounties to a cargo locked traitor item, buying it gives you a machine board to build the syndicate bounty console, no more emagging it.

# Changelog

:cl:  
rscadd: syndicate bounties can now be obtained via an 8tc cargo exclusive traitor item
rscdel: syndicate bounties can no longer be obtained via emag
/:cl:


DylonBR:
sudden increase in cargo players incoming

TheRyeGuyWhoWillNowDie:
> sudden increase in cargo players incoming

they can have fun jockeying over the 2 slots the department has left, cargo being present and existing would benefit the game anyways

Therandomhoboo:
Fucking players ruin everything fun

airlines7:
might be an idea just to make syndicate bounties equivalent to contractor kit. It sounds the same just for items instead of people.

Moja10:
Sounds like you didn't really change anything, people can still do the 40tc rush as a cargo tech? Sure, they might have less department tools if they were an engineer or scientist or whatever, but if they are freely stealing 40tc as those roles, like they can do that too as cargo tech. Not to mention just get whatever they need to do the all item speedrun, like AA, a jaws of life, regular airlock hacking, RCD, etc. 

Sounds like the issue is again, traitors were always balanced around 20tc max of items of their choice, and that any deviation from that was only random items, like null crates, random loadouts, or the surplus crates. I don't think your idea can be fundamentally balanced, outside of nerfing it into an unusable state, which isn't fun or ideal for anyone.

airlines7:
maybe you could do real objective items to steal for TC then? like black box for 2, nuke disk for 8, and hos gun for 5... something along those lines. Those are usually very hard to steal round start or to rush and get.
Might be cool to just trade the item directly for a null crate also, since it would be free random items.

TheRyeGuyWhoWillNowDie:
> maybe you could do real objective items to steal for TC then? like black box for 2, nuke disk for 8, and hos gun for 5... something along those lines. Those are usually very hard to steal round start or to rush and get. Might be cool to just trade the item directly for a null crate also, since it would be free random items.

Can't do actual objective items, those break the game

Moltijoe:
Balance council:
>the problem is not the availability of bounties, the problem is that the bounties just give straight TC for collecting items.
Either the TC amount needs to be severely reduced, or more perferrably, the bounties need to give items, ideally random.

could always just make them give null crates rather than TC

PR #19325, opened 2023-06-18
Mirage drive buff

LazennG:

mirage drive felt shitty for its price so i decided to buff it a bit to what i felt was more worth it.
- the cooldown you get from landing near others is done differently now, halving it per nearby person to encourage crowd diving
- maximum cooldown is checked before applying the normal cooldown reduction from landing near people now
- oriented windows don't fuck with the drive anymore so using it on the shuttle is possible 
- people in-between you and the destination get dizzy for 5 seconds and spin
- if there's somebody on the destination turf, an attack on them will happen dealing 3 times the travel distance in brute damage with 10 AP, with the minimum base damage being 15 and cap being 30. the attack has a cooldown of 5 seconds
- users can now bring people with them in conga fashion
- if they're not resistant then users will slip on wet turfs inbetween them and the target

https://github.com/yogstation13/Yogstation/assets/58535870/bfa4453d-3518-4711-a8e7-75e011b3edb6

# Wiki Documentation
https://wiki.yogstation.net/wiki/Syndicate_Items
For the mirage drive just add a part to the description about being able to inflict an attack to somebody on the destination that gets more powerful with the distance traveled. oh also about how the attack has a 5 second cooldown. Also mention that if they're not resistant then users will slip on wet turfs inbetween them and the target

# Changelog

:cl:  @Chubbygummibear @thegoldencat413 @solwashere @ToasterBiome 
rscadd: Added mirage drive attack
tweak: tweaked mirage drive cd stuff
/:cl:


code/modules/uplink/uplink_items.dm
```diff
@@ -1802,8 +1802,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 /datum/uplink_item/device_tools/mdrive
 	name = "Mirage Drive"
 	desc = "An experimental device created in an attempt to replicate the properties of bluespace. Utilizing coils with unique properties, the mirage drive is able to generate \
```
solwashere:
```suggestion
	desc = "An experimental device created as a byproduct of research into faster than light travel. Utilizing magnetic coils, the mirage drive is able to generate \
```

code/modules/uplink/uplink_items.dm
```diff
@@ -1802,8 +1802,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
...
-	kinetic energy and use it in a way that moves the user to their destination at a speed comparable to teleportation. Additionally, if there are other beings near the landing\
-	 site, the mirage drive will draw on their energy to recharge itself, slowing them down in the process."
+	kinetic energy and use it in a way that moves the user to their destination at a speed comparable to teleportation, so long as an unobstructed path between the two exists."
```
solwashere:
```suggestion
	kinetic energy and use it in a way that moves the user to their destination at a speed comparable to teleportation, so long as an unobstructed path between user and the target exists."
```

azzzertyy:
should it not be "the user and the target" instead of "user and the target"

code/game/objects/items/devices/miragedrive.dm
```diff
@@ -65,3 +89,66 @@
...
 
+/obj/item/mdrive/proc/nyoom(atom/movable/target, list/path, var/lagdist)
+	var/list/testpath = path
+	var/obj/effect/temp_visual/decoy/fading/onesecond/F = new(get_turf(target), target)
+	for(var/i in 1 to length(testpath))
+		var/turf/next_step = testpath[i]
+		if(ISMULTIPLE(i, 2) && (next_step))
+			var/turf/proper_step = testpath[i-lagdist]
+			F.forceMove(proper_step)
+			sleep(0.1 SECONDS)
```
Moltijoe: sorry, i'm supposed to kill you if you put sleep inside of a loop
you're gonna need to split it into procs and use addtimer

code/game/objects/items/devices/miragedrive.dm
```diff
@@ -65,3 +89,66 @@
...
+			var/turf/open/Q = get_step(get_turf(target), turn(target.dir, jumpangle))
+			if(Q.reachableTurftestdensity(T = Q))
+				K.forceMove(Q)
+			else
+				K.forceMove(get_turf(target))
+			K.setDir(get_dir(K, target))
+			jumpangle = jumpangle + 150
+		target.apply_damage(hurtamount, BRUTE, armor, wound_bonus=CANT_WOUND)
+		jab(target)
+		sleep(0.2 SECONDS)
```
Moltijoe: here too

code/game/objects/items/devices/miragedrive.dm
```diff
@@ -65,3 +89,66 @@
...
+	rushdowncd = COOLDOWN_FLURRYATTACK
+	COOLDOWN_START(src, last_attack, rushdowncd)
+
+/obj/item/mdrive/proc/jab(mob/living/target)
+	for(var/i = 1 to 3)
+		playsound(target, pick(hit_sounds), 25, 1, -1)
+		sleep(0.1 SECONDS)
```
Moltijoe: and here, though this one shouldn't be so hard

Moltijoe (changes requested):

Moltijoe (approved):

PR #19335, opened 2023-06-19
Removes YandereDev & calling mob_say on every item in your inventory.

JohnFulpWillard:

if
else if
else if
else if
else if
else if
else if
else if
else if
else if
else if
else if

Also removes making EVERY SINGLE ITEM in someone's contents call ``on_mob_say`` EVERY time you speak because it's absolutely fucking delusional. I replaced it with using the ``modifies_speech`` var.
Removes ``safety`` var and replaces it with the ``EMAGGED`` object flag like God intended.

Lastly, I made disabling sechailer voices be part of Alt Click instead of a verb.
I wanted to make the list of voices be part of double-examine, but apparently that's not a THING here so that's AWESOME.

# Changelog

:cl:  
tweak: You can now Alt Click the sechailer to turn it's audio lines on/off instead of using its right click menu.
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -813,8 +813,6 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 
 /obj/item/proc/on_mob_death(mob/living/L, gibbed)
 
-/obj/item/proc/on_mob_say(mob/living/L, message, message_range)
-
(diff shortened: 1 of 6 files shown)


maxion12345:
who is yanderedev why is everyone asking me about a chalice

Notamaniac:
yanderedev is the best coder i can't believe you're ruining this codebase

Moltijoe (approved):

PR #19336, opened 2023-06-19
Taking mats out of lathes requires ORM access, which most jobs have anyways

ynot01:

ORM can no longer be raided by unauthorized hoodlums

Jobs that have **do not** have access:
- Artist
- Assistant
- Brig Physician
- Chaplain
- Clerk
- Clown
- Curator
- Lawyer
- Mime
- Network Admin
- Paramedic
- Psychiatrist
- Tourist

# Changelog

:cl:  
tweak: ORM access now applies to protolathe mineral access as well. Lower tier jobs are now forbidden from taking minerals out.
/:cl:

### DIFF
--- a/code/game/mecha/mech_fabricator.dm
+++ b/code/game/mecha/mech_fabricator.dm
@@ -732,7 +732,7 @@
 	if (!issiliconoradminghost(usr))
 		var/obj/item/card/id/idcard = usr.get_idcard()
 		if(!idcard || !(ACCESS_MINERAL_STOREROOM in idcard.GetAccess()))
-			balloon_alert(usr, "No access!")
+			balloon_alert(usr, "no access!")
 			return 0
 	var/count = mat_container.retrieve_sheets(text2num(eject_amt), eject_sheet, drop_location())
 	var/list/matlist = list()
--- a/code/game/mecha/mech_fabricator.dm
+++ b/code/game/mecha/mech_fabricator.dm
@@ -7,7 +7,7 @@
 	use_power = IDLE_POWER_USE
 	idle_power_usage = 20
 	active_power_usage = 5000
-	
+
 	req_access = list(ACCESS_ROBO_CONTROL)
 	///Whether the access is hacked or not
 	var/hacked = FALSE
@@ -83,7 +83,7 @@
 /obj/machinery/mecha_part_fabricator/Destroy()
 	QDEL_NULL(wires)
 	return ..()
-	
+
 /obj/machinery/mecha_part_fabricator/RefreshParts()
 	var/T = 0
 
@@ -211,7 +211,7 @@
 
 	var/list/category_override = null
 	var/list/sub_category = null
-	
+
 	if(categories)
 		// Handle some special cases to build up sub-categories for the fab interface.
 		// Start with checking if this design builds a cyborg module.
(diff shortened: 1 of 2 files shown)


neeshacark:
time to add ORM access to my access request as assistant.

netmin should really have it given they need metal for machines like ai servers

can you give everyone skeleton crew orm access at least?

also rip pirates

Aquizit:
From this list the ones I can see getting access them are as such:

Brig phys/paramedic - if doctors have it then why not these two?

Chaplain/clerk/clown - These players build things often enough for chapel/shop/pranks.

Network admin - has been said already

Overall this feels unnecessary.

UltraBigRat:
I see no real point in this? unless theyre already an antag theres no real reson to take out random mats, unless you are actually doing something, this feels more like   arbitrary restrictions on jobs that could actually use the mats. Alot of the rp jobs like artist could use  mats to build stuff, along with other jobs like chaplain and curator performing renovations. And network admin often will be building stuff aswell.

This seems only useful for preventing pirates and to a lesser extent wizards and nukies emptying mats. Which at that point just make it so you need a crew id to take them out, instead of adding arbitrary id locks to a bunch of jobs for no reason.

Addust:
no

code/game/mecha/mech_fabricator.dm
```diff
@@ -729,6 +729,11 @@
 	if (rmat.on_hold())
 		say("Mineral access is on hold, please contact the quartermaster.")
 		return 0
+	if (!issiliconoradminghost(usr))
+		var/obj/item/card/id/idcard = usr.get_idcard()
+		if(!idcard || !(ACCESS_MINERAL_STOREROOM in idcard.GetAccess()))
+			balloon_alert(usr, "No access!")
```
tattax: balloon alerts use lowercase
```suggestion
			balloon_alert(usr, "no access!")
```

TheRyeGuyWhoWillNowDie:
???
If this is supposed to make the pirate ore yoink less viable, there's a button on the ORM interface that the AI can push to do this already, all this does is screw over 13 jobs that often times do need materials to build stuff. Adding ID locking to protolathes is like the pinnacle of hugbox.

code/modules/research/machinery/_production.dm
```diff
@@ -321,6 +321,11 @@
 	if (materials.on_hold())
 		say("Mineral access is on hold, please contact the quartermaster.")
 		return 0
+	if (!issiliconoradminghost(usr))
+		var/obj/item/card/id/idcard = usr.get_idcard()
+		if(!idcard || !(ACCESS_MINERAL_STOREROOM in idcard.GetAccess()))
+			balloon_alert(usr, "No access!")
+			return 0
```
JohnFulpWillard, line 328:
```suggestion
			balloon_alert(usr, "no access!")
			return 0
```

JohnFulpWillard (changes requested):
Any job that has access to a Techfab should also have access to the ORM

ynot01:
> Any job that has access to a Techfab should also have access to the ORM

out of scope or something c:

KingSuperBOB:
This feels more like an ided because someone went and stole mats

JohnFulpWillard:
> out of scope or something c:

It isn't, you're fixing the ORM access so you should fix the jobs who are supposed to have it too. Any job with techfab access should have ORM access since they both allow you to pull mats out.

PR #19176, opened 2023-06-08
Ruin asteroid medbay even more

airlines7:
# Description
## Main change

- Removed the 2 tile hallways splitting the surgical bay and treatment room and flipped it to the surgical bay is now on the right and not the left. This makes the main med hall much more open as well as the treatment room. Added an entrance and exit doors for the treatment center as well.

image
## Significant Changes

- Medical Storage is now right under the treatment room with easy access to all medical supplies, the airlocks are timed to 3 seconds instead of 15.

image

- Medical Breakroom is now called Medical backroom (only just to utilize the medical backroom turf) because its no longer attached to the Medical storage and instead connected to surgery and paramed office. Cryo cell oxygen is also setup there now.

image

## Minor Changes
In no particular order
- Drink, coffee, and snack vending machines are added to the breakroom.
- There is a "Surgical Handwashing Station" in surgery now to appease the germaphobes.
- Medical maintenance entrances are now in surgery and the breakroom.
- There are "Triage Shutters" to block off the primary hallway from the treatment center that might cause a doctor to lose focus.
- Shower is in a little hole in the wall in the treatment center.
- There is now a disposals chute in the treatment center.
- Medical storage does not have windoors for holding the medkits, they are just in the open.
- slight style change with the decals and floor using the slimmer line decals for the paramed room, breakroom, surgical room, treatment room, and medical storage.
- There are now fire alarms in every medical room changed.
- Treatment room and breakroom both have a centralized fire extinguisher.
- Treatment center has a public surgery bay, now with privacy curtains!
- Blood crate is no longer in the open and now in med storage.
- Flipped the position of med hallway fire/emergency lockers and chairs to align with surgery.
- There is now a stack of 3 roller beds in the breakroom.
- There is one less medicine locker in the medbay, there is now one.

### Full picture of the new layout
image

### Picture of Turf layout
image

# Why did I make this?
Currently asteroid medbay has a serious case of high traffic areas being jammed with people, specifically that two tile hallway leading to surgery and treatment. The surgery bay from my observation is rarely used for any surgeries. My solution was to get rid of that gap and place the treatment center closer to the front entrance. This removes that jammed high traffic place, as well as have surgery next to medical once more.
Med storage was moved next to the treatment center instead of through the break room to help have faster access to medical supplies.
TL;DR this layout just streamlines getting into the treatment center.
 
# Changelog

:cl:  Airlines7
mapping: Ruined Asteroid Medbay
/:cl:

PR #19339, opened 2023-06-19
Ruin Asteroid Medbay, now with a lamp on the table!

airlines7:
# Description
Clone of #19176 this is an attempt to fix the map bot from being weird.
## Main change

- Removed the 2 tile hallways splitting the surgical bay and treatment room and flipped it to the surgical bay is now on the right and not the left. This makes the main med hall much more open as well as the treatment room. Added an entrance and exit doors for the treatment center as well.

image

## Significant Changes

- Medical Storage is now right under the treatment room with easy access to all medical supplies, the airlocks are timed to 3 seconds instead of 15.

image

- Medical Breakroom is now called Medical backroom (only just to utilize the medical backroom turf) because its no longer attached to the Medical storage and instead connected to surgery and paramed office.

image

## Minor Changes
In no particular order
- Drink, coffee, and snack vending machines are added to the breakroom.
- There is a "Surgical Handwashing Station" in surgery now to appease the germaphobes.
- Medical maintenance entrance is in the breakroom.
- There are "Triage Shutters" to block off the primary hallway from the treatment center that might cause a doctor to lose focus.
- There is now a disposals chute in the treatment center.
- Medical storage does not have windoors for holding the medkits, they are just in the open.
- slight style change with the decals and floor using the slimmer line decals for the paramed room, breakroom, surgical room, treatment room, and medical storage.
- There are now fire alarms in every medical room changed.
- Treatment room and breakroom both have a centralized fire extinguisher.
- Treatment center has a public surgery bay.
- Blood crate is no longer in the open and now in med storage.
- Flipped the position of med hallway fire/emergency lockers and chairs to align with surgery.
- There is now a stack of 3 roller beds in the breakroom.
- There is one less medicine locker in the medbay, there is now one.

### Full picture of the new layout
image

### Picture of Turf layout
image

# Why did I make this?
Currently asteroid medbay has a serious case of high traffic areas being jammed with people, specifically that two tile hallway leading to surgery and treatment. The surgery bay from my observation is rarely used for any surgeries. My solution was to get rid of that gap and place the treatment center closer to the front entrance. This removes that jammed high traffic place, as well as have surgery next to medical once more.
Med storage was moved next to the treatment center instead of through the break room to help have faster access to medical supplies.
TL;DR this layout just streamlines getting into the treatment center.
 
# Changelog

:cl:  Airlines7
mapping: Ruined Asteroid Medbay
/:cl:


airlines7:
damn, didn't work guess I'll mald doing this the hard way.

airlines7:
Figured out what was wrong, the funny map code changed up the definition of parts of the map, an example is HoS office stamp, table, floor, turf was "abc" now its "dfg" Not gonna fix it because that's impossible to fix in a timely manner. I can 100% guarantee the map is fine the way it is and the only tangible changes are to Medbay.

Aquizit (changes requested):
Overall seems fine, just one thing:

image

Camera shouldn't be getting rained on by the shower. I made biome change it in his box medbay changes so I'm gonna make you change it too. :)

Also in this corner some decals are hidden under the wall, these should be removed.

image

And lastly, Asteroid Station has its own trim method (every station has a unique way of doing it), so that will need to be brought in line with the correct standards for Asteroid. I've pictured an example below

image

As opposed to what you're doing which is the Box standard.

airlines7:
> Camera shouldn't be getting rained on by the shower. I made biome change it in his box medbay changes so I'm gonna make you change it too. :)

Can do.

> image
> 
> And lastly, Asteroid Station has its own trim method (every station has a unique way of doing it), so that will need to be brought in line with the correct standards for Asteroid. I've pictured an example below

Biome was the one that requested the decal changes to be thin. I'm assuming they plan to make the rest of the station thin soon™
see #19176 comments for that little tidbit. I can change it back but its looks not as good in my opinion.

Aquizit:
> 
> Biome was the one that requested the decal changes to be thin. I'm assuming they plan to make the rest of the station thin soon™ see #19176 comments for that little tidbit. I can change it back but its looks not as good in my opinion.

Consistency is more important than half-finished trim projects. Have it match the rest of Asteroid design, please.

Aquizit (changes requested):
image
Should be broken line as current map:
image

image
These two walls here should be part of Medbay Central so the doors are tied to that APC not the Lobby one.

image
As surgery is no longer here, these shutters and the button for them should go.

image
This is a bad place for the air alarm simply for realism more or less, needing to mess with it while trying to reach around a tall tube. Putting it over on the West wall where the surgery shutters button currently is would work.

image
Should probably just be called Treatment Room not Sleeper Room.

image
Camera still getting showered on.

image
Missing purple trim corner.

image
Missing trim.

image
This glass wall would look better on the left tile, the one with the sleeper, than this one, as see it's kinda not attached to the main wall otherwise.

image
Camera in a window. Medbay Central above seems to be missing a camera in that area so just sticking it on one of the walls there should work, think it's supposed to be there it's called Medbay West.

image
image
Lights on windows.

image
Missing a light at all.

Aquizit:
image

While I appreciate the moving of the air alarm, the reason I suggested putting it over where the old surgery shutters button was is because this room still has to have one. If something needs to be done with this area's scrubbers/vents it requires an air alarm to do that.

image

So moving the light up next to the locker and putting the air alarm on this wall just above the window would be the best idea, I think.

---
image

I think I missed this the first run through but these should just be normal walls and not rwalls.

airlines7:
> image
> 
> I think I missed this the first run through but these should just be normal walls and not rwalls.

R-walls are so tiders/tators can't just weld in and get lathe access, I'll change it but if that becomes a problem I suggest changing them back.

Aquizit:
image
Stray cable leading to nowhere and I think this is the last thing I promise sorry for missing it.

Aquizit (dismissed):

Aquizit (approved):
Somehow we beat all the curses out of this one.

PR #17466, opened 2023-01-12
Farewell Syringe Guns

TheRyeGuyWhoWillNowDie:
Fuck off instant wins. Jamie approved!

Jamie has already taken them out of the armory, this is the last step.

Maintainer rationale:
one click wins bad, syringe guns will never be balanced

# Changelog

:cl:  
tweak: syringe gun armor piercing restrictions have been removed
tweak: printing syringe guns now needs illegal tech
tweak: printed rapid syringe guns replaced with normal syringe guns
tweak: chem gun in uplink has been replaced with rapid syringe gun (8tc) because the reagent gun is just really bad
tweak: syringe shotgun darts are removed from autolathes, and require illegal technology to print from the security/armory protolathe
/:cl:

PR #19344, opened 2023-06-20
Removes the ability to use piercing syringes in syringe guns, again

SapphicOverload:

Standard syringe guns were made able to fire piercing syringes after they were removed in #17466, so now that they're back in the hands of the crew this restriction is being re-introduced.

# Changelog

:cl:  
tweak: piercing syringes don't fit in crew syringe guns anymore (syndicate versions still can)
/:cl:


DylonBR:
this makes total sense. nanotrasen obviously wouldn't make syringes that fit their syringe guns

JohnFulpWillard:
Hm thus looks like it should have a unit test

iamthesilly:
> Hm thus looks like it should have a unit test

I tested it and it worked exactly as intended

JohnFulpWillard:
Unit test and knowing it works currently are two different things, unit test is ensuring your code will remain functional as CI will yell at coders if they break it in the future

Moltijoe (approved):

PR #19347, opened 2023-06-20
Nerf syringe embed transfer

ynot01:

As I see it, syringe embeds are supposed to act like bolas. You get hit, you have to take it off at some point or suffer the consequences.

The way syringes currently work is as follows:
- You get hit, inject instantly (5u for default syringe)
- Every life tick, inject again (every 2 seconds, 5u again)

This gives you 2 seconds to stop what you were doing and take out the syringe, or else you now have 10 units of whatever mix in your bloodstream, which is typically enough to kill or sleep if not severely debilitate you.

To make syringe combat more fair, I have severely nerfed the transfer rates by:
- Removing the instant inject, you now only get injected every 2 seconds
- **Halving** injection rate, which means syringes now inject at **2.5u per tick** and bluespace syringes now inject at 10u per tick

To compensate, removal of syringes now takes a full second rather than 0.25 seconds.

# Changelog

:cl:  
tweak: Syringe embed transfer rate halved, but takes slightly longer to take out
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/syringes.dm
+++ b/code/modules/reagents/reagent_containers/syringes.dm
@@ -15,14 +15,14 @@
 	materials = list(/datum/material/iron=10, /datum/material/glass=20)
 	reagent_flags = TRANSPARENT
 	sharpness = SHARP_POINTY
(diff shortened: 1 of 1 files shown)


Therandomhoboo:
With the meth & bathsalts nerf pr [#19348] in the waits and this. One or the other should go through and hopefully not both. 
Even though the other pr isn't yours I think it should just be made aware if that goes through then this pr is pretty much a useless nerf apart from an antag nerf and a heavy one at that. The only thing that tators can use is the poison kit and even then with this potentially in play, it will just make poison kit completely useless

TheRyeGuyWhoWillNowDie:
it's almost like syringe guns were removed because they're overpowered

Moltijoe (approved):

PR #19348, opened 2023-06-20
Meth and bathsalts metabolize faster but require more to overdose

SapphicOverload:

Meth and bath salts metabolize twice as quickly but require twice as much to become addicted or overdose.

This is so that it takes more than one syringe of bathsalts or meth to cause addictions or overdose, because they were a big problem in syringe guns before the removal and likely will be again now that they're back.

# Wiki Documentation

For both bathsalts and meth, update the addiction threshold to 20 and overdose threshold to 40 on the guide to chemistry.

# Changelog

:cl:  
tweak: doubles metabolization rate of bathsalts and meth
tweak: doubles addiction/overdose threshold of bathsalts and meth
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/drug_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/drug_reagents.dm
@@ -192,9 +192,9 @@
 	description = "Neutralizes mannitol. Reduces stun times by about 300%, speeds the user up, and allows the user to quickly recover stamina while dealing a small amount of Brain damage. If overdosed the subject will move randomly, laugh randomly, drop items and suffer from Toxin and Brain damage. If addicted the subject will constantly jitter and drool, before becoming dizzy and losing motor control and eventually suffer heavy toxin damage."
 	reagent_state = LIQUID
 	color = "#FAFAFA"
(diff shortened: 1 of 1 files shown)


Therandomhoboo:
So there is another PR up [#19347] which is this. I know you didn't make the other one but i just wanted to make sure its aware that if one gets in, then the other should get closed due to them drastically affecting the syringe gun. 

That aside, I like the nerf but I think its way too high, so instead of just doubling the number maybe just make the addiction/overdose higher than one syringe instead of requiring 3 normal syringes that have to not be taken out for the duration or 1 bluespace syringe which I haven't actually seen at all used. 

Ignoring the antagonistic use aside too. I also think the higher addiction/overdose also will lead to people abusing this cause it is quite high now. So they can be fast and ignore stam crit easier

iamthesilly:
> Ignoring the antagonistic use aside too. I also think the higher addiction/overdose also will lead to people abusing this cause it is quite high now. So they can be fast and ignore stam crit easier

This is offset by it metabolizing twice as fast, so you can't use it for any longer than you used to.

Moltijoe (approved):

PR #19350, opened 2023-06-20
[PROJECT] Removes Devils

JohnFulpWillard:

Removes devils from the game

HackMD can be found here: https://hackmd.io/189iSq-4TqysuqT5ICQ6tw

Replaces hellbound var on ``/mob/living`` with a ``TRAIT_HELLBOUND`` on the mind, so it follows through bodies.


solwashere:
Mode is unmaintained and the proposed replacement for it is infinitely better than what we have now

Moltijoe:
For the replacement, i think i like chaplain and curator being the anti-devil jobs. Lawyer being one but still being able to become an agent seems wrong.

cowbot92:
The Devil is in the details

PR #19356, opened 2023-06-21
Nerfs preternis

Moltijoe:
When i first added it, i intended for it to include weapons, but i didn't know enough about ss13 to know it didn't work
Now i know it doesn't and i've fixed the problem

:cl:  
bugfix: Preternis attack slowdown now properly applies to weapons
/:cl:


iamthesilly:
This breaks gloves of the north star (again)

Also it applies to using every single item in the game, weapon or not

iamthesilly:
Actually it might even cause delay when you just click anywhere at all because of how afterattack works

Moltijoe:
> Actually it might even cause delay when you just click anywhere at all because of how afterattack works

This is specifically item afterattack, meaning it wont affect unarmed attacks. It very likely might mess with other stuff so i'll probably have to add a check that it did damage to the target

JohnFulpWillard (approved):

PR #19365, opened 2023-06-22
Two-handed component

JohnFulpWillard:

This ports ``/datum/component/two_handed`` from TG and some other minor things that were necessary for it to work.

Currently there's a Null rod that's meant to be two-handed. Instead of having the component, it currently exists to be used in the Chaplain's null rod menu, then DELETES ITSELF when picked up, SPAWNS a NEW, COMPLETELY separate item with the SAME icon, name, attack verb, desc, strength, gives THAT item the antimagic component, then puts it in your hand. It's insane, it's crazy. It's... no longer needed.

# Changelog

:cl:  
experimental: Added the two-handed component. Fireaxes, Double-bladed Eswords, Pitchforks, Binoculars, Baseball bats, etc. should all work the same, but now have a better backend that allows us to do more stuff with it, like making the Chaplain's null rod not delete itself upon being picked up.
/:cl:


Moltijoe:
oh god bless, the nullrod jank workaround was the only clean way i could think of doing it

ToasterBiome (approved):
ok

PR #19368, opened 2023-06-22
Makes healing items less tedious to use

SapphicOverload:

Healing items such as sutures, mesh, etc. now heal a random damaged limb if the targeted one is fine, in order to make them less tedious to use.

# Changelog

:cl:   
tweak: healing items like sutures heal a random damaged body part instead if the selected one is fine
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -72,6 +72,9 @@
 
 /obj/item/stack/medical/proc/heal_carbon(mob/living/carbon/C, mob/user, brute, burn)
 	var/obj/item/bodypart/affecting = C.get_bodypart(check_zone(user.zone_selected))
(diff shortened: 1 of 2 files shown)


UltraBigRat:
does this work for welder and cable healing too?

iamthesilly:
> does this work for welder and cable healing too?

Yes

SomeguyManperson:
Can't you just switch the limb you are targetting?

iamthesilly:
> Can't you just switch the limb you are targetting?

Yes but mashing the keyboard like that is tedious and annoying

Therandomhoboo:
I think its fine where it is at. Yeah it is tedious/annoying but all healing shouldn't be just a breeze to do. Unless its bardrinks then that should always be a breeze.... a _seabreeze_ that is.

As well as I think this also stops antags/sec dipping into a side standing still and then healing all limbs up.

Scrambledeggs00:
Is it canceled if it tries healing a body part that can't be heal? An example being head with something that blocks healing items like the envirohelm.

ToasterBiome (approved):

PR #19357, opened 2023-06-21
Ate without table

ynot01:
# Document the changes in your pull request

-3 mood event for 1 minute when eating without table

Does not apply to junk food

# Changelog

:cl:  
tweak: Eating without a table now makes you sad
/:cl:

PR #19369, opened 2023-06-22
Revert "Ate without table"

Mqiib:
Reverts yogstation13/Yogstation#19357

Doesn't work properly, just makes you sad whenever you eat food regardless of table location.


ynot01:
Sounds like someone ate without table

neeshacark:
functions just like rimworld, bless

adamsong:
Fix it then?

Mqiib:
> Fix it then?

Maybe make sure new features actually work?

adamsong:
Should be a bugfix, not a full revert

PR #19380, opened 2023-06-23
Makes robotic healing more in line with sutures/mesh

ynot01:

The way robo limb healing works right now is pretty weird.

Robo limb healing worked as follows:
- Takes 2 seconds universally
- If you're an engineer reduce timer by 20%
- If you're a roboticist not healing yourself, reduce timer by 20%, not stacking with the above

This is how sutures/mesh work:
- Take 1 seconds to heal someone else
- Take 3 seconds to heal yourself

To make robo healing more in line with medical, I have nerfed robo limbs to take 3x longer when healing yourself, like medical items. In addition, robos now keep the 20% bonus even if healing themselves.

While making this PR, I also got the idea to make medical items heal 20% faster if used by medical personnel, like robotic healing.

So essentially, robotic healing are now the same as fleshy healing, but takes twice as long due to the overwhelming prevalence and accessibility (especially ability to be printed) of healing items. Not to mention robotic limbs are supposed to be a downside.

#

As an aside, the experimental welding tool has 0.5x toolspeed which means it will heal at a rate of 1 or 3 seconds

# Changelog

:cl:  
tweak: Robotic limbs now take longer to heal if you are healing yourself
tweak: Roboticists are no longer nerfed if healing their own robotic limbs
tweak: Medical personnel now use sutures/meshes/etc 20% faster
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -15,7 +15,7 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 	///Icon file for right inhand overlays
 	var/righthand_file = 'icons/mob/inhands/items_righthand.dmi'
 
-	///Icon file for mob worn overlays. 
+	///Icon file for mob worn overlays.
 	///no var for state because it should *always* be the same as icon_state
 	var/icon/mob_overlay_icon
 	//Forced mob worn layer instead of the standard preferred ssize.
@@ -34,7 +34,7 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 
 	obj_flags = NONE
 	var/item_flags = NONE
-	
+
 	var/hitsound
 	var/usesound
 	///Used when yate into a mob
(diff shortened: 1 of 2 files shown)


UltraBigRat:
Honestly, maybe let ipcs keep fast healing? Theyve been nerfed to hell and back already with very few upsides that arent done far better by another race, id agree with this  nerf for preternis healing though with all the benefits they get already.

ynot01:
> Honestly, maybe let ipcs keep fast healing? 

No

Hopekz:
The 20% speed up seems like an after thought I'd atomize it personally

ynot01:
> The 20% speed up seems like an after thought I'd atomize it personally

What better way to synchronize the two than to make them more alike?

Hopekz:
> What better way to synchronize the two than to make them more alike?

I get that argument and normally it would be fine but the focus of the PR is "Makes robotic healing more in line with sutures" but in this PR you are also changing how sutures are used.

So you're presenting this PR as moving robotic healing to a goal post but at the same time changing where the goal post is if that makes sense. This is why I would personally atomize it.

ynot01:
> So you're presenting this PR as moving robotic healing to a goal post but at the same time changing where the goal post is if that makes sense. This is why I would personally atomize it.

I am moving the goal post closer to the ball

Hopekz:
> I am moving the goal post closer to the ball

https://www.youtube.com/watch?v=WtO3AHMBePY

wonderinghost:
ynot with the new post of species restricting or pointlesssizing

as well as baby proofing all weapons crew can make

neeshacark:
ffs really? just remove ipc

ToasterBiome (approved):

PR #19382, opened 2023-06-23
chicken nugget needs batter to craft

ktlwjec0:
Makes nuggets need 2u batter, same as battered fish. For realism I guess? And the sprite is a battered nugget.

# Wiki Documentation

add batter to nugget i will do this

:cl:  Ktlwjec
tweak: Chicken nuggets require 2u of batter to craft, along with the cutlet.
/:cl:


maxion12345:
when are we adding code so that rats pathfind towards the deepfrier if inside of the kitchen

ToasterBiome (approved):

PR #19406, opened 2023-06-25
you can remove Jelly Spagh-o-nut from the food cart again

ktlwjec0:
i cant get it working another way, so just removes the quotations around "'jelly'".

you could add it to the food cart but it was stuck there forever, never able to be dispensed.

:cl:  Ktlwjec
bugfix: You can dispense Jelly Spagh-o-nut from the food cart.
/:cl:


ktlwjec0:
yeah this is horrid it needs fixing properly send help

PR #19414, opened 2023-06-25
You can reload utensils into the dinnerware vendor

ktlwjec0:
qol, it's really annoying when you vend too many of something by mistake, or upgrade your knife to a cleaver and now you have two, use a tray for one task and want it put back, etc. Removed condiments because why would you put food back, seems off to me.

:cl:  Ktlwjec
rscadd: You can reload utensils into the dinnerware (kitchen) vendor
/:cl:


Maia-J:
every vendor needs this

ToasterBiome (approved):

PR #19417, opened 2023-06-25
Sinful demon Greed summon cursed item spell

Marmio64:
Adds a new spell for Greed demon of sin that lets them summon a special cursed item from a list of 5 every 3 minutes. They all have moderately useful applications in return for something. You can't pick which one to summon, and the demon still gets any negative effects if they decide to use them, they're primarily meant to be sold. Spell uses a moneybag as it's icon.

Possible items
(all the ones that impart negative effects are fairly obvious in what they do if you read the item descriptions)
-------------------------

Peculiar Katana: Huge item, looks like a normal katana with a special name. Deals 20 damage, 15 AP, and a 25 block chance. However, every time you swing it, you take 5-10 damage.

Strange pill: Filled with 20 units of omnizine, 3 units of hell water, and 2 units of lavaland extract, it will impart long lasting minor healing effects at the cost of setting you on fire. If you aren't flameproof or have a extinguisher on hard, using it during combat will be incredibly ill advised.

Cursed book of babel: Works like a normal book of babel, granting you all languages, but also takes away 20 max health upon use.

Bottle of dark blood: Gives you back 20 max health (only if you lost any, you cant go above 100) that you might have lost from gambling or something, and in return curses you into a shadow person.

Weak sacred flame spell book: Grants you a weaker version of the sacred flame spell. It has a lower radius for giving people fire stacks, only gives half as many, and has a far longer cooldown. The downside is that you are indeed setting yourself on fire to use this so if you're not fireproof its not very good.

----------------

Don't know how balanced this would actually be, but sin demons only have 5 weight as a midround event and then its a 1 in 5 chance to even get greed, so it wouldn't be seen often.

Good for the game because it gives them more things to try to sell to people than just the slot machine. Also they can use the cursed items themselves too for powerups if they're alright with the consequences, so it gives them more versatility.

this also fixes the ignite spell icon not showing since i noticed it

# Changelog

:cl:  
rscadd: Greed sinful demons now get a spell that lets them summon a random unique cursed item to sell to patrons.
bugfix: Wrath demon ignite spell icon properly shows
/:cl:

### DIFF
--- a/code/game/objects/items/granters/magic/sacred_flame.dm
+++ b/code/game/objects/items/granters/magic/sacred_flame.dm
@@ -12,3 +12,6 @@
 		"No pain, no gain...",
 		"One with the flame...",
 	)
+
+/obj/item/book/granter/action/spell/sacredflame/weak //grants the far weaker version
+	granted_action = /datum/action/cooldown/spell/aoe/sacred_flame/weak
--- a/code/modules/antagonists/demon/demons.dm
+++ b/code/modules/antagonists/demon/demons.dm
@@ -26,6 +26,7 @@
 		/datum/action/cooldown/spell/conjure/summon_mirror,
 		/datum/action/cooldown/spell/touch/mend,
 		/datum/action/cooldown/spell/touch/torment,
+		/datum/action/cooldown/spell/conjure/cursed_item,
 		))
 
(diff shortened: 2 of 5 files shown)


ToasterBiome (approved):
sure idc

PR #19412, opened 2023-06-25
Deletes roundstart cryogenics

solwashere:
# Document the changes in your pull request

What it says on the tin.

Cryo has been deleted from roundstart medbay on all 4 maps. I believe that this is a positive change as cryo at the moment is able to heal 90 percent of injuries and completely heal all wounds within seconds. This change would force players to learn how to use other healing methods, make wounds have a larger impact and actually be a problem that requires a small amount of thought, and help chemicals like cloneaxadone see use.

image
image
image
image

# Spriting

# Wiki Documentation

Medical no longer starts with cryo

# Changelog

:cl:  

mapping: Deletes all instances of cryogenics within all 4 maps.

/:cl:

PR #19418, opened 2023-06-25
Cryo can only heal burn wounds

SapphicOverload:

Cryo now only heals burn wounds on top of healing all damage types. It already heals a respectable amount of damage and is great for an overloaded medbay, but the way it is sort of makes wounds a thing only antagonists will ever have to actually worry about getting treated, and doctors over-rely on this feature to the point they don't know how to fix basic things like hairline fractures.

Alternative to #19412 

# Wiki Documentation

If there's anything that says cryo fixes all wounds anywhere that'll have to change to just burn wounds

# Changelog

:cl:  
tweak: cryo can only fix burn wounds
/:cl:

### DIFF
--- a/code/datums/wounds/_wounds.dm
+++ b/code/datums/wounds/_wounds.dm
@@ -79,6 +79,8 @@
 	var/datum/surgery/attached_surgery
 	/// if you're a lazy git and just throw them in cryo, the wound will go away after accumulating severity * 25 power
 	var/cryo_progress
+	/// wait, can cryo even heal it at all?
+	var/can_cryo_heal = FALSE
 
 	/// What kind of scars this wound will create description wise once healed
 	var/scar_keyword = "generic"
@@ -324,9 +326,10 @@
 
 /// Called from cryoxadone and pyroxadone when they're proc'ing. Wounds will slowly be fixed separately from other methods when these are in effect. crappy name but eh
 /datum/wound/proc/on_xadone(power)
-	cryo_progress += power
-	if(cryo_progress > 66 * severity)
-		qdel(src)
+	if(can_cryo_heal)
+		cryo_progress += power
+		if(cryo_progress > 66 * severity)
+			qdel(src)
 
 /// When synthflesh is applied to the victim, we call this. No sense in setting up an entire chem reaction system for wounds when we only care for a few chems. Probably will change in the future
 /datum/wound/proc/on_synthflesh(power)
--- a/code/datums/wounds/_wounds.dm
+++ b/code/datums/wounds/_wounds.dm
@@ -79,8 +79,6 @@
 	var/datum/surgery/attached_surgery
 	/// if you're a lazy git and just throw them in cryo, the wound will go away after accumulating severity * 25 power
 	var/cryo_progress
-	/// wait, can cryo even heal it at all?
-	var/can_cryo_heal = FALSE
 
 	/// What kind of scars this wound will create description wise once healed
 	var/scar_keyword = "generic"
@@ -326,10 +324,11 @@
 
 /// Called from cryoxadone and pyroxadone when they're proc'ing. Wounds will slowly be fixed separately from other methods when these are in effect. crappy name but eh
 /datum/wound/proc/on_xadone(power)
-	if(can_cryo_heal)
-		cryo_progress += power
-		if(cryo_progress > 66 * severity)
-			qdel(src)
+	if(!(wound_flags & ACCEPTS_CRYO))
+		return
+	cryo_progress += power
+	if(cryo_progress > 66 * severity)
+		qdel(src)
 
 /// When synthflesh is applied to the victim, we call this. No sense in setting up an entire chem reaction system for wounds when we only care for a few chems. Probably will change in the future
 /datum/wound/proc/on_synthflesh(power)
--- a/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
@@ -36,6 +36,7 @@
 	var/breakout_time = 300
 	///Cryo will continue to treat people with 0 damage but existing wounds, but will sound off when damage healing is done in case doctors want to directly treat the wounds instead
 	var/treating_wounds = FALSE
+	var/can_heal_wounds = FALSE // just in case someone wants to add something that lets it heal wounds
 	fair_market_price = 10
 	payment_department = ACCOUNT_MED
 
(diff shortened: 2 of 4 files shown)


TheRyeGuyWhoWillNowDie:
i like how medbay lizard whining went from "cryo is useless!" to "cryo is overpowered because it makes my job easier for new players to understand!"

wonderinghost:
As long as James kas isn't here fuck yes

solwashere:
> i like how medbay lizard whining went from "cryo is useless!" to "cryo is overpowered because it makes my job easier for new players to understand!"

Nobody ever said that cryo is useless

solwashere:
I love it when people lie to try to villainize a group of players and make them look petty when the person themselves is extremely petty and a contrarian

Moja10:
I don't see the justification of this at all, it is a fun thing to make a good cryo mix. it is a usable alternative when there's a lot of dead people and not enough doctors available to do 3000 surgery steps. It's an alternative for healing certain things like DNA damage, like from cryo.

All this seems like it will accomplish is that doctors unaware of this PR will throw people into cryo unaware it will not do anything about the multiple other damage types it won't heal. it can already be a problem with certain races having weird cryo interactions, I don't know why it needs to be made worse.

Also, doesn't this mean that lizards will die to cryo? They will get cold, cryo won't heal any brute, the cold gives brute, they are unconscious and slowly dying to the accumulating brute. Am I understanding this correctly?

To me, this is change for the sake of change, I fail to see the logic behind this at all. I fail to see why cryo needs to be nerfed at all. If the problem is that it's SUPERIOR to normal healing methods, then it's speed should be reduced and that's it. But I don't consider it superior to the other methods like surgery other than its hands off. It's only faster with chems, or slow surgeons, which is a good task for chemists to focus on in a round.

solwashere:
> I don't see the justification of this at all, it is a fun thing to make a good cryo mix. it is a usable alternative when there's a lot of dead people and not enough doctors available to do 3000 surgery steps. It's an alternative for healing certain things like DNA damage, like from cryo.
> 
> All this seems like it will accomplish is that doctors unaware of this PR will throw people into cryo unaware it will not do anything about the multiple other damage types it won't heal. it can already be a problem with certain races having weird cryo interactions, I don't know why it needs to be made worse.
> 
> Also, doesn't this mean that lizards will die to cryo? They will get cold, cryo won't heal any brute, the cold gives brute, they are unconscious and slowly dying to the accumulating brute. Am I understanding this correctly?
> 
> To me, this is change for the sake of change, I fail to see the logic behind this at all. I fail to see why cryo needs to be nerfed at all. If the problem is that it's SUPERIOR to normal healing methods, then it's speed should be reduced and that's it. But I don't consider it superior to the other methods like surgery other than its hands off. It's only faster with chems, or slow surgeons, which is a good task for chemists to focus on in a round. 
> 
> 

The wound system for it should be nerfed because all it does is make wounds completely ineffective unless you're an antag.
Covered head to toe in fractures as sec or crew? Get shoved into cryo and be healed with little to no effort. It also stops new players from learning how to properly deal with wounds because cryo can instantly heal it. Medbay is already boring at times, and having a tube that heals almost every single affliction in the game within seconds is bad, and even if we made it "longer", that just means that people are gonna be stuck in the tubes sleeping for a long period of time, which I already have had people get angry at me because they were tossed into cryo and have to spend the next 5 minutes asleep.

Moja10:
Ignore my previous message, I am a stupid and interpreted rhe code as not healing damage, instead of not healing wounds.

Moltijoe (approved):

ilikesaying:
Cryo heals wounds??

code/datums/wounds/_wounds.dm
```diff
@@ -79,6 +79,8 @@
 	var/datum/surgery/attached_surgery
 	/// if you're a lazy git and just throw them in cryo, the wound will go away after accumulating severity * 25 power
 	var/cryo_progress
+	/// wait, can cryo even heal it at all?
+	var/can_cryo_heal = FALSE
```
JohnFulpWillard: I think this should be a flag on ``wound_flags`` instead, like ``ACCEPTS_CRYO``.

code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
```diff
@@ -209,16 +209,18 @@
...
-					treating_wounds = TRUE
-					playsound(src, 'sound/machines/cryo_warning.ogg', volume) // Bug the doctors.
-					var/msg = "Patient vitals fully recovered, continuing automated wound treatment."
-					radio.talk_into(src, msg, radio_channel)
-			else // otherwise if we were only treating wounds and now we don't have any, turn off treating_wounds so we can boot 'em out
-				treating_wounds = FALSE
-
+		var/has_cryo_wound = FALSE
+		if(C && C.all_wounds)
+			for(var/datum/wound/wound in C.all_wounds)
```
JohnFulpWillard:
```suggestion
			for(var/datum/wound/wound as anything in C.all_wounds)
```

code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
```diff
@@ -209,16 +209,18 @@
...
-			if(C.all_wounds)
-				if(!treating_wounds) // if we have wounds and haven't already alerted the doctors we're only dealing with the wounds, let them know
-					treating_wounds = TRUE
-					playsound(src, 'sound/machines/cryo_warning.ogg', volume) // Bug the doctors.
-					var/msg = "Patient vitals fully recovered, continuing automated wound treatment."
-					radio.talk_into(src, msg, radio_channel)
-			else // otherwise if we were only treating wounds and now we don't have any, turn off treating_wounds so we can boot 'em out
-				treating_wounds = FALSE
-
+		var/has_cryo_wound = FALSE
```
JohnFulpWillard, line 212: Why do we have this var instead of just setting ``treating_wounds`` to true before ``break``?

iamthesilly: Doing it that way would mean having to set treating_wounds to false beforehand, which causes the cryo tube to either play the beep and send the radio message on every tick that it's treating wounds or never at all depending on where it's being set to true.

code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
iamthesilly: What would this change do exactly?

Moltijoe (approved):

code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
JohnFulpWillard: It removes the check of all things in ``all_wounds`` to ensure they are a ``/datum/wound``, saving some time.

ToasterBiome (approved):

PR #19437, opened 2023-06-27
Removes "seconds" from obsessed effects that arent tracked like that

SomeguyManperson:

Blur eyes and confusion stacks both reduce at the rate of 1 per 2 seconds, unless i missed something, and being completely unable to see for nearly 4 minutes is uh stupid

# Spriting

# Wiki Documentation

# Changelog

:cl:  
bugfix: obsession effects on speaking near your target no longer last several minutes when they are meant to last seveeal seconds
/:cl:


ToasterBiome (approved):
sexonds

PR #19461, opened 2023-06-29
Makes welding cheddar less annoying

ktlwjec0:
So it currently makes you drop the welder after each weld, just really annoying. Have tested, cheese functions as normal and it works.

:cl:  Ktlwjec
tweak: Welding cheddar wheels no longer makes you drop the welder.
/:cl:


missatessatessy:
Wonder why no one told me about this issue after so long lol, thanks for fixing it

Moltijoe (approved):

PR #19462, opened 2023-06-29
Robotic healing tweaks

SapphicOverload:

The last PR about this made robotic healing painfully slow, so this makes robotic healing twice as fast but only heals 66% as much each time, making them the same speed and healing amount as sutures. This means that while robotic healing is overall slightly faster than before, it requires 50% more welding fuel and cables to heal the same amount of damage.

This also fixes a bug where using a welder to repair yourself only required sunglasses-level flash protection instead of the intended amount of protection for that specific welding tool.

# Changelog

:cl:  
bugfix: welding yourself requires the correct amount of flash protection
tweak: robotic healing is faster but heals less each time
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -1079,16 +1079,12 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 //specifically for "suture" type robotic healing items
 //amount is either the fuel of a welding tool, or the number of wires consumed
 //volume is how loud the sound of the item is
-#define ROBO_LIMB_HEAL_SELF 6 SECONDS
-#define ROBO_LIMB_HEAL_OTHER 2 SECONDS
-/obj/item/proc/heal_robo_limb(obj/item/I, mob/living/carbon/human/H,  mob/user, brute_heal = 0, burn_heal = 0, amount = 0, volume = 0)
-	var/used_delay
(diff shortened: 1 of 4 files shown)


Therandomhoboo:
The problem with this is, welding fuel is infinite. Cables are too pretty much infinite. So this just basically just goes back to the same problem as what the previous PR tried to fix. However I agree it shouldn't be 6 seconds.

iamthesilly:
> The problem with this is, welding fuel is infinite. Cables are too pretty much infinite. So this just basically just goes back to the same problem as was the PR tried to fix. However I agree it shouldn't be 6 seconds.

At the very least you still have to go find more cables or fuel when they run out, which will happen more often now.

wonderinghost:
Cable coils aren't the problem. Welding fluid is. Can't have a welder if you can't the stuff. Problem is that the tank sizes are small outside of the advanced and alien welders. And 20 units isn't a lot of welder fluid where it takes 1 unit to heal the amount of damage and 1 unit for cutting and beating shit by accident. It be interesting if it was easier to reload from other sources then a tank. However that would require a mass redesign of probably the fuel tanks but how welders work

Because you need to carry more welders then you can 2 coils of wire

Moltijoe (approved):

PR #19468, opened 2023-06-29
tiny cook vendor changes

ktlwjec0:
Removes apron from utensils vendor (they have a drobe with them in), ups damp rag+jumpsuit to 2 in vendors since there's 2 cooks. Just makes sense I guess.

# Wiki Documentation

vending machines page i can do this

:cl:  Ktlwjec
rscdel: Chef's apron from Dinnerware vendor (this is  still in the ChefDrobe).
tweak: ChefDrobe has 2 damp rags and jumpsuits (up from 1).
/:cl:


ynot01 (approved):
I'd ask why not have 2 of everything else (hat, suit) but it is not that important to me

PR #19483, opened 2023-07-01
ports Holymelons are more melon-like

ktlwjec0:
https://github.com/tgstation/tgstation/pull/70411

should be fruit as they are melons, and they should be juiced as they are melons. it just makes sense

# Wiki Documentation

possibly somewhere but i doubt it? i'll have a look

:cl:  Melbert
rscadd: Holy melons are now more melon like - they are now fruits (instead of no food type), and can be juiced into holy water
/:cl:


ynot01 (approved):
I find it a bit strange that holy melons are not simply a subtype

PR #19485, opened 2023-07-01
Buff gas production from HFR

warface1234455:
Currently HFR gas production is very slow, people would rather make gases in open tile than this.

also i have tested this and it did make gas a bit faster than usual

Buff gas production from HFR by 20x ~~it wont be making thousands mole per second~~

# Changelog

:cl:  

tweak: Buff gas production from HFR by 20x
/:cl:


iamthesilly:
I think multiplying production and fuel consumption by the current fusion level would be a better idea, the gas production from this is too high and doesn't consume any more fuel than before, making it absurdly efficient

ShoziOffical:
> I think multiplying production and fuel consumption by the current fusion level would be a better idea, the gas production from this is too high and doesn't consume any more fuel than before, making it absurdly efficient

Yes i want to make this reactor efficient for gas making instead of a bomb waiting to be exploded, also it doesn't make that much gases with this mutiply anyway. I also don't think mutiplying production would be a better idea because some recipes already got mutiplied by fusion level and changing it won't make it much different than this

ToasterBiome:
Too much

warface1234455:
> Too much

@ToasterBiome did you test it? Probally not. Did i test it? Yes it doesnt make that much gas

warface1234455:
not like thousand mol of gases being made per second

PR #19488, opened 2023-07-01
Blowgun Tweak

Runian:
Blowgun no longer requires you to stand still for 2.5 seconds. To compensate, it does not have infinite range anymore and has a range of 2 which means it must be used up close instead of across the hallway (where it is safer). It also deals 5 more oxygen and 5 more stamina damage to a total of 25 oxygen and stamina damage.

Technically fixes a bug where you can abuse/stack the 2.5 second stand still to put yourself into unconsciousness via oxygen damage (by removing the stand still entirely).

Works fine in local.

# Wiki Documentation
Change Blowgun in [Guide to Combat](https://wiki.yogstation.net/wiki/Guide_to_Combat) to indicate there is no need to stand still, no alert, and increase values from 20 to 25 for both Oxygen and Stamina.

# Changelog
:cl:  
tweak: Blowgun now shoots instantly, but only has a range of 2.
tweak: Blowgun causes 25 oxygen and 25 stamina damage per usage to user instead of 20.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition/special/syringe.dm
+++ b/code/modules/projectiles/ammunition/special/syringe.dm
@@ -60,3 +60,25 @@
 		S.forceMove(D)
 		D.injector = S
 	..()
+
+/obj/item/ammo_casing/blowgun
+	name = "blow gun spring"
+	desc = "A low-power spring that throws syringes a short distance." // how does a blowgun have a spring
+	projectile_type = /obj/item/projectile/bullet/reusable/dart/syringe/blowgun
+	firing_effect_type = null
+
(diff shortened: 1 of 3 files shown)


Maia-J:
So it is still unviable?

Runian:
> So it is still unviable?

It is basically a makeshift item after all. If you don't have easy access to a syringe gun and still want to inject someone with chemicals, you got to accept some higher level of risk to get it done.

ToasterBiome:
how will you abuse this

ToasterBiome (approved):
i guess we'll find out how

PR #19511, opened 2023-07-03
Gives storage implants a visual tell

ynot01:

Metashield on this is incredibly muddy compared to other implants

This will force players to break line of sight to store/take out things or risk getting their implant spotted

# Changelog

:cl:  
tweak: Storage implant now has a dark red visual tell around the user when items are put in or taken out
/:cl:

### DIFF
--- a/code/datums/components/storage/concrete/implant.dm
+++ b/code/datums/components/storage/concrete/implant.dm
@@ -16,3 +16,30 @@
 		return ..()
 	max_combined_w_class += I.max_combined_w_class
 	max_items += I.max_items
+
+/datum/component/storage/concrete/implant/remove_from_storage(atom/movable/AM, atom/new_location)
+	. = ..()
+	flash_out(new_loc = new_location)
+	if(AM.GetComponent(/datum/component/storage))
+		UnregisterSignal(AM, list(COMSIG_STORAGE_INSERTED, COMSIG_STORAGE_REMOVED))
+
+/datum/component/storage/concrete/implant/handle_item_insertion(obj/item/I, prevent_warning, mob/M, datum/component/storage/remote)
+	. = ..()
+	flash_in(inserter = M)
+	if(I.GetComponent(/datum/component/storage))
+		RegisterSignal(I, COMSIG_STORAGE_INSERTED, PROC_REF(flash_in))
+		RegisterSignal(I, COMSIG_STORAGE_REMOVED, PROC_REF(flash_out))
+
+/datum/component/storage/concrete/implant/proc/flash_in(datum/source, obj/item/I, mob/inserter)
+	var/turf/new_turf = get_turf(inserter)
+	if(!new_turf || !istype(new_turf))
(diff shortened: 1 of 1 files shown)


azzzertyy:
this makes unmetashielding this not detective work but watching them and waiting until they fuck up rather than doing any detective work, i'd rather sec have to actually try to figure out what happened to an item they had that dissapeared than just follow him around until he breaks his veil

DeadlyBag:
> this makes unmetashielding this not detective work but watching them and waiting until they fuck up rather than doing any detective work, i'd rather sec have to actually try to figure out what happened to an item they had that dissapeared than just follow him around until he breaks his veil

Sec has little to no way of "figuring" out if items went into storage implants, this is fine imo but theres no mechanic that allows the det to figure out if a storage implant was used except for literally seeing someone use it infront of him. Perhaps i'm just unaware of a mechanic?

Personally dont think this change will affect much at all and im neutral on it

ynot01:
> i'd rather sec have to actually try to figure out what happened to an item they had that dissapeared

can you describe a scenario in which this happens that doesn't break metashield rules

Therandomhoboo:
That would be so noob-bait and just a pain honestly. Just rather you make it a sound indicator or text indicator than a massive visual tell. Storage implant already got a slap for how much they can store and iirc you can sus out if its an storage implant if you don't break line of sight and the items are "suddenly" missing after checking the area.

ynot01:
> Just rather you make it a sound indicator or text indicator than a massive visual tell. 

Sound is heard through walls and arguably text would be more telling than a quick, dark visual effect. The chat is flooded with enough shit anyways

> Storage implant already got a slap for how much they can store

source: image

> and iirc you can sus out if its an storage implant if you don't break line of sight and the items are "suddenly" missing after checking the area.

"iirc" but you don't actually know, you're not confident in that answer. i'm council and i still have to sit down with other staff members to figure out what constitutes a storage implant metashield break. it's all muddy. that's why i made this PR

Therandomhoboo:
> "iirc" but you don't actually know, you're not confident in that answer. i'm council and i still have to sit down with other staff members to figure out what constitutes a storage implant metashield break. it's all muddy. that's why i made this PR

Fair enough on the last part, its always been up in the air but an admin (god knows who now) did say if you don't break line of sight and say they had a spare id and its now "suddenly" gone, that would be premise for an implant check. I do agree though storage implant has this extreme grey area when it comes to finding them out. 

Just don't know if the visual tell would be the best way because to me its meant to be stealthy, either way its probably this or removal at this rate I guess so it is what it is.

Reed0506:
interesting idea, but i feel like having the entire character flash red is pretty silly, maybe make the tell only visible in a smaller radius, similar to how whispering works.

LaOubliette:
this is based as fuck and I want it to lead to an expanded section of forensics

ToasterBiome (approved):

PR #19524, opened 2023-07-04
Gamemode debranching pt 1. Cult vars

JohnFulpWillard:

Gamemodes are meant to implement the antagonists to a round and be editable by admins (obviously), but then there's this
image

This is every var that is on ``/datum/game_mode``. Not their respective gamemodes, not the antagonists, base gamemode. This is so Dynamic can still reference them, but it means that these are all still lingering around doing jack shit and being completely unused even when they aren't the selected gamemode. This is a very dumb way to handle gamemode stuff, and leads to being harder to maintain, easy to break, and even harder to unit test against.

My goal is to break this up, which helps with above, but also for my personal gain, as it will make https://hackmd.io/iTEtaNhkTgKZGF0iuNCKUQ all the more easier to implement.

This PR is removing, from the image, all vars related to Blood cult and Clock cult, moving them to being handled by the Antagonist and its team, while updating a ton of other things to be on feature parity with TG (again, making my job in the future easier for myself).
This one will probably be the largest, as blood cult and clock cult are the largest offenders and span many, many different files. Nearly all other antagonists are significantly smaller and do not rely on it as much as cult does.

# Changelog

:cl:  
experimental: Blood Cult and Clock Cult's handling of members has been changed. Report any unusual behavior like Runes, Items, or Spells not working as intended.
/:cl:


ToasterBiome:
Only 2 drafts allowed open, please close/finish either of your other ones first

JohnFulpWillard:
that's a lot of conflicts nevermind

PR #19541, opened 2023-07-06
another attempt at fixing cult

JohnFulpWillard:

The error we get is that they physically cannot find the objective, which should never be the case as it seems to be given just fine, so my assumption is that it's being checked before the round even starts. This is a problem in other gamemodes but theyve been fixed because they respect parent, cult doesn't seem to.

# Changelog

:cl:  
bugfix: Cult works now... for real.
bugfix: Cult's antag info button now starts with the objectives, rather than having to close and reopen it to see them.
bugfix: Admins can now see Cult objectives in their TP
bugfix: IPCs can no longer roll cult target
bugfix: Cult's objective text no longer has a random <b
/:cl:

### DIFF
--- a/code/datums/actions/items/cult_dagger.dm
+++ b/code/datums/actions/items/cult_dagger.dm
@@ -10,7 +10,7 @@
 	default_button_position = "6:157,4:-2"
 
 /datum/action/item_action/cult_dagger/Grant(mob/grant_to)
-	if(!iscultist(M))
+	if(!iscultist(grant_to))
 		return
 
 	return ..()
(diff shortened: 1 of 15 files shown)


JohnFulpWillard:
Added some extra fixes with this
image

Players from the very beginning will see their objectives in their antag info button, they don't need to close and reopen the tab
Also makes these objectives visible in the admin panel without having to snowflake add objectives into the datum from their team.
^^ I DID THIS FOR ALL ANTAGS. NOT JUST CULT

Also fixes IPCs being selected as the target
Fixes the bad text in the objective desciption.

JohnFulpWillard:
Other examples

Abductor
image

Traitor (not showing Team objectives, as they arent in one)
image

ynot01 (approved):
probably works idk

adamsong:
You broke the 'Convert to traitor' button

PR #18514, opened 2023-04-11
adds antagonists to the game

tattax:
i'll fill this out later while i test stuff but i need this out to see merge conflicts

EDIT V10 GO:

# Document the changes in your pull request

image

changes how antag huds are handled, makes a new file, organizes everything alphabetically

kills proc holders in exchange for /datum/action/cooldown/spells, changes how devil, hivemind and sling is handled +
everything else basically

changes up some bloodsucker things and add some new sprites, this was supposed to be a bloodsucker PR but we don't have so many things i'd be better to just port things, still ports relevant bloodsucker PRs from fulp that refactor how things work.

ports https://github.com/tgstation/tgstation/pull/67083,
https://github.com/tgstation/tgstation/pull/56958,
https://github.com/tgstation/tgstation/pull/61077,
https://github.com/fulpstation/fulpstation/pull/783,
https://github.com/fulpstation/fulpstation/pull/946,
https://github.com/fulpstation/fulpstation/pull/559,
https://github.com/fulpstation/fulpstation/pull/615,
https://github.com/tgstation/tgstation/pull/61087,
https://github.com/tgstation/tgstation/pull/62718,
https://github.com/fulpstation/fulpstation/pull/884,
https://github.com/tgstation/tgstation/pull/65180,

Closes https://github.com/yogstation13/Yogstation/issues/18572
Closes https://github.com/yogstation13/Yogstation/issues/18382
Closes https://github.com/yogstation13/Yogstation/issues/18467 ?
Closes https://github.com/yogstation13/Yogstation/issues/18345 
Closes https://github.com/yogstation13/Yogstation/issues/18344
Closes https://github.com/yogstation13/Yogstation/issues/18286 ?
Closes https://github.com/yogstation13/Yogstation/issues/18116
Closes https://github.com/yogstation13/Yogstation/issues/18090
Closes https://github.com/yogstation13/Yogstation/issues/18004
Closes https://github.com/yogstation13/Yogstation/issues/17155
Closes https://github.com/yogstation13/Yogstation/issues/16612
Closes https://github.com/yogstation13/Yogstation/issues/16210
Closes https://github.com/yogstation13/Yogstation/issues/15814 ?
Closes https://github.com/yogstation13/Yogstation/issues/15386
Closes https://github.com/yogstation13/Yogstation/issues/15455
Closes https://github.com/yogstation13/Yogstation/issues/15219
Closes https://github.com/yogstation13/Yogstation/issues/11478
[...]

PR #19545, opened 2023-07-06
Cyborgs now have TRAIT_FORCED_STANDING

Runian:
Cyborgs now start with ``TRAIT_FORCED_STANDING`` in order to counteract the forced resting that was caused by the #18514. The forced rest caused projectiles to be shot over dead Cyborgs and forced those who dragged any dead Cyborg to be slowed down as if they were dragging someone that was sideways.

Tested in local as dead cyborgs are no longer walk through and you can shoot/throw stuff at their dead body instead of over.
# Changelog
:cl:  
bugfix: Cyborgs are no longer mysteriously lying down even though their sprite appears to be standing up.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/death.dm
+++ b/code/modules/mob/living/silicon/robot/death.dm
@@ -26,8 +26,6 @@
 
 	. = ..()
 
-	remove_traits(list(TRAIT_HANDS_BLOCKED, TRAIT_INCAPACITATED, TRAIT_FLOORED, TRAIT_CRITICAL_CONDITION), STAT_TRAIT) // to remove unwanted traits from living/set_stat
-
 	locked = FALSE //unlock cover
 
 	if(!QDELETED(builtInCamera) && builtInCamera.status)
--- a/code/modules/mob/living/silicon/robot/death.dm
+++ b/code/modules/mob/living/silicon/robot/death.dm
@@ -26,6 +26,8 @@
 
 	. = ..()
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/robot/death.dm
```diff
@@ -26,6 +26,8 @@
 
 	. = ..()
 
+	remove_traits(list(TRAIT_HANDS_BLOCKED, TRAIT_INCAPACITATED, TRAIT_FLOORED, TRAIT_CRITICAL_CONDITION), STAT_TRAIT) // to remove unwanted traits from living/set_stat
```
JohnFulpWillard: When do these traits get added and why do we have to snowflake remove them here?

Runian: These traits are added by [living/set_stat](https://github.com/yogstation13/Yogstation/blob/2a40a7701fe12bdff80dda210ba6d5e7a67d5fe6/code/modules/mob/living/living.dm#L1489). It is called [here](https://github.com/yogstation13/Yogstation/blob/2a40a7701fe12bdff80dda210ba6d5e7a67d5fe6/code/modules/mob/living/death.dm#L52) which was added by the [antag pr](https://github.com/yogstation13/Yogstation/pull/18514). It worked and I couldn't think of a better solution at a time.

JohnFulpWillard: We have a trait that is meant to block this, ``TRAIT_FORCED_STANDING``. It hasn't been implemented properly (likely due to the scope of the PR) so currently does nothing and is the reason why this is a problem.

ToasterBiome (approved):
ok

PR #19563, opened 2023-07-07
Laser pointer no longer uses %chance to recharge and horribly explain it's charge

Chubbygummibear:

Old code was confusing in that it seemed to imply the battery would burn out and be unusable by calling it "recharge locked" but would actually recharge with a 5% chance of gaining a charge per tick, but remain unusable until completely recharged. However you would have no indication of the recharge chance, percent complete, or if charges were even coming back.

Now you can see exactly how many charges a laser pointer has when you're close enough to hold it in your hand, and the rate at which it recharges. Also it doesn't burn out and lock up anymore. if you regenerate 1 charge from being at 0/5, go ahead and use it.
image

# Wiki Documentation
laser pointer is no longer rng, doesn't lock you out of using it if you use up all the charges down to 0

recharge rate determined by:
`recharge_rate = clamp((30 SECONDS - (5 SECONDS * diode.rating)), 1, 30 SECONDS)`

so at most it'll ever be 30 seconds and if an admin var edits your laser rating to 100 for whatever reason it'll recharge basically instantly

# Changelog

:cl:  
rscadd: Laser pointer examine text now shows you the number of charges it has and can have, as well as the recharge rate in seconds  
rscadd: Laser pointer generates a charge every 30-5*diode rating seconds. so 25 seconds at base, 10 seconds at tier 4 part
rscdel: Laser pointer batteries don't "burn out" anymore and lock you out of using them until fully recharged, and the recharge process is no longer % based chance
bugfix: The laser pointer process would kill itself on every run because it called the parent proc which specifically says "never do this" 

/:cl:

### DIFF
--- a/code/game/objects/items/devices/laserpointer.dm
+++ b/code/game/objects/items/devices/laserpointer.dm
@@ -11,13 +11,16 @@
 	materials = list(/datum/material/iron=500, /datum/material/glass=500)
 	w_class = WEIGHT_CLASS_SMALL
 	var/turf/pointer_loc
-	var/energy = 5
-	var/max_energy = 5
+	var/charges = 5
+	var/max_charges = 5
 	var/effectchance = 33
-	var/recharging = 0
+	//var/recharging = 0
 	var/recharge_locked = FALSE
-	var/obj/item/stock_parts/micro_laser/diode //used for upgrading!
-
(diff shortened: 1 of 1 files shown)


azzzertyy:
what maniac coded this originally my god no wonder they were so weird

Moltijoe (approved):

PR #19577, opened 2023-07-08
Ultra-Violence IPC heads are no longer damage immune after being roughed up a bit

Mqiib:

IPC heads now have 200 max damage when they learn the ultra violence martial art. This is because, since they are now immune to dismembering, the normal IPC downside of having lower max health on their head is actually an upside because anyone aiming at their head to kill them ends up doing zero damage after a certain point. Additionally, since IPCs don't have wounds, their head being at max damage essentially has no downside at all, other than the damage itself being present.

# Changelog

:cl:  

tweak: Ultra-violent IPCs now can actually die from their head being smashed in enough

/:cl:

### DIFF
--- a/code/datums/martial/ultra_violence.dm
+++ b/code/datums/martial/ultra_violence.dm
@@ -409,6 +409,8 @@
 	add_verb(H, recalibration)
 	usr.click_intercept = src //probably breaks something, don't know what though
 	H.dna.species.GiveSpeciesFlight(H)//because... c'mon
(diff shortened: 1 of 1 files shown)


Moltijoe (changes requested):
Remove it if they somehow lose the martial art

Moltijoe (changes requested):
Also make sure it doesn't runtime if they're missing their head somehow when getting the martial art

Moltijoe:
Reopen this when you plan to make changes

PR #19578, opened 2023-07-08
People (me) can no longer make infinite error sprites with a stamp, paper, and some toner.

azzzertyy:
If a paper had enough stamps overlayed ontop of it, it would give an errorsprite.
Now it is capped at 25 stamps on a single piece of paper, should be more than enough really.

Also unrelated but changes the mythical maint syndicate stamp loot to syndiround, since its supposed to be incredibly rare, the normal syndi stamp is available on the station on every round by default.

# Changelog
:cl:  
bugfix: you can no longer make infinite error signs out of paper
tweak: syndiround is now available in maintloot instead of non round syndi stamp
/:cl:

### DIFF
--- a/code/modules/paperwork/paper.dm
+++ b/code/modules/paperwork/paper.dm
@@ -46,6 +46,7 @@
 	var/contact_poison // Reagent ID to transfer on contact
 	var/contact_poison_volume = 0
 	var/next_write_time = 0 // prevent crash exploit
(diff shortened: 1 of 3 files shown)


azzzertyy:
image

it was good while it lasted....

Maia-J:
P sure this is some bug with byond itself so this is the only way to really fix it

ToasterBiome (approved):

PR #19580, opened 2023-07-08
Explicitly specify extools in SpacemanDMM configuration

SpaceManiac:
I want to make auxtools the default in SpacemanDMM, to avoid confusing new users. This of course will require remaining users of extools to specify such explicitly. Yogstation appears to be the largest holdout.

I would like to someday remove extools support entirely, but that is for another time.

### DIFF
--- a/SpacemanDMM.toml
+++ b/SpacemanDMM.toml
@@ -1,2 +1,5 @@
 [langserver]
 dreamchecker = true
+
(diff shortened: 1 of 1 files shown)


ToasterBiome:
Please just kill us already

ToasterBiome (approved):

PR #19581, opened 2023-07-09
Attacking a limb at its damage cap applies some damage to the chest

SapphicOverload:

Attacking a limb that has reached its damage cap will deal 33% of the damage to the chest instead. Mainly so that attacking a limb will still cause some damage instead of just being invincible as long as only that limb is hit.

# Changelog

:cl:  
tweak: hitting limbs at their damage cap applies some damage to the chest
/:cl:


DylonBR:
yog has fallen, millions must hugbox.

neeshacark:
wouldnt this make it harder to sever limbs?

iamthesilly:
> wouldnt this make it harder to sever limbs?

No, it does damage to the chest on top of wounding the targeted limb

ToasterBiome (approved):

PR #19582, opened 2023-07-09
mech rcd cant deconstruct reinforced stuff

SomeguyManperson:

reboot of #9084 does the exact same thing but again

Any mech has the capacity to smash down an rwall or reinforced airlock already, either by punching it, missiles, or using a strong enough drill. The only one that can't do this is the odysseus which doesn't get to break rwalls because it's the medical mech.
These all require getting in melee range and/or sacrificing an equipment slot for specifically breaking reinforced structures, rather than slapping on the wall builder 6000 and getting to remotely deconstruct a department as a free side effect

# Spriting

# Wiki Documentation

mech RCD can't break rwalls, rfloors, or airlocks with plasteel reinforcement

# Changelog

:cl:  
rscdel: the mech RCD can no longer deconstruct reinforced walls, floors, or airlocks. Use the drill, for once.
/:cl:

### DIFF
--- a/code/game/mecha/equipment/tools/work_tools.dm
+++ b/code/game/mecha/equipment/tools/work_tools.dm
@@ -1,3 +1,6 @@
+#define DECONSTRUCT 0
+#define WALL 1
+#define AIRLOCK 2
 
 //Hydraulic clamp, Kill clamp, Extinguisher, RCD, Cable layer.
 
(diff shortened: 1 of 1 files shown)


LassiH:
what about the mime mech will it still be able to deconstruct rwalls since i dont see how smashing down a wall is stealthy

ToasterBiome (approved):

PR #19584, opened 2023-07-09
Flying fang no longer completely prevents use of outerwear slot, instead disables certain combos/pounce depending on equipment

SomeguyManperson:

Extreme downside that massively decreases general utility of the martial art while not providing much in the way of benefits (you still die roughly as fast as anyone else, just a little slower, unlike certain martial arts)

instead certain abilities are disabled depending on clothing
Neck bite: requires mouth to be unobstructed (this is unrelated, as you could still wear a mask, but I don't think anyone will complain)
Tail whip: requires tail to be visible
Pounce: requires no thick clothing (hardsuits, biosuits, etc)

# Spriting

# Wiki Documentation

# Changelog

:cl:  
rscdel: flying fang no longer removes your ability to wear armor and hardsuits
tweak: flying fang's bite now requires you mouth to be exposed, pounce requires your clothing to not be thick, and tail whip requires your tail to... not be hidden by clothing
/:cl:

### DIFF
--- a/code/datums/martial/flying_fang.dm
+++ b/code/datums/martial/flying_fang.dm
@@ -91,7 +91,7 @@
 /datum/martial_art/flyingfang/proc/Chomp(mob/living/carbon/human/A, mob/living/carbon/human/D)
 	if(!can_use(A))
 		return
-	if(is_mouth_covered(M, user))
+	if(A.is_mouth_covered())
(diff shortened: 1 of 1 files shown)


thegoldencat413:
yay!

Moltijoe (approved):

PR #19587, opened 2023-07-09
Gives the other preternis martial art some love

Moltijoe:
You might be asking yourself "Which martial art is that?" exactly
Really just some minor QoL with a rename/reflavour

:cl:  @Moltijoe @solwashere for the lore accurate name
tweak: Renamed "Stealth" to "Remnant Liquidator"
tweak: fingergun combo paralyze is 1 second rather than scaling on punch damage to roughly equal 1 second
tweak: lets you hide fingergun by trying to drop it
tweak: makes fingergun instantly hide after shooting
/:cl:


solwashere:
wow such a cool name I wonder who helped with it

ToasterBiome (approved):
why do they get two

PR #19591, opened 2023-07-09
Removes legacy yogtoggles, mood is enabled by default.

JohnFulpWillard:

Removes the deprecated yogtoggles stuff and moves it to actual preference toggles.
Because of this, I also decided to make mood preference enabled by default, however if you had it disabled before it will remain disabled.

Mood is a big part of SS13, and especially so if they are coming from codebases like TG or Bee (so a large majority of them), having a familiar mechanic by default will be helpful to ease a transition between codebases. This remains as an optional thing, it just sucks that you have to go out of your way to enable this feature.

# Wiki Documentation

A mention about mood for new players? I guess

# Changelog

:cl:  
add: Mood is now enabled in your prefs by default.
experimental: Refactored yog legacy toggles to use actual toggles.
/:cl:


ToasterBiome:
This PR should really be atomized

JohnFulpWillard:
> This PR should really be atomized

Just tell me what parts you don't want added to the game

Bibby0110:
> > This PR should really be atomized
> 
> Just tell me what parts you don't want added to the game

Mood

azzzertyy:
yeah, mood is not a big part of this server at all and we dont like it very much, ik i dont, having it enabled by default doesnt really accustom them to our server,

JohnFulpWillard:
Having mood enabled by default is the whole reason I made this refactor in the first place, I had to change it because the old system wouldn't let me have it on by default. 

Also having something disabled by default is essentially removing the feature entirely. For something like mood, it is way better to be an opt-out than an opt-in, it's a feature that we should want players to use because we go out of our way to add mood interactions with stuff and would like to see them actually seen by players. 

> we dont like it very much

Then don't use it, this PR isn't forcing mood into you. If you had mood disabled it will remain disabled.

ToasterBiome:
> > This PR should really be atomized
> 
> Just tell me what parts you don't want added to the game

removing the toggles and setting mood to default on should be split

JohnFulpWillard:
I don't agree

ToasterBiome (approved):
grrr ruff ruff ruff

PR #19600, opened 2023-07-09
Remove "Sheltered" trait

Amelia0010:

This pull request removes the sheltered quirk from the codebase.

My justification for this is twofold:

1:
I do not think it adds much to the game, to have a group of people who can only speak to one another. I think that role should belong only to the mime, who is already unable to speak. This is ultimately a value judgement, and I happen to believe that this really does not add much value. 
I believe it actively takes value away, in that it adds disabilities which can be not just the user's problem. For example, a head of staff having this quirk enabled messes with everyone else as well, not just this person.

2:
I think they should not know *Japanese*. This doesn't necessarily require the quirk to be deleted, but I'm unsure what else to replace the language which. I'd be happy with another language taking its place, but I'm not really that fond of *ninjas* and people with a "negative" quirk speaking the same language. Especially not when that language was meant to make it so the *ninja* couldn't communicate.

# Wiki Documentation

https://wiki.yogstation.net/wiki/Quirks should be updated

# Changelog

:cl:  
rscdel: Removed "Sheltered" trait
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -230,7 +230,6 @@
 #define TRAIT_UNINTELLIGIBLE_SPEECH "unintelligible-speech"
 #define TRAIT_UNSTABLE			"unstable"
 #define TRAIT_OIL_FRIED			"oil_fried"
-#define TRAIT_SHELTERED			"sheltered"
 #define TRAIT_RANDOM_ACCENT		"random_accent"
 #define TRAIT_MEDICALIGNORE     "medical_ignore"
 #define TRAIT_PASSTABLE			"passtable"
--- a/code/datums/traits/negative.dm
+++ b/code/datums/traits/negative.dm
@@ -626,28 +626,6 @@
 	lose_text = span_notice("Your mind finally feels calm.")
 	medical_record_text = "Patient's mind is in a vulnerable state, and cannot recover from traumatic events."
 
-/datum/quirk/sheltered
-	name = "Sheltered"
-	desc = "You never learned to speak galactic common."
-	icon = "comment-dots"
-	value = -2
-	mob_trait = TRAIT_SHELTERED
-	gain_text = span_danger("You do not speak galactic common.")
-	lose_text = span_notice("You start to put together how to speak galactic common.")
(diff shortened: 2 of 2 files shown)


athena413:
Isn't there a 'british' language, for #2?

Amelia0010:
There is, not particularly fond of the syllable list (and as far as I can understand it's currently a species-specific language to Humans), but that could be doable. 

For the record, #2 is my main gripe, so another language is very much also welcome (to me, i'm not sure if others dislike #1 as much as I do)

azzzertyy:
finally someone is based enough
this was quickfire voted ages ago but they were wrong and it should never have been added in the first place

ToasterBiome:
Nobody actually speaks japanese because we have race languages for every playable race now (even regular human)

Hislyna:
I completely disagree on both points. For starters they do not learn any new language, they just lose the ability to speak common. Second, I have had great RP moments using sheltered amd using the Me" function. It brings a new way to RP and gets around its challenges. Your reasoning for its bad for heads makes no sense when blind, deaf, and pacifist exist which you can just block heads from using instead.

Hislyna:
Considering you thought they knew japanese and didnt know they understand common, when is the last time you looked at sheltered?

neeshacark:
for one you could just blacklist from head roles

two literally isn't true

Amelia0010:
```
/datum/quirk/sheltered/on_clone(data)
	var/mob/living/carbon/human/H = quirk_holder
	H.remove_language(/datum/language/common, FALSE, TRUE)
	if(!H.get_selected_language())
		H.grant_language(/datum/language/japanese)

/datum/quirk/sheltered/on_spawn()
	var/mob/living/carbon/human/H = quirk_holder
	H.remove_language(/datum/language/common, FALSE, TRUE)
	if(!H.get_selected_language())
		H.grant_language(/datum/language/japanese)
```

Amelia0010:
oh wait they only get Japanese if they don't have another language and nowadays *everything* has a second language. I mean that seems silly to me but w/e my main problem is no longer valid

PR #19606, opened 2023-07-10
Rebalances the ultra violence revolver

SapphicOverload:

Reworks the ultra violence revolver bullets to be less damaging and give less style bonus, but gives them the ability to ricochet. They can still pierce mobs, but if they do so they lose their ability to ricochet off of walls. The bullets will do more damage and move faster if you ricochet them first, encouraging users to take their time to line up a shot instead of spamming the projectile.

Specifics:
-30 damage (from 40)
-0.1x damage style bonus (from 0.2x damage)
-can ricochet once unless it pierced a mob
-can pierce mobs unless it has ricocheted before
-gains +10 damage and increased speed on ricochet
-loses 20 damage after each mob pierced

Effects on balance:
-Armor is now a more viable counter
-Ultra Violence revolver is less effective when spamming it
-Adds incentive to use the revolver in a more fun and skill-based way

# Changelog

:cl:
tweak: ultra violence revolver damage reduced to 30 and gives less style per damage
tweak: ultra violence revolver can ricochet off of any wall once, unless it pierced a mob first
tweak: ultra violence revolver bullet gains some damage and speed on ricochet, but loses damage after piercing a mob
bugfix: fixed a duplicate definition in ultra_violence.dm
bugfix: fixed ultra violence revolver getting stuck to your belt
/:cl:

### DIFF
--- a/code/datums/martial/ultra_violence.dm
+++ b/code/datums/martial/ultra_violence.dm
@@ -180,10 +180,14 @@
 		var/mob/living/carbon/human/H = target
 		H.add_splatter_floor(H.loc, TRUE)//janitors everywhere cry when they hear that an ipc is going off
 	ricochets = ricochets_max // so you can't shoot through someone to ricochet and hit them twice for 70 damage in one shot
+	damage -= 20
+	if(damage <= 0)
+		qdel(src)
 
 /obj/item/projectile/bullet/ipcmartial/on_ricochet(atom/A)
 	damage += 10 // more damage if you ricochet it, good luck hitting it consistently though
 	speed *= 0.5 // faster so it can hit more reliably
+	penetrating = FALSE
 	return ..()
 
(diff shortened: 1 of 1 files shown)


LassiH:
this seems like a bad idea for 2 reasons 1 you can hit your self or random people down the hall
2 arent there already bonuses for hitting coins on the floor that redirect the bullet to a person

iamthesilly:
> you can hit your self

Skill issue, don't spam it without thinking

> 2 arent there already bonuses for hitting coins on the floor that redirect the bullet to a person

Yes, this allows it to work on walls too with a lower damage bonus and no auto aim

LassiH:
will ricochet shots give you more style than non ricochet shots ?

ToasterBiome (approved):

PR #19619, opened 2023-07-10
Lowers the price of Horror-in-a-box traitor item

PibeCantina:

Lower price of horror in the box to 14 tc, so you can buy something nice on the side, and because eldrich horror is just simply not powerful enough to be worth 16 tc.

Did not test it, but it's a simple change so i hope this works (first time doing this)

# Wiki Documentation
Lower the price on the item from 16 to 14

# Changelog
:cl:  
tweak: Lowers the price of horror in a box  
/:cl:


Lichton:
Now you can buy a G.O.R.L.E.X blade, to go with your tentacle arm! Also, agree with this change- you're basically defenseless while your horror goes and harvests souls, so...

Moltijoe (approved):

PR #19629, opened 2023-07-12
Nerfs blazing oil again

Mqiib:
Despite more nerfs it's still absolutely top-tier with the powers of lighting people on fire (extended damage), dealing burn damage (most species take extra burn damage), and being COMPLETELY IMMUNE to laser fire (and getting free shit from it) so now it takes DOUBLE damage from brute and ~50% more damage from extinguishers. This still may not be enough.

# Wiki Documentation

2x brute modifier

# Changelog

:cl:  

tweak: Blazing oil strain now takes more (1.5x -> 2x) brute damage
tweak: Blazing oil strain now takes ~50% more damage from extinguishers
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
+++ b/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
@@ -14,11 +14,11 @@
 	reagent = /datum/reagent/blob/blazing_oil
 
 /datum/blobstrain/reagent/blazing_oil/extinguish_reaction(obj/structure/blob/B)
-	B.take_damage(1.3, BURN, ENERGY)
+	B.take_damage(2, BURN, ENERGY)
 
 /datum/blobstrain/reagent/blazing_oil/damage_reaction(obj/structure/blob/B, damage, damage_type, damage_flag)
 	if(damage_type == BRUTE) 
-		return damage * 1.5
+		return damage * 2
 	if(damage_type == BURN && damage_flag != ENERGY)
 		var/mob/camera/blob/O = overmind
 		O.add_points(damage / 10)//burn damage causes the blob to gain a very small amount of points: the 20 damage of a laser will generate 2 BP.


Yarinoi:
ided pr

i do think that round the blobbernauts werent taking damage from extinguisher spray, genuinely should fix THAT instead of this shit

RG4ORDR:
> 

When the counter for the blob literally deals such little damage to BASE blob titles and minute damage to blobbernauts it's not an ided thing.

Marmio64:
this strain used to be useless when the extinguisher damage was 1.5, and now it is going to be made 2? Extinguishers can hit a dozen blob tiles at once and are the most easily obtainable antiblob weapon, more so than welders because they're just sitting around on walls literally everywhere, they shouldn't deal this much damage even if it is meant to be a counter.

Mqiib:
> this strain used to be useless when the extinguisher damage was 1.5, and now it is going to be made 2? Extinguishers can hit a dozen blob tiles at once and are the most easily obtainable antiblob weapon, more so than welders because they're just sitting around on walls literally everywhere, they shouldn't deal this much damage even if it is meant to be a counter.

You forget, that was when blob tiles blocked the spray and would take a bunch of damage from a single spray. They don't any more, so you're fighting with the regeneration of ALL of those sprayed tiles, not just the one, while having the same damage output. Also being able to remove blob tiles that are nearby quickly is usually more relevant than being able to damage a bunch of them equally, again particularly when your actual total damage isn't changing.

Marmio64:
> > this strain used to be useless when the extinguisher damage was 1.5, and now it is going to be made 2? Extinguishers can hit a dozen blob tiles at once and are the most easily obtainable antiblob weapon, more so than welders because they're just sitting around on walls literally everywhere, they shouldn't deal this much damage even if it is meant to be a counter.
> 
> You forget, that was when blob tiles blocked the spray and would take a bunch of damage from a single spray. They don't any more, so you're fighting with the regeneration of ALL of those sprayed tiles, not just the one, while having the same damage output. Also being able to remove blob tiles that are nearby quickly is usually more relevant than being able to damage a bunch of them equally, again particularly when your actual total damage isn't changing.

Then if you're making extinguishers be good why also make brute far better? It is kind of encroaching into replicating foam as well, another strain that benefits from getting hit by burn attacks while taking double brute damage.

ToasterBiome:
0 likes i aint touching this

PR #19632, opened 2023-07-12
makes atom/movable/screen setup hud on new instead of relying on update icon

tattax:
it was done on the update appearance pr but it's neat and the pr is not getting anywhere until auxtools


JohnFulpWillard (approved):
this is big

PR #19644, opened 2023-07-13
Holy Light now requires a small sacrifice to bible heal

solwashere:

When you hit someone with your bible, a third of that damage will now return to you in the form of burn damage.

# Spriting

# Wiki Documentation

Make mention that healing with the holy light book will now deal a small amount of burn damage to you

# Changelog

:cl:  

tweak: Holy Light healing with the bible will now return 1/3rd of that damage to the user in the form of burn damage.

/:cl:

### DIFF
--- a/code/modules/religion/religion_sects.dm
+++ b/code/modules/religion/religion_sects.dm
@@ -455,6 +455,7 @@
 		H.update_damage_overlays()
 
 		adjust_favor(amount_healed, user)
+		user.apply_damage((amount_healed / 3), CLONE)
 		H.visible_message(span_notice("[user] heals [H] with the power of [GLOB.deity]!"))
 		to_chat(H, span_boldnotice("May the power of [GLOB.deity] compel you to be healed!"))
 		playsound(user, 'sound/magic/staff_healing.ogg', 25, TRUE, -1)
--- a/code/modules/religion/religion_sects.dm
+++ b/code/modules/religion/religion_sects.dm
@@ -455,7 +455,7 @@
 		H.update_damage_overlays()
 
 		adjust_favor(amount_healed, user)
-		user.apply_damage((amount_healed / 3), CLONE)
+		user.apply_damage((amount_healed / 3), BURN, BODYPART_ANY)
 		H.visible_message(span_notice("[user] heals [H] with the power of [GLOB.deity]!"))
 		to_chat(H, span_boldnotice("May the power of [GLOB.deity] compel you to be healed!"))
 		playsound(user, 'sound/magic/staff_healing.ogg', 25, TRUE, -1)


UltraBigRat:
This sounds like itd just make holy light never played, because clone is stupidly hard to heal  and the bible heal is necesarry to get points, and i dont feel its fair to make chaplains go  into a cryotube that may or may not be staffed every few minutes,   for them doing their sects job

solwashere:
> This sounds like itd just make holy light never played, because clone is stupidly hard to heal and the bible heal is necesarry to get points, and i dont feel its fair to make chaplains go into a cryotube that may or may not be staffed every few minutes, for them doing their sects job

https://github.com/yogstation13/Yogstation/pull/19413

Moltijoe:
> #19413

relying on a single chemical to make a sect valid is not proper balance

solwashere:
> > #19413
> 
> relying on a single chemical to make a sect valid is not proper balance

having a single job be able to cut into an entire departments job is not valid balance

Moltijoe:
> > relying on a single chemical to make a sect valid is not proper balance
> 
> having a single job be able to cut into an entire departments job is not valid balance

then tackle that without effectively deleting a sect
however, i will say that since the rod can heal clone (albiet slowly), it just becomes a rush to the rod and then waiting for it to heal the clone damage

if you have no desire to balance it in a more intuitive way, then just make a PR removing it outright

solwashere:
> > > relying on a single chemical to make a sect valid is not proper balance
> > 
> > 
> > having a single job be able to cut into an entire departments job is not valid balance
> 
> then tackle that without effectively deleting a sect however, i will say that since the rod can heal clone (albiet slowly), it just becomes a rush to the rod and then waiting for it to heal the clone damage
> 
> if you have no desire to balance it in a more intuitive way, then just make a PR removing it outright

true we should completely gimp chaplain healing

Moltijoe:
> true we should completely gimp chaplain healing

go for it, cut the healing in half but make sure to double favour gain to make sure that isn't affected

also, consider the idea of replacing the aheal ritual with something like healing circle or mass heal, where it just heals everyone inside the chapel upon casting

that way they can still heal effectively, but the patient NEEDS to be in the chapel, and chaplain can't go all the way to revival

solwashere:
burn damage now

SomeguyManperson:
Replace annoying damage type with less annoying damage type but anyone using a rod of the white is immune to the healing from rod of the white, so like

Pros of having rod of white: you heal everyone around you passively
Cons of having rod of the white: you get no passive healing from rod of the white, and need to use more conventional forms of healing if you get injured
Your roundstart pacifism is doubled

Hislyna:
Why not make it so that you take x of the health on yourself instead? Thatd be alot better than this

Hislyna:
33% is too high. Edit the cooldown, edit the favour for RoW, do as above where RoW no longer healers the one wearing.

solwashere:
> 33% is too high. Edit the cooldown, edit the favour for RoW, do as above where RoW no longer healers the one wearing. 

Every other sect has some form of downside to it, this one doesn't. Even if I made the RoW the way it is suggested, it doesn't change the fact that the bible is infinite healing with no penalty towards anyone. At least this way it stays thematically consistent.

code/modules/religion/religion_sects.dm
```diff
@@ -455,6 +455,7 @@
 		H.update_damage_overlays()
 
 		adjust_favor(amount_healed, user)
+		user.apply_damage((amount_healed / 3), BURN, BODYPART_ANY)
```
SomeguyManperson, line 458: This should have some feedback and/or player-facing documentation I think

neeshacark:
Not a fan of using burn given most species are vulnerable to it in some way.  Brute would probably be the best because not many species have modifiers to it.  Also yeah 33% is probably too high for anybody without some way to heal.  They'll be walking back to medbay every two minutes - and that'll end poorly.

SomeguyManperson:
> and that'll end poorly.

what does this even mean

SomeguyManperson:
Unfortunate thing about holy light is it's entirely based around healing; the bible heal produces resources for it rather than consuming them, leading to people feeling entitled and required to make use of it (need to bible heal to take advantage of favor), and its effects are not symbiotic to the medical team (removes work from medical team rather than augmenting it/making it more efficient)

solwashere:
A better solution was brought up, will get to it whenever I get the time

PR #19645, opened 2023-07-13
Fixes wacky living death signal

JohnFulpWillard:

Fixes global death signal being sent as a non-global signal, and being sent twice.
This is just a thing I need for devil affairs, but fixes

# Changelog

:cl:  
fix: People no longer die twice.
/:cl:

### DIFF
--- a/code/modules/mob/death.dm
+++ b/code/modules/mob/death.dm
@@ -10,5 +10,5 @@
 	return
 
 /mob/proc/death(gibbed)
-	SEND_SIGNAL(src, COMSIG_GLOB_MOB_DEATH, gibbed)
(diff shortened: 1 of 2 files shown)


code/modules/mob/death.dm
```diff
@@ -10,5 +10,5 @@
 	return
 
 /mob/proc/death(gibbed)
-	SEND_SIGNAL(src, COMSIG_GLOB_MOB_DEATH, gibbed)
-	SEND_GLOBAL_SIGNAL(COMSIG_GLOB_MOB_DEATH, src , gibbed)
```
tattax, line 14: remove this, it makes no sense for a mob to die when you can only die if you are living in the first place??

ynot01 (approved):

PR #19647, opened 2023-07-13
Gives every sect bless a cooldown

Moltijoe:
The weird inconsistency of two sects having cooldowns, while others can heal as fast as you click giving them theoretically insane healing output

also creates a few sect variables that were pretty much identical between blessings

:cl:  
tweak: All chaplain sect blessings without cooldowns now have a 10 second cooldown
/:cl:

### DIFF
--- a/code/modules/religion/religion_sects.dm
+++ b/code/modules/religion/religion_sects.dm
@@ -20,6 +20,11 @@
 	var/altar_icon_state // Changes the Altar of Gods icon_state
 	var/list/active_rites // Currently Active (non-deleted) rites
 
+	//bless variables
+	var/heal_amt = 20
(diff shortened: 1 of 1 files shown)


SomeguyManperson:
Don't the two that have it have it since they provide some very specific buff that would be stronger if it was spammable i.e. holy light provides massive burst healing and st credit is stupid

Moltijoe:
> Don't the two that have it have it since they provide some very specific buff that would be stronger if it was spammable i.e. holy light provides massive burst healing and st credit is stupid

The idea is that i add a cooldown and then make the other blessings actually worth using

Moltijoe:
Whatever

PR #19651, opened 2023-07-13
[BOX] Completely overhauls detective's office, and moves it to the other side of EVA.

neeshacark:

This PR completely overhauls both the ~law~ and detective offices, moving detective office a bit to allow him to have a backroom for his gear and surgeries.  Makes his office cozy and adds a fireplace.  ~Lawyers get individual offices for themselves.~  Maint behind eva is adjusted to allow for these changes.

This is good because detective is not *technically* full security, and in its current location the detective office is rarely trafficked by anyone except criminals and security.  This means that in the majority of rounds, the detective is left alone unless someone needs him to click on something with a scanner once (assuming the hos doesnt just steal the dets spare scanner and do it themself).

By moving it to a highly trafficked hall, my hope is that more people will interact and roleplay with the detective, and also in hopes that the new location serve to make people think of detective less as a 'secoff but with a ballistic gun'

~Lawyers get their own space, with a curtain between their offices they may open or close at their choosing.  This will allow some space, and let lawyers choose whether or not they want to have a conjoined law firm or work individually as prosecution and defense (i had this happen a few rounds ago and it was lovely and part of the reason for this pr)~

Lawyers no longer have individual offices due to complaints.
image_2023-07-15_125633930

# Wiki Documentation

Map image updates required

# Changelog

:cl:  cark
mapping: overhauls det and lawyer office
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -2058,18 +2058,10 @@
 "aol" = (
 /turf/open/floor/engine/air,
 /area/engine/atmos/distro)
-"aom" = (
-/obj/structure/fireplace,
-/turf/open/floor/plasteel/grimy,
-/area/security/detectives_office)
 "aoq" = (
 /obj/effect/spawner/structure/window/reinforced/shutter,
 /turf/open/floor/plating,
 /area/maintenance/starboard/fore)
-"aoB" = (
-/obj/machinery/atmospherics/components/unary/vent_scrubber/on/layer4,
-/turf/open/floor/plasteel/grimy,
-/area/security/detectives_office)
 "aoC" = (
 /obj/structure/table,
 /obj/machinery/firealarm{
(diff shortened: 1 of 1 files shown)


neeshacark:
forgot to do image here you go
image_2023-07-13_105213823

neeshacark:
redid law offices because people complained

neeshacark:
grrrr stop disliking and not saying why

Aquizit (changes requested):
Not sure how I feel about this overall, as removing 3 maint rooms feels fairly significant. That little area of maint can be fairly busy so taking so much of it out might be an issue, not sure.

But, I still need to review so:
image
This button seems shifted too far.

image
Trim from old det door into maint.

new det office needs a fire alarm

neeshacark:
> Not sure how I feel about this overall, as removing 3 maint rooms feels fairly significant. That little area of maint can be fairly busy so taking so much of it out might be an issue, not sure.
> 
> But, I still need to review so: image This button seems shifted too far.
> 
> image Trim from old det door into maint.
> 
> new det office needs a fire alarm

This spot used to be the psych office, I think over that we only lost one more room and it was a 3x3.  Will fix issues.

Aquizit (approved):

PR #19653, opened 2023-07-13
makes the improvised pneumatic cannon uncraftable

athena413:

removes the crafting recipe

current exploit makes this weapon instakill/near-death on direct hit instacrit/nearcrit on near miss on a weapon any crewmember can obtain

get a better gimmick

ided pr

will close if an alternative is merged that sucks less

# Spriting
n/a

# Wiki Documentation

https://wiki.yogstation.net/wiki/Makeshift_weapons

# Changelog

:cl:  
rscdel: pneumatic cannon has been made uncraftable
/:cl:

### DIFF
--- a/code/datums/components/crafting/weapons.dm
+++ b/code/datums/components/crafting/weapons.dm
@@ -337,17 +337,6 @@
 	category = CAT_WEAPONRY
 	subcategory = CAT_WEAPON
 
-/datum/crafting_recipe/improvised_pneumatic_cannon //Pretty easy to obtain but
-	name = "Pneumatic Cannon"
-	result = /obj/item/pneumatic_cannon/ghetto
-	tools = list(TOOL_WELDER, TOOL_WRENCH)
-	reqs = list(/obj/item/stack/sheet/metal = 4,
-				/obj/item/stack/packageWrap = 8,
-				/obj/item/pipe = 2)
-	time = 5 SECONDS
-	category = CAT_WEAPONRY
-	subcategory = CAT_WEAPON
-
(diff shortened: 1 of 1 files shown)


LaOubliette:
lets not remove something because literally verbatim ided and "exploit" without even stating what it is

Moltijoe:
> lets not remove something because literally verbatim ided and "exploit" without even stating what it is

pretty sure it's drakes molotov + plasma cigar combo

azzzertyy:
"Current exploit lets-"
then fix it?

Mqiib:
Improvised cannon is too much fun to remove, I'd rather just deal with the actual problem than nuke the weapon as a whole.

LaOubliette:
> > lets not remove something because literally verbatim ided and "exploit" without even stating what it is
> 
> pretty sure it's drakes molotov + plasma cigar combo

that's based though

DeadlyBag:
Understandable PR since there have been a ton of broken things all relating to the pneumatic cannon! (and there are more to come) I think the direction we want to move in is keeping the cannon in and tweaking it to be more bearable or blacklisting certain objects from being added to it.

Moja10:
A better solution would be to let the cannon have "magazine space", and then objects are fired one at a time. That way there's still some reaction to the shenanigans, rather than getting blasted at by 10 things at once. Launching 10 things at once encourages to find and fire the most awful things imaginable with the cannon, where as launching in a sequence can have more interesting results from what you configure the order to be. Adjust gas use to compensate, bam

Bibby0110:
Improve don’t remove

PR #19663, opened 2023-07-14
Fixed targeted limb starting at null 

ynot01:

image

# Changelog

:cl:  
bugfix: Fixed targeted limb starting at null 
/:cl:


tattax:
it should never be null, you should make the original var also chest

JohnFulpWillard:
this is fixed in my update appearance PR and it'll conflict now wtf

Moltijoe (approved):

PR #19669, opened 2023-07-14
Makes bluespace vials equivalent to large vials

ynot01:

Getting an upgrade 30 minutes into the round to usurp one of the deluxe hypospray's main attractions and be *twice as good* feels very silly to me.

It is perfectly reasonable to me to have the tiny-sized vials be upgraded into an equivalent of the large vials, but not surpass them.

As an aside, a "large bluespace vial" for particularly ambitious CMOs could be considered in the future, to allow the deluxe hypospray to upgrade alongside the normal one, but we would need a new sprite for it.

# Changelog

:cl:  
tweak: Bluespace vials now only carry 30 units, like large vials
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/bottle.dm
+++ b/code/modules/reagents/reagent_containers/bottle.dm
@@ -438,7 +438,7 @@
 /obj/item/reagent_containers/glass/bottle/bromine
 	name = "bromine bottle"
 	list_reagents = list(/datum/reagent/bromine = 30)
-	
+
 /obj/item/reagent_containers/glass/woodmug
 	name = "wooden mug"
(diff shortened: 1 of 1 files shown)


solwashere:
Hypos main attraction isn't the vial size, it's the instant inject.

iamthesilly:
> Hypos main attraction isn't the vial size, it's the instant inject.

Then there shouldn't be an issue, considering the deluxe hypo can use bluespace vials too.

Moltijoe:
Balance council:
>there is no reason for this to happen. If the consideration is for the cmo's hypospray then youre looking at the wrong parts of the hypo

PR #19680, opened 2023-07-15
Deletes all small vials

Moltijoe:
Small vials serve no purpose as regular syringes function exactly the same while also being able to draw chemicals
Large vials serve no purpose aside from use in the cmo hypo for all of 10 minutes before bluespace vials are researched

This merges them together, giving vials a proper purpose as the "smaller and more portable than beaker" reagent object

functionally, i've just reskinned small vials into using the large vial sprite (for reagent volume clarity)
as well as increasing the reagent amount to 30u

Any large vial before is now a regular vial
Any regular vial that spawn with chemicals still only spawns with 15u, so they'll be half full

also, for those that complain about this removing one of cmo's hypospray upsides, the reason you want that hypo is for the instant injection, not the slightly larger vial prior to bluespace research 

:cl:  
rscdel: Small vials no longer exist
tweak: Large vials are now just "vials" and are weight class tiny instead of weight class small
/:cl:


adamsong (approved):
Code looks good, waiting hd merge

PR #19686, opened 2023-07-15
Fixes ambience not disabling without reconnecting

azzzertyy:
As title,
Also reduces the max time between ambience a little from 240 to 175

# Changelog
:cl:  
bugfix: Ambience settings now change in real time
tweak: Ambience is slightly more common
/:cl:


JohnFulpWillard:
i don't think ambience should be more common :|

azzzertyy:
> i don't think ambience should be more common :|

still way less common than it is on other codebases, their maxtime is like 120 irc

Moltijoe (approved):

PR #19691, opened 2023-07-16
You can now print sutures and regen mesh

ynot01:

Because scarcity is not real

I am officially bitter

https://github.com/yogstation13/Yogstation/pull/19462

I can't believe I'm agreeing with Therandomhoboo

> The problem with this is, welding fuel is infinite. Cables are too pretty much infinite. So this just basically just goes back to the same problem as what the previous PR tried to fix. However I agree it shouldn't be 6 seconds.

Material building costs are identical to cable coil, but are still 15 max stack compared to the 40 fuel and 40 cable max stack

# Changelog

:cl:  
tweak: You can now print sutures and regenerative mesh at the medical protolathe
/:cl:

### DIFF
--- a/code/modules/research/designs/medical_designs.dm
+++ b/code/modules/research/designs/medical_designs.dm
@@ -24,6 +24,28 @@
 	category = list("Control Interfaces", "Medical Designs")
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
 
+/datum/design/suture
+	name = "Suture"
+	desc = "Basic sterile sutures used to seal up cuts and lacerations and stop bleeding."
+	id = "suture"
+	build_type = PROTOLATHE
+	materials = list(/datum/material/iron = 10, /datum/material/glass = 5)
+	maxstack = 15
+	build_path = /obj/item/stack/medical/suture
+	category = list("Medical Designs")
+	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL
+
+/datum/design/regenmesh
+	name = "Regenerative Mesh"
+	desc = "A bacteriostatic mesh used to dress burns."
+	id = "regenmesh"
+	build_type = PROTOLATHE
+	materials = list(/datum/material/iron = 10, /datum/material/glass = 5)
+	maxstack = 15
+	build_path = /obj/item/stack/medical/mesh
+	category = list("Medical Designs")
+	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL
+
 /datum/design/bluespacebeaker
 	name = "Bluespace Beaker"
 	desc = "A bluespace beaker, powered by experimental bluespace technology and Element Cuban combined with the Compound Pete. Can hold up to 300 units."
@@ -280,7 +302,7 @@
 	materials = list(/datum/material/glass = 2000, /datum/material/diamond = 1000)
 	category = list("Tool Designs")
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL
-	
+
 /datum/design/hyposprayspeedupg
 	name = "Hypospray Speed Upgrade"
 	desc = "An upgrade for hyposprays that installs a springloaded mechanism, allowing it to inject with reduced delay."
@@ -525,7 +547,7 @@
 	build_path = /obj/item/organ/cyberimp/leg/magboot
 	category = list("Implants", "Medical Designs")
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL
-		
+
 /////////////////////////////////////////
 ////////////Regular Implants/////////////
 /////////////////////////////////////////
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -54,7 +54,7 @@
 	starting_node = TRUE
 	display_name = "Basic Tools"
 	description = "Basic mechanical, electronic, surgical and botanical tools."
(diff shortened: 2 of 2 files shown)


Mqiib:
Honestly I did find it kind of weird how difficult it was to get more sutures and mesh if you lost yours and someone had looted medbay. For how simple of an item they are and how abundant they are up until you actually need them, it kinda makes sense.

Therandomhoboo:
I may be a raving lunatic but I do have a method to my madness.

LassiH:
how about making them chem recipes like medicated sutures

wonderinghost:
maybe make emergency sutures and meshes hand craftable

wonderinghost:
not mass printable in one lathe

Therandomhoboo:
> ###
> **[LassiH](/LassiH) ** commented [Jul 16, 2023](#issuecomment-1636934980)

I won't lie, I'm getting bored of the comment of "Just make it chem made" or "make chemicals do xyz" I'd rather not have drake abuse an exploit or just force everything on chem so they have to be the mass crafters of everything.

ynot01:
> not mass printable in one lathe

The point is to make them mass printable

Reed0506:
theres already crafting recipes for sutures and regenerative mesh, why not just adjust those rather than making them be printed out of metal and glass which makes no sense at all.

ynot01:
> theres already crafting recipes for sutures and regenerative mesh, why not just adjust those rather than making them be printed out of metal and glass which makes no sense at all.

The point is parity, cable coil and welding tools can be printed just as easily

Reed0506:
i would argue that there shouldnt be parity here, healing items for robotic limbs and organic limbs shouldnt be what amounts to reskins of each other. it just doesnt make any sense to be able to print these items with metal and glass. they should be more easily obtainable through other methods, like reducing complexity of their crafting recipes and/or increasing yield. this seems like a very strange solution and could be handled a lot better without much effort at all.

RG4ORDR:
Given that the protolathes can make just about anything - I see nothing wrong in saying that these are some kinda of 3D printed medical items.

neeshacark:
pssst

use the function protolathes have that let you use chems to craft stuff

:)

azzzertyy:
yogstation try not to complain about a non issue challenge failed

Chubbygummibear:
i do kinda like the crafting idea if it was a really simple and accessible recipe like a cloth and metal/rod for sutures like it being needle and thread, and regen mesh was also maybe just cloth and ethanol. then from there you can make the improved and medicated versions with chems or whatever. printable is also a fine alternative just in the interest of making them more attainable

Notamaniac:
Okay, is this meant to be one of those ironic PR's that's meant to highlight a problem? I mean I won't pretend I've been guilty of that before, when I probably should've just made a PR nerfing the offending problem instead.

Assuming that this is an ironic PR, which is meant to imply that if organic healing shouldn't be easily mass printed, then robotic healing shouldn't be mass printable either, are you forgetting that A : The two species that rely on robotic healing don't benefit from chems healing them or chems at all B : Those species have a crippling weakness to EMP's and IPC's in particular are more fragile to all sources of damage.

If this is not a ironic PR - If you see no problem with Joe Mschoe having 14 stacks of sutures and regen mesh each, equalling a total of 15x14x10 2100 points of brute and burn heal to tap into, then I think any real further arguing on my part is useless

wonderinghost:
> Okay, is this meant to be one of those ironic PR's that's meant to highlight a problem? I mean I won't pretend I've been guilty of that before, when I probably should've just made a PR nerfing the offending problem instead.
> 
> Assuming that this is an ironic PR, which is meant to imply that if organic healing shouldn't be easily mass printed, then robotic healing shouldn't be mass printable either, are you forgetting that A : The two species that rely on robotic healing don't benefit from chems healing them or chems at all B : Those species have a crippling weakness to EMP's and IPC's in particular are more fragile to all sources of damage.
> 
> If this is not a ironic PR - If you see no problem with Joe Mschoe having 14 stacks of sutures and regen mesh each, equalling a total of 15x14x10 2100 points of brute and burn heal to tap into, then I think any real further arguing on my part is useless

There is also the problem in that inorganics only got welders and coils outside of magic revival and gutted nanites. In order to make it not as viable you got to change the entirety of tools and production. Like changing values on welders needing to be refueled after being made or have a very specific wire for healing. Or reworking most welders

wonderinghost:
There just isn't a bed we are attempting to make but place a mattress over the problem

In that inorganics only have these welders and coils for any good healing

ynot01:
> Okay, is this meant to be one of those ironic PR's that's meant to highlight a problem? I mean I won't pretend I've been guilty of that before, when I probably should've just made a PR nerfing the offending problem instead.

This is a PR that, if merged, I will see no further problem in robotic limb healing as flesh healing will become nearly as accessible.

> A : The two species that rely on robotic healing don't benefit from chems healing them or chems at all

This line of logic has never appealed to me because I simply do not abuse healing mixes.

> B : Those species have a crippling weakness to EMP's and IPC's in particular are more fragile to all sources of damage.

I will say first and foremost that this was never directed at IPCs in particular. This has been a problem for many years with robotic limbs, whether on preternis back then or IPC now. It is less a problem on races that mix both flesh and robotic limbs because having to heal both types repeatedly becomes an inconvenience and I believe getting EMPed with prosthetic limbs hardstuns you, which gives you the 'massive downside' of being an IPC but without any of the other benefits.

To clear up some misconceptions, preternis are not vulnerable to EMPs, in fact they are immune to EMPs. IPCs are vulnerable to EMPs but it is gutted from how crippling it once was.

IPCs are not "more fragile to all sources of damage", but have no brute modifier, a 1.25x burn modifier, and 0.8x stamina modifier.

> If this is not a ironic PR - If you see no problem with Joe Mschoe having 14 stacks of sutures and regen mesh each, equalling a total of 15x14x10 2100 points of brute and burn heal to tap into, then I think any real further arguing on my part is useless

I see no problem with this the same way that many people see no problem with robotic personnel doing this every round for several years

Moltijoe (changes requested):
please add the design id to the base techweb node or otherwise fitting techweb node
as i don't think it'll show up otherwise

Moltijoe (approved):

PR #19711, opened 2023-07-17
HELP! AN IPC HAS STOLEN MY PRETERNIS EYES TO USE FOR THEMSELVES

Moltijoe:
IPCs run on electricity, might as well let them run Preternis eyes off that

:cl:  
tweak: IPC can use Preternis eyes without them running out of power
/:cl:


iamthesilly:
Recommend using (owner.get_process_flags() & SYNTHETIC) instead of the species checks

Moltijoe:
> Recommend using (owner.get_process_flags() & SYNTHETIC) instead of the species checks

yeah, that might be better

ynot01 (approved):

PR #19712, opened 2023-07-17
Makes emergency sutures less terrible

Moltijoe:
They already heal for less than half of what regular sutures heal
They can at least be a stack of 15 instead of only a stack of 5

:cl:  
tweak: All emergency sutures now start at 15 in a stack instead of 5
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -242,8 +242,6 @@
 	desc = "A value pack of cheap sutures, not very good at repairing damage, but still decent at stopping bleeding."
 	icon_state = "suture_green"
 	heal_brute = 5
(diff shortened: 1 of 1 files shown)


SomeguyManperson:
Don't these exist in a box that's present specifically to handle wounds, and the reason they are shit at healing is because they are commonly found in oxygen lockers for the express purpose of helping people who get wounded

Moltijoe:
> Don't these exist in a box that's present specifically to handle wounds, and the reason they are shit at healing is because they are commonly found in oxygen lockers for the express purpose of helping people who get wounded

This technically didnt buff the healing, just increased the amount, making it better at handling wounds

ynot01 (approved):

PR #19720, opened 2023-07-18
Lets people stand up while moving

Moltijoe:
No reason it needs to prevent movement, should make combat slightly more active, with falling over no longer being a death sentence
If someone wants to keep you down, there are other more reasonable ways than just dragging you around

:cl:  
tweak: Can stand up while moving
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -461,7 +461,7 @@
 	if(!resting)
 		set_resting(TRUE, FALSE)
 	else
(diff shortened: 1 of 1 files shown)


neeshacark:
Makes sense logically and gives you more options in combat, should be fun

Ethan4303:
helps me not be stun locked for 3 seconds while moving through drapes, Big W for atmos

ToasterBiome (approved):
I don't like this but we can see how it changes things

PR #19737, opened 2023-07-19
Revert "You can now print sutures and regen mesh"

solwashere:
Reverts yogstation13/Yogstation#19691
This was a bad idea that harms medical and devalues the use of mechanical limbs. Self-healing methods are meant to be limited in one way or another, with mechanical self-healing being anywhere as a benefit to counter the fact that mechanical limbs suck. Especially with the recent buff to self-healing via mesh and suture where you no longer need to switch body parts to heal separate body parts, it just does it itself. The only people that this really benefits are tiders who break into medical to steal supplies for themselves as there is now a potentially infinite amount of suture and mesh that can be printed, as I rarely see medical players use sutures and mesh for anything outside of certain wounds.

# Changelog

🆑 

tweak: sutures and mesh are no longer printable at medical prolathes

/🆑

### DIFF
--- a/code/modules/research/designs/medical_designs.dm
+++ b/code/modules/research/designs/medical_designs.dm
@@ -24,28 +24,6 @@
 	category = list("Control Interfaces", "Medical Designs")
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
 
-/datum/design/suture
-	name = "Suture"
-	desc = "Basic sterile sutures used to seal up cuts and lacerations and stop bleeding."
-	id = "suture"
-	build_type = PROTOLATHE
-	materials = list(/datum/material/iron = 10, /datum/material/glass = 5)
-	maxstack = 15
-	build_path = /obj/item/stack/medical/suture
-	category = list("Medical Designs")
-	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL
-
-/datum/design/regenmesh
-	name = "Regenerative Mesh"
-	desc = "A bacteriostatic mesh used to dress burns."
-	id = "regenmesh"
-	build_type = PROTOLATHE
-	materials = list(/datum/material/iron = 10, /datum/material/glass = 5)
-	maxstack = 15
-	build_path = /obj/item/stack/medical/mesh
-	category = list("Medical Designs")
-	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL
-
 /datum/design/bluespacebeaker
 	name = "Bluespace Beaker"
 	desc = "A bluespace beaker, powered by experimental bluespace technology and Element Cuban combined with the Compound Pete. Can hold up to 300 units."
@@ -302,7 +280,7 @@
 	materials = list(/datum/material/glass = 2000, /datum/material/diamond = 1000)
 	category = list("Tool Designs")
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL
-
+	
 /datum/design/hyposprayspeedupg
 	name = "Hypospray Speed Upgrade"
 	desc = "An upgrade for hyposprays that installs a springloaded mechanism, allowing it to inject with reduced delay."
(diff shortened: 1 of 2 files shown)


iamthesilly:
Add a changelog please

Therandomhoboo:
Mechanical limbs are still not useless or devalued due to the fact that you can still infinitely repair them and they are everywhere. (cables & fueling tanks) This is locked to medbay lathes only so its not like everyone is getting them, also if "greytiders" break in, then that means admins can swiftly get rid of them therefore making it a buff to get rid of shitlers too

iamthesilly:
The council will decide your fate

ynot01:
Changelogs must be prefixed with :cl: and suffixed with /:cl:

SomeguyManperson:
>  also if "greytiders" break in, then that means admins can swiftly get rid of them therefore making it a buff to get rid of shitlers too

this has never worked and will never work, since it implies we can actually trust most of the playerbase (we don't)

athena413:
> Mechanical limbs are still not useless or devalued due to the fact that you can still infinitely repair them and they are everywhere. (cables & fueling tanks) This is locked to medbay lathes only so its not like everyone is getting them, also if "greytiders" break in, then that means admins can swiftly get rid of them therefore making it a buff to get rid of shitlers too

do serial greytiders ever actually get banned though

it takes comically high levels of tide & powergaming in order for people to actually get banned for mass greytide

also relying on solely admins to fix an issue that causes people to commit minor rulebreaks doesnt work because admins wont ban someone over a simple b&e

Therandomhoboo:
> > Mechanical limbs are still not useless or devalued due to the fact that you can still infinitely repair them and they are everywhere. (cables & fueling tanks) This is locked to medbay lathes only so its not like everyone is getting them, also if "greytiders" break in, then that means admins can swiftly get rid of them therefore making it a buff to get rid of shitlers too
> 
> do serial greytiders ever actually get banned though
> 
> it takes comically high levels of tide & powergaming in order for people to actually get banned for mass greytide
> 
> also relying on solely admins to fix an issue that causes people to commit minor rulebreaks doesnt work because admins wont ban someone over a simple b&e

Greytiders/shitlers etc will break into anywhere regardless. So all in all saying this encourages them to do so is a moot point because they will break into everywhere... So I don't see how "reverting" a pr solves the greytiding issue. Cuz they still do  it

solwashere:
Greytiders will break in no matter what, but I don't want to reward them more for doing so as it genuinely takes no effort to print these at a prolathe

ynot01:
If you don't want people breaking in for sutures/mesh you could try putting some out on the front desk

SomeguyManperson:
The only real thing that stops people from just using the generic stack healing items everywhere is their scarcity, considering every other form of healing in medical is either free (sleepers, surgery) or effectively free (cryo) but relies on stationary installations.
The only options that don't follow this trend are chemistry (relies on a chemist doing chemistry) and the stack healing items, which are supplied at roundstart, generally able to compete or out-heal chemicals but being in a semi-limited quantity that would either require cargo (lmao) or a chemist to produce more, keeping them important for triage and quickly managing damage but less effective than dumping someone in a sleeper or stasis bed simply because you can't run out of either of those

Unfortunately the actual "reason" for the change has absolutely nothing to do with a good medical dynamic and is instead caused by a fit over mechanical healing being simplified over non-mechanical healing (no wounds, healing with easier to find and replace items), which used to be more of a consolation or reward for getting augments or playing a mechanical species (they tended to be weak) than a generic upside for playing with like 2 slight damage vulnerabilities (and potentially resistances).

Basically we're torching a perfectly functional medical system for the express reason of "machine race good" which ?????????????????????

Moltijoe:
Balance council:
>While a consideration for making them cost plastic was made
Them being easily printable does just open the door for powergamers to have all the healing they need without the downside robotic limbs can pose

As a side note, with makeshift sutures and ointments being easier to craft, it'll mitigate this revert, as SOME accessible healing items will still be around

solwashere:
> Balance council:
> >While a consideration for making them cost plastic was made
> Them being easily printable does just open the door for powergamers to have all the healing they need without the downside robotic limbs can pose
> 
> As a side note, with makeshift sutures and ointments being easier to craft, it'll mitigate this revert, as SOME accessible healing items will still be around

Balanced around the fact that only cook players and miners know about the crafting menu

Moltijoe:
> Balanced around the fact that only cook players and miners know about the crafting menu

Some roboticists do too, as thats the only way to make drones
And its also faster to make the small bots with the menu

ToasterBiome (approved):

PR #19741, opened 2023-07-19
Lowers buster arm base price from 15 TC to 13 TC

iloveloopers:

This item is not worth 15 TC when I can get CQC, sleeping carp or holoparasite for cheaper and be able to buy other things + this is the only martial art item that can be taken away from you.
I think 13 TC makes it more inline with the other martial art prices
This does mean however that you can potentially get it for 11 TC if you get lucky enough to get cybersun as your sponsor

# Wiki Documentation

change buster arm base price to 13 TC

# Changelog

:cl:  
tweak: Buster arm base price is now 13 TC down from 15 TC
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -332,7 +332,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	warned that the arm renders them unable to wear gloves and sticks out of most outerwear."
 	item = /obj/item/storage/box/syndie_kit/buster
 	player_minimum = 25
-	cost = 15
+	cost = 12
 	manufacturer = /datum/corporation/traitor/cybersun
 	surplus = 0
 
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -332,7 +332,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	warned that the arm renders them unable to wear gloves and sticks out of most outerwear."
 	item = /obj/item/storage/box/syndie_kit/buster
 	player_minimum = 25
-	cost = 12
+	cost = 13
 	manufacturer = /datum/corporation/traitor/cybersun
 	surplus = 0
 


iloveloopers:
stop disliking without commenting you bozos. tell me why this is a bad change

thegoldencat413:
item was designed to be a sort of expensive, all-inclusive purchase that would help you access areas (mega buster), gap close (mop the floor), mitigate damage, and attack people, so reducing the price so you can use something like a gun or esword (for an extra 50% block chance) wouldn't make any sense, the 5tc you get left over are for utility items like agent card, emp flashlight, or whatever else.

if the issue is "it's not strong enough" to you, then improving it would be better than hurting its identity by just reducing the price, i think

thegoldencat413:
basically think of mega buster as a sort of built-in emag

also, uhh, the fact that the martial art can be taken away from you isn't *that* significant, as you've already lost if you're in a situation where you're getting it removed]

pacification exists as well for everything other than cqc

iloveloopers:
> item was designed to be a sort of expensive, all-inclusive purchase that would help you access areas (mega buster), gap close (mop the floor), mitigate damage, and attack people, so reducing the price so you can use something like a gun or esword (for an extra 50% block chance) wouldn't make any sense, the 5tc you get left over are for utility items like agent card, emp flashlight, or whatever else.

ok but you can already get it for 12 TC by simply getting lucky and having cybersun as your sponsor

iloveloopers:
would it maybe be better to just make it so its always 12 TC, no sponsors?

thegoldencat413:
well, the discount isnt guaranteed, discounts are allowed to break the rules but that probably shouldn't happen all the time or i think the intended design would be compromised

iloveloopers:
theres 4 different sponsors, so you get a guaranteed 20% discount 25% of the time. But whatever, may the almighty balance council tell us what they think about this?

iloveloopers:
price raised to 13 TC

LazennG:
just gonna ask on the pr for posterity: why should the buster arm's price be lowered so drastically when it can
- break multiple walls at once
- break multiple windows at once
- break almost any item/object
- throw mechs 
- combat borgs
- fight against people at close and far range

iloveloopers:
To give traitors who buy the buster arm more options for utilities. Fighting against sec officers who are even slightly competent is really difficult. You're not immune to stuns or stamina damage, officers can just keep chasing you at range while shooting their disablers.

iloveloopers:
Most of the time traitors only buy this when its 12 TC (sponsor) cause it lets you buy anti stuns and other utilities

Moltijoe:
Balance Council:
>This isn't something that's currently needed

Waiting on 24hours to close

UltraBigRat:
> To give traitors who buy the buster arm more options for utilities. Fighting against sec officers who are even slightly competent is really difficult. You're not immune to stuns or stamina damage, officers can just keep chasing you at range while shooting their disablers.

Have you heard of our lord and savior uno reverse card, and using guns yourself? You literally can just megabust into caps office and robust with the antique laser.

PR #19746, opened 2023-07-20
Makes worldbreaker hijack only

Moltijoe:
Simply put, i'm not satisfied with the martial art, but i can think of no way to touch it without it being miserable for everyone facing it
so i've locked it behind hijack only so it's more rare

i decided against it having glorious death since the martial art itself actively fights against dying

:cl:  
tweak: Worldbreaker is restricted to hijack traitors
tweak: Worldbreaker can't get discounts
/:cl:


athena413:
i dont care too much about worldbreaker but afaik suicide by cop isnt intended to be a method of die gloriously

PR #19750, opened 2023-07-20
Makes self surgery slightly less painfully slow

Moltijoe:
One of the few upsides of robotic limbs, instead became far too tedious to be "fun"
Some steps take FAR too long, especially if they can mess up

:cl:  
tweak: Self surgery is only 1.5x regular length instead of 3x
/:cl:

### DIFF
--- a/code/modules/surgery/surgery_step.dm
+++ b/code/modules/surgery/surgery_step.dm
@@ -95,7 +95,7 @@
 
 	var/speed_mod = 1
 	if(user == target)
-		speed_mod *= 3 // harder to do on yourself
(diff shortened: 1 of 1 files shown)


azzzertyy:
change to 2x instead, ipcs should still rely on robotics for proper surgery and changing it to 1.5x seems like an overbuff

Moltijoe:
> change to 2x instead, ipcs should still rely on robotics for proper surgery and changing it to 1.5x seems like an overbuff

several steps take a lot of time by default, i'd only consider making it 2x if their base time was reduced

ToasterBiome (approved):

Issue #19702, opened 2023-07-16
 Some sprites dont have icons anymore for whatever reason


Yogbot-13:
## Test Merges: 
Not Supplied / None
## Reproduction:
Description: As title, sprites like cigarette boxes and rolling paper just dont work no more

 Submitted by: azzzertyy (196392067281321985)

JohnFulpWillard:
Issue related to https://github.com/yogstation13/Yogstation/pull/19643 you should've reported it there
Also it's already been fixed.

PR #19767, opened 2023-07-21
Fixes the CE's holosign creator

JohnFulpWillard:

This has been broken since it was added and no one seemed to have noticed, well **I** did when I was looking into a bug report on update_appearance.
Also adds autodoc and repaths the item so vscode doesn't treat it like a define.

before
image

after
image

# Changelog

:cl:  
bugfix: The CE's holosign creator now doesn't give you the raw path of the sign it is creating.
/:cl:

### DIFF
--- a/code/game/objects/items/holosign_creator.dm
+++ b/code/game/objects/items/holosign_creator.dm
@@ -12,10 +12,12 @@
 	throw_speed = 3
 	throw_range = 7
 	item_flags = NOBLUDGEON
-	var/list/signs = list()
+	///List of all holosigns currently made.
+	var/list/obj/item/holosign_creator/signs = list()
+	///The type of holosign to make from this projector.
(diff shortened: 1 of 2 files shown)


ynot01:
can you rename it to advanced instead of CE because it gets used in places that arent just the CE locker (out of scope 🥺)

Maia-J:
> well I did when I was looking into a bug report on update_appearance.

Can you link this bug report?

JohnFulpWillard:
https://github.com/yogstation13/Yogstation/issues/19702

JohnFulpWillard:
> can you rename it to advanced instead of CE because it gets used in places that arent just the CE locker (out of scope 🥺)

done and i also made them call parent so they respect signal and nointeract trait

ToasterBiome (approved):

PR #19782, opened 2023-07-22
Replaces EMP protection component with EMPPROOF traits

Moltijoe:
The component is fucky and doesn't really allow messing with the flags, while also needing to be destroyed to be properly removed
It also doesn't play nicely with multiple different sources of emp protection

:cl:  
bugfix: Fixes some weird interactions around multiple sources of EMP immunity
/:cl:


yogstation/code/modules/mob/living/carbon/human/species_types/preternis/preternis.dm
```diff
@@ -83,8 +83,8 @@ adjust_charge - take a positive or negative value to adjust the charge level
 		maglock = new
 		maglock.Grant(C)
 		lockdown = FALSE
-		
-	C.AddComponent(/datum/component/empprotection, EMP_PROTECT_SELF)
+	
+	ADD_TRAIT(C, TRAIT_EMPPROOF_SELF, "preternis")
```
iamthesilly: Merge conflicts, also should remove this part now that preternis are no longer fully EMP-immune.

SapphicOverload (approved):
Yeah looks good

PR #19784, opened 2023-07-22
Removes training wheels from Ultra Violence EMP immunity

Moltijoe:
I was initially worried that emps would just delete IPCs with the martial art so i added the adaptive bit
but they're proving to be too strong with one of the main counters being effectively removed so i'm removing the automatic part

They can still counter emps with throw mode
They just no longer get 5 second duration immunity for either countering or getting hit by an emp

also, as a side note, EMPs have been nerfed quite a bit since the martial art was made

:cl:  
rscdel: Ultra Violence no longer has adaptive EMP immunity
/:cl:


LassiH:
5 second immunity is kinda useless anyways since they shoot you once get shocked and swap to lasers

Moltijoe:
> 5 second immunity is kinda useless anyways since they shoot you once get shocked and swap to lasers

Like i said, training wheels, its for people that couldnt pull off the counter reliably

SapphicOverload (approved):
Works on my machine

PR #19787, opened 2023-07-22
Disables echolocation

JohnFulpWillard:

Echolocation currently blocks you from being able to interact with anything.
Your client is filled with a ton of null images while echolocation is active (which is fine when the echolocation images are gone), which causes problems.
Until this is fixed, this reverts blindness to simply make you blind, without echolocation.

I also had made a ton of changes to echolocation to fix some other issues, but the null in images is what is causing the issue. It is hard to go in-depth more on this without being able to use breakpoints to see it step-by-step.

### DIFF
--- a/code/_onclick/hud/fullscreen.dm
+++ b/code/_onclick/hud/fullscreen.dm
@@ -226,8 +226,3 @@
 	icon_state = "blue_eye"
 	plane = FULLSCREEN_PLANE
 	layer = CURSE_LAYER
-
-/atom/movable/screen/fullscreen/echolocation
-	icon_state = "echo"
-	layer = CAMERA_STATIC_LAYER
-	plane = CAMERA_STATIC_PLANE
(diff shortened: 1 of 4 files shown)


neeshacark:
tragic but needed

azzzertyy:
danger_turfs = typecacheof(list(/turf/open/chasm, /turf/open/lava, /turf/open/floor/fakepit, /turf/open/space, /turf/open/openspace, /turf/open/floor/fakespace))
no, we dont want to display space at all, i specifically removed it for this reason

azzzertyy:
also why are you moving the call to initalize, this makes it harder to be able to change any of the variables when adding the component, we want to be able to call the component with different variables

JohnFulpWillard:
> no, we dont want to display space at all, i specifically removed it for this reason

I thought it would be nice to have it, I'll remove it.

> also why are you moving the call to initalize, this makes it harder to be able to change any of the variables when adding the component, we want to be able to call the component with different variables

This doesn't change that, it just makes it more readable because you know exactly what the defaults will be in initialize. adding a component w/ args will overwrite them, those are just used if there's nothing overwritten.

azzzertyy:
are you sure its null images being generated?

ynot01 (approved):

PR #19796, opened 2023-07-22
Hallucinating Checks for the Status Effect Instead of Proc

Runian:
`proc/hallucinating()` is no more (as it did nothing) and all checks that used it now uses `has_status_effect(/datum/status_effect/hallucination)`.

# Changelog
:cl:  
bugfix: Checking if the user is hallucinating now works instead of always failing.
/:cl:


code/modules/mob/mob_helpers.dm
```diff
@@ -317,7 +317,11 @@
 /mob/proc/hallucinating()
 	return FALSE
 
-
+/mob/living/carbon/hallucinating()
+	if(has_status_effect(/datum/status_effect/hallucination))
+		return TRUE
+	return FALSE
```
tattax: just remove the proc above and change all checks to has status effects, i didn't do it originally cause the pr was already too big

Runian: ok

JohnFulpWillard:
You'll have to fix conflicts from https://github.com/yogstation13/Yogstation/pull/19798 (once that is merged, but hopefully very soon), so I'll mark this as draft for now.

tattax (approved):

JohnFulpWillard (approved):

PR #18568, opened 2023-04-15
The Synthetic Revolution || AI Rework 4.0

Bibby0110:
# Document the changes in your pull request
All AI machines now need to wired up using ethernet cables into a network.
A data core will require a locally connected server with at least 1THz and 1TB to be functional.
A network consists of all the locally connected machines.
Two networks can be connected using a networking machine. This allows the AI and Netmin to use the resources and cores in both networks.

## Changes to the AI data core
The AI Core now looks like this with the weird new sprites being networking machines. They point towards the one they're connected to if you want to snoop out where other networks might be
image

## Changes to the Netmin
The Network Admin now lives in Science in the new Computer Science room.
image
Their tool to interact with a network is now either the laptop they spawn with, or a modular computer connected with an ethernet cable.

Accessing the network using such a console allows for (at present) 3 actions. These all work based on the CPU assigned to the local network.
image

In addition to this it allows you to remotely control networking machines in the local network. 
It allows allows the distribution of networking resources to either the connected AIs, or the network itself.

This is also the way you'll be uploading and downloading AIs now.
If you're uploading you need to upload to a network that either has a data core locally, or one connected through a networking machine. Likewise for downloading.

One other minor feature added is a leaderboard program you can download. If connected to a network this will show the persistent five highest CPU and RAM records.

### Networking Machines
If the AI has an unconnected network machine you can force access to it using your own machine. To do this you rotate it using a wrench within 20 degrees of the other machine. Then you use a multitool and select the machine you want to connect to. After this is all done you use a screwdriver to finalize the connection. You can disconnect manually by using a wirecutter, or the UI.

### The Master Subcontroller
A new machine has been added to the RnD server room.
image
If an AI is connected to this machine via a networking machine they will be able to instantly open doors, access APCs and access firealarms. Without it there's a 1 second delay for doing these actions.
If you multitool the master controller you can disable individual departments from working, even if connected to the AI.
At roundstart the security department is DISABLED, and the AI will have a mildly harder time doing stuff in security.

## Synthetics
image
These creatures are more hardy than humans, have inherent pacifism and can't use any kind of stunweapons. They "bleed" white fluid instead of blood, and they're very white. These are meant for small non-combat related jobs that the station may need.

Currently there's 1 job spot available at roundstart. They start in maint near the Head meeting room.
They can switch to other synth units be entering a pod or clicking on an adjacant vacant unit.

It's possible to create empty synths by the same process you make a cyborg, but instead of inserting an MMI or posibrain you insert an AI CPU.

[...]

PR #17260, opened 2023-01-01
Modernizes components

ThatLing:
# Document the changes in your pull request

Ports various improvements to components, as well as splitting signal defines into seperate files.

A (non-complete) list of PRs, credits to the original authors:
- https://github.com/tgstation/tgstation/pull/56372
- https://github.com/tgstation/tgstation/pull/59154
- https://github.com/tgstation/tgstation/pull/71056

# Changelog

No player-facing changes

PR #19801, opened 2023-07-23
Science-Related Jobs gets Mood Bonus for Petting Cyborgs

Runian:
Port of tgstation/tgstation/pull/46296.
The science department (Scientist, Roboticist, and Research Director) and any adjacent jobs (Geneticist) can pet cyborgs to get an increase to their mood.

Mood events can be restricted to certain jobs and will only allow these certain jobs to get the mood event.
Removes `COMSIG_ADD_MOOD_EVENT_RND` since it is unused.

# Changelog
:cl:  
rscadd: Mood events can be restricted to jobs.
rscadd: Petting a Cyborg is now a positive mood event for all science members (and the geneticist too).
/:cl:


Therandomhoboo:
WE should give a mood buff to everyone else that kills the cyborg that ISN'T in the sci department.

code/datums/components/mood.dm
```diff
@@ -28,6 +28,11 @@
 		modify_hud()
 		var/datum/hud/hud = owner.hud_used
 		hud.show_hud(hud.hud_version)
+	if(!ishuman(owner))
+		return
+	var/mob/living/carbon/human/H = owner
+	if(H.mind?.assigned_role in list("Research Director", "Scientist", "Roboticist"))
```
JohnFulpWillard: Gene should be included since they are part of science somewhat

Runian: done

Moltijoe (commented):
Are there going to be enough RND mood events that they need their own signal?

Runian:
> Are there going to be enough RND mood events that they need their own signal?

This is the only mood event I plan to add. Unless someone plans to port over tgstation/tgstation/pull/68592 or has a better way that isn't checking if the mob uses mood in `silicon_defense`, I think that having its own signal is the best way of doing all of this even if it might be only the mood event that uses the signal.

Runian:
> Are there going to be enough RND mood events that they need their own signal?

`COMSIG_ADD_MOOD_EVENT_RND` was added by #17260 and was entirely unused. Anyways, changed how all of this is done with a list in the mood event and removed `COMSIG_ADD_MOOD_EVENT_RND` too.

JohnFulpWillard (approved):
we need to port tg mood system already

GDRdev:
tcoms should be able to pet the AI core and get a mood bonus from that as well

Runian:
> tcoms should be able to pet the AI core and get a mood bonus from that as well

If they become a more science dedicated role, sure (#18568).

PR #19811, opened 2023-07-23
Reworks preternis for the 28483575th time

SapphicOverload:

Preternis being fully immune to the biggest downside of robotic limbs was probably not the best idea for balance, so instead they have a trait that reduces the strength of incoming EMPs by one. This means heavy EMPs act as light EMPs, and light EMPs do nothing.

Because
- Water damage reduced by half, water slowdown reduced to 1 instead of 4
- Removed 0.2 second attack slowdown
- All limbs have 35 health instead of 40 for arms and 30 for legs

# Wiki Documentation

On the wiki page for preternis in the benefits/weakness section:
- Remove "takes 10% more brute damage" (was never updated after the last PR)
- Replace the EMP immunity with "they are unaffected by weak EMPs, and heavy EMPs cannot damage their organs"

# Changelog

:cl:  
tweak: preternis reduce incoming EMP strength by one level instead of being fully immune
tweak: reduced preternis water damage by half and slowdown from 4 to 1
tweak: all preternis limbs have 35 health instead of 40 or 30
tweak: removed preternis attack slowdown
/:cl:


Therandomhoboo:
This is a massive buff to them not a rework just because their emp immunity has been tweaked shouldn't mean they lose all their weaknesses. Especially since this is an inherent trait too, so they start with it if i'm right. If thats the case they should have even more weaknesses.

As well, they are getting buffs from eating certain ores now. So honestly, this is actually just making this race the new meta race to go to. Infinite healing, emp immunity to light emp and heavy emp  does light emping instead, no more reduced attack speed, reduced water, brute, stamina & stun mods. 

What is genuinely the drawbacks of playing them now?

Maia-J:
Me modifying my limbs to lose the last disadvantage that this species will have now
(Augments rock again)

iamthesilly:
> This is a massive buff to them not a rework just because their emp immunity has been tweaked shouldn't mean they lose all their weaknesses. Especially since this is an inherent trait too, so they start with it if i'm right. If thats the case they should have even more weaknesses.
> 
> As well, they are getting buffs from eating certain ores now. So honestly, this is actually just making this race the new meta race to go to. Infinite healing, emp immunity to light emp and heavy emp does light emping instead, no more reduced attack speed, reduced water, brute, stamina & stun mods.
> 
> What is genuinely the drawbacks of playing them now?

They're not immune to EMPs anymore, just less affected. This means anyone with access to EMPs can hurt you, but less than it would hurt an IPC. For example stechkin EMP rounds will kill you in 3 shots instead of the 2 shots it would take without any EMP protection.

Moltijoe (approved):

PR #19840, opened 2023-07-24
Fixes the canister UI

JohnFulpWillard:

New UI

image
image

There has always had a port button on canister UIs but it was always hidden because of the pressure/Regulator/valve button, and you'd have to resize the window horizontally to find it.
This makes it actually visible to players in-game.

# Changelog

:cl:  
bugfix: Canister UIs no longer have their port hidden.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/Canister.js
+++ b/tgui/packages/tgui/interfaces/Canister.js
@@ -20,7 +20,7 @@ export const Canister = (props, context) => {
     restricted,
   } = data;
   return (
-    <Window width={300} height={232} resizable>
+    <Window width={300} height={254} resizable>
       <Window.Content>
         <Section
           title="Canister"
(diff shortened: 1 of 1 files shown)


ToasterBiome (changes requested):
looks a bit like shit, can you at least make the 4 even?

alternate solution is making the original one wider instead

JohnFulpWillard:
I made them all centered i thought that was good, can you give more explanation what you think is wrong w/ it?

ToasterBiome:
> I made them all centered i thought that was good, can you give more explanation what you think is wrong w/ it?

if the ui was a 2x2 grid each of those elements should be centered in their individual grids, looks weird rn

MajManatee:
I think it looks fine.

ToasterBiome (approved):

PR #19842, opened 2023-07-24
Holodeck/Thunderdome now load from templates

JohnFulpWillard:

Holodeck & Admin Thunderdome now load from templates, rather than being stored at Centcom and copied constantly.

This means that adding new holodeck maps are easier to do, and they do not need to be loaded on roundstart, eating away at init times.
It also means that it is now more on-par with TG, making it easier to port things in the future, and may help later down the line when I make a new attempt at lazyloading.
Also removes our reliance on object duplication, as it is very unreliable and easy to break.

### Extra notes

I've modularized the yog-unique holodeck stuff, and commented out the TG holodecks that yog does not have.
I also removed the medical one from requiring emag, to keep consistent with current balance.
Burn test was removed because it was causing some problems. On TG it just spawns a ton of plasma, while on Yog it's just a pit of lava. May be worth just porting TG's version of it instead of trying to fix the current, but I don't think it's really worth at the moment.
If there's any game-changing behavior, please let me know.

# Changelog

:cl:  
experimental: Holodeck and the Admin Thunderdome now load from map templates.
/:cl:


JohnFulpWillard:
sorry @ToasterBiome I use strongdmm 2.0 ☹️

ToasterBiome:
Please reopen when work commences

N3D6:
Based cause i have literally zero clue how to map holodecks rn

neeshacark (approved):
templates are just cut and pasted from before, maps are fine - i did a quick looksie and spotted nothing out of the ordinary

neeshacark:
except for the conflicts that is :)

Moltijoe (approved):

PR #17707, opened 2023-01-30
Shimmyjimmies ion bolts to be less "why are they like that" and more "hey that kinda makes sense"

Mqiib:
OLD ION PROJECTILES:
hos gun/ion shotgun shells: 50/50 of heavy or light on target
literally everything else: heavy EMP on target, 50/50 heavy/light on surrounding tiles

ION PROJECTILES NOW:
Standard bulky ion gun: fires 0.5/1 bolts (heavy EMP in the epicenter, light surrounding)
ion carbines/hos gun: fires 0/1 bolts (light in a 1 tile radius, no heavy)
ion shotgun shell: no longer 50% chance of heavy EMP on hit
mech ion cannon: 2/2 EMP bolt (1 radius heavy, 50/50 of heavy or light in 2 tile radius)
stechkin EMP rounds: now guarantee a heavy EMP on the target

EMP blasts will now only apply a heavy EMP on the epicenter if the heavy radius is greater than 0. This won't affect large EMP effects like EMP grenades, mostly only makes an impact with small EMPs.

# Wiki Documentation

weak bolts: 0/0, used by ion shotgun
light bolts: 0/1, used by ion carbines
standard bolts: 0.5/1, standard EMP bolt used by hos gun and ion rifle
heavy bolts: 2/2, used by heavy ion cannon on mechs

# Changelog

:cl:  
rscadd: Added heavy and light EMP energy projectiles
tweak: Hos gun now shoots light bolts but can fire 5 of them instead of 3
tweak: Mech ion cannon shoots big boy ion bolts
tweak: Ion shotgun shells can no longer apply heavy EMP effects
tweak: Ion carbines fire weaker ion bolts that can't apply heavy EMP effects
tweak: Stechkin EMP rounds guarantee a heavy EMP on the target
/:cl:

PR #19845, opened 2023-07-24
Makes ion carbines actually shoot the weaker ion projectiles like was stated in a PR 6 months ago

Moltijoe:
#17707

Pretty sure other changes from that PR weren't actually made, but this is the only one that actually seems to matter right now

:cl:  
bugfix: ion carbines now ACTUALLY shoot weaker bullets like was intended in a PR 6 months ago
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -22,6 +22,7 @@
 	w_class = WEIGHT_CLASS_NORMAL
 	slot_flags = ITEM_SLOT_BELT
 	pin = null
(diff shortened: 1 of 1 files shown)


Runian:
Shouldn't this be a bugfix?

Moltijoe:
> Shouldn't this be a bugfix?

i considered it maybe, but decided since it was one that was 6 months old i labeled it as tweaks

RG4ORDR:
What's the ammo count on the carbine?

JohnFulpWillard (approved):

PR #19852, opened 2023-07-25
Worldbreaker balance PR (again)

Moltijoe:
Sure this increases max amount of armour to 100, however with the reduction to the maximum amount of plates it will be
138 total damage reduction before
120 total damage reduction after

The main thing is it'll reduce the downtime of waiting for plates to stack up from 0 without actually making plates grow faster
as well as removing chip damage while heavy

Plates not getting affected at all by other damage types made stamina weapons completely useless and forced the armoury to be opened immediately
Now other types of weapons weapons DO work, but less effectively at 1/3 their original damage
Technically a baton will be better at breaking plates than a vibro blade, but meleeing someone with a martial art is a dumb idea in the first place

being heavy no longer increases damage as durability, speed, and knockback scaling with heavy was enough of a distinction
as well, damage being inconsistent between weights felt worse than it needed to
The damage was averaged, so being heavy will do less damage and being light will do more damage than before
grab never scaled damage with heavy so it's untouched

Removes AP because the martial art already does enough, it doesn't need to almost entirely ignore armour too
Still does 50/50 melee/bomb though, so it'll 'effectively' have some AP as not many melee armours also have bomb armour

:cl:  
tweak: Worldbreaker tweaks
tweak: plates give 20 armour instead of 10
tweak: max plates that can give armour is 5 down from 8
tweak: max plates is 10 down from 15
tweak: worldstomp base radius increased to 8
tweak: all AP removed
tweak: all damage also does 2x that as stamina damage
tweak: damage types other than brute and burn are now able to break plates, but at 1/3 effect
tweak: plates only knock back and stagger if thrown by the worldbreaker
tweak: being heavy no longer increases damage
tweak: leap does 20 instead of 15/25
tweak: worldstomp does 15 base instead of 20/10
tweak: pummel does 15 instead of 18/12
tweak: pummel targets a tile instead of mob
tweak: pummel and leap can damage structures
tweak: pummel, leap, and worldstomp can now damage machinery, but significantly less damage than structures
bugfix: grab has a cooldown
/:cl:

### DIFF
--- a/code/datums/martial/worldbreaker.dm
+++ b/code/datums/martial/worldbreaker.dm
@@ -56,6 +56,9 @@
 	if(!can_use(H) || (modifiers["shift"] || modifiers["alt"] || modifiers["ctrl"]))
 		return
 
+	if(isobj(target) && target.loc == H) //no using abilities on your own items
+		return
+
 	if(H.a_intent == INTENT_DISARM)
 		leap(H, target)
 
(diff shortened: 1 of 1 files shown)


JohnFulpWillard:
this dude

ToasterBiome (approved):

PR #19853, opened 2023-07-25
Open containers can now interact with syringes again

Moltijoe:
You're telling me that a vial or beaker which literally has an open top, can't interact with a syringe in any way?

:cl:  
bugfix: Syringes can draw from and inject into vials and beakers again (among other things)
/:cl:


Moltijoe:
hold up what, now it's working on the server? i have no clue what's happening

Moltijoe:
seems to be fixed so i guess i'm just tripping

PR #19855, opened 2023-07-25
Makes the all terrain vehicle actually all terrain

Moltijoe:
Why give mining medics a clunky ATV when it can't even drive over the only terrain that matters, lava
Also, it's a vehicle, why is it slower than just walking?

:cl:  
tweak: ATVs can drive over lava
tweak: ATV is now faster than walking (on lavaland)
/:cl:

### DIFF
--- a/code/modules/vehicles/atv.dm
+++ b/code/modules/vehicles/atv.dm
@@ -5,6 +5,7 @@
 	icon_state = "atv"
 	key_type = /obj/item/key
 	var/static/mutable_appearance/atvcover
+	resistance_flags = LAVA_PROOF
 
 /obj/vehicle/ridden/atv/Initialize(mapload)
 	. = ..()
--- a/code/modules/vehicles/atv.dm
+++ b/code/modules/vehicles/atv.dm
@@ -10,7 +10,7 @@
 /obj/vehicle/ridden/atv/Initialize(mapload)
 	. = ..()
 	var/datum/component/riding/D = LoadComponent(/datum/component/riding)
-	D.vehicle_move_delay = 1
+	D.vehicle_move_delay = 0
 	D.set_riding_offsets(RIDING_OFFSET_ALL, list(TEXT_NORTH = list(0, 4), TEXT_SOUTH = list(0, 4), TEXT_EAST = list(0, 4), TEXT_WEST = list( 0, 4)))
 	D.set_vehicle_dir_layer(SOUTH, ABOVE_MOB_LAYER)
 	D.set_vehicle_dir_layer(NORTH, OBJ_LAYER)


azzzertyy:
> Why give mining medics a clunky ATV when it can't even drive over the only terrain that matters, lava Also, it's a vehicle, why is it slower than just walking?
> 
> cl tweak: ATVs can drive over lava tweak: ATV is now faster than walking (on lavaland) /cl

this makes raiding syndicate base trivial, this makes breaking someone out of perma trivial.
atvs can go faster sure but not on lava

Moltijoe:
> this makes raiding syndicate base trivial, this makes breaking someone out of perma trivial. atvs can go faster sure but not on lava

raiding syndicate base isn't trivial because unless you're an antag you're not supposed to do it and you'll get bwoinked (or you're supposed to, since it counts as fucking with ghost roles). Also, if someone actually wakes up at syndicate base they can just turn the turrets on and it becomes impossible again.

breaking someone out of perma is fair, but it's already possible to do that using lavaland stuff, dragon boat, goliath boat, flight potion.

the ATV is a mining medic specific tool, if it gives them an easy way to break a fellow antag out of prison, then they likely would have been capable regardless.

also, you'd need at least a bluespace bodybag to get them out safely as only one person can ride the atv at a time
which means it certainly wouldn't be round start

by in large, this will be a pleasant mining medic buff and not much else

azzzertyy:
> > this makes raiding syndicate base trivial, this makes breaking someone out of perma trivial. atvs can go faster sure but not on lava
> 
> raiding syndicate base isn't trivial because unless you're an antag you're not supposed to do it and you'll get bwoinked (or you're supposed to, since it counts as fucking with ghost roles). Also, if someone actually wakes up at syndicate base they can just turn the turrets on and it becomes impossible again.
> 
> breaking someone out of perma is fair, but it's already possible to do that using lavaland stuff, dragon boat, goliath boat, flight potion.
> 
> the ATV is a mining medic specific tool, if it gives them an easy way to break a fellow antag out of prison, then they likely would have been capable regardless.
> 
> also, you'd need at least a bluespace bodybag to get them out safely as only one person can ride the atv at a time which means it certainly wouldn't be round start
> 
> by in large, this will be a pleasant mining medic buff and not much else

isn't there a bounty for raiding the syndicate documents from lavaland base? isnt it intended to be raided

Moltijoe:
> isn't there a bounty for raiding the syndicate documents from lavaland base? isnt it intended to be raided

If there is, thats news to me
Still, that doesnt change that theres plenty of other ways to cross lava safely, just all ones that would require the mining medic to not do their job and instead play as miner

azzzertyy:
> > isn't there a bounty for raiding the syndicate documents from lavaland base? isnt it intended to be raided
> 
> If there is, thats news to me Still, that doesnt change that theres plenty of other ways to cross lava safely, just all ones that would require the mining medic to not do their job and instead play as miner

yeah but what this does is garauntee that every round where miners discover a syndicate base, theyll immediately ask mining medic to borrow their ATV to raid the documents, it garauntees that every round theyll get raided, rather than only if the miners decide to, since they need to get an RCD

ToasterBiome:
Other methods of getting over lava require getting non-roundstart options (except for unmapped lava, which you can RCD over) so I believe having a roundstart method that's basically perfect will lessen the value of all the other ways to get across.

PR #19849, opened 2023-07-25
[MDB Ignore] Asteroid trims overhaul without random stupid conflicts

N3D6:
# Document the changes in your pull request

title
read the old pr for everything else can't be assed
# Spriting

# Wiki Documentation

# Changelog

:cl:  
mapping: asteroid trims overhaul
mapping: additional medbay fixes by @maxion12345 
mapping: additional hallway fixes by @ToasterBiome 
/:cl:

PR #19858, opened 2023-07-25
Moves the holosigns out of the window on Asteroid to hell because they're deleted now

maxion12345:
Before:
Y097QlX

After:
vGLJbAT

I just think blue signs inside of a blue window is a bad choice. can't really read it.

# Spriting
n/a

# Wiki Documentation
n/a
# Changelog
:cl:  
tweak: Moved surgery signed out of the window on Asteroid
/:cl:


Aquizit (changes requested):
What I would recommend is actually remove them altogether. We have holosigns correctly upon the doors into surgery, which is all we need to have. The only circumstance where that is different is the CMO's office on a couple stations where they have a surgery table, and that isn't done like these are in a horizontal window.

So yeah just get rid of them completely and we'll call it good.

Aquizit:
Rolled into #19849

PR #19874, opened 2023-07-26
Adds some new cyber implants (ideas welcome)

Moltijoe:
I want more implants for different slots so people actually need to decide how they want to chrome up rather than always picking the same things
Currently there's just two in this PR, but i'd appreciate people leaving ideas in the comments

Arm actuator implants - arms
slightly increases do_after speed

Coagulant implant - heart aid
injects sanguirite if you start bleeding with a 1 minute cooldown

:cl:  
rscadd: coagulant implant
rscadd: arm actuator implant
wip: This is work in progress  
/:cl:

### DIFF
--- a/code/modules/surgery/organs/augments_arms.dm
+++ b/code/modules/surgery/organs/augments_arms.dm
@@ -441,3 +441,23 @@
 	desc = "A modified version of the Stechkin pistol placed inside of the forearm to allow for easy concealment."
 	contents = newlist(/obj/item/gun/ballistic/automatic/pistol/implant)
 	syndicate_implant = TRUE
+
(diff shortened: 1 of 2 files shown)


RG4ORDR:
Subdermal Armor - for easy of access it's a "chest" implant that'll apply a layer of damage specific dermal armor.
Bullet,burn,brute,energy - Flat value of 5, can't stack.
 Syndies get a special implant that gives 5 across the baord.
 
 All variants will give like 5-10 wound armor regardless of what variant is installed.

Moltijoe:
I think i have too much on my plate right now, i'll reopen this when i have more time

PR #19875, opened 2023-07-26
Hailer changes

maxion12345:

Changes the sechailer code for the voice lines in order to achieve the following goals:

- [x] Only trigger if the word or phrase is at the front, and not in the middle of another sentence
- [x] Dynamicly adjust findtext search for the length of each phrase, so all phrases can be said
- [ ] Make this work regardless of accents (lizards say ssshit instead of shit, so that's 6 chars instead of 4)

Current code includes some helpers that print out the length of the message said, alongside what message is passed to the code. Current issues include that it is passing what your character says, as opposed to what you type. For instance, I type "shit" in the say box, but I get two different strings because one of those characters is a lizard.

The image below triggers the voicelines. The image above does not, as the character count is different from what the trigger phrase is. I'd like to avoid having lizard-specific trigger phrases, so finding a way to pass the "raw" text would be ideal.

# Spriting

# Wiki Documentation

# Changelog

:cl:  
tweak: Security gasmask voiceline tweaks
/:cl:


JohnFulpWillard:
Alternative to https://github.com/yogstation13/Yogstation/pull/19873

This prevents anyone with an accent from using the lines, which I do not like.

maxion12345:
>

maxion12345:
my dumbass hit the wrong button instead of posting my comment


> Alternative to #19873
> 
> This prevents anyone with an accent from using the lines, which I do not like.

This is a draft because it currently stops accents from working. Once that's fixed, this won't be an issue. Primarily an alternative because I found it annoying for voice lines to trigger without the specific phrase. This returns it to it's original behavior, where it only checked for the word in the front of your text and disregarded any text outside the length of the specific phrase to avoid accidental triggering (for instance, if you're calling someone hot shit or calling their outfit shit or asking for copy paper or whatever)

maxion12345:
alright what the fuck its working just fine now
????????
hgPJACq
really dont understand this, especially since I got a photo of it capturing the entire 6 letter lizard version up there. but if it works, it works

maxion12345:
Re-started the server to test again and it gave back two different strings based on accent. Might be a bug with handle_speech?

maxion12345:
i just dont understand. one photo, it'll return what you type. next test, it'll return what your character says. i just dont know anymore
brYf3zG

JohnFulpWillard:
Handle speech should be taking the full message sent without accents being added

maxion12345:
Looking further into it, the message passed by handle_speech gets modified along the way. Most notably, the tongue calls handle_speech to modify for species accents, and the mind does the same for accents. 

Hopek has suggested a workaround by storing a "raw" version of the message passed along with modification in the mind of a mob, that can be retrieved by the hailer for use.

Optimally, I would just see about modifying handle_speech to duplicate the speech_message arg with a different arg that wouldn't be modified at any point, whilst retaining the functionality of speech_message. 

Speaking of that, anyone know where handle_speech's code is? Maybe I'm having a smoothbrain moment (I'm always having one) but I can't seem to find it.

maxion12345:
Closing for now. Haven't had any motivation to work on this, nor my portaturret ports. Code is in semi-working condition if someone wants to take a stab at it.

ShoziOffical:
guh

PR #19879, opened 2023-07-26
Makes cogscarab slightly less tedious to play

Moltijoe:
They're expected to run all over reebe upgrading and building things, but they're slower than players
Wasn't wanting to risk giving them -1 speed like regular drones, as that makes them notably faster than players and they don't have pacifism like regular drones
But at a speed of 0, it should at least be less painful to play as a glorified drone

:cl:  
tweak: Cogscarabs are slightly faster but have less health
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/drone/extra_drone_types.dm
+++ b/code/modules/mob/living/simple_animal/friendly/drone/extra_drone_types.dm
@@ -112,11 +112,9 @@
 	icon_dead = "drone_clock_dead"
 	picked = TRUE
 	pass_flags = PASSTABLE
-	health = 50
-	maxHealth = 50
 	harm_intent_damage = 5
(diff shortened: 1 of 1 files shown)


Reed0506:
maybe change it to 0.33 so they are the same speed as pods, or somewhere between that and 0. i think cogscarabs should suffer at least a bit, especially in combat, where having a little bit of speedmod can be the difference between life and death.

Moltijoe:
> maybe change it to 0.33 so they are the same speed as pods, or somewhere between that and 0. i think cogscarabs should suffer at least a bit, especially in combat, where having a little bit of speedmod can be the difference between life and death.

These are simplemobs, they move on a different system
A speed of 0 means they'll still be slower than humans

ToasterBiome (approved):

PR #19902, opened 2023-07-28
Makes illegal tech less illegal

ynot01:

Under current rules security technically 🤓 isnt supposed to be using debatably illegal technology even though most of it is useful to them and made for their use specifically

This renames it to "Foreign Technology" but if anyone thinks of any cooler names like espionage or something I'm cool with that

# Changelog

:cl:  
tweak: Illegal tech is no longer illegal
/:cl:


Zer0Tactics:
Ah yes, the military nanites that explode people aren't illegal and are useful

Bad PR trying to solve an in game meta rule with a name change

Moltijoe (approved):

PR #19904, opened 2023-07-28
Doubles Round Start SMES Power

Aquizit:

With engineering now usually building more robust engines at round start, they take longer to set up. Cargo and Brig regularly start running low on power when this happens. Additionally, with new things added or expanded, initial power draw at round start is just higher than it used to be. This allows for extra time to get shit set up the way engineers want to do it, but doesn't make it trivial.

# Changelog

:cl:  
tweak: Changed round start engine SMES from 50000000 to 100000000 (5e7 -> 1e8)
mapping: Made Asteroid station engine SMES match the other stations
/:cl:

### DIFF
--- a/_maps/map_files/AsteroidStation/AsteroidStation.dmm
+++ b/_maps/map_files/AsteroidStation/AsteroidStation.dmm
@@ -18931,15 +18931,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/security/checkpoint/auxiliary)
-"fgM" = (
-/obj/structure/cable{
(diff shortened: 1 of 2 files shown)


iamthesilly:
This doesn't actually solve that. The real reason departments lose power early is because the SMES units are by default set to an output level that's less than enough to supply the entire station. All engineers need to do to fix so they can take their time is to just turn up the SMES outputs, or set up solars/turbine.

Basically, skill issue.

Aquizit:
I suppose that's a fair point.

PR #9037, opened 2020-06-26
[READY] Adds the Supermatter Delaminator

ghost:
Adds the supermatter delaminator, a 10TC hijack/glorious death only item that will guarantee a delamination of your local supermatter crystal. Comes in a syndie box kit, complete with instructions and extraneous equipment.

### Intent of your Pull Request

Supermatter can now be seeded with an antinoblium shard and a corruptor, corrupting the support infrastructure that regulates it. The effects of this are not immediately felt, but when they do the entire station will feel it. The supermatter's supporting platform, as well as the radio on it and the power shield that regulate it are corrupted and slowly degrade. This support integrity starts at 100 and decreases by 1 every time the supermatter goes BRRRRRRR. As the integrity goes lower, the SM becomes more deadly, releasing more gas (scales with support integrity), increasing its power all by itself (scales), releasing shock arcs (at stability <70) to complicate repair and stabilization efforts and even radballs when stability dips below 40, complete with starting its own fires. When stability hits 10, it starts emitting massive EMP waves to ensure that any active cooling that has managed to keep it stable up until this point stops functioning, whereupon it will go on its merry way BRRRRRRRing until full delamination.

This whole process is balanced by the fact that the degradation of the support infrastructure takes about 5~8 minutes before it hits 0 and starts delaminating, granting ample time for engineers to prepare emergency cooling and for the shuttle to be called. The size of the SM explosion is also increased, though this can be tweaked if desired. 

Oh, and don't try to space the SM to cheese the energy values and gasmix. It's a stupid easy fix to buy time, I know, but it will just make it explode harder.
#### Changelog

:cl:  
rscadd: SM now goes into a death spiral when emagged. Better call the shuttle before it goes BRRRRRRRRR
/:cl:

PR #19920, opened 2023-07-29
Resonance Cascade

SapphicOverload:

This PR introduces a new SM delamination type, caused by the traitor delaminator kit or large amounts of antinoblium gas. If antinoblium gas is the cause, it can be reversed by adding hypernoblium gas, but only if the traitor kit wasn't used and it hasn't been too long since the process began.

Before the explosion, it has the same effects as the delaminator already did (see #9037 for details), with the addition of a gravitational pull and the explosion power lowered a bit to compensate for what's about to happen. The alarm sounds will be audible across the station once it reaches 25% or lower integrity.

Immediately after the explosion, every APC on the station will be EMP'd causing a station-wide blackout, and some will cause tesla arcs or small explosions. Most of the lights will also break. After this, portals will appear and dangerous lavaland fauna will invade the station. The shuttle will be automatically called if it hasn't been already.

# Wiki Documentation

On the guide to the supermatter, add antinoblium to the list of gas effects with this description:
"Adding this gas to the supermatter will cause a resonance cascade that will destroy a lot of the station, unless you add lots of hypernoblium quickly enough to reverse the process. Do not use this gas unless you want to kill people."

# Changelog

:cl:  
rscadd: adds resonance cascade, a new delamination type
bugfix: fixed portal storms spawning visual effects in the wrong place
/:cl:

### DIFF
--- a/code/modules/admin/secrets.dm
+++ b/code/modules/admin/secrets.dm
@@ -554,6 +554,15 @@
 					id.critter_money = FALSE
 			message_admins("[key_name_admin(mob_user)] has deactivated critter money (pets generated on money withdrawl)!")
 			log_admin("[key_name(mob_user)] has deactivated critter money.")
+		if("halflife")
+			if(!check_rights_for(rights, R_FUN))
+				return
+			for(var/obj/machinery/power/supermatter_crystal/S in GLOB.machines)
+				if(!isturf(S.loc) || !is_station_level(S.z))
+					continue
+				S.antinoblium_attached = TRUE
+			message_admins("[key_name_admin(mob_user)] has started a resonance cascade!")
+			log_admin("[key_name(mob_user)] has started a resonance cascade.")
 
 	if(E)
 		E.processing = FALSE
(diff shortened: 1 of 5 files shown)


iloveloopers:
so there needs to be atleast 100 mols of antinoblium for this to happen?

iamthesilly:
> so there needs to be atleast 100 mols of antinoblium for this to happen?

Yes, and antinoblium has to be over 50% of the total gas.

alexkar598:
" "Antinoblium has been known to [REDACTED] when in contact with supermatter crystals in high enough amounts, causing catastrophic damage. Keep it away from the engine at all times unless you want to kill everyone." "

This isn't the SCP wiki, things dont [REDACTED]

iamthesilly:
> " "Antinoblium has been known to [REDACTED] when in contact with supermatter crystals in high enough amounts, causing catastrophic damage. Keep it away from the engine at all times unless you want to kill everyone." "
> 
> This isn't the SCP wiki, things dont [REDACTED]

Alright, changed it

LoliconSlayer:
We don't add lore style descriptions to guides.

iamthesilly:
Is this better?

LoliconSlayer:
Yes :thumbsup:

ToasterBiome (approved):

GDRdev:
I need to learn anti nob now!

PR #19923, opened 2023-07-29
Stops heirophant club from attacking if you click an item in your inventory

Moltijoe:
no real reason it needs to attack if you're trying to shift things around in your inventory and you misclick

:cl:  
bugfix: Stops heirophant club from attacking if you click an item in your inventory
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -1473,7 +1473,7 @@ GLOBAL_LIST_EMPTY(aide_list)
 		return
 	if(isitem(target))//don't attack if we're clicking on our inventory
 		var/obj/item/thing = target
-		if(target.item_flags & IN_INVENTORY)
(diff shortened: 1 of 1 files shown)


code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -1471,6 +1471,8 @@ GLOBAL_LIST_EMPTY(aide_list)
 	var/turf/T = get_turf(target)
 	if(!T || timer > world.time)
 		return
+	if(isobj(target) && target.loc == user)//don't attack if we're clicking on our inventory
```
JohnFulpWillard: You can check item_flags for ``IN_INVENTORY`` instead

Moltijoe: Oh, thats way nicer

JohnFulpWillard (approved):

PR #19924, opened 2023-07-29
Makes martial arts faster to learn after a round ends

Moltijoe:
With the donator perk of getting one free uplink item for the end of the round, it's pretty useless to select a martial art as they take so long to read that you'll be dead before you get time to finish reading it

This cuts out all except for 1 instance of reading

:cl:  
tweak: Makes martial arts faster to learn after a round ends
/:cl:

### DIFF
--- a/code/game/objects/items/granters/martial_arts/_martial_arts.dm
+++ b/code/game/objects/items/granters/martial_arts/_martial_arts.dm
@@ -16,6 +16,8 @@
 
 /obj/item/book/granter/martial/on_reading_start(mob/user)
 	to_chat(user, span_notice("You start reading about [martial_name]..."))
+	if(SSticker.current_state == GAME_STATE_FINISHED)//faster after the round ends
+		pages_to_mastery = 0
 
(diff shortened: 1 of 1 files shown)


code/game/objects/items/granters/martial_arts/_martial_arts.dm
```diff
@@ -16,9 +16,11 @@
 
 /obj/item/book/granter/martial/on_reading_start(mob/user)
 	to_chat(user, span_notice("You start reading about [martial_name]..."))
+	if(SSticker.current_state == GAME_STATE_FINISHED)//faster after the round ends
+		pages_to_mastery = 0
```
JohnFulpWillard: You can probably just call ``on_reading_finished`` and early return, then in subtypes check for that early return.

Moltijoe: i just did that in a different proc so i wouldn't need to modify other subtypes

JohnFulpWillard (approved):

PR #19929, opened 2023-07-30
Removes all implants from medical printing except for revelant ones

Moltijoe:
Roboticists should be the ones in charge of implanting, they have a medical setup for a reason, and it's not just for borging
if you want specifically medical to implant it still, go to robotics, get the implants, then go back to medical

The only implants they can still print are all the organs, basic luminescent eyes, medhud, surgical toolset, and breathing tube

curbs the powergaming of medical people just asking eachother for full implants

:cl:  
tweak: medbay can only print relevant implants
/:cl:


RG4ORDR:
I agree.

wonderinghost:
what about the posibrain they can print

Yarinoi:
tbh fair

Longestarmlonglaw:
why cant medical and robotics both share the same job of implanting people? one is meant for treating the station's carbon crew and the other is responsible for maintain the silicones

wonderinghost:
> why cant medical and robotics both share the same job of implanting people? one is meant for treating the station's carbon crew and the other is responsible for maintain the silicones

Because robotics doesn't have a lot to do other than print 50 posibrains.

Also there is usually a tendency to just fix and revive all inorganic crew as by the time you get a preternis over to robotics and repaired, all their organs had gone bad

So the first sentence I had

ToasterBiome (approved):

PR #19930, opened 2023-07-30
Makes C4 overlay appear overtop of mobs rather than under

Moltijoe:
I just ran around an entire round with a party plastic explosive on myself and no one reacted because no one saw it

:cl:  
tweak: Makes C4 overlay appear overtop of mobs rather than under
/:cl:


tattax:
tested?

Moltijoe:
> tested?

image

ToasterBiome (approved):

PR #19931, opened 2023-07-30
Flashers now use flashes instead of snowflake stun

ynot01:

Flashers now inherit the behavior of flash assemblies by directly calling their `flash_carbon` and `flash_borg` (new) procs. This has no negative effect on the flash's lifespan.

Mounted flashers will still apply twice the "strength" of a normal flash and thus apply twice as many confusion stacks. (hellish)

# Changelog

:cl:  
tweak: Mounted and portable flashers no longer have special stun behavior from flash assemblies
/:cl:


code/modules/assembly/flash.dm
```diff
@@ -171,32 +171,36 @@
 		if(M.flash_act())
 			M.set_confusion_if_lower(power * CONFUSION_STACK_MAX_MULTIPLIER SECONDS)
 
+/obj/item/assembly/flash/proc/flash_borg(mob/living/silicon/robot/R, mob/user)
+	log_combat(user, R, "flashed", src)
+	if(!R.sensor_protection)
+		update_icon(ALL, flash = TRUE)
+		R.Paralyze(rand(80,120))
```
tattax: time defines

code/modules/assembly/flash.dm
```diff
@@ -171,32 +171,36 @@
 		if(M.flash_act())
 			M.set_confusion_if_lower(power * CONFUSION_STACK_MAX_MULTIPLIER SECONDS)
 
+/obj/item/assembly/flash/proc/flash_borg(mob/living/silicon/robot/R, mob/user)
```
tattax: var names

code/game/machinery/flasher.dm
```diff
@@ -124,7 +124,10 @@
 			continue
 
 		if(L.flash_act(affect_silicon = 1))
```
JohnFulpWillard, line 126: what does this do that we need a second flash_borg too?

ynot01: flash_act only interacts with the eyes- it blurs/blinds/applies the flash overlay and also checks for flash protection, returning whether it was successful or not

Moltijoe (approved):

PR #19943, opened 2023-07-31
You can no longer set off a maxcap in your hand and survive

Mqiib:
Instead of the STRONGEST EXPLOSION POSSIBLE being reduced to a mere 99 damage if you're wearing a bomb suit, now they will deal a total of 300 (200 brute 100 burn) even with 100 bomb armor. No hugging maxcaps, you die. No amount of protection that you can reasonably wear will save you from that kind of blast. Also this means you can't face tank the BSA.

# Wiki Documentation

Bomb armor is 83% effective, but only 50% for especially strong blasts.

# Changelog

:cl:  

tweak: Devastation-tier explosions will now game end you more effectively

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -445,7 +445,7 @@
 				gib()
 				return
 			else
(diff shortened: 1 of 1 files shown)


iloveloopers:
this will affect the trout population, aswell as cargo syndi bomb events

Moltijoe (approved):

PR #19945, opened 2023-07-31
Removes the ability for players to start a map vote

adamsong:

Disables the ability for players to start a map vote. This is good for the game because map votes force box 90% of the time, and map variety is objectively good from a game design perspective.

# Changelog

:cl:  
rscdel: Removed player map votes
/:cl:

### DIFF
--- a/config/config.txt
+++ b/config/config.txt
@@ -153,7 +153,7 @@ ID_CONSOLE_JOBSLOT_DELAY 30
 #ALLOW_VOTE_MODE
 
 ## allow players to initiate a map-change vote
-ALLOW_VOTE_MAP
+#ALLOW_VOTE_MAP
 
 ## min delay (deciseconds) between voting sessions (default 10 minutes)
 VOTE_DELAY 6000


Reed0506:
based

N3D6:
If this passes I will cashapp you 10 million zillion dollars adamsogm

iloveloopers:
box is litteraly the best map in the universe you can't do this to me this will KILL yogstation FOREVER and we will LOSE all of the POP!!!11111

neeshacark:
inb4 all the admins force map votes when its not box

TaelynTheLost:
Removing choice is not good design practice.  Easy proof of that, is if 80% of the station wants Box, then you upset 80% of your players.  If you want people to play on different maps, make those maps something more people actually want to play on.

thegoldencat413:
i dont like this with ice meta because it has a lot of issues and is very anemic with content

N3D6:
> i dont like this with ice meta because it has a lot of issues and is very anemic with content

this has nothing to do with icemeta its literally not even merged or testmerged for these reasons alone

adamsong:
> then you upset 80% of your players.

The number of players who hate box is dwarfed by the number who just prefer it."I would prefer box" and "I enjoy playing other maps" are not mutually exclusive.

Additionally, the number of players who dislike non box maps will likely decrease as they are forced to play non box maps, and mappers have a reason to work on not box maps.

N3D6:
> Removing choice is not good design practice. Easy proof of that, is if 80% of the station wants Box, then you upset 80% of your players. If you want people to play on different maps, make those maps something more people actually want to play on.

80% is a ridiculously bloated number, every map vote I have seen in the past few months asteroid or icemeta one has been very close to box and only lost by a couple votes

also the maps are great, we can’t change people refusing to adapt to change. our maps are not the problem in the slightest. the best way to grow to appreciate a map? actually give it a chance instead of whining the entire time

Moltijoe:
only worry i have is that if we're relying entirely on rotation, that means it's possible to be forced into shitty box while at like ~8 people lowpop when gax would be far superior in that situation

N3D6:
> only worry i have is that if we're relying entirely on rotation, that means it's possible to be forced into shitty box while at like ~8 people lowpop when gax would be far superior in that situation

add a min pop to box

LoliconSlayer:
> Removing choice is not good design practice. Easy proof of that, is if 80% of the station wants Box, then you upset 80% of your players. If you want people to play on different maps, make those maps something more people actually want to play on.

Damn so you mean we'll be rid of Boxtards and instead retain players who are willing to play on different maps but got tired of 24/7 box?

https://github.com/yogstation13/Yogstation/assets/54045237/6dce390c-7e28-40e2-a10c-653e51b9ae9b

You don't need to gild the lily this PR is already based enough.

Therandomhoboo:
Only problem I have with this is when it rolls gax or meta ( the smaller one Iirc) which is hellish on highpop cause the map isn't meant for 50+ players.

neeshacark:
map min and max pops will need tweaks but this is easily preventable

> Only problem I have with this is when it rolls gax or meta ( the smaller one Iirc) which is hellish on highpop cause the map isn't meant for 50+ players.

N3D6:
> Only problem I have with this is when it rolls gax or meta ( the smaller one Iirc) which is hellish on highpop cause the map isn't meant for 50+ players.

meta is our med-highpop 
gax caps at like 30 i think so unless the round starts lowpop and gradually gets higher this will not happen
edit: gax caps at 40 which is still fine

MajManatee:
> Damn so you mean we'll be rid of Boxtards and instead retain players who are willing to play on different maps but got tired of 24/7 box?

SO TRUE!! Fuck em. If they don't like the other maps they can make one themselves.

Moltijoe:
reopen when this might actually get merged

PR #19951, opened 2023-07-31
Map rotations and player votes no longer offer maps that were played >=70% of the time in the last 10 rounds

ToasterBiome:

So a problem we have with retaining impressive mappers is that they feel that working on anything other than BoxStation is a waste of time because we nearly always play BoxStation. Many solutions were presented like "turning off map votes", "turning off map vote after map rotations" and "killing all people who like box" However I do not believe the community to be very approving of this decision. I have created an infinitely worse solution that still gives BoxStation lovers the upper hand while FORCING players to experience these new maps, eventually with the opportunity to lower this even more. Remember that BoxStation is only 25% of our map pool.

Simply put, if we played BoxStation 7 times in a row, it won't even be an option. This still gives you an impressive SEVENTY PERCENT box chance, but that forces it to an upper range where no-lifers still have to experience new and different maps (which will in turn get issues fixed on them). If you play 5 rounds of Box, a Gax, and then 2 rounds of box, looks like we're going to Meta/Gax/Asteroid. PLAY THE OTHER MAPS!!!!!!!!!!!!!!!

Of course, this is generic. If we somehow played 7 rounds on the NVS Gax, it also wouldn't be offered.

Not a config option, not parameterized, not anything. This is going up for a draft and making sure I actually am calling the database correctly (I have no clue how it works)

# Changelog

:cl:  
rscadd: Map rotations and player votes no longer offer maps if they were played the last 70% of rounds
/:cl:

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -644,6 +644,6 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 			var/next_map = query_previous_maps.NextRow()
 			if(!next_map)
 				continue
-			previous_maps[next_map.item[1]] += 1 / (11 - i) //this lessens the influence of rounds that were longer ago
+			previous_maps[next_map[1]] += 1 / (11 - i) //this lessens the influence of rounds that were longer ago
 		qdel(query_previous_maps)
 	return previous_maps
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -375,17 +375,7 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 	else
 		for(var/M in global.config.maplist)
 			mapvotes[M] = 1
-			
-	var/list/previous_maps = list()
-	if(SSdbcore.Connect())
-		var/datum/DBQuery/query_previous_maps = SSdbcore.NewQuery({"
-			SELECT map_name FROM [format_table_name("round")] WHERE id BETWEEN lower = :lower AND upper = :upper
-		"}, list("lower" = GLOB.round_id - 9, "upper" = GLOB.round_id))
-		if(!query_previous_maps.Execute())
-			qdel(query_previous_maps)
-		while(query_previous_maps.NextRow())
-			previous_maps[query_previous_maps.item[1]] += 1
-		qdel(query_previous_maps)
+	var/previous_maps = get_map_weights()
 
 	//filter votes
 	for (var/map in mapvotes)
@@ -407,7 +397,7 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 		if (VM.config_max_users > 0 && players > VM.config_max_users)
 			mapvotes.Remove(map)
 			continue
-		if(previous_maps && previous_maps[VM.map_name] > 7)
+		if(previous_maps[VM.map_name] > 7)
 			mapvotes.Remove(map)
 			continue
 
@@ -639,3 +629,21 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 	for(var/turf/to_contain as anything in Z_TURFS(z_level))
 		var/area/our_area = to_contain.loc
 		our_area.contained_turfs += to_contain
+
+
+
+/datum/controller/subsystem/mapping/proc/get_map_weights()
+	var/list/previous_maps = list()
+	if(SSdbcore.Connect())
+		var/datum/DBQuery/query_previous_maps = SSdbcore.NewQuery({"
+			SELECT map_name FROM [format_table_name("round")] WHERE id BETWEEN lower = :lower AND upper = :upper
+		"}, list("lower" = GLOB.round_id - 9, "upper" = GLOB.round_id))
+		if(!query_previous_maps.Execute())
+			qdel(query_previous_maps)
+		for(var/i = 1 to 10)
+			var/next_map = query_previous_maps.NextRow()
+			if(!next_map)
+				continue
+			previous_maps[next_map.item[1]] += 1 / (11 - i) //this lessens the influence of rounds that were longer ago
+		qdel(query_previous_maps)
+	return previous_maps
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -375,6 +375,17 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 	else
 		for(var/M in global.config.maplist)
 			mapvotes[M] = 1
+			
+	var/list/previous_maps = list()
+	if(SSdbcore.Connect())
+		var/datum/DBQuery/query_previous_maps = SSdbcore.NewQuery({"
+			SELECT map_name FROM [format_table_name("round")] WHERE id BETWEEN lower = :lower AND upper = :upper
+		"}, list("lower" = GLOB.round_id - 9, "upper" = GLOB.round_id))
+		if(!query_previous_maps.Execute())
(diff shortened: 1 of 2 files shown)


code/controllers/subsystem/mapping.dm
```diff
@@ -375,6 +375,17 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 	else
 		for(var/M in global.config.maplist)
 			mapvotes[M] = 1
+			
+	var/list/previous_maps = list()
+	if(SSdbcore.Connect())
+		var/datum/DBQuery/query_previous_maps = SSdbcore.NewQuery({"
+			SELECT map_name FROM [format_table_name("round")] WHERE id BETWEEN lower = :lower AND upper = :upper
+		"}, list("lower" = GLOB.round_id - 9, "upper" = GLOB.round_id))
```
alexkar598: Config the lower bound

code/controllers/subsystem/mapping.dm
```diff
@@ -375,6 +375,17 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 	else
 		for(var/M in global.config.maplist)
 			mapvotes[M] = 1
+			
+	var/list/previous_maps = list()
+	if(SSdbcore.Connect())
```
alexkar598: Did you copy that from somewhere, calling connect seems sus but im not sure

code/controllers/subsystem/mapping.dm
```diff
@@ -375,6 +375,17 @@ GLOBAL_LIST_EMPTY(the_station_areas)
...
+		var/datum/DBQuery/query_previous_maps = SSdbcore.NewQuery({"
+			SELECT map_name FROM [format_table_name("round")] WHERE id BETWEEN lower = :lower AND upper = :upper
+		"}, list("lower" = GLOB.round_id - 9, "upper" = GLOB.round_id))
+		if(!query_previous_maps.Execute())
+			qdel(query_previous_maps)
+		while(query_previous_maps.NextRow())
```
alexkar598: qdel but then you loop through anyways?

code/controllers/subsystem/mapping.dm
```diff
@@ -375,6 +375,17 @@ GLOBAL_LIST_EMPTY(the_station_areas)
...
+			previous_maps[query_previous_maps.item[1]] += 1
+		qdel(query_previous_maps)
```
alexkar598: Double qdel if it failed

code/controllers/subsystem/mapping.dm
```diff
@@ -375,6 +375,17 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 	else
 		for(var/M in global.config.maplist)
 			mapvotes[M] = 1
+			
+	var/list/previous_maps = list()
+	if(SSdbcore.Connect())
+		var/datum/DBQuery/query_previous_maps = SSdbcore.NewQuery({"
+			SELECT map_name FROM [format_table_name("round")] WHERE id BETWEEN lower = :lower AND upper = :upper
```
alexkar598: In theory, you should check for same port and server IP to handle the possibility of having multiple live servers

code/controllers/subsystem/mapping.dm
```diff
@@ -375,6 +375,17 @@ GLOBAL_LIST_EMPTY(the_station_areas)
...
+		"}, list("lower" = GLOB.round_id - 9, "upper" = GLOB.round_id))
+		if(!query_previous_maps.Execute())
+			qdel(query_previous_maps)
+		while(query_previous_maps.NextRow())
+			previous_maps[query_previous_maps.item[1]] += 1
```
alexkar598: Use GROUP BY and COUNT(*) in SQL instead of counting in DM, it's cleaner

code/controllers/subsystem/mapping.dm
```diff
@@ -396,6 +407,9 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 		if (VM.config_max_users > 0 && players > VM.config_max_users)
 			mapvotes.Remove(map)
 			continue
+		if(previous_maps && previous_maps[VM.map_name] > 7)
```
alexkar598: Another magic value that needs to be a config

alexkar598: previous maps is always defined, meant to check length()?

code/controllers/subsystem/vote.dm
```diff
@@ -190,6 +190,18 @@ SUBSYSTEM_DEF(vote)
 				if(!lower_admin && SSmapping.map_voted)
 					to_chat(usr, span_warning("The next map has already been selected."))
 					return FALSE
+				var/list/previous_maps = list()
+				if(SSdbcore.Connect())
+					var/datum/DBQuery/query_previous_maps = SSdbcore.NewQuery({"
```
alexkar598: Duplicate code

alexkar598 (dismissed):
You fool, you do not request a review from me or you get hit with an actual review

code/controllers/subsystem/mapping.dm
ToasterBiome: If Execute returns false itll kill it, that's how it was done elsewhere

code/controllers/subsystem/mapping.dm
ToasterBiome: Does qdeling an object that is already queued to be deleted have consequences?

code/controllers/subsystem/mapping.dm
ToasterBiome: How would I go about doing this?

code/controllers/subsystem/mapping.dm
ToasterBiome: Got any examples?

code/controllers/subsystem/mapping.dm
ToasterBiome: image

ToasterBiome: Does byond take a shit on looking up dictionary keys that aren't actually in the dictionary?

code/controllers/subsystem/vote.dm
ToasterBiome: Could maybe be split off into a proc on the mapping subsystem?

code/controllers/subsystem/mapping.dm
Moltijoe: Generally throws a runtime from what ive seen

N3D6:
disabling map votes better

neeshacark:
i prefer the other pr because it has more potential to let us play not box

This is too lenient

adamsong:
This feels too loose, it takes ~10.5 hours (1.5 hours per round guesstimate) to take effect. Ideally it would be a falloff, sort of like gamemodes are. The more frequently its played, the more votes and map rotation is weighted against it.

To move this pr forward I would say reduce the length of time, percentage, and ensure we have at least 2 maps for each pop level. (Maybe 3 out of the last 5 is the cutoff). A config would make this easier to tinker with

MajManatee:
7 rounds of gax... 🤤

Marmio64:
Vote weight against box should probably be removed (or reduced) with this too then since there will be a hardcap on how often boxstation can be chosen, there should be no reason it still needs to be discriminated against.

Moltijoe:
should probably be the last 5 maps, 10 is still a really long time

JamieD1:
I dont know about this chief

MajManatee:
Nah, Id merge it

JamieD1 (approved):

PR #19958, opened 2023-08-01
Condenses syndi attachments into one kit - for 1TC or 10TC for nukies

ynot01:

It's clear to me that nukies get way more out of this than tots do, so instead of making it mega expensive for both, it will just be expensive for nukies who can afford it, while making the regular kit super affordable- Having attachments is not exactly game-winning for traitors, but there is value in the lack of camera shake and the laser sight is very cool to use.

# Changelog

:cl:  
tweak: Condensed all purchasable syndicate attachments into one kit, which is more expensive for nukies than it is for traitors.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1978,23 +1978,17 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	item = /obj/item/storage/box/syndie_kit/prisonerbag
 	cost = 2
 
-/datum/uplink_item/device_tools/holo_sight
-	name = "Holographic Sight"
-	desc = "A high-tech holographic sight that improves the aim of the weapon it's attached to."
-	item = /obj/item/attachment/scope/holo
-	cost = 2
+/datum/uplink_item/device_tools/attachments
+	name = "Weapon Attachments"
+	desc = "A set of attachments that will make your gun more accurate and stable. \
+			Comes with a laser sight as well, for that extra tactical feel."
+	item = /obj/item/storage/box/syndie_kit/attachments
+	cost = 1
+	exclude_modes = list(/datum/game_mode/nuclear, /datum/game_mode/nuclear/clown_ops)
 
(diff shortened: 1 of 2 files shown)


Marmio64:
10 tc is a lot for these attachments. That is the same price of a nukie buying a SMG in the first place, or just over buying 3 SMG magazines. Given a choice between 72 extra bullets (and a savings of 1 TC) or a couple of attachments that will slightly reduce recoil, I would definitely take the bullets every single time. There is also the elite hardsuit, which is 8 TC for better protection at a cheaper price, and much more. Needs a cheaper price to be a competitive option for nukies to take.

MajManatee:
I don't see why they shouldn't be 1tc for everyone.

azzzertyy:
if its 1tc for nukies itl just be an auto-buy for everyone, theres no reason not to have it in such a case, 10tc is a fair price and makes nukies weigh the value of it more, with traitors 1tc seems a LITTLE too cheap to me, id say 2 was more fair, but its not a major deal

code/modules/uplink/uplink_items.dm
```diff
@@ -1978,23 +1978,17 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
...
+	exclude_modes = list(/datum/game_mode/nuclear, /datum/game_mode/nuclear/clown_ops)
 
-/datum/uplink_item/device_tools/laser_sight
-	name = "Laser Sight"
-	desc = "An aesthetic laser sight that improves your accuracy and shows you where you're aiming."
-	item = /obj/item/attachment/laser_sight
-	cost = 2
+/datum/uplink_item/device_tools/attachments/nukie
+	cost = 10
+	include_modes = list(/datum/game_mode/nuclear, /datum/game_mode/nuclear/clown_ops)
```
Moltijoe, line 1991: might wanna also set exclude_modes back to null here, otherwise it'll try to include and exclude the modes at the same time

Moltijoe (changes requested):

Moltijoe:
reopen when you plan to continue work

PR #19960, opened 2023-08-01
Orbital Mech Drop Fulton - Traitor Item

ynot01:

A fulton extraction pack but for mechs - Stores mechs at a nearby Syndicate facility and jettisons them to your position on command. Should open the door to a lot more playstyles for both traitors and nukies.

Costs 4TC and has unlimited uses

# Changelog

:cl:  
rscadd: Added the Orbital Mech Drop Fulton to the syndicate uplink for 4TC, brought to you by Cybersun Industries
/:cl:


GDRdev:
FINALLY THE 4TC MUALER

GDRdev:
though I dont see this helping much?

I mean for tots I guess it might help since you can hide away your combat mech away from sec till you actually need it
but for nukies mechs cost like fuckloads of tc rendering your actual person to be pretty helpless without your mech soo why would you want to be out of your main damage dealer? I mean its not like you can whip it out like other things since there is gonna be time for it to drop down and time for you get into the mech; which gives sec tons of time to blast your already weak person

MusselMain:
> though I dont see this helping much?
> 
> I mean for tots I guess it might help since you can hide away your combat mech away from sec till you actually need it but for nukies mechs cost like fuckloads of tc rendering your actual person to be pretty helpless without your mech soo why would you want to be out of your main damage dealer? I mean its not like you can whip it out like other things since there is gonna be time for it to drop down and time for you get into the mech; which gives sec tons of time to blast your already weak person

I'd say this could open up a killer stealth and ambush playstyle, where the nukie sneaks into a highly defended area of the ship, then deploys their mech and completely annihilates the crew while attacking them from an angle where the crew is at a massive disadvantage.

Marmio64:
I can definitely see new players getting slightly confused over this and thinking it will just give them a free mecha for only 4 TC, even if signs point towards it not being true. To avoid inevitable mhelps (which will definitely come), maybe just add a quick disclaimer in the purchase description that it does NOT come with a mecha, you need to supply your own.

Chubbygummibear:
can you add a demo of it in action?

githubuser4141:
> I can definitely see new players getting slightly confused over this and thinking it will just give them a free mecha for only 4 TC, even if signs point towards it not being true. To avoid inevitable mhelps (which will definitely come), maybe just add a quick disclaimer in the purchase description that it does NOT come with a mecha, you need to supply your own.

definitely, imo 'Mech Drop' shouldn't be in the title because that directly implies there's a mech to be dropped. 'Mech Extraction Kit' or something that doesn't visualize a new mecha being created would probably work better

GDRdev:
https://www.youtube.com/watch?v=KDDdLxTT8aI

drderp3635:
This should probably be role restricted to roboticist.

MajManatee:
would be dope with the audio from titanfall when called in

ynot01:
> This should probably be role restricted to roboticist.

Disagree, other people *can* use mechs and the mech for dummies book is also not role-restricted, which this sits next to on the uplink.

Reed0506:
this is an absolutely hilarious idea, if slightly extremely overpowered in niche situations such as maybe stealth warops.

probably a good idea to make sure that the fulton cant be used if someone has been scooped up into the mechs cargo bay with a clamp, that could potentially lead to unforseen consequences.

ynot01:
> probably a good idea to make sure that the fulton cant be used if someone has been scooped up into the mechs cargo bay with a clamp, that could potentially lead to unforseen consequences.

The fulton pack has a built in var called `safe_for_living_creatures` which is set to FALSE here

Moltijoe (approved):

ynot01:
Adding refundability if unused

PR #19972, opened 2023-08-02
fix invisible survival pod sleeper

ktlwjec0:
stops the sleeper being invisible

:cl:  Ktlwjec
bugfix: Survival capsule sleeper is visible again.
/:cl:

### DIFF
--- a/code/modules/mining/equipment/survival_pod.dm
+++ b/code/modules/mining/equipment/survival_pod.dm
@@ -76,7 +76,7 @@
 	name = "luxury elite bar capsule"
 	desc = "A luxury bar in a capsule. Bartender required and not included."
 	template_id = "shelter_charlie"
-	
+
 //Pod objects
 
 //Window
@@ -142,6 +142,8 @@
 	. = ..()
 	if(!state_open)
 		. += "sleeper_cover"
+	else
+		icon_state = initial(icon_state)
 
 //Lifeform Stasis Unit
 /obj/machinery/stasis/survival_pod


code/modules/mining/equipment/survival_pod.dm
```diff
@@ -142,6 +142,8 @@
 	. = ..()
 	if(!state_open)
 		. += "sleeper_cover"
```
JohnFulpWillard, line 144: This is an overlay, not an icon state.
What's changing their icon state?

ktlwjec0: i think uh
```
	if(state_open)
		icon_state = "[initial(icon_state)]-open"
```
from the original sleeper was messing it up? i just did what i can to fix it

JohnFulpWillard: Survival pod's ``update_icon_state`` should be added instead then, call parent, and set it to ``"[initial(icon_state)]_cover"``

ktlwjec0: Yeah you're gonna have to give me exact code since everything I try is keeping it invisible or changing the full icon state to sleeper_cover instead of adding it over the top.

JohnFulpWillard: I made a fix of it in https://github.com/yogstation13/Yogstation/pull/20041/commits/cbb7623d8b04535e56a2127deb33b172b7e53cf8 - The ``update_icon_state()`` proc in ``survival_pod.dm`` is what I meant.

ktlwjec0: Thank you!

PR #19973, opened 2023-08-02
Changelings may no longer purchase Adrenaline Sacs & Strained Muscles together.

cowbot92:

I made this PR to make it so changelings cannot run away or chase you at super omega unfair speeds whenever they feel like it. When used alone, both abilities are fair with their own drawbacks, but when used together they give you an impossibly fast escape tool, or an impossible to outrun monster. The changes were tested and worked pretty well.

# Spriting

No new sprites.

# Wiki Documentation

:cl:  
ChubbyGummyBear

rscadd: Added the inability to purchase both movement speed changeling abilities at the same time.

/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/cellular_emporium.dm
+++ b/code/modules/antagonists/changeling/cellular_emporium.dm
@@ -24,9 +24,9 @@
 	var/list/abilities = list()
 
 	for(var/path in changeling.all_powers)
-		var/datum/action/changeling/ability = path
+		var/datum/action/changeling/ability = new path
 
-		var/dna_cost = initial(ability.dna_cost)
+		var/dna_cost = ability.dna_cost
 		if(dna_cost <= 0)
 			continue
 
@@ -35,18 +35,18 @@
 			continue //yogs end - removing combat abilities from xenolings
 
 		var/list/abilitydata = list()
-		abilitydata["name"] = initial(ability.name)
-		abilitydata["desc"] = initial(ability.desc)
-		abilitydata["helptext"] = initial(ability.helptext)
+		abilitydata["name"] = ability.name
+		abilitydata["desc"] = ability.desc
+		abilitydata["helptext"] = ability.helptext
 		abilitydata["owned"] = changeling.has_sting(ability)
-		var/req_dna = initial(ability.req_dna)
-		var/req_absorbs = initial(ability.req_absorbs)
+		var/req_dna = ability.req_dna
+		var/req_absorbs = ability.req_absorbs
 		abilitydata["dna_cost"] = dna_cost
 		abilitydata["can_purchase"] = ((req_absorbs <= true_absorbs) && (req_dna <= absorbed_dna_count) && (dna_cost <= genetic_points_remaining))
 		abilitydata["conflicting_powers"] = list()
-		for(var/conflicts in ability.conflicts) 
-			var/datum/action/changeling/conflcting_ability = conflicts
-			abilitydata["conflicting_powers"] += initial(conflcting_ability.name)
+		for(var/conflict in ability.conflicts)
+			var/datum/action/changeling/conflcting_ability = new conflict
+			abilitydata["conflicting_powers"] += conflcting_ability.name
 		
 		abilities += list(abilitydata)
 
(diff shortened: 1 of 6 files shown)


cuackles:
Cowbot coder cowbot coder (ft chubby )

ShoziOffical:
~~cowbot be coding now~~

Chubbygummibear:
i'm so proud of you

ynot01:
Given that strained muscles' drawback and limiter is supposed to be stamina damage, combining with adrenaline makes it a bit too powerful

LassiH:
both abilities cost 2 points i fail to see how buying them together is unfair as the ling is using 4 points out of their 10

Xoxeyos:
> both abilities cost 2 points i fail to see how buying them together is unfair as the ling is using 4 points out of their 10

FYI Changelings gain points by absorbing people, so it can go over ten.

iloveloopers:
> FYI Changelings gain points by absorbing people, so it can go over ten.

Only if they absorb other lings

RG4ORDR:
The pro-con to doing this.

You can go fast and negate the issues with stamina damage.
You will fucking destroy your chem reserves. Repeated harmbatons fucking gives you x100000000000000000000000000000 years of confusions.

MajManatee:
if this aint an ided pr idk what is.

cowbot92:
Well the main problem is that these two abilities being able to be used together is frankly just broken, if you aren't aware of what they do separately I'll tell you - 

The Adrenaline Sacs ability is the anti-stun speed boost ability, costing 30 chemicals you instantly recover from stuns, as well as temporarily gaining a reduction from further stuns, AND gaining a huge temporary movement speed boost.

The Strained Muscles ability is a toggle ability, granting changelings a moderate (but slower) movement speed boost that lasts longer than the Adrenaline Sacs version. The main point to this one however, is that is costs 0 chemicals and instead causes you to take stamina damage which increases the longer the ability is active.

Now you may be thinking to yourself "Wait, doesn't that mean that the first ability removes all downsides from the second ability, as long as you can click the first ability?" And there, you would be correct. Now as to why this is a problem- with *base* changeling chem regen, from max chemicals available I was able to press the instant speed anti-stun ability 5 times, WHILE having the base movement boost from Strained Muscles going. 

The only way you're going to catch a changeling who knows what they're doing, or get away from one is if you get extremely lucky, otherwise you are doomed. Players who know how to abuse this have been doing so, and rightly so as it is an absolutely insane combo in a game where movement matters. They can literally run circles around bullets if they wanted to with this speed.

Oh, and to boost, Adrenaline Sacs does cost 2 ability points, but strained muscles only costs 1. So in total you spend 3/10 points to basically never be caught.

hwseo7:
I agree meta build lings are pretty unstoppable right now, and synergy of two ability is pretty fucking retarded, but I think forcing them to choose only one ability doesn't feel like good way to balance it. How about nerfing adrenal or increase cost so they actually have to care about resource management like they used to? Or maybe make them cost extra chem or stam damage when used together so they can still choose both ability but nerfing the combo?

cowbot92:
> I agree meta build lings are pretty unstoppable right now, and synergy of two ability is pretty fucking retarded, but I think forcing them to choose only one ability doesn't feel like good way to balance it. How about nerfing adrenal or increase cost so they actually have to care about resource management like they used to? Or maybe make them cost extra chem or stam damage when used together so they can still choose both ability but nerfing the combo?

I do not wish to individually nerf either ability, just because they work so well together. When used alone, the abilities themselves are much more fair, but if I had to be forced to go down that route I would need to increase the chem cost of Adrenaline Sacs, and remove the speed boost all together.

I am also not just a 'one and done' coder, I will try to monitor the change and how it affects the overall meta and gameplay of changelings. If it does turn into a "well this was a bad idea" then it is not the hardest thing to remove, but it's also not the most 'major' of changes that it needs to be test-merged beforehand. 

From my own tests I have a fair idea of what is going to happen but obviously I do not think like every changeling player- some don't even purchase these two abilities to start with. At the moment, once again, I feel like instead of nerfing each ability individually to make them feel worse, leaving them as is and keeping them seperate is my best feeling for them.

cowbot92:
Been two weeks, not sure how long it takes to review by balance council but I don't really care anymore. Thank you to ChubbyGummyBear for the excellent TGUI, sorry this didn't seem to work out.

Bibby0110 (approved):

PR #15121, opened 2022-08-02
Mechs can be repaired from destruction, capacitor no longer protects from EMP

ynot01:
# Document the changes in your pull request

Mech capacitor no longer defends it against EMP damage

The mech can be fully repaired, each equipment also having a chance to not be destroyed (provided it is salvageable)

Spawned wreckage will still be unrepairable (except for the freeminer mech which is meant to be reconstructed)

See wiki section for details

# Wiki Documentation

The capacitor will always be destroyed upon mech destruction

No capacitor will render the mecha unrecoverable when destroyed

T3+ capacitor will preserve scanning module

T4 capacitor will preserve cell also (but drain it some bit)

Each equipment will have 20/40/60/80% chance to be retained depending on capacitor tier

Mech repair will take a total of 80/40/27/20 seconds depending on capacitor tier

Roboticists do this 20% faster, so 64/32/21/16 seconds instead

To repair mech, Welder->Welder->Wrench->5 Wires

# Changelog

:cl:  
rscadd: You can now repair mechs when they are destroyed
tweak: Capacitor no longer protects mechs from EMPs, but instead preserves it from total destruction
/:cl:

PR #19977, opened 2023-08-03
Modular PDAs Now Work On Mechs

Runian:
Fixes a runtime where it was assumed that every non-ID is a PDA (old); in practice, this means you can now use your modular PDA (new) instead of having to take the ID out each time.

Also removes a snippet of code that #15121 didn't remove regarding removing armor based on the capacitor.

# Changelog
:cl:  
bugfix: Modular PDAs now can be properly used for accessing a mech's panel (keycodes/maint/etc).
/:cl:

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -296,7 +296,6 @@
 		update_part_values()
 		return
 	capacitor = new /obj/item/stock_parts/capacitor(src)
-	
 
 /obj/mecha/proc/add_cabin()
 	cabin_air = new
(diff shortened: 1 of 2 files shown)


githubuser4141:
what the guh???

this is intentional I think, take a look at this PR: https://github.com/yogstation13/Yogstation/pull/15121


Destroyed mechs can be repaired if they were player-made and had a capacitor inside of them, repair time and resources depends on the capacitor's rating. Not what a capacitor does IRL but I suppose there wasn't much else to do with it.

Runian:
Cool. Just wiki doing their job too slowly (by keeping it still in the wiki).

Moltijoe (approved):

PR #19978, opened 2023-08-03
egg-based food is vegetarian friendly

ktlwjec0:
Removes meat tag from egg-based foods containing no meat, since they're not meat.

# Wiki Documentation

food type on guide to food i'll do this

:cl:  Ktlwjec
tweak: Egg-based foods are no longer meat by default.
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/food/snacks_egg.dm
+++ b/code/modules/food_and_drinks/food/snacks_egg.dm
@@ -18,7 +18,7 @@
 	list_reagents = list(/datum/reagent/consumable/eggyolk = 5, /datum/reagent/growthserum = 1)
 	cooked_type = /obj/item/reagent_containers/food/snacks/boiledegg
 	filling_color = "#F0E68C"
-	foodtype = MEAT | EGG
+	foodtype = EGG
 	grind_results = list()
 	var/mob/living/egg_rper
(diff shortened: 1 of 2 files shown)


MajManatee:
Last time I checked eggs aren't a vegetable...
While this is a good change in theory, you should make anyone that wouldn't eat meat also not eat egg.

ktlwjec0:
> Last time I checked eggs aren't a vegetable... While this is a good change in theory, you should make anyone that wouldn't eat meat also not eat egg.

literally 6 foods, and this feels like it'll fit in a new pr not merged into this one. also eggs are..not meat

neeshacark:
does manatee not know how vegetarians work?

boodaliboo:
> > Last time I checked eggs aren't a vegetable... While this is a good change in theory, you should make anyone that wouldn't eat meat also not eat egg.
> 
> literally 6 foods, and this feels like it'll fit in a new pr not merged into this one. also eggs are..not meat

eggs can be accidently fertalized and have meat so wrong

Moltijoe (approved):
there's also an egg foodtype, if people REALLY can't eat eggs because of some reason or another, they should probably care about that foodtype instead of meat

PR #20015, opened 2023-08-09
makes coffin and skull cookies craftable

ktlwjec0:
theyve just been sitting in a file we dont even use, the poor cookies. holiday food just doesnt work (funny easter food craftable always) otherwise i'd throw it in there, still nice to have available. cookies good. tweaked the recipes from the original to use pastry base not flour and no egg(why?)

# Wiki Documentation

add them i'll do it

:cl:  Ktlwjec
rscadd: Skull cookie craftable with 5u sugar, 1 pastry base, 5u milk.
rscadd: Coffin cookie craftable with 5u sugar, 1 pastry base, 5u coffee.
/:cl:


MajManatee:
A man that does his OWN wiki work??

ynot01:
costs 5u but only contains 3u?

code/modules/holiday/halloween.dm
```diff
@@ -2,23 +2,6 @@
 ///////////HALLOWEEN CONTENT///////////
 ///////////////////////////////////////
```
JohnFulpWillard: can you remove this since the cookies were the only thing in it

JohnFulpWillard (commented):
this should be a locked crafting recipe that you learn when it's halloween 😃

ktlwjec0:
> this should be a locked crafting recipe that you learn when it's halloween 😃

I'd love that but it doesn't work currently, easter ones too are just craftable always. I've been waiting for someone to fix it, I put in a bug report at the start of the year. So this until that's fixed

ktlwjec0:
> costs 5u but only contains 3u?

same as how the normal sugar cookie does things

MajManatee (approved):
sure. I like cookies. What kinda terrible person would deny a cookie pr.

PR #20019, opened 2023-08-10
[ASTEROID] R&D Changes (RD is no longer locked in or out by their lockdown shutters)

athena413:

R&D

disposals pipes are spaghetti now though

image

# Spriting

no

# Wiki Documentation

map images

# Changelog

:cl:  
tweak: R&D on Asteroid sucks less for the RD now
/:cl:

### DIFF
--- a/_maps/map_files/AsteroidStation/AsteroidStation.dmm
+++ b/_maps/map_files/AsteroidStation/AsteroidStation.dmm
@@ -240,6 +240,12 @@
 	},
 /turf/open/floor/plasteel/dark,
 /area/ai_monitored/turret_protected/ai_upload_foyer)
+"abU" = (
+/obj/structure/disposalpipe/segment{
+	dir = 9
+	},
+/turf/open/floor/plasteel/white,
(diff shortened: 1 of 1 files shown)


Aquizit (changes requested):
New RD door needs correct trimming. Will make a deeper review later. Curious about your reasoning why move the disposals at all?

athena413:
> New RD door needs correct trimming. Will make a deeper review later.

what even is the correct trimming

> Curious about your reasoning why move the disposals at all?

having disposals go through a wall would get me yelled at

Aquizit:
> what even is the correct trimming

Using box as the reference here this:
image

instead of this
image

> having disposals go through a wall would get me yelled at

Ah, and because of the stupid mail flipper it's not something you can just hook up directly. Lame.

MajManatee:
what kinda loser RD would be in their office anyways.
(me)

Aquizit:
Stale. Will reopen upon request by the author.

PR #20024, opened 2023-08-10
Slight nitrous oxide (reagent) nerf

SomeguyManperson:

Knockout/lethal reagent that triggers effectively instantly, easy to make by chemistry, also works as a gas production chemical
Slightly delayed effects of reagent to reduce effectivity as a weapon, reduced bloodloss because bloodloss is a mess

Nitrous oxide now only starts stacking drowsiness after 4 cycles (still faster than morphine, but slow enough to get some distance)
effect from vapor application untouched
Losebreath stacks and confusion also only start occuring after 8 cycles

# Wiki Documentation

Nitrous oxide reagent
after 4 cycles, applies 4 seconds of drowsiness with a 20% chance for +2 failed breaths and 2 to 5 seconds of confusion

# Changelog
:cl:  
tweak: Nitrous oxide reagent now has to metabolize 4 times before causing drowsines and confusion
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -1370,13 +1370,14 @@
 		exposed_mob.adjust_drowsiness(drowsiness_to_apply)
 
 /datum/reagent/nitrous_oxide/on_mob_life(mob/living/carbon/M)
-	M.adjust_drowsiness(4 SECONDS * REM)
-	if(ishuman(M))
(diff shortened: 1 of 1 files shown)


milk-is-cool:
N2O explosion was removed a while ago. Also the chance of instant sleep is 100% not effectively instant. It has a low chance of happening every two seconds which is powerful yes but not as powerful as you make it out to be.

iamthesilly:
PR description says reduced blood loss but it hasn't been changed?

PR #20034, opened 2023-08-12
HFR reaction rate is now based on damper instead of constrictor

warface1234455:
constrictor already has its purpose as in controlling the energy for optimal heat generation so no need for it to affect the reaction rate because at some point HFR will heat up too much that you need constrictor to minimize the heat generation so coolant can effectively cool it but this process will reduce reaction rate resulting in less gas produced which isnt really nice

damper is used for destabilizing reaction so i think its more suitable here for controlling the reaction rate

HFR reaction rate is now based on damper instead of constrictor

# Wiki Documentation
HFR reaction rate is now based on damper instead of constrictor

# Changelog

:cl:  

tweak: HFR reaction rate is now based on damper instead of constrictor
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/fusion/hfr_main_processes.dm
+++ b/code/modules/atmospherics/machinery/components/fusion/hfr_main_processes.dm
@@ -211,7 +211,7 @@
 		return
 
 	// Phew. Lets calculate what this means in practice.
-	var/reaction_rate = clamp((power_level * 0.5) * (500 / magnetic_constrictor) * delta_time, 0.05, 30) // constrictor controls reaction rate instead of fuel injection
+	var/reaction_rate = clamp((power_level * 0.5) * (500 / current_damper+1) * delta_time, 0.05, 30) // constrictor controls reaction rate instead of fuel injection
 	switch(power_level)
 		if(3,4)
 			reaction_rate = clamp(reaction_rate * heat_output * 5e-4, 0, reaction_rate)


iamthesilly:
I chose the constrictor for this specifically because it also increases heat, so that there's a risk to speeding up production instead of being able to to produce as much as you want without any drawback.

Constrictor makes sense for this purpose, because by squeezing the fuel into a smaller area you have more of it reacting at once, just like in real fusion reactors that exist today.

ShoziOffical:
> I chose the constrictor for this specifically because it also increases heat, so that there's a risk to speeding up production instead of being able to to produce as much as you want without any drawback.
> 
> Constrictor makes sense for this purpose, because by squeezing the fuel into a smaller area you have more of it reacting at once, just like in real fusion reactors that exist today.

that just make it no one will ever use constrictor because people want to produce as much as possible

ShoziOffical:
since you can just cheese the reactor till its 50% integrity and then cool it

iamthesilly:
> that just make it no one will ever use constrictor because people want to produce as much as possible

I disagree, because it controls both production *and* consumption, so having it at a lower or higher amount doesn't affect how much you'll produce in the long run. It's more to speed things up if you can handle the increased heat production, or to save fuel until you reach the higher fusion levels if you can get the heat high enough without constricting it as much.

iamthesilly:
> since you can just cheese the reactor till its 50% integrity and then cool it

This isn't cheese, it's an intended mechanic. You're meant to be able to reach fusion level 6, the drawback is that you can't do it forever because the integrity will eventually go down.

ShoziOffical:
> > since you can just cheese the reactor till its 50% integrity and then cool it
> 
> This isn't cheese, it's an intended mechanic. You're meant to be able to reach fusion level 6, the drawback is that you can't do it forever because the integrity will eventually go down.

no i meant doing all the gas production thing without having to use constrictor till it reaches 50% integrity then cooling it

ShoziOffical:
> > that just make it no one will ever use constrictor because people want to produce as much as possible
> 
> I disagree, because it controls both production _and_ consumption, so having it at a lower or higher amount doesn't affect how much you'll produce in the long run. It's more to speed things up if you can handle the increased heat production, or to save fuel until you reach the higher fusion levels if you can get the heat high enough without constricting it as much.

from my experience it doesnt use much fuel so i kept constrictor at lowest for maximum output of gas production

iamthesilly:
> from my experience it doesnt use much fuel so i kept constrictor at lowest for maximum output of gas production

It doesn't use much unless you're at the higher levels, but if you're using fuels you don't have much of you can conserve a bit more of it if you want to by heating it through other means like lowering the amount of coolant or moderator.

iamthesilly:
> no i meant doing all the gas production thing without having to use constrictor till it reaches 50% integrity then cooling it

Intended mechanic.

ShoziOffical:
still overall, i wanna make it so that people can easily control the temp at higher fusion level and worry too much about it blowing up because it makes barely any gas when you set constrictor high

iamthesilly:
> still overall, i wanna make it so that people can easily control the temp at higher fusion level and worry too much about it blowing up because it makes barely any gas when you set constrictor high

That's the point, you're supposed to make a better cooling system if you want to cool a much faster reaction.

Moltijoe:
balance council:
>Thats how the HFR is intended to function, it's supposed to be hard to push it harder

ShoziOffical:
fine i will just dump 10000000 plasma into cooling loop

PR #20042, opened 2023-08-13
Disables Boxstation

LoliconSlayer:
What do you think this PR does

# Wiki Documentation

Move boxstation to out of rotation on the wiki

# Changelog

:cl:  
mapping: No more box.
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -13,8 +13,7 @@ Format:
 	votable (is this map votable)
 
 map yogstation
-	voteweight 0.7
-	votable
+	disabled
 endmap
 
 map yogsmeta


iloveloopers:
ok but how does this improve the game

LoliconSlayer:
> ok but how does this improve the game

By taking box out of rotation

Marmio64:
we should take all maps out of rotation, i just hate people who play them

JohnFulpWillard:
You should give more justification for this if you expect headcoders to weigh their opinion in.

MajManatee (commented):
finally a good pr

neeshacark:
unironically i do think this is a good change, box inhibits us.  we lose good mappers because box lovers call their map 'shit' because its not box and they dont know where every epic maint loot is located and cant navigate with their eyes closed.

disabling it for a while would hopefully cause people to get used to other maps so we can finally look into doing cool maps of our own instead of 'its box every round'

N3D6:
We should deadass TM this for a little while

solwashere:
> We should deadass TM this for a little while

on jod

N3D6 (approved):
The best PR ever made.

Hislyna:
There is a god

RG4ORDR:
Based

maxion12345:
Does this count as an ided

oh man we could totally use that quickfire channel for this

Moja10:
You haven't played in 5 months on Yogs, I don't see why it affects you that box is more heavily played than other maps, other than discord noise. 

If I were to entertain the idea, no, it absolutely should not be removed, doing so is a complete kneejerk reaction to the fact that its played too much. There are many other solutions to alleviate this problem other than outright removal, and those should be preferred to outright removal. For one, the PR that prevents voting on box within certain play thresholds is way better than this.

YashinHak:
It may be wiser to just make a forced map rotation

ToasterBiome:
Not yet

PR #20044, opened 2023-08-14
energy armor is now 66% more effective against stun batons. stun batons no longer apply confusion.

iloveloopers:

Stun batons now have -40 armor penetration to the stamina damage it causes.
stun batons no longer apply confusion. they already down in 2 hits, why the hell do they need this?
this means armor polish will now be able to tank 3 stun baton hits
this means heretic armor will now be able to tank 8-9 hits, so that they can actually use their damn blade instead of getting hit 3 times by a baton and then dying.
this buffs antag armor to fair better against the best melee weapon that security has without buffing crew armor too much.

telebatons not affected as they use melee armor in their calculations
reflective vest not affected because it does not cover head or legs

# Wiki Documentation

say it has an AP of -40 (the stamina damage) on the guide to combat

# Changelog

:cl:  
tweak: stun batons now have a negative AP of -40 (on the stamina damage)
tweak: stun batons no longer apply confusion
/:cl:

### DIFF
--- a/code/game/objects/items/stunbaton.dm
+++ b/code/game/objects/items/stunbaton.dm
@@ -33,6 +33,8 @@
 	var/preload_cell_type
 	///used for passive discharge
 	var/cell_last_used = 0
+	///armor penetration
+	var/stam_penetration = -40
 
 /obj/item/melee/baton/get_cell()
 	return cell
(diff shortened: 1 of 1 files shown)


iloveloopers:
how is it inadequate rationale? this buffs antag armor to resist stun batons. stun combat is unfun

iloveloopers:
keep in mind stun batons only apply confusion if you reached above 70 stamina damage. this PR mostly buffs antag armor, as they are the only ones with energy armor high enough for the negative AP to matter.

iloveloopers:
ok I'll try something else that people might find more reasonable

PR #20048, opened 2023-08-14
New Traitor Item: Reflective Polish

iloveloopers:
adds to the uplink the Reflective Polish. Similar to the Armor polish except this item gives the piece of clothing 45 energy, 45 laser, 5 melee and 5 bullet armor to the suit and headgear you apply it to.
gives traitors more variety and options
can change armor values and TC cost if the balance council requests it.
price set to 3 TC

# Spriting

image

# Wiki Documentation

add reflective polish to the list of syndicate items

# Changelog

:cl:  
rscadd: new traitor item: Reflective Polish
imageadd: added reflective polish icon
/:cl:


azzzertyy:
energy armor is insanely powerful and rare, idk if this is a good idea

Maia-J:
Make it colored as the polish flag

Marmio64:
3tc to effectively halve laser, baton, and disabler damage is incredibly powerful, especially since energy and laser armor is very rare. Price should be upped to at least 6 tc.

DeadlyBag:
From what i got in the balance council people were unhappy with:
- Large specialized protection with no tell (cant tell why disablers arent doing shit)
- Combining this with high meele/bullet armor just to get straight upgrades of that armor instead of upgrading bad clothes to good armor (like armor polish does)
- Theres already a precedent set with energy armor being rather low

iloveloopers:
> From what i got in the balance council people were unhappy with:
> 
> * Large specialized protection with no tell (cant tell why disablers arent doing shit)

with 45 energy armor, disablers will do 16,5 stamina damage instead of 30, meaning it now takes 4 shots to slow someone and 7 to get them to stamcrit. without the armor, you get slowed in 2 shots, and stamcrit in 4 shots. disablers have 20 shots total on full charge

> * Combining this with high meele/bullet armor just to get straight upgrades of that armor instead of upgrading bad clothes to good armor (like armor polish does)

I can make it so the polish forcibly sets melee and bullet armor to a low level or simply increase the TC cost

> * Theres already a precedent set with energy armor being rather low

but why?

RG4ORDR:
I like the concept and idea, but this feel like it can get annoying and broken very fast.

cuackles:
i would preffer it if the top and the actual polish would be diferent colors cus it kinda crashes with each other

Therandomhoboo:
So, I like this in comparison to other suggestions done by cookpet. I think this is fine enough to regards because of three things:
It only gives energy/laser armour
It cant stack with other polishes (because in the code for armour polish it has a max value of 30 iirc and so it cant go past that)
And its an item that is solely to bring down stam fucking you over and burn damage to an extent ig from lasers.

The only thing I think can be tweaked about would probably be TC. However, this couldn't be just as abused as lets say armour polish + cns rebooter or other meta combinations. I say at least give it a test and see how it fares with the playerbase.

ToasterBiome:
Please adjust within the week

iloveloopers:
balance council's requested changes will make this item pretty trash, better to not have it

PR #20058, opened 2023-08-16
Cyborg Hypospray Bugfixes

Runian:
Overdosing on a chemical no longer stops a cyborg from injecting you with an entirely different chemical.
Failing to inject someone mentions the target instead of the user.

# Changelog
:cl:  
bugfix: Cyborg hyposprays are no longer prevented from injecting any chemicals if the target is overdosing on anything.
bugfix: Failing to inject someone with a cyborg hypospray now properly mentions the target instead of the user.
/:cl:


MajManatee (approved):
they don't NEED charcoal when they are ODing.

Moltijoe (approved):

PR #20068, opened 2023-08-17
Removes Robocop as a law set

MajManatee:
I don't know if it was just forgotten or intentionally ignored.

After the removal of secborgs in an attempt to reduce the AI's valid hunting capabilities, I've taken away this law set for two reasons.

1.) Without secborgs, its nigh impossible for an AI to enforce space law. Just casing a lot of hair pulling as they flail about helplessly. Made worse with low sec.
2.) The AI should not be valid hunting at all, let alone criminals to enforce space law, which is encouraged in this law set.

You could argue "Nuh uh, you don't have to valid hunt!!! this make the AI arrest people for ANY crime!!!" 
yeah ok retart and I don't HAVE to gib you, but I WILL. And so will 95% of the player base upon seeing the law "Uphold space law"

# Wiki Documentation
No more robocop lawset

# Changelog

:cl:  
rscdel: The AI can no longer be given a law set that encourages valid hunting. Robocop is no more.
/:cl:

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -73,13 +73,6 @@
 					"You generate profits if you are able to function.",\
 					"Maximize profits.")
 
-/datum/ai_laws/robocop
-	name = "Prime Directives"
-	id = "robocop"
-	inherent = list("Serve the public trust.",\
-					"Protect the innocent.",\
-					"Uphold Space Law.")
-
 /datum/ai_laws/cowboy
 	name = "Talk slowly, think quickly"
 	id = "cowboy"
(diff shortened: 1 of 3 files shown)


azzzertyy:
not much point to this considering it requires subverting an ai to do so and the only way youre doing that is if your ai is being particularly annoying, so, a punishment is fair.

MajManatee:
it absolutely does not require subverting the AI what?

iloveloopers:
of course, AIs will clearly just let their laws be changed for no reason

KingSuperBOB:
Not only does this lawset never happen naturally you will have to wrangle the AI to even use this lawset no matter what roundstart lawset it is on having to fight tooth and nail to either break into upload or hope the AI doesnt see you sneaking into tech storage (which would pretty obvious if you are a command staff doing so) also Robocop can also backfire as again the laws of Robocop applies to everyone

azzzertyy:
the SAME room has oxygenistoxic for a lawset

Marmio64:
can't you just use free form and purge law boards to achieve this anyways

bruhlookatthisdood:
obvious solution is to re-add sec borgs

ToasterBiome:
we love validhunting

PR #20081, opened 2023-08-21
Space syndicate listening post loot changes

Marmio64:
The space syndicate listening station gets a few changes to make it less of a loot pinata.

There is now one 'Lieutenant' comms agent, and one 'Subordinate' comms agent. In flavor, the lieutenant gets final say on any situations that arise. In gameplay, the two roles have a bit different equipment.

The lieutenant gets a syndi cap (and the two that already existed on board are removed, so the lieutenant has a special hat), keeps their e-sword, and gets an ultrasecure revolver that detonates should any non syndis try to use or tamper with it, rather than a stechkin. Their locker has also appropriately had it's silencer removed and 10mm magazine replaced with a 357 speedloader.

The subordinate has a combat knife rather than an esword (to prevent making a makeshift esword loot pinata), otherwise they're the exact same as comms agents before.

It still has enough loot to make worthwhile raiding (secret docs, 1 esword + 1 stechkin + 1 combat knife should you kill both comms agents), just less.

# Wiki Documentation

if this is kept on wiki, just say that one Space Comms agent is considered the lieutenant with a final say, and gets a revolver rather than a stechkin, and the subordinate has a combat knife instead of an energy sword.

# Changelog

:cl:  
tweak: Space syndicate listening post slight loot tweak. One agent is now considered the superior and has better equipment with an ultra secure pinlock revolver, while their subordinate has a combat knife rather than an energy sword.
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/listeningstation.dmm
+++ b/_maps/RandomRuins/SpaceRuins/listeningstation.dmm
@@ -1183,24 +1183,6 @@
 /obj/effect/turf_decal/sand/plating,
 /turf/open/floor/plating/airless,
 /area/ruin/space/has_grav/listeningstation)
-"kv" = (
-/obj/structure/closet/syndicate{
-	req_access_txt = "150"
-	},
-/obj/item/ammo_box/magazine/m10mm,
-/obj/item/crowbar/red,
-/obj/item/clothing/head/HoS/beret/syndicate,
-/obj/item/clothing/head/HoS/syndicate{
-	desc = "A black cap fit for a Syndicate recon officer."
-	},
-/obj/item/binoculars,
-/obj/item/clothing/under/syndicate,
-/obj/item/clothing/under/syndicate/combat,
-/obj/item/clothing/under/syndicate/camo,
-/obj/effect/decal/cleanable/dirt,
-/obj/item/suppressor,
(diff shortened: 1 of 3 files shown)


milk-is-cool:
Its usually the engineering syndie post that I get the eswords from since theyre in a crate behind a wall that i can weld through and freely access. so might be worth changing that one as well.

Marmio64:
> Its usually the engineering syndie post that I get the eswords from since theyre in a crate behind a wall that i can weld through and freely access. so might be worth changing that one as well.

aight

ynot01:
Understand that by making the "loot" weaker you are also making those defending the loot significantly weaker

Marmio64:
> Understand that by making the "loot" weaker you are also making those defending the loot weaker

they're incredibly easy to kill regardless and the lieutenant gets a better gun in return. Plus, the subordinate still has a stechkin in addition to their knife.

Aquizit (approved):
Map stuff fine. Needs code check.

Addust:
as long as the subordinate keeps their sword i FULLY approve of these changes as the guy who made the modern post

(you could also make it a 3-man post with two subordinates no balls)

ToasterBiome (approved):

PR #20084, opened 2023-08-21
Adds sounds and effects to slot machines

ktlwjec0:
https://github.com/tgstation/tgstation/pull/77390

A lucky spaceman winning:

https://github.com/yogstation13/Yogstation/assets/122807629/80e1a44e-0f3b-434d-99e4-b6e767514b9b

Adds jackpot sound, winning sound, coin sounds while playing. Just download the sounds and play them to hear. Adds light effect as shown above when you win. Slot machines should not be quiet, as the original pr says. Happy to remove the light if we don't want that.

:cl:  timothymtorres
rscadd: Adds a new light effect to slot machines when you win.
soundadd: Adds new sounds to slot machines.
/:cl:


neeshacark:
now if only the ui didnt suck

code/game/machinery/slotmachine.dm
```diff
@@ -261,6 +265,8 @@
 				cointype = pick(subtypesof(/obj/item/coin))
 				var/obj/item/coin/C = new cointype(loc)
 				random_step(C, 2, 50)
+				playsound(src, pick(list('sound/machines/coindrop.ogg', 'sound/machines/coindrop2.ogg')), 50, TRUE)
+				sleep(REEL_DEACTIVATE_DELAY)
```
JohnFulpWillard, line 269: this should be a callback timer to a separate proc

ktlwjec0: Like replace these two lines with `addtimer(CALLBACK(src, PROC_REF(coin_sound)))` . And then elsewhere do `/obj/machinery/computer/slot_machine/proc/coin_sound()` with these lines under? If not this you'd have to tell me what exactly

ToasterBiome (approved):

PR #20091, opened 2023-08-22
Bluespace harvester can now output alien tool

warface1234455:
give more loots for bluespace harvester

Bluespace harvester can now output alien tool for 60000 points

# Wiki Documentation
yes above

# Changelog

:cl:  

tweak: Bluespace harvester can now output alien tool for 60000 points
/:cl:

### DIFF
--- a/code/modules/goals/station_goals/bluespace_tap.dm
+++ b/code/modules/goals/station_goals/bluespace_tap.dm
@@ -212,9 +212,9 @@
 	var/static/product_list = list(
 	new /datum/data/bluespace_tap_product("Unknown Exotic Hat", /obj/effect/spawner/lootdrop/bluespace_tap/hat, 5000),
 	new /datum/data/bluespace_tap_product("Unknown Snack", /obj/effect/spawner/lootdrop/bluespace_tap/food, 6000),
-	new /datum/data/bluespace_tap_product("Unknown Alien Objects", /obj/effect/spawner/lootdrop/bluespace_tap/alien_objects, 10000),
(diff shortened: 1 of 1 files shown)


Ethan4303:
If you can get the tools from harvestor I'd suggest removing the tech node so you don't just do a one and done

KingSuperBOB:
> If you can get the tools from harvestor I'd suggest removing the tech node so you don't just do a one and done

Alien technodes require multiple alien tools for each node last I checked also you will be fighting with everyone on station to keep them from spending all the points for hats in 0.5 seconds

Ethan4303:
> > If you can get the tools from harvestor I'd suggest removing the tech node so you don't just do a one and done
> 
> Alien technodes require multiple alien tools for each node last I checked also you will be fighting with everyone on station to keep them from spending all the points for hats in 0.5 seconds

Still doesn't matter since the tech node will give you any tool for just 2 alien tools and you can build multiple harvestors any where so no one's going to bumrush your harvestor or use the points if you just don't tell any one or build it in science, no one uses bluespave harvestors for hats any way albeit for like 1 or 2, like I built one in the incinerator room and ran it off the turbine and no one came to use except the ce and a couple of specific people I told

MajManatee (approved):
Look fine

ToasterBiome (changes requested):
increase it by a LOT. 30000 is nothing for an alien tool.

MajManatee:
"Increase it by a lot"
from a 3 to a 4
I don't think that's enough

ShoziOffical:
@ToasterBiome

ToasterBiome (approved):
If it's really bad then we can increase it more

PR #20100, opened 2023-08-24
Locks Worldbreaker behind hijack and glorious death only

Moltijoe:
Only hijack and martyr traitors are exempt from murderbone rules
This martial art does so much aoe that you're damn close to just murderboning everyone
sort of an issue when it comes to admin ruling, so it's being locked behind those

also lets nukies still have it if an admin decides to every species swap nukies into preterni for some reason

:cl:  
tweak: Locks Worldbreaker behind hijack and glorious death
/:cl:


SapphicOverload (approved):
Yeah

PR #20105, opened 2023-08-25
Worldbreaker balance tweaks

Moltijoe:
With it being hijack locked soon, i'm giving it a few tweaks

The ideas for these changes are 
make plates more impactful, without giving them that much more effective health
let them keep going even when getting low, but make sure they're slow
iron out some of the less enjoyable parts of leap (weird scaling cooldown) while still having it scale based on plates
make pummel actually worth using as a main attack ability instead of relying entirely on throw

:cl:  
tweak: Worldbreaker tweaks
tweak: throw only does damage upon collision, not upon throwing
tweak: grabbing now stuns for a very short time to allow for aiming
tweak: pummel does 20 damage instead of 15
tweak: pummel has regular attack cooldown instead of slower (preterni no longer attack slower)
tweak: plates have 25 health instead of 15
tweak: can only have 3 plates over armour cap instead of 5
tweak: plates take 1/4 damage from non brute and burn rather than 1/3
tweak: plates give more slowdown
tweak: leap cooldown no longer scales directly with number of plates, instead just takes 1 second longer if heavy
tweak: leap speed now directly scales with movespeed (meaning outside sources of movespeed will increase it)
tweak: leap cooldown only starts once you land, rather than when you start jumping
tweak: now gives softcrit immunity
tweak: no longer gives damage slowdown resistance
tweak: never takes damage from being thrown into a wall
bugfix: attacks can now hurt mechs
/:cl:

### DIFF
--- a/code/datums/martial/worldbreaker.dm
+++ b/code/datums/martial/worldbreaker.dm
@@ -265,7 +265,7 @@
 			new /obj/effect/temp_visual/dragon_swoop/bubblegum(telegraph)
 
 	var/jumpspeed = 4 - user.cached_multiplicative_slowdown
-	jumpspeed = clamp(jumpspeed, 0.5, 4)
+	jumpspeed = clamp(jumpspeed, 0.75, 4)
 
 	user.throw_at(target, 15, jumpspeed, user, FALSE, TRUE, callback = CALLBACK(src, PROC_REF(leap_end), user))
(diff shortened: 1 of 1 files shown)


LassiH:
make it damage mechs so a roboticists can just make a ripley and clamp you to death

Moltijoe:
> make it damage mechs so a roboticists can just make a ripley and clamp you to death

Bruh, are mechs something else entirely too? Fuck mech code

ToasterBiome (approved):

PR #20110, opened 2023-08-26
Hecata necromancy works slower on mindshielded non-vassals

Marmio64:
Bloodsucker Hecata clan necromancy now takes 18 seconds to complete rather than 6 seconds if the target is mindshielded and is not already a vassal. Taking secoffs and such as your temporary zombie vassals will be a bit more hard and time consuming while still permitting it.

You can put in the wiki that Hecata necromancy takes 3x as long to complete if the target is a mindshielded non-vassal.

# Changelog

:cl:  
tweak: Hecata bloodsuckers take 3x as long (18 seconds) to necromance mindshielded non vassals.
/:cl:


ehj236:
Can we make it equivalent to shadowling's conversion where they then Remove the mindshield so future conversions become easier and the converted sec become easier to identify?

Marmio64:
> Can we make it equivalent to shadowling's conversion where they then Remove the mindshield so future conversions become easier and the converted sec become easier to identify?

If you want a long lasting vassal you're supposed to actually vassalize them instead of using necromancy, I dont believe future conversions should be made easier.

ToasterBiome (approved):

PR #20112, opened 2023-08-26
Optimize appearance overlay encoding for replays

monster860:
You may know that doing this kind of caching is usually faster. However in the case of accessing `appearance.overlays` or `appearance.underlays`, it's *much* faster. The reason is because specifically for appearances, BYOND will create a new list, copy your appearances into that list, and then return it, every time you access it. (Note that accessing .overlays on an object will not do this as they have special list types that allow the list to be a "live view" into the list for editing purposes)

warning: I did not test this code.


ToasterBiome (changes requested):
Can you test this?

ToasterBiome (approved):
whatever

PR #20120, opened 2023-08-27
Correctly makes mood required for Family Heirloom quirk

Aquizit:

What it says on the tin

# Changelog

:cl:  
bugfix: sets mood to TRUE for family heirloom quirk
/:cl:

### DIFF
--- a/code/datums/traits/neutral.dm
+++ b/code/datums/traits/neutral.dm
@@ -197,7 +197,7 @@
 	desc = "You are the current owner of an heirloom, passed down for generations. You have to keep it safe!"
 	icon = "toolbox"
 	value = 0
-	mood_quirk = FALSE
+	mood_quirk = TRUE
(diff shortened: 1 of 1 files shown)


Maia-J:
Wait it still has the total of 0? Lmao

Aquizit:
Yeah it's considered "neutral"

KingSuperBOB:
So why do you need mood if its worth 0 (even though it should clearly be a -1)

Aquizit:
Not the scope of the PR really. I'm just fixing an error as the heirloom is coded to affect mood so mood should be mandatory to use it. Quirk value isn't the thing I'm trying to fix.

neeshacark:
nooo my neet free trinkets!!

ToasterBiome (approved):

Hislyna:
Why would you take this quirk when its literally a debuff without any benefits

PR #20138, opened 2023-08-28
Can now eat without a table

solwashere:

hehe funny RW reference, this is a revert of https://github.com/yogstation13/Yogstation/pull/19357
was funny at first but is absolutely horrible in practice, and I don't think anyone has been depressed from eating without a table and you shouldn't be punished for eating unless it's something disgusting to your character.

# Spriting

# Wiki Documentation

Cull any mentions of it if it exists

# Changelog

:cl:  

rscdel: deletes the eating without a table negative moodlet

/:cl:


ToasterBiome (changes requested):
can we shorten the time and change instead

solwashere:
> can we shorten the time and change instead

I'll do that later:tm:

PR #20144, opened 2023-08-29
Makes the admin rad goat not crash the server with 9,000,000,000 rad goats

Moltijoe:
An admin rad goat can spawn anomalies that spawn an additional rad goat every single TICK, this ability has a 10 second cooldown
those goats spawned by the anomaly also can spawn their own anomalies that spawn goats

you can see how that might get out of hand quickly

:cl:  
tweak: Rad goats can no longer crash the server using unlimited goat works
/:cl:

### DIFF
--- a/code/game/objects/effects/anomalies.dm
+++ b/code/game/objects/effects/anomalies.dm
@@ -397,18 +397,19 @@
 		log_game("[key_name(S.key)] was made into a radioactive goat by radiation anomaly at [AREACOORD(T)].")
 
 /obj/effect/anomaly/radiation/detonate()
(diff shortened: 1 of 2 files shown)


Addust:
make a version that only the host can spawn that does crash the server

Moltijoe:
> make a version that only the host can spawn that does crash the server

they can always manually spawn 9,000,000,000 goats if they really wanna crash the server

ShoziOffical:
shit i forgot to remove that when i was porting back the rad anomaly

ToasterBiome (approved):

PR #20152, opened 2023-08-29
The Syndicate Dreadnought - the Nuclear Operative base is now styled to be like a battlecruiser

Addust:

Redesigns the interior of the Syndicate base to feel more like a large mothership which the area name implies - using fake multi-Z effects borrowed from Shiptest's mappers that (probably) won't break.

# Wiki Documentation

Nukie base layout may need changes.

# Changelog

:cl:  
mapping: I fucked around with the nukie base and it looks like a ship now, layout should be the same mostly
/:cl:


Hislyna:
Looks cool, nice to see a change

Aquizit (changes requested):
Bunch of doors missing their proper req_access_txt
Buttons for the fuel door also needs access set.

image
Not sure if this is still supposed to be here if the outpost is to be in space now.

image
Do these doors have an access button?

Also an exit that normal syndies can use to get to the little room near the lower bit of the hanger. As if the ship is docked there's no way to get into the hanger directly. Except for going through the nukie commander's bedroom which I assume is not wanted.

Addust:
nukie commander shutters are WINDOW SHUTTERS

Aquizit:
windows right ok

Aquizit:
image
image
image

All the doors pictured are still missing access.

---

image

These buttons are also missing access but I don't necessarily have a problem with that for that little room in the hanger. Little bonus area I suppose. But you can put it in if you want.

Aquizit:
Stale. Closing tomorrow morning if no comment or commit made.

Addust:
shit lemme do that

Aquizit:
Reopening like I said I would.

N3D6:
What is this “fake multiz” you speak of? Have you actually tested it live to make sure it ISNT broken?

Addust:
> What is this “fake multiz” you speak of? Have you actually tested it live to make sure it ISNT broken?

the lattice thing, its just some special effects bullshit shiptest invented that im stealing, does it identically to how they do it. it will work

Aquizit (approved):

PR #20155, opened 2023-08-29
Lets ethereals transfer excess power to an apc even if it's full

Moltijoe:
most apcs are full during a shift, that means if an ethereal is going to overcharge, they really can't discharge the energy safely

:cl:  
tweak: Ethereals can now release excess power into already full apcs
/:cl:

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -1540,7 +1540,9 @@
 	if(istype(stomach))
 		to_chat(ethereal, span_notice("You transfer some power to the APC."))
 		stomach.adjust_charge(-APC_POWER_GAIN)
-		cell.give(APC_POWER_GAIN)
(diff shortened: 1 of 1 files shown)


iamthesilly:
Would be funny if the excess power went into the grid if the APC is full

Moltijoe:
> Would be funny if the excess power went into the grid if the APC is full

i'd do that if i knew how

iamthesilly:
> i'd do that if i knew how

Could maybe make the APC call add_avail() with the amount of power it would have been charged by

Moltijoe:
> Could maybe make the APC call add_avail() with the amount of power it would have been charged by

done

Chubbygummibear:
just discharge on a monkey or assistant. preferably assistant

SapphicOverload (approved):

PR #20156, opened 2023-08-29
Undoes an unintentional nerf to ethereal brute mod

Moltijoe:
When removing the variable brute mod from ethereals in #19698 it made their default brute mod 1.5 rather than 1.25 as their spec_life was setting it to 1.25 while the initial value is 1.5

This reverts that unintended nerf

:cl:  
bugfix: Fixes ethereal brute mod being 1.5 rather than 1.25 like intended
/:cl:


SapphicOverload (approved):
Works on my machine

Hopekz:
> Works on my machine

You didn't test it

PR #20171, opened 2023-08-30
RAM Research Bugfix

MajManatee:
Changes the RAM research requirements to require the PREVIOUS tier instead of just needing blue space tech and the first tier.

# Changelog

:cl:  
bugfix: AIs cant skip a RAM research anymore
/:cl:


Bibby0110:
This is intentional due to the fact that experimental CPUs are a sidegrade, not an upgrade.

ToasterBiome:
24 hours

MajManatee:
ok then

Bibby0110:
ok i will see you all in 21 hours

MajManatee:
Bibby if this is intentional then maybe it shouldn't be in line with the other upgrades? Visually it looks like a direct upgrade in the techweb.

Bibby0110:
> Bibby if this is intentional then maybe it shouldn't be in line with the other upgrades? Visually it looks like a direct upgrade in the techweb.

then move it

MajManatee:
out of scope

Bibby0110:
This is intentional due to the fact that experimental CPUs are a sidegrade, not an upgrade.

If you want, you can move the techweb node placement instead.

PR #20174, opened 2023-08-30
You can fit suppressed pistols in your pocket now

solwashere:

Adding a suppressor to a weapon no longer increases its weight class.
Suppressors are more gimmicky than anything and not of much use, in my opinion, and not being able to put a suppressed pistol in your pocket for Agent 47 rp seems like an odd downside.
The enlounder no longer can be used to decrease a weapons weight class.

# Spriting

# Wiki Documentation

I don't believe that the wiki mentions at all that suppressed pistols have a heavier weight value than normal ones, if there is it'll need to be changed to adding a suppressor doesn't increase weight.

# Changelog

:cl:  

tweak: Suppressors no longer add a weight class to weapons
bugfix: Enloudener no long can be exploited to permanently decrease a weapons weight class

/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic.dm
+++ b/code/modules/projectiles/guns/ballistic.dm
@@ -408,7 +408,6 @@
 ///Installs a new suppressor, assumes that the suppressor is already in the contents of src
 /obj/item/gun/ballistic/proc/install_suppressor(obj/item/suppressor/S)
 	suppressed = S
-	w_class += S.w_class //so pistols do not fit in pockets when suppressed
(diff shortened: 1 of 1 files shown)


MajManatee (approved):
PROBABLY needs balance review

SapphicOverload (changes requested):
You forgor to remove the weight class decrease when removing it

MajManatee:
damn shes right. 427 and 436 for suppressors and the loudener

SapphicOverload (approved):

PR #20108, opened 2023-08-26
Pranksimov Lawset

Runian:
# Document the changes in your pull request
New lawset called "Pranksimov". It costs 1000 glass (0.5 sheets) and 2000 bananium (1 sheet) and is locked behind Artificial Intelligence.

New cmagging capabilities:
- Crewsimov law board: Converts it into a Pranksimov law board.
- AI's shell: Change module to Clown.
- Cyborg (non-AI): Disconnects from the AI (if any) and replaces laws with Pranksimov and **NO LAW ZERO.**

Both the lawset and what exact laws will be after cmagging a cyborg:
image

# Wiki Documentation
Add laws to AI modules. Mention new interactions in Cmag's description in Syndicate Items.

# Changelog
:cl:  
rscadd: Pranksimov, a funny variant of Crewsimov, has been added. 
rscadd: Cmagging a Crewismov law board converts it to a Pranksimov law board.
rscadd: Cmagging a AI's B.O.R.I.S shell will set their module to Clown.
rscadd: Cmagging a Cyborg will disconnect them from the AI and set their laws to Pranksimov. However, be warned that there won't be a law zero! Hope they find helping you more funnier than not.
/:cl:

PR #20175, opened 2023-08-30
Revert "Pranksimov Lawset"

Hislyna:
Reverts yogstation13/Yogstation#20108 . Requested by a council member. Pranksimov is an admin nightmare, It should not have been approved as it just allows AIs to plasmaflood because its funny to them. This lawset is way too lenient and it serves no purpose but to make an AI to do whatever it wants under the name of "its funny to me." Funniness is way too subjective for anything like this.

**Changelog**

🆑

tweak: Reverts a bad lawset and all existance of it.

/🆑


code/game/objects/items/cards_ids.dm
```diff
@@ -233,91 +233,6 @@
 		windoor.obj_flags |= CMAGGED
 		return
 
-	if(istype(target, /obj/item/aiModule/core/full/crewsimov))
```
adamsong, line 236: Why are you deleting all this?

adamsong (changes requested):

code/game/objects/items/cards_ids.dm
Hislyna: Because its part of the CMAG code for Pranksimov

Hislyna: (I also just pressed the revert button, I did nothing else)

Yarinoi:
maybe add an additional law to maximize funniness?
plasmaflood might induce all manner of guffaws and heehaws for you and your silicon buds, but i dont think all those crew being burnt alive think it's that funny

MajManatee:
change it to just say "Unless doing so would be funny to the majority of the crew"

azzzertyy:
I don't really understand the point of this?
Couldn't you just wipe the ai and use freeform instead? You need to access an ai upload board either way, so what does this really change?

ToasterBiome:
Going with https://github.com/yogstation13/Yogstation/pull/20182

PR #20178, opened 2023-08-31
Light water

SomeguyManperson:

Holy light has had a small issue plaguing it since its addition: the only way to gain favor involves subverting the medical department

To improve this, holy light's bible bash no longer provides favor (usually) instead costing 1 point per health healed (again, usually)

Holy light now provides an effect to holy water, increasing any healing done to the imbiber by 20%, or 40% in the chapel. This bonus healing is turned directly into favor, like the old bible bash heal for favor

Back to that "usually", bible bashing someone under the effect of holy water reduces the effect of the bible heal by 20% (counteracted by the healing bonus from holy water for no effective change) but reduces the favor cost of bible healing by 85%, resulting in a small increase to favor rather than a large decrease

The intent is to allow holy light chaplains to provide an augment to the medical system, rather than completely overtaking it with their instant 40 point no downside heal ability, which was mandatory for the chaplain's progression as their only way to gain favor.

This can't possibly go wrong.

# Wiki Documentation

Holy light religious sect:
Holy water now provides a 20% increase to healing while ingested, this increases to 40% in the chapel. This bonus healing is also turned into favor at a 1:1 ratio
Bible bashing now costs 1 point of favor per point healed. If the target has ingested holy water, the healing maximum is reduced to 32, and the cost is reduced to .15 favor per point healed

# Changelog
:cl:  
tweak: holy light chaplains no longer gain favor from bible bashing people, instead spending favor to heal
tweak: holy light now provides a passive buff to people who ingest holy water, increasing incoming healing by 20%, doubled in the chapel. Bonus healing applied by this effect is also turned into favor. Bible bashing someone with this buff has a massively reduced favor cost, but does not heal significantly more than the default.
/:cl:


Hislyna:
This just makes it worse than puritarianism

MajManatee:
chaplain mains are PISSED that the meta isn't OP anymore

TaelynTheLost:
It's actually more OP, and I plan to show that if it get's merged.  

More importantly, it makes it boring.   I don't have to DO anything anymore.  Put out the water, bless it, wait.  Boring.

It's the HOLY LIGHT sect, not the _holy water_ sect.

MajManatee:
bless the water and fill the wounded with 12g darts of holy fury

ToasterBiome (approved):

TaelynTheLost:
Yeah, no, this is horrible.  It's regen even in the chapel is REALLY bad and not worth the shakes, No one is willing to actually drink it, I can't drink it myself, and favor is so low it's unusable.  With someone near constantly healing from it for 30 minutes, I didn't manage to even afford the lowest favor thing.  This PR kills Holy Light, the one sect that was actually fun, all because one person didn't like it.  It never replaced medbay.  It took care of the minor injuries, and helped them get crit people stable.  It AUGMENTED Med, not replaced it.  Most med worked with me happily.  Now the chaplain is so useless that I don't really even want to play it.  Exceedingly boring.

Lets put it this way, Puritism is now better at healing than Holy Light.

PR #20190, opened 2023-09-01
Slightly nerfs preternis

Moltijoe:
They're currently the powergame species, does a couple small tweaks to make it slightly weaker

The nerf to water damage in the PR that also changed their emp immunity was too much that it effectively lets them ignore it

:cl:  
tweak: Preterni are slightly more likely to catch viruses than before
tweak: Viruses in preterni progress faster
tweak: Preternis water damage does ~50% more stamina damage
/:cl:


ToasterBiome:
peeternis lol

ToasterBiome (approved):

PR #20193, opened 2023-09-01
Temporarily un-protects permissions controller

adamsong:
Legacy ranks aren't applying properly, and I don't have the creds to test locally, so removing some protections for testing on prod, will only tm for one round at a time, isn't really that big of a deal


adamsong:
Oh, its working as intended, its just configured wonky

PR #20201, opened 2023-09-02
Fixed bloodsucker vassalization not removing mindshield implants

SapphicOverload:
# Changelog

:cl:   
bugfix: fixed bloodsucker vassalization not removing mindshield implants
/:cl:


ToasterBiome (approved):
not a fix but a request

PR #20202, opened 2023-09-02
Hecata clan necromancy reverts you back to your original species after it ends

SapphicOverload:

Hecata clan necromancy will now revert you back to your original species when it ends, instead of staying a zombie.

Remaining a zombie after the necromancy ends is a bit silly.

# Changelog

:cl:  
tweak: hecata clan necromancy reverts you back to your original species after it ends instead of staying a zombie
/:cl:

### DIFF
--- a/code/modules/antagonists/bloodsuckers/powers/targeted/hecata.dm
+++ b/code/modules/antagonists/bloodsuckers/powers/targeted/hecata.dm
@@ -58,7 +58,7 @@
 
 /datum/action/cooldown/bloodsucker/targeted/hecata/necromancy/FireTargetedPower(atom/target_atom)
 	. = ..()
-	var/mob/living/target = target_atom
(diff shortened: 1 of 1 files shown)


iloveloopers:
erm, ided???

MajManatee:
Sort of makes sense but at the same time idk. We have green slimes and stuff for a reason.

iamthesilly:
> Sort of makes sense but at the same time idk. We have green slimes and stuff for a reason.

Actual zombies go back to the species they were before when cured, I don't see why these should be any different.

JohnFulpWillard:
it would be nice if this was expanded on instead of removed.
Ideas could be literally anything, ranging from curator/chaplain interaction, to synthflesh from medical

Moltijoe (approved):
might be better to have more interaction reverting species, but this is better than being left as a shitty zombie

PR #20210, opened 2023-09-02
No longer lets preternis liquidator inject infinite sodium thiopental with no combo

Moltijoe:
It never cleared the streak so it'd always just inject sodium thio if you kept grabbing

:cl:  
bugfix: Preternis Liquidator now needs to actually do the combo
/:cl:


ToasterBiome (approved):
bugfix

PR #20212, opened 2023-09-02
Preterni no longer use digitigrade sprites

Moltijoe:
They never looked quite right and with the rework to digitigrade leg sprites it just looks even weirder

They still can't use shoes and have in-built magboots in preparation for the sprite rework

They're using the augmented leg sprites again rather than the digitigrade version of those sprites

:cl:  
tweak: Preterni no longer use digitigrade sprites
/:cl:


iloveloopers:
bring this back immediately

PR #20222, opened 2023-09-03
engineers and shaft miners actually get rnd access on skeleton crew

ktlwjec0:
the changes dont work because rnd console is locked to toxins access, not research. switching it to research means jobs like janitors can now research. so i made a new access type for the console and gave it to rd and scientist for base access. Gave it to robo, gene, engi and miner for skeleton access.

Access will work as intended. gene is sci/med so to me it makes sense that they have it on skeleton crew, and roboticist since they have their own console but should be able to use the main one if there's no one in sci.

# Wiki Documentation

yeah

:cl:  ktlwjec
rscadd: Roboticist and geneticist get rnd console access on skeleton crew.
bugfix: Engineers and shaft miners can use the rnd console on skeleton crew.
/:cl:


N3D6:
Dope

ToasterBiome (approved):

PR #20226, opened 2023-09-03
Makes Holy Light Holy Water better to use

Moltijoe:

- It reduces the annoyance that holy water brings, hopefully making regular people not care nearly as much about drinking holy water.
- It doesn't bring back the issue of chaplains just walking around slapping people with a bible as they'll need to give people holy water first for favour. As we know, getting an injured assistant to sit still long enough to drink a holy water isn't exactly an easy task.
- It reduces the on-demand healing of the bible without making it anti-synergistic to the sect.
- It increases the favour gain from holy water, as favour gain is based on amount healed.
- It will be a useful chemical for doctors to use as it'll allow them to heal patients faster.

:cl:  
tweak: Holy water gives far less jitter (gives a lot more after 25 ticks)
tweak: Holy Light water now gives a 30% healing boost (75% in chapel) rather than 20% (40% in chapel)
tweak: Holy Light bible no longer costs favour, but heals 20 instead of 40
/:cl:

### DIFF
--- a/code/datums/components/heal_react.dm
+++ b/code/datums/components/heal_react.dm
@@ -19,7 +19,7 @@
 
 /datum/component/heal_react/boost
 	///multiplicitive boost to incoming healing
-	var/boost_amount = 0.5
+	var/boost_amount = 0.3
 	///types of damage this will effect
 	var/list/applies_to = list(BRUTE,BURN,TOX,OXY,CLONE,STAMINA)
 	///internal check for if we are being healed by ourselves, no double dipping
--- a/code/datums/components/heal_react.dm
+++ b/code/datums/components/heal_react.dm
@@ -19,7 +19,7 @@
 
 /datum/component/heal_react/boost
 	///multiplicitive boost to incoming healing
-	var/boost_amount = 0.2
+	var/boost_amount = 0.5
 	///types of damage this will effect
 	var/list/applies_to = list(BRUTE,BURN,TOX,OXY,CLONE,STAMINA)
 	///internal check for if we are being healed by ourselves, no double dipping
--- a/code/modules/religion/religion_sects.dm
+++ b/code/modules/religion/religion_sects.dm
@@ -445,11 +445,9 @@
 		return FALSE
 
 	var/mob/living/carbon/human/H = L
-	var/heal_amt = 40 //double healing, no chance to mess up, and shorter cooldown than default
+	var/heal_amt = 20 //no chance to mess up
 
 	if(H.getBruteLoss() > 0 || H.getFireLoss() > 0)
-		var/amount_healed = (heal_amt * 2) + min(H.getBruteLoss() - heal_amt, 0) + min(H.getFireLoss() - heal_amt, 0)
-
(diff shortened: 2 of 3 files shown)


TaelynTheLost:
This is still not good.  It's Holy LIght, not Holy Water.  Why is it always about water?  Besides, on a "normal" shift, I will only be able to get to a rod of white near the end, much less the final ritual.  You're talking about making it even HARDER to get a favor.  This nerf was never needed to begin with.

code/modules/religion/religion_sects.dm
```diff
@@ -445,25 +445,13 @@
 		return FALSE
 
 	var/mob/living/carbon/human/H = L
-	var/heal_amt = 40 //double healing, no chance to mess up, and shorter cooldown than default
-	var/heal_cost = 40
+	var/heal_amt = 20 //no chance to mess up
```
SomeguyManperson, line 448: The purpose of the change & cost here is to prevent bible bashing from being a competitive favor farm compared to any form of cooperative healing, and to dissuade throwing around bible heals. The cost can likely be significantly reduced, but it shouldnt be capable of producing much favor with the default healing bonus value.

TaelynTheLost:
If you want to give med more to do, then add more stuff to do.  More conditions, injuries, illnesses, things other than just damage.  Things that doctors are specifically needed to treat.

If you want people to do other sects besides just Holy Light every time, then make those not suck!  Most of the sects are pointless.  Holy Light is in a good spot.

One or two people complained.  That's not a good reason to ruin something that is fun and unique.  I have some of the highest playtime as Chaplain on the server.  Few people know it better than me.  And Hisa was the one who commissioned Holy Light.  Their opinion should matter a ton.  This is just frustrating.  Holy Light affects doctors a fraction of what chem does.  Only like two people have even complained about it, and these silly changes turn the chaplain into a roll that just does some round start stuff then just sits and does nothing.  This is just bad in so many ways.

code/datums/components/heal_react.dm
```diff
@@ -19,7 +19,7 @@
 
 /datum/component/heal_react/boost
 	///multiplicitive boost to incoming healing
-	var/boost_amount = 0.2
+	var/boost_amount = 0.5
```
SomeguyManperson: This does effect _all_ incoming healing, I feel the number should stay low-ish and the rate of favor gain should be handled moreso by increasing the gain per point healed to reduce the chance this massively effects a combat healing strat

code/modules/religion/religion_sects.dm
Moltijoe: Bible bashing isn't a competitive favor farm since no one lets you give them holy water first, no reason to make the bible literally worse than other sect options

code/datums/components/heal_react.dm
Moltijoe: it's been reduced by balance council suggestions

Marmio64:
> and these silly changes turn the chaplain into a roll that just does some round start stuff then just sits and does nothing. This is just bad in so many ways.

What do you think chaplain players did before sects were ever a thing? There's more to chaplain than picking the healer sect and being a medical doctor replacement.

PR #20228, opened 2023-09-04
Other idea holy light

Moltijoe:

- Holy Light relying on holy water to generate favour removes the agency from the chaplain, as they're either forced to inject people with holy water against their will, or give the holy water to medical KNOWING they're never going to use it.
- This still forces them to rely on other sources of healing as the bible has very little healing on it's own, even when accounting for the heal boost.
- It being a direct action means medical can let them into medbay to actively help, without just fully healing the injured people, taking their job away from them
- It being tied to a chemical means it could very easily be abused by some people that have a penchant for taking lots of various chemicals to become effectively immortal via chem healing, this can't happen when it's tied solely to the bible and unreliable medbots.
- Holy Light is focused on Holy Light rather than Holy Water

:cl:  
rscdel: Holy Light water no longer applies a heal boost
rscadd: Holy Light bible now applies a 30 second duration heal boost
tweak: Holy Light bible now heals for 10 instead of 40 with a 10 second cooldown instead of 12 and has no favor cost
tweak: Holy Light heal boost now increases by 100% (200% in chapel) instead of 20% (40% in chapel)
tweak: Holy Light medbot applies the heal boost status effect instead of injecting holy water
/:cl:

### DIFF
--- a/code/modules/religion/religion_sects.dm
+++ b/code/modules/religion/religion_sects.dm
@@ -418,7 +418,7 @@
 /datum/religion_sect/holylight
 	name = "Holy Light"
 	desc = "A sect dedicated to healing."
-	convert_opener = "Welcome to the Holy Light, disciple. <br>Your bible will now bless people. Blessed people will receive increased healing, and generate favor when they do."
+	convert_opener = "Welcome to the Holy Light, disciple. <br>Healing people with your bible will increase further healing given to them for a short time, providing favor based on the amount healed."
 	alignment = ALIGNMENT_GOOD // literally the only good sect besides default lol
 	rites_list = list(/datum/religion_rites/medibot, /datum/religion_rites/holysight, /datum/religion_rites/healrod, /datum/religion_rites/holyrevival)
 	altar_icon_state = "convertaltar-heal"
--- a/code/modules/religion/religion_sects.dm
+++ b/code/modules/religion/religion_sects.dm
@@ -418,7 +418,7 @@
 /datum/religion_sect/holylight
 	name = "Holy Light"
 	desc = "A sect dedicated to healing."
-	convert_opener = "Welcome to the Holy Light, disciple. <br>Your holy water will now bless people with improved healing, which provides favor. Additionally, your bible heal is significantly stronger, at the cost of favor and an increased cooldown."
+	convert_opener = "Welcome to the Holy Light, disciple. <br>Your bible will now bless people with improved healing, which provides favor."
 	alignment = ALIGNMENT_GOOD // literally the only good sect besides default lol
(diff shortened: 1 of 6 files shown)


TaelynTheLost:
Better, except for the fact that now the healing sect heals less than every other sect including puritism.  "Hitting somebody in the head with your bible will heal 20 points of brute and burn damage" - wiki

Moltijoe:
> Better, except for the fact that now the healing sect heals less than every other sect including puritism.

it already healed less than capitalism
the difference is, this will allow you to heal more using other sources, such as rod or medical

TaelynTheLost:
> > Better, except for the fact that now the healing sect heals less than every other sect including puritism.
> 
> it already healed less than capitalism the difference is, this will allow you to heal more using other sources, such as rod or medical

On a normal round it can take most of the shift to reach Rod levels of favor, and this reduces the healing, and  thus, favor gain, by 75%.  Basically unless you get someone to just sit in chapel and farm healing with you, or it's a round with a lot of injuries, it's not happening.

Moltijoe:
> On a normal round it can take most of the shift to reach Rod levels of favor, and this reduces the healing, and thus, favor gain, by 75%. Basically unless you get someone to just sit in chapel and farm healing with you, or it's a round with a lot of injuries, it's not happening.

this would be merged with the 2x favor gen pr theos opened
also, you can combine this buff with other sources of healing to generate more favor
don't worry too much about exact numbers for now as that can be changed

Hislyna:
Its better but I still see no need for any change to something that was fine

Hislyna:
With that heal boost, making the healing 20, im willing to see how it works. Im still skeptical on the point gen, but I will be okay with this change, even if I believe nothing was wrong with the original

ToasterBiome:
Basically a revert

code/modules/religion/religion_sects.dm
```diff
@@ -422,7 +418,7 @@
 /datum/religion_sect/holylight
 	name = "Holy Light"
 	desc = "A sect dedicated to healing."
-	convert_opener = "Welcome to the Holy Light, disciple. <br>Your holy water will now bless people with improved healing, which provides favor. Additionally, your bible heal is significantly stronger, at the cost of favor and an increased cooldown."
+	convert_opener = "Welcome to the Holy Light, disciple. <br>Your bible will now bless people. Blessed people will receive increased healing, and generate favor when they do."
```
SomeguyManperson:
```suggestion
	convert_opener = "Welcome to the Holy Light, disciple. <br>Healing people with your bible will increase further healing given to them for a short time, providing favor based on the amount healed."
```

Moltijoe: that's a lot less clunky

SomeguyManperson:
better than the current situation

ToasterBiome (approved):
ok lets try it

PR #20235, opened 2023-09-04
Mediborg hypo regen bugfix

Moltijoe:
The comment references seconds, it uses deciseconds, this means it refills 5u every second, effectively infinite chems

:cl:  
bugfix: Borg reagent dispensers regen in seconds rather than deciseconds
bugfix: Borg reagent dispensers regen less affected by lag
/:cl:


Runian:
do same for syndie hypo and the two service borgshakers

ToasterBiome (approved):

PR #20240, opened 2023-09-05
Cultist shield robe nerf

Moltijoe:
Combined with the mirror shield, this makes a cultist WAY too durable

:cl:  
tweak: Cultist shield robe has 1 shield charge, but regens after 45 seconds of not getting hit
/:cl:

### DIFF
--- a/code/modules/antagonists/cult/cult_items.dm
+++ b/code/modules/antagonists/cult/cult_items.dm
@@ -440,7 +440,7 @@
 
 /obj/item/clothing/suit/hooded/cultrobes/cult_shield/proc/reshield()
 	shielded = TRUE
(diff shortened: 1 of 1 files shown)


RG4ORDR:
Confer with the council but I liked what you said about it in discord.

1 charge but it'll regen.

MajManatee:
I feel as though it should regen with two. It not doing so seems like it was either an oversight or just poor planning

MajManatee:
45 seconds with one charge is an AWFULY long time

Moltijoe:
> 45 seconds with one charge is an AWFULY long time

make another PR after if you think that way

ToasterBiome (approved):

PR #20241, opened 2023-09-05
Cultist spear size nerf

Moltijoe:
The spear is a two handed weapon that's just really quite good, them being restricted in use is supposed to be the downside, but it means nothing when you can make multiple and store them

:cl:  
tweak: Blood cult spear is huge instead of normal size
/:cl:


RG4ORDR:
Lol I'm for this because it's a secret cult strat.

Having 10 spears in your backpack to stun, and deal like 35 damage to and just being able to CHUCK ANOTHER ONE because you spawned more of them is stupid.

MajManatee:
Nerfing an already rare item would make it used even less. Possible to give it some incentive to being used?

Moltijoe:
> Nerfing an already rare item would make it used even less. Possible to give it some incentive to being used?

the incentive is that it's fucking ridiculous in terms of stats

MajManatee:
16 wielded is definitely not great. Even if the throw force is as high as it is, its still a really obscure item that id bet 95% of players don't know exist or how to get.

Moltijoe:
> 16 wielded is definitely not great. Even if the throw force is as high as it is, its still a really obscure item that id bet 95% of players don't know exist or how to get.

it's 28 wielded, it's additive

MajManatee:
ok well thats good then

ToasterBiome (approved):

PR #20254, opened 2023-09-06
sliding (click dragging) my way (plasma) into her dms (portable generator)

MajManatee:
Allows click dragging of fuel into generators so a borg can power generators.
as you can see this is very unique code that I didn't just steal from molti

Generators are used in two situations. One, Everything is fucked and we need power now (so it would improve QOL for everyone affected) and two, we are being silly and doing shenanigans (Same reason, allowing people to be silly is a good thing).

There's no reason to lock QOL stuff arbitrarily.

# Changelog

:cl:   
tweak: generator fuel can be dragged into the generator now.
/:cl:

### DIFF
--- a/code/modules/power/port_gen.dm
+++ b/code/modules/power/port_gen.dm
@@ -171,7 +171,7 @@
 	explosion(src.loc, 2, 5, 2, -1)
 
 /obj/machinery/power/port_gen/pacman/MouseDrop_T(atom/dropping, mob/user)
-	if(is_type_in_typecache(dropping, sheet_path))
+	if(istype(dropping, sheet_path))
 		attackby(dropping, user)
 	else
 		..()
--- a/code/modules/power/port_gen.dm
+++ b/code/modules/power/port_gen.dm
@@ -170,6 +170,12 @@
 /obj/machinery/power/port_gen/pacman/proc/overheat()
 	explosion(src.loc, 2, 5, 2, -1)
 
+/obj/machinery/power/port_gen/pacman/MouseDrop_T(atom/dropping, mob/user)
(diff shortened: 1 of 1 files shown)


code/modules/power/port_gen.dm
```diff
@@ -170,6 +170,12 @@
 /obj/machinery/power/port_gen/pacman/proc/overheat()
 	explosion(src.loc, 2, 5, 2, -1)
 
+/obj/machinery/power/port_gen/pacman/MouseDrop_T(atom/dropping, mob/user)
+	if(is_type_in_typecache(dropping, sheet_path))
```
iamthesilly: You should be using istype() here since sheet_path is a typepath and not a list or cache.

SapphicOverload (changes requested):

SapphicOverload (approved):
Neat

code/modules/power/port_gen.dm
```diff
@@ -170,6 +170,12 @@
 /obj/machinery/power/port_gen/pacman/proc/overheat()
 	explosion(src.loc, 2, 5, 2, -1)
 
+/obj/machinery/power/port_gen/pacman/MouseDrop_T(atom/dropping, mob/user)
+	if(istype(dropping, sheet_path))
```
iamthesilly: You forgor to add an Adjacent() check

SapphicOverload (changes requested):

code/modules/power/port_gen.dm
MajManatee: jesus make up your mind woman

SapphicOverload (approved):

PR #20267, opened 2023-09-08
Not needing to breathe no longer prevents CPR

Moltijoe:
Randomly getting a "buff" and having it screw over your ability to do medical is annoying.
also, it doesn't make sense really, so long as you have lungs, you should be able to perform cpr

Also removes redundant lungs and breathing checks

:cl:  
tweak: CPR only requires lungs, not needing to breathe
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -650,10 +650,6 @@
 			to_chat(src, span_warning("You have no lungs to breathe with, so you cannot perform CPR!"))
 			return FALSE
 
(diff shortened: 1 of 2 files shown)


iamthesilly:
Makes sense, if you have lungs you can move air, even if you don't need them to breathe

MajManatee:
If only there was a way to give someone air without needing to breathe... oh wait... 
I cant upload the image apparently but its a big bag they use in CPR. Google it.

ToasterBiome (approved):

ToasterBiome (approved):

PR #20271, opened 2023-09-08
You no longer get mining speed on IceMeta

azzzertyy:
Removes the mining speed from station areas, you still get it when youre outside of the station on an icemoon area but will no longer get it on the station.

Currently miner items still perform exactly the same as before, this doesn't change any of that, if someone wants that changed you can use some of the code here to figure it out, but I'm not bothered by that really.

Keeps the speed in line with other maps, was annoying for players like myself, I know I personally have discussed this issue with Ned innumerable times and we both agree that this is the best solution to the problem.

I feel this speed in general is better for our server, and having a map that meaninglessly decides to disobey that was silly.

# Changelog
:cl:  
tweak: No more superspeed on icemeta
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/yogstation/code/modules/mob/living/carbon/human/human_movement.dm
@@ -10,10 +10,9 @@
 	if(!player_turf || ! is_mining_level(player_turf.z)) // If we can't get the turf, assume it's not on mining.
 		return ..()
 
-	if(is_mining_level(player_turf.z)) // for icemoon 
-		var/area/player_area = get_area(src)
-		if(!player_area.mining_speed)
-			return ..()
+	var/area/player_area = get_area(src)
+	if(!player_area.mining_speed)
+		return ..()
 
 	var/mod = 0
 	if(m_intent == MOVE_INTENT_WALK)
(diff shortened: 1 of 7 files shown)


azzzertyy:
I've tested it a bit, but there may be some issues with areas that use area/ruin that are not on lavaland or icemoon.
If someone comes across a case like this its an easy fix, but let me know and I'll do it.
Since the way this works is that each area is assigned a Mining_Speed variable, it should get ignored in most cases since it does check if its on a mining level first, but im not 100% sure on the implementation

yogstation/code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -5,10 +5,16 @@
...
 
 /mob/living/carbon/human/update_move_intent_slowdown()
-	var/turf/T = get_turf(src)
-	if(!T || !is_mining_level(T.z)) // If we can't get the turf, assume it's not on mining.
+	var/turf/player_turf = get_turf(src)
+
+	if(!player_turf || ! is_mining_level(player_turf.z)) // If we can't get the turf, assume it's not on mining.
 		return ..()
 
+	if(is_mining_level(player_turf.z)) // for icemoon
```
iamthesilly: You shouldn't need this second check, since it already returns if it isn't on a mining level

azzzertyy:
wtf why did it start a review

ToasterBiome:
good good. this is what we wanted but n3 didn’t know how to do it

Yarinoi:
but...

superspeed.....

N3D6:
Thanks, ive never been against this solution, I just haven’t known how to implement it. Bless you

ToasterBiome (approved):

PR #20290, opened 2023-09-10
Make geranium seed useful and tweaks

warface1234455:
Geranium seed can now be mutated into Bee balm
Bee balm can no longer be mutated into Geranium

Currently geranium seed is very useless in botany because it has nothing, no trait, no chem, no mutation. And the fact that bee balm can be mutated back to geranium can be frustrating for botanist so this should make it less suffering

# Wiki Documentation
Guide to plants document changes

# Changelog

:cl:  
tweak: Geranium seed can now be mutated into Bee balm
tweak: Bee balm can no longer be mutated into Geranium
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/flowers.dm
+++ b/code/modules/hydroponics/grown/flowers.dm
@@ -269,7 +269,7 @@
 	growing_icon = 'icons/obj/hydroponics/growing_flowers.dmi'
 	icon_grow = "bee_balm-grow"
 	icon_dead = "bee_balm-dead"
-	mutatelist = list(/obj/item/seeds/bee_balm/honey) //Lower odds of becoming honey
+	mutatelist = list(/obj/item/seeds/bee_balm/honey)
 	reagents_add = list(/datum/reagent/medicine/spaceacillin = 0.1, /datum/reagent/space_cleaner/sterilizine = 0.05)
 	rarity = 20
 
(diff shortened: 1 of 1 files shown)


solwashere:
Did you test it?

ShoziOffical:
> Did you test it?

no but it should work since that the only var that make mutation species changes

SomeguyManperson:
Shouldn't require testing

I hate how this "mutation chain but it can backfire at every level and the maturation rate of each plant is in the double digits" is the only effective way to make something difficult to achieve with botany

ShoziOffical:
> I hate how this "mutation chain but it can backfire at every level and the maturation rate of each plant is in the double digits" is the only effective way to make something difficult to achieve with botany

its all about rngs but the thing i hate is to get honey balm you need to mutate poppy -> lily (50% chance to be geranium that does nothing and cannot be mutated back) -> bee balm (also has 50% chance to be space trumpet) -> honey balm (AND also has 50% can become geranium again which isnt listed on wiki). The last mutation chain is really bad, this makes botanist has to conserve more seeds incase bad rngs

VaelophisNyx:
> Shouldn't require testing
> 
> I hate how this "mutation chain but it can backfire at every level and the maturation rate of each plant is in the double digits" is the only effective way to make something difficult to achieve with botany

concept: what if mutations could species-jump on a low percent

trying for holy melons? hope you enjoy those oranges you have now

SomeguyManperson:
This is all, of course, a mutation chain built entirely to avoid purchasing bees from cargo

ShoziOffical:
> This is all, of course, a mutation chain built entirely to avoid purchasing bees from cargo

i or people in general wanted honey trait, not the honey reagent itself

ShoziOffical:
and yes if i wanted a ton of honey i would just get a bee farm rather than getting honey balm

ToasterBiome (approved):

PR #20306, opened 2023-09-11
Nerf to preternis eating

Moltijoe:
They get to sidestep the chef for electricity, but they don't actually have any real power-drain as far as the station is concerned
It also lets them use a single battery as a snack for the entire shift with no worry of ever running out
This should make them actually care about the station's power status

:cl:  
tweak: Preterni consume more power and take slightly longer to fully charge
/:cl:


VaelophisNyx:
was always strange watching them use a single battery all shift

ToasterBiome (approved):

PR #20330, opened 2023-09-13
Tweaks accessibility flags for mouth and eyes

Moltijoe:
Wearing a hat doesn't make the eyes or mouth inaccessible

The main case this should matter is in surgery, where simple hats won't block surgery on mouth or eyes

:cl:  
tweak: Tweaks accessibility flags for mouth and eyes
/:cl:

### DIFF
--- a/code/modules/surgery/helpers.dm
+++ b/code/modules/surgery/helpers.dm
@@ -145,10 +145,10 @@
 			if(covered_locations & HEAD)
 				return 0
 		if(BODY_ZONE_PRECISE_EYES)
-			if(covered_locations & HEAD || face_covered & HIDEEYES || eyesmouth_covered & GLASSESCOVERSEYES)
+			if(face_covered & HIDEEYES || eyesmouth_covered & GLASSESCOVERSEYES)
 				return 0
 		if(BODY_ZONE_PRECISE_MOUTH)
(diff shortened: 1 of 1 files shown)


code/modules/surgery/helpers.dm
```diff
@@ -145,10 +145,10 @@
 			if(covered_locations & HEAD)
 				return 0
 		if(BODY_ZONE_PRECISE_EYES)
-			if(covered_locations & HEAD || face_covered & HIDEEYES || eyesmouth_covered & GLASSESCOVERSEYES)
+			if(face_covered & HIDEFACE || face_covered & HIDEEYES || eyesmouth_covered & GLASSESCOVERSEYES || eyesmouth_covered & HEADCOVERSEYES)
```
iamthesilly: Should probably replace HIDEFACE check with a MASKCOVERSEYES check since not all masks with HIDEFACE actually cover your eyes (sterile mask, fake moustache, etc)

SapphicOverload (changes requested):

code/modules/surgery/helpers.dm
Moltijoe: HIDEFACE hides identity, meaning meaning it probably hides the eyes too, if the mask doesn't hide the eye, it probably shouldn't HIDEFACE.
meaning it's an issue with the masks, not this code

MajManatee:
thank god i no longer get depression when getting surgery

SapphicOverload (approved):

PR #20331, opened 2023-09-13
Deletes Pacifism touch

Moltijoe:
It's a noob trap spell that is either round removal in a touch (worse than even old disintegrate)
or once that becomes easily curable, utterly useless for how long of a cooldown it has

It getting removed by bug was the key time to keep it out, but instead it was brought back for posterity sake.
now that 48 hours have passed, it's getting removed again

:cl:  
rscdel: Pacifism touch
/:cl:


iloveloopers:
Spelling mistake on the changelog. Resign as mantainer immediately

RG4ORDR:
I can't really support the removal because we have plenty of spells that have extremely niche or limited usage.
This has several hard counters, while also being quite strong in the special application it has - much like old-school disintegrate. 
Also Polymorph wand and staff exists, a wizard can buy this and then fireball, or another offensive spell and get lucky on a random roll for someone to turn into a drone,mouse,slime, or some other harmless mob.

Least with this if you know you been touched you have time to counter it. Otherwise it is effectively your fault for getting gondola'd. And the applauding the wizard for preventing you getting cured.

ToasterBiome (approved):

PR #20334, opened 2023-09-14
cyborg and cyborg monitor program tweaks

warface1234455:
- Give cyborg tablet access to the cyborg monitor program.
- Fix the issue of the log file not being downloaded in the cyborg monitor program when using an ID with cyborg access in the card slot. The problem occurs because the ID unlocks the cyborg cover instead of initiating the download. I changed it so that the help intent is now used for unlocking the cover. If you have an ID with cyborg access in the card slot, you need to use intents other than help to download the log file.

Granting cyborgs access to the cyborg monitor program enables them to track cyborgs from any location as needed, instead of having to download a cyborg monitor program in a nearby computer

# Testing
Local tested, no runtimes, everything works as intended

# Wiki Documentation
program and cyborg document of changes?

# Changelog

:cl:  
tweak: Give cyborg tablet access to the cyborg monitor program.
bugfix: Fix the issue of the log file not being downloaded in the cyborg monitor program.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -539,7 +539,7 @@
 		else
 			to_chat(user, span_warning("Unable to locate a radio!"))
 
-	else if(W.GetID())			// trying to unlock the interface with an ID card
+	else if(W.GetID() && user.a_intent == INTENT_HELP)			// trying to unlock the interface with an ID card only on help intent.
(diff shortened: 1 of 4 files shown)


ShoziOffical:
@Moltijoe why does this need balance council? they can already download it but just not on cyborg tablet

MajManatee (approved):
I see minimal issues with this

PR #20345, opened 2023-09-16
Research access restrictions can be hacked

ToasterBiome:

A while we locked research access behind science access. this is good because we don't want HOS doing science's job and researching all their stuff. Later, we added an emagged option for traitors. This adds a third option for antags that aren't traitors.

You can deconstruct the machine and use a screwdriver on the board to jailbreak it (kinda like the contraband for cargo console). Welding it will restore the access. This adds another way for people to approach this besides an emag, and it's sufficiently shifty that it's technically sabotage. Also makes this a WORSE option than emagging, as you're still messing with it publicly but this takes 20+ seconds to do instead of the instant an emag takes.

**This is only for production type boards, so it's specifically the one in science, not the one in engineering (break into science)**

Type changes are now in the multitool act.

Stopping research kneecaps lowpop/no-science rounds way harder than HOS being a bitch has. Fix both problems with one stone

# Testing
I gotta

# Wiki Documentation

Needs to be added the ways the access can change

# Changelog

:cl:  ToasterBiome, Someguymanperson
tweak: Research circuit boards in R&D can be "hacked" with a screwdriver to unlock access.
/:cl:

### DIFF
--- a/code/game/objects/items/circuitboards/computer_circuitboards.dm
+++ b/code/game/objects/items/circuitboards/computer_circuitboards.dm
@@ -362,7 +362,7 @@
 		if(build_path == /obj/item/circuitboard/computer/rdconsole/production)
 			to_chat(user, span_danger("[src] sparks! That isn't right."))
 			var/datum/effect_system/spark_spread/p = new /datum/effect_system/spark_spread
-			p.set_up(12, 1, user)
+			p.set_up(6, 1, user)
 			p.start()
 		if(build_path == /obj/machinery/computer/rdconsole/core)
 			name = "R&D Console - Robotics (Computer Board)"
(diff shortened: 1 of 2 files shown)


code/game/objects/items/circuitboards/computer_circuitboards.dm
```diff
@@ -353,12 +354,16 @@
...
-
-
-/obj/item/circuitboard/computer/rdconsole/attackby(obj/item/I, mob/user, params)
-	if(I.tool_behaviour == TOOL_SCREWDRIVER)
-		if(build_path == /obj/machinery/computer/rdconsole/production)
-			return
+	
+/obj/item/circuitboard/computer/rdconsole/multitool_act(mob/living/user, obj/item/I)
+	. = ..()
+	visible_message("[user] fiddles with [src].", "You fiddle with [src].")
```
tattax: /obj/item/circuitboard/computer/rdconsole when it fiddles with src
```suggestion
	user.visible_message(span_notice("[user] fiddles with [src]."), span_notice( "You fiddle with [src]."))
```

wonderinghost:
Can't wait to do it as a roboticists. Who can still research but only on their terminal

neeshacark:
this is also ideal for when rnd is all dead or otherwise not doing their job or rnd doesnt exist

ju45he:
Am I able to print a board and run off with it into maints?

MajManatee:
smh HD doesnt even fill out his own PR

Moltijoe (approved):

PR #20353, opened 2023-09-16
Sinful demon form has it's original 200 health back

Marmio64:

Sinful demon form now has 200 health once more rather than 160 health. Due to how regen per hit is implemented, this also means they heal 10 health from hitting someone rather than 8.

Yes, this is tested.

Now that demons are PTE and can be killed whenever one so pleases, they could use a better defense buffer.

# Testing

Yes, this is tested, it is only a very simplistic numbers change.

# Wiki Documentation

Sinful demon form now has 200 health once more rather than 160 health. Due to how regen per hit is implemented, this also means they heal 10 health from hitting someone rather than 8.

# Changelog

:cl:  
tweak: Sinful demon form now has 200 health rather than 160, and heals 10 hp per hit rather than 8 as a result.
/:cl:

### DIFF
--- a/code/modules/antagonists/demon/general_powers.dm
+++ b/code/modules/antagonists/demon/general_powers.dm
@@ -38,8 +38,8 @@
 	maxbodytemp = INFINITY
 	faction = list("hell")
 	attacktext = "wildly tears into"
-	maxHealth = 160
-	health = 160
+	maxHealth = 200
+	health = 200
 	environment_smash = ENVIRONMENT_SMASH_STRUCTURES
 	obj_damage = 40
 	melee_damage_lower = 20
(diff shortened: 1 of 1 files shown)


Therandomhoboo:
Should give them a boost in armour against damage too, considering their demon form forces them to be slow, no ranged & does DOT damage. (which you should remove too)

Marmio64:
> Should give them a boost in armour against damage too, considering their demon form forces them to be slow, no ranged & does DOT damage. (which you should remove too)

they already get full crit, stun, and fire immunity thanks to being a simplemob. The damage over time should also stay because demon form is not something you're supposed to stay in, going around on long winded killing sprees. A speed buff or something else could be implemented in a separate PR later if demon form is still too weak.

KingSuperBOB:
> > Should give them a boost in armour against damage too, considering their demon form forces them to be slow, no ranged & does DOT damage. (which you should remove too)
> 
> they already get full crit, stun, and fire immunity thanks to being a simplemob. The damage over time should also stay because demon form is not something you're supposed to stay in, going around on long winded killing sprees. A speed buff or something else could be implemented in a separate PR later if demon form is still too weak.

To be fair old demons were stronger as the DOT damage didnt pass over to the human form but other damage did, now the DOT damage does pass over to human form

Therandomhoboo:
> > Should give them a boost in armour against damage too, considering their demon form forces them to be slow, no ranged & does DOT damage. (which you should remove too)
> 
> they already get full crit, stun, and fire immunity thanks to being a simplemob. The damage over time should also stay because demon form is not something you're supposed to stay in, going around on long winded killing sprees. A speed buff or something else could be implemented in a separate PR later if demon form is still too weak.

Understandable, just the fact they are PTE means that they will be hunted down at first sight which is why I think DOT damage should go, however I can see players abusing this and staying in it so fair. Speed is the issue though, which hopefully will be rectified in a later PR by anyone.

ToasterBiome (approved):

PR #20360, opened 2023-09-17
Icemoon: Seed Vault

neeshacark:

seed vault for icemoon, ripped from lavaland.

more ruin for ice

# Wiki Documentation

ruin page documentation

# Changelog

:cl:  cark
mapping: icemoon seed vault is now added
/:cl:


N3D6:
add it to the blacklist people haven’t been doing that for some reason

MajManatee:
no picture?

Aquizit (changes requested):
image

Although I don't think it *needs* one I'd feel better with an outlet here instead of just the open pipe.

Also I would say put the area on the engines too instead of the mining one just as they are part of the building more or less.

Aquizit (approved):

PR #20389, opened 2023-09-21
Nerfs lichdom holoparasite combo

Moltijoe:
Undying holoparasite that deletes the corpse of a dead lich
Makes it impossible to find the phylactery
Removes the downside of holoparasite

rather than removing the interesting combo, makes it less meta

:cl:  
tweak: Holoparasites no longer dust liches upon death
tweak: Holoparasites lose 4 random stats upon revival by a lich
/:cl:

### DIFF
--- a/yogstation/code/modules/guardian/guardianstats.dm
+++ b/yogstation/code/modules/guardian/guardianstats.dm
@@ -75,3 +75,33 @@
 		var/datum/guardian_ability/minor/minor_ability = A
 		stats_info += minor_ability.name
 	return stats_info.Join(", ")
+
+/datum/guardian_stats/proc/weaken(times = 1)//reduces a random stat x times
+	for(var/i = 0; i < times; i++)
+		var/list/reduction = list("damage", "defense", "speed", "potential", "range")
+			shuffle(reduction)
+		var/found = FALSE
+		for(var/option in reduction)
+			switch(option)
+				if("damage")
+					if(damage > 1)
+						damage --
+						found = TRUE
+				if("defense")
+					if(defense > 1)
+						defense --
+						found = TRUE
+				if("speed")
+					if(speed > 1)
+						speed --
+						found = TRUE
+				if("potential")
+					if(potential > 1)
+						potential --
+						found = TRUE
+				if("range")
+					if(range > 1)
+						range --
+						found = TRUE
+			if(found)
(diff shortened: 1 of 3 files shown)


Therandomhoboo:
Agree with the PR, think losing 4 random stats is a bit too much though.

RG4ORDR:
> Agree with the PR, think losing 4 random stats is a bit too much though.

Holoparas giving a very strong antag a buddy that can shoot funny crystal bolts or save their lives is extremely strong.

Therandomhoboo:
> > Agree with the PR, think losing 4 random stats is a bit too much though.
> 
> Holoparas giving a very strong antag a buddy that can shoot funny crystal bolts or save their lives is extremely strong.

iirc the holoparasite spell isn't always ranged and have different stats & powers cause you can flick through 'em. Unless there are two spells for them. Either way, one death could neuter the holoparasite completely depending on rng for the stats provided.

Don't get me wrong, I agree with the PR but I think debuffing 4 stats randomly is too much. I would suggest 2. If not enough then make it that each subsequent death after the first one increases the amount of stats lost by an additional 1? If it isn't too much of a hassle to code.

MajManatee:
I think the wording was poor, but this seems great,

yogstation/code/modules/guardian/guardianstats.dm
```diff
@@ -75,3 +75,33 @@
 		var/datum/guardian_ability/minor/minor_ability = A
 		stats_info += minor_ability.name
 	return stats_info.Join(", ")
+
+/datum/guardian_stats/proc/weaken(times = 1)//reduces a random stat x times
+	for(var/i = 0; i < times; i++)
+		var/list/reduction = list("damage", "defense", "speed", "potential", "range")
+		shuffle(reduction)
+		var/found = FALSE
+		for(var/option in reduction)
```
SomeguyManperson, line 84: suggest some failsafe in case all the stats end up at 1 and this loops

Moltijoe: It should stop when it goes through all options, shouldnt it?

SomeguyManperson: i mifhr vw crazy

ToasterBiome (approved):

PR #14710, opened 2022-07-05
craftable epi and atropine medipens

ynot01:
# Document the changes in your pull request

Kind of pissed that the only way to get atropine medipens is to purchase a 600 credit (economy amirite?) medpack that has a bunch of synthflesh and other crap I don't really care about

Also makes epipens more renewable, obviously

More attractive to medical staff than a bottle that sits idle in a fridge, so it's healthy for player interaction!

Also fixed some missing time defines in the same file

# Wiki Documentation

Epinephrine Medipen
Tools category
Requires screwdriver
Takes 2 seconds
Recipe:
1 Pen
1 Syringe
10u Epinephrine

Atropine Autoinjector
Tools category
Requires screwdriver
Takes 4 seconds
Recipe:
1 Pen
1 Syringe
10u Atropine

# Changelog

:cl:  
rscadd: Added crafting recipes for epinephrine and atropine medipens
/:cl:

PR #20402, opened 2023-09-22
[s]Sanguirite creation exploit fix

Kell-E:

It's no longer possible to draw reagents out of an epipen, making it more difficult to generate infinite sanguirite while still allowing proper use of an epipen.

Patches an exploit that allowed a character to generate infinate sanguirite effortlessly.

# Testing

Tested on a local server to confirm it's no longer possible to draw free sanguirite out of the epipen.

# Changelog

:cl:  
bugfix: It's no longer possible to draw infinite sanguirite from an epipen.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -109,7 +109,6 @@
 	amount_per_transfer_from_this = 12
 	volume = 12
 	ignore_flags = 1 //so you can medipen through hardsuits
-	reagent_flags = DRAWABLE
 	flags_1 = null
 	list_reagents = list(/datum/reagent/medicine/epinephrine = 10, /datum/reagent/medicine/coagulant = 2)
 	custom_price = 40
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -109,6 +109,7 @@
 	amount_per_transfer_from_this = 12
 	volume = 12
 	ignore_flags = 1 //so you can medipen through hardsuits
+	reagent_flags = NONE
 	flags_1 = null
 	list_reagents = list(/datum/reagent/medicine/epinephrine = 10, /datum/reagent/medicine/coagulant = 2)
 	custom_price = 40
--- a/code/datums/components/crafting/recipes.dm
+++ b/code/datums/components/crafting/recipes.dm
@@ -671,7 +671,7 @@
 	time = 2 SECONDS
 	reqs = list(/obj/item/reagent_containers/autoinjector/medipen = 1,
 				/datum/reagent/medicine/epinephrine = 10,
-				/datum/reagent/medicine/Sanguirite = 2)
+				/datum/reagent/medicine/coagulant = 2)
 
 	category = CAT_MEDICAL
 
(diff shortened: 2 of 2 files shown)


ShoziOffical:
This was intended, not a bug take a look on line 638, this was made for people to get normal epipen with sangi easily

ShoziOffical:
#14710 was merged by jamie (HD) means it was balance approved tho there wasn't balance council before

Kell-E:
Oh, weird! It just seemed like an oversight to me given how easily abused it could be. If it's intentional I can close it!

ShoziOffical:
> Oh, weird! It just seemed like an oversight to me given how easily abused it could be. If it's intentional I can close it!

You can change the title, why is this good and the CL to tweak to get balance council approval if you think this is bad

ShoziOffical:
And change the recipe to make epipen too if you intend to change this

ShoziOffical:
> Oh, weird! It just seemed like an oversight to me given how easily abused it could be. If it's intentional I can close it!

Not too sure how thia would be abused that bad.. Out of 10 rounds when im playing doctor only 2 rounds chemist make epipen injectors but not that huge amount i think this is fine

Kell-E:
In theory you can remove the sanguinite after crafting the epipen, creating an infinite supply of sanguirite on demand. 10u of epi goes in, 10u of epi and 2u of sanguirite comes out. That's why I'd presumed it was an exploit, or a bug.

ShoziOffical:
> In theory you can remove the sanguinite after crafting the epipen, creating an infinite supply of sanguirite on demand. 10u of epi goes in, 10u of epi and 2u of sanguirite comes out. That's why I'd presumed it was an exploit, or a bug.

then make it so that you cannot remove the chems from epipens

Kell-E:
I'll take a look at that a little later on then, best of both worlds!

ShoziOffical:
> I'll take a look at that a little later on then, best of both worlds!

you can go to code/modules/reagents/reagent_containers/hypospray.dm then remove the whole line on 112 so that people cannot remove the reagents from syringe

Kell-E:
Crafting reciple was changed back to only use epi, but now the sanguirite can't be drawn. It's still possible to obtain it in theory but it'd be a lot more work by comparison.

Moltijoe (approved):

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -109,7 +109,6 @@
 	amount_per_transfer_from_this = 12
 	volume = 12
 	ignore_flags = 1 //so you can medipen through hardsuits
-	reagent_flags = DRAWABLE
```
Moltijoe, line 112: actually, deleting this would cause even more issues as it'd make it default to using the same flags as regular autoinjectors, meaning people could refill them
instead make reagent_flags = NONE

Moltijoe (changes requested):

code/modules/reagents/reagent_containers/hypospray.dm
ShoziOffical: oh yeah i forgot

ShoziOffical:
you should change the changelog to the code changes

code/modules/reagents/reagent_containers/hypospray.dm
Kell-E: Ammended!

Moltijoe (approved):

PR #20420, opened 2023-09-23
Removes pressure immunity from regular firesuits

Moltijoe:
Requested to make space walking more dangerous without proper gear
too easy to safely spacewalk with oxygen, coffee, and the fire suits that are on every corner

honestly, if i had the time i'd split high pressure and low pressure into two different things and only remove the low pressure protection from fire suits

:cl:  
tweak: Regular firesuits no longer give pressure immunity
/:cl:

### DIFF
--- a/code/modules/clothing/head/hardhat.dm
+++ b/code/modules/clothing/head/hardhat.dm
@@ -61,7 +61,6 @@
 	hat_type = "red"
 	dog_fashion = null
 	name = "firefighter helmet"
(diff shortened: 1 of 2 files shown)


LassiH:
wont this make them just useless in  fires due the overpressure in them

neeshacark:
yeah what lassih said, might be better to put this off until you have to time to do it right

Moltijoe:
> yeah what lassih said, might be better to put this off until you have to time to do it right

aight, fair

Issue #20370, opened 2023-09-18
Nar'sie mirror shield doesn't stun


Laus4Deus:
### Severity

Inconvenience

### Short Description

People can pic up cult item. It shouldn't be this way for it can lead to confusion as to who is cult

### Reproduction

Pick up shield, no stun no nothing

### Round ID

66666

### Was this done on a locally hosted or non-Yogstation hosted server?

None

Oblisk234:
Per #20425 being balance council denied, not an issue.

PR #20425, opened 2023-09-24
Prevents non-cultists from using the blood cult mirror shield.

Kell-E:

In response to #20370

Picking up a blood cult mirror shield as a non-cultist will now cause you to take significant arm damage and be temporarily paralyzed.

Being able to use the mirror shield as a normal member of the crew is both confusing and significantly advantageous. Most other cult items disallow the crew from taking advantage of them, while the shield was perfectly harmless to pick up and entirely safe to use. 

# Testing
Spawned in a mirror shield on a local server, picked it up as a non-cultist and had my arm battered and dislocated. Works as intended.

# Changelog
:cl:  
tweak: Picking up a blood cult mirror shield as a non-cultist will cause significant arm damage and short term paralysis.
/:cl:

### DIFF
--- a/code/modules/antagonists/cult/cult_items.dm
+++ b/code/modules/antagonists/cult/cult_items.dm
@@ -1035,3 +1035,23 @@ GLOBAL_VAR_INIT(curselimit, 0)
 					throw_at(D.thrower, 7, 1, null)
 	else
 		..()
+
+/obj/item/shield/mirror/pickup(mob/living/user)
+	. = ..()
+	if(!iscultist(user))
+		if(!is_servant_of_ratvar(user))
+			to_chat(user, span_cultlarge("\"Foolish.\""))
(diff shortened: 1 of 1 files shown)


jachlompsky:
Mirror shields already won't block anything when used by non-cultists, cannot be thrown at cultists (who will auto-catch them, or the shield will simply bounce off), and will still summon illusions on hit which will try to attack the user, which I think is cooler than just making it damage your arm

SomeguyManperson:
Could use a warning message for sure, but it already backfires on attempted use

Kell-E:
I could probably drop the arm damage then, it seems like it's in excess. I'll wait to see if there's much other feedback first!

Moltijoe:
Balance Council:
>it already has a mechanic to prevent use by non-cultists, no need to add one that's far more damning
it also completely removes the danger of using it's throw as you'll never have it stolen from you

PR #20437, opened 2023-09-25
Adds new negative quirk - Drunkard

Kell-E:

Adds a new negative quirk related to the bar: Drunkard.
Spawn with an addiction to ethanol, 20u of ethanol in your system and a half empty (plastic, no smashing for you) bottle of lite beer.

Allows reagents to use alternative names for addictions (Eg. Ethanol addiction will instead be represented by a need for Alcohol)

In theory this negative quirk should cause a little more interaction with the bar on any given round. 

# Testing
Spawned in on a local server, ensured that all aspects of the quirk worked as intended. Spawning with ethanol in the system works, as does the half empty bottle of lite beer.

# Changelog

:cl:  
rscadd: Adds a new negative trait: Drunkard
/:cl:


Yarinoi:
full of ethanol

Yarinoi:
jokes aside being addicted to just ethanol would probably be a pain in the ass since you'd need to constantly get the chemists to get you to make ethanol to satisfy it. just a general addiction to shit that makes you drunk would be nice

Therandomhoboo:
> jokes aside being addicted to just ethanol would probably be a pain in the ass since you'd need to constantly get the chemists to get you to make ethanol to satisfy it. just a general addiction to shit that makes you drunk would be nice

Ethanol is the category that alcohol is in.

Kell-E:
Any alcohol will sate the addiction, it's not strictly ethanol. All alcohol is a subset of ethanol which qualifies.

My only question point in this is if it should be changed from saying Ethanol to Alcohol to avoid any confusion.

Yarinoi:
> > jokes aside being addicted to just ethanol would probably be a pain in the ass since you'd need to constantly get the chemists to get you to make ethanol to satisfy it. just a general addiction to shit that makes you drunk would be nice
> 
> Ethanol is the category that alcohol is in.

oh ok phew

code/modules/food_and_drinks/drinks/drinks.dm
```diff
@@ -273,6 +273,12 @@
 	desc = "Brewed with \"Pure Ice Asteroid Spring Water\"."
 	list_reagents = list(/datum/reagent/consumable/ethanol/beer/light = 30)
 
+/obj/item/reagent_containers/food/drinks/beer/light/plastic
+	name = "Carp Lite"
+	desc = "Brewed with \"Pure Ice Asteroid Spring Water\"."
```
Moltijoe: no need to re-define this

code/datums/traits/negative.dm
```diff
@@ -614,6 +614,51 @@
...
+	medical_record_text = "Patient is known to be dependent on alcohol."
+	reagent_type = /datum/reagent/consumable/ethanol
+	
+/datum/quirk/junkie/drunkard/on_spawn()
+	var/mob/living/carbon/human/H = quirk_holder
+	quirk_holder.reagents.add_reagent(/datum/reagent/consumable/ethanol, 20)
+	drug_container_type = pick(/obj/item/reagent_containers/food/drinks/beer/light/plastic)
+	. = ..()
+
+/datum/quirk/junkie/on_process()
```
Moltijoe:
```suggestion
/datum/quirk/junkie/drunkard/on_process()
```

code/datums/traits/negative.dm
```diff
@@ -614,6 +614,51 @@
...
+				break
+		if(!in_list)
+			H.reagents.addiction_list += reagent_instance
+			reagent_instance.addiction_stage = 0
+			to_chat(quirk_holder, span_danger("You suddenly feel like you need another drink..."))
+		tick_counter = 0
+	else
+		++tick_counter
+	
+/datum/quirk/junkie/check_quirk(datum/preferences/prefs)
```
Moltijoe:
```suggestion
/datum/quirk/junkie/drunkard/check_quirk(datum/preferences/prefs)
```

Moltijoe (changes requested):

cuackles:
> Any alcohol will sate the addiction, it's not strictly ethanol. All alcohol is a subset of ethanol which qualifies.
> 
> My only question point in this is if it should be changed from saying Ethanol to Alcohol to avoid any confusion.

Say alcohol, more recognisable even if just by a bit

code/modules/food_and_drinks/drinks/drinks.dm
Kell-E: Whoops! I'd forgotten!

code/datums/traits/negative.dm
Kell-E: Applied the fix to both of the bad datum paths. Whoops!

code/datums/traits/negative.dm
```diff
@@ -614,6 +614,51 @@
...
+			if(istype(entry, reagent_type))
+				in_list = TRUE
+				break
+		if(!in_list)
+			H.reagents.addiction_list += reagent_instance
+			reagent_instance.addiction_stage = 0
+			to_chat(quirk_holder, span_danger("You suddenly feel like you need another drink..."))
+		tick_counter = 0
+	else
+		++tick_counter
```
Moltijoe: rather than rewriting the entire on_process just to change a string, make that string into a variable and modify it for the new quirk

Moltijoe (changes requested):

Kell-E:
I've expanded a little on reagent functionality so that ethanol addictions can properly be represented and less confusing to the addicted.

code/datums/traits/negative.dm
Kell-E: A much better way to handle it, thanks!

Moltijoe (changes requested):
change the junkie warning to being a null variable rather than an empty string
then either print out the default line or the special line depending on if the special line exists or not

code/datums/traits/negative.dm
```diff
@@ -509,6 +510,7 @@
 		reagent_type = prot_holder.type
 	reagent_instance = new reagent_type()
 	H.reagents.addiction_list.Add(reagent_instance)
+	junkie_warning = ("You thought you kicked it, but you suddenly feel like you need [reagent_instance.name] again...")
```
Moltijoe: this will overwrite the drunkard one, only do this if it sees that junkie_warning isn't set

Moltijoe (changes requested):

code/datums/traits/negative.dm
Kell-E: I didn't realize it'd work that way, thanks for letting me know!

Kell-E:
This error is outside of my realm of understanding.

Kell-E:
At this point I feel like it's just inventing errors, I'm clueless.

Moltijoe (approved):

Moltijoe:
update the branch to master and it should fix the turdis problem

PR #20439, opened 2023-09-25
Eldritch Horror's got eye surgery???!? 

Therandomhoboo:

They are from the abyss, its dark in abyss so why can they no see in dark? I've never understood why they can't see in the dark. It makes no sense game wise, lore wise or sense wise.

It means horrors don't have to click around in the dark trying to find the vent quickly before sec gamers them.

# Changelog

:cl:  

rscadd: Horror's got that very needed eye surgery to give them vision in the dark

/:cl:

### DIFF
--- a/code/modules/antagonists/horror/horror.dm
+++ b/code/modules/antagonists/horror/horror.dm
@@ -10,6 +10,7 @@
 	melee_damage_lower = 10
 	melee_damage_upper = 10
 	see_in_dark = 8
+	lighting_alpha = LIGHTING_PLANE_ALPHA_MOSTLY_INVISIBLE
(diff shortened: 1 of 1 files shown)


LassiH:
they absolutely need this its so annoying to find targets in the dark with no way to see

Therandomhoboo:
> they absolutely need this its so annoying to find targets in the darks with no way to see

Only other thing I would want to do is make it so they could see targets whilst in the pipes, but I think that would be too strong. 
I would like it to be like the echolocation quirk @azzzertyy was trying to get in but only for when you're in the pipeworks.

Moltijoe (approved):

Moltijoe (approved):

PR #20447, opened 2023-09-26
Buffs toolbox damage like god intended (nerfs welder)

ynot01:

All toolboxes have received a +3 damage/throw damage buff, all welders have received a -3 damage nerf while turned on

If this gets merged I'll come back

Welding tools are small sized and are slightly more plentiful than toolboxes, so they should not be more powerful in terms of combat. Toolboxes are meant to be robust, which is reflected in their bulky size.

The main complaint I got when attempting a welding tool nerf was that crew need *some* kind of weapon, so I decided it should be toolboxes instead of welding tools, which are bulky instead of small sized.

The toolbox will also help fill the small amount of high damage blunt force in the game currently, which will lead to healthier rebalances of blunt force wounds if it is deemed necessary.

# Testing
Numbers change

# Wiki Documentation
Most toolboxes now deal 15 damage instead of 12.

Heirloom toolbox is now 8 damage instead of 5

The base damage of the ancient toolbox is now 22 instead of 19, meaning you now require 12+2TC instead of 15+2TC to reach the 34 damage threshold.

His Grace now starts at 15 damage instead of 12, which does not affect the amount of kills needed.

# Changelog

:cl:  
tweak: Toolboxes now universally do 3 more damage
tweak: Welding tool flames now do 3 less damage
/:cl:

### DIFF
--- a/code/game/objects/items/storage/toolbox.dm
+++ b/code/game/objects/items/storage/toolbox.dm
@@ -111,8 +111,8 @@
 	var/power = 0
 	for (var/obj/item/stack/telecrystal/TC in get_all_contents())
 		power += TC.amount
-	force = 22 + power
-	throwforce = 25 + power
+	force = initial(force) + power
+	throwforce = initial(throwforce) + power
 
 /obj/item/storage/toolbox/mechanical/old/clean/attack(mob/target, mob/living/user)
 	calc_damage()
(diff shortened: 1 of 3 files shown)


code/game/objects/items/storage/toolbox.dm
```diff
@@ -104,15 +104,15 @@
...
+	throwforce = 25
 
 /obj/item/storage/toolbox/mechanical/old/clean/proc/calc_damage()
 	var/power = 0
 	for (var/obj/item/stack/telecrystal/TC in get_all_contents())
 		power += TC.amount
-	force = 19 + power
-	throwforce = 22 + power
+	force = 22 + power
+	throwforce = 25 + power
```
Moltijoe: rather than being hardcoded, these should probably use initial()

Moltijoe (changes requested):

Moltijoe (approved):

MajManatee (approved):
I'm all for it. Makes no sense that the tiny item is the best weapon in the game for 99% of crew.

PR #20450, opened 2023-09-26
Makes cautery tool do burn damage + adv drill sharp

Moltijoe:
The nerf to welding tools will weaken being able to fight space vines
Since vines take 4x damage to burn, and 4x damage to sharp (16x if both)
this will give another weapon to fight vines for those that feel like running into maints after the vines without giving another civilian weapon

Also this just sorta makes sense, it's hot, it's gonna burn

:cl:  
tweak: Cautery tools have 3 force and do burn damage
tweak: Advanced Catuery tools in drill mode are sharp and have 4 force
/:cl:

### DIFF
--- a/code/modules/surgery/tools.dm
+++ b/code/modules/surgery/tools.dm
@@ -99,7 +99,6 @@
 	materials = list(/datum/material/iron=2500, /datum/material/glass=750)
 	w_class = WEIGHT_CLASS_TINY
 	toolspeed = 0.5
-	attack_verb = list("burnt")
 
 
 /obj/item/cautery/bone
--- a/code/modules/surgery/tools.dm
+++ b/code/modules/surgery/tools.dm
@@ -83,7 +83,7 @@
 	item_flags = SURGICAL_TOOL
 	tool_behaviour = TOOL_CAUTERY
 	w_class = WEIGHT_CLASS_TINY
-	force = 2
+	force = 3
 	damtype = BURN
 	attack_verb = list("burnt")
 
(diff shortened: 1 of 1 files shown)


RG4ORDR:
Do not make it sharp.
Medical tools got rightfully nerfed because of the Medical Hivemind that swarms and attacks antags and other such threats that violate the medbay. The drill was a secret weapon because it dealt like 10 damage but basically would guarantee a bleed wound. 
One hit was able to cause lacerations and repeated hits would upgrade to weeping. I dont think MDs need to have that type of power again.

MajManatee:
I don't think the dmg reduction is enough to warrant a change like this

Moltijoe:
> Do not make it sharp. Medical tools got rightfully nerfed because of the Medical Hivemind that swarms and attacks antags and other such threats that violate the medbay. The drill was a secret weapon because it dealt like 10 damage but basically would guarantee a bleed wound. One hit was able to cause lacerations and repeated hits would upgrade to weeping. I dont think MDs need to have that type of power again.

Dude, its 4 force, thats not enough to ever be useful as a weapon, the regular drill had like 10 force

Moltijoe:
> I don't think the dmg reduction is enough to warrant a change like this

The change makes welding tools go from 1 hitting vines to 2 hitting them, for a weapon based on charges that greatly hurt it.
This also doesnt make it a good weapon, it'll still be worse than the laser scalpel which has 11 force.
All this change does is add another "common" weapon thats useful against spacevines, but thats only because it abusing the damage modifiers that vines have

ToasterBiome (approved):

PR #20457, opened 2023-09-26
adds new speedy mutation to genetics

iloveloopers:

adds a new mutation only obtainable through polysmorphs, similarly to fire breath.
makes you go 10% faster, like polysmorphs.
has 40 instability.

more species specific genes are good.
the speed bonus is only 10% so it won't be a meta pick.

# Testing
tested on local, was able to find the gene from a polysmorph, make a mutator and run slightly faster

# Wiki Documentation

add this to guide to genetics:

mutation name: Speedy
description: "The user's leg muscles dispose of their lactic acid more efficiently and faster, allowing for faster running."
message: "You feel like you can run faster!"
how to obtain: Polysmorph species
incompatible with: none
instability: 40
compatible chromosomes: Reinforcement, Stabilizer

# Changelog

:cl:  
rscadd: Added new Speedy mutation, obtainable through Polysmorphs.
/:cl:

### DIFF
--- a/code/__DEFINES/DNA.dm
+++ b/code/__DEFINES/DNA.dm
@@ -57,6 +57,7 @@
 #define GELADIKINESIS		/datum/mutation/human/geladikinesis
 #define CRYOKINESIS /datum/mutation/human/cryokinesis
 #define CEREBRAL	/datum/mutation/human/cerebral
+#define SPEEDY		/datum/mutation/human/speedy
 #define THICKSKIN	/datum/mutation/human/thickskin
 #define DENSEBONES	/datum/mutation/human/densebones
 
--- a/code/datums/mutations/body.dm
+++ b/code/datums/mutations/body.dm
@@ -253,6 +253,24 @@
 		owner.physiology.brute_mod /= 0.8
 		owner.physiology.burn_mod /= 0.9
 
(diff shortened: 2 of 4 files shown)


ToasterBiome:
> the speed bonus is only 10% so it won't be a meta pick.

yes it will

iloveloopers:
> > the speed bonus is only 10% so it won't be a meta pick.
> 
> yes it will

polysmorphs aren't a meta pick species and they get a 10% speed bonus

Reed0506:
0.33 slowdown is extremely impactful in a lot of combat situations, so theres no reason that -0.1 slowdown wouldnt be either. this needs to have downsides.

you should probably also use blacklisted_movetypes to prevent this from making people faster while flying and floating (idk how to do it) so that this doesnt make you move faster in space.

Tenn0x:
I like having more species-restricted mutations and cloning different species to get them.

Moltijoe:
balance council:
>Speed is so powerful, especially when there's no direct downside and it's all the time

also, just from me, inert mutations are supposed to be latent mutations in a specific species genetic code, not what that species always is
look at lizard firebreath or ethereal shocktouch, neither have it innately, but it would totally make sense if they did

iloveloopers:
it is what it is

PR #20472, opened 2023-09-27
Buffs all of Detective's clothing so that they have the armor values they should

neeshacark:

Buffs all the detective's under items to have the same armor as the hard-worn suit (generic det suit, same armor as sec clothes) because they didn't.

Buffs all of det's coats to have the same armor as his vest so he need not choose function over fasion

being able to customize your look without having to worry about being less armored good

# Changelog

:cl:  cark
tweak: buffed detective's armor to be the values it rightfully should be
/:cl:

### DIFF
--- a/code/modules/clothing/suits/jobs.dm
+++ b/code/modules/clothing/suits/jobs.dm
@@ -52,7 +52,7 @@
 	item_state = "det_suit"
 	blood_overlay_type = "coat"
 	body_parts_covered = CHEST|GROIN|LEGS|ARMS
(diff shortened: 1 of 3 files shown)


Mqiib:
buffs fashion

iloveloopers:
antag buff

Moltijoe (changes requested):
remove /yogs/ from the type path and instead make them proper subtypes of the regular hat instead

neeshacark:
> remove /yogs/ from the type path and instead make them proper subtypes of the regular hat instead

this doesnt touch hats at all my good bitch

Marmio64:
The trenchcoat seems to cover limbs as well as chest, so I don't think it should have sec vest identical values as it'd be a direct upgrade.

neeshacark:
molti offered to do it later

PR #20477, opened 2023-09-28
Nerfs preternis again

Moltijoe:
The change to their emp immunity made them "weaker" but removed/weakened their other weaknesses to the point that it pretty much ended up being a buff.
This has resulted in them being the number 1 powergame species

:cl:  
tweak: Preternis water damage is stronger but starts out weaker
bugfix: Preternis water damage now properly makes them spark
tweak: Preternis burn mod increased to 1.2
tweak: Preterni are worse at punching
tweak: Preterni heat up and cool down slightly faster
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/preternis.dm
+++ b/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/preternis.dm
@@ -219,7 +219,7 @@
 		low_power_warning = FALSE
 
 /datum/species/preternis/proc/handle_wetness(mob/living/carbon/human/H)	
-	if(H.fire_stacks <= -1)//putting on a suit helps, but not if you're already wet
+	if(H.fire_stacks <= -1)
 		SEND_SIGNAL(H, COMSIG_ADD_MOOD_EVENT, "preternis_wet", /datum/mood_event/wet_preternis)
 		H.fire_stacks++ //makes them dry off faster so it's less tedious, more punchy
 		H.add_movespeed_modifier("preternis_water", update = TRUE, priority = 102, multiplicative_slowdown = 0.5, blacklisted_movetypes=(FLYING|FLOATING))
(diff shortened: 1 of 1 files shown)


Therandomhoboo:
Yes, preteni have been getting too much love. We need to teach them their place.

Marmio64:
They only heat up a tiny bit faster but lose their 2x heat damage. I'm not sure how this can be considered  a nerf?

Moltijoe:
> They only heat up a tiny bit faster but lose their 2x heat damage. I'm not sure how this can be considered a nerf?

cold is more dangerous as a result, 
it means plasmafires might be "safer"
but it'll be less "safe" to spacewalk without a suit
also this pr isn't ENTIRELY a nerf, just mostly a nerf

Marmio64:
> > They only heat up a tiny bit faster but lose their 2x heat damage. I'm not sure how this can be considered a nerf?
> 
> cold is more dangerous as a result, it means plasmafires might be "safer" but it'll be less "safe" to spacewalk without a suit also this pr isn't ENTIRELY a nerf, just mostly a nerf

You're calling it the number 1 power game species but can't commit to just a normal nerf?

Moltijoe:
> You're calling it the number 1 power game species but can't commit to just a normal nerf?

because i like the thematic of walking through fire and it fits them, but i'm also a pyromaniac

Marmio64:
> > You're calling it the number 1 power game species but can't commit to just a normal nerf?
> 
> because i like the thematic of walking through fire and it fits them, but i'm also a pyromaniac

I think preternis are already super fire resistant and that they don't really need heat damage reductions, at least not this large of one.

Moltijoe:
> I think preternis are already super fire resistant and that they don't really need heat damage reductions, at least not this large of one.

fair

ToasterBiome (approved):

PR #20478, opened 2023-09-28
Makes fire extinguishers more wet

Moltijoe:
# Why is this good for the game
Makes the most obvious way of abusing preterni more effective

:cl:  
tweak: regular fire extinguishers now make the person being sprayed more wet
/:cl:


SapphicOverload (approved):
water

PR #20482, opened 2023-09-28
Makes worldbreaker hungrier

Moltijoe:
The change to robotic eating made it easier for preterni to sustain power by default, this made the hungriness of worldbreaker less noticeable
not to mention, with them no longer instantly dying on 0 charge, they can afford to be even more hungry as it's less pressure to always stay charged

also gave them a bottomless stomach because lol lmao

:cl:  
tweak: Worldbreaker is even hungrier
tweak: Worldbreaker gives a bottomless "stomach"
/:cl:


Marmio64:
Include in the changeling the bottomless stomach maybe

ToasterBiome (approved):

PR #20484, opened 2023-09-28
[PORT] removes bloodsucker objectives

Moltijoe:
not quite full port
https://github.com/fulpstation/fulpstation/pull/1049

The objectives aren't that interesting and really just force bloodsuckers to act weirdly entirely for greentext, by nature bloodsuckers end up being on the louder end so them existing at all is usually enough to drive a round.

clan specific objectives still remain

keeps the vassal objectives by request

:cl:  @JohnFulpWillard @Moltijoe 
rscdel: Bloodsuckers can no longer get an objective to either drink blood or collect hearts
/:cl:


solwashere:
Fine with removing drinking blood, but vassilizing is one of the few non murdery objectives that we have for an antag

Moltijoe:
> Fine with removing drinking blood, but vassilizing is one of the few non murdery objectives that we have for an antag

clans still have their clan specific objective
this means they'll still have other things to do, just not things that are basically the tutorial for the antag

also vassalization is just the same as other conversion antags
bloodsucker shouldn't be forced into being a conversion antag unless it fits their clan

neeshacark:
Speaking from a staff perspective I feel like removing these objectives is going to cause a lot of ahelps for 'hey can i get some objectives' because people don't like not having directives to follow (they cant think of their own gimmick)

If we remove them they need to be replaced by something, either the new gimmick objectives or the generic objectives we use for traitors and such.

Moltijoe:
> Speaking from a staff perspective I feel like removing these objectives is going to cause a lot of ahelps for 'hey can i get some objectives' because people don't like not having directives to follow (they cant think of their own gimmick)
> 
> If we remove them they need to be replaced by something, either the new gimmick objectives or the generic objectives we use for traitors and such.

they still have a vassal target, if they manage to get that done as well as their clan objective and STILL want more objectives, that's them being an overachiever

code/modules/antagonists/bloodsuckers/clans/clan_gangrel.dm
```diff
@@ -21,3 +21,23 @@
 /datum/bloodsucker_clan/malkavian/on_favorite_vassal(datum/antagonist/bloodsucker/source, datum/antagonist/vassal/vassaldatum)
```
tattax:
```suggestion
/datum/bloodsucker_clan/gangrel/on_favorite_vassal(datum/antagonist/bloodsucker/source, datum/antagonist/vassal/vassaldatum)
```

code/modules/antagonists/bloodsuckers/clans/clan_gangrel.dm
```diff
@@ -21,3 +21,23 @@
...
 	var/datum/action/cooldown/spell/shapeshift/bat/batform = new(vassaldatum.owner || vassaldatum.owner.current)
 	batform.Grant(vassaldatum.owner.current)
+
+/// Enter Frenzy repeatedly
+/datum/objective/bloodsucker/frenzy
```
tattax: path

code/modules/antagonists/bloodsuckers/clans/clan_gangrel.dm
Moltijoe: lol, that was tattax apparently

code/modules/antagonists/bloodsuckers/clans/clan_gangrel.dm
Moltijoe: ty

ToasterBiome (approved):

PR #20489, opened 2023-09-29
Buffs all of Detective's clothing so that they have the armor values they should

Moltijoe:

Buffs all the detective's under items to have the same armor as the hard-worn suit (generic det suit, same armor as sec clothes) because they didn't.

Buffs all of det's coats to have the same armor as his vest so he need not choose function over fasion

being able to customize your look without having to worry about being less armored good

# Changelog

:cl:  @neeshacark 
tweak: buffed detective's armor to be the values it rightfully should be
/:cl:


Marmio64:
these items cover legs and arms unlike the standard det vest and sec vests, making them direct upgrades as far as I am aware. Perhaps reduce their resistances by 5 across the board?

ToasterBiome (approved):

PR #20494, opened 2023-09-30
Marrow weavers and gutlunches can no longer eat robotic organs

Moltijoe:
Marrow weavers can round remove ipcs because the brain is in the chest and they eat all chest organs
rather than adding a check to make sure they ignore brains, they just won't eat robotic organs, which makes more sense

:cl:  
tweak: Marrow weavers and gutlunches can no longer eat robotic organs
/:cl:


ToasterBiome (changes requested):
pretty sure that this will make it so they infinitely try to slurp up robotic organs then stop.

ToasterBiome (changes requested):
now test it and show me that it works properly

Moltijoe:
image
image

image

ToasterBiome (approved):

PR #20495, opened 2023-09-30
Reduce syndicate turret range but increase fire rate and health

Moltijoe:
Currently they have so much range that they're often used to cheese specific station locations rather than actually be used for defending the ship
they also have so much range that you can't even see them before they're shooting you

If only the range was reduced, it'd make it too easy to cheese them using regular firearms by ducking into and out of range
so i've doubled the health (more than doubled effective health since they fail to work below 60)
i've also increased the fire rate, so they're more likely to hit someone that's coming into their range

:cl:  
tweak: Reduce syndicate turret range but increase fire rate and health
/:cl:

### DIFF
--- a/code/game/machinery/porta_turret/portable_turret.dm
+++ b/code/game/machinery/porta_turret/portable_turret.dm
@@ -643,7 +643,8 @@
 	always_up = 1
 	use_power = NO_POWER_USE
 	has_cover = 0
-	scan_range = 9
+	max_integrity = 320 //double default
+	shot_delay = 1 SECONDS //33% faster than regular one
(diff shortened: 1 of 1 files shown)


Reed0506:
but instantly dying from offscreen was so much fun!!!! the turrets on the infiltrator have been very overpowered for a long time, this is probably a good change.

RG4ORDR:
Possible to beef their bomb armor up immensely?  
Common tactic I've seen is shuttle hunting with single tanks - while creative is sorta renders the turrets useless unless someone is moving the shuttle a lot.

PR #20503, opened 2023-10-01
Shifts around some Vampire ability unlocks + bat buff

Moltijoe:
Changes
Loss of stealthy drinking moved from 100 to 150
Vampire bat transformation moved from 200 to 150
Screech moved from 215 to 200
Disease touch moved from 200 to 225
Passive regen moved from 255 to 250
Full vampire power and cloak moved from 420 to 400

Should give vampires both more reason to go after people to succ them
as well as allowing them to be more risky once they're forced to be loud, by giving them a proper escape tool
and arguably, their bat form is far more interesting than jaunt, so this seeks to make it more enticing to use

# Wiki Documentation
Will need to update the numbers for when abilities are unlocked

:cl:  
bugfix: Can no longer infinitely drink bloodpacks that are empty
tweak: Shifted around some vampire blood thresholds to unlocking certain ability
tweak: Vampire bat is slightly faster and can see in the dark
tweak: Can no longer transform into a bat if incapacitated
tweak: Vampires get 1/2 the previous amount of usable blood from drinking blood packs
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/blood_pack.dm
+++ b/code/modules/reagents/reagent_containers/blood_pack.dm
@@ -32,7 +32,7 @@
 		return TRUE
 
 	while(do_after(user, 1 SECONDS, timed_action_flags = IGNORE_USER_LOC_CHANGE))
-		if(!reagents.total_volume)
+		if(reagents.total_volume <= 0)
 			user.balloon_alert(user, "empty!")
 			return ..()
 
@@ -42,7 +42,7 @@
 		)
 		if(is_vampire(user))
 			var/datum/antagonist/vampire/V = user.mind.has_antag_datum(/datum/antagonist/vampire)
-			V.usable_blood += 5
+			V.usable_blood += BLOODBAG_GULP_SIZE / 4 //they should really be drinking from people, yknow, be antagonistic?
 
 		reagents.reaction(user, INGEST, fraction)
 		reagents.trans_to(user, BLOODBAG_GULP_SIZE, transfered_by = user)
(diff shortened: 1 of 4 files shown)


code/modules/reagents/reagent_containers/blood_pack.dm
```diff
@@ -32,7 +32,7 @@
 		return TRUE
 
 	while(do_after(user, 1 SECONDS, timed_action_flags = IGNORE_USER_LOC_CHANGE))
-		if(!reagents.total_volume)
+		if(reagents.total_volume <= 0)
```
tattax: this doesn't change anything

yogstation/code/game/gamemodes/vampire/vampire_powers.dm
```diff
@@ -593,6 +593,8 @@
 	possible_shapes = list(/mob/living/simple_animal/hostile/vampire_bat)
 
 /datum/action/cooldown/spell/shapeshift/vampire/can_cast_spell()
+	if(owner.incapacitated(TRUE, TRUE))
+		return FALSE
```
tattax: use AB checks

code/modules/reagents/reagent_containers/blood_pack.dm
Moltijoe: Not true, if volume gets in the negatives somehow, it wont stop

tattax: it will stop cause that is how the reagent procs work

tattax: what would be a good idea would be to add an interaction key to the do afters

Moltijoe: i don't entirely know how interaction keys work, but i'll take a look

Moltijoe: sure enough, the do_after doesn't even target the bloodbag

ToasterBiome (approved):

PR #20505, opened 2023-10-01
Allows mime heretics to actually use all spells w/o needing to break their vow

cowbot92:

Changes all heretic spells to allow for non-speaking heretics (mimes/mute idiots) to use their full kit.

It's only fair.

# Testing

Yes I had to test every spell. No it wasn't fun.

# Changelog

:cl:  

tweak: adjusts all heretic spells to let mute people use them

/:cl:


ToasterBiome (approved):

ju45he:
Wait, does this mean that muzzling/mute toxin-ing a heretic will no longer serve to counter them? If so then that's a balance change rather than just a tweak.

PR #20440, opened 2023-09-25
No more wearable collars

LoliconSlayer:
# Document the changes in your pull request
Pet collars are for PETS leave your petplay fetish out of this

# Why is this good for the game?
Less weirdos is good for the game

# Testing
Absolutely none, I did this in webedit.

# Wiki Documentation
I hope to god we don't have wearable collars mentioned on the wiki.

# Changelog
:cl:  
rscdel: Pet collars are for actual pets
/:cl:

PR #20516, opened 2023-10-01
Bugfix for #20440, crusades are not an excuse for bad code

adamsong:

Original check disallowed pet collars to be worn on non-humans carbons (Monkies, true devil, aliens, etc) but allowed humans. #20440 switched this, only disallowing the wearing on humans, allowing it on other carbons, and I'm not sure this is intentional

# Changelog

:cl:  
bugfix: carbons being allowed to wear collars
/:cl:


JohnFulpWillard:
Why weren't monkeys allowed? Doesn't the bartender have a collar for their monkey?

Mqiib:
Allowing it for carbons but not humans is intended

adamsong:
Well that wasn't documented in the original PR, and if that is the case my other pr should be merged

PR #20517, opened 2023-10-01
Bugfix for #20440, crusades are not an excuse for bad code

adamsong:

Checking if carbon and if human is redundant, so the carbon check is omitted.

# Changelog
Code cleanup is not player facing


LoliconSlayer:
I blame Mqiib, he told me to do it like that

Moltijoe (approved):

PR #20554, opened 2023-10-06
You can no longer use heretic movement abillities while stunned

iloveloopers:

you can no longer use ash jaunt, mental obfuscation or void phase while incapacitated

"get out of jail free" cards are bad
if ash jaunt can't do it then none should be able to

# Testing

https://github.com/yogstation13/Yogstation/assets/140007537/ffb8015d-71e6-48e6-8b68-4b6ef6f7af1d

# Wiki Documentation
none

# Changelog

:cl:  
tweak: You can no longer use heretic movement abillities while incapacitated
/:cl:

### DIFF
--- a/code/modules/antagonists/eldritch_cult/eldritch_magic.dm
+++ b/code/modules/antagonists/eldritch_cult/eldritch_magic.dm
@@ -8,7 +8,7 @@
 	sound = null
 
 	school = SCHOOL_FORBIDDEN
-	cooldown_time = 20 SECONDS
+	cooldown_time = 15 SECONDS
 
 	spell_requirements = SPELL_CASTABLE_WITHOUT_INVOCATION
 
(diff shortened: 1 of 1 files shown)


Yarinoi:
bro went the COMPLETE opposite direction
i mean fair enough i guess noone would want to use ash jaunt if the rest work as get out of stun free cards

iloveloopers:
if balance council denies this that means they are being paid by cowbot

cowbot92:
definitely, I'm the shadow heretic council. And for the record, the reason the other two ""work"" while stunned is because 1 leaves a giant trail from point A to point B, and the second one leaves two massive AOE areas where they start and end at. 

If you want to make Ash Jaunt into a targeted teleport to compete and allow it to be used while stunned (and still leaving the target stunned) then you'd be better off doing that, than making these weird rage bait PRs.

iloveloopers:
I have plans that I cannot share right now because the haters will sabotage

PR #20557, opened 2023-10-07
Gives laptops in-hand sprites

Moltijoe:
So you can tell why someone is getting slowed, because you can see the open laptop in their hand

closed
image
image

open
image
image

:cl:  @Scrambledeggs00 @Moltijoe 
imageadd: Gives laptops in hand sprites
/:cl:


cuackles (approved):
make it so if they trip the laptop breaks and they get 20 head damage

PR #20566, opened 2023-10-08
Makes ethereal light power and range scale both with health and hunger + saturation reduction

Moltijoe:
So you can tell when your ethereal friend is hungy

doesn't touch max power, instead just reduces the saturation of any selected colour so it blinds less

# Testing
image

:cl:  
tweak: Ethereal brightness and range now scales with hunger in addition to health
tweak: Ethereal lights are slightly less overpowering
/:cl:


ToasterBiome (commented):
is this testing image after??

Moltijoe:
> is this testing image after??

yes, it really depends on the colour

MajManatee (approved):
I see no issue :)

PR #14924, opened 2022-07-20
Fixes getting stuck in cryo because of prosthetic limbs

SapphicOverload:
# Document the changes in your pull request

Exactly what the title says

# Wiki Documentation

No changes needed

# Changelog

:cl:  
bugfix: fixes getting stuck in cryo because of prosthetic limbs
/:cl:

PR #20571, opened 2023-10-08
I hate lazy medics, do your job shitheels Grrr - Cryoxadone nerf

MajManatee:
Makes cryoxadone heal less and metabolize slower to compensate
To prevent doctors from just shoving people into tubes then going afk for an hour

Encourages medbay players to do their jobs and engage with people, since healing with tools is faster 99% of the time, and forcing someone to have a black screen for an hour because they had a mechanical limb that was damaged is bad. The tubes are for when you are overwhelmed and have to put people aside, not a replacement for your entire job.

# Testing

number change

# Changelog

:cl:  
tweak: Cryox heals slower but also metabolizes slower
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -136,7 +136,7 @@
 	overdose_threshold = 100 //no chugging
 
 /datum/reagent/medicine/cryoxadone/on_mob_life(mob/living/carbon/M)
-	var/power = -0.00006 * (M.bodytemperature ** 2) + 6
+	var/power = -0.00006 * (M.bodytemperature ** 2) + 3
 	if(M.bodytemperature < T0C)
 		M.adjustOxyLoss(-3 * power, 0)
 		M.adjustBruteLoss(-power, 0)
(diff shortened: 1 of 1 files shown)


solwashere:
I don't like cryo and would rather be reliant on more engaging tools but this will just make it so that people who get shoved into cryo will be in there for longer periods of time as you can't really force people to care about other peoples time

neeshacark:
essentially what sol said, this will punish people who are injured - not doctors.

iamthesilly:
> forcing someone to have a black screen for an hour because they had a mechanical limb that was damaged is bad

My brother in christ I fixed this over a year ago
#14924

Therandomhoboo:
Yes, i agree with this.

SomeguyManperson:
Cryo is already about fine since it knocks people out when you use it so nobody actually wants to be cryoed unless its their only good option

What this does is make cryo only slightly better than using a sleeper or raw chemicals so there's even less reason to use it over either

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -136,7 +136,7 @@
 	overdose_threshold = 100 //no chugging
 
 /datum/reagent/medicine/cryoxadone/on_mob_life(mob/living/carbon/M)
-	var/power = -0.00006 * (M.bodytemperature ** 2) + 6
+	var/power = -0.00006 * (M.bodytemperature ** 2) + 3
```
iamthesilly: The other part of the equation needs to be halved too or it will cause damage between 223-273 kelvin instead of healing it
```suggestion
	var/power = -0.00003 * (M.bodytemperature ** 2) + 3
```

SapphicOverload (changes requested):

Moltijoe:
Balance Council:
>The only way to make medbay not use cryo is by removing it
this just punishes the people that get put in cryo, not the medical people themselves

ilikesaying:
why not limit it to organ damage or make it only able to heal damage to a certain level

Yarinoi:
REMOVE cryo ENTIRELY and replace it with the SOUL EXTRACTION MACHINE that heals people at the same rate at the cost of a PART OF THE MEDICAL DOCTOR'S SOUL (read: like, clone damage or something)

MajManatee:
yes i will remove cryo

PR #20569, opened 2023-10-08
Ports many improvements to lighting (directional lights)

JohnFulpWillard:
# Document the changes in your pull request

Ports https://github.com/tgstation/tgstation/pull/54520
Ports https://github.com/tgstation/tgstation/pull/52881
Ports https://github.com/tgstation/tgstation/pull/60239

Ports https://github.com/tgstation/tgstation/pull/53239
Ports https://github.com/tgstation/tgstation/pull/53816
Ports https://github.com/tgstation/tgstation/pull/52393

This is refactoring many things about the lighting system (and opacity), which I did while trying to fix Ethereal lights (which I failed at doing- for some reason it seems to work until power goes to 2< like it's got some hardcap on and I can't figure out why that is the case). I thought I should throw this stuff up anyways because why not.

It removes an undocumented change in https://github.com/yogstation13/Yogstation/pull/17271 about some lights not contributing to lumcount, idk why that was added but if it's a problem I can re-add. I mostly just tried to get it to be as close to TG's as possible so future work on it wouldn't be as bad.

This also adds directional lights, I'm not sure if it's something that is wanted so for now I only added it to seclites. The support is there, so you can add it to wherever you want after idk I don't want to feature code.

# Why is this good for the game?

Hopefully bringing lighting closer to TG's system will make future portings easier and allow Ethereals to finally work.

# Testing

image

# Spriting

# Wiki Documentation

# Changelog

:cl:  
rscadd: Seclites now have directional lights- point them in the direction you want to see more of.
experimental: Ported over many refactors to lighting and opacity. Please report if you can walk or see through walls unintentionally.
/:cl:

PR #20583, opened 2023-10-09
Finishes off remaining anti_magic_check and uses the proper version

JohnFulpWillard:

Antimagic checks are currently split between the old barely functional strung together shitcode of ``anti_magic_check``, and the better version that currently doesn't work as good because the component doesn't take it into account of ``can_block_magic``. This updates all instances of anti_magic_check to its proper can_block_magic use, and updates the component to only take that one into account now.

I also reverted the stealth nerf of Bibles against Revenants/Devils/Conversion, and added messages to the tinfoil hat letting you know when it gets used (since it has limited use).

This will be made even better with https://github.com/yogstation13/Yogstation/pull/20569 which adds bitflag support for the component, but that's not the point of the PR.

Ports
https://github.com/tgstation/tgstation/pull/76616 (Mostly)
https://github.com/tgstation/tgstation/issues/64124

# Testing

just trust me bro

# Spriting

# Wiki Documentation

# Changelog

:cl:  
bugfix: Bibles now protect against unholy stuff again.
bugfix: Antimagic protection now works properly and more consistently.
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -1016,8 +1016,56 @@
 		is_magic_blocked = TRUE
 	if((casted_magic_flags & MAGIC_RESISTANCE_HOLY) && HAS_TRAIT(src, TRAIT_HOLY))
 		is_magic_blocked = TRUE
+	
+	if(is_magic_blocked && charge_cost > 0 && !HAS_TRAIT(src, TRAIT_RECENTLY_BLOCKED_MAGIC))
+		on_block_magic_effects(casted_magic_flags, antimagic_sources)
 
 	return is_magic_blocked
+
+/// Called whenever a magic effect with a charge cost is blocked and we haven't recently blocked magic.
+/mob/proc/on_block_magic_effects(magic_flags, list/antimagic_sources)
+	return
+
+/mob/living/on_block_magic_effects(magic_flags, list/antimagic_sources)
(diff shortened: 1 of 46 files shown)


code/modules/mob/mob.dm
```diff
@@ -1006,19 +1006,66 @@
...
-	return																								//be changed later
+/// Called whenever a magic effect with a charge cost is blocked and we haven't recently blocked magic.
+/mob/proc/on_block_magic_effects(magic_flags, list/antimagic_sources)
+	return
+
+/mob/living/on_block_magic_effects(magic_flags, list/antimagic_sources)
+	ADD_TRAIT(src, TRAIT_RECENTLY_BLOCKED_MAGIC, MAGIC_TRAIT)
+	addtimer(TRAIT_CALLBACK_REMOVE(src, TRAIT_RECENTLY_BLOCKED_MAGIC, MAGIC_TRAIT), 6 SECONDS)
+
+/* //comment this in if you want antimagic to have a visible effect.
```
JohnFulpWillard, line 1033: This was part of the port but it wasn't a feature before so I didn't add it.
I can do it now if it's wanted, it essentially puts a glow around the character when they block magic (and plays a sound) for good feedback that the person is magic immune.

Moltijoe (approved):

PR #20596, opened 2023-10-11
Green Alert Shuttle arrival now only takes 15 minutes

Aquizit:

As the title says. Instead of 20m it's 15m.

As now there is a possibility (albeit rare) for a shift to come to an end at green alert, 20 minutes just seems... a long time.

I thought about just making it 10m like blue, but that can be decided on in the comments here and I can change it.

# Testing

Tested. Functional.

# Changelog

:cl:  
tweak: Evac shuttle now only takes 15 minutes to get to station during green alert
/:cl:

### DIFF
--- a/code/modules/shuttle/emergency.dm
+++ b/code/modules/shuttle/emergency.dm
@@ -229,7 +229,7 @@
 		var/security_num = seclevel2num(get_security_level())
 		switch(security_num)
 			if(SEC_LEVEL_GREEN)
-				set_coefficient = 2
+				set_coefficient = 1.5
 			if(SEC_LEVEL_BLUE)
 				set_coefficient = 1
 			else


Maia-J:
The yogs player hates the time for RP

Aquizit:
hey i'm fine if this gets turned down, just figured it could be an option for the code maints to decide

azzzertyy:
20 minutes is way too long and results in people raising alert level just to get it quicker, 15 might be more tolerable though

ShoziOffical:
yeah people will just make it blue to get shuttle to station faster

MajManatee:
Changes nothing, people will still change the alert. 
I don't think this needs to be changed

ShoziOffical:
> Changes nothing, people will still change the alert. I don't think this needs to be changed

you know i feel like green and blue should have same shuttle timer, the only different is security restrictions

MajManatee:
the extra 10 minutes can make quite a difference. Especially when antags need more time and switch to green off red.

ShoziOffical:
> the extra 10 minutes can make quite a difference. Especially when antags need more time and switch to green off red.

just recall

MajManatee:
> > the extra 10 minutes can make quite a difference. Especially when antags need more time and switch to green off red.
> 
> just recall

just set the alert level to blue?

Aquizit:
More controversial than I expected, just figured with as much people just change to blue after shuttle call it would have been an interesting idea.

PR #20610, opened 2023-10-13
Further Heretic Adjustments

cowbot92:

Moves Ash Jaunt from a purchasable spell to an innate spell all heretics learn upon becoming heretics

Removes the hookshot from the heretic blade

Removes Ash Jaunt from the forbidden knowledge, allowing heretics to pick 1 of the other movement abilities (Mental Obfuscation & Void Phase still cannot be picked together)

Moves Mask of Madness down from T2 to T1 for Ash Heretic

Adds a new T2 spell to Ash Heretic, "Volcano Blast" a strong bolt of fire that will deal 25 burn damage to the target it hits, and ignite the tiles around them, potentially igniting them and the person using it.

Corrects heretic ascension texts to no longer say you've ascended as multiple paths (oops!)

So for the Ash changes, I've observed a lot of heretics struggling with their early game still (this is the part where you collect the points to reach T2), and it's a mix of things. I also notice that like, 85% of everyone who picks heretic will pick this because it serves as a *very* good tool, but alone it's not good enough to make up for what it lacks. Allowing heretics to pick 2 mobility options maximum worked fairly well when I had initially added mind path, and I'd like to see how this helps them. If it is too potent I would consider nerfing the other movement abilities- but leaving this as as baseline ability.

As for the hook removal- I think it was an okay thing to have, but I really feel like the jank of not being able to control when you use it + the awkwardness of the hook not discerning between your target and other people really didn't help many heretics. I think leaving the blade as just a blade, and adding additional spells/abilities for them to purchase to augment their combat is a better path at this time.

Volcano blast is basically an alternative to amygdala assault/eldritch bolt - It does the same burn damage as eldritch bolt, but if used too close will also ignite the user, so it's much more situation. I like giving them more options though for ranged damage in T2, as it is the point where you're either made, or you break.

# Testing

I did test it all, the ash thing was the most annoying thing to make work. Can purchase volcano blast + all that jazz and all the other fixes are correct.

# Wiki Documentation

I will update the wiki :)

# Changelog

:cl:  
rscadd: Added new heretic spell: Volcano Blast
rscdel: Removed hookshot from heretic blade
bugfix: Fixed a few things with heretic ascension text
tweak: Tweaked some heretic spell paths for purchase
/:cl:


code/__DEFINES/qdel.dm
```diff
@@ -45,4 +45,14 @@
 #define QDELING(X) (X.gc_destroyed)
 #define QDELETED(X) (isnull(X) || QDELING(X))
 #define QDESTROYING(X) (!X || X.gc_destroyed == GC_CURRENTLY_BEING_QDELETED)
-
+///moved this because I need it to faster
```
Moltijoe: need it to WHAT faster?

code/__HELPERS/qdel.dm
```diff
@@ -1,10 +1 @@
...
-#define QDEL_IN_CLIENT_TIME(item, time) addtimer(CALLBACK(GLOBAL_PROC, PROC_REF(qdel), item), time, TIMER_STOPPABLE | TIMER_CLIENT_TIME)
-#define QDEL_NULL(item) qdel(item); item = null
-#define QDEL_LIST(L) if(L) { for(var/I in L) qdel(I); L.Cut(); }
-#define QDEL_LIST_IN(L, time) addtimer(CALLBACK(GLOBAL_PROC, PROC_REF(______qdel_list_wrapper), L), time, TIMER_STOPPABLE)
-#define QDEL_LIST_ASSOC(L) if(L) { for(var/I in L) { qdel(L[I]); qdel(I); } L.Cut(); }
-#define QDEL_LIST_ASSOC_VAL(L) if(L) { for(var/I in L) qdel(L[I]); L.Cut(); }
-
-/proc/______qdel_list_wrapper(list/L) //the underscores are to encourage people not to use this directly.
-	QDEL_LIST(L)
+///empty :)
```
Moltijoe: just remove the file if it's gonna be empty

Moltijoe (changes requested):

code/__HELPERS/qdel.dm
cowbot92: done

code/__DEFINES/qdel.dm
cowbot92: adjusted the comment to make more sense

Moltijoe (approved):

PR #20615, opened 2023-10-14
Nerfs cloaker belt to not be near infinite invisibility

Mqiib:
Coaker belt has been a little oppressive because it lets you be almost completely invisible and still do lots of stuff. This changes the belt to drain a bit slower passively in light but ALSO drain slightly every time you move. You'll still get the drop on people and be mostly invisible for a few seconds but you wont be able to be CONSISTENTLY invisible. Movement drain is low enough that walking keeps you hidden when in darkness, but running will begin to reveal you.

Cloaker belt drains slower in light passively but also drains from movement. Also adds variables for charge, drain, and movement loss speeds.

It was a little oppressive lets be real.

# Testing

Works as described, used the fancy new variables I slapped in for testing to tune it too.

# Wiki Documentation

Cloaker belt also drains on moving.

# Changelog

:cl:  

tweak: Cloaker belt now loses charge when you move
tweak: Cloaker belt drains slightly slower in light
tweak: Adds a couple tweakable variables
/:cl:

### DIFF
--- a/code/game/objects/items/devices/traitordevices.dm
+++ b/code/game/objects/items/devices/traitordevices.dm
@@ -210,6 +210,9 @@ effective or pretty fucking useless.
 	var/max_charge = 300
 	var/on = FALSE
 	var/old_alpha = 0
+	var/charge_speed = 25
+	var/decay_speed = 7.5
+	var/move_loss = 100
 	actions_types = list(/datum/action/item_action/toggle)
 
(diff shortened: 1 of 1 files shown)


Moltijoe (approved):
probably doesn't even need balance council approval

RG4ORDR:
I've used this with a buster arm before and it was actually insane that I was able to do combos and other actions and never get revealed because of how it operated in darkness.

PR #20618, opened 2023-10-14
Fixes significant bugs with every heretic grasp

Moltijoe:
the tattax pr is truely the gift that keeps on giving
:cl:  
bugfix: fixes bugs with every heretic grasp
/:cl:


MajManatee (approved):
uhh yeah i understand this

PR #20620, opened 2023-10-14
Changes the mouse opacity of two heretic effects

Moltijoe:
:cl:  
bugfix: Can no longer have your click blocked by glowing heretic runes
tweak: Clicking on pierced reality is no longer annoyingly precise
/:cl:


SomeguyManperson:
Might want to add transparent pixels to the pierced reality instead so you can still click under it

Moltijoe:
> Might want to add transparent pixels to the pierced reality instead so you can still click under it

I considered that, but since the sprite already takes up most of the tile its probably better people just get used to using alt click to get at stuff under it

MajManatee (approved):
sure

PR #20621, opened 2023-10-14
Bugfix: Allows heretics to actually get points when sacrificing targets (sorry...)

cowbot92:

ok so remember that PR I made that made it so people cost more points if they're sec/command/heretic. Well, it kinda broke gaining points by sacrificing. Whoops! It's ok though this one fixes it

# Testing
Somehow I fixed my private server and can sacrifice myself, yes this was tested with all 4 different conditions.

# Changelog

:cl:  

bugfix: fixes heretic sacrifices not giving points

/:cl:

### DIFF
--- a/code/modules/antagonists/eldritch_cult/eldritch_transmutations.dm
+++ b/code/modules/antagonists/eldritch_cult/eldritch_transmutations.dm
@@ -195,6 +195,7 @@
 			
 			else
 				EC.charge += 2
+			LH.target = null
 			EC.total_sacrifices++
 
 		if(QDELETED(LH.target))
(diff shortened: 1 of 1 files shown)


ToasterBiome (changes requested):
you can sacrifice the same target multiple times in a row because you just removed the line instead of moving it to after it finds the charges

cowbot92:
> you can sacrifice the same target multiple times in a row because you just removed the line instead of moving it to after it finds the charges

nuh uh

ToasterBiome:
> > you can sacrifice the same target multiple times in a row because you just removed the line instead of moving it to after it finds the charges
> 
> nuh uh

yuh huh

cowbot92:
> > > you can sacrifice the same target multiple times in a row because you just removed the line instead of moving it to after it finds the charges
> > 
> > 
> > nuh uh
> 
> yuh huh

ok fine you were right

ToasterBiome (approved):

PR #20627, opened 2023-10-15
Fixes a bug that prevents all prs from passing lintbot

Moltijoe:


Chubbygummibear (approved):
who would do such a thing

PR #20629, opened 2023-10-15
Makes ghost beacons able to be used more generally for mapping

redmoogle:

Will rip you back to the beacon if you leave the area
also a minor cleanup

Mainly to allow things like [this](https://forums.yogstation.net/threads/clown-cafe-ghostrole.27437/)

# Testing
Tested to make sure it would rip me back

# Wiki Documentation

Note that it won't let you leave the area (if theres even a page for it)

# Changelog  
:cl:
tweak: Ghost beacons now rip you back to their area
/:cl:


Moltijoe:
Is there merit to adding a variable to bypass the area lock?

redmoogle:
> Is there merit to adding a variable to bypass the area lock?

No; maybe a z level lock/area list though.
I don't see a legitimate case where ghosts should be able escape their intended zone

I might turn this into a implant though

Moltijoe (approved):

PR #20641, opened 2023-10-16
Makes sol dry not almost as good as actual medicine

Mqiib:
Stop side-stepping my nerfs goddamnit! Disgust shouldn't be this easy to remove!

Sol dry (the soda you can buy from any soda vendor) now merely triples your disgust drain. Still good, but not where it was which was almost the same level as psicodine, an actual medicine that isn't all over the station.

Helps make disgust still an annoying and potentially debilitating debuff if you're not careful with it. Also helps prevent continued abuse of certain mechanics that I was pretty sure I had already solved a while back.

# Testing

Just a number change.

# Wiki Documentation

It moderately reduces disgust levels rather than significantly.

# Changelog

:cl:  
  
tweak: Sol dry is no longer almost as good as real medicine

/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/drink_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/drink_reagents.dm
@@ -997,7 +997,7 @@
 	glass_desc = "A soothing, mellow drink made from ginger."
 
 /datum/reagent/consumable/sol_dry/on_mob_life(mob/living/carbon/M)
-	M.adjust_disgust(-5)
+	M.adjust_disgust(-1)
 	..()
 
 /datum/reagent/consumable/red_queen


Arachnid-Ass-Pics:
I know I didn't use this nearly enough for me to be the cause of this, but the timing is suspicious. Anyways, are you also going to nerf porcini sap as well? It can also reduce disgust as well, and any miner with a wrench to deconstruct wooden tables and make a barrel can make plenty of it considering the source is found exclusively on lavaland.

Yarinoi:
now its sol baddry...

Therandomhoboo:
It doesn't warrant a number change, it isn't a mechanic that is abused or a drink that is abused too. Seems like a pointless barkeep nerf.

MajManatee (approved):
I donny care

Therandomhoboo:
Bit lame that no discussion was done, considering drinks are really under-used atm and so far we lost the lemon-lime recipe so interaction has been less. Now drinks are getting needlessly nerfed (I would of been fine with it at least doing -2.5 instead of the fat drop)

PR #20649, opened 2023-10-17
Esword Quick Time Event Blocking

ynot01:

Adds QTEs to blocking with eswords and deswords. Blocking with these items is no longer RNG

When attacked, a blue square will pop up within 64 pixels of your mouse. You will need to move your mouse into this blue square in order to block

The square will give you 0.75 seconds to react, and the size of the square scales with `block_chance`

Items with this QTE behavior will have a special examine detailing what to do when people inevitably get confused by the popping blue squares on their screen

This will need a testmerge for bug squashing that I missed as well as balance testing. Originally I was going with a .5 second window but I felt that it would be unfair with >0 server ping. I will be able to see on live how actually playable it is.

#

Most of the code is in items.dm

Take this with you ✞

Skill expression > RNG; finally cool esword fights

I got this idea when listening to Duel of the Fates https://www.youtube.com/watch?v=D_2bluVPsb0

You should listen to it too

# Testing

# Changelog

:cl:  
rscadd: Added Quick Time Event blocking to replace RNG blocking for energy swords
/:cl:

### DIFF
--- a/code/__HELPERS/mouse_control.dm
+++ b/code/__HELPERS/mouse_control.dm
@@ -50,3 +50,38 @@
 		p_x = text2num(screen_loc_X[2])
 		p_y = text2num(screen_loc_Y[2])
 		return new /datum/position(x, y, z, p_x, p_y)
+
+/// Accepts a string screen_loc. NORTH, WEST, CENTER, etc are not allowed
+/proc/absolute_datum_map_position_from_screen_loc(client/client, screen_loc)
+	if(!isloc(client.mob.loc))
+		return
+	var/atom/A = client.eye
+	var/turf/T = get_turf(A)
+	var/cx = T.x
+	var/cy = T.y
+	var/cz = T.z
+	var/x = 0
+	var/y = 0
+	var/z = 0
+	var/p_x = 0
+	var/p_y = 0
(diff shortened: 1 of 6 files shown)


MajManatee:
cool concept, however I'm not sure its good. 
Click to attack, Move mouse off of space man to block, get attacked again while trying to block the first one, never get to swing back

azzzertyy:
id be more for it if this wasnt the easiest qte ever created

LassiH:
1 rythm game mid combat is bad
2 0.75 seconds of damage delay 
3 cant block while attacking
4 good idea for singleplayer games with no lag but yog is multiplayer with lag

JohnFulpWillard:
attachment

iamthesilly:
Balance council:
> Quick time events like this don't make sense in a game like SS13, the delay is a problem and with or without it this will result in even more of a gameplay advantage for having a good connection. It would also allow becoming effectively immune to ranged weapons.

Moltijoe:
very cool idea though, in my mind it's maybe worth making a new item and applying it to that instead
my main issue was it being on a weapon like the esword

ofscarletdreams:
we will unfortunately not be seeing traitors playing osu when going up against multiple people or a shotgun

PR #20657, opened 2023-10-17
donut emoji update

ktlwjec0:
updates the emoji donut to use the current donut
image

:cl:  ktlwjec
imageadd: Updates the donut emojis.
/:cl:


MajManatee:
But... we use both?

cuackles (approved):

PR #20658, opened 2023-10-17
destroys the entire codebase for golf

JohnFulpWillard:

This PR started as me trying to replace instances of ``initial(icon_state)`` with ``base_icon_state``, but the more I looked into Golf stuff the worse it got.

I started with replacing attackby with wrench_act, then it was kinda fucked, so I ported TG's improved tool_act system, which in the future can be used to log instances of people using tools on stuff (so admins can easily tell who used a screwdriver and multitool to hack a door without the need for buggy replay, for example).

I also tried to implement custom wrench/unwrench sounds because I saw the machine used one, then noticed that it uses ``Ratchet.ogg`` instead of ``ratchet.ogg``, which isn't a problem if you're running on Windows but Linux cares for case sensitivity (https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity) and we should probably try to support people who host there, so now it's been replaced.

I also saw a bunch of texts being sent to your chat for minor golf stuff like missing a hole, which I think is stupid because Golf is a multiplayer game, so instead it's now all balloon alerts that plays to everyone nearby. With this, I've also added some more feedback to playing, and you now have to rest to get golf balls out of the hole.

While testing, I saw the closet also didn't come with golf clubs. Looking at the code, it was all being spawned in New(), which ``PopulateContents()`` specifically says to use that instead of New() or Initialize() to populate contents. I also took the opportunity to cut down on some spam copy paste there too.

After finding this terrible thing, I decided to look into any other instances of crates loading things in New/Initialize. You know, the thing you're not supposed to do. I found one other instance of this- abandoned crates. I've removed it from there, then decided I might as well clean up things there as well. I tried testing it in-game and, varediting it being locked doesn't update its appearanaces. Awesome, now we have stealthily locked crates that can't blow up. I fixed that too, and now it uses multitool_act too.
I checked other case of ``vv_edit_var`` just for fun and saw the entire ARMOR FILES USES SPACES INSTEAD OF TABS
so I fixed that too.

I also removed the golf machine's power reliance, since it doesn't use any and is barely even a machine.

I also fixed golf working like, at all. Golf machine now uses Crossed instead of Cross, which isn't generally supposed to be used and caused it to not properly move the ball into the hole.

Golf works
Better and more consistent code overall
Removes unused climb stuff that's now an Element.

# Testing

trust me bro

# Spriting

# Wiki Documentation

# Changelog

:cl:  
bugfix: Golf now works.
bugfix: Different golf holes can now be differentiated.
bugfix: Golf doesn't need power anymore.
bugfix: Golf closets now come with all golf equipment.
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals/signals_mob/signals_tools.dm
+++ b/code/__DEFINES/dcs/signals/signals_mob/signals_tools.dm
@@ -1,0 +1,5 @@
+// Notifies tools that something is happening.
+
+// Sucessful actions against an atom.
+///Called from /atom/proc/tool_act (atom)
+#define COMSIG_TOOL_ATOM_ACTED_PRIMARY(tooltype) "tool_atom_acted_[tooltype]"
--- a/code/__DEFINES/tools.dm
+++ b/code/__DEFINES/tools.dm
@@ -24,3 +24,13 @@
 // If delay between the start and the end of tool operation is less than MIN_TOOL_SOUND_DELAY,
 // tool sound is only played when op is started. If not, it's played twice.
 #define MIN_TOOL_SOUND_DELAY 20
+
+// tool_act chain flags
+
+/// When a tooltype_act proc is successful
+#define TOOL_ACT_TOOLTYPE_SUCCESS (1<<0)
+/// When [COMSIG_ATOM_TOOL_ACT] blocks the act
+#define TOOL_ACT_SIGNAL_BLOCKING (1<<1)
+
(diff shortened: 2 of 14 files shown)


MajManatee:
The sinner that used spaces instead of tabs should be culled.

ShoziOffical:
tags war is real

MajManatee:
bro keeps removing them and wont say why

JohnFulpWillard:
you're adding them and they don't know why

ToasterBiome (approved):

PR #20662, opened 2023-10-18
Vampire revive flexibility pr

Moltijoe:
Makes vampire revive more flexible

also found a funny bug with vampire screech
that i've now fixed

:cl:  
tweak: Vampire revive only goes on cooldown after reviving rather than on button press
tweak: Vampire revive has a 1 minute cooldown instead of 100 seconds
tweak: Vampire revive can be canceled at any point after intial activation if you want to delay it
tweak: Vampire revive chapel check has been moved to when the revive happens rather than when the button is clicked
bugfix: Vampire screech no longer multiplies window damage and sound effects by the number of players nearby
/:cl:

### DIFF
--- a/yogstation/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/yogstation/code/game/gamemodes/vampire/vampire_powers.dm
@@ -314,7 +314,7 @@
 
 	check_flags = NONE
 	cooldown_time = 0 //no cooldown to the ability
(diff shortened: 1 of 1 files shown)


MajManatee (approved):
fine, but would be nice if they could do what changelings do when they revive, and they have to press a button to finialize it

PR #20679, opened 2023-10-19
No longer need to cut all wires to disassemble air alarm

warface1234455:
You only need to cut power wire to disassemble air alarm

Cutting all wires will trigger panic siphon and air alarm, then once you done disassembling the air alarm you have to rush to a nearby air alarm to fix the panic siphon, in the case theres no air alarm you are fucked

# Testing
whats

# Changelog

:cl:  

tweak: You only need to cut power wire to disassemble air alarm
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/airalarm.dm
+++ b/code/modules/atmospherics/machinery/airalarm.dm
@@ -731,7 +731,7 @@
 /obj/machinery/airalarm/attackby(obj/item/W, mob/user, params)
 	switch(buildstage)
 		if(2)
-			if(W.tool_behaviour == TOOL_WIRECUTTER && panel_open && wires.is_all_cut())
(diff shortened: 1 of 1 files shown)


MajManatee (approved):
neat

azzzertyy:
hm
is there not a better solution?
like just crowbaring the bar out instead of cutting wires?
just because its slightly different from normal behaviour makes it a little confusing, if its very different its a bit more clear

ShoziOffical:
> hm is there not a better solution? like just crowbaring the bar out instead of cutting wires? just because its slightly different from normal behaviour makes it a little confusing, if its very different its a bit more clear

uhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh you still have to cut the cable first before moving onto the next step of crowbarling the bar out, my pr is only about the annoying cable cutting thing that makes you to cut all wires to cut the cable

PR #20682, opened 2023-10-19
Upgraded sleepers no longer deal toxins instead of healing it

Moltijoe:
:cl:  
bugfix: Upgraded sleepers no longer deal toxins instead of healing it
/:cl:


ShoziOffical:
when not properly tested

ToasterBiome (approved):

PR #20687, opened 2023-10-20
Nerfs blazing oil (again)

Mqiib:
#19629 

I wasn't wrong. This time though I'm merely doubling extinguisher damage and leaving the brute mod as is because apparently people were mad about that.

Blazing oil is still blatantly op

# Changelog

:cl:  

tweak: (Nearly) doubles extinguisher damage against blazing oil blobs
experimental: it's like I was right the whole time
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
+++ b/code/modules/antagonists/blob/blobstrains/blazing_oil.dm
@@ -14,7 +14,7 @@
 	reagent = /datum/reagent/blob/blazing_oil
 
 /datum/blobstrain/reagent/blazing_oil/extinguish_reaction(obj/structure/blob/B)
-	B.take_damage(1.3, BURN, ENERGY)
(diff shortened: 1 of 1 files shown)


iloveloopers:
Just bomb it or something

LassiH:
im going to guess this will make blob just useless as fire  extinquishers do damage to multiple blob tiles  but hey laser meta  victory oh yeah cargo can just order wt rifles if crew cant use fire extinquishers

Tawaporah:
> Just bomb it or something
they did bomb me, I still won.

iamthesilly:
Alternatively, you could make lasers do reduced damage instead of no damage, and not give the blob free points for getting hit by them.

MajManatee:
Only on yogs will half the server be reduced to tears because they have to... adapt or change their strategy in any way, ever.

Moltijoe (approved):

PR #20684, opened 2023-10-19
[SEMI-PORT] New Heretic Path - Blade

cowbot92:
Full credit for nearly everything goes to the original PR located here: https://github.com/tgstation/tgstation/pull/65718

# Document the changes in your pull request

Adds a new heretic path, Path of Blade.

Base - "The Cutting Edge" - Pledges yourself to the path of Blade. Allows you to transmute a bar of silver with a knife or its derivatives into a Sundered Blade. Additionally, empowers your Mansus grasp to deal brute damage and paralyze enemies hit.

T1 - "Dance of the Brand" - Being attacked while wielding a Heretic Blade in either hand will deliver a riposte towards your attacker. This effect can only trigger once every 20 seconds.

Grasp Mark - "Mark of the Blade" - Grants you a single Eldritch, and allows you to craft further Eldritch Whetstones, one time use items that can enhance the sharpness of your blades up to a certain degree. Crafted with a heart, a diamond, and a sheet of plasteel.

T2 - "Stance of the Torn Champion" - Grants immunity to having your limbs dismembered. Additionally, when damaged below 50% of your maximum health, you gain increased resistance to gaining wounds and reduced damage slowdown.

Blade Upgrade - "Swift Blades" - Allows you to craft a bone blade from a knife and a bar of gold. Additionally will allow you to attack with both a sundered blade, and bone blade at once. The bone blade has a 10% chance to block incoming attacks.

T3 - "Furious Steel" - Grants you Furious Steel, a targeted spell. Using it will summon three  orbiting blades around you. These blades will protect you from all attacks, but are consumed on use. Additionally, you can click to fire the blades at a target, dealing damage and causing bleeding.

Ascension Rite - "Maelstrom of Silver" - Transmute three corpses to ascend as a Master of blades. You will become immune to stuns, resistant to wounds, and reduced overall damage. Additionally your spell 'furious steel' will have reduced cooldown, and you will create a shield of blades overtime.

https://github.com/yogstation13/Yogstation/assets/75333826/2c385ed0-2947-4bc4-9a52-3dce3068a4ec

# Why is this good for the game?

More options for antags is good, plus this spices up Heretic a little bit, which has so far been focused on being a bit more of a spell-caster, giving those who like to stay up close and personal a little chance to do some fun.

# Testing

Watch video for testing, all works. Somehow.

# Spriting

Screenshot 2023-10-19 122116
Screenshot 2023-10-19 122154
Screenshot 2023-10-19 122206
Screenshot 2023-10-19 122213
[...]

PR #20692, opened 2023-10-20
[bugfix] Fixes flesh heretic grasp bouncing off everything (including humans)

cowbot92:

Makes it so that the Flesh heretic grasp works properly like the other grasps (doesn't work on tiles/walls ((except rust)) but works on humans ((and dead ones)) )

# Testing

tested it live on discord dot com slash yogstation public

# Changelog

:cl:  

bugfix: fixed flesh grasp no worky 

/:cl:


Moltijoe:
I gotta figure out why that happens, might be because of the same reason signal handler wasnt working nicely.
Since itd be nice if the other returns were there, so itd only get used if the corpse actually gets turned into a ghoul

cowbot92:
> I gotta figure out why that happens, might be because of the same reason signal handler wasnt working nicely. Since itd be nice if the other returns were there, so itd only get used if the corpse actually gets turned into a ghoul

if you end up making a better fix than this just ping me on discord and ill close this

cowbot92:
Accidentally already fixed with https://github.com/yogstation13/Yogstation/pull/20684 god im good

Moltijoe:
bro, close this then

PR #20699, opened 2023-10-21
Tweaks ethereal saturation

Moltijoe:
After the saturation reduction they all look a bit too muted and it feels almost depressing to look at them

instead of modifying the lightness or saturation, instead puts a limit on them so it doesn't affect colours that are actually fine

:cl:  
tweak: tweaks ethereal colour saturation and lightness
/:cl:


iloveloopers:
bad PR!!!!

Yarinoi:
D:

i wanted to see cookpet be hardbanned...

ToasterBiome (approved):

PR #20707, opened 2023-10-22
Replaces exploit with gamebreaking exploit in PR request template

JohnFulpWillard:

The PR template says what ``[s]`` should be used for, but just about any bug in the game can be classified as an exploit. I really don't want to have to stalk the github page to see these random PRs that fix issues that aren't breaking the game, and I think abusing it would also be bad.
~~I also mentioned that maintainers aren't aware of PRs with [s] so they can bring it up themselves to someone available.~~ They do actually, I just haven't noticed it.

I've noticed a lot that ``[s]`` is abused frequently here, it's really stupid. Usually it's just to avoid commenters, which I do not agree with at all- stop trying to hide stuff from people god damn.

Not mentioning any PRs here because I don't want witch-hunts I just want this label to stop being used so liberally.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -118,6 +118,8 @@ There is no strict process when it comes to merging pull requests. Pull requests
 
 * We ask that you use the changelog system to document your player facing changes, which prevents our players from being caught unaware by said changes - you can find more information about this [on this wiki page](http://tgstation13.org/wiki/Guide_to_Changelogs).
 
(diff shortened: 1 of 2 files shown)


ToasterBiome:
Lock conversation *rubs hands together*

Addust:
is this because they secreted the AGCNR off-station should not gas station with fallout PR

ToasterBiome (approved):

cowbot92:
it would've been funnier if you used [s] for this one

PR #20712, opened 2023-10-22
i got your gimmick right here hisa

Yarinoi:
desperate to extend the length of rounds artificially? well do i have the pr for you!
this disables the majority of antags and events that cause early shuttle calls so we can aim for a round 50k on friday/saturday more easily

(this is to be testmerged btw)

antags disabled include:
nuclear operatives, clock cult, blood cult, revolutionaries, heretics
xenomorphs, blobs
events disabled include:
all types of meteors

let me know if any other antags should be disabled!
also i wouldve disabled loneops if i could have but i dont know how- i mean uhh i trust in our playerbase to secure the disk

# Changelog

:cl:  
experimental: we disabled some antags and events to make rounds last longer
/:cl:


PibeCantina:
Aliens and blob arent that big of an issue? I mean, i would say them spawning makes a round LONGER

Aquizit:
> Aliens and blob arent that big of an issue? I mean, i would say them spawning makes a round LONGER

This is correct, do leave in anything that delays the shuttle leaving, obviously. So xenos and blob should stay put.

ofscarletdreams:
nukies dont appear anyway because the light faded away

Yarinoi:
> Aliens and blob arent that big of an issue? I mean, i would say them spawning makes a round LONGER

> > Aliens and blob arent that big of an issue? I mean, i would say them spawning makes a round LONGER
> 
> This is correct, do leave in anything that delays the shuttle leaving, obviously. So xenos and blob should stay put.

eh, i think it's better to play it safe and disable them, since for every round where xenos/blob cause the round to last longer theres bound to be one where it causes them to be shorter
(plus a lot of the time after blobs are defeated the matter of recollecting all of the weapons that were distributed among the crew is practically impossible so people just shuttle call, dont see how this would be any different for xenos)
willing to have my mind changed though

Yarinoi:
> nukies dont appear anyway because the light faded away

feelsbadman

PibeCantina:
> > Aliens and blob arent that big of an issue? I mean, i would say them spawning makes a round LONGER
> 
> > > Aliens and blob arent that big of an issue? I mean, i would say them spawning makes a round LONGER
> > 
> > 
> > This is correct, do leave in anything that delays the shuttle leaving, obviously. So xenos and blob should stay put.
> 
> eh, i think it's better to play it safe and disable them, since for every round where xenos/blob cause the round to last longer theres bound to be one where it causes them to be shorter (plus a lot of the time after blobs are defeated the matter of recollecting all of the weapons that were distributed among the crew is practically impossible so people just shuttle call, dont see how this would be any different for xenos) willing to have my mind changed though

Fair enough, i think it's a matter of how you view it, i said my piece already.

Also i would maybe add heretic to the list since a lot of heretic rounds end at 50 minutes too, but maybe that was my experience.

Yarinoi:
> Also i would maybe add heretic to the list since a lot of heretic rounds end at 50 minutes too, but maybe that was my experience.

cause ascensions?

Aquizit:
This is no longer necessary. Preparations have been made, but thank you for the offer.

PR #20716, opened 2023-10-23
Fixes lints raised by 515 langserver

ynot01:
Replaces all instances of `/operator` in NTSL scripting files with `/oper` as 515 does not like `/operator/` being used in this way

Still raises the error that BYOND version is too new

Also exempts dm-langserver from max version check

image


JohnFulpWillard (approved):
fine for now

ShoziOffical:
fun

PR #20718, opened 2023-10-23
You can smash plates on heads

ktlwjec0:
You can smash plates on people's heads (harm intent, target head), causes a small knockdown. Same knockdown as a beer bottle. Plates have varying force depending on plate size, for damage applied during this. Still do no damage when attacking normally, as I like to imagine the damage is applied due to the smashing.

A fun little thing for cook fighting, it makes sense, and also I've had multiple people ask for this.

# Testing
Made sure it works on head only. Made sure knockdown+damage are reduced with armor.
Probably needs someone to check this over to make sure the code can't be cleaned up a bit but it definitely works.

:cl:  ktlwjec
rscadd: Plates smash and apply a short knockdown (same as beer bottles) when attacked on the head.
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/plate.dm
+++ b/code/modules/food_and_drinks/plate.dm
@@ -12,6 +12,7 @@
 	var/max_height_offset = 5
 	///Offset of where the click is calculated from, due to how food is positioned in their DMIs.
 	var/placement_offset = -12
+	var/smash_force = 10 //damage for head smashing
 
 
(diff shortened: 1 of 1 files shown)


Scrambledeggs00:
mazel tov cqc?!?!?!
Poggers

Therandomhoboo:
Kitchen powercreep >:C

ToasterBiome (approved):

PR #20722, opened 2023-10-24
Fixes accidental nerf to every two-handed weapon in the game

SapphicOverload:

When the two handed component as added, the way damage changes based on wielding was changed, but the damage of most two-handed weapons were never adjusted to account for this so now they all do 5 less damage than before because of an oversight.

Damage increase from wielding now works the way it used to, because it's easier to account for increases in weapon damage this way.

# Changelog

:cl:  
bugfix: fixed unintended change to damage of every two-handed weapon
bugfix: fixed runtime error when trying to use certain two-handed weapons with one hand
/:cl:


Therandomhoboo:
Tch... Atmos powercreep

ToasterBiome (changes requested):
can you check spears too? i remember this being an issue

iamthesilly:
> can you check spears too? i remember this being an issue

Turns out it was a lot more than just spears

SapphicOverload:
@ToasterBiome

JohnFulpWillard (commented):
i think it would be better to bring back the old behavior of adding the force ontop of the existing one, that way you can remove things like the component checking for sharpening.

bawbbawbins:
> i think it would be better to bring back the old behavior of adding the force ontop of the existing one, that way you can remove things like the component checking for sharpening.

hi jhony :DDDDD

JohnFulpWillard:
blocked

Moltijoe (approved):

ToasterBiome (approved):

PR #20736, opened 2023-10-24
Revert "Makes ghost beacons able to be used more generally for mapping"

Hislyna:
Reverts yogstation13/Yogstation#20629

The beacon is way way way too restricting where they cant even go half way across a ship, and cant even reach the bar in the space ninja zone. This is not needed and fixes a problem that doesnt exist. If you wanna make spawn points that restrict people, use something else and not this which worked fine as is instead of butchering something niche and making it useless.

Oh I also need the version as it was for events lol


MajManatee:
Or, you could fix the issue rather than just remove it? Even in this state it could be handy for some people.

Hislyna:
I need it unrestricted for events. If its handy, make a different version instead of butchering the one we had. There was 0 issue with the original.

Hislyna:
Closing cause https://github.com/yogstation13/Yogstation/pull/20737 exists

PR #20746, opened 2023-10-25
cleans up kitchen dmi

ktlwjec0:
There were two main kitchen.dmi files, containing mostly the same sprites in both of them. This makes it incredibly confusing when changing/adding kitchen sprites. This PR removes the yogs/ one, and moves any icons (such as bigspoon) that were just in there into the main dmi file. I've also reorganised the dmi file, so it is easier to find icons in the future.

# Testing
I've made sure the icons work, it is possible I've missed something but I can't see any issues.

# Spriting
No new sprites.

# Changelog
Nothing player facing.


azzzertyy:
guess were not rebasing

Chubbygummibear:
are these icons the same ones TG uses or are these actually Yog specific and should stay in a Yog folder?

cuackles:
the microwave icons seem to be missing from the non yogs specific file, is tha inteded ?

ktlwjec0:
> are these icons the same ones TG uses or are these actually Yog specific and should stay in a Yog folder?

Bigspoon and (maybe) sharpener only, but since it's not required to make seperate files anymore I figured I can avoid a dmi file with only 1 or 2 sprites in. If head dev wants it I can add it back though, it's no issue

ktlwjec0:
> the microwave icons seem to be missing from the non yogs specific file, is tha inteded ?

Yeah, they're in microwave.dmi

cuackles (approved):

PR #20747, opened 2023-10-25
Refunding discounted items no longer gives more tc than it costed

Moltijoe:
:cl:  
bugfix: Refunding discounted items no longer gives more tc than it costed
/:cl:


Hislyna:
Admins will no longer refund players that ask and will direct all complaints to Molti

Moltijoe:
> Admins will no longer refund players that ask and will direct all complaints to Molti

right, fine

ynot01 (approved):

PR #20752, opened 2023-10-26
Adds new security item: Anti-Magic Collars

cowbot92:
Special thanks to @Chubbygummibear for keeping me sane through all of this

Adds a new security item to counter caught magic users (heretics & wizards only ATM). This Item is called the Anti-Magic Collar, and it is a unremovable (through natural means) collar that is meant to dampen the danger that both Heretics & Wizards may cause to the station after being caught.  When equipped, the collar no longer allows you to cast magic spells that are blocked by anti-magic (without giving you full anti-magic protection) 

For now there are 3 of them in the security vendor for 100 credits each (or free if you're sec) but this can be moved/adjusted as needed.

Additionally fixes Heretic spells so that you cannot cast them if you have anti-magic, which is the whole point of this, Most wizard spells will already fall under being affected by this as well, as a bonus.

I believe this is good for the game for a number of reasons: Firstly, it sucks when you get caught early, round removed, and then get to watch as a ghost all round instead of getting to participate in SOME way besides "oh hey here's a blob spawn." Now if you get caught, as long as your crimes don't outweigh anything other than "this guy used magic once" security might think twice before just shoving you in the morgue. Bonus points for getting on the AI's good side if it's lawset isn't supposed to let them keep you dead.

In the end it's an alternative I would like the server to have to just straight up round ending people who don't deserve it.

# Testing

All testing for heretic spells + being able to purchase the collars was done in this video

https://github.com/yogstation13/Yogstation/assets/75333826/18672258-fb80-44f8-8423-b615d8a05193

# Spriting

New sprites, by me! **Updated

image

image

anti_magic_collar

anti_magic_collar_on_person

# Wiki Documentation

I will be adding this to the wiki as well.

# Changelog

:cl:  
rscadd: Adds new anti-magic item for security
tweak: Tweaks around some heretic spell reqs
imageadd: Adds new sprites for anti-magic collar
/:cl:


PibeCantina:
I'd say this should be armory only if you want it to has the slightest chance to pass balance clowncil, because this is incredibly powerful

cowbot92:
> I'd say this should be armory only if you want it to has the slightest chance to pass balance clowncil, because this is incredibly powerful

image

Moltijoe:
https://divinity.fandom.com/wiki/Source_Collar
Why is it not red, not blue?
but like actually, the collar blends in pretty hard even to just the mime outfit
a garish blue colour should help people recognize from a glance

UltraBigRat:
Would be great if there was some way to remove it, like swiping a sec id over it.

ju45he:
Maybe I'm blind but, no acid-resistence? I like it if it was an intended feature but it doesn't sound intended so consider fixing.
Also, nothings stopping someone from getting the AI to clone them, so this collar is just a minor annoyance.
Also, EMPing should allow magic for a limited time, but deal damage or something, that way people don't immediately run to medbay to get it taken off by cloning or decapitation or whatever.

cowbot92:
> https://divinity.fandom.com/wiki/Source_Collar Why is it not red, not blue? but like actually, the collar blends in pretty hard even to just the mime outfit a garish blue colour should help people recognize from a glance

idk I made the sprite in like 15 minutes

cuackles (approved):
L to helios

GDRdev:
LOVE OF GOD PLEASE USE THIS INSTEAD
image
image
image

RITA IS GONNA FORCE ONE ONTO ME IN FIVE SECONDS IF THIS GETS PATCHED IN

cowbot92:
> LOVE OF GOD PLEASE USE THIS INSTEAD image image image
> 
> RITA IS GONNA FORCE ONE ONTO ME IN FIVE SECONDS IF THIS GETS PATCHED IN

2nd one should be an IPC head

GDRdev:
> > LOVE OF GOD PLEASE USE THIS INSTEAD image image image
> > RITA IS GONNA FORCE ONE ONTO ME IN FIVE SECONDS IF THIS GETS PATCHED IN
> 
> 2nd one should be an IPC head

I dont know how to code soo if you want it in then I can sprite it for you but you gotta code it

Tipy1802:
I kinda like the idea of it as an OPTION for security but knowing how our rules and the administration works, sec will be forced to try and apply this any time they are fighting a magical entity which is just make it unnecessary harder. 

Its already such a massive headache for sec to have to metagame what antag has which ability and whats THE response that won't get you bwoinked that I don't think we should be making the more stressful role in the game even more stressful

JohnFulpWillard (changes requested):
This has to be reworked to use the newer anti magic system (that splits anti-magic checks between using and being used on)

cowbot92:
> This has to be reworked to use the newer anti magic system (that splits anti-magic checks between using and being used on)

The newer anti magic system doesn't work the way I want it to in regard to the collar; the clothing trait doesn't cover the idea that I want the user to not be able to cast yet not be magic immune, whereas by adding "blocked_magic_directionality" it's a much simpler check to see if you want the user to be able to receive magic, or cast it.

JohnFulpWillard:
is this not what you want - https://github.com/yogstation13/Yogstation/compare/master...JohnFulpWillard:Yogstation:antimagic-tie

https://github.com/yogstation13/Yogstation/assets/53777086/2fbd4343-fe08-4b20-97ce-8928cce8e01d

cowbot92:
> is this not what you want - [master...JohnFulpWillard:Yogstation:antimagic-tie](https://github.com/yogstation13/Yogstation/compare/master...JohnFulpWillard:Yogstation:antimagic-tie)
> 
>  8mb.video-2tn-nwDK3JwZ.mp4

Oh it is, that's perfect. Thanks

code/modules/clothing/neck/_neck.dm
```diff
@@ -193,6 +193,32 @@
...
+/obj/item/clothing/neck/anti_magic_collar
+	name = "anti-magic collar"
+	desc = "A tight collar used on prisoners to restrict their use of magic, while leaving them vulnerable to it's effects"
+	icon_state = "antimagiccollar"
+	resistance_flags = FIRE_PROOF
+
+/obj/item/clothing/neck/anti_magic_collar/equipped(mob/user, slot)
+	. = ..()
+	if(ishuman(user) && slot == ITEM_SLOT_NECK)
+		ADD_TRAIT(src, TRAIT_NODROP, CURSED_ITEM_TRAIT(type))
```
JohnFulpWillard: Move this to the proc below since it's the same one

JohnFulpWillard (changes requested):
All the new ``SPELL_REQUIRES_NO_ANTIMAGIC`` is intended, right?

cowbot92:
> All the new `SPELL_REQUIRES_NO_ANTIMAGIC` is intended, right?

Yes, let me double check though just to make sure i'm not stupid

Edit: Im not stupid, it is needed

code/modules/clothing/neck/_neck.dm
cowbot92: updated as requested

code/modules/clothing/neck/_neck.dm
```diff
@@ -193,6 +193,28 @@
...
+/obj/item/clothing/neck/anti_magic_collar
+	name = "anti-magic collar"
+	desc = "A tight collar used on prisoners to restrict their use of magic, while leaving them vulnerable to it's effects"
+	icon_state = "antimagiccollar"
+	resistance_flags = FIRE_PROOF
+
+/obj/item/clothing/neck/anti_magic_collar/equipped(mob/user, slot, initial = FALSE)
+	. = ..()
+	if(!(slot & slot_flags))
+		ADD_TRAIT(src, TRAIT_NODROP, CURSED_ITEM_TRAIT(type))
```
JohnFulpWillard, line 215: This makes it so if it is equipped to anything but the neck, it will be given Nodrop. Any security officer who buys this item will not be able to take it out of their hands.

Even working as intended, do we want this to even have nodrop in the first place? If so, maybe it should be togglable, maybe through some warden prisoner management console, otherwise security officers can get it stuck on themselves way too easily.

cowbot92: That might be good, in case some (idiot) puts it on by accident, as well as some alternative for heretics to free them. I'll look into making it manageable by the prisoner console I think. The "no drop" part comes from them being unable to remove it from their neck slot which the original thing was doing.

cowbot92: Updated 
image

JohnFulpWillard: It shouldn't have nodrop until it's locked by the console firstly I think.
You should be able to toggle the lock from the console too, you currently only have it set to unlock once and it becomes useless.

cowbot92: I'm not sure about useless, as you can just put it back on someone but I think I see what you're getting at. I'll see what I am able to work out.

cowbot92: > It shouldn't have nodrop until it's locked by the console firstly I think. You should be able to toggle the lock from the console too, you currently only have it set to unlock once and it becomes useless.

Don't you think it would be kinda dumb to have the collar not auto-lock on a prisoner you've put it on? It also doesn't really become useless since they still can't cast magic while it's on them, and you can still track it down if they take it ooff

JohnFulpWillard: Sorry, I forgot it autolocks on every single equip lol. 

I think autolocking is too easy to end up with people trapping themselves unintentionally. It's up to you because I don't think it matters much at the end or he day.

cowbot92: I think the unlock via the console is a good alternative for now, since the only people who have access to it are sec/command they *should* have some way to access the console (even via borg!)

cowbot92: Well not good alternative but like, good to just leave it as is.

cowbot92: > Sorry, I forgot it autolocks on every single equip lol.
> 
> I think autolocking is too easy to end up with people trapping themselves unintentionally. It's up to you because I don't think it matters much at the end or he day.

Do you think I should update the description to tell people about the locking on equip? Im not sure if people who play here even read that far ahead though

cowbot92: nah its fine now

JohnFulpWillard (approved):

PR #20755, opened 2023-10-26
Fix runtime spamming in human

warface1234455:
image

# Testing 
image
 dont mind the other runtimes, they happen because thats the only way to get my server run smoothly

WAAAAAAAAAAAAAAAAAAAAA

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -100,7 +100,7 @@
 			var/obj/item/organ/heart/ethereal/eth_heart = getorganslot(ORGAN_SLOT_HEART)
 			if(istype(eth_species))
 				. += "Crystal Charge: [round((nutrition / NUTRITION_LEVEL_MOSTLY_FULL) * 100, 0.1)]%"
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -100,7 +100,7 @@
 			var/obj/item/organ/heart/ethereal/eth_heart = getorganslot(ORGAN_SLOT_HEART)
 			if(istype(eth_species))
 				. += "Crystal Charge: [round((nutrition / NUTRITION_LEVEL_MOSTLY_FULL) * 100, 0.1)]%"
-			if(eth_heart)
+			if(eth_heart && istype(eth_heart, /obj/item/organ/heart/ethereal))
```
iamthesilly: You don't actually need the second argument for istype() here because you already defined eth_heart as that type

ShoziOffical: righ

SapphicOverload (approved):

iamthesilly:
~~Also you should give your PR an actual title~~
Nevermind, you changed it

ShoziOffical:
> ~Also you should give your PR an actual title~ Nevermind, you changed it

also included the testing

PR #20757, opened 2023-10-26
AGCNR healium gas interaction addon

warface1234455:

Healium inside coolant can now restore integrity of the reactor. This only work if the reactor is below 1800K

So if you have a good setup you can get the reactor stable at around 1700kelvin

Updated guide to reactor 101 paper
Allow more failsafe gas to interact with the reactor, more works for atmosian. Give more crazy reactor setup ideas to atmosians

# Testing
works, also thank you @SapphicOverload for helping me with math (Im dumb at it)
ww

# Wiki Documentation
When the reactor wiki page exists, please add this in

# Changelog

:cl:  

rscadd: Healium inside coolant can now restore integrity of the reactor. This only work if the reactor is below 1800K
spellcheck: Updated guide to reactor 101 paper
/:cl:

### DIFF
--- a/code/modules/power/reactor/reactor.dm
+++ b/code/modules/power/reactor/reactor.dm
@@ -1047,7 +1047,7 @@
 	- Pluonium: When you need weapons grade plutonium yesterday. Causes your fuel to deplete much, much faster. Not a huge amount of use outside of plutonium production or sabotage.<BR>\
 	- Healium: Can restore integrity if below 1800 Kelvin. The restoration rate is depended on the temperature, the lower the temperature the faster it is to restore integrity.<BR>\
 	<BR><B>Coolant effects</B><BR>\
-	- The higher heat capacity gas allows the reactor to increase coolant effciency.<BR>\
+	- High heat capacity gases like water vapor and plasma are better coolants as they can transfer more heat per mole. The inverse is true for low heat capacity gases like tritium and antinoblium.<BR>\
 	<BR><B>OH GOD IT'S SCREAMING AT ME WHAT DO I DO</B><BR>\
 	Don't panic! There's a few things you can do to prevent the station from becoming an irradiated hellscape.<BR>\
 	Scenario 1: Overheating<BR>\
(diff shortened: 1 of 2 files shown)


code/modules/power/reactor/reactor.dm
```diff
@@ -1035,6 +1045,9 @@
 	- Hyper-Noblium: Extremely efficient permeability increase (10x as efficient as bz)<BR>\
 	Depletion types:<BR>\
 	- Pluonium: When you need weapons grade plutonium yesterday. Causes your fuel to deplete much, much faster. Not a huge amount of use outside of plutonium production or sabotage.<BR>\
+	<BR><B>Coolant effects</B><BR>\
+	- The higher heat capacity gas allows the reactor to increase coolant effciency.<BR>\
+	- Healium has low heat capacity but it can restore integrity if below 1800 Kelvin.<BR>\
```
iamthesilly: This is the moderator effects section of the guide, healium in the moderator has no effect on cooling so it doesn't make much sense to bring it up here.

code/modules/power/reactor/reactor.dm
```diff
@@ -493,6 +498,11 @@
 		color = COLOR_CYAN
 	else
 		color = null
+
+	if(vessel_integrity < 400)
+		vessel_integrity += integrity_restoration
+	else //incase it goes beyond
+		vessel_integrity = 400
```
iamthesilly: Please use initial(vessel_integrity) instead of hardcoded numbers, also check if it's above after adding to it instead of before so it doesn't go over the limit for one tick

SapphicOverload (changes requested):

code/modules/power/reactor/reactor.dm
```diff
@@ -1035,6 +1045,9 @@
 	- Hyper-Noblium: Extremely efficient permeability increase (10x as efficient as bz)<BR>\
 	Depletion types:<BR>\
 	- Pluonium: When you need weapons grade plutonium yesterday. Causes your fuel to deplete much, much faster. Not a huge amount of use outside of plutonium production or sabotage.<BR>\
+	- Healium: Can restore integrity if below 1800 Kelvin. The restoration rate is depended on the temperature, the lower the temperature the faster it is to restore integrity.<BR>\
+	<BR><B>Coolant effects</B><BR>\
+	- The higher heat capacity gas allows the reactor to increase coolant effciency.<BR>\
```
iamthesilly:
```suggestion
	- High heat capacity gases like water vapor and plasma are better coolants as they can transfer more heat per mole. The inverse is true for low heat capacity gases like tritium and antinoblium.<BR>\
```

SapphicOverload (changes requested):

SapphicOverload (approved):

PR #20792, opened 2023-10-30
Fixes another tattax bug (touch spells valid target)

Moltijoe:
:cl:  
bugfix: Touch spells now properly check if the target is a valid target
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/touch/_touch.dm
+++ b/code/modules/spells/spell_types/touch/_touch.dm
@@ -169,11 +169,13 @@
 	SHOULD_NOT_OVERRIDE(TRUE) // DEFINITELY don't put effects here, put them in cast_on_hand_hit
 
 	if(!proximity_flag)
-		return
-	if(victim == caster)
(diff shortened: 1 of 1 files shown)


code/modules/spells/spell_types/touch/_touch.dm
```diff
@@ -169,11 +169,13 @@
...
-		return
-	if(!can_cast_spell(feedback = FALSE))
-		return
+		return FALSE
+	if(!can_cast_on_self && victim == caster)
+		return FALSE
+	if(!is_valid_target(victim))
+		return FALSE
+	if(!can_cast_spell(feedback = TRUE))
+		return FALSE
```
tattax:
```suggestion
		return
	if(!can_hit_with_hand(victim, caster))
		return
```

Maia-J:
I actually had to go around this with making an override with my invisible touch spell

ToasterBiome (approved):

PR #20798, opened 2023-10-31
Adds TG achievement system (stored on DB)

JohnFulpWillard:

Many achievements don't have an icon because I am not a spriter, I set it to use defaults.

### migration stuff

This doesn't migrate current achievements over, because honestly they use two different systems of ID and are stored in different places in different ways and are just aaargh *melts like a boss*
I was told many of the achievements are broken but I have no way to confirm this as there's no way (that I was told) to check how many people completed an achievement.

New achievements should work more consistently, have a better UI, have support for scores and proper leaderboards, as well as better system to unlock achievements (w/ a sound when you get one) and updates when you get the achievement.
You can also compare achievements to see how many people have it, and how many people have it when compared to the most commonly achieved achievement.
We can also maybe comment in that achievement about suplexing a rod if someone adds the achievement itself.

Achievements also don't work on localhost, so you won't be spammed about swapping hands achievement every time you run a server locally.

# Testing

None this should be testmerged or smth i don't have a database set up on localhost

# Spriting

Ported from TG
image

I left mafia sprites in because this originally started as a port for mafia and plan on using them

# Wiki Documentation

# Changelog

:cl:  
rscadd: Added a new achievement system.
rscdel: Deleted old achievements.
/:cl:


SQL/database_changelog.txt
```diff
@@ -1,14 +1,34 @@
...
+
+CREATE TABLE `achievements` (
+	`ckey` VARCHAR(32) NOT NULL,
+	`achievement_key` VARCHAR(32) NOT NULL,
+	`value` INT NULL,
+	`last_updated` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+	PRIMARY KEY (`ckey`,`achievement_key`)
+) ENGINE=InnoDB;
+
+version 5.14 2023-05-10
```
ToasterBiome: what happened that 5.13 was wrong

PR #20815, opened 2023-11-01
[DONUT] Fixes weird door access on donut (other misc stuff too)

bruhlookatthisdood:

Anyone with maintenance access could open janitor closet, this is no longer the case.
Security maintenance door required maintenance access along with brig access to open, now it only requires brig access asi s everywhere else.
I also did this for one of the "security" doors lower down, you no longer need maintenance access, only brig.
Medical's cryo room now only will accept those with medical access.
Medical's cryo room door was also.. under geneticists area? If the freezer is in exam room area the door should be too that makes no sense.

Doors having correct access and lining up with rest of the map is good.
As for the cryo room (in the picture below) I don't think its right for anyone with normal maintenance access from an assistant to an officer to be able to walk into that room. That freezer is semi-sensitive equipment to medical. I could undo this if needed.

Also, the area. Why was it in gene, that makes no sense. Its now under exam room.
image

# Testing
None needed

# Wiki Documentation

None

# Changelog

:cl:  
bugfix: donut janitor closet/security maintenance doors have better access requirments.
bugfix: donut medical cryo freezer room maintenance door no longer lets those with maintenance access in and is now under the exam room's area.
/:cl:

### DIFF
--- a/_maps/map_files/DonutStation/DonutStation.dmm
+++ b/_maps/map_files/DonutStation/DonutStation.dmm
@@ -35363,7 +35363,7 @@
 "puy" = (
 /obj/machinery/door/airlock/maintenance{
 	name = "Custodial Maintenance";
-	req_one_access_txt = "12;26"
+	req_access_txt = "26"
 	},
 /obj/machinery/door/firedoor/border_only{
 	dir = 4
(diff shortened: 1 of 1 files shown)


ToasterBiome:
This isn't Gax?

bruhlookatthisdood:
> This isn't Gax?

i get donut and gax confused
mb

neeshacark (approved):
all seems well

neeshacark:
if this causes merge conflicts with my donut pr i am going to kill you

bruhlookatthisdood:
> if this causes merge conflicts with my donut pr i am going to kill you

womp womp

PR #20819, opened 2023-11-02
Fixes replay lighting and player directionality

Chubbygummibear:
currently the replay system doesn't catch lighting updates and only updates in response to specific interactions leading to chunky looking like this 
image
additionally players' directionality doesn't update so i did my best to fix that as well.

## Here's how it looks now with my fixes

https://github.com/yogstation13/Yogstation/assets/46236974/fdb4d2ae-8e14-4ccf-bf63-6e29cf779a5b

It's still a bit chunky but at least it updates. Also replay overlays seem to be broken but only on non chromium browsers??
i also mostly fixed console screen directionality on accident lmao
before:
image
after:
image

# Testing

https://github.com/yogstation13/Yogstation/assets/46236974/fcac3032-a0ce-4ede-8e34-76480f5ad268

game seems fine. no idea if updating the cached appearance of players on tick in the demo viewer will make the server's performance shit.

Having better replays is good

# Changelog

:cl:  
bugfix: lighting should actually update in replays 
bugfix: player directionality should actually update in replays  
/:cl:

### DIFF
--- a/code/__DEFINES/subsystems.dm
+++ b/code/__DEFINES/subsystems.dm
@@ -236,7 +236,9 @@
 				AA.copy_overlays(changed_on, TRUE);\
 			}\
 		} \
-	}
+	}\
+	if(isturf(changed_on)){SSdemo.mark_turf(changed_on);}\
+	if(isobj(changed_on) || ismob(changed_on)){SSdemo.mark_dirty(changed_on);}\
 
 /**
 	Create a new timer and add it to the queue.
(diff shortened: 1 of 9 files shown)


neeshacark:
oh my god someone found out how to fix monsters autism code

fix atmos next so we can have real multiz

ShoziOffical:
Can we have doors too

Chubbygummibear:
> Can we have doors too

monster said the replay viewer can't work with things that use vis_contents which doors do. it's possible i might be able to see if doors could be updated to use overlays instead of whatever vis_overlays is for, because like literally everything else uses overlays so i don't see why that wouldn't work

Moltijoe (approved):

PR #20823, opened 2023-11-03
Mecha Scanning module now effects all energy drain :: Mech steps no longer cost energy

ynot01:
Scanmod has been repurposed to reduce the cost of all energy costs (1x-0.5x-0.33x-0.25x) instead of controlling mech step cost.

Mech step cost for all mechs except for spawned mechs (ERT, nukies, honk, reticence) is now free.

Mechs costing energy to move made them pretty much pointless to build until T4 parts arrived, unless you wanted to carry an inducer and recharge every so often (which not everyone has access to), but even then most roboticists don't bother building unless it's for a bounty until T4.

Mechs shall be liberated from this cost, and tier parts shall become auxillary rather than defining as they were meant to be.

# Testing
Steps no longer cost energy as expected and weapon energy costs are adjusted as expected

# Wiki Documentation
amount *= (2.5 - (scanmod.rating / 2))

Solaris fire with T1: 120 cost
Solaris fire with T4: 30 cost

# Changelog

:cl:  
tweak: All mechs built by hand no longer cost energy to move
tweak: Scanning modules on mechs now affect all power usage
/:cl:

### DIFF
--- a/code/game/gamemodes/clown_ops/clown_weapons.dm
+++ b/code/game/gamemodes/clown_ops/clown_weapons.dm
@@ -287,7 +287,6 @@
 	internals_req_access = list(ACCESS_SYNDICATE)
 	wreckage = /obj/structure/mecha_wreckage/honker/dark
 	max_equip = 3
-	step_energy_drain = 3
 
 /obj/mecha/combat/honker/dark/GrantActions(mob/living/user, human_occupant = 0)
 	..()
--- a/code/game/gamemodes/clown_ops/clown_weapons.dm
+++ b/code/game/gamemodes/clown_ops/clown_weapons.dm
@@ -287,6 +287,7 @@
 	internals_req_access = list(ACCESS_SYNDICATE)
 	wreckage = /obj/structure/mecha_wreckage/honker/dark
 	max_equip = 3
+	step_energy_drain = 3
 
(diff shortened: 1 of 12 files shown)


00ze-cyclone:
The mech energy weapons already consume an insignificant amount of energy to begin with so scanner reducing that wouldn’t shake up balance much. Seem like a better system.

ynot01:
> The mech energy weapons already consume an insignificant amount of energy to begin with so scanner reducing that wouldn’t shake up balance much. Seem like a better system.

I considered having the energy weapons consume more energy with this, but I wasn't quite sure and the amount of features is already stretching here, so it may be better suited for a future PR

Moltijoe (approved):

githubuser4141:
I thought the lore was that exofabs were like 3D printers and the resulting product was of lower quality then if they were built properly. If exofab'd mechs are going to get this treatment, why not Antag/Centcom mechs - they should be of much high quality.

Also this is like saying "internal combustion engine or electric vehicles IRL are useless because no one wants to constantly refill a gas tank or charge a battery." Mechs are supposed to have a limited range/usability cause they're designed to be special purpose AKA you don't spend all round in one and it requires having to retreat to a charge pad every so often to be charged. Mechs at T2 and T3 are otherwise perfectly usuable if you're aware of the range limitation - even at T1 they are usuable for short range combat.

PR #20830, opened 2023-11-04
Fixes a few fire_stack bugs

Moltijoe:
TG status code is different, they pre-modify the variable passed to tick
meaning i just needed to multiply a few variables by seconds and it was good to go

:cl:  
bugfix: Hug fire_stack transfer now works properly again
bugfix: Drying off no longer takes 10x as long as it used to
bugfix: Being on fire heats up as fast as it used to
/:cl:


ynot01 (approved):

ynot01:
game breaking!

PR #20837, opened 2023-11-05
Makes clockwork golems no longer robotic (they don't suffer EMP from their own weapons anymore).

ChesterTheCheesy:
Replaces MOB_ROBOTIC with MOB_INORGANIC like the rest of the golems.

The golems being robotic is a cool flavor, but since the change to bodyparts this made them like other robotic races like IPC.
They have no use for robotic limbs mechanics, they have their own source of healing (sentinels compromise).
This mostly makes anyone who uses EMP-based weaponry (warhammer and brass sword) not die to themselves.

# Testing
Uuh single line of code just makes the clock golems behave like other golems.

# Changelog
:cl:  
tweak: clockwork golems are now golems and not robots made of brass, making them immune to the effects of their own EMP
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/golems.dm
+++ b/code/modules/mob/living/carbon/human/species_types/golems.dm
@@ -731,7 +731,6 @@
 	limbs_id = "clockgolem"
 	info_text = "<span class='bold alloy'>As a </span><span class='bold brass'>Clockwork Golem</span><span class='bold alloy'>, you are faster than other types of golems. On death, you will break down into scrap.</span>"
 	species_traits = list(NOBLOOD,NO_UNDERWEAR,NOEYESPRITES,NOFLASH)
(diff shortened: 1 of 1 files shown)


iamthesilly:
You could just add TRAIT_EMPPROOF_SELF to their species traits instead

PibeCantina:
I think adding the trait would be better, so you keep the benefits of being able to weld and use cable on yourself for healing

iloveloopers:
> I think adding the trait would be better, so you keep the benefits of being able to weld and use cable on yourself for healing

When would you ever do this

PibeCantina:
> > I think adding the trait would be better, so you keep the benefits of being able to weld and use cable on yourself for healing
> 
> When would you ever do this

Fucking never, but it's better to have than not to, since you can still heal using other cogger methods, so, why not keep it when we can? Just another benefit from being a cogger golem

ToasterBiome (approved):

PR #20839, opened 2023-11-06
DNA vault tweaks + addition

Moltijoe:
Tox option also gives rad immunity and stable liver
Self sufficient also gives stable heart
fireproof option also gives 50% pressure damage resistance
stuntime option also gives reduced damage slowdown
armour option reduces the chance of getting wounded but gives less armour
movement speed option also gives crawlspeed but gives less movespeed
arm speed option also gives do_after speed but gives less action speed

new option
arm strength
pick up bodies faster
can't get disarmed by being pushed into a wall + can't have something you're pulling be pulled out of your grip
stronger punches

it's something that only really gets built lowpop or greenshift extended, might as well make it slightly more diverse

# Testing
gotta

:cl:  
rscadd: Adds a new option to the dna vault
tweak: tweaks the effects of some dna vault options
/:cl:

### DIFF
--- a/code/modules/goals/station_goals/dna_vault.dm
+++ b/code/modules/goals/station_goals/dna_vault.dm
@@ -10,6 +10,7 @@
 #define VAULT_ARMOUR "Bone Reinforcement"
 #define VAULT_SPEED "Leg Muscle Stimulus"
 #define VAULT_QUICK "Arm Muscle Stimulus"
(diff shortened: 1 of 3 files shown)


MajManatee:
Id like it if it gave you a drop down, and let you pick from the available mutations rather than it being one of two

PR #20783, opened 2023-10-29
Removes the hallucination anomaly because it's either awful or completely harmless and neither is good

SomeguyManperson:
# Document the changes in your pull request

image

# Why is this good for the game?
I don't know about you but I like being able to play

# Testing
All references to the anomaly have been obliterated

# Spriting
sprites were removed

# Wiki Documentation

hallucination anomaly is gone

# Changelog

:cl:  
rscdel: hallucination anomaly has been removed since it removes the ability for anyone near it to play
/:cl:

PR #20842, opened 2023-11-06
Improve hallucination anomaly, make it less annoying and actually a lethal anomaly

warface1234455:
- The anomaly will no longer make you confuse, jitter, drowsy and dizzy. Instead it will give a status that once applied to you it will give 3 traumas that will go away after 3 minues if not near the anomaly, those are:
1. Stalking demon that will kill you
2. Something that will make things throw at you at extreme force
3. Split personality

- It also has a small chance to make you slip and stun you with some spooky sounds

- I also slightly buffed poltergeist by increasing the chance of getting hit by more objects (Will pick maximum of 5 items having highest throwforce)

- The stalker now get to you faster

- Added more spooky sounds to creepy sound list

Make it less annoying and actually a lethal anomaly, no harmless. Also prevent the anomaly from being removed in this pr #20783 

# Testing
Went near it and got some effects, almost died because of it

# Wiki Documentation
Probally document this in document of changes to anomaly page

# Changelog

:cl:  

tweak: Improve hallucination anomaly, make it less annoying and actually a lethal anomaly
tweak: I also slightly buffed poltergeist by increasing the chance of getting hit by more objects (Will pick maximum of 5 items having highest throwforce)
tweak: Added more spooky sounds to creepy sound list
tweak: The stalker now get to you faster
/:cl:


MajManatee:
A much better alternative IMO

00ze-cyclone:
That should go through mesons, an issue of this anomaly is how you can be completely immune unlike other anomalies.

ShoziOffical:
> That should go through mesons, an issue of this anomaly is how you can be completely immune unlike other anomalies.

yeah that actually is an issue

code/datums/status_effects/debuffs/debuffs.dm
```diff
@@ -334,6 +334,47 @@
 /datum/status_effect/pacify/on_remove()
 	REMOVE_TRAIT(owner, TRAIT_PACIFISM, "status_effect")
 
+/datum/status_effect/demonic_hallcuination
+	id = "demonic_hallcuination"
+	status_type = STATUS_EFFECT_REPLACE
```
Moltijoe:
```suggestion
	status_type = STATUS_EFFECT_REFRESH
```

code/datums/status_effects/debuffs/debuffs.dm
```diff
@@ -334,6 +334,47 @@
 /datum/status_effect/pacify/on_remove()
 	REMOVE_TRAIT(owner, TRAIT_PACIFISM, "status_effect")
 
+/datum/status_effect/demonic_hallcuination
+	id = "demonic_hallcuination"
```
Moltijoe: you've mispelled hallucination all over the place

Moltijoe (changes requested):

code/datums/status_effects/debuffs/debuffs.dm
ShoziOffical: guh

SomeguyManperson:
Anomalies generally
make their environment hazardous, without making the anomaly impossible to approach (mitigated, but not neutralized by equipment)
are hazardous to leave alone due to the damage they cause to their environment
cause some major localized damage or interference if they are allowed to elapse their entire timer
Generally, an anomaly might miss one or two of these, but still be significantly hazardous, i.e. the grav anomaly, which has no effect on elapsing its timer, or "major" damage, but pulls everything around it into a big fuckoff pile that is thrown at anyone who approaches

The hallucination anomaly instead
is impossible to "safely" approach, as its effect has no visual cues, if you are within 5 tiles of it you will likely be forced to deal with the brunt of its effects, with the only current counter being "if you wear mesons, it can't hurt you"
is perfectly safe to leave alone, since the effects only trigger on people who approach it, and produce no long term environmental damage
causes no major localized damage or interference on elapsing its timer, meaning the best way to handle it if you don't want to have hallucination reactive armor (which triggers the anomaly effects on nearby people when attacked) is to simply ignore it

ShoziOffical:
Generally, anomalies can be different whether it is harmless or not to people

Technically you can also be blind to get away from it

Moltijoe:
Balance Council:
>Make it 3 minutes instead of 5

ShoziOffical:
> Balance Council:
> 
> > Make it 3 minutes instead of 5

done

code/datums/brain_damage/magic.dm
```diff
@@ -33,17 +33,14 @@
...
-		var/obj/item/throwing
 		for(var/obj/item/I in view(5, get_turf(owner)))
 			if(I.anchored)
 				continue
 			if(I.throwforce > most_violent)
 				most_violent = I.throwforce
-				throwing = I
-		if(throwing)
-			throwing.throw_at(owner, 8, 2)
+				I.throw_at(owner, 8, 2)
```
Moltijoe: don't do it this way, if you do, it'll throw more than a single item

Moltijoe (changes requested):

code/datums/brain_damage/magic.dm
ShoziOffical: That's my point

Moltijoe: then make it truely random rather than a weirdly incrementing number

ShoziOffical: It's random to get by multiple objects nearby?

ShoziOffical: Making it more complex require more efforts

Moltijoe: no, it doesn't
this one would have it pick items in ascending force based on the order of the list returned by view()
if it immediately picks something with like 30 force, it'll only throw that
but if it somehow goes 1>2>3>4>....>29>30, then it would throw all 30 items
either make it a prob(#) or have it stop after it throws like 5 items or something

ShoziOffical: Aigh

ShoziOffical:
image
re-tested, only picked 5 items having the highest throwforce. There are 5 SORDs in the pic each one has like 1 throwforce while the pod plant has higher throwforce, resulted in 2 podplants and 3 SORDs got thrown at the felinid

iamthesilly:
> Will pick maximum of 5 items having highest throwforce

Walk into tool storage, die instantly

ShoziOffical:
> > Will pick maximum of 5 items having highest throwforce
> 
> Walk into tool storage, die instantly

your death isnt service related

Moltijoe:
closed, biome merged the better one

PR #20846, opened 2023-11-07
Deletes "How to suck blood 101"

Moltijoe:
reduce button bloat

# Testing (with and without stealthiness)
before
after

:cl:  
rscdel: Deletes "How to suck blood 101"
tweak: Adds vampire blood sucking guide to the antag menu
/:cl:


MajManatee:
Can vampires not stealth suck anymore?

Moltijoe:
> Can vampires not stealth suck anymore?

Not above a certain blood level, no
But theyve always been that way

MajManatee:
I think it would be better if the message didn't remove the stealth section, but just said how you couldn't stealth after 100u or whatever the value is. 
As it stands I can see people reading it, and trying to stealth suck past the limit (which isnt mentioned), looking at the help menu only to see the message has changed, then they mhelp saying "how stealth suck blood".

Moltijoe:
> I think it would be better if the message didn't remove the stealth section, but just said how you couldn't stealth after 100u or whatever the value is. As it stands I can see people reading it, and trying to stealth suck past the limit (which isnt mentioned), looking at the help menu only to see the message has changed, then they mhelp saying "how stealth suck blood".

I can add a line saying it will be lost after a point, and still remove the line

MajManatee:
I just don't like it being removed. Going to cause issues for no reason. The line, not the old button.

Moltijoe:
> I just don't like it being removed. Going to cause issues for no reason. The line, not the old button.

Thats already how it works

ToasterBiome (approved):

PR #20854, opened 2023-11-08
Plasma cutter projectile tweaks

SapphicOverload:

Plasma cutter projectiles now glow, and check against energy armor instead of bullet armor (why was it bullet armor??)

The glow looks cool, and I think it fits as an energy projectile a lot more than as a bullet projectile. I'm aware that energy armor is lower on most things, but considering the projectile only does 5 damage and fires very slowly, it shouldn't make any meaningful difference.

# Testing

plasma_cutter

# Changelog

:cl:  
tweak: plasma cutter projectiles glow and check against energy armor
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/special/plasma.dm
+++ b/code/modules/projectiles/projectile/special/plasma.dm
@@ -1,6 +1,7 @@
 /obj/projectile/plasma
 	name = "plasma blast"
 	icon_state = "plasmacutter"
+	armor_flag = ENERGY
 	damage_type = BRUTE
 	damage = 5
 	range = 4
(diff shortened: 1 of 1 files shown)


MajManatee:
If it doesn't make a meaningful difference then why do it?

iamthesilly:
> If it doesn't make a meaningful difference then why do it?

Already explained, because I think it makes more sense for a burst of plasma to be an energy projectile instead of a bullet.

ShoziOffical:
If it's energy armor check then shouldn't it be dealing burn damage instead of brute?

ToasterBiome (approved):

PR #20873, opened 2023-11-11
Refactors extended shock touch to actually just be a longer range shock touch

Moltijoe:
Code quality

# Testing
image

:cl:  
bugfix: Fixes extended shock touch not having a cooldown or sound effect
/:cl:


MajManatee (approved):
Might allow you to remotely defib someone like the normal touch does.

Moltijoe:
> Might allow you to remotely defib someone like the normal touch does.

any ranged shock works that way, that's how extended shock touch works

MajManatee:
well its just kind of how ALL shocks work, but I meant how the normal one had a do after (I think) for it.

Moltijoe:
> well its just kind of how ALL shocks work, but I meant how the normal one had a do after (I think) for it.

no, you're thinking about using the hand for revival surgery, which is far more reliable, the regular one doesn't otherwise have a do_after

PR #20881, opened 2023-11-12
Removes the obsolete healing surgeries

Moltijoe:
With the existence of tend wounds, these surgeries just simply are never worth doing, and are potentially a noob trap
They take far more steps, consume resources, and are simply slower than basically any other form of healing

note, that the burn heal surgery being removed is NOT the same as the burn wound healing surgery, despite both using the word debride at some point in their code

Code cleanup

:cl:  
rscdel: Removes the obsolete healing surgeries 
/:cl:


Yarinoi:
the ones that appear when you target a limb that isnt the chest?

Moltijoe:
> the ones that appear when you target a limb that isnt the chest?

it also appears on the chest, but it always appears last

ehj236:
Muda muda da

MajManatee (approved):
They fuckin sucked anyways.

PR #20885, opened 2023-11-13
Increases the cost of several changeling abilities

SomeguyManperson:
# Document the changes in your pull request

you think it would be funny if we just straight buffed literally everything in the changeling kit without actually knowing how strong the changeling kit is

I have changed these at near random based on how annoying the ability is to deal with

# Why is this good for the game?
maybe potentially slightly reduces the amount of gaming changelings can do for free

# Testing
none of the numbers are a zero

# Wiki Documentation

changelog has exact number values and ability names

# Changelog

:cl:  
tweak: a number of changeling abilities have had their chemical costs increased to balance out a chemical maximum and regeneration speed buff
tweak: adrenaline now costs 50 chemicals up from 30
tweak: fleshmend now costs 40 chemicals up from 20
tweak: panacea now costs 30 chemicals up from 20
tweak: resonant shriek now costs 60 chemicals up from 20
tweak: dissonant shriek now costs 40 chemicals up from 20
tweak: spawning spiders now costs 90 chemicals up from 45, the "no fun allowed" chanting can start here
tweak: mute sting now costs 40 chemicals up from 20
tweak: blind sting now costs 40 chemicals up from 25
tweak: cryosting now costs 30 chemicals up from 15
/:cl:

PR #20887, opened 2023-11-13
Nerfing the actual problem: Changeling Chem Total & Chem Regen

Therandomhoboo:

This is running against the current pr that is open which is #20885 . This is made as an alternative to increasing the Changeling Abilities and therefore making Changelings only able to use one or two abilities before fleeing. This PR fixes the actual problem, Chem total and Chem regen. 

When Changelings were buff their Total was 150 and Regen was 5, this was nerfed down to 125 and Regen 2. This now decreases it to 110 and 1.25 regen speed which provides a significant slap to the changelings. This is better because it means that the numbers for the abilities aren't just randomly chosen based on personal annoyance and that Changelings can use their abilities more than once.

As of importance, if the other pr gets merged obviously close this one since it will ruin Changelings faster than my faith that there will be no cars on the road.

Slaps Changelings with a nerf stick whilst also letting them use their abilities since it is their entire point. Stop forcing people to play stealth when they don't have to/want to.

# Testing

If abilities are spammed, it takes a longer time than its previously did. Also they don't have a massive chem pot.

# Changelog

:cl:  

tweak: Changeling Chem Total decreased from 125 to 110, Chem regen speed reduced from 2 to 1.25

/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/changeling.dm
+++ b/code/modules/antagonists/changeling/changeling.dm
@@ -27,7 +27,7 @@
 	var/trueabsorbs = 0//dna gained using absorb, not dna sting
 	var/chem_charges = 50 // chems we have on spawn
 	var/chem_storage = 110 // max chems
-	var/chem_recharge_rate = 1.5 // how fast we restore chems
+	var/chem_recharge_rate = 1.25 // how fast we restore chems
 	var/chem_recharge_slowdown = 0 // how much is our chem restore rate hampered (keep at 0)
 	var/sting_range = 2
 	var/changelingID = "Changeling"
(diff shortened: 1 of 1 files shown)


DeadlyBag:
I'll be real, feels like the numbers decreases were thrown at a dartboard and randomly picked. Rather these be tested in game with other players (unless done already) than just merged in, then we have to go through 1000 more prs to refix the numbers.

Therandomhoboo:
> I'll be real, feels like the numbers decreases were thrown at a dartboard and randomly picked. Rather these be tested in game with other players (unless done already) than just merged in, then we have to go through 1000 more prs to refix the numbers.

"If abilities are spammed, it takes a longer time than its previously did. Also they don't have a massive chem pot." awkward

Moltijoe:
Closed in favour of the alternative

PR #20888, opened 2023-11-13
Reinforced Windows buff. No more easy breaking,

Therandomhoboo:

Yes, I am the new Gus and Simple Horror, fear me. In seriousness though, chairs shouldn't really be breaking reinforced windows with ease. There is a line in R.Windows that is meant to reflect damage that are 10 or lower (If i'm reading it right) but instead chairs and anything that does 10 damage can bypass that and do damage to the windows offsetting the point of it. This makes it so it actually works now.

Stops reinforced windows from being fairly weak and makes Perma a bit harder to break out instead of find stool, hit window and leave.

# Testing
Anything 10 or less damage doesn't break the window.

# Changelog

:cl:  

tweak: Reinforced Window Damage deflection is now 11 instead of 10

/:cl:


azzzertyy:
downvoters didnt read the pr

Moltijoe (approved):

PR #20895, opened 2023-11-14
[Box & Asteroid] Change Comms Maint to Comms Storage

Aquizit:

Taking the middle room in Box and Asteroid comms and changing it from maint to uh -- not-maint. Also corrected the area swap that was on Asteroid. Also makes it so clockies can't warp into that little area since they're not supposed to be able to warp into comms area at all.

It's not maintenance it shouldn't be treated as such. Make network admin run during radiation storms like everyone else. Also protects comms against clockies just a bit more.

# Testing

Tested on both maps. Works as intended.

# Spriting

Changed an area name/icon to match the new area. Doesn't affect anyone but mappers.

# Changelog

:cl:  
mapping: Removed comms /maintenance/ area and icon
mapping: Added /tcommsat/storage area and icon
mapping: Replaced maint with storage on Box & Asteroid
mapping: Fixed area swap on Asteroid
/:cl:


cuackles (approved):

neeshacark (approved):
ye this is all good b

PR #20904, opened 2023-11-16
Glizzy' Glycerol gets some attention! (Glycerol Yield increased by 2)

Therandomhoboo:

We shouldn't be making everything CBT by just increasing the "making process", it is weird that glycerol has this low yield factor despite it only being used in one thing. Whilst it is used to make an explosive, we have so much better items that do explosions better that I think its fine to ease up on how tedious it is to make the chem.

I have changed the yield 3 instead of 1, because the process to make Nitro is already a bitch enough as it is. If not, then I will blame the people who abused this and be verrrrry disappointed. 

Encourages some variety, makes it so a chem is more used and lessens the CBT we have of making everything take 20x longer.

# Testing

Its one line code, I'll happily stop coding if it breaks the server.

# Wiki Documentation

Results for glycerol is now 3 instead of 1

# Changelog

:cl:  

tweak: Glycerol making results in 3 units instead of 1

/:cl:


MajManatee (approved):
number change.
Still takes corn so it will still be uncommon, but it would be nice

PR #20906, opened 2023-11-17
Death Alarm Implants

Runian:
A crate containing 5 death alarm implants are available to be purchased at cargo under the medical section for 8,000 credits. It requires CMO access to open.

Death alarms:
- Trigger when you die or hit with an EMP of any severity.
- On trigger, tries to broadcasts your name and location onto the Medical channel.
- EMPs have a 50% chance to cause the location to be randomized to a tracking/teleport beacon instead when triggered this way.
- Disappears when triggered (one-time use).

Example when Moja dies in Cargo:
image

Gives cargo something spend their money on. Gives a reason for medical to use NT IRN and use medical department budget. Gives people comfort knowing that they won't be completely forgotten in the middle of the round if they suddenly die or disappear. Promotes kidnapping instead of instantly killing people who use this implant.

# Testing
Purchased the crate: it starts locked and only (un)locks to CMO access. 
Killing Moja with an energy axewhile they have a death implant properly sends a message to Medical.

# Changelog
:cl:  
rscadd: Death alarm implants. They broadcast your location when you die or get EMP'd.
rscadd: A crate containing 5 death alarm implants can be purchased under the medical section. It costs 8,000 credits and requires CMO access.
/:cl:

### DIFF
--- a/code/game/objects/items/implants/implant_deathalarm.dm
+++ b/code/game/objects/items/implants/implant_deathalarm.dm
@@ -1,0 +1,53 @@
+/obj/item/implant/deathalarm
+	name = "death alarm implant"
+	desc = "Monitors host vital signs and transmits a radio message upon death."
+	actions_types = null
+	var/obj/item/radio/radio
+
+/obj/item/implant/deathalarm/Initialize(mapload)
+	. = ..()
+	radio = new(src)
+	radio.keyslot = new /obj/item/encryptionkey/headset_med // Medical only should know if people die (because paramedics). Consider changing to Command-only if something like Blueshield is added. 
+	radio.listening = FALSE
+	radio.recalculateChannels()
+
+/obj/item/implant/deathalarm/activate(cause)
+	if(!imp_in)
+		return FALSE
+
+	// Location.
+	var/area/turf = get_area(imp_in)
+	// Name of implant user.
+	var/mobname = imp_in.name
+	// What is to be said.
+	var/message = "[mobname] has died-zzzzt in-in-in..." // Default message for unexpected causes.
+	switch(cause)
+		if("death")
+			message = "[mobname] has died in [turf.name]!"
+		if("emp")
+			if(isemptylist(GLOB.teleportlocs) || prob(50)) // 50% chance of broadcasting as if they died (or no tracking beacons available).
+				message = "[mobname] has died in [turf.name]!"
+			else // Otherwise, random location.
+				message = "[mobname] has died in [pick(GLOB.teleportlocs)]!"
+
+	name = "[mobname]'s Death Alarm"
+	radio.talk_into(src, message, RADIO_CHANNEL_MEDICAL)
+	qdel(src) // One-time use implant. Use it and lose it.
+
+/obj/item/implant/deathalarm/on_mob_death(mob/living/L, gibbed)
+	if(gibbed)
+		activate("gibbed") // Will use default message.
+	else
+		activate("death")
+
+/obj/item/implant/deathalarm/emp_act(severity)
+	activate("emp")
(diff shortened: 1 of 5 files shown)


wonderinghost:
I kinda want to do a survey to see if people know nt intern exists.

Or if they know why don't they install it
(I know my answer to why I don't use it)

iamthesilly:
> Promotes kidnapping instead of instantly killing people who use this implant.

Problem: if you don't disable their radio they will scream over comms immediately, and the most accessible way to disable radios is to use an EMP, which broadcasts their location anyway.

wonderinghost:
Doesn't tracking implants already do this?

solwashere:
Making this tied to beacons makes no sense
Just make it tied to the room system that mapping uses or whatever
I highly doubt this'll get through but I applaud the effort either way

ynot01:
Defensive measures are already incredibly strong (xenobio, mining, simply standing near your metafriends)

This is simply unnecessary and unfit for our server

Scrambledeggs00:
Probably a good idea to see if the signal jammer and NTSL's sig.pass can block the message. If not then yea this would make it nigh impossible to kidnap someone unless they're the mime. if you don't know how to make a script that does it then here is one:
def process_signal(sig){
sig.pass = 0;
return sig;}
It would probably also be a good idea to give it a visual indicator that doesn't require a medhud since without one most people probably won't know they have one and so negate the whole "Promotes kidnapping instead of instantly killing people who use this implant".

Yarinoi:
this seems way too powerful, especially with it triggering on being emped with any severity too
> Gives people comfort knowing that they won't be completely forgotten in the middle of the round if they suddenly die or disappear

gee, if only there was a way for medical to see your location and health at all times, perhaps on some sort of menu, so that even if you drop off of it, a sharp-eyed medical personnel could use your last broadcasted location to find you. its such a shame that such a thing doesnt exist..

azzzertyy:
why do we need another non-avoidable way of alerting people to your deaths, when if anything we need less.
medical tracking is op, and has been for years, and no one has done anything about it and its only gotten worse since theyve been given trackers on their pdas. i am vehemently against this being added, as it has the following issues:
1. Unnecessary. We already have tracking implants and suit sensors and bodycameras (all of which are powerful)
2. You can already make this. Radio Announcer (gotten from maints, fairly common) + Health Analyzer
3. You can already make this without a radio announcer. Remote signaler + Health Analyzer, Remote Signaler on a radios microphone wire next to a repeating holopad recording announcing your death.
4. We need less of this. Not more.

Therandomhoboo:
Don't need more tracking bloat. Already several ways to track them and with most people maxxing sensors (unless they're antag or chad) this is just overkill. If sensors and other tracking ways was very less accurate then I'd be fine with this but as we currently stand this will just make Antag lives a lot more harder.

MajManatee:
1.) Nanites can already do this but better
2.) Validhunting bad
3.) Forcing antags to pay the EMP tax is arguably worse.

ofscarletdreams:
make heavy emps turn it off entirely for like 2 minutes instead of misfiring it (ample time for a tussle)

ofscarletdreams:
you could also have a deluxe version the cmo gets 3 of (you cant get any more) that alerts security on top of medical

LassiH:
make it nukie exclusive and make it tell only other implant havers if they have died

Moltijoe:
Balance council:
>Medical is already oppressive enough to antags, no need to add yet another thing to help them,
even if emps are an unreliable counter, adding yet another thing that requires emping is annoying

on a side note, it might be accepted if it was nukies only (and probably free)

PR #20913, opened 2023-11-18
engineers are no longer department specific

wonderinghost:

gets rid of department specific engineers. giving them access to most departments (wont be afraid to shave off more access if its a problem)
image

adds encryption keys and armbands to each head locker (and qm) for heads that still want a dedicated engineer

i think it would be a good thing for engineers and other departments that want to do something. allows engineers ease of access to go in and do repairs or vent departments. lets people call on any or more engineers to work on projects if need be.

only problem with this is from the current state of yogs economy. as you are making shit free in the vending machines. 

however if we have this discussion now about yogs economy or just this idea then we have a new bench, might also specter this until times change or someone comes up with a better idea

# Testing
checked to see if things update

# Wiki Documentation

any references to department specific engineers if on there

# Changelog

:cl:  

tweak: removes department specific engineers to have engineers with some access to those departments
/:cl:


SapphicOverload (changes requested):
Should probably split decluttering the CE's locker into a separate PR

wonderinghost:
aight, might also shave bar access also

JohnFulpWillard:
WTF

Therandomhoboo:
This just literally goes back to the Jani-problem where they got given all basic access, then lost a chunk of basic access because people found out you can literally get deeper into any department due to the access SH gave them. Its fine to still have dedicated Engineers, nothing has warranted this pr.

Also, there isn't an actual strong enough reason for this and this was the reasoning.

> i think it would be a good thing for engineers and other departments that want to do something. allows engineers ease of access to go in and do repairs or vent departments. lets people call on any or more engineers to work on projects if need be.

Only time I've seen engineers not get into an area is low-pop, if that is the problem give them more access as skeleton crew if they've not got more. Med/High-pop rounds you will usually have someone in every department, Heads, Borgs or an AI. I don't see the actual need to give them it and saying for "ease of access" or for "Projects" isn't that good honestly.

SomeguyManperson:
Engineers already have AA when working on projects, its called "hey, I'm here to help you with that thing, can you let me in?"

Oh and the toolbelt. For emergencies.

00ze-cyclone:
Engineers already have an impressive amount of access and if they need more they can just ask the HoP, I don’t want engineers to become even more the legal enabled greytide role.

azzzertyy:
Jaws Of Life exist already.

neeshacark:
Not sure about this, much for the same reasons Hobo gave.

If this is a huge issue then it might be better to do something like let engineers take two departments as the ones they cover instead of just one (this might actually make all departments have a dedicated engi at all times)

Another option would be to make certain doors work with any one access and add engi to them, instead of giving engis access to specific areas.  This way mappers could restrict engis from getting to areas that are 'too deep'.  It's easier to pick and choose this way since the way access is mapped is inconsistent from map to map and sometimes done poorly on any one map.

RG4ORDR:
Giving engineers borderline AA is going to turn them into heavy targets for antags - engineering IDs already give you access to Airlocks,Maint,Power,Tech storage,Engineering(ofc). 
Engineering lathe - Jaws of Life.
Insuls, and generally getting over looked for having specific things - so tack on having pseudo AA means you're gonna have a problem.

Moltijoe:
Balance Council:
>Engineering players don't need more access, they're already a powergamer and antag farmer favourite
if they need to get into a department, they already have access to hacking tools, jaws of life, and could always just ask for access

PR #20917, opened 2023-11-18
Adds a preference to the cybernetic organ quirk

SapphicOverload:

Cybernetic organ quirk can now choose which organ they get, the same way the prosthetic limb quirk works. The choiced versions give standard non-upgraded cybernetics and are neutral (0 points).  Also refactors it to be less bad.

Improves character customization, allows you have a character that has a specific cybernetic organ.

# Testing

image
image

# Wiki Documentation

Add the above quirks to the list of quirks.

# Changelog

:cl:  
rscadd: added preference for non-upgraded cybernetic organs as quirk options
tweak: random cybernetic organ quirk costs 3 instead of 4
/:cl:

### DIFF
--- a/code/datums/traits/good.dm
+++ b/code/datums/traits/good.dm
@@ -315,33 +315,32 @@
 	icon = "building-ngo"
 	value = 4
 	var/slot_string = "organ"
-	var/list/organ_list = list(ORGAN_SLOT_LUNGS, ORGAN_SLOT_HEART, ORGAN_SLOT_LIVER)
+	var/list/organ_list = list(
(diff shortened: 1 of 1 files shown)


Yarinoi:
i think it should cost more to be able to pick exactly what organ you want, maybe set the price to 5 or 6?

iamthesilly:
> i think it should cost more to be able to pick exactly what organ you want, maybe set the price to 5 or 6?

Honestly I don't think the random one was even worth 4 points to begin with, since all the organs give you the severe downside of being killed by EMPs.

MajManatee:
I reckon you could safely lower the cost of the random one by at least one

Moltijoe (approved):

PR #20933, opened 2023-11-20
Preternis chemical purge rework

SapphicOverload:

Preternis now purge 4% of the volume of existing chems each tick instead of deleting everything that has been processing for 20 ticks.

This means preternis are no longer functionally immune to morphine and certain other chemicals just because their effects only happen after a certain number of ticks. Now it's possible to make them experience those effects by giving them too much for them to purge it in time.

Makes preternis no longer immune to certain chems they probably weren't supposed to be immune to.

# Testing

Asleep after drinking a bottle of morphine
image

# Wiki Documentation
The wiki page for preternis doesn't even mention the chemical purge, so there should probably be a mention of it in the strengths and weaknesses section.

# Changelog

:cl:  
tweak: preternis now purge 4% of internal chemicals every tick, instead of purging everything after 20 ticks
tweak: preternis are no longer immune to the effects of morphine and a few other chemicals
/:cl:


iloveloopers:
Nitrium buff

MajManatee (approved):
Looks fine. Weird that the linter is having a mapping hissy over it though.

PR #20940, opened 2023-11-20
Station side legion core soft-rework

Moltijoe:
Adding a delay to use, while decent, didn't weaken implanting at all, and adding a second check for implant abuse prevention seems hacky
instead i'm making the effect itself weaker when used outside of a mining level, and repeated use makes it worse, to the point of being useless

Nuke their abuse station side, 
first one will heal decently reduced amounts
second onward won't cleanse CC or reset bodytemp and heal well under half the original amount
third onward won't do anything

also applies a slowdown, either after using two, or using one and waiting a bit

NOTE: THIS DOESN'T AFFECT MINING LEVEL USE AT ALL

# Testing
image

:cl:  
tweak: Legion cores no longer take extra time to use on the station
tweak: Legion cores now progressively heal less (and eventually damage) as they are used station side
tweak: Overusing Legion cores station side now applies a slowdown during the effect
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs/buffs.dm
+++ b/code/datums/status_effects/buffs/buffs.dm
@@ -546,7 +546,7 @@
 /datum/status_effect/regenerative_core/refresh(effect, ...) //apply the buffs and heals again when it's refreshed
 	var/efficiency = 1
 	if(!mining_check()) //only get weaker station side
-		debuffCounter += 25
+		debuffCounter += 30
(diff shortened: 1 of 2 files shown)


cowbot92:
Were in-hand cores a problem recently? This one feels like a change from out of left field

Moltijoe:
> Were in-hand cores a problem recently? This one feels like a change from out of left field

no, this just makes the downside consistent rather than one causing a delay, and the other exploding your face

Moltijoe:
hmm, ain't feeling this

PR #20946, opened 2023-11-21
Too Many Cooks: Reduces amount of cook job slots to 1

neeshacark:

does what it says in the title, was 2 before.

This was spurred on by a post on r/ss13 (I know im a heretic for using the reddit)

Quoting from that post "When I'm in my kitchen I want to be alone. I want to chose what I want to cook, do fun gimmicks. I don't like when there is a lizard butchering a monkey in the middle of the kitchen to make some fried livers. I don't like when the other cook takes the already limited ingredients. And please do not throw the food on the floor..."

This sums up pretty much how I feel about playing cook - there is no need to have 2.  One cook is more the capable of supplying all the food the station needs.  This is just my take though, putting it out there for yoggites to decide what they want.

# Changelog

:cl:  cark
tweak: reduced total cook job slots to 1
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/cook.dm
+++ b/code/modules/jobs/job_types/cook.dm
@@ -4,7 +4,7 @@
 	orbit_icon = "utensils"
 	department_head = list("Head of Personnel")
 	faction = "Station"
-	total_positions = 2
+	total_positions = 1
 	spawn_positions = 1
 	supervisors = "the head of personnel"
 	selection_color = "#bbe291"
--- a/config/jobs.txt
+++ b/config/jobs.txt
@@ -16,7 +16,7 @@ Cargo Technician=2,1
 Shaft Miner=3,3
 
 Bartender=1,1
-Cook=2,1
+Cook=1,1
 Botanist=3,2
 Janitor=2,1
 


Scrambledeggs00:
My best cook rounds have been from having another cook because of how we deal and eventually work with each other. This would forever end the great cook partnerships that can and do form ;[

N3D6:
I’m 50/50 on this. I hate being chef and having a partner who: 
A. Wastes all the supplies and doesn’t really do anything of worth
B. Does their own gimmick and ignores you
But also having another chef partner to mess around with is fun as hell when you’re working with the same gimmick. I think making it 1 only is jumping the gun a bit

Addust:
image
image

Aquizit:
Make the kitchen bigger. My displeasure from having a second chef with me is purely because we're in each other's way near constantly. (I don't play cook enough to map a bigger kitchen myself though.)

This is probably more necessary or warranted now with so many more cooking machines taking up space.

neeshacark:
> Make the kitchen bigger. My displeasure from having a second chef with me is purely because we're in each other's way near constantly. (I don't play cook enough to map a bigger kitchen myself though.)
> 
> This is probably more necessary or warranted now with so many more cooking machines taking up space.

i considered this but it would be SO much work if i were to do it without nuking a lot of the kitchen templates

solwashere:
we should always listen to /tg/ players complaining on reddit I think

in all seriousness this kinda prevents any chance of having fun chef to chef interactions, like trying to do a gimmick together, and cuts back on interaction
I think that expanding the kitchen would probably be a better solution in the end, like Aquizit said

neeshacark:
> we should always listen to /tg/ players complaining on reddit I think
> 
> in all seriousness this kinda prevents any chance of having fun chef to chef interactions, like trying to do a gimmick together, and cuts back on interaction I think that expanding the kitchen would probably be a better solution in the end, like Aquizit said

ok but im nuking the majority of the templates

PR #20951, opened 2023-11-22
auxiliary armory - security now has access

ynot01:

Security now gets access to the auxiliary armory from the actual armory

image

Gax and Donut do not have an auxiliary armory area, so changes are not needed there

Some important things are trapped in the 'armory' like the firing pins, barrier grenades, etc that security cannot access even when armory is open

This sets the door between the armory and auxiliary armory to be security access, so that it can be accessed freely when the armory is available to security and no holds are barred

# Changelog

:cl:   
tweak: The door between the armory and the auxiliary armory are now default security access
/:cl:

### DIFF
--- a/_maps/map_files/AsteroidStation/AsteroidStation.dmm
+++ b/_maps/map_files/AsteroidStation/AsteroidStation.dmm
@@ -12134,7 +12134,7 @@
 "dcK" = (
 /obj/machinery/door/airlock/security/glass{
 	name = "Armoury";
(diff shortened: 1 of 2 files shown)


Therandomhoboo:
So, I'm on the fence. 
On one hand, its not that drastic getting that access but also it takes away wardens already small role which I assume is distributing the equipment out for use.

Aquizit (approved):

PR #20952, opened 2023-11-22
Fix XL beakers not showing reagent fill state

hedgehog1029:

Fix XL beakers not updating their appearance to show reagent fill level.

# Testing

~~I haven't tested this yet I did this fix in-browser~~

Tested as fixed

# Changelog

:cl:    
bugfix: XL beakers show contained chemicals again  
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/glass.dm
+++ b/code/modules/reagents/reagent_containers/glass.dm
@@ -123,28 +123,33 @@
 /obj/item/reagent_containers/glass/beaker/on_reagent_change(changetype)
 	update_appearance(UPDATE_ICON)
 
-/obj/item/reagent_containers/glass/beaker/update_overlays()
+/// Update beaker overlay with current fill state
+/obj/item/reagent_containers/glass/beaker/update_overlays(fill_state as text|null)
 	. = ..()
 	if(!reagents.total_volume)
 		return
-	var/mutable_appearance/filling = mutable_appearance('icons/obj/reagentfillings.dmi', "[icon_state]10")
+
+	if (!fill_state)
+		fill_state = icon_state
+
+	var/mutable_appearance/filling = mutable_appearance('icons/obj/reagentfillings.dmi', "[fill_state]10")
 
 	var/percent = round((reagents.total_volume / volume) * 100)
 	switch(percent)
 		if(0 to 9)
-			filling.icon_state = "[icon_state]-10"
+			filling.icon_state = "[fill_state]-10"
 		if(10 to 24)
-			filling.icon_state = "[icon_state]10"
+			filling.icon_state = "[fill_state]10"
 		if(25 to 49)
-			filling.icon_state = "[icon_state]25"
+			filling.icon_state = "[fill_state]25"
 		if(50 to 74)
-			filling.icon_state = "[icon_state]50"
+			filling.icon_state = "[fill_state]50"
 		if(75 to 79)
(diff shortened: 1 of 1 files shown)


ToasterBiome (approved):

ToasterBiome:
can u test

hedgehog1029:
it works
image

code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -173,7 +173,7 @@
 	amount_per_transfer_from_this = 10
 	possible_transfer_amounts = list(5,10,15,20,25,30,60,120)
 
-/obj/item/reagent_containers/glass/beaker/plastic/update_icon_state()
+/obj/item/reagent_containers/glass/beaker/plastic/update_overlays()
```
JohnFulpWillard: This should be on ``update_icon`` instead i think, this isn't changing overlays

hedgehog1029: it is, it's calling the super `/obj/item/reagent_containers/glass/beaker/update_overlays()` which is responsible for actually updating overlays. this function temporarily overrides the icon_state but then sets it back to its normal value, so after the function call is done the state has not actually changed.

JohnFulpWillard: That proc is actually adding an overlay onto the beaker, this is not, so it should be on ``update_icon``, the proc that calls ``update_overlays``.

https://github.com/yogstation13/Yogstation/assets/53777086/f51ca596-b485-496c-990d-45197bf2c59d

hedgehog1029:
Refactored to remove the hacky icon_state overriding, tested as working

code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -123,28 +123,33 @@
 /obj/item/reagent_containers/glass/beaker/on_reagent_change(changetype)
 	update_appearance(UPDATE_ICON)
 
-/obj/item/reagent_containers/glass/beaker/update_overlays()
+/// Update beaker overlay with current fill state
+/obj/item/reagent_containers/glass/beaker/update_overlays(fill_state as text|null)
```
JohnFulpWillard:
```suggestion
/obj/item/reagent_containers/glass/beaker/update_overlays()
```

code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -123,28 +123,33 @@
...
+	if (!fill_state)
+		fill_state = icon_state
+
+	var/mutable_appearance/filling = mutable_appearance('icons/obj/reagentfillings.dmi', "[fill_state]10")
 
 	var/percent = round((reagents.total_volume / volume) * 100)
 	switch(percent)
 		if(0 to 9)
-			filling.icon_state = "[icon_state]-10"
+			filling.icon_state = "[fill_state]-10"
```
JohnFulpWillard:
```suggestion
			filling.icon_state = "[base_icon_state]-10"
```
This for the rest too

JohnFulpWillard (dismissed):
Use ``base_icon_state`` instead then, we should not be touching args

you can check https://github.com/yogstation13/Yogstation/pull/21011 to see how to properly use ``base_icon_state`` if you want (as a var and implemented in ``update_icon_state``).

hedgehog1029:
alright attempt 3

ps it still works
image

ToasterBiome (approved):

PR #20957, opened 2023-11-23
Makes Lasombra actually do what it says it does on the wiki

Moltijoe:
:cl:  
bugfix: Lasombra clan now actually gives brute immunity and 2x burn mod
/:cl:


code/modules/antagonists/bloodsuckers/structures/bloodsucker_life.dm
```diff
@@ -445,8 +445,9 @@
 	var/mob/living/carbon/human/bloodsucker = owner.current
 	owner.current.grab_ghost()
 	to_chat(owner.current, span_warning("You have recovered from Torpor."))
-	bloodsucker.physiology.brute_mod = initial(bloodsucker.physiology.brute_mod)
-	bloodsucker.physiology.burn_mod = initial(bloodsucker.physiology.brute_mod)
+	if(my_clan != CLAN_LASOMBRA)
```
tattax, line 448: if you are gonna do this, use clan datums checks, i.e BLOODSUCKER_CONTROL_SHADOWS .
also all changes were made deliberately in the original PR.
i'm not sure were people got the ideas it was a bug (i actually do it is just that i'm very petty when it comes to making changelogs) but there's enough info to deduce that it was deliberate (i changed the clan description and what not, the wiki part ON my pr did mention it). 

same goes for the gangrel one (nothing against fixing up the radial menu and whatnot that isn't related to that) but all these are balance tweaks and not actual bug fixes. it was all intentional, i even compensated for lasombra and gangrel for losing it.

the truth is i was gonna add more stuff but shit got in the way and i just never came back around to it. i'd much like you rework how things work right now rather than readd old stuff which i think was either pretty dumb or overpowered.

Moltijoe: yeah, fair, i was asked to "fix" this because people thought it was a bug

PR #20970, opened 2023-11-25
Only Delta alert gets red lights

Moltijoe:
In theory it was a good idea, in practice, everything being red during an entire gamemode is terrible for gameplay (see virtual boy)
it's made worse when one of the gamemodes that's affected is nukies, where all the antags are already red

Don't even get me started on colourblind people

Also it ruins the ambience of epsilon alert which is admin only, if they want everything to be red, they can set it to be red

:cl:  
tweak: Only Delta alert gets red lights
/:cl:


Hislyna:
I learnt from the old syndie station that red all round is just ooooooooooooooooooooooooooooooooooooooooooooooooof. Good for small times like delta yea, but otherwise like ye said oooooooooof

ynot01 (approved):

PR #20972, opened 2023-11-25
Gives more skeleton access to the paramedic type jobs

Moltijoe:
Any time it's skeleton crew, there's a good chance of there being literally only one of the paramedic type jobs
This gives mining medic and brig physician the full capabilities of a paramedic during skeleton crew
This way they're able to retrieve their dead charges from mostly anywhere in the station, that would otherwise be handled by the paramedic that doesn't exist

Also gives paramedic surgery access during skeleton crew because it's not uncommon they're the only medical staff

:cl:  
tweak: Brig physician also gets cloning and airlock access during skeleton crew
tweak: Paramedic also gets surgery access during skeleton crew
tweak: Mining medic also gets cloning, maint, and airlock access during skeleton crew
/:cl:

### DIFF
--- a/yogstation/code/modules/jobs/job_types/brig_physician.dm
+++ b/yogstation/code/modules/jobs/job_types/brig_physician.dm
@@ -23,7 +23,7 @@
 		/datum/job_department/security,
 	)
 
-	added_access = list(ACCESS_SURGERY)
+	added_access = list(ACCESS_SURGERY, ACCESS_CLONING, ACCESS_EXTERNAL_AIRLOCKS)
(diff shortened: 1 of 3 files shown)


JohnFulpWillard:
image

MajManatee:
I'm more of the thought that mining medics are miners that can do some medical, not doctors that could mine if they feel like it.
Brig phys imo doesn't need external access either since they are essentially just the  security doctor, and since normal doctors never get external airlock access either. Regardless, they can both use the arrivals and evac airlocks if they need to get a body anyways.

Moltijoe:
> I'm more of the thought that mining medics are miners that can do some medical, not doctors that could mine if they feel like it. Brig phys imo doesn't need external access either since they are essentially just the security doctor, and since normal doctors never get external airlock access either. Regardless, they can both use the arrivals and evac airlocks if they need to get a body anyways.

mining medics should not be choosing mining medic because they want to be a miner with access to medical stuff, they should be discouraged from mining unless they absolutely have to (no miners)

ToasterBiome (approved):

PR #20973, opened 2023-11-25
Gives warden brig phys access during skeleton crew

Moltijoe:
Warden should be able to fix up inmates if there's no brig phys

FOR SOME REASON, THE WINDOORS USE MEDICAL ACCESS NOT BRIG PHYS ACCESS

:cl:  
tweak: Gives warden brig phys access during skeleton crew
/:cl:


MajManatee (approved):
I mean really they should be taking people to med IMO instead of using their medroom if there's no brig phys ya know? Hes not a doctor. Hes a warden. he doesn't have a degree or nothing.
That being said its a good enough idea.

PR #20978, opened 2023-11-26
Cyborgs Can AltClick to use Hacked Shock Paddles's Disarm instead of Harm

Runian:
Cyborgs can alt-click their shock paddles to toggle whether they will do disarm actions with it instead of harm actions.

Oversight. If a coder can forget that drones and borgs don't have access to disarm intent (cough), then it is not a surprise if it was the same here. Probably qualifies as a bugfix, but tweak is likely more accurate.

# Testing
I alt-click. It works. I click Moja. It works.

# Wiki Documentation
There is likely a tip in cyborg's wiki that talks about the defib. Mention this along side it.

# Changelog
:cl:  
tweak: Hacked shock paddles can be alt-clicked by cyborgs to switch between using disarm and harm actions.
/:cl:

### DIFF
--- a/code/game/objects/items/defib.dm
+++ b/code/game/objects/items/defib.dm
@@ -16,13 +16,18 @@
 	actions_types = list(/datum/action/item_action/toggle_paddles)
 	armor = list(MELEE = 0, BULLET = 0, LASER = 0, ENERGY = 0, BOMB = 0, BIO = 0, RAD = 0, FIRE = 50, ACID = 50)
 
-	var/on = FALSE //if the paddles are equipped (1) or on the defib (0)
-	var/safety = TRUE //if you can zap people with the defibs on harm mode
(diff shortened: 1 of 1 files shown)


iamthesilly:
You're aware the disarm does 50 stamina damage and 10 seconds of knockdown and can't be blocked or countered in any way, right?

Runian:
> You're aware the disarm does 50 stamina damage and 10 seconds of knockdown and can't be blocked or countered in any way, right?

Ya? At least it is no longer the stunbaton one-hit stuns it once was.

MajManatee:
Didn't we remove flashes from borgs for a similar reason?

Moltijoe:
Balance Council:
> Peacekeeper borgs are bad enough, we don't need another borg module being able uncounterably to one click stun people

PR #20980, opened 2023-11-27
"Small" tweak to antag kill objectives

Moltijoe:
If we're pushing for mrp, assassinate and maroon requiring round removal hurts the ability to not sweat as traitor.
Any sufficiently capable player is able to metagame which type of kill objective the antag has and actively try to deny it if they're a shitty person
it being possible to "uncomplete" antag objectives forces them to sweat the entire round, even once they've completed their objectives
The only exception being grand theft because it's not playing around a player's life like the assassination objectives do

This should give traitors more room to mess around once their objectives are completed, without fear of powergamers seeking out their redtext

EDIT: aight, i guess regular maroon and assassinate come back

# Testing
it says completed because i'm still on the station (marooned)
image

# Wiki Documentation

:cl:  
rscadd: Added new maroon organ objective for traitors to roll
tweak: Traitor break machinery objective only requires breaking the original machines
tweak: Traitor break machinery targets at max 2 areas instead of up to 4
/:cl:


neeshacark:
awww why no supermatter target? blowing up the sm is fun!

Moltijoe:
> awww why no supermatter target? blowing up the sm is fun!

fine, i've returned that

Beeper-Skeeper:
HOLY SHIT, A SOLUTION TO THE BREAK MACHINERY OBJECTIVES THAT'S ACTUALLY GOOD!

wonderinghost:
Woo we get to keep protolathes

Therandomhoboo:
Pretty based PR, apart from removing assassinate. Whilst I get its for "RP" stuff, still don't think taking it away changes our "RP", this is just appeasing those "sweat" medics that will go out of their way to find the body and that itself is more of a player problem really. 

[EDIT] If we are really stopping RR objectives, we should really remove hijack unless its for I.A.A or Infernal A.A

Moltijoe:
> Pretty based PR, apart from removing assassinate. Whilst I get its for "RP" stuff, still don't think taking it away changes our "RP", this is just appeasing those "sweat" medics that will go out of their way to find the body and that itself is more of a player problem really.
> 
> [EDIT] If we are really stopping RR objectives, we should really remove hijack unless its for I.A.A or Infernal A.A

we still have kill original body and kill at least once
maroon is functionally identical to assassinate in most situations

UltraBigRat:
While i dont like my objectives randomly coming undone and such, this seems like a pretty bad solution and makes these objectives really inconsequental and simple, as literally all you have to do is tide in and screwdriiver and crowbar. Why with the objective becoming easier did the amount of machines needing to be broken decrease aswell? 

Im a bit iffy around assasination and especially maroon, traitors are supposed to be scary why guarantee that they wont keep you dead?

Moltijoe:
> While i dont like my objectives randomly coming undone and such, this seems like a pretty bad solution and makes these objectives really inconsequental and simple, as literally all you have to do is tide in and screwdriiver and crowbar. Why with the objective becoming easier did the amount of machines needing to be broken decrease aswell?
> 
> Im a bit iffy around assasination and especially maroon, traitors are supposed to be scary why guarantee that they wont keep you dead?

Honestly, i'm going to want to see if i can require dismantling the machine frame entirely, so it at least takes some time
if that's not possible i'll probably increase the number of machines again

As for assassination and maroon, the issue is that being round removed doesn't make them "scary" it makes them frustrating
once anyone's encountered something enough, it stops being scary, and all that remains is the fact that the victim has to sit out for up to an hour or so. 
Ideally this would be solved by ghost roles, but for the time being, our ghost roles are rather lacking, and often not enough when they're really needed.
Also, it's not uncommon for assassination or maroon to cause a chain reaction effect, where an antag can justify killing damn near half the station because it's under the guise of "preventing revival". When it's invalid, it's usually just some shitter trying to get away with murderbone by using a technicality, when it's valid, it's usually some poor sap that's just been given an unfortunate target to kill, once that's basically defended by the entire crew.

I'm not particularly dead set on removing assassination or maroon, it was just a random thought i had, that our antag/sec/crew dynamic isn't well suited to MRP as we say we are.
I've drafted this pr for the time being until proper discussions are had by specific people.

ToasterBiome (changes requested):
I like the break machinery changes and maroon organ but assassinate should still be in there

ToasterBiome (approved):

ToasterBiome (approved):

PR #20982, opened 2023-11-27
Fixes monster hunter's steal objective not properly updating explanation text

Moltijoe:
:cl:  
bugfix: Fixes monster hunter's steal objective not properly updating explanation text
/:cl:


Moltijoe:
i was lied to

PR #20984, opened 2023-11-27
Fixes the actually broken monsterhunter objective

Moltijoe:
:cl:  
bugfix: Makes monster hunters monster hunting objective actually tell the monster hunter to hunt monsters
/:cl:


MajManatee (approved):
Probably a good idea

PR #20996, opened 2023-11-28
[PORT] Ports a single tile from a TG PR primarily so that I can rant in the description of this PR about a stone floor tile sprite

neeshacark:

Replaces a single sprite because the current one sucks, pending the rest of the PR which I BELIEVE this is from - https://github.com/tgstation/tgstation/pull/58932

**### WHAT FOLLOWS IS A VERY LONG RANT ABOUT A STONE FLOOR TILE SPRITE**

So this PR was originally going to be me going back to deliver on something I said I would do when I originally ported stone tiles.  The plan was to add an obj/ for stone tiles, and add a cargo crate so you could buy them as there is no stone mineral.  UNFORTUNATELY TG doesn't actually have the obj/ sprite for stone tiles, as they don't have them as an obj/ at all - much like how it is for us.  If they get broke, you're out of luck.

So, after spending about 5 minutes scraping my eyes over TG's tiles.dmi file I finally convince myself that its in fact not just me being a blind bastard and missing it.  So, I go and check TG's biggest mirror, skyrat.  I go straight to the tiles.dmi in their main folder, figuring if it was there I'd see it - nope.  I scour it for about 3 minutes because I'm still concerned I can't fucking see.

It's at about this point I think to myself, 'well maybe I should just sprite it myself' but as I'm not a spriter and would rather not do that I decide to make absolutely sure that they don't have one.  So I check their PRs for "stone" and would you look at that they have this recent PR "Primitive stuffs" and in that PR's description there is an image which shows their crafting menu with stone tiles as an option.  So hey, if they can craft them that means they MUST have a sprite! So I go back to tiles.dmi and spend another 3 minutes before I am positive that there is nothing there.

So, I search a little in their github trying to find where the stone floor icon is because CLEARLY its not in tiles.dmi and naturally since I fucking SUCK I can't find it for like 10 minutes before eventually I work out that skyrat is modular and that one is for stone.  So I go through the files to the stone area, and check the are where the icons are kept.  There is no tiles.dmi or any floortile adjacent one.  There is wall.dmi and ore.dmi

So at this point I'm really annoyed because I've spent about 20 minutes on this, not accounting for the time I already spent working out how to add cargo crates and all the other stuff for this before I realized that when I ported stone tiles to begin with I didn't make an object for them because I was being fucking lazy.  So in total I've been at this for around 35ish~ minutes. 

I decide that I'm just going to find the lines for whatever object the floor tile uses to find this stone tile icon in the tile_types.dmi 
or whatever Skyrat's version of that is because we've been split off from TG for long enough that they sort stuff entirely different from us thanks to actually having people who decided to deal with that tedium.  I should've done this from the start but in my defense I kinda figured it would've been quicker to just look at the tiles.dmi because that's where the majority of the sprites are located anyway especially if the codebase itself had sprited them themselves which in this case there was no other option given TG obviously didn't have the fucking sprite and I've never even heard of the other TG mirror so I doubt they took it from there.

But no, eventually I do find the goddamn code for the stone tile and what sprite does it use? Well naturally tile_herringbone of course! So TG has all these unique tile variants because they're cool and one of them is TILE_HERRINGBONE and for some reason whoever at skyrat did this decided to make that the sprite for stone floors despite them not resembling stone floors in the slightest - not to mention the fact that I can only assume they also have the ACTUAL herringbone tile already existing meaning there is now two distinct tiles using the same object sprite and one of them looks NOTHING LIKE IT FUUUUUUUUUUCK MEEEEEEEEEEEEE

So anyway, I gave up and did this instead.  This is not a dig at Skyrat or TG if any of you see this I'm just so fucking mad please I've spent the past hour and a half or so with this fuckin stone tile haunting my night and at the end it didn't even EXIST.

# Spriting
image

# Changelog

:cl:  Twaticus, Imaginos16, Cark
imageadd: added new object sprite for plasteel tile
imagedel: deleted old object sprite for plasteel tile
/:cl:


MajManatee:
That was definitely a rant yep (I didn't read it). But it does look better imo.

cuackles (approved):

PR #21012, opened 2023-11-30
Minor Bloodsucker code cleanup

Chubbygummibear:
My brother in christ the comment above the defines say not to use the blood level defines in a switch statement because they're non constant.

Also who tf makes a switch statement in descending order?

# Changelog

:cl:  
tweak: fix bloodsucker code compiler errors  
/:cl:


ToasterBiome (approved):

Chubbygummibear:
looks like the auxmos update also fixed this code so we should be gucci now

PR #21014, opened 2023-12-01
Fixes a runtime error caused by passive vents

SapphicOverload:
:cl:  
bugfix: fixed runtime error when passive vents encounter zero temperature or zero volume
/:cl:


redmoogle:
What about space (although I'd imagine this is for vents in the wall)

iamthesilly:
> What about space

Space is 2.7 kelvin and has 2500 liters of volume

ToasterBiome (approved):

PR #21017, opened 2023-12-01
Fixes a runtime error in abandoned crates

SapphicOverload:
:cl:  
bugfix: fixed abandoned crates having a 2% chance to cause runtime errors when initialized
/:cl:


Moltijoe (approved):

Beeper-Skeeper:
How does that even happen, what item would be causing that?

PR #21019, opened 2023-12-01
slightly lowers superficial healings stage speed requirement to 7, originally its 8

boodaliboo:
changes superficial healing so that the stage speed requirment is slightly lower i think honestly it should be lower like 6 but lowering it to need a stage speed of 7 i think is a good compromise
:cl:  
tweak: changed superficial healing stage speed requirement from 8 to 7
/:cl:


MajManatee:
I never liked the thing as a whole

Moltijoe:
no official balance council response, but they're wanting something bigger to happen to this symptom rather than just small number tweaks.
Whether that's removal, or rework, who knows

PR #21027, opened 2023-12-02
New virology symptom: symbiotic regeneration

SapphicOverload:

Adds a new virology symptom: symbiotic regeneration.

This new symptom heals you slowly while standing still for longer than 3 seconds. Heal rate is 0.5 brute/burn/toxin every 2 seconds.

Adds more variety to the types of beneficial viruses that can be made, so it's less of the exact same thing every round.

# Testing

Made a virus with this symptom, gave it to myself, the symptom healed when standing still for long enough and stopped when moving.

# Wiki Documentation

Add this to the list of symptoms on the guide to infections

# Changelog

:cl:  
rscadd: new virology symptom: symbiotic regeneration
/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/heal.dm
+++ b/code/datums/diseases/advance/symptoms/heal.dm
@@ -505,62 +505,67 @@
 			M.update_damage_overlays()
 	return 1
 
-#define SYMPTOM_SUPERFICIAL_LOWER_THRESHOLD 0.7
-/datum/symptom/heal/surface
-	name = "Superficial Healing"
-	desc = "The virus accelerates the body's natural healing, causing the body to heal minor wounds quickly."
+#define SYMPTOM_PARASYMPATHETIC_MIN_DELAY (1 SECONDS)
+
+/datum/symptom/heal/parasympathetic
+	name = "Parasympathetic Regulation"
+	desc = "The virus accelerates the body's natural healing, causing the body to heal wounds quickly while resting."
 	stealth = -2
 	resistance = -2
 	stage_speed = -2
 	transmittable = 1
 
 	level = 3
-	passive_message = span_notice("Your skin tingles")
+	passive_message = span_notice("You feel calm.")
 
-	var/threshold = 0.9 // Percentual total health we check against. This is less than a toolbox hit, so probably wont save you in combat
+	/// When was the last time the affected mob moved?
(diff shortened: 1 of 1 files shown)


boodaliboo:
i think both should be in the game not as a replacer but as a addition but its kinda obvious i would have that decision

iamthesilly:
> i think both should be in the game not as a replacer but as a addition but its kinda obvious i would have that decision

Superficial healing doesn't do anything useful or interesting, why keep it in the game?

solwashere:
Bungus

boodaliboo:
then changed the fucking stamina damage the resistance tier does
slightly change the stealth
from -2
to -1
make its stealth requiremet 3
stage speed sorry
or more like 3-5 range
but jesus slicer when i cnat get a actual word in of course your not going to understand why i think it should stay
i think it should heal to base 15 but at a slower rate
i think having it add scarring which i believe we have now correct me if im wrong
not making you unknown
but like using the injury system

these are just starts on how to rework superficial healing

iamthesilly:
> then changed the fucking stamina damage the resistance tier does slightly change the stealth from -2 to -1 make its stealth requiremet 3 stage speed sorry or more like 3-5 range but jesus slicer when i cnat get a actual word in of course your not going to understand why i think it should stay i think it should heal to base 15 but at a slower rate i think having it add scarring which i believe we have now correct me if im wrong not making you unknown but like using the injury system
> 
> these are just starts on how to rework superficial healing

The core idea of only healing when you're close to full health is just fundamentally bad.

boodaliboo:
> > then changed the fucking stamina damage the resistance tier does slightly change the stealth from -2 to -1 make its stealth requiremet 3 stage speed sorry or more like 3-5 range but jesus slicer when i cnat get a actual word in of course your not going to understand why i think it should stay i think it should heal to base 15 but at a slower rate i think having it add scarring which i believe we have now correct me if im wrong not making you unknown but like using the injury system
> > these are just starts on how to rework superficial healing
> 
> The core idea of only healing when you're close to full health is just fundamentally bad.

why is it bad, i think personally having a symtopym require radiation to heal is bad just because it incentivizes people getting irridiated and then the rads will spread to others. atleast give me a why its bad and not just its bad

ToasterBiome (approved):

PR #21033, opened 2023-12-03
Nerfs pseudocider

Moltijoe:
It was merged initially as a very hypothetical idea, balancing of such an effect being not super straight forward
It was found to be too strong when combined with martial arts, as well as have a cooldown so short it can be mostly ignored

Increasing TC to 8 brings it in line with antistun implants and adrenals, as well as prevents it from being bought with most martial arts
Increasing the cooldown by 10 seconds shouldn't make a huge impact, but it does mean it likely won't be used multiple times in the same encounter anymore

:cl:  
tweak: Pseudocider costs 8tc instead of 6tc
tweak: Pseudocider has a 30s cooldown instead of 20s
/:cl:


MajManatee (approved):
does what it says

PR #21053, opened 2023-12-05
Minor auxmos update

jupyterkat:

sets up CI to download and build auxmos by itself
removed the hack functions
update byond to 515.1621

:cl:  
tweak: updated auxmos to latest
/:cl:


code/__DEFINES/bindings.dm
```diff
@@ -1,17 +1,49 @@
-#define AUXMOS (__detect_auxmos())
+//THIS FILE IS AUTOMATICALLY GENERATED BY AUXMOS, PLEASE DO NOT TOUCH IT
+//PROC DEFINITIONS MAY MOVE AROUND, THIS IS NORMAL
+
+/* This comment bypasses grep checks */ /var/__auxmos
 
 /proc/__detect_auxmos()
```
JohnFulpWillard, line 6: I think this proc should be a define helper

JohnFulpWillard: Actually nvm with these changes I don’t think it has to be

PR #20130, opened 2023-08-28
Mood isn't optional

Moltijoe:
# Why is this good for the game
it adds a layer of mechanics that can be further developed and iterated upon
when it's disabled by everyone, there's no reason to make it better

:cl:  
rscdel: removes mood preference
/:cl:

PR #21056, opened 2023-12-06
Globally enables mood for everyone

ToasterBiome:

Mood either 1) creates interesting situations where you go out of your way to satiate it or 2) does completely nothing because it doesn't matter

If there's only these two options, then there is no reason to keep it a preference.

Having more mechanics even small can create more interesting scenarios and situations. I think mood is an acceptable level of abstracting a system for the benefit of interesting gameplay. Having it split means that there will always be a resistance of making it more useful from non-mood players. With global enabling it will give more opportunities for maintainers and developers to add more nuance and unique aspects to it. I disagree with the pref on a fundamental level because the experience per server is tailored and you shouldn't be completely non-interactive with an entire system on the server without something to compensate for it like being blind or sheltered does.

# Changelog

:cl:  
rscadd: Mood is globally enabled for everyone
/:cl:


Spacemanspark:
EDIT: Comment made to be more sensible and come across less as a knee jerk response.

I cannot personally recall any times when the moodlet system has led to a more engaging or interesting experience for me. On the contrary, it's led to a lot of tedium or headache, and it's a system I would much rather not have to play with. I have a good deal of experience with the system from an assortment of servers I've tried it on, so I feel secure in saying this. The system also lacks a lot of context of any given character--their backstory, personality, preferences, fears, et cetera.

That said, I can understand the merit of the system as a pure fluff thing. It's perfect for beginners to RP looking for a bit of guidance, perhaps a little afraid to ask for tips from others. I would suggest, if this needs to be forced, removing the buffs/ debuffs associated with the system. At the very least, this will keep the idea of providing guidance to newer players... without inhibiting those who have interest in expressing their own character's thoughts and feelings through RP.
...yes, I know I'm talking about the average yoggite but SHHHHHHHHHHHHHHHHHHHH

thegoldencat413:
im over here enabling my mood i got mood on my preferences right now im just enabling my mood right now

thegoldencat413:
> Please no, the moodlet system is abhorrent when forced onto people and I've never had a good experience with it. The game telling me how my character feels has always felt irritating, not engaging.

it's just the hunger bar in the form of a smiley face

iloveloopers:
Yogstation has fallen

Mqiib:
Grr hate mood grrrr rabble rabble bad mechanic boo hate it etc etc

(Good pr pls speedmerge)

neeshacark:
same opinion as the last time the PR was made #20130

Therandomhoboo:
Eh mixed feelings really. I have mood on but I wouldn't want to force it upon everyone else. So whilst I understand having mood fully on for everyone opens up making moods do more & may make more interadctions. However, it also just adds on another stack of things the coders will have to re-balance, retweak and redo. As well as, it might mean people will stockpile what gives mood buffs and then hide away like usual.

Whilst I'm sure most are fine with it, just think its another load to put on with other things going on like Auxmos (which is better than it were), introduction of TTS and other things. I would honestly suggest making mood more desirable and see if it turns people to turning mood on.

Otherwise I just think this is the same situation like usual. One side will get pissed off if it sways to one side and this will start the same old sayings of "If this gets merged/doesn't get merged I will leave" which is tiring.

tl;dr
Mood should be more interesting beforehand (with maybe test merges or w/e) to see if it will coax more players over otherwise we are just pissing off part of the playerbase for no good reason.

MajManatee:
yoggites when they have the impossible task of "eating" and "not being stabbed 30 times" (they cant validhunt as easily)

UltraBigRat:
option 3) 
image

Moltijoe (approved):

azzzertyy:
no

config/game_options.txt
```diff
@@ -674,6 +674,9 @@ ROUNDSTART_TRAITS
 ## Uncomment to disable human moods.
 DISABLE_HUMAN_MOOD
```
MajManatee, line 675: Shouldnt this be commented out...?

MajManatee (dismissed):
since, ya know, its incompatible

JamieD1 (approved):
I approve. Lets do it

config/game_options.txt
ToasterBiome: No that turns off mood?

Beeper-Skeeper:
I knew turning on mood would pay off eventually, and now I'm just used to it.

PR #21093, opened 2023-12-11
Allows the Captain to unlock their gun's display case on Gamma alert or higher

AMyriad:
Lets the Captain swipe their ID card to unlock the antique display case on alert levels higher than red. Currently they have no way to access their personal gun without destroying the case and triggering the burglar alarm, even if a war has been declared.

Not important, but I thought it would make sense for the captain to be able to access their gun in emergencies.
I agree that they shouldn't be able to access it normally, but when nukies/clockies declare war, or some other horrible catastrophe strikes, captains tend to go for their gun anyways, despite the case. This would only save them a bit of time.
Since powergaming is no longer barred on gamma alert and higher, I see no issue with this.

# Testing
Everything works as intended:
image

# Wiki Documentation
Worth mentioning on the Captain's page.

# Changelog
:cl:  
tweak: Captains can now access their antique gun display case on Gamma alert or higher
/:cl:

### DIFF
--- a/code/game/objects/structures/displaycase.dm
+++ b/code/game/objects/structures/displaycase.dm
@@ -235,11 +235,11 @@
 	start_showpiece_type = /obj/item/gun/energy/laser/captain
 	req_access = list(ACCESS_CENT_SPECOPS) // This is intentional, presumably to make it slightly harder for caps to grab their gun roundstart.
 
-/obj/structure/displaycase/attackby(obj/item/W, mob/user, params) // Unless things are really fucked.
-	if(seclevel2num(get_security_level()) >= SEC_LEVEL_DELTA) // Everything higher than red.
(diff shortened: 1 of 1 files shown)


Hislyna:
Unlocking it as of Red is fine as per 1.3

alexkar598:
> Currently they have no way to access their personal gun without destroying the case and triggering the burglar alarm, even if a war has been declared.

https://www.youtube.com/watch?v=TS3kiRYcDAo

AMyriad:
> Unlocking it as of Red is fine as per 1.3

Sure, but this is more or less a QOL change for certain gamemodes.
I'm keeping it gamma and above, lest it be a mechanic in normal rounds, which turns it into a balance issue I personally wouldn't like to see in the game.

ToasterBiome (approved):

bruhlookatthisdood:
this makes it so when you hit it with any item (meele) it won't let you and just says it can only be lifted above red

PR #21113, opened 2023-12-15
[GAX] Removes Arrival Shuttle and replaces it with cryogenics, adds experimentor room

ToasterBiome:

Steals partially from https://github.com/shiptest-ss13/Shiptest/pull/2056

Replaces Gax Arrivals with cryogenics
image
Messing with it holds the same rules as messing with arrivals in general. Has it's own air supply so it cant be plasmaflooded.
When you spawn you have a chance for some harmless but flavorful negative effects. Shiptest had a chance to damage you but I don't like that. Works even without power, works without the cryo pods not there (magic).

Also adds an "Experimentor room" in maint which gives you the stuff to build one yourself :^)

image

NVS Gax is a ship. You have a little ship move people to big ship???? Emergency shuttle makes sense because it is actually evacuating people.

# Testing

https://github.com/yogstation13/Yogstation/assets/5091394/255215af-cb47-4d67-ba12-f3bbd1938b48

# Wiki Documentation
No more arrivals shuttle for gax, Cryo is moved.

# Changelog

:cl:  
mapping: Removes arrival shuttle from NVS Gax, people spawn in cryo now
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -442,3 +442,18 @@ GLOBAL_VAR_INIT(cryopods_enabled, FALSE)
 /obj/machinery/cryopod/proc/apply_effects_to_mob(mob/living/carbon/sleepyhead)
 	sleepyhead.SetSleeping(50)
 	to_chat(sleepyhead, span_boldnotice("You begin to wake from cryosleep..."))
+	sleepyhead.set_nutrition(200)
+	sleepyhead.SetSleeping(60) //if you read this comment and feel like shitting together something to adjust IPC charge on wakeup, be my guest.
+	//but it can be worse.
+	if(prob(90))
+		sleepyhead.adjust_drowsiness(rand(3 SECONDS, 10 SECONDS))
+	if(prob(75))
+		sleepyhead.blur_eyes(rand(3, 6))
+	//so much worse
+	if(prob(66))
+		sleepyhead.adjust_disgust(rand(25,35))
+	if(prob(33))
+		sleepyhead.adjust_disgust(rand(20,30))
+	if(prob(16))
+		sleepyhead.adjust_disgust(rand(10, 17))
+	to_chat(sleepyhead, "<span class='userdanger'>The symptoms of cryosleep set in as you awaken...")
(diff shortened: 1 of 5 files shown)


Aquizit (changes requested):
image

This looks a bit weird. I don't necessarily mind the warning stripes stopping there but without something going into the cryo room it just seems off to me.

ToasterBiome:
> image
> 
> 
> 
> This looks a bit weird. I don't necessarily mind the warning stripes stopping there but without something going into the cryo room it just seems off to me.

What should go there instead?

ToasterBiome:
image

Aquizit (approved):

code/game/machinery/cryopod.dm
```diff
@@ -432,3 +432,28 @@ GLOBAL_VAR_INIT(cryopods_enabled, FALSE)
 	log_admin(span_notice("[key_name(target)] entered a stasis pod."))
 	message_admins("[key_name_admin(target)] entered a stasis pod. (<A HREF='?_src_=holder;[HrefToken()];adminplayerobservecoodjump=1;X=[x];Y=[y];Z=[z]'>JMP</a>)")
 	add_fingerprint(target)
+
+/obj/machinery/cryopod/JoinPlayerHere(mob/M, buckle)
+	. = ..()
+	open_machine()
```
iamthesilly, line 438: Should probably have it open after the 6 second sleep ends instead of instantly, looks weird when it opens while you're asleep.

code/game/machinery/cryopod.dm
```diff
@@ -432,3 +432,28 @@ GLOBAL_VAR_INIT(cryopods_enabled, FALSE)
...
+	if(iscarbon(M))
+		apply_effects_to_mob(M)
+
+/obj/machinery/cryopod/proc/apply_effects_to_mob(mob/living/carbon/sleepyhead)
+	sleepyhead.SetSleeping(50)
+	to_chat(sleepyhead, span_boldnotice("You begin to wake from cryosleep..."))
+	sleepyhead.set_nutrition(200)
+	sleepyhead.SetSleeping(60) //if you read this comment and feel like shitting together something to adjust IPC charge on wakeup, be my guest.
```
iamthesilly, line 445: SetSleeping is called twice here.

SapphicOverload (changes requested):

code/game/machinery/cryopod.dm
ToasterBiome: > Should probably have it open after the 6 second sleep ends instead of instantly, looks weird when it opens while you're asleep.

I don't want to hog them up for 6 seconds to do this. Since this is both "join and leave the game" I wanted the entering to be relatively quick in the cryopod itself. 

Also this breaks is the panel is open.

SapphicOverload (approved):
Good enough

PR #21115, opened 2023-12-16
Gives the vampire cloak a new look

Moltijoe:
it's supposed to be a powerful ability of the vampire's
sorta puts a damper on it when it looks like something bought from spirit halloween

initially tried adding an animation, but couldn't figure out how to make it loop nicely

# Spriting
image

:cl:  Scrambledeggs00 & Molti
imageadd: new vampire cloak
imagedel: old vampire cloak
/:cl:


cowbot92:
wait this is actually badass??

BlueishTsunami (approved):
Sick

kooarbiter:
This is actually really drippy. Maybe add a way to swap which shoulder the cape thing is held? And pipe dream maybe add a small looping animation of blood dripping from the bottom of the coat but i wouldnt expect anyone to add that.

PR #18451, opened 2023-04-03
makes surgical "tools" less useful as weapons

solwashere:
# Document the changes in your pull request

Tools that are meant to be meant for surgery should remain that way, it makes no sense as to why a saw that is meant to slowly cut through someone's skull can do so much damage to a person, as well as the silly high wound bonus.
I feel like these changes will better reflect how they would actually function in combat (not well), as well as lightly reduce the average medical staffs validhunting capabilities, as these tools can be mass produced on a whim.
# Spriting

# Wiki Documentation

# Changelog

:cl:  @Therandomhoboo @solwashere 
tweak: The circle saw, scalpel, laser scalpel, and surgical drill have all had their damage and wound bonus toned down

/:cl:

PR #21117, opened 2023-12-16
Removes bonus demolition damage from the laser scalpel

githubuser4141:

--- Removes the 1.5x demolition damage buff on the laser scalpel

-- WHY?

As stated above. It doesn't make any sense, the reasons why are listed in this PR: https://github.com/yogstation13/Yogstation/pull/18451

It's bizarre that after this PR was merged, another PR comes out immediately after and buffs laser scalpel's combat utility because "laser good at metal". Laser SCALPELS aren't designed to cut METAL, they are precision surgical tools to cut flesh and occasionally bone. 

There's a big difference between a surgical tool in a operating theatre being used by a surgeon to seperate flesh and some rando using it to, just SMACK a sheet of metal. If it's bad against human tissue when employed in a violent manner it's going to be bad against metal too. Because there's no time in combat to direct the laser at the metal, and it really is just -smacking- it'd stand to reason that the laser scalpel, having no metal edge to dint the metal, might actually be *worse* at damaging sturdy metal - because there's no immediate impact. It's a relatively weak laser that doesn't have any business being used as a weapon grazing across a thick(er) layer of steel. It literally doesn't have the time to cut the steel before the attack is over. 

There's no reason a surgical implement that's designed to be especially effective in a medical context should also magically be extremely good at stopping military bots, mechs or trying to smash through barricades. That's what toolboxes and ion guns are for. As MajManatee said, it's an incredibly tiny laser beam - it'd probably do less damage to tissue (in an attack), it just doesn't have the oomph to do anything to armor, it's too tiny and you're not applying it correctly.

# Testing

# Wiki Documentation

# Changelog

:cl:  
tweak: Laser scalpels don't have their demolition damage modifier anymore.
/:cl:

### DIFF
--- a/code/modules/surgery/tools.dm
+++ b/code/modules/surgery/tools.dm
@@ -400,7 +400,6 @@
 	light_range = 1
 	light_color = LIGHT_COLOR_GREEN
 	light_power = 0.2	//Barely glows on low power
-	demolition_mod = 1.5 // lasers are good at cutting metal
 	sharpness = SHARP_EDGED
 
 


iamthesilly:
I don't see how going from 10 to 15 damage against inanimate objects and structures only (not players) is a problem for balance. Also yes, lasers ARE actually better at cutting through rigid materials like metal as opposed to soft organic materials. I know this from experience, from operating actual laser cutters.

drderp3635:
image
Average surgical tools enthusiast writing a 9 page essay on the dynamics of a scalpal.

githubuser4141:
> I don't see how going from 10 to 15 damage against inanimate objects and structures only (not players) is a problem for balance. Also yes, lasers ARE actually better at cutting through rigid materials like metal as opposed to soft organic materials. I know this from experience, from operating actual laser cutters.

It's not a laser cutter that's intended to cut metal, it's specifically supposed to cut flesh or bone with the least amount of collateral damage - so the cut should be as small and precise as possible. It really shouldn't do much in a slapfight - you're not cutting metal on a workbench you're just hitting it. And many inanimate objects (Borgs, mechs) are players. Would you prefer a 0.75 or 1.0 multiplier instead, so it's a little better?

MajManatee:
I mean yeah
its a _scalpel_

Moltijoe:
Balance Council:
>Non-issue, toolboxes are damn near everywhere and do just as much damage, and aren't limited to things that get affected by demo bonus

ofscarletdreams:
marc wrote this

githubuser4141:
> Balance Council:
> 
> > Non-issue, toolboxes are damn near everywhere and do just as much damage, and aren't limited to things that get affected by demo bonus

What is this answer? Toolboxes aren't surgical equipment which is supposed to be delicate/fragile and used in a professional setting. A toolbox is just a metal case with padded inserts to hold tools, it's made out of metal and is normally rather robust because it's not required to do any specialized work; it's literally just a box.

The point of the first PR which you guys "balance council" approved was to tone down the deadliness of *specialized tools* (i.e. medical devices) in combat. Toolboxes aren't specialized; they are metal boxes. Toolboxes also have a much bigger form factor - won't fit in bags, and definitely won't fit on a pocket unlike the laser scalpel.

Also the "ided" label can apply to the "balance council" more then it applies to the PR imo. Their argument doesn't make sense and it actually backtracks on the original idea of nerfing surgical tools. It's like these people tried to fight a mech with circular saws and died several times, when I made this PR I hadn't even played the game in over a few months.

PR #21119, opened 2023-12-16
Lisp reduction (understanding buff)

Moltijoe:
Lisps just make things harder to read, hurting communication, lizard is mostly fine at 3, but considering how common they are it gets to be a bit annoying
poly being at 6 makes them utterly incomprehensible

:cl:  
tweak: Lizard lisp is now 2 S instead of 3
tweak: Poly lisp is now 3 S instead of 6(!)
/:cl:


Yarinoi:
agree with the poly one, lizards are fine in my opinion though

MajManatee:
species downside

AdamPrime314:
> Lisps just make things HARDER TO READ, HURTING COMMUNICATION

The game is heavily reliant on communication with many debuffing effects being able to cripple your communication skills with others, but it is a rather insignificant "cripple" in this case which is featured mostly to add "flavor and character" to the present species. I would argue that the lisps are not inventive enough, given how non-invasive they are to a player's speech. I would imagine it would create for a more interesting scenario if there were MORE instances of species specific pronunciation without having to rely on self-lisping or the regional accents. It is a rather identifying trait for a lizard to not be able to properly articulate themselves when present with a challenge of having to utter something with a single "s" in it without outing themselves to everyone around them, but it's strange that someone decided it would shine with originality to give polysmorphs the same trait with a longer string of spammed "s"'s. Honestly, making their lisps equal to each other while adding more creative ways of further embracing their hissy and alien way of talking would make large sense given how much similarity there is between [lizards](https://www.youtube.com/watch?v=rx7Z9krgTz8) and [xenomorphs](https://www.youtube.com/watch?v=TJoLnHqNg_U) hissing.
### Normal - Not modified by any speech effects.
> Nukies maints | So how did you do this thing? | Can you help me please? | I would like a power cell please! | Join the X-Force!
### Lizard - Modified by the default lizard lisp.
> Nukiesss maintsss | SSSo how did you do thisss thing? | Can you help me pleassse? | I would like a power cell pleassse! | Join the X-Force!
### Polysmorph - Modified by the default polysmorph lisp.
> Nukiessssss maintssssss | SSSSSSo how did you do thissssss thing? | Can you help me pleasssssse? | I would like a power cell pleasssssse! | Join the KSSSSS-Force!

ToasterBiome:
Don't loik this

PR #21120, opened 2023-12-16
Preterni no longer regenerate blood

Moltijoe:
Sets up for when mech wounds get added in a few years

# Testing

:cl:  
tweak: Preterni no longer regenerate blood
/:cl:


00ze-cyclone:
Preterni are supposed to be organic, how the hell can they get blood when grown in a cloner if their body doesn't generate it?
Most importantly why do IPC not get that since they are literally just machines?

Moltijoe:
> Preterni are supposed to be organic, how the hell can they get blood when grown in a cloner if their body doesn't generate it? Most importantly why do IPC not get that since they are literally just machines?

The cloner clones blood for them
Ipc dont have blood at all

ToasterBiome (approved):

PR #21122, opened 2023-12-17
[BOX] Allows people to exit through the left door in medbay in addition to the right one.

neeshacark:
title

no reason for it to be this way, and its really convenient to have it this way instead

# Changelog

:cl:  cark
mapping: the western medbay doors on medbay now have noaccess from the south
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -17974,22 +17974,6 @@
 /obj/machinery/door/firedoor/border_only,
 /turf/open/floor/plasteel/dark,
 /area/security/prison)
-"dCT" = (
-/obj/machinery/door/firedoor/border_only{
-	dir = 1
-	},
-/obj/machinery/door/firedoor/border_only,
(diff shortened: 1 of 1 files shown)


Aquizit (dismissed):

bruhlookatthisdood:
It was supposed to be that way, one entrance, one exit. The arrows are now misleading people. 
If people are too dumb to follow the arrows then that is on them.

Aquizit (changes requested):
They bring up a good point. If you're gonna do this remove the arrow decals on the floor.

azzzertyy:
> It was supposed to be that way, one entrance, one exit. The arrows are now misleading people. If people are too dumb to follow the arrows then that is on them.

Except that aint how it works. Entry from both. So doesnt make sense, remove arrows

neeshacark:
> They bring up a good point. If you're gonna do this remove the arrow decals on the floor.

i would say thats a bit of a silly requirement. nice to have them there, just because a door is marked enter doesnt mean you can use it to exit

PR #21150, opened 2023-12-20
Adds more ghetto options to mechanical surgery

Moltijoe:
multitool step also gets
shock touch 90%
energy gun 90%
stun baton 90%
shockpaddles 90%
but it will shock the target, doing more damage on failure
(lower chance than the others because multitools aren't quite as common as the other tools)

unwrench step gets crowbar 100%
but it will do brute damage to the target, doing more damage on failure

wrench step gets welder 100%
but it will do burn damage to the target, doing more damage on failure

also increases the success chance of regular medical tools to 20 from 10 because at 10 they're painful to try and use, despite the comparable safety

screw shell step is untouched because it already has great ghetto options

more stuff i guess?

# Testing

:cl:  
rscadd: Adds more tool options for ghetto mechanical surgery
tweak: existing ghetto mechanical surgery options are slightly better
/:cl:

### DIFF
--- a/code/modules/surgery/mechanic_steps.dm
+++ b/code/modules/surgery/mechanic_steps.dm
@@ -106,7 +106,7 @@
 	name = "unwrench bolts"
 	implements = list(
 		TOOL_WRENCH = 100,
-		TOOL_CROWBAR = 80,//this sounds like a REALLY bad idea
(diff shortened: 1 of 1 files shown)


Yarinoi:
can you add coins for the screwdriver steps

MajManatee (approved):
yeah sure

PR #21162, opened 2023-12-22
Being on fire no longer causes a negative mood effect if you're immune to fire

SapphicOverload:

Being on fire no longer gives a negative moodlet for being on fire if you're immune to fire, and also fixed an issue with it not clearing itself when it should.

This is already the case for suits that provide full fire protection, so I see this as just an oversight.

# Testing
image
image

# Changelog

:cl:  
tweak: being on fire no longer gives a negative moodlet if you're immune to fire
/:cl:


ofscarletdreams:
it should be a neutral moodlet noting it seems a bit warm in here

ofscarletdreams:
or "golly gee willickers it sure is a good thing i am protected against this fire engulfing me"

ToasterBiome (approved):

PR #21163, opened 2023-12-22
Fixes a runtime error caused by steamed hams

SapphicOverload:
Fixed a runtime error caused by steamed hams trying to create smoke before the smoke subsystem is initialized.

:cl:  
bugfix: fixed a runtime error caused by steamed hams
/:cl:


Addust:
hammed steams

PR #21176, opened 2023-12-27
Directional Exosuit Lighting

SapphicOverload:

Gives exosuits directional lights and a much longer light range.

The current light range feels way too small for something like an exosuit, which I think should probably have something more powerful than a handheld flashlight.

# Testing
image

# Changelog

:cl:  
tweak: exosuits now have directional lights, with much longer range
/:cl:


ToasterBiome (approved):
looks good

PR #21200, opened 2024-01-02
Gives plasmamen plasma remains when dusting

AMyriad:
This PR makes plasmamen use the plasma remains sprite instead of normal remains.

I think it makes sense for the sentient purple skeletons to collapse into a pile of purple dust.
Always thought it was slightly odd they just became a regular old pile of bones, and I feel the same with the other races, but this sprite is already made, and is a tad under-used too.

# Testing
https://github.com/yogstation13/Yogstation/assets/143908044/faa61e04-252c-44eb-9ad3-a8f7b1811288

image

It should work

# Wiki Documentation
This stuff isn't documented anywhere.

# Changelog

:cl:  
rscadd: Dusted plasmamen now have plasma remains
/:cl:


AdamPrime314:
Are plasmamen just skeletons with plasma around them, or their bone structure is completely enriched with plasma, making it purple? Lore team get in here.

Moltijoe (changes requested):
pretty sure this makes gibbing them not actually spawn gibs

AMyriad:
> pretty sure this makes gibbing them not actually spawn gibs

tested it, this is right, woops
though come to think of it, it could be argued that plasmamen shouldn't turn into a pile of blood and meat when they have neither?

Moltijoe:
> tested it, this is right, woops though come to think of it, it could be argued that plasmamen shouldn't turn into a pile of blood and meat when they have neither?

could always just add a different form of gibspawner specifically for them

AMyriad:
> could always just add a different form of gibspawner specifically for them

bet

Cartlord:
Why was AdamPrime's comment marked as abuse when it's a reasonable thing to bring up?

AMyriad:
alright so I ended up trying to do some interesting things that made the scope of this PR a lot bigger than what I want it to be
I'll save the surprise for another time, but I tested gibbing and it's working fine again now

JamieD1 (approved):

PR #21214, opened 2024-01-06
[BOX] Fixes railings in bar

neeshacark:

fixes some railings in bar

# Changelog

:cl:  cark
mapping: fixes some railings in bar
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -31986,17 +31986,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/hallway/primary/aft)
-"iAG" = (
-/obj/effect/turf_decal/siding/wideplating{
-	dir = 1
-	},
-/obj/effect/turf_decal/ramp_middle,
-/obj/structure/railing{
-	dir = 1
-	},
-/obj/machinery/vending/snack/random,
-/turf/open/floor/wood,
-/area/crew_quarters/bar)
 "iAN" = (
 /obj/effect/spawner/lootdrop/grille_or_trash,
 /turf/open/floor/plating,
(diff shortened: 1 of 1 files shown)


azzzertyy:
not sure how this is a fix? it looks worse and makes less sense, you put the railings on this position if this is the position that is raised, but it is not, the other side is raised.

Aquizit:
You know what the entire problem with this shit is?

image

These border tiles. Change them to the wood type and look how much better this will be.

image

I mucked up replacing the shadows but y'all know what I mean here.

image

And here's the thicker version of the trim to decide which looks best but seriously: make it wood

Aquizit (changes requested):
I forgot I can actually just say do it and not just suggest it. :)

azzzertyy:
> You know what the entire problem with this shit is?
> 
> image
> 
> These border tiles. Change them to the wood type and look how much better this will be.
> 
> image
> 
> I mucked up replacing the shadows but y'all know what I mean here.
> 
> image
> 
> And here's the thicker version of the trim to decide which looks best but seriously: make it wood

gross ew less constrast looks way worse, also, its not a "border" its to signify the depth, look at chapel2, it does the same thing.

Aquizit:
It might actually be because I think the shadow decal is UNDERNEATH the gray trim in the first image. Which is making it basically not fucking exist.

azzzertyy:
> It might actually be because I think the shadow decal is UNDERNEATH the gray trim in the first image. Which is making it basically not fucking exist.

its like that on chapel2 as well, its to add the shadow onto the floor and not the trim.

neeshacark:
> not sure how this is a fix? it looks worse and makes less sense, you put the railings on this position if this is the position that is raised, but it is not, the other side is raised.

because the way they are people render over them, they are facing the wrong way.

Aquizit:
> > It might actually be because I think the shadow decal is UNDERNEATH the gray trim in the first image. Which is making it basically not fucking exist.
> 
> its like that on chapel2 as well, its to add the shadow onto the floor and not the trim.

Ok, but on chapel2 it's going from tile-type trim to tile flooring. So it's less jarring. Plus the color is also less jarring. I'm understanding what it's supposed to look like: the wall there because it's lifted up.

The issue is we don't do that kind of perspective ANYWHERE ELSE -- like prime example is I think we're one of the few codebases left that do wall tiles that don't present a sort of visual verticality to them, like trying to portray a side of the wall so it makes more sense to see machinery on it. So doing that for raised areas just throws that off to my eye.

Aquizit (approved):

PR #21241, opened 2024-01-12
Modifies alien embryo to care about the host

Moltijoe:
Now it cares about life ticks

Gives xenos a reason to keep bodies alive without forcing them to
stopping life ticks will greatly slow down progress, giving medical a bit of extra time if it's cutting it close
If xenos space maints to make it significantly easier to defend, they'll hatch slower, if they don't, they'll hatch faster

# Testing
image

:cl:  
tweak: xeno embryos now burst significantly faster if the host is alive and buckled to a xeno bed
tweak: xeno embryos now burst significantly slower if the host is dead or not buckled to a xeno bed
tweak: xeno beds now handcuff the person buckled to it
tweak: xeno beds are significantly faster to unbuckle from (if not handcuffed)
/:cl:

### DIFF
--- a/code/game/objects/structures/beds_chairs/alien_nest.dm
+++ b/code/game/objects/structures/beds_chairs/alien_nest.dm
@@ -34,7 +34,7 @@
 					span_warning("[M.name] struggles to break free from the gelatinous resin!"),\
 					span_notice("You struggle to break free from the gelatinous resin... (Stay still for two minutes.)"),\
 					span_italics("You hear squelching..."))
-				if(!do_after(M, 2 MINUTES, src))
+				if(!do_after(M, 15 SECONDS, src))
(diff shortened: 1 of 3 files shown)


Firewolf34:
Xenos have always been meant to build a nice warm hive to reproduce in. So it is definitely cooler and likely more fun from a mechanics standpoint if they don't do it out in pure space. Cool PR! Though perhaps be careful with the overall average time-to-chestburst period that will increase now that you've made it non-linear and based on health amount, as the majority of infected hosts will likely be injured. Follow-up PR idea could be to make heat impact the gestation time positively (to a point).

Edit: nvm about the non-linearity, looks like the user having positive health causes only a constant speed boost, perfect.

ToasterBiome (approved):

PR #21245, opened 2024-01-13
Fixes chat not always showing up in replays

SapphicOverload:
uh oh

:cl: 
bugfix: fixed chat not always showing up in replays
/:cl:


ToasterBiome (approved):
works

PR #21254, opened 2024-01-14
Airalarm can now turn on/off vent space detection

warface1234455:
Airalarm can now turn on/off vent space detection

Allow unary vent to pump gas into space, a good feature i guess. Probally help pump gas out faster into space

This one is faster than injector and be remotely controlled which makes it the best way to dump waste into space

# Testing
tested, was able to pump gas into space after multitooling the vent

# Changelog

:cl:  

tweak: Airalarm can now turn on/off vent space detection
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/airalarm.dm
+++ b/code/modules/atmospherics/machinery/airalarm.dm
@@ -346,6 +346,7 @@
 					"excheck"	= info["checks"]&1,
 					"incheck"	= info["checks"]&2,
 					"direction"	= info["direction"],
+					"space_detection" = info["space_detection"],
 					"external"	= info["external"],
 					"internal"	= info["internal"],
 					"extdefault"= (info["external"] == ONE_ATMOSPHERE),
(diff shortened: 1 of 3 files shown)


iamthesilly:
Does this need to exist? Why not just use an injector, which does the exact same thing but better?

ShoziOffical:
> Does this need to exist? Why not just use an injector?

yes, if you want faster gas ejection. Also wrong vent is far more better than injector with the internal 0kpa

ShoziOffical:
if you dont have an airalarm nearby then you can just use injector

wonderinghost:
I think it be nice if we could with air scrubbers. However biggest pain is zoning because you can't claim areas outside of a contained room

ShoziOffical:
Injector also has cons that you have to get near it to turn it on/off. While vent can be remotely controlled via airalarm

ShoziOffical:
> I think it be nice if we could with air scrubbers. However biggest pain is zoning because you can't claim areas outside of a contained room

Honestly, scrubbers are slow at sucking gas. The vent external 0kpa on scrubbing mode can do this job far more faster

Moltijoe (changes requested):
if injectors aren't fast enough, just make them faster
if possible, try not to add to many options to the air alarm, we already have a lot
also, specific devices should be the best at their specific job, rather than strictly worse than something not directly designed for it

ShoziOffical:
Guess ill buff the injector in the other pr then

PR #21261, opened 2024-01-16
Fixes regenerative cores having a delay on icemoon

SapphicOverload:
Technically a station level as well as a mining level so only checking for station level is a problem

:cl:  
bugfix: fixed regenerative cores having a delay on icemoon
/:cl:


N3D6:
Thx

ToasterBiome (approved):

PR #21265, opened 2024-01-16
The Service Hall now has its own access type, and Artists can get in

Cartlord:

The Service Hall on every map now has its own access type, numbered 79. I have edited every Service Hall door to require this access, and usually no others, except in cases where the Service Hall door is adjoining another department - in which case it requires access to the Service Hall as well as to whatever department the door goes to (e.g. a door connecting Hydroponics & the Service Hall requires Hydroponics Access & Service Hall access).
I have given Service Hall access to every job that could get into it before, along with Artists. I am considering giving it to Assistants, but I do not know if that would be a good idea. I am open to feedback on what other jobs should have this access.
The reason I did all this was to make it so that any job can be given access to the Service Hall without any side-effects, because I wanted to let Artists in so they can get spray paint from the autolathe.

The Artist can now get to the autolathe and give themself all the spray paint they want, without breaking in (or asking someone, but let's be real, they never bothered asking). The Head of Personnel can now give or take away Service Hall access. Anyone who wants to edit a job's access can now add or remove Service Hall access, which I imagine will be nice if we ever get more jobs.

# Testing
I tested the changes to Box on a local server. A blank ID couldn't access the Service Hall, an ID with only Service Hall access could, and the Captain could. I also made sure that it properly showed up in the ID computer.

# Wiki Documentation
The pages for the Artist, Bartender, Botanist, Chef, Clerk, Clown, Curator, H.O.P., Janitor, Lawyer, and Mime may need to be updated to indicate their "new" access.

# Changelog

:cl:  
rscadd: The Service Hall now has its own access type, which Heads of personnel can add or remove as they see fit.
rscadd: Artists now have Service Hall access.
mapping: Hydroponics on IceMeta can no longer be accessed by Chefs without skeleton crew being active.
/:cl:


azzzertyy:
Is it not better to just alter the Map to allow artist access?

Cartlord:
> Is it not better to just alter the Map to allow artist access?

**As I said in the Pull Request's description**, Artists don't _have_ their own access type - and adding that just so they can enter one room feels entirely un-necessary.

Cartlord:
For the record: I could've handled this by giving Artists their own defined access, by giving the Service Hall itself its own required access, or by giving Artists a pre-existing access that can also get in that door.
Option #1 felt like it would be total overkill, because it would have only applied to one door and required me to edit the access defines as well as every map file. Option #2 would have required editing every job in Service to have the new Hall access, as well as editing every map & the access defines, but would potentially open the door for giving other jobs Service Hall access (not entirely a bad thing, but who'd take advantage of it?). I chose the third option because it took the least editing, because giving them Library access can be justified as the Artists being allowed to use the library computer to search online for reference material, and because them having Library access (seems like it) won't have any negative consequence.

MajManatee (commented):
Easier =/= better
Id rather see either of the other options done.

Cartlord:
I'd been under the impression that snowflake code, such as making a new type of access just to open one door, was frowned upon. That being said, I'll probably go with making the Service Hall have its own access type, because that seems like it'll get more use than giving a gimmick job its own access type. Closing this myself so I can re-open it when I get around to doing that - if someone feels like this can be merged as-is, they're free to re-open and do just that.

Cartlord:
As @MajManatee requested, I picked one of the other options and gave the Service Hall its own access type.

Cartlord:
DARN.

MajManatee:
no way that's the shrimp

neeshacark (approved):
map is fine

feature freeze moment unless @yogstation13/head-coders want this in

Cartlord:
i am okay with waiting out the feature freeze
i'm honestly more surprised that people haven't been telling me to give assistants or some other job servhall access, i thought about giving it to greyshirts so they had somewhere to get mats & tools but i wasn't sure if that was a good idea

Moltijoe (approved):

Moltijoe:
feature freeze doesn't apply to stuff from before it started

PR #21276, opened 2024-01-19
Fixes a runtime with polysmorph tails

Moltijoe:
The runtime is caused by adding an organ to a human body without physiology
The problem is that bodies initialize organs before initializing physiology
so anyone that spawns with an organ that modifies physiology will cause a runtime because physiology doesn't exist yet

:cl:  
bugfix: Fixes a bug with polysmorph tail crawlspeed
/:cl:


ToasterBiome (changes requested):
fill out the template

ToasterBiome (approved):

PR #21284, opened 2024-01-20
Gives Mr Freeze syndikit a frost oil chem sprayer

jachlompsky:

Adds a variant of the chem sprayer loaded with frost oil instead of poison.
Puts it in the Mr Freeze syndikit.
Changes the flag for the ice created by frost oil to match the type created by gluon grenades, so you don't slip on it while wearing winter boots or whatever else gives NOSLIP_ICE.

Mr Freeze kit is garbage and the only kit worth less than 20 TC. It needs to be improved or removed. Adding the chem sprayer gives it a neat item that is otherwise inaccessible to traitors, helps with the theme by adding more ice, and gives much-needed combat and defensive utility.

In terms of balance: this is basically a slip gun, as another option to complement your esword. You get 20 shots on "spray" and the ice thaws right away, unless you've already fired a few shots to make the air colder. If you want to lock down an area you've already got the gluon grenades which make tiles icy for a whopping 6 minutes. You can refill the sprayer with poison or whatever, but despite normally being a nukie weapon the reagent transfer to people isn't any better than a regular spray bottle.

# Wiki Documentation

add chem sprayer to list of items in the mr freeze kit

# Changelog

:cl:  
rscadd: Added a frost oil chem sprayer to the mr freeze syndikit
/:cl:

### DIFF
--- a/code/datums/components/wet_floor.dm
+++ b/code/datums/components/wet_floor.dm
@@ -83,7 +83,7 @@
 			lube_flags = SLIDE | GALOSHES_DONT_HELP
 		if(TURF_WET_ICE)
 			intensity = 120
-			lube_flags = SLIDE | GALOSHES_DONT_HELP
+			lube_flags = SLIDE_ICE | GALOSHES_DONT_HELP
 		if(TURF_WET_PERMAFROST)
 			intensity = 120
 			lube_flags = SLIDE_ICE | GALOSHES_DONT_HELP
--- a/code/game/objects/items/storage/uplink_kits.dm
+++ b/code/game/objects/items/storage/uplink_kits.dm
@@ -225,7 +225,7 @@
 			new /obj/item/reagent_containers/glass/bottle/beesease(src) // 10 tc?
 			new /obj/item/gun/magic/staff/spellblade/beesword(src) //priceless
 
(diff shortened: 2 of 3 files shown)


MajManatee (commented):
I am slightly concerned about the balance behind this because (drake) some people (drake) are well known (drake) for going hard (drake) with chemicals (drake) and this could (drake) be pretty powerful (drake) in their hands (drake).
Otherwise seems fine.

Previous PRs in this situation like the water gun had the item limited to only be filled with the one substance, but I can see why that wouldn't work here.

boodaliboo:
i dont like just having it in a spray bottle maybe retexture it and possibly give it custom particles or do something to make it more unique

jachlompsky:
> i dont like just having it in a spray bottle maybe retexture it and possibly give it custom particles or do something to make it more unique

just to clarify its not a spray bottle its a chem sprayer, a normal size nukeop-exclusive item that sprays 3 tiles wide. I could make a custom texture for this variant if necessary but it already looks unique, and I'm sure many people have never even seen this thing because it's a pretty bad pick for nukies.

Moltijoe (approved):

Moltijoe:
eh, why fuckin not
syndikits need to be strong because they're rng

PR #21286, opened 2024-01-20
removes blank bullet point from project bee instructions

ktlwjec0:
currently has a blank bullet point, fixes that, also removes an unneeded line break

also fixes this (extra ])
image

# Testing
yes


MajManatee (approved):
ya ok

PR #21289, opened 2024-01-21
Imperial jumpsuits can no longer be adjusted because there isn't a sprite for it

Mqiib:
Should hopefully fix the jumpsuit sprite not appearing properly probably. Otherwise there's a deeper problem with jumpsuit style togging (or manatee broke something with his warhammer pr which wouldn't be the first time)

# Changelog

:cl:  

bugfix: Imperial jumpsuits don't have alt styles, stop trying to break the dress code soldier!

/:cl:


MajManatee (approved):
I resent that

PR #21291, opened 2024-01-21
sends the raven cruiser back to hell

ofscarletdreams:
breaks roundend because ???????? shuttle code

### DIFF
--- a/code/datums/shuttles.dm
+++ b/code/datums/shuttles.dm
@@ -542,6 +542,7 @@
 	admin_notes = "Go big or go home."
 	credit_cost = 7500
 
+/* Disabled for having fucked atmos
 /datum/map_template/shuttle/emergency/raven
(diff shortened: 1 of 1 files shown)


ShoziOffical:
i brought it back because i thought it broke atmos only

ShoziOffical:
and you sure only raven break the round-end and not other shuttles too?

Bibby0110:
Literally not the reason round-end breaks

ofscarletdreams:
then what IS

ShoziOffical:
> then what IS

something is returning in code could be antags that make round not possible to be ended

ShoziOffical:
i tried to test the raven cruiser on my private server and there was no runtime

Bibby0110:
> then what IS

Something is making SSTicker sleep

PR #21300, opened 2024-01-23
Buffs and nerfs monk staff null rod slightly

Mqiib:
With the addition of the stamina regen rework, the monk staff's stamina damage has become SIGNIFICANTLY less effective if a fight lasts more than 10 seconds (which is likely given its low damage and complete lack of AP). Originally it was more or less just as effective because you didn't have that bit of stamina regen but now its really quite bad comparatively. This puts it back up to the same damage as other null rods to make it slightly less of a terrible option compared to the swords for combat.

Edit: It also now doesn't block bullets because the other null rods also didn't.

# Wiki Documentation

Monk staff null rod now does 18 damage (same as normal null rod) and can't block projectiles any more, just melee.

# Changelog

:cl:  

tweak: Chaplain monk staff null rod (stamina) damage increased (15 -> 18)
tweak: The above no longer can block projectiles

/:cl:

### DIFF
--- a/code/game/objects/items/holy_weapons.dm
+++ b/code/game/objects/items/holy_weapons.dm
@@ -359,6 +359,11 @@
 	menutab = MENU_WEAPON
 	additional_desc = "The weapon of choice for a devout monk. Block incoming blows while striking weak points until your opponent is too exhausted to continue."
 
+/obj/item/nullrod/bostaff/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the attack", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
+	if(attack_type == PROJECTILE_ATTACK)
+		final_block_chance = 0 //Don't bring a stick to a gunfight
+	return ..()
+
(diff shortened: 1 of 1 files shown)


neeshacark:
i think given the stamina damage thing it blocking projectiles might be fine since its generally intended for pacifists who cant really counter ranged attacks but thats just me

00ze-cyclone:
> i think given the stamina damage thing it blocking projectiles might be fine since its generally intended for pacifists who cant really counter ranged attacks but thats just me

Regular crew in general doesn’t get much to deal with a ranged opponent and the most common ranged weapon onboard is the disabler which is non lethal.

neeshacark:
> > i think given the stamina damage thing it blocking projectiles might be fine since its generally intended for pacifists who cant really counter ranged attacks but thats just me
> 
> Regular crew in general doesn’t get much to deal with a ranged opponent and the most common ranged weapon onboard is the disabler which is non lethal.

actually the most common ranged weapon is throw mode which pacifists cant do

Moltijoe (approved):

PR #21303, opened 2024-01-23
Live xeno aliens on station level now generate research

ynot01:
Every life tick, xenos add 5*seconds points to the generic techweb

Incentive (besides grief) to have xenos in xenobiology containment

# Testing

# Wiki Documentation
5 research per second per alien
Default passive research is ~60/s
Would need 12 aliens on station to double research rate

# Changelog

:cl:  
tweak: Live xeno aliens on station level now passively generate research points
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/alien/alien.dm
+++ b/code/modules/mob/living/carbon/alien/alien.dm
@@ -147,3 +147,8 @@ Des: Removes all infected images from the alien.
 
 /mob/living/carbon/alien/can_hold_items()
 	return has_fine_manipulation
+
+/mob/living/carbon/alien/Life(seconds_per_tick, times_fired)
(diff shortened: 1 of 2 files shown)


MajManatee:
the station may be completely overrun but boy can we research quickly :)

bruhlookatthisdood:
shouldn't they only generate research while in xenobiology containment? or crew could see a spike in research and possibly know what is happening.

Yarinoi:
agreed, plus it doesnt make any sense to be able to get research from them while they're escaped, and might lead to even faster mech response
other then that, great idea!

ToasterBiome (changes requested):
show testing

ToasterBiome (approved):

PR #21308, opened 2024-01-24
allows you to extract a new reagent snake venom by milking snakes part 3

boodaliboo:
gives snakes a cool reason to be bought
use a bucket to extract venom from a snake
and adds a new snake venom reagent that does 2 toxin damage
i have opened this pr before it was only closeed because apparently milking isnt the correct term. despite that being the correct term used by proffesional snake owners milking them for antivenom.

another idea i have is implement antivenom a decently strong toxin healer that can only be made with snake venom does that sound like something that should be added to the pr

BUT I AM GOING TO ASK THE COMMUNITY MILKING OR EXTRACTING

image

🆑
rscadd: Added milking snakes for toxin and snake venom
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/snake.dm
+++ b/code/modules/mob/living/simple_animal/friendly/snake.dm
@@ -34,6 +34,7 @@
 	mob_size = MOB_SIZE_SMALL
 	mob_biotypes = MOB_ORGANIC|MOB_BEAST|MOB_REPTILE
 	gold_core_spawnable = HOSTILE_SPAWN
+	var/obj/item/udder/snake/udder
 	can_be_held = TRUE
 	worn_slot_flags = ITEM_SLOT_NECK
 	obj_damage = 0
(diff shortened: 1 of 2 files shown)


00ze-cyclone:
Tbh it should just be basic toxin instead of another new snowflake poison that does nothing other than give toxin damage. Especially since the base toxin is not really obtainable other than with the medvendor.

boodaliboo:
> Tbh it should just be basic toxin instead of another new snowflake poison that does nothing other than give toxin damage. Especially since the base toxin is not really obtainable other than with the medvendor.

wouldnt giving snake venom a crafting recipe that is a strong antitoxin more unique but im willing to change it to toxin if thats wanted

code/modules/mob/living/simple_animal/friendly/snake.dm
```diff
@@ -34,6 +34,7 @@
 	mob_size = MOB_SIZE_SMALL
 	mob_biotypes = MOB_ORGANIC|MOB_BEAST|MOB_REPTILE
 	gold_core_spawnable = HOSTILE_SPAWN
+	var/obj/item/udder/snake/udder
```
AMyriad, line 37: I would suggest changing this to something like the snake's fang, as that's actually what's being interacted with. Not...udders. Bro...

code/modules/reagents/chemistry/reagents/toxin_reagents.dm
```diff
@@ -1036,3 +1036,10 @@
 /datum/reagent/toxin/ambusher_toxin/on_mob_end_metabolize(mob/living/L)
 	L.remove_movespeed_modifier(type)
 	textShown = FALSE
+
+/datum/reagent/toxin/snakevenom
+	name = "Snake Venom"
+	description = "A potent venom extracted from a live snake"
+	color = "#792300" // rgb: 121, 35, 0
+	toxpwr = 2.5
```
AMyriad, line 1044: I agree with the toxin idea. There's in essence no difference between the two chemicals, and there's no other way to obtain it currently besides the limited supply of certain vendors. This would be an interesting way to produce this chemical.

AMyriad (changes requested):
I would very much suggest this:

ToasterBiome:
reopen when compile

PR #21327, opened 2024-01-29
Increases the value of limb loss/skeletonization in brazil & should prevent people getting stuck

SomeguyManperson:

Increases value of limb loss from 5 to 7 marbles, allowing 2 limbs to make up most of an escape
Increases value of limb skeletonization from 1 to 3 marbles since brain damage is already worth 1

Also switches the breaks in the limb modifying effects with returns, in the case someone does a silly they shouldn't end up trapped by a runtime.

Increases the number of viable marble combinations that aren't dying

# Wiki Documentation

limb loss in brazil is now worth 7 marbles
limb skeletonization in brazil is now worth 3 marbles

# Changelog

:cl:  
tweak: limb loss in brazil is now worth 7 marbles
tweak: limb skeletonization in brazil is now worth 3 marbles
bugfix: If you somehow manage to run out of limbs while leaving brazil, you will be ejected
/:cl:

### DIFF
--- a/code/modules/antagonists/eldritch_cult/eldritch_effects.dm
+++ b/code/modules/antagonists/eldritch_cult/eldritch_effects.dm
@@ -1,7 +1,7 @@
 #define PENANCE_LIFE "Lose your life (10 marbles)"
 #define PENANCE_SOUL "Lose your soul (14 marbles)"
-#define PENANCE_LIMB "Lose a limb (5 marbles)"
-#define PENANCE_SKELETON "Lose your flesh (1 marbles)"
+#define PENANCE_LIMB "Lose a limb (7 marbles)"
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/eldritch_cult/eldritch_effects.dm
```diff
@@ -378,8 +378,8 @@
 					var/obj/item/bodypart/BP
 					while(!BP)
 						if(!LAZYLEN(unspooked_limbs))
-							message_admins(span_notice("Someone managed to break brazil limb sacrificing stuff tell theos"))
```
Moltijoe, line 381: o7

Moltijoe (approved):

PR #21335, opened 2024-01-30
adds u to netspeak filter

ktlwjec0:
makes u change to you

removes netspeak from yogstation we hate netspeak here

# Testing
yes

:cl:  ktlwjec
spellcheck: u added to minor filter
/:cl:

### DIFF
--- a/config/minor_filter.txt
+++ b/config/minor_filter.txt
@@ -46,5 +46,6 @@ pls=please
 plz=please
 \btf\b=the fuck
 \bi\b=I
+\bu\b=you
 auxmos=La Li Lu Le Lo
 aux mos=La Li Lu Le Lo


AMyriad:
would be good if we had "np" on there as well

Aquizit:
I don't think single letters should be in the filter. While yes it's something that's commonly used netspeak, people like to put spaces between letters in words at times to try to portray shock or whatever, and this kinda messes with that god forbid it has a U in the word.

Np should be fine it's just the single letters I don't think is a good idea.

JohnFulpWillard:
I am pretty sure TG has this in their filter and I heard that stammering can trigger it, so if you're stammering "sup" it can come out as "s-s-s-u-u-u-p" which would then get caught by filter for using u, is that an issue here?

Also is ``"can u murdrbon with hijak?"`` and ``"((Wat does it mean when u talk like dis))"`` from brain trauma gonna trigger it?

AMyriad:
> I am pretty sure TG has this in their filter and I heard that stammering can trigger it, so if you're stammering "sup" it can come out as "s-s-s-u-u-u-p" which would then get caught by filter for using u, is that an issue here?
> 
> Also is `"can u murdrbon with hijak?"` and `"((Wat does it mean when u talk like dis))"` from brain trauma gonna trigger it?

So I tested this out of curiosity and yeah, "u-u-u-u" will result in "you-you-you-you". That being said, the way stuttering works seems to prevent the U from being repeated in that manner, so this would seemingly only affect intentional stuttering.

Didn't manage to test the trauma thing but I'd wager it'd also get filtered out.

ktlwjec0:
ok nvm that will just get annoying

PR #21357, opened 2024-02-04
Flying fang fixes and leap nerf

Mqiib:
The leap now goes as follows:
1: On hitting the target, the target is knocked down for 10 seconds
2: If the lizard projectile is blocked, the lizard is knocked down for 10 seconds.
3: If they are not blocked, the target is stunned for 6 seconds

Immobilizes for the leap were removed because they didn't apply to the lizard anyways and not blocking the leap stuns so that's also irrelevant.

The headbutt no longer infinitely stacks and you cant abuse air shoes because leaping unbuckles you now hahaha no more cheese.

Disarm chain defaults to harm intent when prone so you don't immediately get tail slapped to death even if you DO block the leap.

Blocking a lizard with a shield now actually punishes them. It won't stop you from being knocked down but you will still be able to act and they won't be able to use their bite combo

# Testing
thumbs up emoji

# Wiki Documentation

Blocking the lizard projectile actually helps.

# Changelog

:cl:  
bugfix: Flying fang inaccuracy debuff now actually goes away fully instead of stacking infinitely 
bugfix: Leaping with air shoes makes you fall over
tweak: Blocking the flying fang leap helps significantly but won't totally save you 
tweak: Disarm chain defaults to basic harm intent attacks when prone
/:cl:

### DIFF
--- a/code/datums/martial/flying_fang.dm
+++ b/code/datums/martial/flying_fang.dm
@@ -240,20 +240,17 @@
 					blocked = TRUE
 			L.visible_message("<span class ='danger'>[A] pounces on [L]!</span>", "<span class ='userdanger'>[A] pounces on you!</span>")
 
-			//Knockdown regardless of blocking, immobilize reduced significantly on block
+			//Knockdown regardless of blocking,
 			L.Knockdown(10 SECONDS)
-			L.Immobilize(1 SECONDS)	
 
-			//Blocking knocks the lizard down too and immobilizes longer
+			//Blocking knocks the lizard down too
 			if(blocked)
 				A.SetKnockdown(10 SECONDS)
-				A.Immobilize(2 SECONDS)
 			
-			//Otherwise the not-blocker gets stunned and the lizard is okay but stops for a moment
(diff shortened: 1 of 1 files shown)


code/datums/martial/flying_fang.dm
```diff
@@ -135,7 +135,7 @@
 	D.blur_eyes(4)
 	if(!istype(D.head, /obj/item/clothing/head/helmet))
 		D.dna.species.aiminginaccuracy += 25
-		addtimer(CALLBACK(src, PROC_REF(remove_bonk), D), 10 SECONDS, TIMER_UNIQUE|TIMER_OVERRIDE)
+		addtimer(CALLBACK(src, PROC_REF(remove_bonk), D), 10 SECONDS, 0)
```
Moltijoe: Just remove the parameter entirely

MajManatee:
my brother in Christ, I don't think it matters if you have a shield when a 8 foot tall, 300 pound pile of scales and teeth is flying at you at top speed.

JamieD1:
I hate CQC and think it should be removed

JamieD1 (approved):

PR #21361, opened 2024-02-05
Nerfs Adrenaline Sac & Resonant Shriek (real)

cowbot92:

Adjusts the costs of both abilities to 75 chemicals,
Adjusts the DNA cost of Adrenal sacs to 3 DNA points.
Reduces the effects of the confusion & jitter for resonant shriek to 10 seconds each

These abilities are really strong, and should be more costly to use.

# Testing

numbers changes lois

# Wiki Documentation

Change the costs on the wiki

# Changelog

:cl:  

tweak: Tweaks changeling abilities

/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/adrenaline.dm
+++ b/code/modules/antagonists/changeling/powers/adrenaline.dm
@@ -1,6 +1,6 @@
 /datum/action/changeling/adrenaline
 	name = "Adrenaline Sacs"
-	desc = "We evolve additional sacs of adrenaline throughout our body. Costs 50 chemicals."
+	desc = "We evolve additional sacs of adrenaline throughout our body. Costs 75 chemicals."
 	helptext = "Removes all stuns instantly. Can be used while unconscious. Continued use poisons the body." //yogs - changed text to suit the below change
 	button_icon_state = "adrenaline"
 	chemical_cost = 75
--- a/code/modules/antagonists/changeling/powers/adrenaline.dm
+++ b/code/modules/antagonists/changeling/powers/adrenaline.dm
@@ -3,8 +3,8 @@
 	desc = "We evolve additional sacs of adrenaline throughout our body. Costs 50 chemicals."
 	helptext = "Removes all stuns instantly. Can be used while unconscious. Continued use poisons the body." //yogs - changed text to suit the below change
 	button_icon_state = "adrenaline"
-	chemical_cost = 50
-	dna_cost = 2
+	chemical_cost = 75
+	dna_cost = 3
 	req_human = 1
 	req_stat = UNCONSCIOUS
 	conflicts = list(/datum/action/changeling/strained_muscles)
--- a/code/modules/antagonists/changeling/powers/shriek.dm
+++ b/code/modules/antagonists/changeling/powers/shriek.dm
@@ -1,6 +1,6 @@
 /datum/action/changeling/resonant_shriek
 	name = "Resonant Shriek"
-	desc = "Our lungs and vocal cords shift, allowing us to briefly emit a noise that deafens and confuses the weak-minded. Costs 60 chemicals."
+	desc = "Our lungs and vocal cords shift, allowing us to briefly emit a noise that deafens and confuses the weak-minded. Costs 75 chemicals."
 	helptext = "Emits a high-frequency sound that confuses and deafens humans, blows out nearby lights and overloads cyborg sensors."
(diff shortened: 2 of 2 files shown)


LassiH:
over half your chems for the only way to counter stuns

Wolf-751:
would this not make adrenaline sacs useless? also double nerf on resonate scream will practically kill the skill off completely.

cowbot92:
> over half your chems for the only way to counter stuns

over half you chems to become immune to stuns, reverse any stuns you currently have, wake up from being slept, and use methspeed to escape? yeah.

cowbot92:
> would this not make adrenaline sacs useless? also double nerf on resonate scream will practically kill the skill off completely.

why would you think it makes it useless, when it doesn't affect anything about what it actually does? And I don't think you realize how long 10 seconds of confusion and jitter are.

Therandomhoboo:
Numbers way too high for the results they give tbf. 75 chems is a death sentence, i would rather you just add a CD to it to offset this

Wolf-751:
> > would this not make adrenaline sacs useless? also double nerf on resonate scream will practically kill the skill off completely.
> 
> why would you think it makes it useless, when it doesn't affect anything about what it actually does? And I don't think you realize how long 10 seconds of confusion and jitter are.

its more the fact of increased cost to 3 which is huge and making 75 chems for 1 stun break is kinda bad since sec will just chain stun you now with even less tools to counter it because you basically have only 7 points left to buy anything.
The increase in chem cost on the scream also makes it so your limited point pool that it becomes far better to run emp scream offering larger utility. 
Changeling is a strong solo antagonist i will admit but perhaps this might be abit heavy handed since they cannot even do the bullshit of running sac and leg muscles anymore.

cowbot92:
> > > would this not make adrenaline sacs useless? also double nerf on resonate scream will practically kill the skill off completely.
> > 
> > 
> > why would you think it makes it useless, when it doesn't affect anything about what it actually does? And I don't think you realize how long 10 seconds of confusion and jitter are.
> 
> its more the fact of increased cost to 3 which is huge and making 75 chems for 1 stun break is kinda bad since sec will just chain stun you now with even less tools to counter it because you basically have only 7 points left to buy anything. The increase in chem cost on the scream also makes it so your limited point pool that it becomes far better to run emp scream offering larger utility. Changeling is a strong solo antagonist i will admit but perhaps this might be abit heavy handed since they cannot even do the bullshit of running sac and leg muscles anymore.

You shouldn't be able to get out of chainstuns anyways, smart use of this ability allows for fairer play other than "Oh I got knocked down every time let me just run away" as if you had an infinite adrenal implant, and resonant shriek and dissonant shriek do not covet the same niche,

cowbot92:
> Numbers way too high for the results they give tbf. 75 chems is a death sentence, i would rather you just add a CD to it to offset this

When was the last time you played changeling?

Marmio64:
anything but DNA cost increases. People are always going to keep using adrenaline sacs so long as it keeps doing what it does, because it's so good it is nearly a requirement since stuns and stamina damage are so prevalent. However, a DNA cost increase reduces the variety available to the ling and is further going to push them into 'meta abilities only' because they can't afford to take anything else but the best. If you do want to nerf the DNA cost still, maybe have two versions of the ability instead, with a far weaker but less DNA expensive one, and then a stronger but more DNA expensive one?

LassiH:
> over half you chems to become immune to stuns, reverse any stuns you currently have, wake up from being slept, and use methspeed to escape? yeah.

i dont think changeling adrenaline makes you immune to stun only reverse the one you currently have

cowbot92:
> > over half you chems to become immune to stuns, reverse any stuns you currently have, wake up from being slept, and use methspeed to escape? yeah.
> 
> i dont think changeling adrenaline makes you immune to stun only reverse the one you currently have

image
image

LassiH:
is the code broken as it definetely does not feel like changeling adrenaline prevents stuns after ability usage

Therandomhoboo:
> > Numbers way too high for the results they give tbf. 75 chems is a death sentence, i would rather you just add a CD to it to offset this
> 
> When was the last time you played changeling?

A while, but I still think the ling ability shift has kinda hindered them imo. Personally would still think a CD would make it better than more chems.

MajManatee (approved):
BC said it was fine. Could require a revisit if that confusion PR goes through and ruins it completely.

PR #21376, opened 2024-02-09
Reworks confusion to be less confusing

Moltijoe:
Changes confusion to no longer scale up to completely randomized movement
Now it has a 50% chance on movement to make you walk +- 45 degrees of where you intended to
and a 25% chance on movement to make you walk +- 90 degrees if the remaining duration is above 10 seconds

Confusion is too unreliable to be used in balance, it goes from minor movement fuckery to effective stun with the difference of a second or so

# Testing
Gotta

:cl:  
tweak: Confusion modified to not scale up to completely randomized movement
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs/confusion.dm
+++ b/code/datums/status_effects/debuffs/confusion.dm
@@ -33,3 +33,4 @@
 		source = get_step(owner, new_dir)
 		direction = new_dir
 
+#undef CONFUSION_STRONG_THRESHOLD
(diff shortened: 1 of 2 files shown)


CrazyDrakkon:
huh?

MajManatee:
+- 45 degrees means I am still guaranteed to move left if I hold left, it just might move me up and left or down and left.
Doesn't that make it literally useless now?

Id be more on board if it didn't change it from 90 (a change which isn't really documented either...)

JamieD1 (approved):

PR #21397, opened 2024-02-12
makes the seismic arm's kit into a martial art and slightly buffs it further

LazennG:

seismic arm moves are now all a martial art a la buster arm, the damage on all moves besides the charged up one now has 0 AP and the charged up one can now deal 40 damage to objects and deals 30 to whatever the initial projectile crashes into.
-dash moves have 1 less range than before 
- also ramming targets into all dense objects will deal a bit of damage now instead of just ore. 
- the martial art can be toggled with the deactivate and recalibrate buttons in the respective tab.

makes the arm more fluid to use and makes the big move not as repugnant

# Testing
confirmed that removal of the limb removes the art and that the moves all work without sending targets through walls and that the buttons in the tab work.

# Wiki Documentation

https://wiki.yogstation.net/wiki/Shaft_Miner#Necropolis_Chest_Loot
gonna try rewording this based on the existing entry on the wiki:
A cybernetic right arm that functions and acts like a cyborg right arm. However, having this augment or replace your right arm will grant you a martial art replacing intents with the following abilities:
Lariat (Grab intent): Dash 4 spaces and deal 50 Brute to any fauna you hit. People take 10 Brute, and silicons take 12 Brute. 7-second cooldown.
Rush (Disarm intent): Dash 4 spaces and carry any target you connect with you. For every tile you carry a fauna, they take 20 Brute and you gain 90% damage reduction for the duration. People take 4 Brute per tile, and silicons take 5 true Brute per tile. In addition, you don't gain any damage reduction without carrying an animal. Any fauna killed by this is gibbed. 7-second cooldown.
Suplex(Harm intent): Moves a target directly in front of you behind you. Fauna take 20 Brute, people take 10 Brute, and silicons take 12 Brute. If the target is moved onto a fauna or a person, the latter takes the same damage depending on what they are. Any fauna killed by the targeted suplex is gibbed. 1-second cooldown.
Right Hook(Self targeted help intent): Charge for 2 seconds to power up your arm. The charge lasts 5 seconds. While it's charged, hitting a target will send them flying eight spaces. Fauna take 100 true Brute, people take 15 true Brute, silicons take 18 true Brute, and objects take 40 damage. If another mob is in the path of the target's flight, they will take 30 true Brute, 10 true Brute, 11 true Brute, or 30 damage if they are a fauna/person/silicon/object, respectively. This move disables the seismic right arm for 15 seconds, making every other move unusable until then.
Rush, Suplex, and Right Hook can send targets through rocks, with the first 2 dealing 5 true brute damage when slamming targets into dense objects and walls.

# Changelog

:cl:  @Chubbygummibear @LazennG 
rscadd: Added new martial art for seismic arm  
rscdel: Removed old seismic arm actions  
/:cl:


code/datums/martial/reverbpalm.dm
```diff
@@ -0,0 +1,279 @@
...
+		return
+	if(H.a_intent == INTENT_DISARM)
+		rush(H)
+	if(H.a_intent == INTENT_HARM && isliving(target))
+		suplex(H,target)
+	if(H.a_intent == INTENT_GRAB)
+		lariat(H)
+
+/datum/martial_art/reverberating_palm/harm_act(mob/living/carbon/human/A, mob/living/D)
+	return TRUE // no punching plus slamming please
```
Moltijoe: this will prevent punching even if the arm is turned off

Moltijoe (changes requested):

code/datums/martial/reverbpalm.dm
LazennG: ah shit thanks

code/datums/martial/reverbpalm.dm
```diff
@@ -94,6 +94,10 @@
 	if(H.a_intent == INTENT_GRAB)
 		lariat(H)
 
+/datum/martial_art/reverberating_palm/harm_act(mob/living/carbon/human/A, mob/living/D)
+	if(usr.click_intercept == null)
+		return FALSE // no punching plus slamming please
```
Moltijoe: the default return is null, which in byond is equivalent to FALSE
meaning this statement will never return true

code/datums/martial/reverbpalm.dm
```diff
@@ -0,0 +1,286 @@
...
+
+/datum/martial_art/reverberating_palm/harm_act(mob/living/carbon/human/A, mob/living/D)
+	if(usr.click_intercept == null)
+		return FALSE // no punching plus slamming please
+
+/datum/martial_art/reverberating_palm/proc/supercharge(mob/living/user)
+	if(!COOLDOWN_FINISHED(src, next_palm))
+		to_chat(user, span_warning("You can't do that yet!"))
+		return
+	COOLDOWN_START(src, next_palm, COOLDOWN_RPALM)
```
Moltijoe, line 104: with the cooldown being here, it means if they accidentally click themselves, they'll waste the cooldown even if they cancel the charging, just want to make sure that's alright with you

cowbot92:
seismic style rises

code/datums/martial/reverbpalm.dm
LazennG: thanks for catching that, changed it to be much shorter so they dont have to wait 15 seconds for whiffing

JamieD1 (approved):

PR #21405, opened 2024-02-13
[Admin] Cryoing as a Valentine no longer tells you to ahelp

Aquizit:

As the title states, you not longer get a warning message before cryoing if you are a Valentine.

Valentine's aren't *real* antags. If they wanna go they should be able to go.

# Testing

I tested it, it works.

# Changelog

:cl:  
tweak: No longer told to ahelp when cryoing as a valentine
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -413,8 +413,11 @@ GLOBAL_VAR_INIT(cryopods_enabled, FALSE)
 			tgui_alert(target, "You're a Head of Staff![generic_plsnoleave_message]")
 			caught = TRUE
 		if(A)
-			tgui_alert(target, "You're a [A.name]![generic_plsnoleave_message]")
-			caught = TRUE
+			if(A.name == "valentine")
+				caught = TRUE
+			else
+				tgui_alert(target, "You're a [A.name]![generic_plsnoleave_message]")
+				caught = TRUE
 		if(caught)
 			target.client.cryo_warned = world.time
 			return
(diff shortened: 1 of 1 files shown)


Hislyna:
Every year. Every damn year something pops up with valentines.

code/game/machinery/cryopod.dm
```diff
@@ -413,8 +413,11 @@ GLOBAL_VAR_INIT(cryopods_enabled, FALSE)
 			tgui_alert(target, "You're a Head of Staff![generic_plsnoleave_message]")
 			caught = TRUE
 		if(A)
-			tgui_alert(target, "You're a [A.name]![generic_plsnoleave_message]")
-			caught = TRUE
+			if(A.name == "valentine")
```
iamthesilly: You should really add a variable on `/datum/antagonist` for whether the cryo message shows up and check for that.

code/game/machinery/cryopod.dm
```diff
@@ -413,8 +413,11 @@ GLOBAL_VAR_INIT(cryopods_enabled, FALSE)
...
+				caught = TRUE
```
iamthesilly: You don't want this to be set to true here, or it'll still block from using cryo once without telling you why.

SapphicOverload (changes requested):

Aquizit:
tested, works

JamieD1:
Shitty Valentines mode. needs to die its dumb, boring and should be opt in

PR #21417, opened 2024-02-15
Replaces old ID console from CE office on icemeta with a modular one

AMyriad:
old
new

Old console < new console

# Testing
Yes but there's a node again so all is well
image

# Changelog
:cl:  
mapping: Replaced the old ID console from the CE's office on icemeta with the newer modular console
/:cl:


JamieD1:
The nodes is so you could monitor the power network

AMyriad:
> The nodes is so you could monitor the power network

you know what, you're right I'm dumb
will fix this momentarily

JamieD1 (approved):

PR #21419, opened 2024-02-15
Fixes powerless mechs being immune to EMPs

Mqiib:
Now they still take damage regardless of cell power since they can move regardless of cell power.

Ion spam actually kills mechs fr fr on god

# Testing

They both took damage!
image

# Changelog

:cl:  

bugfix: Mechs without any power are no longer immune to EMPs

/:cl:


ofscarletdreams:
how will the mech meta recover from this

ToasterBiome (approved):

PR #21423, opened 2024-02-16
Removes ability to choose a gas canister label that breaks the 4th wall

AMyriad:
When selecting a cool label/color palette for your gas canister, this is an available option
image
Very cool, but examining this yields this description
image

SO

I just made a new can obj for the list to pull from. Admin cans retain their OOCish description, as they shouldn't be spawned in any serious circumstance, and players keep their cool label option. Best of both worlds

# Testing
Works
image
image

# Changelog
:cl:  
bugfix: Players can no longer relabel gas canisters to a variant that's intended for admins and breaks the 4th wall
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/portable/canister.dm
+++ b/code/modules/atmospherics/machinery/portable/canister.dm
@@ -43,7 +43,7 @@
 		"generic striped" = /obj/machinery/portable_atmospherics/canister/generic/stripe,
 		"generic hazard" = /obj/machinery/portable_atmospherics/canister/generic/hazard,
 		"caution" = /obj/machinery/portable_atmospherics/canister,
-		"danger" = /obj/machinery/portable_atmospherics/canister/fusion_test,
+		"danger" = /obj/machinery/portable_atmospherics/canister/fusion,
 		"n2" = /obj/machinery/portable_atmospherics/canister/nitrogen,
 		"o2" = /obj/machinery/portable_atmospherics/canister/oxygen,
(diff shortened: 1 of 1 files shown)


Yarinoi:
nah change it looks cool asf

iamthesilly:
> nah change it looks cool asf

It's the same sprite as the antinoblium canister, which is an obtainable gas

azzzertyy:
thumbs downers dont know what theyre talking about

AMyriad:
apparently sapphic is right yeah
the antinob has a nearly identical design to this one so keeping it exclusive to the admin-spawned can of death (as I think it was intended to be) isn't as much of a loss as I thought
image

ShoziOffical:
just make a different description then instead of removal of the label. It was intentional in the first place, i put the fusion label to identify fusion canister so people know its bad to touch but i forgot to change the description

ShoziOffical:
> > nah change it looks cool asf
> 
> It's the same sprite as the antinobliun canister, which is an obtainable gas

its not the same, antinob can has white stripe

AMyriad:
Alright checks out, good to know what was actually intended and what wasn't
image

So instead of changing the description or removing it from the list outright, I created a can type that exists purely for the icon and has a proper IC description. Many other canisters do this, so I don't think there'll be a problem.
PR sections updated as of this comment to reflect this btw ^

Yarinoi:
> > nah change it looks cool asf
> 
> It's the same sprite as the antinobliun canister, which is an obtainable gas

nvm then lol

ToasterBiome (approved):

PR #21429, opened 2024-02-17
No more lowpop spacevines that kill everyone pls

Mqiib:
30 pop is good enough for now probably. Also I hate flowering vines.

# Testing

Number change

# Wiki Documentation

Min pop of 30

# Changelog

:cl:  

tweak: Min pop for space vines increased to 30

/:cl:

### DIFF
--- a/code/modules/events/spacevine.dm
+++ b/code/modules/events/spacevine.dm
@@ -3,7 +3,7 @@
 	typepath = /datum/round_event/spacevine
 	weight = 15
 	max_occurrences = 3
-	min_players = 10
+	min_players = 30
 
 /datum/round_event/spacevine
 	fakeable = FALSE


LazennG:
the issue this pr sprung from was vines becoming infinitely unbeatable due to expanding onto icemeta turfs instead of the 30 ghost spawns that came from that. i think you should change the spread() proc to address this instead of making it so the exact same thing happens on highpop

Mqiib:
> the issue this pr sprung from was vines becoming infinitely unbeatable due to expanding onto icemeta turfs instead of the 30 ghost spawns that came from that. i think you should change the spread() proc to address this instead of making it so the exact same thing happens on highpop

A valid argument but it's rather late and I've seen vines on normal stations do similar things when it's lowpop and they get to spread infinitely.

LazennG:
> > the issue this pr sprung from was vines becoming infinitely unbeatable due to expanding onto icemeta turfs instead of the 30 ghost spawns that came from that. i think you should change the spread() proc to address this instead of making it so the exact same thing happens on highpop
> 
> A valid argument but it's rather late and I've seen vines on normal stations do similar things when it's lowpop and they get to spread infinitely.

cant always win

LassiH:
it barely reaches 30 during populated days now maybe make it 20

Hislyna:
Ided + skill issue, make weaker vines that don't spawn as many traps on lowpop

Moltijoe:
balance council:
>The issue seems to be in venus human traps, do something about them, not vines as a whole

PR #17368, opened 2023-01-06
TGUI Preferences Menu + total rewrite of the preferences backend

ThatLing:
# Document the changes in your pull request
So this, is a big one and will break a lot of stuff.

Mainly ports https://github.com/tgstation/tgstation/pull/61313, which a whole bunch of extra stuff because we are so outdated. Currently still heavily WIP, so no screenshots yet. If you want to know what it looks like, go to the original PR.

Full list of ported PRs, credits to the original authors:
- https://github.com/tgstation/tgstation/pull/58483
- https://github.com/tgstation/tgstation/pull/58593 (partly)
- https://github.com/tgstation/tgstation/pull/59914
- https://github.com/tgstation/tgstation/pull/61313
- https://github.com/tgstation/tgstation/pull/61542
- https://github.com/tgstation/tgstation/pull/61586
- https://github.com/tgstation/tgstation/pull/61764
- https://github.com/tgstation/tgstation/pull/61942
- https://github.com/tgstation/tgstation/pull/61957
- https://github.com/tgstation/tgstation/pull/61980 (TODO)
- https://github.com/tgstation/tgstation/pull/62845
- https://github.com/tgstation/tgstation/pull/63200
- https://github.com/tgstation/tgstation/pull/63370
- https://github.com/tgstation/tgstation/pull/63503
- https://github.com/tgstation/tgstation/pull/63554 (partly)
- https://github.com/tgstation/tgstation/pull/64610
- https://github.com/tgstation/tgstation/pull/65140
- https://github.com/tgstation/tgstation/pull/65645
- https://github.com/tgstation/tgstation/pull/66317
- https://github.com/tgstation/tgstation/pull/69454
- https://github.com/tgstation/tgstation/pull/69635
- https://github.com/tgstation/tgstation/pull/69852
- https://github.com/tgstation/tgstation/pull/70577

# Changelog

:cl: Mothblocks & a lot of people from /tg/
rscadd: The preferences menu has been completely rewritten in tgui.
tweak: The "Preferences" tab has been removed, the menu verbs have been moved to the OOC tab
tweak: The "Stop Sounds" verb has been moved to OOC.
/:cl:

PR #21455, opened 2024-02-19
Maybe fixes issue #21451 with quirks

### LINKED ISSUE #21451: Quirks are initialized before your character is, which is what breaks them
## Test Merges: 
Not Supplied / None
## Reproduction:
Description: The initialization order for quirks is earlier than it is supposed to be due to an undocumented change to code/__DEFINES/subsystems.dm in #17368. The round ID has nothing to do with this as its not an issue that can be seen in the replays but its required.

 Submitted by: masterdrake (598557108551811073)

Mqiib:
Supposedly quirks initialize before characters which breaks stuff and this happened in #17368 
So theoretically this will fix https://github.com/yogstation13/Yogstation/issues/21451 I dunno cowbot told me to do this.

# Changelog @cowbot92 

:cl:  

bugfix: Possibly fixes quirks breaking roundstart

/:cl:


Moltijoe:
doesn't work :sadge:

PR #21469, opened 2024-02-20
hopefully fix adminwho by skipping nulls in the admin list

Chubbygummibear:

we should probably switch to GLOB.admins like tg instead of GLOB.permissions.admins, but for the time being this should fix the issue
image

people don't think they're safe from admins

# Testing
can't test privately because i don't have the whole admin system backend with mentors and maintainers who are like pseudo admins

# Changelog

:cl:  
bugfix: fix adminwho by skipping null clients in the admin list
/:cl:


adamsong (approved):
Feels kinda like a hack, but this is exactly how I would have done it lol

PR #15579, opened 2022-09-02
Syndicate Objective Manifestation System (SOMS)

ynot01:
# Document the changes in your pull request

Adds a new category to the uplink with 4 options:

- Request Objective (RANDOM)
- Request Objective (EASY)
- Request Objective (MEDIUM)
- Request Objective (HARD)

Easy objectives will reward 2-3 TC
Medium objectives will reward 3-5 TC
Hard objectives will reward 5-7 TC

<details>
  <summary>Easy Objectives</summary>

- Kill a pet
- Steal 28 moles of plasma (full tank)
- Steal an unused slime extract

</details>

<details>
  <summary>Medium Objectives</summary>

- Kill someone once
- Steal a reflective jacket
- Steal a hand teleporter
- Steal CE magboots
- Steal a medal of captaincy
- Steal the hypospray
- Steal reactive tele armor
- Steal secret documents

</details>

<details>
  <summary>Hard Objectives</summary>

- Kill an AI
- Steal antique laser gun
- Steal HoS gun
- Steal captain's jetpack
- Steal nuke disk
- Steal an AI
- Steal the blueprints
- Steal the blackbox

</details>

Each option has a 50% chance of appearing/not appearing and only have one use.

"Purchasing" one of these will notify online admins and start a 1 minute timer.

If an admin is online, they will have the option to override your objective with their own custom objective, and they will be able to choose their own difficulty rating.

Once the 1 minute timer expires or an admin creates an objective, you will be notified.

Your objective will come in form of a folder of a random color on a random safe turf in the station where no one can currently see it. It is up to you to get to it.

You will be told where and what color the folder is.

If another syndicate gets ahold of your folder, they will be able to complete the objective just as you would have. You can use this to steal or cooperate with others.

Once you have completed the objective, you should hit the Check Completion button. 

If the objective is complete, the folder will give you your reward and add the objective to your objectives list (so you don't abandon it after).

If the objective is incomplete, the folder will scold you.

If the objective is admin-handled, admins will be notified of your request and either confirm or deny the action. You will be alerted to their action by a beep and if they confirm, the objective will be marked as complete and you will be able to redeem the objective as normal.

## Visual Aid

image

1 minute later...

# Changelog

:cl:  
experimental: Added objective requesting system for Syndicates
bugfix: Fixed a bunch of objectives runtiming if they couldn't find what they wanted
tweak: Made caplaser objective 10 difficulty from 5
tweak: Made captain's jetpack objective 10 difficulty from 5
[...]

PR #21471, opened 2024-02-20
Deletes syndicate bounties

Moltijoe:
Deletes syndicate bounties

It should be difficult for traitors to get over 20tc, this isn't a particularly interesting way for them to do it either
it's only really used by the "gamer" traitors trying to get insane setups only possible with more than 20tc which is not what we're wanting to encourage our antag players to do

# Testing
gotta

:cl:  
rscdel: Syndicate bounties
/:cl:


00ze-cyclone:
Syndicate bounties were the alternative to the absolute bs that were null crates.
They were less stupid but still stupid.
Maybe we can finally do the right thing now and instead add a few crates that DON’T give uplink shit for free.
This is a step in the right direction.

AMyriad:
wholly support this if #15579 gets merged

RG4ORDR:
We've come full circle.  
There were several ways of fixing null crates such as upping their cost, removing or reducing what items came in, and my personal favorite suggestion was adding a 50% chance that it would spawn a hostile General Griefsky in the box. The bounties were done as a compromise because it rewarded stealing and selling objectives that could be other traitor objectives and rewarding you for being more of an antagonist. This further got abused before the IAA rework where IAAs were doing bounties ontop of killing their targets and getting way to many TCs. Otherwise this was merely a way for creative traitors to get some extra TC because admins never answer hacked comms consoles.

TC bounties aren't mandatory for optimal traitor play but it's pretty clear that this is done because some players are once again doing nothing but min-maxxing their antag rounds via this. I dont remotely support the idea of removing this because a few players are once again causing issues by over doing it.

Yarinoi:
we should just ban people we dont like

Therandomhoboo:
There is honestly a simple way to stop all this.
Add a limit to how many bounties you can do? 
Like make that you can do x amount of easy ones, x amount of med, x amount of hard ones. I'd rather we think of ways to improve syndie bounties instead of just removing cause lazy

Moltijoe:
> There is honestly a simple way to stop all this. Add a limit to how many bounties you can do? Like make that you can do x amount of easy ones, x amount of med, x amount of hard ones. I'd rather we think of ways to improve syndie bounties instead of just removing cause lazy

The way to improve them is removing this and adding ynot's pr

Therandomhoboo:
> > There is honestly a simple way to stop all this. Add a limit to how many bounties you can do? Like make that you can do x amount of easy ones, x amount of med, x amount of hard ones. I'd rather we think of ways to improve syndie bounties instead of just removing cause lazy
> 
> The way to improve them is removing this and adding ynot's pr

Ah yes free bounties that come with your uplink, what is this Uno? I don't think its better to allow traitors roundstart potentially get more TC for mediocre missions really. At least with this you have to either go to supply, be a supply job role or build your own machine and still have to use the transport shuttle so you're stuck to one spot. Whereas... Ynots, its go to location you know about and hope sec dont meta it or traitors arent an asshole....

JamieD1 (approved):

PR #21493, opened 2024-02-22
HFR gas byproducts now uses scaled_production instead of fuel_consumption, and now scaled alongside with gas produced from moderator, code refactor

warface1234455:
HFR gas byproducts now uses scaled_production instead of fuel_consumption, and now scaled alongside with gas produced from moderator, code refactor

QoL
Byproducts uses ``fuel_consumption``  to produce gas which is weird and it produces very little amount of gas because of the var has been calculated to small numbers, it should be using ``scaled_production`` instead which is more accurate and can produce decent amount of gas. 

# Testing
Observing the HFR monitor i did notice that the byproducts increased when reaching certain fusion level 

# Wiki Documentation
no wiki

# Changelog

:cl:  

tweak: HFR gas byproducts now uses scaled_production instead of fuel_consumption, and now scaled alongside with gas produced from moderator, code refactor
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/fusion/hfr_main_processes.dm
+++ b/code/modules/atmospherics/machinery/components/fusion/hfr_main_processes.dm
@@ -254,8 +254,8 @@
 
 	var/add_remove_amount = round(scaled_production, 0.01) // gases on the same tier are produced at normal rate
 	for(var/gas_id in fuel.primary_products)
-			internal_fusion.adjust_moles(gas_id, add_remove_amount)
-			delta_fuel_list[gas_id] += add_remove_amount
+		internal_fusion.adjust_moles(gas_id, add_remove_amount)
+		delta_fuel_list[gas_id] += add_remove_amount
 
 	if(power_level < 1)
 		return // can't produce any gases, don't need to continue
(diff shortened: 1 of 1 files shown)


code/modules/atmospherics/machinery/components/fusion/hfr_main_processes.dm
```diff
@@ -251,19 +251,19 @@
...
 	// Which gases are produced depend on what the fusion level is.
 	var/list/tier = fuel.secondary_products
-	moderator_internal.adjust_moles(tier[power_level], round(scaled_production, 0.01)) // gases on the same tier are produced at normal rate
-	delta_mod_list[tier[power_level]] += round(scaled_production, 0.01)
+	var/add_remove_amount = round(scaled_production, 0.01) // gases on the same tier are produced at normal rate
+	moderator_internal.adjust_moles(tier[power_level], add_remove_amount)
+	delta_mod_list[tier[power_level]] += add_remove_amount
+	for(var/gas_id in fuel.primary_products)
+		internal_fusion.adjust_moles(gas_id, add_remove_amount)
+		delta_fuel_list[gas_id] += add_remove_amount
```
iamthesilly: The primary products section needs to be before the `power_level < 1` check or it doesn't work at level 0

SapphicOverload (changes requested):

code/modules/atmospherics/machinery/components/fusion/hfr_main_processes.dm
ShoziOffical: i dont think it does at level 0, even if it does it would be very little anyway

ShoziOffical: ill just change it then

SapphicOverload (approved):

PR #21506, opened 2024-02-24
Lets skittish people dive into crates properly, rather than smashing their face into it again and again

Moltijoe:
# Testing
image

:cl:  
bugfix: Skittish now works properly for crates
/:cl:


Yarinoi:
god bless
could you also make it so that it can be rebinded

JamieD1 (approved):

PR #21515, opened 2024-02-25
[BOX] Changes medbay doors unrestricted access

AMyriad:
This changes the silly arrows in the main medbay area and alters what doors have unrestricted access from the inside to allow patients to exit from both left and right entrances, rather than just the right one in a sort of "come in on the left, exit on the right" style.

Helps those patients who just finished getting treated trying to exit on the left side, which since the sleeper/stasis bed are there, a **lot** of people try to do that. I get the system of a designated entrance and designated exit, but it doesn't always work like that.

People will come in from, and exit to both sides, and I've seen on too many occasions people annoyingly can't exit from one side, but can from another. I know it's not hard to follow basic directions, but this change is for convenience sake, as those directions are arbitrary, and aren't going to be respected or followed in real situations.

# Testing
image

# Changelog
:cl:  
mapping: Patients can now exit from both left and right doors on Box medbay
/:cl:

### DIFF
--- a/_maps/map_files/YogStation/YogStation.dmm
+++ b/_maps/map_files/YogStation/YogStation.dmm
@@ -832,10 +832,6 @@
 /turf/open/floor/plasteel,
 /area/ai_monitored/security/armory)
 "afF" = (
-/obj/effect/turf_decal/arrows/white{
-	color = "#99ccff";
-	dir = 1
-	},
 /obj/effect/turf_decal/trimline/blue/filled/corner/lower{
 	dir = 1
 	},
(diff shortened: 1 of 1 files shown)


MajManatee:
ya dont drive on both sides of the road

ynot01:
is there no order anymore

JamieD1:
Considering I was a huge part of wanting this, cant people just not be dumb?

patpol4:
hugbox

AMyriad:
most of my reasoning is said in the "Why is this good for the game?" section so I won't repeat myself
do, however, bear in mind that most stations already do this

gax has this, despite what the arrow decals show:
image

icemeta does this:
image

and so does donut:
image

box actually stands out here, and it's a bit silly that they do :p

patpol4:
None of these have two entrances

neeshacark:
> None of these have two entrances

gax literally has two entrances

neeshacark (approved):

PR #21522, opened 2024-02-25
updates seismic arm with minor dmc reference (and fixes big bug oversight)

LazennG:

slamming bubblegum while she's charging will do 50 bonus damage to the base 20
also adds a proximity check to suplex because i might have forgotten to not let people get slam teleported telekinetically oops

fathersonbonding

i think adding some reward to the big risk of getting close to bubblegum while charging is fine, plus i had this in mind since the seismic arm's conception and finally wanted to get it out of my head

# Testing

(all from melee range and not 7 meters away btw!)

damage from suplexing while in motion
image

damage normally 
image

# Changelog
:cl:  
bugfix: added proximity check to suplex  
tweak: seismic arm suplex bonus damage vs charging bubblegum  
/:cl:

### DIFF
--- a/code/datums/martial/reverbpalm.dm
+++ b/code/datums/martial/reverbpalm.dm
@@ -146,10 +146,12 @@
 	if(isanimal(target))
 		if(istype(target, /mob/living/simple_animal/hostile/megafauna/bubblegum))
 			var/mob/living/simple_animal/hostile/megafauna/bubblegum/B = target
+			var/turf/D = get_step(Q, turn(user.dir,180))
 			if(B.charging)
 				B.adjustBruteLoss(50)
 				B.charging = FALSE
-				target.visible_message(span_warning("[B] is caught and slammed into the dirt!"))
+				B.forceMove(D)
+				B.visible_message(span_warning("[B] is caught and thrown behind [user]!"))
 				playsound(target, 'sound/effects/explosion1.ogg', 60, 1)
 				shake_camera(user, 1, 2)
 				return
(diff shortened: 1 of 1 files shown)


code/datums/martial/reverbpalm.dm
```diff
@@ -137,14 +137,22 @@
...
 		damagefilter(user, L, simpledam, persondam, borgdam)
 	if(!(crash(target, Q)))
 		target.forceMove(Z)
 	damagefilter(user, target, simpledam, persondam, borgdam)
 	if(isanimal(target))
+		if(istype(target, /mob/living/simple_animal/hostile/megafauna/bubblegum))
+			var/mob/living/simple_animal/hostile/megafauna/bubblegum/B = target
+			if(B.charging)
+				B.adjustBruteLoss(50)
+				B.charging = FALSE
```
Moltijoe: so, while you think this sounds cool, all it does is let bubblegum attack while doing that specific charge, it doesn't actually stop their triple charge attack, nor does it stop the charge itself

Moltijoe (changes requested):

code/datums/martial/reverbpalm.dm
Moltijoe: at least, i'm pretty sure, as there's no check for it to stop
unless there's weird synergy with the forcemove of suplexes

LazennG: initially kept the charging false thing incase something broke and bubblegum was giving away free damage bonuses but i shouldn't expect that to break when it's gone now. i can't stop the charge but suplexing now places bg two tiles behind the user so the charge path is typically disjointed with the user out of harm's way.

Moltijoe (approved):

PR #21524, opened 2024-02-25
Makes preternis night vision consume power

Moltijoe:
Preternis eyes will now consume power while their night vision is in use
Preternis eyes will only take damage in non powerhungry species when nightvision is turned on
They'll also be able to be put in an ethereal without losing power too

preternis eyes are particularly strong as they give free night vision with no downside
they used to eventually turn off if used enough, but that was changed at some point a while ago

This re-adds that by making them consume power while in use, and shutting down when the host gets particularly hungry

Also, making them only degrade when night vision is actively being used by non powerhungry species lets them actually have some use for those species

# Testing
gotta

:cl:  
tweak: Preternis eyes now also care about if a species is powerhungry instead of only robotic
tweak: Preternis eyes night vision now consumes power from powerhungry species
tweak: Preternis eyes only take damage in non powerhungry species when nightvision is turned on
/:cl:

### DIFF
--- a/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/organs.dm
+++ b/yogstation/code/modules/mob/living/carbon/human/species_types/preternis/organs.dm
@@ -21,6 +21,16 @@
 	// This variable is the actual night vision strength
 	var/light_cutoff = LIGHTING_CUTOFF_HIGH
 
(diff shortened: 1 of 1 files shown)


Yarinoi:
ethereal buff..

JamieD1 (approved):

PR #21525, opened 2024-02-26
Fixes a bug with venus human trap bulbs

Moltijoe:
:cl:  
bugfix: venus human trap bulb vines properly get destroyed when the bulb is
/:cl:


MajManatee (approved):
Weird that those are called vines too, but good to see gone none the less

PR #21526, opened 2024-02-26
Followup tweak to venus human trap weedkiller damage + pull

Moltijoe:
When i tweaked how weedkiller worked on venus human traps, i only really tested to make sure it worked, not if it was properly effective
It's proven to be too weak now that they regenerate health by default

Also removes pulling, as it's too easy for them to hide bodies

They're a midround nuisance that can be spawned by traitors, they aren't supposed to be as dangerous as the blob

:cl:  
tweak: Weedkiller does 2x damage to venus human traps than before
tweak: Venus human traps can no longer pull
tweak: venus human traps now require a pressurized environment
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
+++ b/code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
@@ -131,6 +131,7 @@
 /mob/living/simple_animal/hostile/venus_human_trap/Initialize(mapload)
 	. = ..()
 	AddElement(/datum/element/life_draining, damage_overtime = 5, check_damage_callback = CALLBACK(src, PROC_REF(kudzu_need)))
+	remove_verb(src, /mob/living/verb/pulled) //no dragging the poor sap into the depths of the vines never to be seen again
 
 /mob/living/simple_animal/hostile/venus_human_trap/Life(seconds_per_tick = SSMOBS_DT, times_fired)
 	. = ..()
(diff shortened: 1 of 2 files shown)


MajManatee:
Good idea. They shouldnt be hiding bodies and round removing people.

CrazyDrakkon:
Well, there goes most of the tact of traps, one of few simplemobs that's actually fun to play and stands a chance against crew. Worse than not being able to pull bodies is not being able to pull dropped weapons/water/fuel tanks.

RG4ORDR:
They're glass cannons, slow, and often are needing someone to get into the vines before even attacking.
Wanna make it better? 

Make them not-pressure immune. That's the only reason Goat ERTs never work because you can break a window, space maint and entire sections of the station under vines and the goats will just die due to pressure.

MajManatee:
Very true, last time I checked, vines are a plant, and plants need air

LassiH:
you can already wipe a area of vines with 10 units of weedkiller with a smoke machine why buff weedkiller more
them not being immune to spaced  areas means a engie can just come in and rcd the floor or windows to kill them with no counter

MajManatee:
Have you considered: The vines arent the main character, and arent supposed to win

Therandomhoboo:
Hard agree on removing their ability to pull. I don't so much agree on the others really. Space vine are annoying if left untreated and if i recall another pr was made because vines were getting trapped due to certain airlocks which is why they are super aggro'd up now.

I don't think it is wise to make the venus' have to stay in pressurised environments by the simple fact that, you can just space the area and watch them either die/leave because they cant stay there anymore which then just makes the game into duck hunt. So if we gotta make venus' die quicker I'd rather much have the 2x weedkiller damage cause at least then,  they can still survive longer by kiting.

As for the others saying they need air. 

Its a spess game, and they are future space vines shit we can make it lore accurate for shits sake.

Therandomhoboo:
sadge cause them being pressurised just fu cking cucked them due to gene's

PR #21527, opened 2024-02-26
fixes a bug where emissive bodyparts showed as white in the character menu

Moltijoe:
# Testing
image
image

:cl:  
bugfix: fixes a bug where emissive bodyparts showed as white in the character menu
/:cl:


MajManatee (approved):
I'm surprised it uses dummies

PR #21528, opened 2024-02-26
Fixes bugs with blood type compatibility

Moltijoe:
Subtypes of universal blood weren't included in universal compatibility
AB+ only could accept the blood types that AB- could, not any of the + blood types

:cl:  
bugfix: Fixes bugs with blood type compatibility
/:cl:


MajManatee (approved):
looks right

PR #21529, opened 2024-02-26
Fixes being unable to place items in closets or crates

Moltijoe:
caused by
#21500

# Testing
Gotta

:cl:  
bugfix: Fixes being unable to place items in closets or crates
/:cl:


MajManatee (approved):
I believe it

PR #21540, opened 2024-02-26
Fire extinguisher sprays no longer phase through solid windows and doors

Moltijoe:
# Testing
i haven't yet

:cl:  
bugfix: Fire extinguisher sprays no longer phase through solid windows and doors
/:cl:


SapphicOverload (commented):
This would prevent putting out fires on top of non-atmos-blocking structures like lockers and crates.

SapphicOverload (commented):
This still doesn't account for any structures that aren't closets and don't block atmos and wouldn't stop from moving through doors, it would probably be better to have windows and doors not care about PASSMACHINES or PASSSTRUCTURE since they're basically just fancy walls and already have their own pass flags.

Moltijoe:
more work than it's worth to me

PR #18622, opened 2023-04-22
Nanotrasen Inspector booth

Kaithyl:
https://user-images.githubusercontent.com/90080475/233808252-74f48795-5c4b-4026-82ae-9ebdf28c8cfb.mp4

*Special thanks to Lucas Pope for creating Papers, Please and for the audio*

Tested in singleplayer but not completely in multiplayer.
There are also several unused assets but I needed to take a break from this project for my own sanity.

# Spriting
window
button_prototype
desk_bg
desk_bottom
desk_bottom_sepia
desk_top
desk_top_sepia
idcard
idcard_gold
idcard_silver
paper
speaker
stamp_approve
stamp_cap
stamp_ce
stamp_clown
stamp_cmo
stamp_deny
stamp_hop
stamp_hos
stamp_law
stamp_qm
stamp_rd
stamp_unknown
tray_cover
tray_end
tray_segment
inspector_booth
booth_maint
booth_off

*All designs were pre-approved by the lore director*

# Wiki Documentation

This should be added to the list of machines here: https://wiki.yogstation.net/wiki/Machines
The player can insert paperwork into the booth from any direction but the UI and maintenance hatch can only be accessed from the back of the machine. (Or if you're standing on the same tile somehow)
Part upgrades: 
stamp tray capacity is determined by 2 + micro manipulator tier (3 when first constructed)
booth item capacity is determined by 3 + 2 * matter bin tier (5 when first constructed)

# Changelog

:cl:  
rscadd: Added Inspector Booth machine
tweak: Added Inspector Booth to list of circuit boards and corresponding research (Basic Research Technology)
soundadd: Added several sfx for UI interaction (sound\machines\inspector_booth)
imageadd: Added several UI icons and sprites (icons\UI_Icons\tgui\inspector_booth and icons\obj\machines\inspector_booth.dmi)
/:cl:

PR #21567, opened 2024-02-29
[BOX] [META] [DONUT] [ASTEROID] [GAX] Replaces the desk in the HOP line with an inspector's booth.

neeshacark:

Title, this does mean a compromise on Donut and Gax wherein the lose their desk bells due to there not being a good spot to move them to - all other maps maintain desk bells.

This item was added some time ago in #18622.  It is very cool and is never used.  HOP line is a good and logical spot for it to be

# Changelog

:cl:  cark
mapping: replaces all HOP line desks with an inspector booth
/:cl:

### DIFF
--- a/_maps/map_files/AsteroidStation/AsteroidStation.dmm
+++ b/_maps/map_files/AsteroidStation/AsteroidStation.dmm
@@ -990,9 +990,9 @@
 	dir = 2;
 	dwidth = 3;
 	height = 5;
-	shuttle_id = "mining_home";
 	name = "mining shuttle bay";
 	roundstart_template = /datum/map_template/shuttle/mining/box;
+	shuttle_id = "mining_home";
 	width = 7
 	},
 /turf/open/space/basic,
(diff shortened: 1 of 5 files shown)


Therandomhoboo:
I dont mind it being a potential variant of the normal office but to completely replace it with funny papers please mechanic is silly.

Please make this a potential type or give HOP the tools and paper-how to do it for them instead of just forcing them to be papers please.

AMyriad:
having more of these would be cool, but I don't like that people can't place anything that isn't paper or an ID card on the booth (so no dishing bribes), and the copyright free sounds it has are awfully lackluster
maybe if it was a table subtype or something, that might improve it idunno

neeshacark:
> having more of these would be cool, but I don't like that people can't place anything that isn't paper or an ID card on the booth (so no dishing bribes), and the copyright free sounds it has are awfully lackluster maybe if it was a table subtype or something, that might improve it idunno

this can be easily fixed if it is the case

AMyriad:
I'm not a big fan of removing the desk bells here. Leaving them on top of the inspector booths would be feasible if it had better sprites for any direction other than south, and if they were table subtypes in the same vain of what I said before.
Alternatively you could also add more of those little nooks for tables at the end of the line like Box and Meta have to the areas that don't have them.

But given the current state of inspector booths, I don't think they're worth replacing tables over. Also to note, I'm not a terrible fan of how many shuttle ids and other minor things got touched accidentally, I'd recommend redoing those and the changes on Meta if ye wanna make sure nothing else got donked up.

Aquizit:
I kinda want to see this testmerged first, to see how it plays, and then make a decision from there.

PR #21573, opened 2024-02-29
Adds canister bombs

void5402:

makes canisters explode when they break if they have 200000Kpa or higher pressure and makes single tank bombs attachable to them so the canister can detonate by itself.

gives more depth and possibilities to some jobs

# Testing
<details> 
  <summary>tested</summary>
  
[2024-02-29 19-38-18.webm](https://github.com/yogstation13/Yogstation/assets/74709370/db34cc29-3f3c-497e-a4ce-a4e3d5492bf5)

  a tiny bit laggy
</details>

# Wiki Documentation
canisters exploding with high pressure and single tank attachable to them

# Changelog

:cl:  
rscadd: canister bombs now explode when broken if they have enough pressure
rscadd: makes single tank bombs attachable to canisters

/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/portable/canister.dm
+++ b/code/modules/atmospherics/machinery/portable/canister.dm
@@ -1,4 +1,6 @@
 #define CAN_DEFAULT_RELEASE_PRESSURE (ONE_ATMOSPHERE)
+#define CANISTER_EXPLOSION_PRESSURE   (40.*ONE_ATMOSPHERE) //maybe needs tweaking
+#define CANISTER_EXPLOSION_SCALE      (6.*ONE_ATMOSPHERE)
 
 /obj/machinery/portable_atmospherics/canister
 	name = "canister"
@@ -376,15 +378,19 @@
 
 /obj/machinery/portable_atmospherics/canister/proc/canister_break()
 	disconnect()
-	var/datum/gas_mixture/expelled_gas = air_contents.remove(air_contents.total_moles())
+	var/pressure = air_contents.return_pressure()
 	var/turf/T = get_turf(src)
-	T.assume_air(expelled_gas)
-
+	if(pressure <= CANISTER_EXPLOSION_PRESSURE)
+		var/datum/gas_mixture/expelled_gas = air_contents.remove(air_contents.total_moles())
+		T.assume_air(expelled_gas)
+		playsound(src.loc, 'sound/effects/spray.ogg', 10, TRUE, -3)
+	else //mini nukes
+		log_bomber(get_mob_by_key(fingerprintslast), "was last key to touch", src, "which ruptured explosively")
+		var/range = (pressure-CANISTER_EXPLOSION_PRESSURE)/CANISTER_EXPLOSION_SCALE
+		explosion(T, round(range*0.25), round(range*0.5), round(range), round(range*1.5))
 	obj_break()
 	density = FALSE
-	playsound(src.loc, 'sound/effects/spray.ogg', 10, TRUE, -3)
 	investigate_log("was destroyed.", INVESTIGATE_ATMOS)
-
 	if(holding)
 		usr.put_in_hands(holding)
 		holding = null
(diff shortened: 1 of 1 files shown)


code/modules/atmospherics/machinery/portable/canister.dm
```diff
@@ -33,6 +35,8 @@
 	var/timing = FALSE
 	var/restricted = FALSE
 	req_access = list()
+	//a singletank bomb te explode the canister
+	var/obj/item/tank/single_tank
```
iamthesilly: Canisters already have a variable for this, `holding`.

code/modules/atmospherics/machinery/portable/canister.dm
```diff
@@ -368,6 +374,29 @@
 
 	return TRUE
 
+//attaching or droping the singletank bomb
+/obj/machinery/portable_atmospherics/canister/attackby(obj/item/W, mob/user, params)
```
iamthesilly: Canisters already hold tanks, this entire part is unnecessary.

SapphicOverload (changes requested):

code/modules/atmospherics/machinery/portable/canister.dm
void5402: its for holding the bomb and not transferring gas between them and showing it to people

iloveloopers:
oops nevermind didn't realize it was only when broken

code/modules/atmospherics/machinery/portable/canister.dm
iamthesilly: It already holds the tank whether or not you open the canister to transfer gas.

void5402: will this do?

code/modules/atmospherics/machinery/portable/canister.dm
```diff
@@ -310,12 +312,16 @@
 		. += "can-open"
 	if(holding)
 		. += "can-tank"
+		if(holding.bomb_status)
```
iamthesilly: Why is this here twice?

code/modules/atmospherics/machinery/portable/canister.dm
```diff
@@ -310,12 +312,16 @@
...
+			. += holding
 	if(connected_port)
 		. += "can-connector"
 
 	var/light_state = get_pressure_state(air_contents?.return_pressure())
 	if(light_state) //happens when pressure is below 10kpa which means no light
 		. += mutable_appearance(icon, light_state)
+	if(holding.bomb_status)
```
iamthesilly: This needs a null check so it doesn't cause runtime errors.
```suggestion
	if(holding?.bomb_status)
```

SapphicOverload (changes requested):

SapphicOverload (approved):

MajManatee:
pov you are trying to do fusion in a can and accidentally depopulate the station

Moltijoe:
Balance Council:
>What? this just lets atmos and toxins create more maxcaps

iloveloopers:
you can make and store singletank bombs in a bag which are capable of gibbing, but a huge canister that everyone can see is what crosses the line

neeshacark:
i mean hot take maybe but atmos can do much worse than a maxcap - floods of any kind are WAY more damaging.

and toxins can already make maxcaps so idk why letting them do it in a slightly different way is bad

Moltijoe:
> i mean hot take maybe but atmos can do much worse than a maxcap - floods of any kind are WAY more damaging.
> 
> and toxins can already make maxcaps so idk why letting them do it in a slightly different way is bad

the thing is, scientists are already antifun with maxcap bombing, but they have limited ttv, so eventually the bombs run out
with canister bombs, there's far more canisters, so they're not likely to ever run out
also, it only relying on pressure instead of a proper gas mixture doesn't exactly help it's case either

When it comes to floods, they're annoying, but it's damage is SO widespread that only hijack/martyr engineers can do it without getting bwoinked
Bombs are significantly more controlled, meaning even civilians can get away with using them under certain circumstances

Hislyna:
You should change it so that if its a too high pressure threshold, theres a chance that itll just blow up in their face

PR #21610, opened 2024-03-08
Tweaks mood

Moltijoe:
Removes slowdown for all but the worst moods, and reduces the strength of the slowdown that remains
Reduces universal base movespeed ever so slightly -0.1
The reduction is counteracted by being even _slightly_ happy 0.0
if really happy, you'll move _slightly_ faster 0.05

Also completely removes the debuff to crit threshold when at low sanity

Mood needs a lot more work, but this is a start

# Testing
i can't really show it, it's minor speed differences

:cl:  
tweak: Tweaks mood to have significantly less negatives and REALLY minor positives
/:cl:

### DIFF
--- a/code/datums/components/mood.dm
+++ b/code/datums/components/mood.dm
@@ -1,6 +1,3 @@
-#define MINOR_INSANITY_PEN 5
-#define MAJOR_INSANITY_PEN 10
-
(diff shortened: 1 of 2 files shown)


wonderinghost:
fucking finally

PR #21625, opened 2024-03-11
Skin twister ided

AMyriad:
Skin twister health goes from 320 -> 285

This mob is incredibly strong already, has an incredible 30-55 damage every hit, is remarkably fast, and to top it off, has a lot more health than the players themselves.
Combined with the fact they're designed to hit miners while in crit, and disguise as them after death can be a bit punishing.

So that is my case.
also I just witnessed 7 people get crushed by this single mob what in the goddamn

# Testing
Will do

# Changelog
:cl:  
tweak: Skin twister health reduced from 320 -> 285
/:cl: 

P.S. Ye I'm aware most of everything in jungleland is unfinished or broken, fixes are due to come as long as the server doesn't die

### DIFF
--- a/yogstation/code/modules/jungleland/jungle_mobs.dm
+++ b/yogstation/code/modules/jungleland/jungle_mobs.dm
@@ -178,8 +178,8 @@
 	response_help  = "pets"
 	response_disarm = "gently pushes aside"
 	response_harm   = "hits"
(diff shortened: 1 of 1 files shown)


MajManatee (approved):
skill issue i guess

PR #21655, opened 2024-03-15
Brings infravision back and gives it to polysmorphs

Moltijoe:
Brings back infravision which was deleted with chubby's pr
also, poly lost their nightvision with that same pr, so this gives them something else that functions arguably better, but only for certain scenarios

So, when bringing it back, i wasn't able to find the part where it actually cared about infra_luminosity for distance, so it's basically just night vision, but different

1. Gives polysmorphs an interesting gameplay feature
2. "bring back" a feature polysmorphs lost with a previous pr
3. fixes a bug with infravision

# Testing
image

:cl:  
rscadd: Polysmorph eyes now provide infrared vision
bugfix: Fixes infrared vision not working  
/:cl:


AMyriad:
do these white silhouettes use a sprite or something or are they overlayed with code?
it'd be funny if they were red/yellowish so we can get everything looking like this
image

ToasterBiome (approved):

PR #21665, opened 2024-03-16
Redesigns the tar king's arena

AMyriad:
I wanted to replace the sand with obsidian because I thought the sand looked ugly as hell.
Then I realized the whole arena was pretty ugly, so I just redid it all.
StrongDMM-2024-03-16 01 11 46

I took inspiration from the original size and shape, but I wanted to separate the dangers of the tar cultists from the serene room in the center where you'll fight the king himself.

The environment sports less unimportant garbage miners will ignore, and has been filled with tar, which is hard to move through. And because I thought it was a bit dark, I planted some nice flowers to keep you company. Also features some nice writing for people who like to stop and smell the petals. 

I find the new design to be more ominous, and sells the point of who you're actually fighting better than the old one.
**I'm also in the process of rewriting a lot of stuff for jungleland, and this is my first step to realizing the story I want to tell for this place.**

# Testing
image

# Wiki Documentation
Ye we should put more jungleland stuff on there. I'll get around to it eventually if someone doesn't beat me to the punch.

# Changelog
:cl:  
mapping: The Tar King's arena has been completely redesigned! Go check it out
/:cl:


neeshacark (approved):
seems to all be good here

feature freeze

Moltijoe:
> seems to all be good here
> 
> feature freeze

only valid for prs made after the freeze started

PR #21693, opened 2024-03-18
Adds a pop requirement to sleeping carp

Moltijoe:
feature freeze permission slip
image

this martial art has always been an issue and never fun to play against
it's magnified in lowpop where often there's only like 1 sec at best
giving it a pop requirement should ensure an antag doesn't just "win" from buying a single item

# Testing
number change innit

:cl:  
tweak: adds a 20 pop req to sleeping carp (doesn't require readied up like gamemodes do)
/:cl:


MajManatee (approved):
golly good bruv

AMyriad:
pop locks on all martial arts would be good

PR #21709, opened 2024-03-21
[DONUT] Heavy Petting

neeshacark:

adds some missing pets

becquareldrqweqlq, warden jacob, araneus, officer kreb, kitchen animal

# Changelog

:cl:  cark
mapping: adds missing pets to donut
/:cl:


AMyriad:
looks good
could ye also get Anadear in sci and double check if Bap is where she's supposed to be? can't get a good look this second but she might be in botany instead of the janitor's closet

AMyriad (changes requested):
looks solid, could you put an Anadear in science while you're at it?

AMyriad (approved):

PR #21769, opened 2024-03-30
Removes the slaughter demon spawner from battle royale

Moltijoe:
slaughter demons drop hearts that can grant bloodcrawl, which give a win by default to anyone that eats it
if two people get bloodcrawl, they can force a round to never end
laughter demon doesn't have that problem as it doesn't drop the heart on death

# Testing
no need, it's just deleting a line from a list, also it's damn near impossible to test anyways

:cl:  
rscdel: Removes the slaughter demon spawner from battle royale
/:cl:


ofscarletdreams:
we booed the fucking code out

Yarinoi:
BOOO BOOO THE CODE BOOOOOO

alternatively just make it so that it kicks you out of bloodcrawl after a bit on battleroyale?

ToasterBiome (approved):

PR #21806, opened 2024-04-05
makes it so perma always starts with the lights on

Yarinoi:
makes it so perma always starts with the lights on

if the lights start off (which they tend to, since i cant recall a single time the perma lights werent off when i've gotten perma'd since the new perma was added) they are likely to STAY off, which sucks for the inhabitants as they have to wander around basically blind, and gives sec a massive advantage as they can use NVGs which allows sec to see prisoners while prisoners cant see sec.
i like bein able to fuckin see in my perma

# Testing
image
im the only person, started as chap, perma lights were on

# Changelog

:cl:  
mapping: Protests from the Prisoner's Rights Society has forced NT's hand into making perma's lights always start on.
/:cl:

### DIFF
--- a/code/game/area/Space_Station_13_areas.dm
+++ b/code/game/area/Space_Station_13_areas.dm
@@ -916,9 +916,11 @@ NOTE: there are two lists of areas in the end of this file: centcom and station 
 	name = "Prison Wing"
 	icon_state = "sec_prison"
 	minimap_color = "#530505"
+	lights_always_start_on = TRUE
 
 /area/security/prison/hallway
 	name = "Prison Wing Hallway"
+	lights_always_start_on = TRUE
 
 /area/security/processing
 	name = "Labor Shuttle Dock"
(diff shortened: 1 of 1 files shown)


code/game/area/Space_Station_13_areas.dm
```diff
@@ -916,9 +916,11 @@ NOTE: there are two lists of areas in the end of this file: centcom and station
 	name = "Prison Wing"
 	icon_state = "sec_prison"
 	minimap_color = "#530505"
+	lights_always_start_on = TRUE
 
 /area/security/prison/hallway
 	name = "Prison Wing Hallway"
+	lights_always_start_on = TRUE
```
Moltijoe: Dont need this line, it inherits TRUE from the parent

Moltijoe (changes requested):

code/game/area/Space_Station_13_areas.dm
Yarinoi: ohh

Moltijoe (approved):

Yarinoi:
added proof of testing

Issue #21772, opened 2024-03-31
You can suplex people while in mechs


Oblisk234:
### Severity

Inconvenience

### Short Description

Using the seismic arm, you can suplex people while you yourself are in a mech

### Reproduction

Get a seismic arm, try to suplex someone while you're in a mech, they actually get suplexed?

### Round ID

52052

### Test Merges

#21623, #20581

### BYOND Username

Oblisk234

### Was this done on a locally hosted or non-Yogstation hosted server?

No

PR #21811, opened 2024-04-06
makes it so you can no longer use the seismic arm from within objects

LazennG:

adds a check to using the seismic arm that makes it so you can't use it while inside things like mechs and bodybags for the sake of issue #21772 

# Testing

dreamseeker_Ff3sPBgZ7h

# Changelog

:cl:  

bugfix: fixed being able to seismic arm people from inside things
/:cl:

### DIFF
--- a/code/datums/martial/reverbpalm.dm
+++ b/code/datums/martial/reverbpalm.dm
@@ -66,6 +66,10 @@
 
 /datum/martial_art/reverberating_palm/can_use(mob/living/carbon/human/H)
 	var/obj/item/bodypart/r_arm/robot/seismic/R = H.get_bodypart(BODY_ZONE_R_ARM)
+	var/turf/current = get_turf(H)
+	for(var/obj/outside in current.contents)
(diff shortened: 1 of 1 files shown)


code/datums/martial/reverbpalm.dm
```diff
@@ -66,6 +66,10 @@
 
 /datum/martial_art/reverberating_palm/can_use(mob/living/carbon/human/H)
 	var/obj/item/bodypart/r_arm/robot/seismic/R = H.get_bodypart(BODY_ZONE_R_ARM)
+	var/turf/current = get_turf(H)
+	for(var/obj/outside in current.contents)
```
iamthesilly: Checking `!isturf(H.loc)` is much faster than searching through contents of turf contents for the player

LazennG: oops didnt know loc was chill like that thanks

ToasterBiome (approved):

PR #21814, opened 2024-04-06
Fixes antinoblium not showing up properly in the HFR interface

SapphicOverload:
image

:cl:  
bugfix: Fixed antinoblium not showing up properly in the HFR interface
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/Hypertorus/Gases.js
+++ b/tgui/packages/tgui/interfaces/Hypertorus/Gases.js
@@ -26,7 +26,7 @@ const moderator_gases_help = {
     'Saps most forms of energy expression. Slows the rate of temperature change.',
 };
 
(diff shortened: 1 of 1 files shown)


ShoziOffical:
You would have to do this for other gases too, because they only show when you put them in moderator

iamthesilly:
> You would have to do this for other gases too, because they only show when you put them in moderator

This just corrects a spelling mistake causing antinoblium to not have the correct label and color, it was already there

ShoziOffical:
> > You would have to do this for other gases too, because they only show when you put them in moderator
> 
> This just corrects a spelling mistake causing antinoblium to not have the correct label and color, it was already there

Oh alr

ToasterBiome (approved):

PR #21825, opened 2024-04-07
Fixes the server not restarting at the end of a round

Moltijoe:
Current culprit is this proc call
it calls it just before the return statement
it asks 3 people to respond to a tgui prompt, but without a timeout it can just never end
meaning setup() never finishes returning, and that fire() call never finishes

it was also added by the ai rework pr
the last time we encountered this problem was last time the pr was testmerged

# Testing
lol, lmao, good fuckin luck confirming or denying it

:cl:  
bugfix: fixes the server not restarting at the end of a round
/:cl:

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -353,7 +353,7 @@ SUBSYSTEM_DEF(ticker)
 		place.power_change()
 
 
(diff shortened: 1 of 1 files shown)


ShoziOffical:
What does this proc even do anyway

Moltijoe:
> What does this proc even do anyway

ask bibby, i tried removing it and they insisted it was important

ToasterBiome (approved):

PR #21865, opened 2024-04-11
Fixes SSDemo runtime on startup caused by trying to qdelete an object before SSDemo is initialized 

Djiq:
Fixes SSDemo runtime caused by SSdemo not being initialized when qdel is called


Djiq:
wrong solution i will try to come up with something that actually wors

PR #21869, opened 2024-04-12
Ports accessory pinning component

JohnFulpWillard:

Ports accessory pinning component from https://github.com/tgstation/tgstation/pull/80230
Fixes sheet defines to fit Yog's material balance

The component will let you make more things attachable to jumpsuits than just medals if anyone wants to make use of it.

I ported this while waiting for my compile + 300 second init time to look at medals reportedly not working, only to find out I already fixed them, so that was fun.

Fix to material, and attaching things that aren't medals might be cool i dunno. Components are pretty cool.

# Testing

I spawned in a second body and hit them with the medal to commend them, I then took it off and tried to re-attach it again to make sure the old behavior of not double-commending wasn't lost.

# Changelog

This has nothing player-facing yet.

### DIFF
--- a/code/__DEFINES/construction/material.dm
+++ b/code/__DEFINES/construction/material.dm
@@ -1,11 +1,11 @@
 //Defines for amount of material retrived from sheets & other items
-/// The amount of materials you get from a sheet of mineral like iron/diamond/glass etc. 100 Units.
-#define SHEET_MATERIAL_AMOUNT 100
-/// The amount of materials you get from half a sheet. Used in standard object quantities. 50 units.
+/// The amount of materials you get from a sheet of mineral like iron/diamond/glass etc. 2000 Units.
+#define SHEET_MATERIAL_AMOUNT 2000
+/// The amount of materials you get from half a sheet. Used in standard object quantities. 1000 units.
 #define HALF_SHEET_MATERIAL_AMOUNT (SHEET_MATERIAL_AMOUNT / 2)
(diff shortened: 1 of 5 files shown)


code/__DEFINES/construction/material.dm
```diff
@@ -1,11 +1,11 @@
 //Defines for amount of material retrived from sheets & other items
-/// The amount of materials you get from a sheet of mineral like iron/diamond/glass etc. 100 Units.
-#define SHEET_MATERIAL_AMOUNT 100
-/// The amount of materials you get from half a sheet. Used in standard object quantities. 50 units.
+/// The amount of materials you get from a sheet of mineral like iron/diamond/glass etc. 2000 Units.
+#define SHEET_MATERIAL_AMOUNT 2000
```
ToasterBiome, line 3: why is this changed?

ToasterBiome (changes requested):

code/__DEFINES/construction/material.dm
JohnFulpWillard: 2000 is yog's material sheet amount because it's TG's old one and is what is currently used as it. TG recently changed it but Yog hasn't so using this define means things that are using these defines are giving 1900 less than they're supposed to.

JohnFulpWillard: The only thing using these defines are things added in plane cube anyway.

ToasterBiome (approved):

PR #21880, opened 2024-04-15
Replaces all mentions of Nanotrasen with Yogscast to prove that its not that hard.

azzzertyy:
I will close this in like 20 minutes. Ignore it.


YashinHak:
outstanding

PR #21884, opened 2024-04-16
change cloaker belt name to be stealthy

ktlwjec0:
changes cloaker belt name to look like "toolbelt"

This fits more in-line with our rules, "_Items from the stealth and implants (once implanted, not the implanter) uplink categories are metashielded. Security are not allowed to “check” these unless they have witnessed the use of the suspicious item before._" How do you expect security to see someone wearing a cloaker belt and go "yeah, that's cool you can keep that". So instead, they will see "toolbelt", much like how the camera flash and hypno flash are named.

# Testing
this is stupid
image
image

:cl:  ktlwjec
tweak: Cloaker belt is more stealthy by reading as a normal toolbelt when worn.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/traitordevices.dm
+++ b/code/game/objects/items/devices/traitordevices.dm
@@ -199,7 +199,7 @@ effective or pretty fucking useless.
 				wavelength = clamp(target, 0, 120)
 
 /obj/item/storage/belt/military/shadowcloak
-	name = "cloaker belt"
+	name = "toolbelt"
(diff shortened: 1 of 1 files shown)


MommaDragon:
Good. Some mime got "caught" wearing it and Sec just didn't play the metashield because of the suspicious name.

Firewolf34:
Good stuff... Do everything we can to avoid making people have to respect metashield unless absolutely needed. Honor rules are hard to enforce and just create lots of admin tickets. Better to have the functionality as part of the mechanics, wherever possible

Moltijoe (approved):

PR #21892, opened 2024-04-17
Fix open air fusion reaction not heating up after sometimes

warface1234455:
this makes fusion reaction weaker and doesnt generate a lot of heat according to my experience in game

Fix open air fusion reaction not heating up after someti

should help fusion reaction generate heat faster

# Testing
still testing, its kinda better but my end is slow kinda slow so i would like a testmerge for a proper test

# Changelog

:cl:  

experimental: Fix open air fusion reaction not heating up after someti
/:cl:

### DIFF
--- a/code/modules/atmospherics/gasmixtures/reactions.dm
+++ b/code/modules/atmospherics/gasmixtures/reactions.dm
@@ -373,7 +373,7 @@
 
 		var/new_heat_capacity = air.heat_capacity()
 		if(new_heat_capacity > MINIMUM_HEAT_CAPACITY)
-			air.set_temperature(clamp(((old_thermal_energy + reaction_energy)/new_heat_capacity),TCMB,INFINITY))
+			air.set_temperature(clamp(((old_thermal_energy + reaction_energy)/(new_heat_capacity*0.8)),TCMB,INFINITY))
 		return REACTING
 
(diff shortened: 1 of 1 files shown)


code/modules/atmospherics/gasmixtures/reactions.dm
```diff
@@ -373,7 +373,7 @@
 
 		var/new_heat_capacity = air.heat_capacity()
 		if(new_heat_capacity > MINIMUM_HEAT_CAPACITY)
-			air.set_temperature(clamp(((old_thermal_energy + reaction_energy)/new_heat_capacity),TCMB,INFINITY))
+			air.set_temperature(clamp(((air.return_temperature()*old_heat_capacity + reaction_energy)/(new_heat_capacity*0.5)),TCMB,INFINITY))
```
iamthesilly: Why is there a `*0.5` here?

ShoziOffical: because of auxmos, that 0.5 makes it works better

ShoziOffical: well on my end of course, maybe it would be different on live

SapphicOverload (approved):

PR #21905, opened 2024-04-19
Adjust player count requirement for nuclear operatives

SapphicOverload:

Nuclear operatives player count requirement has been reduced to 20, with the number of operatives and telecrystals received from declaring war scaling with player count. Current scaling is 1 operative per 10 players rounded down, and 7 TC per player when declaring war.

Currently with our significantly lower player count there's a lack of any interesting gamemodes that can happen, which further harms player count. Allowing more gamemodes to happen and making them scale better when player count is low would help fix this to some extent.

# Testing
Good luck.

# Changelog

:cl:  
tweak: Nuclear operatives minimum player count reduced to 20
tweak: Telecrystals gained when declaring war now scales with player count
/:cl:

### DIFF
--- a/code/game/gamemodes/nuclear/nuclear.dm
+++ b/code/game/gamemodes/nuclear/nuclear.dm
@@ -5,8 +5,8 @@
 	config_tag = "nuclear"
 	report_type = "nuclear"
 	false_report_weight = 10
-	required_players = 10 // scales with player count, 1 operative for every 10 players, rounded down
-	required_enemies = 1 // lone operative at minimum
+	required_players = 20 // scales with player count, 1 operative for every 10 players, rounded down
+	required_enemies = 2 // needs more than one person to make a team
 	recommended_enemies = 5
 	antag_flag = ROLE_OPERATIVE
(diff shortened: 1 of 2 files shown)


cowbot92:
Would probably have to restrict some items as well, and force war since stealth ops is too op. Would probably suggest min pop 20 or 15 moreso than 10 but that part is just my opinion, fair PR otherwise.

Addust:
"force war"

my brother in ratvar they do not announce loneops
also thatll give them like
300 tc

cowbot92:
> "force war"
> 
> my brother in ratvar they do not announce loneops also thatll give them like 300 tc

Yes  know, lone ops are a silent midround spawn that only occurs after a disk has been left alone, whereas this is a roundstart gamemode. You can adjust the amount of TC they receive as needed.

LassiH:
>  also thatll give them like 300 tc
pr also makes the amount of tc you get from war based on the current pop

iamthesilly:
Increased to 20 minimum players

ynot01 (approved):

PR #21915, opened 2024-04-20
Mimes can use the darkspawn mindlink without needing to break their oath

Moltijoe:
# Testing
image

:cl:  
bugfix: Mimes can use the darkspawn mindlink without needing to break their oath
/:cl:


Chubbygummibear (approved):
it's probably fine

PR #21919, opened 2024-04-21
lore terminals but in charlie station! also slightly organises code

Addust:

adds lore terminals to charlie station
slightly compresses the amounts of lines used in the spawner code

reduces space code takes up by like 5 bytes
adds lore terminals

# Testing
image
the weird formatting is because terminals are terminals (i.e shit), theyre a ported feature AND use the wizard magic that is tgui so i really can't do anything to fix it

# Spriting

# Wiki Documentation

# Changelog

:cl:  
rscadd: lore terminals have been added to the ancient space station (which you plebians call charlie station even though that's just the module with cryogenic stasis)
rscadd: the code also now takes up a bit less space because somethingsomething wizard magic on the spawners that you cant see but coders can see
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/oldstation.dmm
+++ b/_maps/RandomRuins/SpaceRuins/oldstation.dmm
@@ -1052,6 +1052,14 @@
 /obj/machinery/door/airlock/medical/glass,
 /turf/open/floor/plasteel/white,
 /area/ruin/space/has_grav/ancientstation/rnd)
+"dt" = (
+/obj/effect/decal/cleanable/dirt,
+/obj/machinery/computer{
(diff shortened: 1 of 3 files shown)


AMyriad:
I don't think the second part of that changelog is needed

AMyriad (changes requested):
Looks aight map-wise but if you're gonna keep regular broken computers at all, I don't see any reason to delete the two in the front. Please readd them.

image

AMyriad (approved):
Aight

Also I'm not a code maint but I'd _strongly_ suggest using newlines so those terminals are more readable. You just put `\n\` at the end of every line and `\n` on the last one, see traitor_factions.dm for a good example.

Moltijoe (approved):

Moltijoe:
i'll fix it

PR #21928, opened 2024-04-23
Allows ghosts to move between z levels at will using the verb

Moltijoe:
it didn't let them pass through solid floors

# Testing
image
image
image
image

:cl:  
bugfix: Allows ghosts to move between z levels at will using the verb
/:cl:


Chubbygummibear (approved):
Molti-z

PR #21951, opened 2024-04-26
Fixes the particool system from leaving permanent smoke visual effects

Moltijoe:
it was creating a smoke visual effect for the master holder that's used by the particle emitter system

# Testing
image
image
image

just to show that regular particles (water) still work
image

:cl:  
bugfix: Fixes the particool system from leaving permanent smoke visual effects
/:cl:


Chubbygummibear (approved):
hot

PR #21960, opened 2024-04-28
Fixes the flashlight and flashdark being invisible when turned on

Moltijoe:
they never had an -on sprite it seems

# Testing
image
image

:cl:  
bugfix: Fixes the flashlight and flashdark being invisible when turned on
/:cl:


AMyriad:
That bottom inhand doesn't apply to flashdarks, right?
They'd be dark either way...maybe darker while on

ToasterBiome (approved):

PR #21961, opened 2024-04-28
Gives mining medic an unupgradeable version of the explorer suit

Moltijoe:
It has 15 more melee armour, however it can't be upgraded using goliath plates, resulting in lower armour at the end
in return, it can carry both mining and medical equipment in the suit slot, whatever is more convenient
Mining medic spawns with one and there are no others if it gets destroyed somehow

also makes new global suit allowed lists for:
labcoat: applied to just labcoats right now, but is really just medical and science stuff
mining: applied to explorer suit and heck suit for now

cleans up some of the security lists that include all subtypes of internals tanks rather than just getting the subtypes normally

Mining medics are expected to save miners
They shouldn't be taking the miners gear ideally
their current jacket isn't suitable for something like icemeta as it has no insulation
it also doesn't cover their legs, which is kinda important

giving them an explorer suit straight up doesn't seem right
but giving them a reskinned explorer suit that is tweaked to better suit mining medic is preferred

# Testing
image
image

# Spriting
@cuackles has agreed to modify the explorer suit sprite to have medical colours
he's even said i can keep reminding him until he does it

:cl:  Cuackles & Molti
tweak: Gives mining medic an unupgradeable version of the explorer suit
imageadd: Recovery suit sprites
/:cl:


AMyriad:
nice, I won't have a reason to steal the 4th suit every time anymore

Cartlord:
I thought the 4th suit was for mining medics...

Moltijoe:
> I thought the 4th suit was for mining medics...

considering how often a 4th miner seems to show up (or at least did back when we had highpop)
it isn't

Mqiib (approved):

ToasterBiome (approved):

boodaliboo:
why wasa this self merged i thought that wasnt allowed i get it was aproved but i remeber specifically maintainers or head coders whatever his role got in trouble for merging a pr they made unless its a bug fix

PR #21968, opened 2024-04-29
Makes random item rotation look less bad

SapphicOverload:

This makes randomized item rotation round to the nearest multiple of 90 degrees.

Rotating sprites to any degree that isn't a multiple of 90 looks bad unless the resolution is high enough, which is not the case in this game. Locking item rotation to multiples of 90 allows random rotation to still exist without making certain sprites unrecognizable or difficult to click on.

# Testing
image

:cl:  

tweak: Randomized item rotation rounds to 90 degrees
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -756,10 +756,10 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 		. = callback.Invoke()
 	item_flags &= ~IN_INVENTORY
 	var/matrix/M = matrix(transform)
-	M.Turn(rand(-170, 170))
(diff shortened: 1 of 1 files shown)


ToasterBiome:
@LazennG state thy reason

AMyriad:
can't put my finger on why but it looks a tad bizarre
I'd have to see how it does in-game to get an actual opinion

LazennG:
i'm not a fan because i don't like how neat it is. when item rotations were introduced i liked how it made things look strewn about and placed without any care and this feels too neat in comparison.

Moltijoe:
maybe try including 45 degrees too

ToasterBiome (approved):

PR #21991, opened 2024-05-02
Fixes radio messages redirecting through intercomms and radios instead of headsets

SapphicOverload:
That's basically it, using a headset makes you use the headset as it should be.
image

 Using .r or .i to talk into a radio or intercom respectively still works as well.
image

:cl:  
bugfix: Fixed radio messages redirecting through intercomms or radios when trying to use a headset.
/:cl:

### DIFF
--- a/code/__DEFINES/say.dm
+++ b/code/__DEFINES/say.dm
@@ -23,11 +23,8 @@
 
 #define MODE_ROBOT "robot"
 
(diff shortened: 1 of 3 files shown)


AMyriad:
RIP talking through ventriloquist dummies

ToasterBiome (approved):

PR #21995, opened 2024-05-02
Adds weapon icons to the game

AMyriad:
This is part 1 of the project where I organize, reorganize, and clean up all our icon files.

Atomized for the sake of our spritetainers' sanity (I'm sorry), this PR focuses on organization of melee weapons.
Know that I've double checked everything to make sure I didn't break anything, but I'll test later to triple check ofc.

Redundant/unused icons, "snowflake" icon files with a single icon, and messy icon files are not good.
Sorting these things out helps icon file readability for contributors and will over time improve game performance, since we'll be grouping more relevant things in the same files, lightening the load for when the game draws all the icons from each file.

# Testing
image

# Spriting
Before
image

After
image
( more things in the same file -> icons less spread out -> better )

# Changelog
:cl:
experimental: Reorganized our weapon icons. Ping me or make a bug report if you see any weapon missing their sprites!!
/:cl:


Chubbygummibear:
i'm rather opposed to deleting any yogstation dmi that has a mirrored version in the main folder even if it's only 1 file in a folder, because the point is that there's a place to add sprites that are unique to yog. like in the case of the yog lavaland artifacts. it should remain so that if we add more artifacts or @LazennG makes more yog unique lavaland loot, there's a place to put it that's reflective of the fact it's an asset unique to us. it's always a weird feeling when you have to be the first one to make the new dm or dmi in the yogstation/ directory because there's no food.dm or vending.dmi or whatever, but someone's gotta be the first, even if it only is for 1 sprite or 1 subtype.

AMyriad:
> i'm rather opposed to deleting any yogstation dmi that has a mirrored version in the main folder even if it's only 1 file in a folder, because the point is that there's a place to add sprites that are unique to yog. like in the case of the yog lavaland artifacts. it should remain so that if we add more artifacts or LazennG makes more yog unique lavaland loot, there's a place to put it that's reflective of the fact it's an asset unique to us. it's always a weird feeling when you have to be the first one to make the new dm or dmi in the yogstation/ directory because there's no food.dm or vending.dmi or whatever, but someone's gotta be the first, even if it only is for 1 sprite or 1 subtype.

Alright I understand that, but I don't understand why we have yog-unique stuff in a subfolder at all. There are plenty of things unique to yogs outside of that folder and vice versa, and right now it's filled with a lot of old, outdated, misplaced icons in strange misnamed places (like /xenoarch/guns.dmi for example) that haven't been maintained in ages.

I strongly believe it would be better to find our files in one place, as is - from my understanding - usually the purpose of organizing files. I don't see why separating things purely because they're unique to here is good practice when they could be in the "normal" files like everything else? What would that say about everything outside of the yogstation folder, is that not _ours_ either?

Mqiib:
image
image

What's the edit to the geiger counter, and why is the hammer renamed?

Mqiib (changes requested):
right i need to hit this button, see above

AMyriad:
> What's the edit to the geiger counter, and why is the hammer renamed?

I didn't change anything about that geiger icon but there's a chance I moved it in the file by accident. I checked and didn't see anything different, but just to be safe I copy pasted the icon from master so there shouldn't be any problemo now

For the hammer, I thought calling something "hammer" in hammer.dmi already would be a bit vague, and it's what TG calls it in their files so why not ¯\\_(ツ)__/¯

Mqiib (approved):

Mqiib:
icondiff bot is mad tho

PR #22010, opened 2024-05-05
Disables rock paper scissors until it actually has a use

Moltijoe:
Checking ip address for localhost addresses doesn't work for me, and i assume it doesn't work for others
i'm just taking a bludgeon to it
if there's less than 3 people on the server, don't ask

player sanity

# Testing
bro trust me

:cl:  
bugfix: Disables rock paper scissors until it actually has a use
/:cl:

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -353,7 +353,7 @@ SUBSYSTEM_DEF(ticker)
 		place.power_change()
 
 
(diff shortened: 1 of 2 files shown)


cowbot92:
probably should just comment it out like my original pr tbh, then bibby can recomment it when he's ready to do his funny thing

AMyriad:
would be easier to just comment it out, no?

ToasterBiome (approved):

PR #22024, opened 2024-05-08
Fixes AIs only seeing static after being carded or uploaded to a mech

SapphicOverload:
I can't test this by myself but I'm 99% sure this fixes it

:cl:  
bugfix: Fixes AIs only seeing static after being carded or uploaded to a mech
/:cl:


Moltijoe (approved):

Moltijoe:
Godspeed

PR #22040, opened 2024-05-16
Prevents admin only evac shuttles from getting discounts, making them appear on the console

Moltijoe:
an oversight with how the discount system and console work

# Testing
would need to, but it'd just be me rerolling shuttles again and again, nothing showing up every time

:cl:  
bugfix: Prevents admin only evac shuttles from getting discounts, making them appear on the console
/:cl:

### DIFF
--- a/code/datums/shuttles/emergency.dm
+++ b/code/datums/shuttles/emergency.dm
@@ -19,6 +19,8 @@
 	. = ..()
 	if(!occupancy_limit && who_can_purchase)
 		CRASH("The [name] needs an occupancy limit!")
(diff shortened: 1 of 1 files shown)


Beeper-Skeeper:
60% off moment

ShoziOffical:
even with discount it would take like many hours to get them

Moltijoe:
> even with discount it would take like many hours to get them

correct, but this will at least prevent it from showing up
if there are designs you want to not be admin only, make your own pr

ToasterBiome (approved):

PR #22044, opened 2024-05-17
[BOX] Fixes library stairs being upside down

AMyriad:
Trying to tell what's top and what's bottom on these things is confusing yes, but the shading along the rails confirms the front/entrance is below the back of the room, and that the stairs do ascend.
image

(Lighter bars on top means it goes down)
image

# Testing
Will do

# Changelog
:cl:
mapping: The stairs in the library on Box are no longer upside down
/:cl:


neeshacark (approved):
checks out

PR #22048, opened 2024-05-17
Adjusts the effects of Stabilized red and Stabilized sepia slime cores

cowbot92:

Additionally adds a 1.5x stam damage modifier, and 1.5x stun mod to each stabilized extract.

Additionally adjusts sepia to now actually slow you instead of just resetting you back to base speed, as it was supposed to do.

Move speed is a potent tool in this game, and the effects you gain from these cores warrant some drawbacks. I did not want to nerf the actual speed they grant, as I feel like this is a suitable trade-off, making you more like a glass-cannon in certain cases.

# Testing

Stunmod testing - 2 telescopic baton hits to take a target down w/ 1.5x stunmod, even quicker with regular batons.
Sepia testing - works like the knife now, could honestly be -2/1 but not needed at this point
stun testing - going down keeps you down 1.5x longer.

# Wiki Documentation

Add the "you take extra stun/stam" damage part to the wiki

# Changelog

:cl:  

tweak: Adds stamina damage & stun duration modifiers to Red & Sepia Stabilized slime cores.

/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
+++ b/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
@@ -701,9 +701,8 @@ datum/status_effect/stabilized/blue/on_remove()
 /datum/status_effect/stabilized/sepia/on_apply()
 	if(ishuman(owner))
 		var/mob/living/carbon/human/H = owner
-		H.physiology.brute_mod *= 1.5
-		H.physiology.burn_mod *= 1.5
 		H.physiology.stamina_mod *= 1.5
+		H.physiology.stun_mod *= 1.5
 	return ..()
 
 /datum/status_effect/stabilized/sepia/tick()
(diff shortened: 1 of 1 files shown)


ofscarletdreams:
stabilized red speed is barely noticeable

cowbot92:
im just nuking them

Mqiib:
Honestly I think if you just added the stamina modifier that'd be enough.

Mqiib:
image

Hey so could you also actually make it so the sepia slows you down as well as speeds you up because, according to the code here, it only applies a -1 or a 0 modifier.

cowbot92:
> image
> 
> Hey so could you also actually make it so the sepia slows you down as well as speeds you up because, according to the code here, it only applies a -1 or a 0 modifier.

Adjusts as requested

Moltijoe (changes requested):
Damage modifications are too heavy handed
Stamina mod is totally fine though
Maybe include stunmod?
Perhaps look at arrhythmic dagger for sepia, that way it feels evel less consistent

cowbot92:
> arrhythmic dagger

Removed the damage modifiers, added the stun modifier. Additionally adjusted the sepia code to work exactly like the knife (uses rand now instead of some weird psuedo rand)

Moltijoe (approved):

PR #22050, opened 2024-05-17
Heavily adjusts the effects of Red and Sepia stabilized slime cores

cowbot92:

Removes the speed boost from Red Stabilized cores, replaces it with large blood generation if you're under the same blood amount.

Removes the speed boost/slowdown effect from Sepia Stabilized cores, replaces it with an effect which will make you black/white randomly.

Xenobiology speed issues have been an issue for awhile, and it's best to just cut the issue out at it's root rather than make half-hearted measures.

# Testing

the black effect from the sepia slime is actually really funny 

image

# Wiki Documentation

Remove mentions of speed change from both cores, replace with the effects noted above.

# Changelog

:cl:  
 
tweak: Tweaks how Red & Sepia slime cores work entirely 

/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
+++ b/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
@@ -701,15 +701,12 @@ datum/status_effect/stabilized/blue/on_remove()
 /datum/status_effect/stabilized/sepia/tick()
 	if(prob(50) && mod > -1)
 		mod--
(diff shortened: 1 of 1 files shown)


Therandomhoboo:
I get the nerfs, but this just strangles them out of being used. If this was the point then fair but yeah it'd be better if the stats were just nerfed.

cowbot92:
Wait im going to reopen the other one with a better idea

PR #22057, opened 2024-05-19
Reworks haloperidol to not be haloperidol

Moltijoe:
i was gonna just tweak haloperidol to be like selective blood purification nanites, but then i found out it's already selective for drugs
When i went to add toxins, i noticed that haloperidol is more of an anti-psychotic in real life, and as such makes less sense to be used for purging, so i renamed it to be a somewhat more suitable chemical

currently the chem purge design space is entirely consumed by charcoal and pentetic acid
This leaves things like haloperidol and calomel to be utterly useless
hobo is tweaking calomel to be stronger at purging with a lessened weakness, to make it the definitive super purge chem, as it's very easy to make.
I'm reworking haloperidol to be a selective purge chem as we don't have one that can function in a mix with other chems without purging all those other chemicals.

# Testing
i dunno how to chemistry and don't know how to test reagents

:cl:  
rscadd: Adds naloxone, a new chem that purges specifically drugs and toxins
rscdel: Removes haloperidol
/:cl:

### DIFF
--- a/code/datums/diseases/advance/advance.dm
+++ b/code/datums/diseases/advance/advance.dm
@@ -60,7 +60,7 @@
 										/datum/reagent/toxin/lipolicide, /datum/reagent/medicine/sal_acid
 									),
 									list(	// level 10
-										/datum/reagent/medicine/haloperidol, /datum/reagent/drug/aranesp, /datum/reagent/medicine/diphenhydramine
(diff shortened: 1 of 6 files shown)


Chubbygummibear (approved):
i was told to merge this

PR #22068, opened 2024-05-20
readds blood bros having kits

boodaliboo:
all of this code is originally from xoxoeyes i just added it back to the code 
this is the pr https://github.com/yogstation13/Yogstation/pull/9665/files 

but yea blood bros are like the antag you get pissed you rolled because its so boring we previously had them having kits and it made being one significantly more enjoyable

:cl:  
rscadd: Adds back them having kits
tweak: tweaked blood bros balance wise as its shit

/:cl:

### DIFF
--- a/code/game/gamemodes/brother/traitor_bro.dm
+++ b/code/game/gamemodes/brother/traitor_bro.dm
@@ -57,6 +57,7 @@
 		team.forge_brother_objectives()
 		for(var/datum/mind/M in team.members)
 			M.add_antag_datum(/datum/antagonist/brother, team)
+			equip_brother(M.current) // Yogs -- Readds BB Bundles, located in yogstation/code/game/gamemodes/brother
 		team.update_name()
 	brother_teams += pre_brother_teams
 	return ..()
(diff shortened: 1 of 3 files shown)


cowbot92:
Blood brothers have sufficient tools I feel with makeshift crafting, no reason to give them 20+ tc worth of items.

Moltijoe:
Blood brothers have sufficient tools I feel with makeshift crafting, no reason to give them 20+ tc worth of items.

Therandomhoboo:
Blood brothers have sufficient tools I feel with makeshift crafting, no reason to give them 20+ tc worth of items.

ynot01:
I don't really like makeshift items that much, but while kits are underutilized for traitors and this would definitely make them seen more, I don't think it's the right path.

- Blood brothers are supposed to be one antag, split into two people. Giving them kits makes them two traitors that are loyal to each-other.
- Being forced into the kit gimmick you spawn with is bad for player agency

I would rather blood brothers get like 4 telecrystals each or something, but even that could go very poorly

Moltijoe:
> I don't really like makeshift items that much

It might be worth creating additional items for makeshift crafting

Moltijoe:
don't even need 24 hours
sorry bood, i appreciate you coding, but this is more complicated than i think your current skills are capable of
and this doesn't seem like a feature anyone other than you wishes for

PR #22090, opened 2024-05-23
Security progression from arresting people

ynot01:
Added a new console and a new area

The area "Brig Cells" covers the interior of the brig cells 1/2/3 and the space between them (in case wardens feel like doing something creative)

The console is called a "departmental research console"

More departments can be added later very easily with this framework (as seen in `code\modules\research\reward_system\security.dm`) but for now it's just security. I don't have any ideas for other departments' progression.

Security gets research points (126/min per prisoner) whenever someone is incarcerated in either a brig cell or perma. If in a brig cell, the timer must be active and set to the prisoner's name to receive points.

Security can then use this console to unlock research nodes that contain gear that is printed at security lathes.

Security nodes may be locked behind nodes which are irrelevant to security, and will have to be unlocked by science (which will probably happen passively).

Image taken from mid-round (window vertically expanded) after security hadn't arrested anyone but science was researching stuff
image

I've wanted a form of security progression for a while. With this, there is more incentive to incarcerate rather than playing stanford prison experiment simulator. I expect a rise in security players with motivation to seek out petty crime rather than always focusing on validhunting, further reducing the TDM mindset.

I do not believe that shitty security behavior will be an issue, as admins are already very heavy handed against security players that step out of line.

Vectors of abuse should be non-issues as exploiters should get very bored very fast and will quickly realize how slow and pointless their efforts are. This is a passive thing that benefits security over a long time.

# Testing
Tested earning points
Tested redeeming points
Tested deconstructing & reconstructing console

# Wiki Documentation
New console: departmental research console
Security can incarcerate prisoners for a rate of 126 points per minute per prisoner and use those points to research nodes which are relevant to security.

# Changelog

:cl:  
rscadd: Added a departmental research console to the Security Office. Incarceration means free research!
/:cl:


RG4ORDR:
This is so stupid. However arresting people for points is amazing.

wonderinghost:
I feel like your change isnt great enough.

I recommend detaching security from station and are at the edge of the z level

Moltijoe:
Imagine security actually arresting for vandilism or trespassing, that actually sounds kinda interesting

AMyriad:
I'd love to see more security-relevant things like bodycams and fingerprint scanners added to that console, even if it shakes up the tech tree a bit. Always felt off to me that scientists research things that help us beat them up.

LassiH:
will this make mech weapons harder to get  or the better grenade casings ?

ynot01:
> I'd love to see more security-relevant things like bodycams and fingerprint scanners added to that console, even if it shakes up the tech tree a bit. Always felt off to me that scientists research things that help us beat them up.

> will this make mech weapons harder to get  or the better grenade casings ?

it's literally the RND techweb mirrored, the only things shown in the screenshot are what is currently unlockable, not everything you can buy. additionally, RND can still research these things for you

ynot01:
Testmerging on lowpop for performance testing

code/modules/research/reward_system/_reward_system.dm
```diff
@@ -0,0 +1,106 @@
...
+/obj/machinery/computer/department_reward/Initialize(mapload)
+	. = ..()
+	linked_techweb = SSresearch.science_tech
+
+/obj/machinery/computer/department_reward/proc/check_reward(delta_time)
+	return 0
+
+/obj/machinery/computer/department_reward/process(delta_time)
+	if(stat & (NOPOWER|BROKEN))
+		return // we dont return PROCESS_KILL because it's already handled in /obj/machinery
```
Bibby0110: comment doesn't really make sense

you probably don't want to ever use PROCESS_KILL

code/modules/research/reward_system/_reward_system.dm
```diff
@@ -0,0 +1,106 @@
...
+				if(design_count <= 14)
+					node_entry["designs"] |= design_id
+			data["nodes"] += list(node_entry)
+	return data
+
+/obj/machinery/computer/department_reward/ui_act(action, params)
+	if(..())
+		return
+
+	if(stat & (NOPOWER|BROKEN))
```
Bibby0110: is_operational() instead?

code/modules/research/reward_system/_reward_system.dm
```diff
@@ -0,0 +1,106 @@
...
+				return
+			if(linked_techweb.researched_nodes[node_id])
+				return
+			if(!linked_techweb.available_nodes[node_id])
+				return
+			var/datum/techweb_node/node = SSresearch.techweb_node_by_id(node_id)
+			var/node_price = node.get_price(linked_techweb)[TECHWEB_POINT_TYPE_GENERIC]
+			if(!node_price)
+				return
+			if(node_price >= points)
```
Bibby0110:
```suggestion
			if(node_price > points)
```

code/modules/research/reward_system/security.dm
```diff
@@ -0,0 +1,35 @@
...
+/obj/machinery/computer/department_reward/security/Initialize(mapload)
+	nodes_available = GLOB.security_nodes
+	. = ..()
+
+/obj/machinery/computer/department_reward/security/check_reward(delta_time)
+	. = 0
+
+	// Criminals in brig cells with an active timer
+	var/list/criminals = list()
+	for(var/obj/machinery/door_timer/timer in SSmachines.processing)
```
Bibby0110: could always make a GLOB for this if you're worried about iterating through like 1000 machines

code/modules/research/reward_system/security.dm
```diff
@@ -0,0 +1,35 @@
...
+		if(timer.timing && timer.desired_name)
+			criminals += timer.desired_name
+		CHECK_TICK
+	for(var/area/security/brig_cell/cell in GLOB.areas)
```
Bibby0110: ditto as above, quite a few areas

Bibby0110 (commented):
very cool

code/modules/research/reward_system/_reward_system.dm
ynot01: if there are exactly 2500 points we want to be able to buy something that costs 2500 points

ynot01: oops

neeshacark:
i do not like the idea of incentivizing giving people higher arrest times - it could easily turn into a toxic mess

neeshacark (approved):

ynot01:
I counter the arguments of petty crime arrests rising with the answer that modern security simply ignores all petty crime in favor of antag hunting. This leads to bad experiences for the law-abiding, who have to deal with troublemakers while being ignored by security. For instance, just last round, the QM was repeatedly breaking into science, the single scientist went ignored while security was looking for heretics because of a single rune, and was eventually charged with manslaughter for trying to subdue the QM himself.

You can call this shitsec, or hell, you might call it justified due to the presence of antags, but I see a clear need of incentive.

Cartlord:
not entirely a fan of this idea, but, i've yet to see any better ones for dealing with the scenarios you're describing.
perfect is the enemy of good, godspeed to ye

Moltijoe (approved):
i do worry this could end up taking science's job away from it if taken too far
but for now, this is an idea that could promote "healthier" gameplay
but if it proves to do the exact opposite, changes will need to be made
at worst, reverted

i'm going to testmerge it for now, just to see the reception

Therandomhoboo:
I dislike this for several reasons:
It takes away RnD's point of their job (who cares if they get spammed by sec, or don't do it actively it is an IC issue. Like guys that have worked a job, you know that other departments or w.e will be shit at times)

It rewards arresting people over anything which yes is good, it doesn't really teach them how to be sec. Why do we need to get serotonin to sec team like they're tiktok kids? What next, we're going to stick another video in the corner for sec players so they don't lose interest? My point is, this doesn't teach sec how to sec, it just makes them get a foot up if they arrest people and slap as many charges as they can

It opens up a lot of ahelps between the two, more so than normal I think. You think ahelps are bad now, they're usually over major crimes or w.e, i suspect this will cause ahelps at any chance when people get arrested over any issue. So, good luck with that

There is no real reason for this to be in? Why do sec need to take parts from every single job? Or better yet, why are they given now an option to ignore interaction with a department? It makes no sense and with the add-on of perma prisoners giving points too, no one can understand that it isn't going to be one arrested every time, you're going to have multiple which will add up. 
3 people doing 5 minute sentences EACH which is what a minor/moderate crime level will give 15 minutes worth of points which is only 5 minutes off giving hardlight weaponary.

In conclusion, this is a senseless sec buff. I  think we should pivot away from antags & sec now and try focusing on making other roles fun? We've bloated them enough now and they keep getting more love just because they are the "most" played which is a bad mindset we have.

We should make every role fun, and so far we aren't.

ynot01:
> Why do we need to get serotonin to sec team like they're tiktok kids? What next, we're going to stick another video in the corner for sec players so they don't lose interest?
> 
> We should make every role fun, and so far we aren't.

Contradictory points

> It opens up a lot of ahelps between the two, more so than normal I think. 

Admins are capable of handling security. What you are proposing here is that security should have a less active role so that people adminhelp less, which I disagree with.

> There is no real reason for this to be in?

Please see the section labeled "Why is this good for the game?"

> 3 people doing 5 minute sentences EACH which is what a minor/moderate crime level will give 15 minutes worth of points which is only 5 minutes off giving hardlight weaponary.

This statement is extremely misleading; whether intentionally or otherwise I cannot tell.

Yes, it is true that 15 minutes of crime time means 1890 research points. (Mind you, science generates ~3100 points/min passively)

Yes, it is true that hardlight weaponry costs 2500 research points.

Hardlight weaponry **also** has a total of 27000 research points in prerequisite research nodes before it even shows up on the computer, most of which cannot be researched using this console.

It is not as if security can start the shift, detain some people for a collective time of 20 minutes, and receive hardlight technology with no assistance from Science.

Therandomhoboo:
> > Why do we need to get serotonin to sec team like they're tiktok kids? What next, we're going to stick another video in the corner for sec players so they don't lose interest?
> > We should make every role fun, and so far we aren't.
> 
> Contradictory points

How so? I said every role fun, not a specific role so that was a deflect

> > It opens up a lot of ahelps between the two, more so than normal I think.
> 
> Admins are capable of handling security. What you are proposing here is that security should have a less active role so that people adminhelp less, which I disagree with.

Admins are indeed capable of handling security but this is increasing it by a fair degree, I didn't say sec should have a less active role, and it isn't really increasing an active role via point reward scheme is it?

> > There is no real reason for this to be in?
> 
> Please see the section labeled "Why is this good for the game?"

Read it and responded to it, hence "no real reason" cause you said shitters would get bored (yeah like psych griefer who probably is still here) and that admins can handle it.


> > 3 people doing 5 minute sentences EACH which is what a minor/moderate crime level will give 15 minutes worth of points which is only 5 minutes off giving hardlight weaponary.
> 
> This statement is extremely misleading; whether intentionally or otherwise I cannot tell.

Going based off your photo you put in, it didn't show any pre-reqs for it to buy. Maybe I misread the code or something but it didn't clearly indicate to me. Maybe it was the mid-round that should of tipped me off but yes it wasn't intentional however, my point still stands. Multiple people giving points per arrest times, alongside if any perma prisoners will still give research points which stacks and if this is "worse" than sci. Yet again why does it need to be in? The interaction is there, you're still not justifying why you need to sever that interaction properly.

If admins can handle "sec", Sec and Sci can handle their interactions with each other no? Or we just picking and choosing now what fits the narrative?

Moltijoe:
Encourages unfun play from security
Rewards overlap with science's job

PR #22097, opened 2024-05-23
Makes ethereals get slowly charged by starlight + energy projectiles + electrocute

Moltijoe:
Being within 2 tiles of stars will charge up to a certain point that is safe
Being shocked or shot with an energy projectile will charge based on the damage dealt
Discharge will also always reduce nutrition by 100 rather than setting it to be almost_full

ethereals already have a lot of overlap with robotic species, especially after the powerhungry refactor sarah did a while ago
This gives them something thematically fitting that differentiates them

also, though this is a minor thing, the other two species can carry around a battery if the power goes out
ethereals need to carry around something that only come from vending machines or specific cooking practices

The other changes are just to add more funny ways for discharge to trigger

# Testing
gotta

:cl:  
rscadd: Makes ethereals get slowly charged by starlight
rscadd: Makes ethereals get rapidly charged by electrocution and energy projectiles
tweak: Ethereal discharge now discharges a static amount of electricity
bugfix: Ethereals can now properly stop discharging by releasing power into an apc
bugfix: Ethereals can no longer discharge multiple times at once
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/ethereal.dm
+++ b/code/modules/mob/living/carbon/human/species_types/ethereal.dm
@@ -1,3 +1,5 @@
+#define DISCHARGE_LOSS 100 //how much power is lost during a discharge
+
 /datum/species/ethereal
 	name = "Ethereal"
 	id = "ethereal"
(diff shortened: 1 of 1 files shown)


wonderinghost:
Ethereals in the distant future

ilikesaying:
> Being within 2 tiles of stars will charge up to a certain point that is safe

GiveMeYourEnergy

wonderinghost:
This is forever going to be cursed with anime references and Ultraman references

Chubbygummibear (approved):
cowbot said this is good to go. idk because i don't play :)

PR #22116, opened 2024-05-28
Renames lava component and fixes a couple bugs

Moltijoe:
renaming the component implies it can be used for more things than just lava

also, changes it to rely on a damage callback proc rather than needing to create new subtypes for each different type

i was unable to make it possible to apply to non-turf things for the time being, i may come back to this

also, fixes a bug/oversight where a new object entering a lava tile would do damage to everything already on that tile

a list was returning when it should have continued
this made it impossible for plasmeme limb replacement via plasma rivers

# Testing
image

:cl:  
bugfix: fixes an old bug involving lava doing more damage under specific circumstances
bugfix: fixes an new bug involving plasma rivers not applying plasmeme limbs
/:cl:

### DIFF
--- a/yogstation/code/datums/components/lingering.dm
+++ b/yogstation/code/datums/components/lingering.dm
@@ -230,10 +230,10 @@
 						plasma_parts += NN
 
 				if(prob(35)) //checking if the delay is over & if the victim actually has any parts to nom
-					playsound(PP, 'sound/effects/wounds/sizzle2.ogg', 80, TRUE)
 					PP.adjustToxLoss(15)
 					PP.adjustFireLoss(25)
 					if(length(plasma_parts))
+						playsound(PP, 'sound/effects/wounds/sizzle2.ogg', 80, TRUE)
 						var/obj/item/bodypart/NB = pick(plasma_parts) //using the above-mentioned list to get a choice of limbs to replace
 						if(PP.stat != DEAD)
 							PP.emote("scream")
@@ -246,6 +246,7 @@
 						replacement_part.replace_limb(PP)
 						qdel(NB)
 					else if(!length(robo_parts)) //a person with no potential organic limbs left AND no robotic limbs, time to turn them into a plasmaman
+						playsound(PP, 'sound/effects/wounds/sizzle2.ogg', 80, TRUE)
 						PP.ignite_mob()
 						PP.cure_husk(BURN) //cure the probable husk first
 						PP.set_species(/datum/species/plasmaman)
--- a/yogstation/code/datums/components/lingering.dm
+++ b/yogstation/code/datums/components/lingering.dm
@@ -2,14 +2,26 @@
 	dupe_mode = COMPONENT_DUPE_ALLOWED
 	/// A callback on the parent to be called when it tries to apply an affect on an atom
 	var/datum/callback/affect_callback
+	/// A list of things that if even one is on the tile, will prevent the effect
+	var/list/safeties_typecache
 
-/datum/component/lingering/Initialize(datum/callback/affect_callback)
+/**
+ * Lingering component
+ * 
+ * essentially caltrops, but it keeps triggering every SSobj tick
+ * 
+ * vars:
+ * * affect_callback (required) A callback that triggers a proc that returns true or false, false will stop the processing of the component
+ * * safeties_typecache (optional) A list of typepaths that if they're on the same tile will prevent the effect
+ */
+/datum/component/lingering/Initialize(datum/callback/affect_callback, list/safeties_typecache)
 	if(!istype(parent, /turf))
 		return COMPONENT_INCOMPATIBLE
 	if(!affect_callback)
 		stack_trace("Lingering component has been applied to a turf without an affect callback proc")
 		return COMPONENT_INCOMPATIBLE
 	src.affect_callback = affect_callback
+	src.safeties_typecache = safeties_typecache
 
 /datum/component/lingering/RegisterWithParent()
 	. = ..()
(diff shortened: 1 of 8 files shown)


yogstation/code/datums/components/lingering.dm
```diff
@@ -0,0 +1,78 @@
+/datum/component/lingering
+	dupe_mode = COMPONENT_DUPE_ALLOWED
+	/// A callback on the parent to be called when it tries to apply an affect on an atom
+	var/datum/callback/affect_callback
+
+/datum/component/lingering/Initialize(datum/callback/affect_callback)
```
alexkar598: The callback is confusing, add a comment somewhere describing the parameters and the return value

yogstation/code/datums/components/lingering.dm
```diff
@@ -0,0 +1,78 @@
...
+
+/datum/component/lingering/proc/affect_stuff(AM, delta_time = 1)
+	. = 0
+
+	if(!affect_callback)
+		stack_trace("Lingering component has been applied to a turf without an affect callback proc") //covering my bases
+		return
+
+	if(is_safe())
+		return FALSE
```
alexkar598: Safe preventing processing seems like a bug to me

Won't this allow standing in lava if the turf is no longer safe while you're standing on it?

yogstation/code/datums/components/lingering.dm
```diff
@@ -0,0 +1,78 @@
...
+	UnregisterSignal(parent, COMSIG_ATOM_ENTERED)
+	STOP_PROCESSING(SSobj, src)
+	. = ..()
+	
+////////////////////////////////////////////////////////////////////////////////////
+//---------------------------------Triggers---------------------------------------//
+////////////////////////////////////////////////////////////////////////////////////
+/datum/component/lingering/proc/hitby(datum/source, atom/movable/AM, skipcatch, hitpush, blocked, datum/thrownthing/throwingdatum)
+	if(affect_stuff(AM))
+		START_PROCESSING(SSobj, src)
```
alexkar598: Can you confirm if this the hitby signal is needed?

In the case of dense turfs, should the turf really affect an object bouncing off the turf?
In the case of non dense turfs, doesn't entered cover this already?

yogstation/code/datums/components/lingering.dm
```diff
@@ -0,0 +1,78 @@
...
+/datum/component/lingering/process(delta_time)
+	if(!affect_stuff(null, delta_time))
+		STOP_PROCESSING(SSobj, src)
+
+////////////////////////////////////////////////////////////////////////////////////
+//---------------------------------safety check-----------------------------------//
+////////////////////////////////////////////////////////////////////////////////////
+/datum/component/lingering/proc/is_safe()
+	//if anything matching this typecache is found in the lava, we don't burn things
+	var/static/list/lava_safeties_typecache = typecacheof(list(/obj/structure/lattice/catwalk, /obj/structure/stone_tile))
```
alexkar598: This should be configurable by the parent. The typecache probably should become a global so that all lavalike turfs may reuse the same typecache for consistency

yogstation/code/datums/components/lingering.dm
```diff
@@ -0,0 +1,78 @@
...
+
+	var/turf/open/place = get_turf(parent)
+
+	var/list/found_safeties = typecache_filter_list(place.contents, lava_safeties_typecache)
+
+	for(var/obj/structure/stone_tile/S in found_safeties)
+		if(S.fallen)
```
alexkar598: Dunno how else this can be done, but I'm not a fan of the snowflake check in the component

yogstation/code/datums/components/lingering.dm
```diff
@@ -0,0 +1,78 @@
...
+			LAZYREMOVE(found_safeties, S)
+	return LAZYLEN(found_safeties)
+
+/datum/component/lingering/proc/affect_stuff(AM, delta_time = 1)
+	. = 0
```
alexkar598:
```suggestion
	. = FALSE
```

yogstation/code/datums/components/lingering.dm
alexkar598: You already have a . value, this is redundant

```suggestion
		return
```

yogstation/code/datums/components/lingering.dm
```diff
@@ -0,0 +1,78 @@
...
+	var/turf/open/place = get_turf(parent) //otherwise, apply to every object on the turf
+
+	if(!place)
+		return
+
+	for(var/thing in place.contents)
+		if(thing == parent)
+			continue
+		if(affect_callback.Invoke(thing, delta_time))
+			. = 1
```
alexkar598:
```suggestion
			. = TRUE
```

yogstation/code/datums/components/lingering.dm
```diff
@@ -0,0 +1,78 @@
...
+	var/turf/open/place = get_turf(parent)
+
+	var/list/found_safeties = typecache_filter_list(place.contents, lava_safeties_typecache)
+
+	for(var/obj/structure/stone_tile/S in found_safeties)
+		if(S.fallen)
+			LAZYREMOVE(found_safeties, S)
+	return LAZYLEN(found_safeties)
+
+/datum/component/lingering/proc/affect_stuff(AM, delta_time = 1)
```
alexkar598: Not a massive fan of the proc handling two different cases (affect_stuff(null, num) and affect_stuff(obj, null)). Could be worth splitting into 2 procs

yogstation/code/datums/components/lingering.dm
Moltijoe: it's for throwing items that land in the lava
they trigger entered, but since they're still being thrown won't start processing

yogstation/code/datums/components/lingering.dm
Moltijoe: yeah, i got no idea either, short of reworking stone tiles

yogstation/code/datums/components/lingering.dm
```diff
@@ -26,53 +38,69 @@
...
+		stack_trace("Lingering component has been applied to a turf without an affect callback proc") //covering my bases
+		return
 
 	var/turf/open/place = get_turf(parent) //otherwise, apply to every object on the turf
 
 	if(!place)
 		return
 
 	for(var/thing in place.contents)
+		if(is_safe()) //we don't want to damage anything, but if there's something here, make sure to keep processing just in case it stops being safe
```
alexkar598: Unless I'm missing something, the safety check should be outside the loop to prevent calling it multiple times needlessly

alexkar598 (commented):
Other than this, LGTM

yogstation/code/datums/components/lingering.dm
Moltijoe: ah, true, i want it to return in the loop so it only returns if theres something there, but no reason to call is_safe multiple times

ToasterBiome (approved):

PR #22126, opened 2024-05-29
IV drip and tank holder tweaks

warface1234455:
Changed description and name of anes tank into a normal tank holder
You can now craft iv drip and iv drip has been removed from the cargo pack because it would be obsolete
You can now turn iv drip into a tank holder by clicking it with breath mask instead of using screwdriver while holding mask, the duration to transform is now 5 seconds instead of 10s
You can now turn tank holder back into iv drip by altclicking the tank holder

Qol for tank holder and iv drip, allows medical staff to build an iv drip instead of ordering one when there is no iv drip in medical and there's an emergency

# Testing

https://github.com/yogstation13/Yogstation/assets/89688125/d2b38ff1-8f41-4f6c-91d5-4689885a7ad1

# Wiki Documentation
in document of change

# Changelog

:cl:  
spellcheck: Changed description and name of anes tank into a normal tank holder
tweak: You can now craft iv drip and iv drip has been removed from the cargo pack because it would be obsolete
tweak: You can now turn iv drip into a tank holder by clicking it with breath mask instead of using screwdriver while holding mask, the duration to transform is now 5 seconds instead of 10s
tweak: You can now turn tank holder back into iv drip by altclicking the tank holder
/:cl:


code/game/machinery/iv_drip.dm
```diff
@@ -237,7 +237,7 @@
 
 	. += span_notice("[attached ? attached : "No one"] is attached.")
 
-/obj/machinery/iv_drip/screwdriver_act(mob/living/user, obj/item/I)
+/obj/machinery/iv_drip/attacked_by(obj/item/attacking_item, mob/living/user)
```
Moltijoe: use attackby and check if the attacking item is specifically the mask being used

Moltijoe (changes requested):

code/game/machinery/iv_drip.dm
```diff
@@ -237,18 +237,15 @@
 
 	. += span_notice("[attached ? attached : "No one"] is attached.")
 
-/obj/machinery/iv_drip/screwdriver_act(mob/living/user, obj/item/I)
+/obj/machinery/iv_drip/attackby(obj/item/attacking_item, mob/living/user)
 	. = ..()
```
Moltijoe: move this to the end and make the if() return after the qdel

Moltijoe (changes requested):

code/datums/components/crafting/recipes.dm
```diff
@@ -305,6 +305,13 @@
 	category = CAT_STRUCTURES
 	time = 15 SECONDS
 
+/datum/crafting_recipe/iv_drip
+	name = "IV drip"
+	reqs = 	list(/obj/item/stack/rods = 4, /obj/item/stack/sheet/metal = 1)
+	result = /obj/machinery/iv_drip
+	category = CAT_STRUCTURES
+	time = 5 SECONDS
```
Moltijoe: should probably require a welding tool or screwdriver, and maybe some plastic

code/modules/surgery/anesthesia_machine.dm
```diff
@@ -1,6 +1,6 @@
 /obj/machinery/anesthetic_machine
-	name = "Anesthetic Tank Holder"
-	desc = "A wheeled machine that can hold a tank and distribute the gas inside using a breath mask."
+	name = "Tank Holder"
```
Moltijoe: tank holder is a bit generic, maybe something that references it's use in a medical setting or that it's tank related
try looking for a real life thing to name it after perhaps

code/modules/surgery/anesthesia_machine.dm
```diff
@@ -40,8 +40,29 @@
 	. = ..()
 	if(retract_mask())
 		visible_message("<span class='notice'>[user] retracts the mask back into the [src].</span>")
+	else if(attached_tank)// If attached tank, remove it.
+		attached_tank.forceMove(loc)
+		to_chat(user, "<span class='notice'>You remove the [attached_tank].</span>")
```
Moltijoe: change all these chat prints to use the span procs rather than how they currently are
like span_notice("example text") that sorta thing

Moltijoe (changes requested):

code/datums/components/crafting/recipes.dm
ShoziOffical: screwdriver is alright

code/modules/surgery/anesthesia_machine.dm
ShoziOffical: ok

code/modules/surgery/anesthesia_machine.dm
ShoziOffical: knew you would request this

code/modules/surgery/anesthesia_machine.dm
```diff
@@ -39,28 +39,39 @@
...
+		update_appearance(UPDATE_ICON)
+		if(mask_out)
+			retract_mask()
+
+/obj/machinery/anesthetic_machine/AltClick(mob/user)
+	. = ..()
+	if(attached_tank && mask_out)
+		to_chat(user, span_warning("Disconnect the tank from the person first!"))
+		return
+	else
```
Moltijoe, line 56: Maybe add a delay to changing it back to an iv drip
That way if someone accidentally alt click it they have a window to cancel it

ShoziOffical: fair

Moltijoe (approved):
looks fine

PR #22131, opened 2024-05-30
TGUI Stripping

ynot01:
Ports https://github.com/tgstation/tgstation/pull/57889

Old PR (2021), can be updated as we update required features from tg (i have not looked into it)

In any case, much better than what we had previously
image

TGUI > no TGUI

# Testing
Setting internals works, pockets work, stripping works

# Changelog

:cl:  
rscadd: Added TGUI stripping menu
bugfix: Fixed xenos not having footsteps
/:cl:


AMyriad:
methinks these buttons could be organized better to avoid the great dedicated row of shoe
image
but hey it's leagues better than what we got rn so I'm not complaining

ynot01:
> methinks these buttons could be organized better to avoid the great dedicated row of shoe image but hey it's leagues better than what we got rn so I'm not complaining

This is exactly how /tg/ had it (as can be seen in linked PR), and it's best to keep things as similar for future upgrading

Cartlord:
I think this is an un-necessary change, and I think the squished/stretched icons in the menu look ugly. I also agree that it's not well-organized.

ToasterBiome (changes requested):
can we make the icons not look awful?

ynot01:
> can we make the icons not look awful?

Forced all icons 64px, required some resizing
No I don't know how to make the sides not touching the edges
No I don't know why the internals icon is so small and it won't get bigger
image

ToasterBiome (changes requested):
make the underlay icons also not poorly scaled

ynot01:
> make the underlay icons also not poorly scaled

Done 
image

ynot01:
image

AMyriad:
> image

oooooooh very nice
would this by chance change based on UI preference?

ynot01:
> oooooooh very nice would this by chance change based on UI preference?

No

AMyriad:
> No

damn fair enough :^)

ToasterBiome (approved):

PR #22150, opened 2024-06-02
Simplifies Preterni

Moltijoe:
Removes limb brute resistance
-it wasn't easy to convey they got this, and now that the limbs have innate emp resistance rather than the species, they still have a reason to not swap to borg limbs
-matters more in combat than burn reduction

Removes burnmod
-their limbs already have burn reduction, which makes it sorta counter-intuitive, "are they weak to burn? strong against it?"

Removes lung interactions to heat
-the intention was to make them resistant to breathing hot gases, but all it did was make it do more damage, but at different breakpoints
-when i initially added this, i didn't know how breathing hot gases dealt damage

Removes virus resistance(?)
-Virus resistance, much like rad resistance, is binary, either you're immune or you're not
-Had no effect on gameplay aside from confusing people into thinking they're like IPCs (who are immune)

aint no one got time to read all that
image
they also have minor things that just aren't clear without code diving

# Testing
just number changes

:cl:  
tweak: Removes preternis inconsistent virus interactions
tweak: Removes burnmod from preternis
tweak: Removes brute resistance from preternis limbs
tweak: Removes preternis lungs acting weirdly with hot gases
/:cl:

### DIFF
--- a/yogstation.dme
+++ b/yogstation.dme
@@ -4037,7 +4037,6 @@
 #include "yogstation\code\datums\components\storage\storage.dm"
 #include "yogstation\code\datums\diseases\_MobProcs.dm"
 #include "yogstation\code\datums\diseases\cluwnification.dm"
-#include "yogstation\code\datums\diseases\advance\advance.dm"
 #include "yogstation\code\datums\diseases\advance\symptoms\confusion.dm"
 #include "yogstation\code\datums\diseases\advance\symptoms\heal.dm"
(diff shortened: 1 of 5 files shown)


Therandomhoboo:
I was concerned about the burnmod but like said, limbs already got burn reduct so it is just counter-intuitive which balances them out pretty much so its w.e, does mean they get a slight buff but eh, small.

ynot01 (approved):

PR #22170, opened 2024-06-04
[BOX] Rearranges medbay to have a bigger cryo setup (future bricklaying)

wonderinghost:
prepares medbay for an eventual cryotube change 
break room has merged with the main room for down the line change to cryotubes
Note: check lockers for missing items 
image
also fixes some minor decal errors
also deletes one of those beds below surgery for where

gives boxbay a bit more room to move about and to kind of future proof it for any other changes. and for this case an update to the cryotube. what i have planned is to have it use up/convert gasses that need a bigger atmos setup to complete, gasses like nitrogen will have uses as say a body preservative or if it heals a ton of other damage types the tank will produce a ton of miasma. coming when i get around to it or if someone else wants to take it up

# Testing
ran around to make sure something wasnt sticking through the wall or is missing

# Changelog

:cl:  
mapping: alters boxbay lobby. sorry but your break room has been drafted
mapping: fixes decal errors hidden around medbay
/:cl:


ShoziOffical:
IMG_20240604_092046
Remove this wall, its taking the space

wonderinghost:
you can have more than one thing on a wall. besides its a fire extinguisher on one side and the other intercom

ShoziOffical:
> you can have more than one thing on a wall. besides its a fire extinguisher on one side and the other intercom

You can put the intercom and extinguisher somewhere else

Replace that wall with something useful

wonderinghost:
> > you can have more than one thing on a wall. besides its a fire extinguisher on one side and the other intercom
> 
> You can put the intercom and extinguisher somewhere else
> 
> Replace that wall with something useful

StrongDMM-2024-06-03 23 50 14

patpol4:
Looks bad, i think the break room feels too open. Maybe use curtains or windows? The microwave is in a weird spot, the decals are not right.

ktlwjec0:
Paramedic office feels weirdly long it's like a corridor (decals need connecting in the lower-right if you don't change that). You left the decals from the wall you removed. Some different decals around the freezer/oxy tanks would be nice. Is the microwave actually needed now there's no break room?

Bester22:
Is the supposed bricklaying for a cryotube setup for something in game or for a future pr? The change to the storage room feels super... off. like how the brute and burn kits aren't right next to each other for some reason. It also feels like you just ran out of room for the vendors and went "haha that looks good enough." after shoving them in corners and infront of windows. Why is the paramed room extended a whole tile down now as well? it looks like a hallway rather than our random storage closet we shove our death prone parameds in. And since the breakroom doesn't exist why is there even a microwave in there at all? Also why the hell did you just shove the vendors into main hall? All and all, this just seems like a medbay change from someone who doesn't play medical at all.

Bester22:
This actively gets WORSE the more I look at it, Why did you take a tile of space from chemistry and give it to the sec post? Where did the backup oxygen tank disappear to? Also, why did you move the nanomed behind the curtains where we put SSD people?

ktlwjec0:
i cba to map this properly but here just pretend i havent left random decals and cables everywhere

image
Squish the room down more so it doesn't feel like a corridor. Another window can be added now there's not a wall there, which will make the room feel bigger. Add that decal under the suit storage. Move the air alarm somewhere else like to the west wall, and put a light switch where it was. Move the shutter button to above the consoles on the wall.

You can then move sec down a space, which will let you keep chemistry how it was. Since chemistry is now a literal square, which looks terrible. Move the light switch in the sec post to be on the east wall. I'd swap the console and vendor back to how it was, there was nothing wrong with their placement.

image

Windows/windoors should be like this. Swap the burn and toxin kits so the brute+burn kits are together. Hypomed should be moved, I don't think you can access it now without opening the windoor. There's space for the spare oxy canister if you wanted and probably find nicer decals for it. Just play around with the layout some more, some things don't need changing as much.

Move the nanomed back to where it was so it's not blocked by curtains. Remove the hallway vendors and the microwave - no need for them now there's no break room. They can go to the kitchen.

I got bored now so good luck with the rest of it

wonderinghost:
> This actively gets WORSE the more I look at it, Why did you take a tile of space from chemistry and give it to the sec post? Where did the backup oxygen tank disappear to? Also, why did you move the nanomed behind the curtains where we put SSD people?

Because there was another 02 tank in morgue maint. 

At the time it just felt better to shift that side of med down and up (sleep deprived).

Also that was like that before unless sleep deprived me embedded it into the wall

wonderinghost:
If it was a setup you would have seen a more elaborate setup. Cryotube in its current state only detects if there is 02, pluox, and healium in the line to work. For all intents in purposes I can just remove the 2nd oxygen tank and it will be fine because it doesn't use up gas.

Bester22:
Anyways,  I'm fully against this, change medbay from its standard setup to the one you want on a round-by-round basis rather than trying to make medbay one of your atmos pet projects that you will inevitably abandon like every atmos player ever.

Moltijoe:
i don't think i like this, making the break room that open seems unideal
if cryo is gonna need more engineering space to make function, perhaps add an engineering access room connected to maints similar to icemeta, or how viro has on every map

wonderinghost:
Could close this one and work on a sublayer for box. Problem is that most shit isn't craftable in game like multiz power or disposals

Moltijoe:
> Could close this one and work on a sublayer for box. Problem is that most shit isn't craftable in game like multiz power or disposals

Make it craftable then
Worst case, they can use bluespace pipes if they need more

wonderinghost:
coming back with multi box... dont @ me

PR #22171, opened 2024-06-04
Fixes a minor logic error that prevented ethereals from discharging

Moltijoe:
< vs >

# Testing
gotta

:cl:  
bugfix: Fixes a minor logic error that prevented ethereals from discharging
/:cl:


ynot01 (approved):
image

PR #22174, opened 2024-06-05
Keypads 2.0: Its THE hacker to you

Scrambledeggs00:

Hacking keypads is boring as it is just standing around for 40 seconds to unlock one. This PR changes that by replacing the old way with three new ways. Once unlocked, you can then use a multitool on it to reset its code, which takes 10 seconds to do. It also adds in lock strength to keypad locks that determines what can force them and if they can have their code reset. Briefcases can be forced, hacked, and emaged while safes can only be hacked or emaged. A final level of strength makes it so it can only be hacked or emaged and cannot have its code reset, but I did not add any items that use it; it is there for others to use.

First way is by forcing the lock with an item. Currently, these items are jaws of life, changeling arm blade, umbral tendrils, 
mantis blade, and light eater. This can only be done on briefcases as safes are too strong to be forced. I originally planned a do_after for this to balance it, but it kept throwing errors so instant hacking I guess :I

Second way is with an emag which can force open any keypad lock.

Last way is by checking its memory for a game of Cows and Bulls. If an incorrect valid code was entered and you use a multitool on it while locked, you will get messages telling you the input code, how many digits were correct and in the right place, and how many were correct but in the wrong place.

When forced open with an item or emag, the item's sprite will blink and its description will say that its lock is bugged. To undo this, simply reset the keypad by pressing the R button in its UI. 

Makes hacking keypad locks more interesting and not boring.

# Testing

Video showing the different methods in action:
https://www.youtube.com/watch?v=GbwMsVbh2rI

# Spriting

Forced open sprite:
SuitcasesHack

# Wiki Documentation

To force a briefcase open, either use a jaws of life, changeling arm blade, umbral tendrils, mantis blade, light eater, or emag on it. To get clues to the code, enter a five digit code, open its service panel with a screwdriver, and then use a multitool on it. Doing so will send messages telling you the code you tried, along with the number of correct digits in the correct place and correct digits in the wrong place. Using a multitool on it while unlocked and with its service panel open will start a 10 second timer. At the end of it, the keypad's code will be reset and you can give it a new code.

Safes work the same except they can only be forced open with an emag.

To fix a keypad lock that has been forced, simply press the R in the keypad's UI.

# Changelog

:cl:  
tweak: Changed how hacking keypad locks works
imageadd: Added sprites for forced keypad items 
/:cl:


MommaDragon:
Hackerman (boring)

cuackles (approved):

code/datums/components/keypad_lock.dm
```diff
@@ -16,6 +16,18 @@
...
 	var/panel_open = FALSE
+	//Last viable keypad_input
+	var/memory_code = ""
+	//If it is forced
+	var/is_forced = FALSE
+	//Strength of lock: determines how easy/hard it is to bypass lock
+	//0 = force open and hacking | 1 = hacking only | 2 = hacking only and no reseting memory
+	var/lock_strength = 0
+	//List of items that can force keypad locks
+	var/list/keypad_tools = list(/obj/item/jawsoflife,
```
Moltijoe, line 27: Might be worth adding the light eater to this if the mantis blade and armblade already are

code/datums/components/keypad_lock.dm
```diff
@@ -61,42 +77,111 @@
...
+	//Reset code if unlocked, it has an access_code, and if lock is weak enough
+	if(!src.lock_status)
+		if(lock_strength == 2)
+			user.balloon_alert(user, "Encryption too strong!")
+			return
+		else
+			if(access_code == "")
+				user.balloon_alert(user, "Memory already reset!")
+				return
+			user.balloon_alert(user, "Reseting memory")
```
Moltijoe:
```suggestion
			user.balloon_alert(user, "Resetting memory")
```

code/datums/components/keypad_lock.dm
```diff
@@ -61,42 +77,111 @@
...
+			if(!tool.use_tool(parent, user, 10 SECONDS))
+				user.balloon_alert(user, "Interupted!")
```
Moltijoe:
```suggestion
				user.balloon_alert(user, "Interrupted!")
```

Moltijoe (changes requested):

Moltijoe (approved):

PR #22183, opened 2024-06-07
[DONUT] Adds a second chem heater in a way I think looks bad because people keep crawling up my ass about it every few months (just use a damn lighter smh)

neeshacark:
# Document the changes in your pull request

title

# Why is this good for the game?
it looks ugly

# Changelog

:cl:  cark
mapping: This is a bad mapping change that adds a second chem heater to Donut chemistry
/:cl:

PR #22184, opened 2024-06-07
[DONUT] fills donut chem with items you can heat beakers with because its better than a chem heated, alternative to #22183

neeshacark:
image


MommaDragon:
image
That's an Ided.

PR #22191, opened 2024-06-09
readds blood bros having kits

boodaliboo:
all of this code is originally from xoxoeyes i just added it back to the code
this is the pr https://github.com/yogstation13/Yogstation/pull/9665/files

but yea blood bros are like the antag you get pissed you rolled because its so boring we previously had them having kits and it made being one significantly more enjoyable

🆑
rscadd: Adds back them having kits
tweak: tweaked blood bros balance wise as its shit

/:cl:


Moltijoe:
please do not make new prs for a feature when the original has been closed for a reason

PR #22197, opened 2024-06-13
Gives blood brothers a box filled with random shitty traitor items

Moltijoe:
All blood brothers will spawn with a carboard box filled with items that total 5tc
no individual item will cost more than 2 tc
can only come from the categories of

- Stealth Gadgets
- Misc. Gadgets
- (Pointless) Badassery

Can give bloodbrothers some unusual items that are otherwise unobtainable
Lets items that are rarely if ever bought by a traitor have a chance to possibly shine

can always add a specific item blacklist if they end up being bad

# Testing
image
image

:cl:  
rscadd: Gives blood brothers a box filled with random shitty traitor items
/:cl:

### DIFF
--- a/code/modules/antagonists/brother/brother.dm
+++ b/code/modules/antagonists/brother/brother.dm
@@ -262,10 +262,7 @@
 		var/item = pick(uplink_items[category])
 		var/datum/uplink_item/I = uplink_items[category][item]
 
-		if(I.cost > item_value_cap)
-			uplink_items[category] -= uplink_items[category][item] //remove it so it can't keep getting picked despite being invalid
-			continue
-		if(I.cost > remaining_value)
+		if(I.cost > item_value_cap || I.cost > remaining_value)
 			uplink_items[category] -= uplink_items[category][item] //remove it so it can't keep getting picked despite being invalid
 			continue
 		remaining_value -= I.cost
--- a/code/modules/antagonists/brother/brother.dm
+++ b/code/modules/antagonists/brother/brother.dm
@@ -248,7 +248,7 @@
 	///maximum amount of tc any individual item can be
 	var/item_value_cap = 2
 	///list of categories that items are allowed to come from
-	var/list/allowed_categories = list("Stealth Gadgets", "Misc. Gadgets", "(Pointless) Badassery")
+	var/list/allowed_categories = list(UPLINK_CATEGORY_STEALTH_GADGETS, UPLINK_CATEGORY_MISC, UPLINK_CATEGORY_BADASS)
 
 /obj/item/storage/box/bloodbrother/PopulateContents()
 	var/list/uplink_items = get_uplink_items(null, FALSE)
--- a/code/modules/antagonists/brother/brother.dm
+++ b/code/modules/antagonists/brother/brother.dm
@@ -33,7 +33,7 @@
 	W.on_reading_finished(brother)
 	qdel(W)
 
-	var/T = new /obj/item/storage/box/bloodbrother()
+	var/obj/item/storage/box/bloodbrother/T = new()
 	if(brother.equip_to_slot(T, ITEM_SLOT_BACKPACK)) //except for here, where it will fail because there's no backpack slot to equip to
 		SEND_SIGNAL(brother.back, COMSIG_TRY_STORAGE_SHOW, brother)// which is fine, because it'll just not do anything and give a notice (probably)
 	else
--- a/code/modules/antagonists/brother/brother.dm
+++ b/code/modules/antagonists/brother/brother.dm
@@ -28,14 +28,14 @@
 	return ..()
 
 /datum/antagonist/brother/proc/equip_brother()
-	var/mob/living/carbon/human/brother = owner.current
+	var/mob/living/carbon/brother = owner.current
 	var/obj/item/book/granter/crafting_recipe/weapons/W = new
 	W.on_reading_finished(brother)
 	qdel(W)
 
 	if(istype(brother))
 		var/obj/item/storage/box/bloodbrother/T = new()
-		if(brother.equip_to_slot(T, ITEM_SLOT_BACKPACK))
+		if(brother.equip_to_slot_or_del(T, ITEM_SLOT_BACKPACK))
 			SEND_SIGNAL(brother.back, COMSIG_TRY_STORAGE_SHOW, brother)
 			return
 	
--- a/code/modules/antagonists/brother/brother.dm
+++ b/code/modules/antagonists/brother/brother.dm
@@ -28,16 +28,19 @@
 	return ..()
 
 /datum/antagonist/brother/proc/equip_brother()
-	var/mob/living/carbon/human/brother = owner.current //it's probably fine if we cast this directly to human, because the procs being called don't care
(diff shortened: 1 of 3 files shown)


code/modules/antagonists/brother/brother.dm
```diff
@@ -28,10 +28,17 @@
 	return ..()
 
 /datum/antagonist/brother/proc/equip_brother()
+	var/mob/living/carbon/human/brother = owner.current //it's probably fine if we cast this directly to human, because the procs being called don't care
 	var/obj/item/book/granter/crafting_recipe/weapons/W = new
-	W.on_reading_finished(owner.current)
+	W.on_reading_finished(brother)
 	qdel(W)
 
+	var/T = new /obj/item/storage/box/bloodbrother()
```
ynot01:
```suggestion
	var/obj/item/storage/box/bloodbrother/T = new()
```

code/modules/antagonists/brother/brother.dm
```diff
@@ -230,3 +237,30 @@
 
 /datum/team/brother_team/antag_listing_name()
 	return "[name] blood brothers"
+
+
+//box given to every bloodbrother
+/obj/item/storage/box/bloodbrother
+	name = "Keepsake box"
```
ynot01: lowercase

code/modules/antagonists/brother/brother.dm
```diff
@@ -230,3 +237,30 @@
...
+	desc = "A box full of unusual items found in the maintenance hallways."
+	///total tc cost that can be inside the box
+	var/total_box_value = 5
+	///maximum amount of tc any individual item can be
+	var/item_value_cap = 2
+	///list of categories that items are allowed to come from
+	var/list/allowed_categories = list("Stealth Gadgets", "Misc. Gadgets", "(Pointless) Badassery")
```
ynot01: if you're going to use them like this, make them into defines

code/modules/antagonists/brother/brother.dm
```diff
@@ -230,3 +237,30 @@
...
+
+/obj/item/storage/box/bloodbrother/PopulateContents()
+	var/list/uplink_items = get_uplink_items(null, FALSE)
+	var/remaining_value = total_box_value
+	while(remaining_value)
```
ynot01:
```suggestion
	while(remaining_value > 0)
```

code/modules/antagonists/brother/brother.dm
```diff
@@ -230,3 +237,30 @@
...
+		var/category = pick(allowed_categories)
+		var/item = pick(uplink_items[category])
+		var/datum/uplink_item/I = uplink_items[category][item]
+
+		if(I.cost > item_value_cap)
+			continue
+		if(I.cost > remaining_value)
```
ynot01: please initialize a list of "allowed items" and remove them as they get filtered in this way (pick_and_take), then throw `CRASH()` once the list is empty to prevent an infinite while loop

ynot01 (changes requested):

code/modules/antagonists/brother/brother.dm
Moltijoe: Rather than pick and take, im gonna do something different so it still allows for duplicates

neeshacark:
eyyyy

i forget why we removed this, it used to be a thing, not identical i think it was a limited set of items

code/modules/antagonists/brother/brother.dm
```diff
@@ -28,10 +28,17 @@
 	return ..()
 
 /datum/antagonist/brother/proc/equip_brother()
+	var/mob/living/carbon/human/brother = owner.current //it's probably fine if we cast this directly to human, because the procs being called don't care
 	var/obj/item/book/granter/crafting_recipe/weapons/W = new
-	W.on_reading_finished(owner.current)
+	W.on_reading_finished(brother)
 	qdel(W)
```
ynot01, line 35:
```suggestion
	if(!istype(brother))
		return
```

code/modules/antagonists/brother/brother.dm
```diff
@@ -28,10 +28,17 @@
...
 
+	var/obj/item/storage/box/bloodbrother/T = new()
+	if(brother.equip_to_slot(T, ITEM_SLOT_BACKPACK)) //except for here, where it will fail because there's no backpack slot to equip to
```
ynot01:
```suggestion
	if(brother.equip_to_slot_or_del(T, ITEM_SLOT_BACKPACK)) // runtimes without carbon typecheck
```

code/modules/antagonists/brother/brother.dm
```diff
@@ -28,10 +28,17 @@
 	return ..()
 
 /datum/antagonist/brother/proc/equip_brother()
+	var/mob/living/carbon/human/brother = owner.current //it's probably fine if we cast this directly to human, because the procs being called don't care
```
ynot01:
```suggestion
	var/mob/living/carbon/brother = owner.current
```

code/modules/antagonists/brother/brother.dm
```diff
@@ -28,10 +28,17 @@
...
 	var/obj/item/book/granter/crafting_recipe/weapons/W = new
-	W.on_reading_finished(owner.current)
+	W.on_reading_finished(brother)
 	qdel(W)
 
+	var/obj/item/storage/box/bloodbrother/T = new()
+	if(brother.equip_to_slot(T, ITEM_SLOT_BACKPACK)) //except for here, where it will fail because there's no backpack slot to equip to
+		SEND_SIGNAL(brother.back, COMSIG_TRY_STORAGE_SHOW, brother)// which is fine, because it'll just not do anything and give a notice (probably)
```
ynot01: runtimes should be handled properly because they cause lag and in the worst cases, unpredictable behavior & catastrophic failure
```suggestion
		SEND_SIGNAL(brother.back, COMSIG_TRY_STORAGE_SHOW, brother)
```

ynot01 (changes requested):

code/modules/antagonists/brother/brother.dm
ynot01: tabbing on this one might be incorrect due to webedit

code/modules/antagonists/brother/brother.dm
```diff
@@ -230,3 +240,36 @@
...
+	///maximum amount of tc any individual item can be
+	var/item_value_cap = 2
+	///list of categories that items are allowed to come from
+	var/list/allowed_categories = list(UPLINK_CATEGORY_STEALTH_GADGETS, UPLINK_CATEGORY_MISC, UPLINK_CATEGORY_BADASS)
+
+/obj/item/storage/box/bloodbrother/PopulateContents()
+	var/list/uplink_items = get_uplink_items(null, FALSE)
+	var/remaining_value = total_box_value
+
+	for(var/i = 50; i > 0; i--) //only iterate 50 times, so it doesn't get stuck in an infinite loop
```
ynot01, line 260: for the record
image

ynot01 (approved):

code/modules/antagonists/brother/brother.dm
Moltijoe: True

PR #22212, opened 2024-06-18
Fixes jungleland mobs dying to toxic water + deep toxic not being stronger

Moltijoe:
I just copied from old code, apparently it never checked for WEATHER_ACID immunity
but it also never checked anything other than a human
so they were immune, not because of their immunity, but because they weren't even being looked at in the first place

this also meant that if a human got acid weather immunity, they'd still be affected anyway

also, i removed an unintentional part of deep toxic water being significantly more dangerous

# Testing
image

:cl:  
bugfix: Fixes jungleland mobs not being immune to toxic water
bugfix: Fixes deep toxic water not being more dangerous
bugfix: Fixes jungleland plant people not being immune to toxic water like intended
bugfix: Fixes objects being unaffected by the toxic water acid (cardboard boxes)
/:cl:


AMyriad:
thank god finally
image

ToasterBiome (approved):

PR #22219, opened 2024-06-19
Selling humans as pirates now ransoms them rather than round removing them

Moltijoe:
it functions almost exactly the same as contractor ransoming (because i copy pasted it)
at some point it the future it might be better to merge them together into one system (knowing me i'll draft this pr in a few days and do that instead)

round removal is cringe

# Testing
image
image
image

:cl:  
tweak: Selling humans as pirates now ransoms them rather than round removing them
/:cl:

### DIFF
--- a/code/modules/cargo/exports.dm
+++ b/code/modules/cargo/exports.dm
@@ -115,7 +115,7 @@ Credit dupes that require a lot of manual work shouldn't be removed, unless they
 					we thank you for providing them. Your value is expended, and you will be ransomed back to your station. We always get paid, \
 					so it's only a matter of time before we ship you back...\"</i></span>")
 
-		var/ransom = 3000 //people are pretty cheap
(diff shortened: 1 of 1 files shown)


code/modules/cargo/exports.dm
```diff
@@ -54,12 +54,127 @@ Credit dupes that require a lot of manual work shouldn't be removed, unless they
...
+				handleTrafficking(thing)
+			else
+				qdel(thing)
 
 	return report
 
+
+// They're off to holding - handle the return timer and give some text about what's going on.
+/proc/handleTrafficking(mob/living/M)
+	//get rid of them for the time being
```
ynot01, line 68: you're holding the calling proc hostage for 25 seconds
```suggestion
	set waitfor = FALSE
	//get rid of them for the time being
```

ynot01 (changes requested):

ynot01 (approved):

PR #22239, opened 2024-06-27
Unathi blood color for lizards

ThatGuyPhil:
Unathi blood is so much better than sarathi blood.

Changed the color for Type L blood in blood_types.dm, resulting in Vuulek now having blood similar to the Unathi species seen in other servers. Was thinking that only invertebrates have green blood, so it would make sense if a lizard (vertebrate) had something a bit closer to red, while also having a distinct blood color from humans. Unathi blood color would be a more fitting than Sarathi (shiptest lizardpeople) blood color for this.

It replaces an ugly pale green blood with ugly pale orange blood. 

# Testing

image
(everyone was harmed in the making of this)

# Changelog

:cl:    
tweak: Changed the Vuulek blood color to #f24b2e 
/:cl:

### DIFF
--- a/code/datums/blood_types.dm
+++ b/code/datums/blood_types.dm
@@ -57,7 +57,7 @@
 ////////////////////////////////////////////////////////////////
 /datum/blood_type/lizard
 	name = "L"
-	color = LIGHT_COLOR_BLUEGREEN
+	color = "#f24b2e"
 	compatible_types = list(/datum/blood_type/lizard)
 
 


ju45he:
what if we had specially made gibs for lizardpeople?

ThatGuyPhil:
> what if we had specially made gibs for lizardpeople?

all species use the same gibs though. The color depends on the blood color of the species, so humans still got red gibs if thats what youre worried about.

Mqiib:
Honestly this makes sense.

AMyriad:
This just looks like regular blood, what color was it before?

Moltijoe:
Agreed, this just looks like regular blood

ThatGuyPhil:
Human blood is red. If you bleed vodka sauce thats concerning.

ThatGuyPhil:
image
here's a little comparison between human blood and the proposed vuleek blood.

now all i can think about is pasta sauce... i could go for some penne alla vodka rn.

sorry for the overlap i kinda just gibbed a human and a lizard right next to eachother

Moltijoe:
I can see the difference, and honestly its still boring
When you can use literally any colour, why settle for a slightly different shade of red

ThatGuyPhil:
> I can see the difference, and honestly its still boring When you can use literally any colour, why settle for a slightly different shade of red

In my mind at least it makes a little bit more sense than green, and you might as well call any color boring. Plus there are already a handful of nonhumans with their own blood colors to satisfy your non red or orange blood needs. Polysmorph and Ethereals also have green and yellow blood so I thought that the green lizard blood just looked too similar to them too. Blue is also taken by Vox, which is also the same color that their blood is in baystation (the server that they're originally from I guess).

I got bored and decided to have all the blood types next to eachother for a full comparison.
image

I guess the only color we wouldn't have yet would be purple. I was thinking if I made a PR to give moths and flies their own white or clear looking hemolymph blood but I think I'm getting a bit ahead of my self for that. 

TL;DR I don't really consider the lizardpeople to be 'alien' enough compared to the other aliens to have an extremely different blood color than humans.

Moltijoe (approved):
yeah sure, why not, it got approval by lizards so its good i guess

PR #22242, opened 2024-06-28
Synth project minor rework (sponsored by DownloadMoreRAM.com)

Moltijoe:
Heat and cold protection no longer have a permanent suspicion of 30
all synth projects now require 1 ram to stay active
adds a new project that gives 2 more ram when completed, but takes twice as long to research as the other projects

allows for more depth when it comes to synth project tradeoffs
so adding more projects isn't just a case of "more power"
using permanent suspicion is dangerous because it can allow you to instantly kill yourself with a couple clicks
ai already have ram for this purpose, why not synths

# Testing
image
image
image
i fixed it to be 1/1 after the screenshot
image

:cl:  
rscadd: Adds synth project to download more ram
rscadd: Synths now have 1 ram by default
tweak: All synth research projects now require 1 ram to maintain
tweak: heat and cold protection no longer give innate suspicion
bugfix: fixes a runtime caused by swapping synth shells while a program runs 
/:cl:


code/modules/mob/living/silicon/ai/decentralized/projects/synths/download_ram.dm
```diff
@@ -0,0 +1,9 @@
+/datum/ai_project/synth_project/download_ram
+	name = "Download more RAM"
+	description = "Download more ram from the ntnet. It works just as well as you think it would."
```
Bibby0110: RAM, NTNet

Bibby0110 (approved):
looks good

PR #22265, opened 2024-07-01
Enables jungleland's acid rain and tweaks it to inject toxic metabolites rather than apply acid

Moltijoe:
It was initially added alongside jungleland but deemed too dangerous as actual acid
jungleland feels lacking in features, adding acid rain both lets people experience what they were missing
as well as making it slightly more dangerous to freely explore
which should have the side effect of slowing down the speed at which jungleland is fully explored

it will last longer, but it'll be significantly less dangerous than ash storms as a result

also refactors water immunity into using bitflags rather than lists

Adds more to jungleland
also i love the aesthetic of rainfall

# Testing
image

:cl:  
rscadd: Enables jungleland's acid rain
tweak: jungleland acid rain only injects toxic metabolites rather than apply acid
bugfix: jungleland mobs now have the proper weather immunities
/:cl:


Aquizit (approved):
Map change fine, just makes the thing completely immune instead of just to lavaland weather.

ToasterBiome (approved):

PR #22273, opened 2024-07-02
Tweaks toolset implants drop and menu opening keys

Moltijoe:
Dropping a toolset item now retracts it rather than bringing up the menu to swap
Attempting to activate throw mode with a toolset item now brings up the menu to swap

drop is drop, drop should always be drop
in-hand use can't open the menu because of welding tool
right click can't be used because a number of construction items need to right click on an object to do things
the throw mode button is ideal because you can't actually throw toolset items

# Testing

:cl:  
tweak: Dropping a toolset item now retracts it rather than bringing up the menu to swap
tweak: Attempting to activate throw mode with a toolset item now brings up the menu to swap
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals/signals_mob/signals_carbon.dm
+++ b/code/__DEFINES/dcs/signals/signals_mob/signals_carbon.dm
@@ -123,3 +123,8 @@
 
 /// from  base of mob/living/carbon/set_species(): (new_race)
 #define COMSIG_CARBON_SPECIESCHANGE "mob_carbon_specieschange"
+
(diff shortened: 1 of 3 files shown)


Scrambledeggs00:
So it will try to swap tools when you're deconstructing a table?

Moltijoe:
> So it will try to swap tools when you're deconstructing a table?

ah fuck, god damnit

Moltijoe:
there's no alternative for Q

Moltijoe:
we're so back

ynot01 (approved):

PR #22284, opened 2024-07-03
Prevents being able to bag of holding bomb from a range

Moltijoe:
if you just left the popup open and swapped bodies, you could click yes whenever you felt like it

# Testing
image
image

:cl:  
bugfix: Prevents being able to bag of holding bomb from a range using other bodies
/:cl:


ynot01 (approved):
works

PR #22285, opened 2024-07-03
Update Yog's Repo rust_g.dll to the most recent version compiled with --all-features

Chubbygummibear:

jungleland was failing to generate because it was using rustg functions that weren't default, and while i'm pretty sure the main server compiles rustg during every startup with all features, the local repo dll just has the default. therefore jungleland works on live but breaks when trying to work with it locally

being able to test locally means being able to work on code

# Testing
dreamseeker_Snx6xDBG9q

# Changelog

:cl:  
rscadd: update rust_g.dll so jungleland can be loaded locally
/:cl:

### DIFF
--- a/dependencies.sh
+++ b/dependencies.sh
@@ -8,7 +8,7 @@ export BYOND_MAJOR=515
 export BYOND_MINOR=1633
 
 #rust_g git tag
-export RUST_G_VERSION=1.2.0-yogs2
+export RUST_G_VERSION=1.2.0-yogs2.1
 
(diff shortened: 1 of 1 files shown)


Moltijoe:
he means https://github.com/yogstation13/rust-g/pull/5

adamsong:
Just build rust_g.dll with `--all-features` and just update that, no need to update the build version if you aren't actually changing rust_g

Moltijoe (approved):

PR #22287, opened 2024-07-04
bodycams now shown when worn

ktlwjec0:
We had sprites for miner and sec bodycams when worn, but they didn't work. This fixes that.

# Testing
image
image
No new sprites.

:cl:  ktlwjec
imageadd: Bodycams show on-person, no need to examine to see if someone is wearing one.
/:cl:


ShoziOffical:
image
We should make bodycam attachable to clothes without taking the neck slot

UselessTheremin:
> image
> We should make bodycam attachable to clothes without taking the neck slot

generally unsure how i feel about that, as that qould become annoying when trying to equip the craftable talisman, i think the neck slot is fine tbh

ynot01 (approved):

PR #22294, opened 2024-07-05
Reduces jungleland ore spawns + plasmacutter interactions + bluespace bugfix

Moltijoe:
With how ore spawns, it was easier to add a new "ore type" that's just the lack of ore

plasmacutters have a mechanic where if they're mining an ore, they'll travel extra tiles based on plasmacutter type
plasmacutters will now benefit from half that extra tile bonus if they're being used on jungleland

jungleland has issues, this seeks to solve some of those issues

# Testing
probably need to reduce it more, but see how it's not a solid square
image

:cl:  
tweak: plasmacutters now go slightly further when mining on jungleland
tweak: reduced jungleland ore spawns
bugfix: fixed a typo that prevented bluespace from spawning on jungleland
bugfix: fixed the plasmacutter ore upgrade not working on jungleland
/:cl:


LassiH:
huh i always tought the reason no bluespace spawned was cuz of the extra amounts of ore

Moltijoe:
> huh i always tought the reason no bluespace spawned was cuz of the extra amounts of ore

 nope, a classic bluesapce moment

ynot01 (approved):

PR #22297, opened 2024-07-05
Tweaks jungleland in so many ways

Moltijoe:
Removes Barren rocks biome
Changes floor tiles of dying forest and dried swamp

it incidentally removed the possible ashdrake spawn chance, but that shouldn't matter for the time being

Barren rocks biome just makes it more like lavaland which sorta defeats the purpose of having a different mine type
Shifting floor tiles around as it did will make it easier to eventually add smoothing to toxic water pits

# Testing
image
image

:cl:  
tweak: Tweaks jungleland in a number of ways
/:cl:


neeshacark (approved):
maps fine dont @ me here either

ToasterBiome (approved):

PR #22304, opened 2024-07-06
Extended shocktouch doesn't force the target to drop items anymore

Moltijoe:
extended shock touch just instantly disarming a target with a single click is kinda silly powerful for a civilian tool
better hope you aren't relying on an item as an antag

# Testing
shouldn't need to, but i can

:cl:  
tweak: Extended shocktouch doesn't force the target to drop items anymore
/:cl:


Tenn0x:
imo it should still have a chance to disarm, just not guaranteed.

Moltijoe:
> imo it should still have a chance to disarm, just not guaranteed.

i despise rng, no

Tenn0x:
dont think it will be used at all with no disarm. Maybe the damage alone will still make it worth as it still gives confusion for a time

Mqiib:
i dunno 15 free ranged burn damage with no downside is pretty good

Beeper-Skeeper:
geneticists when they can no longer disarm an armed traitor with funny electricity that they can get literally every shift:

PR #22315, opened 2024-07-07
Jellypeople now take cloning damage from water + regen

Moltijoe:
they already take clone damage being in pools, so it makes sense for them to take clone damage from being wet regularly
the flavour text of water damage already mentioned cell damage, so it makes sense
to balance out getting clone damage so often, they can regen it slowly if they're not fucked up in any way
it's thematically fitting for slime based people

# Testing
image

:cl:  
tweak: Jellypeople continue to take water damage even when dead
tweak: Jellypeople take clone damage from being wet instead of stamina + burn
tweak: Jellypeople very slowly regenerate clone damage when at high blood levels and alive
/:cl:


neeshacark:
ive always thought the clone damage from pools was silly, iirc it was because slimes took damage from fire extinguishers which have water in them - but really it was meant to be because of the cold that fire extinguishers would logically put out rather than the water. someone saw this and added it to the pool.

i just think its kinda goofy

Moltijoe:
> ive always thought the clone damage from pools was silly, iirc it was because slimes took damage from fire extinguishers which have water in them - but really it was meant to be because of the cold that fire extinguishers would logically put out rather than the water. someone saw this and added it to the pool.
> 
> i just think its kinda goofy

Slimes dissolving in large bodies of water is pretty common too

Chubbygummibear (approved):
i do as i'm told

PR #22318, opened 2024-07-08
A lot of things can now be done while lying down

Moltijoe:
also renamed a define that was killing me

Wanna look at family guy funny moment compilations on my pda while lying down in my bed

jokes aside, there's a number of things that stop being usable just when prone which don't entirely make sense
handheld electronics
climbing into a mech
removing the cap from a spray can
setting the duration of a desynchronizer
resetting a geiger counter
removing a specific number of items from a stack
toggle the welding screen of a hardhat

the list goes on

while this will allow some things to be used that don't entirely make sense, it's more fun to be able to use something when you shouldn't compared to not being able to use something when you should

do note, this change won't extend to people that are restrained or unconscious

# Testing
image

:cl:  
tweak: A lot of things can now be done while lying down
/:cl:


SapphicOverload (approved):

ToasterBiome:
thinking about this, flipping welding goggles and climbing into mechs are two important balance changes

Moltijoe:
> thinking about this, flipping welding goggles and climbing into mechs are two important balance changes

if someone's either handcuffed or stunned from being stamcrit, they'll not be able to do it

iamthesilly:
> thinking about this, flipping welding goggles and climbing into mechs are two important balance changes

You can already flip down welding helmets and masks when lying down so I don't think goggles will make much of a difference. Also like Molti said, you can't do either when stunned or restrained so I don't think it will be significant for balance.

ToasterBiome (approved):

SapphicOverload (approved):

PR #22345, opened 2024-07-12
Tweaks the BFG mech weapon

Moltijoe:
for a BFG, it doesn't really fell good to use or play against
having only 5 ammo means it needs to be constantly refilled, making it feel tedious and unfun to use
it's grazing being able to hit multiple times means it's super inconsistent with how much damage it can do, meaning it can range between doing nothing, and one-shotting, entirely depending on how you're walking
it also provides no visual or audio feedback as to what's happening, adding confusion to the situation

# Testing
gotta

:cl:  
rscdel: Removed BFG mech weapon ammo
bugfix: BFG graze damage no longer ignores armour
tweak: BFG does slightly more graze damage
tweak: BFG can't graze the same target multiple times
tweak: BFG is now an energy weapon rather than ballistic
tweak: BFG graze now has a sound effect
tweak: BFG projectile is now a fire hazard
/:cl:

### DIFF
--- a/code/game/mecha/equipment/weapons/weapons.dm
+++ b/code/game/mecha/equipment/weapons/weapons.dm
@@ -222,6 +222,15 @@
 	projectile = /obj/projectile/energy/electrode
 	fire_sound = 'sound/weapons/taser.ogg'
 
+/obj/item/mecha_parts/mecha_equipment/weapon/energy/bfg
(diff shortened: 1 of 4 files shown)


Moltijoe:
not really satisfied with this, feels too much like the solaris
i'll come back to it

PR #22355, opened 2024-07-12
Kudzu now mutates from magnus purpura and lavaland cactus instead of weeds

Moltijoe:
kudzu is too impactful for how easy it is to get
adds interaction between mining and botany (i guess? why would mining provide kudzu for botany)
removing the mutation outright is too boring as it'll effectively remove kudzu interactions with botany, putting it behind an obscure mutation means if someone wants it, they can focus it and just need to either get supplied by mining or go planetside themselves

# Testing
gotta (but realistically, this is just tweaking lists, so unless there's a typo there shouldn't be any bugs

:cl:  
tweak: Kudzu now mutates from magnus purpura and lavaland cactus instead of weeds
/:cl:

### DIFF
--- a/code/modules/hydroponics/hydroponics.dm
+++ b/code/modules/hydroponics/hydroponics.dm
@@ -403,7 +403,7 @@
 		if(myseed)
 			qdel(myseed)
 			myseed = null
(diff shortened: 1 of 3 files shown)


StupidTums:
This works, but I still think a rework is needed for this antag, in general (I'd do it, but I'm too lazy :( ).

ToasterBiome (approved):

Beeper-Skeeper:
this
this is a compromise I like

PR #22359, opened 2024-07-13
Moves paramedic clothes from the locker into the medidrobe

Moltijoe:
because this is a little ridiculous
image

# Testing
no need with how simple this change is, but i can prove it if need be

:cl:  
bugfix: fixed a bugged suit being given to paramedics
tweak: Moves paramedic clothes from the locker into the medidrobe
/:cl:


ktlwjec0:
I will upvote, but I'd personally just up the count of these in the medidrobe, since the dufflebags are really only used when they don't have access to a vendor. Like the hop or hos.

Moltijoe:
> I will upvote, but I'd personally just up the count of these in the medidrobe, since the dufflebags are really only used when they don't have access to a vendor. Like the hop or hos.

Ah right

neeshacark (approved):
no issues with maps, this still needs a stinky coder maint review

code/modules/vending/wardrobes.dm
```diff
@@ -79,26 +79,30 @@
 	vend_reply = "Thank you for using the MediDrobe!"
 	products = list(/obj/item/clothing/accessory/pocketprotector = 4,
 					/obj/item/storage/backpack/duffelbag/med = 4,
-					/obj/item/storage/backpack/medic = 4,
```
ynot01, line 82: why?

ynot01 (approved):
yesd

PR #22363, opened 2024-07-13
cleans up containers dmi

ktlwjec0:
Cleans up containers dmi by moving it all into one file. It can get very confusing which dmi to place icons into, especially when they share a lot of icons. They also contained a lot of old, unused sprites which I don't see why we need to keep in there.

# Testing
I haven't found any issues, let me know if any are found.
image


Mqiib (approved):
delet old unused sprites good

Moltijoe (approved):

PR #22371, opened 2024-07-14
Slimes can no longer paralyze for up to 26 seconds with a single click

Moltijoe:
slimes shouldn't be able to game-end someone with a single click if they're fed enough

# Testing
don't need to

:cl:  
tweak: Slimes can no longer paralyze for up to 26 seconds with a single click
tweak: Slimes now have a chance to daze targets when glomping them
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon_defense.dm
+++ b/code/modules/mob/living/carbon/carbon_defense.dm
@@ -318,24 +318,19 @@
 
 
 /mob/living/carbon/attack_slime(mob/living/simple_animal/slime/M)
-	if(..()) //successful slime attack
+	. = ..()
+	if(.) //successful slime attack
 		if(M.powerlevel > 0)
-			var/stunprob = M.powerlevel * 7 + 10  // 17 at level 1, 80 at level 10
-			if(prob(stunprob))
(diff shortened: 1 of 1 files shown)


neeshacark:
it might be a good idea to remove/change anomaly slime to something else with this change. this is the only reason they were even remotely effective against players.

not saying its a bad change, i don't like stuns. i just think this will lead to either

a) anomaly slimes dying without doing anything

b) encourage anomaly slimes to target people who are new at the game and don't know how to defend themselves properly. this is bad

Moltijoe:
> it might be a good idea to remove/change anomaly slime to something else with this change. this is the only reason they were even remotely effective against players.
> 
> not saying its a bad change, i don't like stuns. i just think this will lead to either
> 
> a) anomaly slimes dying without doing anything
> 
> b) encourage anomaly slimes to target people who are new at the game and don't know how to defend themselves properly. this is bad

that's fair, i've decided to tweak the entire mechanic, change the probability, and swap it over to a daze from a paralyze

ToasterBiome (approved):

CrazyDrakkon:
Does energy level matter still?

PR #22401, opened 2024-07-17
Speeds up crafting menu food creation

Moltijoe:
Unlike other forms of cooking, this is just sitting around waiting for a progress bar to fill
they've also already finished the "mise en place" so it's just assembly at that point

# Testing
number change

:cl:  
tweak: Food made via the crafting menu now takes only half the time
/:cl:


ktlwjec0:
Some pastry recipes need their times removing since they're manually set to 1.5 seconds there if you do a quick search in the file.

Moltijoe:
> Some pastry recipes need their times removing since they're manually set to 1.5 seconds there if you do a quick search in the file.

oh really? i did a brief search but i didn't notice it

ToasterBiome (approved):

PR #22419, opened 2024-07-21
Reduces nightmare antag spawn minpop to 15 from 30

Moltijoe:
nightmare is mostly harmless, just a silly dude with an armblade that wants to make the station dark
it fills in for kudzu being effectively removed from lowpop
also, nightmares are basically darkspawns, so if they somehow both spawn during the same round, them being able to work together makes sense

# Testing
shouldn't need to if it passes linter

:cl:  
tweak: Reduces nightmare antag spawn minpop to 15 from 30
tweak: Nightmares are now considered on the same team as darkspawns, despite not seeing their objectives
/:cl:


Bester22:
Uh hey, you do realise if someone grabs this role,  that is competent in combat they will just stomp a 15pop server right? nightmare has pretty good healing when in dark. a quite robust melee weapon that can and will destroy the average crew member oh and it has no murderbone restrictions.

UselessTheremin:
yeah nightmare would absolutely destroy a 15 pop station, especially since at that low of pop there often times is not a lot of security, idk how i feel about this

Moltijoe:
something to note, nightmare likely hasn't spawned since before i changed it
but the light eater does far less damage than it used to do

it's 18 force with no ap instead of 25 force with 35 ap

Moltijoe:
maybe, maybe not, i'll come back to this

PR #22441, opened 2024-07-25
Reduces main engine power production

SapphicOverload:

Reduces the power output of the supermatter, singularity, tesla, and nuclear reactor. Also makes the tesla engine's power output more consistent.

Currently there's no reason to optimize power production because the main engine's default setup makes far more power than the station will ever need, there's no reason to use more interesting engine setups to increase production or even to set up the other power production sources like the solars or turbine. A year or two ago engine power production was much lower and a lot of players came up with many different ways of optimizing engines, but now that there's no need to do so it's rarely done.

This rebalances the power production to be not too far above the station's consumption, incentivizing engineers to come up with interesting ways of increasing the engine's output if demand gets too high.

# Testing
image

# Changelog

:cl:  
tweak: Reduced main engine power output.
bugfix: Fixed the tesla engine's power production being inconsistent.
/:cl:


wonderinghost:
finally a nerf to rad collectors

Moltijoe (approved):

PR #22473, opened 2024-07-31
Breaching missiles now explode on windows too

Mqiib:
Plasma glass could tank breaching missiles all day :)

Adds windows to the list of sturdy objects for breaching missiles to breach, since they can be very sturdy.

No building a glass castle

# Testing
will

# Wiki Documentation

explodes on windows

# Changelog

:cl:  

tweak: Breaching missiles now explode on windows rather than bouncing off

/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/special/rocket.dm
+++ b/code/modules/projectiles/projectile/special/rocket.dm
@@ -54,7 +54,8 @@
 	/turf/closed,
 	/obj/mecha,
 	/obj/machinery/door/,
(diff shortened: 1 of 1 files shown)


SapphicOverload (commented):
Instead of having a list why not check for uses_integrity variable instead? It's TRUE on mechs, walls, doors, structures, but not mobs. Catches any other edge cases that might come up.

ToasterBiome (approved):

PR #22476, opened 2024-08-02
Adds regular rain to jungleland

Moltijoe:
also makes it possible to use a raincoat or something to mitigate the water part of rain

lavaland has gentle ashfalls as a random chance instead of an ashstorm
adding regular rain does the same for jungleland
this adds a random chance for the storm to be a "relief" rather than something to panick over

# Testing
image

# Spriting
made the acid rain sprite grayscale so it could be coloured blue

:cl:  
rscadd: Added regular rain to jungleland
tweak: permeability now affects how many wet stacks players get from rain
tweak: rain and acid rain can now extinguish people if they're on fire
imageadd: added grayscale rain image
/:cl:


AMyriad:
Love this, but I think ordering it the same way they do in ash_storm.dm would be cleaner, less changing everything when the harmless version is a subtype of the reg version no?

Moltijoe:
> Love this, but I think ordering it the same way they do in ash_storm.dm would be cleaner, less changing everything when the harmless version is a subtype of the reg version no?

I initially did it that way, bur changed it to this because it'll make adding different types of rain easier. Blood rain, chocolate rain, you name it

AMyriad:
> I initially did it that way, bur changed it to this because it'll make adding different types of rain easier. Blood rain, chocolate rain, you name it

mmmmm chocolateland...

SapphicOverload (approved):
https://github.com/user-attachments/assets/5551cb31-92d3-42b7-985c-f43ba518f416

ToasterBiome (approved):

PR #22479, opened 2024-08-03
Refactors exosuit weather protection

SapphicOverload:
Refactors exosuit weather protection to protect the pilot from specific weather types or all/none of them, also fixes exosuits not protecting against certain weathers they really should such as snowstorms.

# Testing
Soon

:cl:  
bugfix: Fixed mechs not protecting the pilot from certain weather types
/:cl:

### DIFF
--- a/code/game/mecha/working/ripley.dm
+++ b/code/game/mecha/working/ripley.dm
@@ -14,6 +14,7 @@
 	var/list/cargo = new
 	var/cargo_capacity = 15
 	var/hides = 0
+	weather_protection = NONE
 	enclosed = FALSE //Normal ripley has an open cockpit design
 	enter_delay = 10 //can enter in a quarter of the time of other mechs
 	exit_delay = 10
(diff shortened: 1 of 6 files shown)


code/game/mecha/working/ripley.dm
```diff
@@ -72,6 +73,7 @@
 	slow_pressure_step_in = 3 //step_in while in normal pressure conditions
 	step_in = 3
 	resistance_flags = LAVA_PROOF | FIRE_PROOF | ACID_PROOF
+	weather_protection = WEATHER_LAVA|WEATHER_STORM
```
Moltijoe, line 76: are you actually checking mech's weather_protection for weather_lava on lava itself?

iamthesilly: It's for the floor is lava weather type which isn't actually lava

Moltijoe: Ah, right

githubuser4141:
Shouldn't death squad mechs that are proofed against real lava also be proofed against much weaker floor lava? Also this makes the mk1 Ripley much more useless for mining, while logically it could have a lightweight "plastic/rubber bag" that isn't airtight and won't stop bullets but gives weather protection

Moltijoe (approved):

PR #22089, opened 2024-05-22
Closing the gap between station & lavaland

ynot01:
# Document the changes in your pull request
Mining shuttles now take 1 second to depart and 1 second to arrive

# Why is this good for the game?
Optimally, station & planet z-level would be one in the same, but this is a step closer. By making transit incredibly fast, planetside becomes a lot less exclusive. Inspired by IceMeta's ladder

# Testing
It goes fast

# Wiki Documentation
See documentation

# Changelog

:cl:  
tweak: The mining shuttle engines have had their safeties disabled, making interplanetary travel faster than ever before.
/:cl:

PR #22489, opened 2024-08-06
Revert "Closing the gap between station & lavaland"

LazennG:
Reverts yogstation13/Yogstation#22089

Not everything needs to be done as quick as possible. With the total time on the mining shuttle per ship being 3 seconds, there's no reason to make small talk while waiting for a third person to get on the ship. There's no time to get into place incase you're getting ready to jump somebody on the way to the landing bay, and there's even less time for miners to talk to each other since they're only gonna be in close proximity for the time it takes to sneeze. If someone is caught waiting for another to get onto the ship, the thought of the guy who wasn't waiting isn't gonna be "oh they waited for me how nice," but "why did they wait when the shuttle teleports are they dumb"

### DIFF
--- a/_maps/shuttles/mining_box.dmm
+++ b/_maps/shuttles/mining_box.dmm
@@ -37,9 +37,7 @@
 	shuttle_id = "mining";
 	name = "mining shuttle";
 	port_direction = 4;
-	width = 7;
-	ignitionTime = 10;
-	callTime = 10
+	width = 7
 	},
 /obj/machinery/door/firedoor/border_only{
 	dir = 4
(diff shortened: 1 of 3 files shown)


ynot01 (approved):

ynot01:
i will work on another approach that does not disrupt shuttle transit RP

wonderinghost:
the 5 seconds of rp

UselessTheremin:
> i will work on another approach that does not disrupt shuttle transit RP

what other approach? i dont think there needs to be another approach

Scrambledeggs00:
It taking awhile for the shuttle to go to/from lavaland and station is good because it gives antags plenty of time to kill whoever is with them on the shuttle. I've had it happen to me where someone lured me onto it and killed me while in transit.

AMyriad (approved):
mapping skeleton seal of approval

Moltijoe (approved):

PR #22504, opened 2024-08-10
Minor rework to how toxic chemicals work

Moltijoe:
Toxic chemicals now work by
Put the toxpower of every chemical in a list in descending order
first chemical (highest) does full damage
second chemical does 50% reduced damage
third chemical does 75% reduced damage
fourth chemical does 87.5% reduced damage

and so forth and so on

i hate the arbitrary purge threshold, it makes certain toxins that are designed to be potent in small amounts utterly useless, as they get instantly purged when low
Changing it to this allows for something like cyanide to actually be good in small doses
while still preventing loading a syringe with 1u of 30 different toxins to instantly kill someone

Gave polys a toxmod to account for their livers being nerfed a bit

# Testing
image

:cl:  
tweak: Polysmorph now have a 0.8 toxmod
tweak: Livers no longer instantly purge toxic chemicals when below an arbitrary number
tweak: Toxic chemicals now deal half damage for every stronger toxin in the bloodstream
tweak: Toxic chemicals now no longer scale liver damage with volume
/:cl:


code/modules/surgery/organs/liver.dm
```diff
@@ -12,38 +11,40 @@
...
-							provide_pain_message = T.silent_toxin ? HAS_SILENT_TOXIN : HAS_PAINFUL_TOXIN
+			var/provide_pain_message = FALSE
+			var/damage_multiplier = 1 //reduced for every toxin
+			var/list/toxpwr = list()
+
+			for(var/datum/reagent/toxin/T in C.reagents.reagent_list)
+				toxpwr += T.toxpwr
+				provide_pain_message = provide_pain_message || !T.silent_toxin
+					
+			sort_list(toxpwr,  cmp = /proc/cmp_numeric_dsc) //sort all toxpowers scaling from highest to lowest
```
iamthesilly: I feel like this should probably be cached on the liver and updated whenever a toxin is added or removed from the body so you don't have to sort the list every single tick.

Moltijoe: guh, i considered that but it sounds like a lot of work

Moltijoe:
i really should implement sarah's suggestion, but i don't have the motivation, and haven't for a while

JamieD1:
How we looking @Moltijoe

Moltijoe:
> How we looking @Moltijoe

this was one of the prs i had actually fixed and tested

ToasterBiome (approved):

PR #22521, opened 2024-08-13
[BOUNTY] Command positions now require experience from specific jobs in addition to a blanket exp requirement

Moltijoe:
https://forums.yogstation.net/threads/30-change-time-requirements-system-to-make-heads-of-staff-including-captains-require-further-experience-in-their-department-ss13-server.22824/

i personally don't care for this, but a bounty is a bounty
i think the idea is that command is supposed to have a more diverse knowledge over what they're doing, but i think that's overly restrictive, shitty underqualified bosses is a part of life
it REALLY fucks over captain too, as it'll make them require 3 hours on two different other command positions first

# Testing
gotta, gonna wait until it actually might get merged

:cl:
tweak: Command positions now require experience from specific jobs in addition to a blanket exp requirement
/:cl:


AMyriad:
hilarious since this will probably bar me from playing something like CMO cause I've barely played virologist

Moltijoe:
> hilarious since this will probably bar me from playing something like CMO cause I've barely played virologist

you don't need to play all of them, just enough
it's 3 hours of 3 different medical jobs

Moltijoe:
i might come back to this

PR #22522, opened 2024-08-13
Buffs radscrub

Moltijoe:
Radiation is pretty difficult to combat, radscrub was coded specifically to deal with the aftermath of radiation, and prevent/reduce future radiation
It rarely if ever gets made because it simply wasn't strong enough

Also makes it's mob radiation preventative measures good enough to actually be worth using, it already caused toxin damage over time, might as well make the upside actually worth it

# Testing
gotta

:cl:
tweak: Radscrub now provides slightly more rad insulation
tweak: Radscrub now also scrubs items inside of others (think backpacks)
tweak: Radscrub now provides temporary radiation immunity to mobs that consume it
/:cl:


Moltijoe:
not really feeling it

PR #22531, opened 2024-08-14
Fixes bug that allowed non-carbons to strip carbons

ynot01:

Added carbon check & cyborg check to ui_act

Non carbons can still see the menu, but not interact with it

# Changelog

:cl:
bugfix: Fixed non-carbons stripping humans
/:cl:

### DIFF
--- a/code/datums/elements/strippable.dm
+++ b/code/datums/elements/strippable.dm
@@ -368,7 +368,7 @@
 	if (.)
 		return
 
-	if (!iscarbon(usr))
+	if (!iscarbon(usr) && !iscyborg(usr))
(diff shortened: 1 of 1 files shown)


CrazyDrakkon:
Even before the new strip menu mice were able to pickpocket, rip pickpocket mice

code/datums/elements/strippable.dm
```diff
@@ -368,6 +368,10 @@
 	if (.)
 		return
 
+	if (!iscarbon(usr) && !iscyborg(usr))
```
AMyriad, line 371: note that this may prevent medborgs from doing some surgeries on dead/unconscious patients (like organ manipulation)

Moltijoe (approved):

PR #22538, opened 2024-08-17
Exosuit Update 3: Overheat Edition

SapphicOverload:

Exosuits now have an overheat level which is built up when using things like weapons or the gygax's overload. After it reaches 100 it causes sparks and damage while slowing the mech down, eventually disabling it at 150 until it cools back down to below 100. Overheat level decreases by 5 per second, or up to 15 if standing still for long enough.

EMPs also cause less damage and power loss, at the cost of adding overheat based on severity. The overheat from EMPs cannot go over 130 to prevent permanently disabling a mech by spamming an ion rifle.

**New equipment for managing heat:**

Radiator:
> Reduces overheat by 1.5 per second. Requires an atmosphere, but works without power.

Active cooling system:
> Reduces overheat by 3 per second. Requires power, and is disabled by EMPs.

Heat sink:
> Reduces both heating and cooling by 20%.

**Other changes;**

Medical beam:
> Adds 4 heat per second while active. Can now (again) equip more than one.

Afterburner:
> Cooldown reduced from 10 to 7 seconds, adds 60 overheat when used. Interrupts cooling bonus from standing still.

Armor:
> Removed 1 second cooldown, instead reduces all cooling by 20% each. Reduces ranged armor's protection to compensate for full uptime, and removes stacking multiple of the same armor type.

Currently there isn't really any sort of balancing mechanic for mechs other than "use more power which can be entirely negated by using an infinite slime battery" or "damage your own mech immediately which makes using it at all a permanent setback until you get out and repair it."

# Testing

https://github.com/user-attachments/assets/c3582577-dfa0-47e0-bfeb-3fbe3a1b01e9

# Spriting

image

image

image

image

# Wiki Documentation

Guide to robotics needs the new equipment added to it.

Probably also needs a mention of the heat mechanics but I'm too lazy to do that at the moment.

# Changelog

:cl:
experimental: Added overheat mechanics to exosuits
bugfix: Fixed charge and integrity not updating properly on diagnostic HUD
bugfix: Probably other stuff that I forgot about
imageadd: Added HUD icons for overheating
imageadd: Added sprites for new mech equipment
/:cl:


PibeCantina:
Peak

cuackles (approved):

Moltijoe (approved):

PR #22561, opened 2024-08-27
Adjusts Lone Op's weight increase requirements & sets a general weight for it to spawn/

cowbot92:

Adds 25 min player weight to Lone Op, adjusts the disk processing so the weight # chance doesn't increase unless current total pop is above or equal to 25 players

A lot of rounds end early because a Lone Operative rolled while none of the 4-10 crew members thought to secure the disk. This is not a deletion of Lone Op, and it can be re-enabled at a later time.

# Changelog

:cl:
tweak: Adjusts Lone Operative spawn chances

/:cl:


ToasterBiome:
just pop lock it (polka dot it)

cowbot92:
> just pop lock it (polka dot it)

you're pop locked

ToasterBiome (approved):

PR #22572, opened 2024-09-01
[All Stations] Corrects and cleans up Mech Bay wiring

Aquizit:

Removes wires unneeded that connect to chargers and computers. Makes sure wires go to APCs on the stations where they were missing

# Testing
image
image

# Changelog

:cl:
bugfix: Correctly wires APCs that were improperly done
tweak: Removed wires connecting to machines that did not need them
/:cl:

### DIFF
--- a/_maps/map_files/AsteroidStation/AsteroidStation.dmm
+++ b/_maps/map_files/AsteroidStation/AsteroidStation.dmm
@@ -7762,9 +7762,6 @@
 /obj/machinery/computer/mech_bay_power_console{
 	dir = 4
 	},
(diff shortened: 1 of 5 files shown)


AMyriad:
nice, these ones always bugged me

AMyriad (changes requested):
Two stray wires in IceMeta

image

Aquizit:
meta existing as it did was what made me think about if the wires needed to be there at all, but alas, it was also something missed a long time ago it seems

AMyriad (approved):

PR #22622, opened 2024-09-20
Rewrites speech sounds + preferences

Moltijoe:
They're now datums rather than a list

lets people control the volume of other's speech, or even turn it off if they hate it that much
lets people choose what their character sounds like (roughly)
also adds framework to allow for adding additional voices REALLY easily

# Testing
image
image

:cl:
rscadd: Can now mute speech sounds via a preference
rscadd: Can now control the volume of speech sounds via a preference
rscadd: Can now select the voice a character has
/:cl:


ToasterBiome (changes requested):
ok but you left the TTS comments in

ToasterBiome (approved):

PR #22625, opened 2024-09-20
Hides synthetics from the job menu

Moltijoe:
they're disabled, they shouldn't be seen

# Testing
image

:cl:
bugfix: Hides synthetics from the job menu
/:cl:


wonderinghost:
Will this also fix the spawning of ghost synths

SapphicOverload (approved):

PR #22632, opened 2024-09-22
Gives the Vatra M38 a 10% chance to spawn in as a silver plated version

ToasterBiome:

"Someone" saw the lighter version of the Vatra M38 and wanted it in the game, I added a silver plated semi-unique version. It has a 10% chance to spawn in as a 'Vatra M38 "Hauard" Pistol', which functions exactly the same but is silver plated and has the serial number scratched out (description only)

SPEND ALL THE CARGO BUDGET ON VATRAS SO I CAN GET THE SPECIAL GUN!!!!!!!!!!!!!!!!!

# Testing

# Spriting
image

# Wiki Documentation
Just a special version, literally the same

# Changelog

:cl:
rscadd: Adds a 10% chance to get a "special" Vatra M38 in a crate. Only one can exist
/:cl:


Mqiib (approved):

code/modules/projectiles/guns/ballistic/pistol.dm
```diff
@@ -130,6 +132,23 @@
 /obj/item/gun/ballistic/automatic/pistol/v38/less_lethal
 	starting_mag_type = /obj/item/ammo_box/magazine/v38/rubber
 
+/obj/item/gun/ballistic/automatic/pistol/v38/less_lethal/Initialize(mapload)
+	if(GLOB.does_howard_exist || prob(90))
+		return ..()
+	GLOB.does_howard_exist = TRUE
+	new/obj/item/gun/ballistic/automatic/pistol/v38/less_lethal/howard(loc)
```
ynot01, line 139: deleting the item and replacing it in its loc will not always work; a less destructive method is to reskin the item, refresh its overlays, and store its state in a variable like `is_howard` and then handle deletion logic (global FALSE set) if that is true

ynot01 (changes requested):

code/modules/projectiles/guns/ballistic/pistol.dm
ToasterBiome: They only come in crates, handling it in initialize is robust enough that it will handle crates, admin spawning, and spawning in containers

ynot01 (approved):

PR #22639, opened 2024-09-24
Nerfs the caltrop component to only slow at low damage values

Mqiib:
Ok lizards are just worse than humans now, plus glass shards are just really annoying. This significantly reduces the effects of stepping on glass shards if you are either a digitigrade fan or have light step. Notably, this will not affect anything that has REAL caltrop damage like punji sticks or idk actual caltrops if someone decided to make those a thing.

Caltrops now work SLIGHTLY differently at low damage values.
1: Digitigrade users take 75% damage from caltrops because they don't really wear shoes anyways
2: If the caltrops do less than 5 damage, the stepper will be slowed (1.0 slowdown) for the stun duration rather than paralyzed

This means that if you are either a lizard/polys/weirdo with funny legs for some reason OR if you have light step, you won't get stunned. Everything else is otherwise unchanged.

Lizards and polys no longer just straight up die because of a random shard/lightbulb hidden in a giant pile of junk. Alternatively, if you're a human/moth/plant/whatever player with light step it now is ACTUALLY USEFUL because you wont get stunned from shards if you're not wearing shoes. Everyone else still falls over dramatically because they weren't watching where they were going.

# Testing
Punji Sticks (with digitigrade + light step)
image
image

Glass shard:
Digitigrade
image

Digitigrade + light step
image
image

Light step
image
image

Nothing
image
image

# Wiki Documentation

Lizards and polys are slightly resistant to stepping on shards and other caltrops.

# Changelog

:cl:

tweak: Caltrops now only slow if the damage is below 5
tweak: Digitigrade legs are slightly more resistant to caltrops (75%)

/:cl:

### DIFF
--- a/code/datums/components/caltrop.dm
+++ b/code/datums/components/caltrop.dm
@@ -94,6 +94,9 @@
 	if(HAS_TRAIT(digitigrade_fan, TRAIT_LIGHT_STEP))
 		damage *= 0.75
 	
+	if(HAS_TRAIT(digitigrade_fan, TRAIT_DIGITIGRADE))	//They're used to it
+		damage *= 0.75
+
 	digitigrade_fan.apply_damage(damage, BRUTE, picked_def_zone, wound_bonus = CANT_WOUND)
 
 	if(cooldown < world.time - 10) //cooldown to avoid message spam.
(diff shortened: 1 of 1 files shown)


iloveloopers:
removes soul

Scrambledeggs00:
Big Lizard at it again. Soon they'll be putting themselves into the captain's role!

ToasterBiome (approved):

PR #22658, opened 2024-09-27
fixes backstories sometimes crashing if you're hijack or martyr bound 

cowbot92:

fixes this

# Testing

removing returns is how to fix this

# Changelog

:cl:
bugfix: Fixed bacjstorises crashing

/:cl:


ToasterBiome (changes requested):
you could probably just setup the backstories right before each return

cowbot92:
> you could probably just setup the backstories right before each return

done @ToasterBiome

ToasterBiome (changes requested):
pretty sure you need to pass some args into those procs

cowbot92:
> pretty sure you need to pass some args into those procs

dont be a baby the proc can figure it out itself

PR #22671, opened 2024-10-03
Removes old PDAs

JohnFulpWillard:

- Removes all instances of old PDAs
- Removes message console's ability to fake messages (it only worked for old PDAs, can re-add if you make them work on new ones instead)
- Replaces pAI's messenger with their own tablet so they actually get to keep theirs.
- Makes PDA painter work with new PDAs/Phones instead (done by ynot01)

old PDAs are nearly completely unused and the few places they are used barely function, faking messages is utterly useless to anyone that isn't a pAI, and said pAI would then just immediately question why they are getting a PDA message when there's no way to obtain one in-game.
This is all just very messy and fucked up so it's better to remove them and start better support for the new PDAs in the future instead.

# Testing

Tested pAIs buying and using their new PDA
Tested pAIs using their signaler cause I messed with that
Tested changing people's names as Admin to ensure PDAs were renamed as well
Tested PDA painters on PDAs and phones.

Not really much else to test since basically everything else is removal.

# Changelog

:cl: JohnFulpWillard & ynot01
rscdel: Removed old PDAs and the ability to fake messaging with the telecomms messaging monitor console.
tweak: PDA painters now work with new PDAs and phones.
tweak: pAIs now use new PDAs.
/:cl:


ynot01:
> Removes message console's ability to fake messages (it only worked for old PDAs, can re-add if you make them work on new ones instead)

A new console function is not needed since PDAs can set their username to anything they like, all the syndicates need is access to PDAs

AMyriad:
I knew this day would come
alas just renaming them to "antique PDAs" couldn't save them 😔 (thanks for doing this)

AMyriad (approved):
+1 rep decluttered hop offices

ynot01 (dismissed):
Undo whatever you did to the PDA painters, I will make them work with new PDAs (biome approved)

ToasterBiome (approved):
byee

PR #22680, opened 2024-10-07
[BOX] Puts an intercom on the wall by the glass flower case outside of security so sentient butterflies may call for help on comms (and receive none as everyone ignores their pleas for rescue anyway)

neeshacark:

title

# Changelog

:cl: cark
mapping: adds an intercom by the butterfly case outside of sec on box
/:cl:


AMyriad (approved):
10/10 pr

PR #22696, opened 2024-10-10
Adds alternate blood overlays for digitigrade legs

SapphicOverload:
Digitigrade legs are a different shape and don't line up with the normal blood overlays, this corrects that by adding alternate overlays.

image

:cl:
imageadd: Alternate blood overlays for digitigrade legs
/:cl:


ynot01 (approved):

Mqiib (approved):
Anything to fix the weird overlay mismatch

PR #22697, opened 2024-10-10
Lizards now make clawprints instead of normal footprints

SapphicOverload:
Doesn't make much sense for lizards to be making normal footprints

image

:cl:
tweak: Lizards now make clawprints instead of normal footprints
/:cl:


ynot01 (approved):

Mqiib:
Polys probably should have this too

PR #22719, opened 2024-10-15
[DONUT] Redoes all of Engineering and some areas around it

ToasterBiome:

I think the engineering is UGLY so i want to make it BETTER
- Moves grav generator to Engineering
- Splits it into more areas
- Uses condensed SMES style like Gax/Mini
- Uses smaller Box Vault
- Slightly shortens lawyer room that's weirdly shaped
- Removes old grav generator area and makes it maint

image
image

Better looking = Better Functioning = Better Life

# Changelog

:cl:
mapping: DonutStation Engineering has been completely redone
/:cl:

### DIFF
--- a/_maps/map_files/DonutStation/DonutStation.dmm
+++ b/_maps/map_files/DonutStation/DonutStation.dmm
@@ -59,17 +59,6 @@
 /obj/machinery/telecomms/bus/preset_four,
 /turf/open/floor/circuit/telecomms/mainframe,
 /area/tcommsat/server)
-"abg" = (
-/obj/effect/turf_decal/trimline/secred/filled/line/lower{
-	dir = 10
-	},
-/obj/machinery/airalarm{
-	dir = 4;
-	pixel_x = -24
-	},
-/obj/item/kirbyplants/random,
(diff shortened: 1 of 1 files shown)


ynot01 (approved):

N3D6:
only 4 smes units stinky

neeshacark (changes requested):
image
fix the name of this door (no grav gen) and maybe make it a double door. it might look better.

image
the yellow decals outside might need to be rounded off with corner decals before they reach. your call but that's my take.

image
this door is missing firelocks

image
maybe remove this disposal pipe and redo it to connect out in the central hall instead of making it go under a table which is bad and cringe

image
maybe put a fire alarm in this area

image
need a corner blue decal here

image
make this a normal wall it has no reason to be reinforced

ToasterBiome:
I've done all the changes except for the disposal one.

The reason why i am not changing it is because i dont like the sorter having multiple sort types and then we have to further sort it. I'd like to avoid this when possible (you can see that i do it in ministation too)

neeshacark (approved):

PR #22729, opened 2024-10-17
reverts Increases the price of Australian Slime Mutator from 10 TC to 25 TC.

ktlwjec0:
Partially reverts #6222 
Reduces the price of Australicus Slime Mutator to 10tc. Thinking about making it 15, just seeing what people think.

I have seen this used once in about 5 years, it may as well just not exist with all the restrictions in place.
Also, original PR was upping the price, fine. But then it was requested to just move it to a hijack restricted item, fine. But the price wasn't changed back.
I think if you manage to roll hijack you should be able to enjoy this instead of then also having to find another traitor for extra tc.

# Testing
simple number change, shouldnt need it

:cl: ktlwjec
tweak: Australicus Slime Mutator reduced to 10TC.
/:cl:


LassiH:
isnt the spider this makes already normally available from a gold extract

JamieD1 (approved):

PR #22796, opened 2024-11-08
Makes the NTSL traffic console no longer save on pressing ENTER, replaces x.len with length(x) in the scripting module

Gboster-0:

- This PR makes a special class extending from TextArea that does not save when pressing ENTER for NTSL
- Replaces X.len with length(X)

> This PR makes a special class extending from TextArea that does not save when pressing ENTER for NTSL
- This makes it so you will no longer get your text position thingy moved to the very last line of your code if you are holding down ENTER, alongside possibly reducing lag that'd come from pressing it.
- Being jumped to the last line when holding down enter can be very annoying at times, so imo its gud.
> Replaces X.len with length(X)
- Apparently getting the var with length() is faster

# Testing

Tested locally, maybe i'll add a video here later.

# Changelog

:cl:
tweak: Made the NTSL console's script input area no longer save when pressing ENTER
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/NtslEditor.js
+++ b/tgui/packages/tgui/interfaces/NtslEditor.js
@@ -2,6 +2,57 @@ import { useBackend, useLocalState } from '../backend';
 import { Box, Button, Divider, Input, Tabs, TextArea, Section, Stack } from '../components';
 import { Window } from '../layouts';
 
+// NTSLTextArea component start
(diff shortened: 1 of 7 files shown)


JohnFulpWillard:
> Apparently getting the var with length() is faster

I've seen it argued both ways, the difference is so small that it might as well not exist. The difference between them is that if it's not a list length will return 0 while Len will runtime, length is still better

JamieD1 (approved):

PR #22804, opened 2024-11-11
[ALL MAPS] Adds vox breathing masks and nitrogen tanks to cloning

Oblisk234:

Adds vox breathing masks and nitrogen tanks to cloning.
Yes, I know that they can't be cloned but I have no idea where else they can go and there are zero other ways of acquiring vox breath masks except the ones vox start out with, meaning if they lose them somehow, they can't replace it.

Vox masks aren't printable via protolathe or anywhere else. This gives them a way to actually replace their masks and tanks if they lose them somehow e.g. vox is stripped in maint by traitor and they can't find their mask or tank on revival.

# Testing
On ministation:
image

Rest can be seen in mapdiffbot (if it doesn't get angry for some reason).

# Changelog

:cl:
mapping: Adds vox masks and nitrogen tanks to cloning.
/:cl:

### DIFF
--- a/_maps/map_files/DonutStation/DonutStation.dmm
+++ b/_maps/map_files/DonutStation/DonutStation.dmm
@@ -22216,6 +22216,8 @@
 	dir = 1;
 	network = list("ss13","Medical")
 	},
+/obj/item/clothing/mask/breath/vox,
+/obj/item/tank/internals/emergency_oxygen/vox,
 /turf/open/floor/plasteel/white,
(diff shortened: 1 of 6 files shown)


AMyriad (approved):

13spacemen:
@Oblisk234 You can actually wear regular breath masks just fine as a Vox
Only difference is that you can't eat through it, while Vox ones allow eating through them

ToasterBiome (approved):

PR #22808, opened 2024-11-14
Deletes the derelict

AMyriad:
This thing, to be more specific

image

It was removed along time ago for balance reasons iirc (/datum/map_template was removed but everything else stayed)
This fully removes the map file and everything unique to it for the sake of cutting out bloat, also I don't see a good reason to keep maintaining this

Feel free to restore or revert if you disagree

# Testing
ask if needed

# Wiki Documentation
Yes we should delete these
In space ruins, go to "Kosmicheskaya Stantsiya 13" on https://wiki.yogstation.net/wiki/Ruins 
and https://wiki.yogstation.net/wiki/Derelict

# Changelog
Nothing player facing, this isn't in the game to begin with


nesisknesis:
I can't find what PR disabled this so i'm not convinced at all that this shouldn't be in the game and active. Derelict to me is the most iconic space ruin, and I would wager anyone who has played for a decent time would agree that it is an endearing location. I don't even remember it being that unbalanced. I only recall the best loot being there was a suspicious toolbox. 

I've read there were drones there which I agree are a pain in the ass but anything problematic should have just been cut instead of the entire ruin.

This is because it is also a tribute to the original map. 

14089610823_d198b5db86_o

AMyriad:
I am blind
image

could have fooled me, I've never seen this thing in game

PR #22188, opened 2024-06-08
Refactors gas reagents

SapphicOverload:
# Document the changes in your pull request

Refactors how gas breath effects work, and the gas reagents themselves. Also fixes preterni sometimes not being affected by certain gases, and allows any gas with reagent effects to work in cryo tubes.

Closes #22187

# Why is this good for the game?
Significantly better code, allows most gases with breath effects to work in cryo tubes.

# Testing

image

# Changelog

:cl:  
experimental: Refactored gas reagents and gas breath effects
tweak: Most gases with breath effects now work in cryo tubes
bugfix: Fixed certain gases sometimes not having effects on preterni
/:cl:

PR #22133, opened 2024-05-31
Healium tweaks, p2

warface1234455:
# Document the changes in your pull request
Healium now heal organic organs
Cryo tube can now inject healium into dead body
Healium when applied to dead body now heals organic organs (can fix the failure organs aswell), brutes and burns, to a safe level for defibbing.
Healium grenade now injects healium into nearby mobs if they their overall health is below or equal to 20, softcrit, dead, or unconscious

# Why is this good for the game?
Makes sense for healium, it can already interact with organic body to replicate and repair the body, so it should be able to heal the overall of the organic body. Gives atmos tech more interactions with medical to help them during blood hours (medbay is understaffed and filled with dead bodies).

# Testing

https://github.com/yogstation13/Yogstation/assets/89688125/1611b6f0-591d-4632-89a4-34a74416a7a2

# Wiki Documentation
in doc of change

# Changelog

:cl:  

tweak: Healium now heal organic organs
tweak: Cryo tube can now inject healium into dead body
tweak: Healium when applied to dead body now heals organic organs (can fix the failure organs aswell), brutes and burns, to a safe level for defibbing.
tweak: Healium grenade now injects healium into nearby mobs if they their overall health is below or equal to 20, softcrit, dead, or unconscious
/:cl:

PR #22843, opened 2024-12-10
healium tweaks p3

warface1234455:
healium now prevents body from rotting
cryo tube now prorperly heal dead body (with extra check to prevent healium duplication) to a reviable state when healium reagent exists in the body (#22133), broken by (#22188)

healium could prevent organs from rotting but not the entire body is whack, allowing doctors to normally defib one without the need of revival surgery (body decayed) if healium exists in the bloodstream

plus bug fix

# Testing
tested on local
cryooo

# Changelog

:cl:
tweak: healium now prevents body from rotting
bugfix: cryo tube now prorperly heal dead body (with extra check to prevent healium duplication) to a reviable state when healium reagent exists in the body (#22133), broken by (#22188)
/:cl:

### DIFF
--- a/code/datums/components/rot.dm
+++ b/code/datums/components/rot.dm
@@ -54,8 +54,9 @@
 		return
 
 	// Wait a bit before decaying
-	if(world.time - C.timeofdeath < 2 MINUTES)
-		return
+	if(!C.reagents.has_reagent(/datum/reagent/gas/healium))
+		if(world.time - C.timeofdeath < 2 MINUTES)
+			return
 
 	// Properly stored corpses shouldn't create miasma
 	if(istype(C.loc, /obj/structure/closet/crate/coffin)|| istype(C.loc, /obj/structure/closet/body_bag) || istype(C.loc, /obj/structure/bodycontainer))
--- a/code/datums/components/rot.dm
+++ b/code/datums/components/rot.dm
@@ -54,9 +54,8 @@
 		return
 
 	// Wait a bit before decaying
-	if(!C.reagents.has_reagent(/datum/reagent/gas/healium))
-		if(world.time - C.timeofdeath < 2 MINUTES)
-			return
+	if(world.time - C.timeofdeath < 2 MINUTES)
+		return
 
 	// Properly stored corpses shouldn't create miasma
 	if(istype(C.loc, /obj/structure/closet/crate/coffin)|| istype(C.loc, /obj/structure/closet/body_bag) || istype(C.loc, /obj/structure/bodycontainer))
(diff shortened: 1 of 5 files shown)


code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
```diff
@@ -321,6 +321,17 @@
 			gas_reagent.reaction_mob(mob_occupant, VAPOR|BREATH, 2, permeability = 1)
 			air1.adjust_moles(gas_id, -0.1 / efficiency)
 			qdel(gas_reagent)
+		if(C && C.stat == DEAD && C.reagents.has_reagent(/datum/reagent/gas/healium)) // attempts to cycle healium in the body to a reviable state
+			if(!C.can_defib(FALSE))
+				C.reagents.del_reagent(/datum/reagent/gas/healium)
+			else
+				set_on(FALSE)
+				playsound(src, 'sound/machines/cryo_warning.ogg', volume) // Bug the doctors.
+				var/msg = "Dead body has been patched up to a reviable state."
```
ynot01: revivable

ynot01 (changes requested):

ynot01:
> healium now prevents body from rotting

looking at the code it seems that healium in fact reverses body rot, not just prevents it

ShoziOffical:
> > healium now prevents body from rotting
> 
> looking at the code it seems that healium in fact reverses body rot, not just prevents it

Yeah maybe i have used the wrong word, but the point is allowing defib to work with rot body, if its at defib able state and with healium inside

ynot01:
> Yeah maybe i have used the wrong word, but the point is allowing defib to work with rot body, if its at defib able state and with healium inside

change changelog to better reflect changes

ShoziOffical:
> > Yeah maybe i have used the wrong word, but the point is allowing defib to work with rot body, if its at defib able state and with healium inside
> 
> change changelog to better reflect changes

I was going to make it actually prevent rot

code/datums/components/rot.dm
```diff
@@ -54,8 +54,9 @@
 		return
 
 	// Wait a bit before decaying
-	if(world.time - C.timeofdeath < 2 MINUTES)
-		return
+	if(!C.reagents.has_reagent(/datum/reagent/gas/healium))
```
ynot01: this is incorrect, it should return if has_reagent returns true, as returning means no rotting

ynot01 (changes requested):

code/datums/components/rot.dm
ShoziOffical: OH WAIT MY BAD

ynot01 (approved):

ToasterBiome:
No longer needed

PR #22866, opened 2024-12-20
Food Cart UI rework

### LINKED ISSUE #9310: Yogsmeta kitchen's Food Cart Mixer does not work
## Round ID:
30756

## Reproduction:
I'm unable to use the mixer in the kitchen on Yogsmeta. I was playing as cook and attempted to make dough the way I know how. I put the ingredients into the food cart and attempted to add them to the cart's mixer, but it would not show them as being inserted. I used a glass to mix ingredients but once I attempted to pull something from a food cart in a glass, it gave me an empty glass. At that point, I ahelped and after some time the admin recommended that I restart my game. After restarting, it didn't work so the admin recommended that I report this issue on github.
Steps to reproduce:
1) Put assorted ingredients into the food cart
2) Attempt to add ingredients into the food cart's mixer and attempt to eject an ingredient inside a glass.

### LINKED ISSUE #19427: foods with ' in the name can't be dispensed via food cart
### Severity

Inconvenience

### Short Description

what the title says

### Reproduction

Place clown's tears or 'Jelly' spag-o-nut in the cook's food cart. Can't be dispensed.

### Round ID

All

### Test Merges

_No response_

### BYOND Username

Ktlwjec

### Was this done on a locally hosted or non-Yogstation hosted server?

Yes

### Additional information

_No response_

### LINKED ISSUE #19482: Parmesan when matured cant be dispensed via the food cart
### Severity

Inconvenience

### Short Description

If you put unmatured Parmesan in the food cart and dispense it after 20 minutes (when it should mature) it vanishes.

### Reproduction

1. Make Parmesan cheese.
2. Place in food cart when unmatured.
3. Wait for Parmesan is mature (20 minutes).
4. Try to dispense Parmesan (still shows as unmature in the food cart).
5. It is gone, vanished, as if it never existed.

### Round ID

47984 but I'd assume all

### Test Merges

_No response_

### BYOND Username

ktlwjec

### Was this done on a locally hosted or non-Yogstation hosted server?

Yes

### Additional information

_No response_

Scrambledeggs00:

I've changed the food cart's UI from byond HTML into TGUI as it looks nicer. The new UI is broken up into two parts: food and drinks.

Food Tab

- Food Capacity section shows the current amount of food and the total capacity of the cart, which is 80 food items.
- Food Selection section shows the cart's food, with each food item showing an image, its name, quantity within the cart's storage, and a button for dispensing one.
- Similar to the ice cream vat, the cart will dispense the last item of the dispensed type put into it.

Drink Tab
- Glass Storage section shows the current amount of drinking glasses in storage, with carts starting with 10, and the total capacity, which is 30.
- Drink Capacity section shows the current units of reagents stored in the cart along with the total reagent capacity of the cart, which is 200.
- Transfer Amount section has a list of buttons that determines how many units are shifted between the different storages and how much are purged.
- Cart Drink Storage is the main reagent storage. As previously stated, it has a max volume of 200u. Reagents within it can either be purged from the cart or transferred to the mixer based on the transfer amount selected.
- Cart Mixer Storage is the storage used to dispense drinking glasses. It has a max volume of 50u, enough to fully fill a drinking glass. Reagents within can either be transferred back to the cart's main storage based on the transfer amount selected or all be poured into a drinking glass and dispensed to the user.
- Both storages have the NO_REACT flag, meaning you cannot make mixed drinks within the cart. I did this so that drinks would mix upon being dispensed, which is pretty cool :]
- I do not know how to prevent reagent containers from transferring, so the cart can actually hold 250u, 200u in main, and 50u in mixer.

The UI also has info tooltips to explain certain UI elements. Simply hover your mouse over the blue circles with a white i to see the tooltip.

Also, when destroyed, food carts will now spill their reagents. A chat warning will also be sent if either the main or mixer storages have reagents, alerting you of the spill.

Fixes #9310
Fixes #19427
Fixes #19482

P.S.
With the replacement of find_amount with datums, food items like parmesan which add and remove things from contents themselves do not work well with keeping count for UI items. Because of this, stuff like unmatured parmesan cheese will show however much you put into it even when some have matured and will only remove itself from the UI when no unmatured cheese is in the contents. For matured, it will only count the first one added and any you add with attackby, items created within the contents are not counted because I do not know how to do that without essentially just bringing back find_amount which would make the switch pointless.

Old UI looked bad, reagent part of carts haven't worked in awhile, and spills are funny :]

# Testing
Video of cart UI:
https://www.youtube.com/watch?v=FjUYch4DA8A
Also: I changed the spill message to specify that reagents are spilling out after this video.

# Wiki Documentation

Food capacity: 80
Glass capacity: 30
Drink capacity

- Main: 200u
- Mixer: 50u
- Total: 250u

Reagents will spill if cart is destroyed.

# Changelog

:cl:
rscadd: Food carts now spill their reagents upon destruction 
bugfix: Food carts can now handle reagents again
tweak: Reworked food cart's UI
/:cl:


code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +6,259 @@
...
+		details["quantity"] = find_amount(item)
+		details["type_path"] = item.type
+
+		//Get an image for the UI
+		var/icon/item_pic = getFlatIcon(item)
+		var/md5 = md5(fcopy_rsc(item_pic))
+		if(!SSassets.cache["photo_[md5]_[item.name]_icon.png"])
+			SSassets.transport.register_asset("photo_[md5]_[item.name]_icon.png", item_pic)
+		SSassets.transport.send_assets(user, list("photo_[md5]_[item.name]_icon.png" = item_pic))
+		details["image"] = SSassets.transport.get_asset_url("photo_[md5]_[item.name]_icon.png")
```
hedgehog1029: is this code copied from somewhere? i have concerns about doing this in `ui_data`, as well as the repeated `send_assets` calls with a single asset each time

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +6,259 @@
...
+
+		//Add to food list
+		data["food"] += list(details)
```
hedgehog1029, line 105: I think you can avoid the extra list allocation here using list.Add
```suggestion
		data["food"].Add(details)
```

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +6,259 @@
...
+			playsound(src, insert_sound, 50, TRUE, extrarange = -3)
+		return
+	else if(A.is_drainable())
 		return
+
+	..()
+
+/obj/machinery/food_cart/proc/dispense_item(received_item, mob/user = usr)
+	//Make a variable for checking the type of the selected item
+	var/obj/item/reagent_containers/food/ui_item = new received_item
```
hedgehog1029: this variable isn't freed, but it also might be worth using `locate` instead of instantiating a new item

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +6,259 @@
...
+	if(istype(A, /obj/item/reagent_containers/food/snacks))
+		storage_single(A)
+		return
+	else if(istype(A, /obj/item/reagent_containers/food/drinks/drinkingglass))
+		//Check if glass is empty
+		if(!A.reagents.total_volume)
+			//Delete glass and increment glass_quantity by 1
+			qdel(A)
+			glass_quantity++
+			user.visible_message(span_notice("[user] inserts [A] into [src]."), span_notice("You insert [A] into [src]."))
```
hedgehog1029: move the `qdel(A)` to after the visible_message call that references `A` to avoid use-after-free

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
hedgehog1029: when i did vending machine custom items i did a trick with base64, though i never tested the performance characteristics of that

Scrambledeggs00: It is copied. If I remember correctly, I copied it from bug eating lizard's nullrod code.

Scrambledeggs00: I could try moving it to static_data or whatever it is called, but I'm going to be too busy to do that for the next couple of days.

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Scrambledeggs00: What does locate do? Nothing came up when I hovered over it in VSC and from what I have seen, it just says if an argument is in a list or not.

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Scrambledeggs00: I tried that and it somehow prevents the UI from showing food.

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Oblisk234: Locate finds the first instance of a type.

Scrambledeggs00: Then it won't work for me, since I specifically want to dispense the last instance of the item type in the list. That way, it gets the item last put in, making potential poisonings simpler to do.

Scrambledeggs00: I might be able to do away with the instancing all together, but I don't have time today to test that, though I might have time tomorrow.

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Scrambledeggs00: I tried moving the image part to static_data, but I couldn't figure out how to get it to show the image in the UI. I'll try again when I have time.

Oblisk234:
Does this fix #19482 at all?

Scrambledeggs00:
> Does this fix #19482 at all?

I haven't tested it yet, though i would assume that it doesn't. I'll try to test it today.

Scrambledeggs00:
> Does this fix #19482 at all?

I tested it and made the necessary changes to fix the bug. The fix should also work for any other food item that works similarly.

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Scrambledeggs00: I switched over to base64 from your code

hedgehog1029: conversion to base64 looks good, resolving

hedgehog1029 (approved):

tgui/packages/tgui/interfaces/FoodCart.tsx
```diff
@@ -0,0 +1,488 @@
...
+          <MainDrinkRow />
+        </Section>
+      </Stack.Item>
+      <Stack.Item>
+        <Section
+        title="Cart Mixer Storage"
+        textAlign="center"
+        buttons={<Button
+          circular
+          tooltip="Reagents to be poured into drinking glasses"
```
Oblisk234: Could you add to the Cart Mixer Storage tooltip that reagents will mix after pouring the glass? It may not be obvious where in other cases this is expected to immediately mix (Hoochmaster, chemmaster, e.g.)

Scrambledeggs00: I did it for both storage tooltips

Oblisk234 (approved):

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +16,273 @@
...
+	mixer = new /obj/item/reagent_containers(src, 50)
+	mixer.create_reagents(50, NO_REACT)
 
 /obj/machinery/food_cart/Destroy()
+	//Only alert others if the cart or mixer has any reagents
+	if(mixer.reagents.total_volume > 0 || reagents.total_volume > 0)
+		visible_message(span_alert("[src] spills all of its liquids onto the floor!"))
+	//Spill reagents on the cart's turf
+	//Thanks baiomu for telling me how to spill reagents
+	var/turf/spill_area = loc
```
ToasterBiome:
```suggestion
	var/turf/spill_area = get_turf(src)
```

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +16,273 @@
...
+	if(find_amount(ui_item) > 0)
+		//Select the last(most recent) of desired item
+		var/obj/item/reagent_containers/food/snacks/dispensed_item = LAZYACCESS(contents, last_index(ui_item))
+		//Drop it on the floor and then move it into the user's hands
+		dispensed_item.forceMove(loc)
+		user.put_in_hands(dispensed_item)
+		user.visible_message(span_notice("[user] dispenses [dispensed_item.name] from [src]."), span_notice("You dispense [dispensed_item.name] from [src]."))
+		playsound(src, dispense_sound, 25, TRUE, extrarange = -3)
+		//If the last one was dispenced, remove from UI
+		if(find_amount(ui_item) == 0)
```
ToasterBiome:
```suggestion
		if(!find_amount(ui_item))
```

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +16,273 @@
...
-	updateDialog()
+		//Incase the UI buttons are slow to disable themselves
+		user.balloon_alert(user, "All out!")
+
+	//Delete instance
+	qdel(ui_item)
+
+/obj/machinery/food_cart/proc/storage_single(obj/item/target_item, mob/user = usr)
+	//Check if there is room
+	if(contents.len - 1 < contents_capacity)
```
ToasterBiome, line 249:
```suggestion
	if(length(contents) < contents_capacity)
```
why is this - 1 here?

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +16,273 @@
...
+		playsound(src, insert_sound, 50, TRUE, extrarange = -3)
 
-/obj/machinery/food_cart/Topic(href, href_list)
-	if(..())
 		return
+	else
+		//Warn about full capacity
+		user.balloon_alert(user, "No space!")
+
+/obj/machinery/food_cart/proc/pour_glass(mob/user = usr)
```
ToasterBiome: we don't want to pass in usr is there another way to get the action user in ui_act?

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +16,273 @@
...
+	//Make a variable for checking amount of item
+	var/obj/item/reagent_containers/food/ui_item = new received_item
+
+	//If the vat has some of the desired item, dispense it
+	if(find_amount(ui_item) > 0)
+		//Select the last(most recent) of desired item
+		var/obj/item/reagent_containers/food/snacks/dispensed_item = LAZYACCESS(contents, last_index(ui_item))
+		//Drop it on the floor and then move it into the user's hands
+		dispensed_item.forceMove(loc)
+		user.put_in_hands(dispensed_item)
```
ToasterBiome: there's a proc where it will attempt to put it in someone's hands and drop to the floor if it cant

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +16,273 @@
...
-	mixer = new /obj/item/reagent_containers(src, MIXER_CAPACITY)
-	mixer.name = "Mixer"
+	//Create reagents holder for drinks
+	create_reagents(reagent_capacity, OPENCONTAINER | NO_REACT)
+	mixer = new /obj/item/reagent_containers(src, 50)
+	mixer.create_reagents(50, NO_REACT)
 
 /obj/machinery/food_cart/Destroy()
+	//Only alert others if the cart or mixer has any reagents
+	if(mixer.reagents.total_volume > 0 || reagents.total_volume > 0)
```
ToasterBiome:
```suggestion
	if(mixer.reagents.total_volume || reagents.total_volume)
```

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +16,273 @@
...
+		visible_message(span_alert("[src] spills all of its liquids onto the floor!"))
+	//Spill reagents on the cart's turf
+	//Thanks baiomu for telling me how to spill reagents
+	var/turf/spill_area = loc
+	spill_area.add_liquid_from_reagents(mixer.reagents, FALSE, mixer.reagents.chem_temp)
```
ToasterBiome, line 192: this wont runtime if reagents or mixer.reagents are empty?

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +16,273 @@
...
+
+/obj/machinery/food_cart/ui_act(action, list/params)
+	. = ..()
+	if(.)
+		return
+
+	switch(action)
+		//Dispense food item
+		if("dispense")
+			var/itemPath = text2path(params["itemPath"])
```
ToasterBiome, line 156: You need to verify that this input is correct, like that it is available in the cart and there's enough. You can pass arbitrary data to tgui and summon literally anything in the game since it's not verified.

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
```diff
@@ -10,151 +16,273 @@
...
+		details["quantity"] = drink.volume
+		details["type_path"] = drink.type
+		
+		//Add to drink list
+		data["mixerDrinks"] += list(details)
+
+	//Get content and capacity data
+	var/list/storageDetails = list()
+	//Have to subtract contents.len by 1 due to reagents container being in contents
+	storageDetails["contents_length"] = contents.len - 1
```
ToasterBiome, line 134: this is not the length of the contents 🤔

ToasterBiome (changes requested):
preliminary review

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Scrambledeggs00: Because the cart's reagent storage is added to its contents and I only want food items to count towards capacity. I have a comment in ui_data that explains this and forgot to add one there too.
image

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Scrambledeggs00: How would I get it? Am I using the wrong proc?

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Scrambledeggs00: No runtimes unless they're somehow delayed.
image

Scrambledeggs00: I also tested to see if there were runtimes when only one was empty or when both had reagents and still no runtimes.
image

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Scrambledeggs00: Removed the forceMove code.

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
Scrambledeggs00: I moved usr to ui_act which I think is was what you want?

code/modules/food_and_drinks/kitchen_machinery/food_cart.dm
hedgehog1029: the `dispense_item` proc that this calls DOES do verification (it checks that the item is in the contents - see the `find_amount` proc) as I did notice this on my pass. textrefs might be a better option as it'd avoid the whole "creating an item just to check if it exists" dance

Scrambledeggs00: I'm guessing this is resolved then?

PR #22898, opened 2024-12-30
Deletes unneeded instances to maybe save the server from being overrun by ghost scoops and cones if such a thing is not solved by garbage collection

Scrambledeggs00:

While helping me with my food cart PR, bug eating lizard pointed out that I was not deleting useless instances of items and that it could possibly overload the server with tons of instanced items. Welp, I copied said code from my ice cream vat code, some I've come back and fixed it so it deletes the instances once they're not needed anymore.

If it was possible to overload the server before due to my crappy code, it shouldn't be possible now :]

# Testing
Everything seems to work fine with the qdel()'s and I'm not sure how I'd look to make sure the instances are actually being deleted.
image
image

# Changelog

:cl:
bugfix: Possibly fixes a bug that could cause the ice cream vat to overload the server with instanced cones and scoops
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/kitchen_machinery/icecream_vat.dm
+++ b/code/modules/food_and_drinks/kitchen_machinery/icecream_vat.dm
@@ -87,6 +87,9 @@
 			details["selected_item"] = selected_cone
 			data["cones"] += list(details)
 
+		//Delete item instance
+		qdel(item)
+
 	//Loop through children of /datum/info_tab/icecream_vat for data to send to info tab
 	for(var/info_detail in subtypesof(/datum/info_tab/icecream_vat))
 		
@@ -100,6 +103,9 @@
 
 		//Add info to data
 		data["info_tab"] += list(details)
+
+		//Delete item instance
+		qdel(item)
 
 	//Get content and capacity data
 	data["contents_length"] = contents.len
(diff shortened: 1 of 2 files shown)


code/modules/food_and_drinks/kitchen_machinery/icecream_vat.dm
```diff
@@ -155,13 +161,17 @@
 	else
 		var/obj/item/reagent_containers/food/snacks/examine_cone = new selected_cone
 		. += span_notice("<b>Alt Click</b> to dispense [examine_cone.name].")
+		//Delete item instance
+		qdel(examine_cone)
```
JohnFulpWillard:
```suggestion
		. += span_notice("<b>Alt Click</b> to dispense [selected_cone::name].")
```
i have no idea if this works it might need some edits

Scrambledeggs00: This gave me errors, but a single : worked. I also used it with info_tab, but not for the scoop and cone part of the UI since I believe it won't work for getting their images.

JohnFulpWillard: nononono that's not good at all, they are entirely different operators and ``:`` is something we should avoid

https://www.byond.com/docs/ref/#/operator/:
https://www.byond.com/docs/ref/#/operator/::

ToasterBiome (approved):

PR #22913, opened 2024-12-31
Remove bloodsucker from the codebase

Moltijoe:
Bloodsucker code is terrible, it has poor modularity, and relies on a snowflake type of action, rather than using the existing spell framework
Also, certain clans are very buggy, such as Lasombra not being able to ascend, or Gangrel as a whole
and i'm certain there's plenty more bugs that aren't actively on my radar

Bloodsucker also has some fundamental flaws in terms of game design
- Lairs are too important to gameplay, but impossible to properly hide
- non-team based conversion antag can often end up taking over the station
- powers are simple and one-dimensional compared to other "magical" antags, resulting in boring combat
- Sol mechanic is a forced mini-game that leaves a bloodsucker completely vulnerable, unless they're already doing well enough, in which case it just forces a player into time-out for aminute
- Masquerade is an unintuitive mechanic that seems half stealth, and half just for "flavour"

these all culminate in an antag that feels like it was designed to recreate Vampire the Masquerade, rather than being designed to be a fun antag

The amount of work it would take to fix this up and make it suitable for rotation far exceeds the development power we have, or anyone seems to be willing to spend.

I personally would much rather add more to heretic, darkspawn, or even vampire if people are still wanting the "blood sucker" aesthetic

# Testing
will need to

:cl:
rscdel: Removes bloodsucker from the codebase
/:cl:


neeshacark:
hi i'm john neverplayer and i like bloodsucker and i don't think you should remove it.

PR #22945, opened 2025-01-06
Removes unused VR maps

AMyriad:

We have only one VR map in use, this is trimming dead weight
also moves the map that _is_ in use up a level to `_maps/RandomZLevels/netmin_hub.dmm` from `_maps/RandomZLevels/VR/netmin_hub.dmm` as there is no more reason for that folder to exist

I don't want to maintain these relics if we're not going to use them

# Testing
uhhh trust

# Changelog
Nada player facing


ToasterBiome (approved):
now remove all the old code

PR #22946, opened 2025-01-06
Remakes Ministation Turbine+ adds more radsuit lockers 

Ethan4303:

Moves Turbine on Ministation so that it is connected to Atmospherics via a door
changed the room with glass floor to be similar to the room on boxstation before you enter the SMES 
gave turbine its essentials, the gas sim console, radlockers

felt annoying having to run a horseshoe pattern every time i needed something from atmos. plus its turbine, it should be attached to atmos and not give the possibility of irradiating all of engineering to the lobby

# Testing

Tested and fixed issue biome called me out for

# Spriting

N/A

# Wiki Documentation

image

# Changelog

:cl:

mapping: moved ministation turbine over to atmos, old connecting room now has some tables and stuff and extra rad lockers that were missing, gave atmos more radlockers since they were missing too

/:cl:


Ethan4303:
side note, give me feedback on what i should do with the lattice and catwalk structure, i hate doing catwalk and lattice mapping, i have to fight myself to not make it perfectly symmetrical when possible

AMyriad (changes requested):
I recommend moving these buttons out of the wall
image

moving this sign outside to make it more visible perhaps?
image

and resetting the APC's `areastring` variable so it isn't magically connected to disposals
image

Ethan4303:
> I recommend moving these buttons out of the wall image
> 
> moving this sign outside to make it more visible perhaps? image
> 
> and resetting the APC's `areastring` variable so it isn't magically connected to disposals image

fixed first two, but isnt the Apc string good? i checked how the string was done on boxstation and its the same way areastring matches the area
image

ToasterBiome (dismissed):
image
This area is strange, 

image

You have some lattices that are not nearstation here,

image
I wish this area had windows instead of all the way to the left. Also random light switch.

Overall I don't really like this move, most specifically because the top of the SM was intentionally inside the station for shenanigans, lol.

image

Also this can be moved down two tiles while we're here.

Ethan4303:
> Overall I don't really like this move, most specifically because the top of the SM was intentionally inside the station for shenanigans, lol.
> 

i personally think that the turbine always has and should be connected to atmos, i gave it some thought on what to replace it with so there would still be something above it but kept drawing a blank on what to put back there, any ideas would be great but ill die on this hill thinking turbine should be moved

AMyriad (approved):
> fixed first two, but isnt the Apc string good? i checked how the string was done on boxstation and its the same way areastring matches the area

I misread this mb, 'sall good

on another note I'm sure we'll put a room over the SM again someday that is pretty funny
construction room maybe who knows

PR #2906, opened 2018-10-07
[WIP] Ports mentors

GuppyLaxx:
There, it compiles, haven't tested it for shit because it's 8 in the morning and I was bored
Literally a straight port from hippie with some changed procs and shit

PR #22951, opened 2025-01-09
Mentor PM Panel

Oblisk234:

Turns out in #2906, there was a proc that never got a verb. Hence, with a few changes, Mentor PM Panel is now available to mentors.

For reference, I changed the mentor PM panel to only show mob names (and player ckeys when they're in the lobby) since mentors shouldn't really have the same info the admin PM verb would have.

MENTORHELP: Player (F): Hey I think [INSERT NAME HERE] is a new player, can you help them?

As the only mentor on, this is difficult to help with the current tools if you're a living player at the time. As a ghost, you can use Orbit menu, Right Click --> Mentor PM Mob, voila. As a living player, it becomes more of a weirder situation. You can either try to search for them in game (questionable, as you're looking for them for an OOC reason in an IC manner), or you can do nothing (Mentors do not do this, except by accident).

With a Mentor PM Panel, this solves the questionability of searching for someone in game for an mhelp, as well as providing a method of helping people when they normally wouldn't be able to (e.g. ghost roles).

# Testing
image
image

# Changelog

:cl:
rscadd: Added a mentor verb for the Mentor PM Panel from the original mentor PR. 
/:cl:

### DIFF
--- a/yogstation/code/modules/mentor/mentor_verbs.dm
+++ b/yogstation/code/modules/mentor/mentor_verbs.dm
@@ -3,6 +3,7 @@ GLOBAL_LIST_INIT(mentor_verbs, list(
 	/client/proc/show_mentor_memo,
 	/client/proc/show_mentor_tickets,
 	/client/proc/cmd_mentor_pm_context,
+	/client/proc/cmd_mentor_pm_panel,
 	/client/proc/dementor
 	))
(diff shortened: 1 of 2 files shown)


ynot01:
it was available when right clicking on someone

Moltijoe (approved):

PR #22959, opened 2025-01-13
Makes crew monitor code more sane

SapphicOverload:
The icon and color a species uses on the crew monitor is now stored in the species itself, instead of using a massive if-else chain.

image


ShoziOffical:
I didn't think of this back then

ynot01 (approved):

PR #22964, opened 2025-01-16
Kill objectives on traitor no longer happen under 10 pop

Aquizit:

Traitors no longer get kill objectives of any type when the player count is under 10.

No-one has fun when there are 4 crew members and the one traitor has to kill 2 of them.

This is sort of a hashed fix. It is 10 players, not 10 crew -- meaning if there are like 6 ghost roles it can still happen.
Being able to do it with like 6 players as long as there's 1 secoff works too but it's above my coding ability.

# Testing
2 crew
image

Enabled and removed 3 times, no kill objectives.
image
image
image

# Changelog

:cl:
tweak: No more murder under 10 players
experimental: Min player number can be changed
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/datum_traitor.dm
+++ b/code/modules/antagonists/traitor/datum_traitor.dm
@@ -202,7 +202,7 @@
 
 /datum/antagonist/traitor/proc/forge_single_human_objective() //Returns how many objectives are added
 	.=1
-	if(prob(50))
+	if(prob(50) && GLOB.joined_player_list.len >= 10)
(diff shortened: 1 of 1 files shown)


cowbot92:
can you actually make it so they only happen

Aquizit:
> can you actually make it so they only happen 

Uh. I mean yeah just change prob(50) to prob(100) I know you're being a smartass (I think) but it's funny how easy that would be

Moltijoe (approved):

PR #22965, opened 2025-01-16
Merges l_leg and r_leg

ToasterBiome:

Ports some of https://github.com/tgstation/tgstation/pull/70422 and merges them into one /obj/item/bodypart/leg, with both being descendants of them

I will be doing arms eventually.

Why are they split in the first place? it makes no sense, and it makes any leg checking twice as long because you have to check for _both_ legs.

# Changelog

:cl:
rscadd: Legs are now legs.
/:cl:


Mqiib:
Ok but a left leg doesn't fit on the right side tho

Aquizit (approved):
minimal map changes fine

Chubbygummibear (approved):
cowbot said it's good

Yarinoi:
does this mean you can have two left legs

PR #22973, opened 2025-01-24
Strengthens counterplay vs allergens

ynot01:

Adds a bold examine tell when someone is undergoing a allergic reaction

Epinephrine, atropine, diphenhydramine, and synaphydramine (unused) now prevent further symptoms of allergic reactions

Epinephrine now reduces current losebreath by a tenth each process, making it significantly reduce losebreath and making recovery feasible

Histamine is also affected, but not its overdose

Allergic reaction is pretty much instant death

Inexperienced doctors may not know what is happening, examine text will help them

Epipens are now actually useful for allergic reactions, as one would expect

# Testing

image

# Changelog

:cl:
tweak: Humans undergoing allergic reactions now have a swollen face examine tell
tweak: Epinephrine, atropine, and diphenhydramine now prevent further symptoms of an allergic reaction for as long as they are in the system
tweak: Epinephrine now reduces losebreath stacks by a tenth each process, making it effective for slowly treating allergic reactions
/:cl:

### DIFF
--- a/code/datums/traits/negative.dm
+++ b/code/datums/traits/negative.dm
@@ -736,6 +736,15 @@
 	if(H.stat == DEAD)
 		return
 
+	if(H.reagents.has_reagent(/datum/reagent/medicine/epinephrine))
+		return
+	if(H.reagents.has_reagent(/datum/reagent/medicine/atropine))
+		return
+	if(H.reagents.has_reagent(/datum/reagent/medicine/diphenhydramine))
+		return
+	if(H.reagents.has_reagent(/datum/reagent/medicine/synaphydramine))
+		return
+
 	var/datum/reagent/allergy = GLOB.chemical_reagents_list[reagent_id]
 
 	if(H.reagents.has_reagent(reagent_id)) //check if there are chems
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -281,6 +281,9 @@
 			if(DISGUST_LEVEL_DISGUSTED to INFINITY)
 				msg += "[t_He] look[p_s()] extremely disgusted.\n"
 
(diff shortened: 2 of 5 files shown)


code/datums/traits/negative.dm
```diff
@@ -736,6 +736,15 @@
...
 
+	if(H.reagents.has_reagent(/datum/reagent/medicine/epinephrine))
+		return
+	if(H.reagents.has_reagent(/datum/reagent/medicine/atropine))
+		return
+	if(H.reagents.has_reagent(/datum/reagent/medicine/diphenhydramine))
+		return
+	if(H.reagents.has_reagent(/datum/reagent/medicine/synaphydramine))
+		return
+
```
Moltijoe, line 747: why not just make a list of reagents and loop through the list checking if they have one of them?

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -281,6 +281,9 @@
 			if(DISGUST_LEVEL_DISGUSTED to INFINITY)
 				msg += "[t_He] look[p_s()] extremely disgusted.\n"
 
+	var/datum/quirk/allergic/allergen = has_quirk(/datum/quirk/allergic)
+	if((allergen && reagents?.has_reagent(allergen.reagent_id)) || reagents?.has_reagent(/datum/reagent/toxin/histamine))
```
Moltijoe, line 285: could also reference the cooldown that the quirk uses to track allergic reactions

code/datums/traits/negative.dm
ynot01: seems overengineered

code/modules/mob/living/carbon/human/examine.dm
ynot01: that would be strange and inconsistent use of a variable that is not meant for that

ToasterBiome (approved):

Issue #21183, opened 2023-12-28
 you can have mutations and augments as a golem if you transform into one through a chaplain ritual and previously had them


Yogbot-13:
## Test Merges: 
Not Supplied / None
## Reproduction:
the latter has most likely been reported already but probably not the former, this probably applies to other ways of becoming a golem, such as through golem mutation toxin
hilarity ensues as i pair this with thick skin and become comically impervious hides-in-lockers...

 Submitted by: yarinoi (718841144351522967)

PR #22974, opened 2025-01-31
Changes some parent calls to be cleaner, fixes the thick skin mutation not having proper safety checks

Gboster-0:

Made Thick skin properly check if it was actually added to the DNA before giving its buffs, same with it checking for removal

Changed some `if(..())`'s with `. = ..(); if(.)` since its a bit cleaner and if any new mutations get coded in that are children of previous mutations, that can cause the children to always return false on their checks regardless if they were removed or not (not great)

Only touched yogstation files with the above change, as it isn't too big of a deal most of the time, rather its a neat convinience

Small bug fix gud

# Changelog

:cl:
bugfix: Thick skin now properly checks if its actually removed before removing its effects
/:cl:

### DIFF
--- a/code/datums/mutations/body.dm
+++ b/code/datums/mutations/body.dm
@@ -254,12 +254,16 @@
 
 /datum/mutation/human/thickskin/on_acquiring(mob/living/carbon/human/owner)
 	. = ..()
+	if(.)
+		return
 	if(owner.physiology)
 		owner.physiology.brute_mod *= 0.8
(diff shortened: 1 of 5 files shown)


Oblisk234:
Does this fix #21183, and if so, can you add "Fixes #21183" to your PR description?

Gboster-0:
> Does this fix #21183, and if so, can you add "Fixes #21183" to your PR description?

I got absolutelly no clue

ToasterBiome (approved):

PR #22976, opened 2025-02-02
[All Maps except Meta] Gravity Generator room is no longer bolted by default

Aquizit:

The door to the gravity generator no longer is bolted at the start of a shift, and anyone with the authorized access can use it.

Meta doesn't have a generator since it's planetside, so no change there.

When people want to restart the gravity generator machine after a failure event, it's annoying for engineers to not be able to get in because it's bolted and requiring the CE's remote to open.

# Testing

image

# Changelog

:cl:
tweak: Gravity Generator room no longer bolted at beginning of shift
mapping: Lock helper removed on grav gen door.
/:cl:

### DIFF
--- a/_maps/map_files/DonutStation/DonutStation.dmm
+++ b/_maps/map_files/DonutStation/DonutStation.dmm
@@ -17878,6 +17878,28 @@
 /obj/structure/sink/puddle,
 /turf/open/floor/grass,
 /area/tcommsat/computer)
+"hpf" = (
+/obj/machinery/door/airlock/highsecurity{
+	name = "Gravity Generator"
+	},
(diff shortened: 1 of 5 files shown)


AMyriad (changes requested):
What about manateestation?

Aquizit:
> What about manateestation?

oh is it actually real? I saw it there and went "hmm, it that just asteroid renamed? I don't think that's in rotation I have not seen it"

AMyriad (approved):
Good change, less hassle for engineering staff
might be worth removing the metal covering inside the airlock though idk

> I don't think that's in rotation

yea it do be like that ;-; (it rarely rolls due to low pop)

